PR #697, opened 2018-01-10
Add mention of UK sort code plugin

SavageCore:
Not sure if you do maintain the plugin list but here you go :slightly_smiling_face:


gschier:
Awesome, thanks for this 😄

PR #711, opened 2018-01-15
Resolve symlinks when accessing a DB file - fixes #710

### LINKED ISSUE #710: [Feature Request] Allow application data folder to be a symlink
- Insomnia Version: 5.12.4.1752
- Operating System: macOS Sierra 10.12.6

## Details

I recently changed laptop and I had a lot of applications settings backed-up using [Mackup](https://github.com/lra/mackup). It works by actually storing the applications settings in Dropbox, iCloud or similar and adding a symlink to the Dropbox folder where the app is looking for them.

I was [trying to add support for Insomnia](https://github.com/lra/mackup/issues/1115), but when I tested I realised that the app crashes if I create such a symlink:

screen shot 2018-01-15 at 18 56 21

Would it be possible to make sure the paths are resolved when accessing DBs? I guess it would involve changing [`getDBFilePath()`](https://github.com/getinsomnia/insomnia/blob/549ce23ce8cf202d787ca4df5efc5f25e03311ed/packages/insomnia-app/app/common/database.js#L27-L32) to call [`realPath`](https://nodejs.org/docs/latest/api/fs.html#fs_fs_realpath_path_options_callback)?

browniebroke:
This would enable the user data to live somewhere else and be back-up in a cloud storage. A use case for that is using [Mackup](https://github.com/lra/mackup).

### DIFF
--- a/packages/insomnia-app/app/common/database.js
+++ b/packages/insomnia-app/app/common/database.js
@@ -4,6 +4,7 @@ import * as models from '../models/index';
 import electron from 'electron';
 import NeDB from 'nedb';
 import fsPath from 'path';
+import fs from 'fs';
 import {DB_PERSIST_INTERVAL} from './constants';
 import uuid from 'uuid';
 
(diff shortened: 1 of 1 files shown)


browniebroke:
Not needed, see the issue referenced.

PR #721, opened 2018-01-19
Replace time and size with URL in history dropdown (#677)

### LINKED ISSUE #677: [Improvement] Add request url in history preview list
In the history preview list (see screenshot below), it would be great to add the request URL (on hover, now that there's no enough real-estate).

screen shot 2017-12-22 at 2 56 03 pm

I always have to click on a number of items before getting the actual request from history.

Let me know if I need to provide any more information.

-na

iansu:
Closes #677

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.js
@@ -2,9 +2,8 @@ import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import {Dropdown, DropdownButton, DropdownDivider, DropdownItem} from '../base/dropdown';
-import SizeTag from '../tags/size-tag';
 import StatusTag from '../tags/status-tag';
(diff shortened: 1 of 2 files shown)


iansu:
Thank you. I use Insomnia regularly so happy to contribute. 😀

ProfNandaa:
Thank you for pulling this off @iansu 👏

roboslone:
This actually breaks my workflow, can this behaviour be optional?

gschier:
Hey @roboslone. What's your workflow? I'd be open to making it configurable, but maybe there's a way to modify the new interface to satisfy your workflow.

gschier:
Probably best to discuss this in a new issue so it doesn't get lost. Do you want to create one?

roboslone:
@gschier my workflow is that I use Insomnia to test backend handlers that I'm currently working on. I make changes on backend, send a request, then make some other changes. After several iterations I need response times to determine which changes made it lower. Seeing response times in history is really useful in that case.

gschier:
Thanks for the info. I actually just added those tags back so your workflow should be back to normal in the next release.

image

PR #731, opened 2018-01-29
Added functions to get and set parameters via plugins

luveti:


gschier:
Added a fully-implemented solution with tests in https://github.com/getinsomnia/insomnia/commit/dc0695de23f3c2b3a1d9086d1c972ca03313fdd6

PR #762, opened 2018-02-14
[Feature] Save request body when switching to 'No Body' (#752)

### LINKED ISSUE #752: [Feature Request] Switch between actions/body types without losing data
I'd like to be able to switch between PATCH/POST/GET etc. requests within a single url without hassle.

My workflow is:

1) I submit a PATCH request, I select JSON body and enter the data.
2) I switch to GET, but it says 400 Bad Request because there's a body present. 
3) I change the body type to No Body and it warns me that it's about to delete my existing body data. 
4) I cancel out of the Switch Body Type dialog, copy my json data somewhere else to not lose it.
5) Switch body type to No Body, change to GET, perform GET action.
6) Switch back to PATCH, select JSON, find my notepad window with the json data and paste it in.

What I'd like my workflow to be is:

1) I submit a PATCH request, I select JSON body and enter the data.
2) I switch to GET. The system recognises that you cannot submit a body with a GET, so it greys out the Body tab and does not attempt to submit my json data.
3) I switch back to POST/PATCH/something with a body. The body tab ungreys and becomes functional again, with my previous json data still there.

Or at the very least:

1) I submit a PATCH request, I select JSON body and enter the data.
2) I switch to GET, but it says 400 Bad Request because there's a body present. 
3) I change it to No Body, and the system does so without warning me because it does not delete my body data.
4) Switch back to PATCH, select JSON. My json data is still there from before.

Please consider whether or not either of my suggested workflows can/should be implemented. Thank you.

Edit: I've just noticed that it is possible to send a body with any request, so perhaps my first option won't work. But my second option - switching to No Body doesn't erase the body but just hides it instead - could still fix this problem.

alan-seymour:
Closes #752 
insomnia-saved-body

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/content-type-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/content-type-dropdown.js
@@ -40,10 +40,9 @@ class ContentTypeDropdown extends React.PureComponent<Props> {
 
     const willBeFile = mimeType === CONTENT_TYPE_FILE;
     const willBeMultipart = mimeType === CONTENT_TYPE_FORM_DATA;
-    const willBeEmpty = typeof mimeType !== 'string';
     const willBeGraphQL = mimeType === CONTENT_TYPE_GRAPHQL;
 
-    const willConvertToText = !willBeGraphQL && !willBeFile && !willBeMultipart && !willBeEmpty;
+    const willConvertToText = !willBeGraphQL && !willBeFile && !willBeMultipart;
     const willPreserveText = willConvertToText && isText;
     const willPreserveForm = isFormUrlEncoded && willBeMultipart;
 
--- a/packages/insomnia-app/app/ui/components/wrapper.js
+++ b/packages/insomnia-app/app/ui/components/wrapper.js
@@ -82,6 +82,7 @@ type Props = {
   handleSetRequestGroupCollapsed: Function,
   handleSendRequestWithEnvironment: Function,
   handleSendAndDownloadRequestWithEnvironment: Function,
+  handleUpdateRequestMimeType: Function,
 
   // Properties
   loadStartTime: number,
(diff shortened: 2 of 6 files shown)


packages/insomnia-app/app/ui/components/dropdowns/content-type-dropdown.js
```diff
@@ -40,10 +40,9 @@ class ContentTypeDropdown extends React.PureComponent<Props> {
 
     const willBeFile = mimeType === CONTENT_TYPE_FILE;
     const willBeMultipart = mimeType === CONTENT_TYPE_FORM_DATA;
-    const willBeEmpty = typeof mimeType !== 'string';
     const willBeGraphQL = mimeType === CONTENT_TYPE_GRAPHQL;
 
-    const willConvertToText = !willBeGraphQL && !willBeFile && !willBeMultipart && !willBeEmpty;
+    const willConvertToText = !willBeGraphQL && !willBeFile && !willBeMultipart;
```
gschier: Can you explain what you did here?

packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -599,7 +585,7 @@ class Wrapper extends React.PureComponent<Props, State> {
             updateRequestParameters={this._handleUpdateRequestParameters}
             updateRequestAuthentication={this._handleUpdateRequestAuthentication}
             updateRequestHeaders={this._handleUpdateRequestHeaders}
-            updateRequestMimeType={this._handleUpdateRequestMimeType}
+            updateRequestMimeType={handleUpdateRequestMimeType}
```
gschier: Nice refactor 👍

packages/insomnia-app/app/ui/components/dropdowns/content-type-dropdown.js
alan-seymour: This was primarily to remove the alert modal that the body would be lost when changing to 'No Body'.

My understanding is that `willBeEmpty` is only true when switching to No Body, that would make `willConvertToText` false, which in turn makes `willPreserveText` false, causing the modal to display.

Based on the naming of the variables, it might have made more sense to add an additional case to the `willPreserveText` calculation, however the `willBeEmpty` check isn't used anywhere else so I decided to remove it rather than just have it negate itself later.

gschier: Got it. Thanks for the explanation! In the past, there have been some minor bugs related to this so I just wanted to make sure. I'm fine merging this in but I'll do some testing (keeping your explanation in mind) after merging.

PR #768, opened 2018-02-17
Ability for plugins to define custom response viewers

gschier:
## To-Do

- [x] Search plugins when loading response viewers
- [ ] Add new `RequestMeta.previewModeChild` 
  - We need this so that the app can support multiple preview modes. This also gives us a way to fallback to the regular `previewMode` if the plugin is uninstalled a new response is returned with an unsupported content type.
- [ ] Publish a sample plugin somewhere
- [ ] ...

## Future Tasks

- [ ] Move reusable Insomnia component into separate library so they can be used within plugins
- [ ] Show child preview modes in sub menu instead of on same level as others
  - [ ] Implement sub menu support in `Dropdown` component

## Sample Plugin

Example plugin that shows responses in all UPPERCASE!

```js
const React = require('react');

module.exports.responseViewers = [{
  match: /json/,
  name: 'Uppercase',
  previewMode: 'preview',
  previewModeChild: 'uppercase',
  component: class extends React.Component {
    render () {
      return (
        <div className="pad wide tall scrollable">
          <pre className="force-pre-wrap">
            {this.props.bodyBuffer.toString('utf8').toUpperCase()}
          </pre>
        </div>
      );
    }
  }
}];
```

The new response type will show up in the dropdown when it matches a supported content type.

image


osvaldopina:
The idea of visualization plugins is fantastic and I'd like to know you would be willing to extend it a bit.
I'm working with some hypermedia apis using HATEOAS. In this type of api controls, such as links for example, are included in the response body to assist the client in assembling the next request. HAL is a exemple (https://en.wikipedia.org/wiki/Hypertext_Application_Language).
What do you think of expanding that concept so that response preview plugins can interact with the application?
I was thinking of something like: The plugin would be a reactive component that would generate events for the application. A HAL plugin would render a visual link that when clicked would change the HTTP address in the navigation bar and submit a GET as it was done by a user.

osvaldopina:
@dschulten my idea was to provide to the plugin a reasonable access to insomnia engine, so that a plugin developer could implement both strategies: keep all requests in the same request or create a new one for each followed link. 

I created a simple POC in my fork (https://github.com/osvaldopina/insomnia in feature/plugin-response-viewer-api) and a test plugin (https://github.com/osvaldopina/ResponseViewerTestPlugin, https://www.npmjs.com/package/response-viewer-test-plugin). This plugin consists of a react component that allows the user to change the request url, method and send a request.

Issue #739, opened 2018-02-04
[Feature Request] Fuzzy Search Results in Switch Requests (⌘P) Palette


roblav96:
- Insomnia Version: `Version 5.12.4 (5.12.4.1752)`
- Operating System: `OSX 10.11.5`

## Details
The `Switch Requests` palette should display fuzzy search results on request names, urls, and **folder names**; eg. Sublime Text. I recommend [fuzzy](https://github.com/mattyork/fuzzy) for `Array<Object>` inputs and [bevacqua/fuzzysearch](https://github.com/bevacqua/fuzzysearch) for `Array<string>` inputs. Both provide bare minimum use cases. I would make a PR but for efficiency's sake I'm sure you would be able to implement this far quicker than me since you built this :D

gschier:
Hey @roblav96, thanks for the request. 

The request switcher already performs a fuzzy match, but only on request name. Adding folder as a prefix to match on seems like a great idea though 😄👍 

It's true that I'd be able to implement this fairly quickly but it would end up taking a while because it's not at the top of the priority list. If you wanted to submit a PR for this I'd be happy to provide guidance if needed.

![image](https://user-images.githubusercontent.com/587576/35961061-3b243a4c-0ce7-11e8-875a-b87ff97e1604.png)

axs221:
I would like to take a look at this one. I can take a look at this one maybe this weekend. I love Insomnia, but I was thinking the same thing. Definitely could use URL filtering. Might need better placeholder text to show fuzzy matching is allowed.

axs221:
Progress update. I have basic searching working for folders, URL, querystring, and method. You can search multiple fields by space delimiting the search.

https://github.com/axs221/insomnia/tree/feature/fuzzy-match-request-parameters

Still to do includes evaluating the UI aspect. As in those 3rd party libraries @roblav96 linked, we should consider showing what text is matched, including folder path.

I might split this into two and put up a first PR for discussion with @gschier and implement any UI tweaks in another PR. What I have now is usable, just not as user friendly. But we will need to determine how the UI will need to change to accommodate showing what parts of search matched, especially for fields not currently visible on the Quick Switch window (URL, grandparent folders, querystrings).

gschier:
That sounds great @axs221. Early PR with what you have. Then we can discuss the rest 😀

axs221:
@gschier Sounds like a plan. ^--- Early PR is up! Let me know your thoughts when you get time.

[...]

PR #797, opened 2018-03-05
Filter requests by folder, URI, method, query string

axs221:
Relevant issues?
-----------------

#739 - This fulfills at least some part of the issue. We may want to update the UI to display either what parts of text are matched, and/or show missing fields (e.g. add URL to the Quick Switch modal).

What does this PR do?
---------------------

* Adds a fuzzyMatchAll method to simplify and standardize how multiple-field searches are performed.
* Applies fuzzyMatchAll to request search bars in the Sidebar and in the Quick Switch modal.
* Updates those request search bars to have more descriptive placeholders.
* Adds unit tests for both fuzzyMatch and fuzzyMatchAll.

What is left to do or discuss on the relevant issues?
----------------------------------------------------

* We might want to make Quick Switch window display URI, 2+ folders (all parents), possibly query string. Either multiple rows, and/or show the field that matched. It might be a little much to include many e.g. query string parameters on there, so some things might have to stay hidden, or we have to get creative.
* Should we keep Fuzzy Matching as implemented? An option to turn it off might be nice. I was wanting to search for POST earlier, and it was matching something like "httP://lOcalhoST". I prefer multiple word non-fuzzy searching for this reason. But I assume others prefer fuzzy matching or it wouldn't be there.
* If we keep Fuzzy Matching as implemented, consider whether labels are needed. Probably overkill, but we could add optional labels to the search, like "method:POST url:localhost".

Example screenshots, gifs, or videos?
-------------------------------------

* [Image of an example search in Quick Switch](https://cl.ly/3H3X2a0G1D32/Image%202018-03-04%20at%208.44.57%20PM.png)
* [Image of same search on Sidebar](https://cl.ly/1C2J2Y0d0V00/Image%202018-03-04%20at%208.47.17%20PM.png)
* [Video of various search attempts, searching on URI, folder, etc, in both request search bars](https://cl.ly/2n0Q2d3G2i3P/Screen%20Recording%202018-03-04%20at%2008.37%20PM.mov)

What code can others reuse?
-----------------------------

* fuzzyMatchAll - This calls out to the existing fuzzyMatch misc utility method. The difference is that it takes a string that will be split on spaces for required search terms, and it takes an array of fields that the search terms can match against.

Testing Steps
--------------

* Search for URI, method, query string, folder, and/or request name on Quick Switch window
* Perform fuzzy search, leaving out some letters for a given field
* Perform a multiple word search, searching multiple types of fields in the list above
* Repeat tests for Sidebar search


gschier:
Haven't forgot about this! Should be able to get to it next week 😄

packages/insomnia-app/app/ui/components/modals/request-switcher-modal.js
```diff
@@ -122,24 +122,58 @@ class RequestSwitcherModal extends PureComponent {
     this._handleChangeValue(e.target.value);
   }
 
+  /**
+   * Appends path of ancestor groups, delimited by forward slashes
+   * E.g. Folder1/Folder2/Folder3
+   */
+  _groupOf (request) {
```
gschier: Small thing. This variable would be more accurately described as `child` or `requestOrRequestGroup`. Can you rename to be more clear?

packages/insomnia-app/app/ui/components/modals/request-switcher-modal.js
```diff
@@ -122,24 +122,58 @@ class RequestSwitcherModal extends PureComponent {
...
+  _isMatch (searchStrings) {
+    return (request) => {
+      // Match request attributes
+      const matchesAttributes = fuzzyMatchAll(searchStrings,
+        [
+          request.name,
+          request.url,
+          request.method,
+          this._groupOf(request),
+          ...(request.parameters
```
gschier: It might be more intuitive to join the parameters to the URL? Thoughts?

See https://github.com/getinsomnia/insomnia/blob/cd1ff32b4d098979cc61de6837475f648bda82e3/plugins/insomnia-plugin-request/index.js#L108-L109 for an example of URL building.

gschier:
Responding to your comments in the PR description:

> We might want to make Quick Switch window display URI, 2+ folders (all parents), possibly query string. Either multiple rows, and/or show the field that matched. It might be a little much to include many e.g. query string parameters on there, so some things might have to stay hidden, or we have to get creative.

I agree, it's probably best to show as much as possible. Perhaps we could hide most things by default and only show them if they were part of the match?

> Should we keep Fuzzy Matching as implemented? An option to turn it off might be nice. I was wanting to search for POST earlier, and it was matching something like "httP://lOcalhoST". I prefer multiple word non-fuzzy searching for this reason. But I assume others prefer fuzzy matching or it wouldn't be there.

I agree that the fuzzy search right now is probably too loose, specifically for the sidebar filter. For the request switcher, the current solution would probably be fine if results were sorted by best match like in Atom/Sublime/etc.

> If we keep Fuzzy Matching as implemented, consider whether labels are needed. Probably overkill, but we could add optional labels to the search, like "method:POST url:localhost".

Oh, that's interesting. Like GitHub does with issues. I think labels are a great idea either way because we could add things like status:200, auth:Basic, content-type:json, etc.

Thanks @axs221!

packages/insomnia-app/app/ui/components/modals/request-switcher-modal.js
axs221: Sounds good, I'll go with requestOrRequestGroup, verbose but clear as what can be expected

packages/insomnia-app/app/ui/components/modals/request-switcher-modal.js
axs221: Yeah that makes sense, I will change it as suggested

axs221:
@gschier Okay I updated the branch with the minor feedback from the PR, minus the items mentioned in `Responding to your comments in the PR description`.

Is what is there so far worth merging, and then doing a second PR for the rest? Otherwise I should be able to get to the additional items in the next week or so.

Thank you!

gschier:
Yep, it looks great!

gschier:
I was busy for the past couple weeks traveling but will now be more available and responsive to help out if needed 😄

axs221:
@gschier Okay thanks! Hope you had a good time traveling; I've been busy too, no worries. I'll put some thought into the UI changes soon and we can discuss.

gschier:
It was amazing, thanks for asking 😃 

Sounds good! Probably best to create a new issue for discussion of the next phase 👍

gschier:
@axs221 just so you're aware, I'm disabling the new fuzzy matching for URL/parameters until we have some UI to make it more clear what's going on.

PR #820, opened 2018-03-17
change header placeholder from My-Header to Header so it says New Header

benatkin:
When I create a header, it has "New My-Header" for the placeholder:

screen shot 2018-03-17 at 1 09 49 am

This is a bit redundant. I think "My-Header" or "New Header" would both be good, but since the value is "New Value", I think it makes sense to change it to "New Header". Here's what it looks like after the change in this PR:

screen shot 2018-03-17 at 1 10 04 am


gschier:
Awesome. Small but valuable change 😄 

Thanks @benatkin!

PR #822, opened 2018-03-18
Add support for additional claims in ASAP Auth

BCook98:
PR adds field in ASAP Auth to supply additional JWT claims beyond those already required/available.

image

image


gschier:
Sorry for the delay @BCook98. I like it! Thanks 😄

PR #837, opened 2018-03-29
Move folders and requests to new workspaces

### LINKED ISSUE #403: [Feature Request] Move requests between workspaces
## Overview
- `Insomnia Version:` _5.7.4_
- `Summary: ` _I would like to be able to move requests from one workspace to another_

I spread lots of requests that actually belong to the same project over different workspaces because I was a little bit stupid. Now I find it awkward that I have to switch workspaces so often.

I would like to consolidate the requests belonging to the project in one workspace but I can't find any way to do it. As far as I can see, this is not a provided use case in insomnia so far and using export/import isn't an option either because you can only export and import whole workspaces but not requests or folders of requests.

gschier:
Closes #403

image

Also an option to move or copy individual requests from the request settings dialog:

image


njleonzhang:
when will this be released?

gschier:
Not sure. Either this week or next

gschier:
I might put it up this week as a beta release

PR #863, opened 2018-04-15
Add Swagger to import format lists

scop:


gschier:
Thanks for adding this!

Issue #912, opened 2018-05-01
[Feature Request] Support for scoped npm packages for plugins


skang0601:
- Insomnia Version: 5.16.1 (5.16.1)
- Operating System: Mac OS 10.13.4

## Details

Insomnia's plugin system does not support scoped npm packages as defined here: https://docs.npmjs.com/getting-started/scoped-packages. 
I'd like for Insomnia to recognized scoped directories in the root of the plugin directory, aka directories that start with "@" and read all the plugins found in that directory.

gschier:
Hey @skang0601 

As far as I know, there's nothing preventing scoped npm modules from working in Insomnia currently.

Insomnia scans the `plugins/` location for folders containing a `package.json` and loads those into the app. The ONE requirement is that the `package.json` MUST include an `insomnia` key with at least a `name` attribute. Here's an example:

```js
{
  "name": "@org/module",
  "version": "1.0.0",
  // ...
  "insomnia": {
    "name": "my-plugin"
  }
  // ...
}
```

Perhaps your scoped module's `package.json` files are missing the `insomnia` property? If you can provide a specific example that would help a lot.

skang0601:
The problem is due to the directory structure of scoped packages. As seen here: https://docs.npmjs.com/misc/scope:
`packages are installed to a sub-folder of the regular installation folder, e.g. if your other packages are installed in node_modules/packagename, scoped modules will be installed in node_modules/@myorg/packagename. The scope folder (@myorg) is simply the name of the scope preceded by an @ symbol, and can contain any number of scoped packages.`

The scope directory doesn't contain a package.json and Insomnia doesn't further traverse down the scope directory.

gschier:
Ok, I understand now. Thanks for clarifying that 😀

So maybe Insomnia should traverse down the tree and stop at the first directory that contains a package.json?

skang0601:
Yep that's exactly what I had in mind. I started on a fix for this a little while ago, I'll have a PR up for you to review most likely tomorrow.

gschier:
Awesome! Let me know if you need any help finishing that up 👍

skang0601:
Let me know if you see anything funky with that PR. It solves the issue I had when running a local build.

gschier:
Oops, this should have been closed by #914

PR #914, opened 2018-05-02
Traversing further into scope directories

skang0601:
Related #912 

Just a little tweak to recursive traverse into the scop directories. I'm pretty unfamiliar with modern JS development so let me know if I'm doing anything egregious. 

I've also noticed that the tests related to the AWS auth headers are failing locally, not sure why that's the case.


gschier:
FYI, those tests were fixed in a later commit 👍

PR #979, opened 2018-06-08
Closes 959 Fix "Copy Curl" to respect "encode url" option

besolov:


packages/insomnia-app/app/common/har.js
```diff
@@ -360,7 +360,8 @@ export async function exportHarWithRenderedRequest (
     queryString: getRequestQueryString(renderedRequest),
     postData: getRequestPostData(renderedRequest),
     headersSize: -1,
-    bodySize: -1
+    bodySize: -1,
+    encode: renderedRequest.settingEncodeUrl
```
gschier: Since the `encode` property isn't officially part of the HAR spec, it might make more sense to prefix this like we do with request settings. 

I vote to keep it consistent and use the same name `settingEncodeUrl`.

gschier: P.S. you'll need to add the new property to the Flow type as well

https://github.com/getinsomnia/insomnia/blob/cd1ff32b4d098979cc61de6837475f648bda82e3/packages/insomnia-app/app/common/har.js#L59-L70

packages/insomnia-httpsnippet/src/index.js
```diff
@@ -174,7 +174,9 @@ HTTPSnippet.prototype.prepare = function (request) {
 
   // update the uri object
   request.uriObj.query = request.queryObj
-  request.uriObj.search = qs.stringify(request.queryObj)
+  request.uriObj.search = qs.stringify(request.queryObj, '&', '=', {
```
gschier: It took me a while to figure out what this was doing but it looks good! 👍

besolov: Do you want me to simplify it somehow?

gschier: If you can, but not a big deal. I can't think of anything obvious that would simplify it.

besolov: Me too, actually(

packages/insomnia-app/app/common/har.js
besolov: Done

PR #990, opened 2018-06-17
Run GraphQL query based on cursor position.

bwlt:
If there are multiple operations in the GraphQL query it automatically choose
which operation to run based on the cursor position.


alessiodionisi (approved):

gschier:
This is a very well done PR, thanks! 

I will also leave some actual review comments but I just wanted to have a brief high-level discussion about this feature first.

While I do think this feature is useful (and I know it's part of GraphiQL), it seems like there is potential to cause confusion since there's no obvious indication that the cursor position is controlling the selected query.

A quick improvement could be to highlight the selected query name based on the currently active `OperationName`. I'm perfectly happy making this (or similar) improvement after merging this PR. However, if you felt like adding this yourself that would be awesome too.

Thoughts?

bwlt:
Ok, how to highlight? (in terms of style)
I thought on these two alternatives:

- the selected query have syntax color
screen shot 2018-06-19 at 14 16 11

- the selected query have a **bolder** operation name
screen shot 2018-06-19 at 14 16 35

(I'm not good on design stuff 😅)

timrs2998 (approved):

gschier:
Ooo I like grayscaling everything else but I'm not sure how difficult that is to implement. If it's not easy, I would also be fine with using Codemirror's `markText` function to highlight the background of the query name.

Here's a code snippet of how it works.

```js
editor.markText({line: 6, ch: 26}, {line: 6, ch: 42}, {className: "styled-background"});
```

Setting up the precise colors to use may be tricky since it has to work with the Insomnia theme system. But, if you want to add the basic implementation I can help fine-tune it if needed.

gschier:
Docs, if needed https://codemirror.net/doc/manual.html#api_marker

bwlt:
I have worked on it using codemirror textmarker.
The result is nice! Using graphql documentAST helped a lot.
Let me work on a little issue introduced in this PR; a little refactor is
necessary :)

Il mer 20 giu 2018, 05:30 Gregory Schier <notifications@github.com> ha
scritto:

gschier:
Awesome! I'm looking forward to it 😄 

Let me know if you need any help! 👍

packages/insomnia-app/app/ui/css/components/graph-ql-editor.less
```diff
@@ -21,4 +21,10 @@
   .graphql-editor__query .cm-attribute {
     color: var(--color-info);
   }
+
+  &__disabled-definition {
+    &&& {
```
bwlt: @gschier sorry, this is a dirty trick to increase the specificity (always better than `!important`, inspired from [styled-components](https://www.styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity))

jsejcksn:
It seems like this change didn't make it into the docs: https://support.insomnia.rest/article/61-graphql

PR #991, opened 2018-06-17
Feature/flow typing

bwlt:
It's not an easy PR to review; 90% of the changes are typing related.

Critical parts:

1. some additional type checking was necessary:
  [packages/insomnia-app/app/ui/components/base/debounced-input.js](https://github.com/getinsomnia/insomnia/pull/991/files#diff-4fb4586d3c56f850045acebeb624d46eR52) `this._input` can be null; generally all refs must be checked against null.
2. [packages/insomnia-app/app/ui/components/base/modal.js](https://github.com/getinsomnia/insomnia/pull/991/files#diff-e83335652acd76fa9c5aa9e32a012743)
  it gets `currentTarget` instead of `target`.
3. [packages/insomnia-app/app/ui/components/base/prompt-button.js](https://github.com/getinsomnia/insomnia/pull/991/files#diff-12887caa6242d0f56ba56130e468d8c2)
  it is too dynamic and difficult to add correct typings (@see _ask method)
4. [packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js](https://github.com/getinsomnia/insomnia/pull/991/files#diff-2e4876dfdc0d76472c210753b6e85f87)
  same for point 1.

### DIFF
--- a/packages/insomnia-app/app/ui/components/base/copy-button.js
+++ b/packages/insomnia-app/app/ui/components/base/copy-button.js
@@ -16,7 +16,7 @@ type State = {
 
 @autobind
 class CopyButton extends React.PureComponent<Props, State> {
-  _triggerTimeout: void | TimeoutID
+  _triggerTimeout: TimeoutID
 
   constructor (props: Props) {
     super(props);
--- a/packages/insomnia-app/app/ui/components/base/copy-button.js
+++ b/packages/insomnia-app/app/ui/components/base/copy-button.js
@@ -1,18 +1,31 @@
-import React, {PureComponent} from 'react';
+// @flow
+import * as React from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 const {clipboard} = require('electron');
 
+type Props = {
+  content: string | () => Promise<string>,
+  children?: React.Node,
+  title?: string,
+  confirmMessage?: string
+}
+type State = {
+  showConfirmation: boolean
+}
+
 @autobind
-class CopyButton extends PureComponent {
-  constructor (props) {
+class CopyButton extends React.PureComponent<Props, State> {
+  _triggerTimeout: void | TimeoutID
+
+  constructor (props: Props) {
     super(props);
     this.state = {
       showConfirmation: false
     };
   }
 
-  async _handleClick (e) {
+  async _handleClick (e: SyntheticMouseEvent<HTMLButtonElement>) {
     e.preventDefault();
     e.stopPropagation();
 
--- a/packages/insomnia-app/app/ui/components/base/dropdown/dropdown-item.js
+++ b/packages/insomnia-app/app/ui/components/base/dropdown/dropdown-item.js
@@ -1,11 +1,23 @@
-import React, {PureComponent} from 'react';
+// @flow
+import * as React from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import classnames from 'classnames';
 
+type Props = {
+  buttonClass?: React.ElementType,
+  stayOpenAfterClick?: boolean,
+  value?: any,
+  disabled?: boolean,
+  onClick?: Function,
+  children?: React.Node,
+  className?: string,
+  color?: string
+}
+
 @autobind
-class DropdownItem extends PureComponent {
-  _handleClick (e) {
+class DropdownItem extends React.PureComponent<Props> {
+  _handleClick (e: SyntheticMouseEvent<HTMLElement>) {
     const {stayOpenAfterClick, onClick, disabled} = this.props;
 
     if (stayOpenAfterClick) {
--- a/packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js
+++ b/packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js
@@ -65,7 +65,9 @@ class Dropdown extends React.PureComponent<Props, State> {
     if (e.key === 'Enter') {
       // Listen for the Enter key and "click" on the active list item
       const selector = `li[data-filter-index="${this.state.filterActiveIndex}"] button`;
-      if (!this._dropdownList) return;
+      if (!this._dropdownList) {
+        return;
+      }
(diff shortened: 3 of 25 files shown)


packages/insomnia-app/app/ui/components/base/copy-button.js
```diff
@@ -1,18 +1,31 @@
...
+}
+type State = {
+  showConfirmation: boolean
+}
+
 @autobind
-class CopyButton extends PureComponent {
-  constructor (props) {
+class CopyButton extends React.PureComponent<Props, State> {
+  _triggerTimeout: void | TimeoutID
```
gschier: Nitpick but the rest of the codebase has simply been using plain ol' `TimeoutID` for this. Probably best to be consistent.

packages/insomnia-app/app/ui/components/base/dropdown/dropdown-item.js
```diff
@@ -1,11 +1,23 @@
-import React, {PureComponent} from 'react';
+// @flow
+import * as React from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import classnames from 'classnames';
 
+type Props = {
+  buttonClass?: React.ElementType,
```
gschier: Nice, I wasn't aware of this type!

packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js
```diff
@@ -10,9 +11,35 @@ import {fuzzyMatch} from '../../../../common/misc';
 import KeydownBinder from '../../keydown-binder';
 import * as hotkeys from '../../../../common/hotkeys';
 
+type Props = {
+  children: React.Element<any>,
```
gschier: Should this not be `React.Node`?

packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js
```diff
@@ -102,10 +136,10 @@ class Dropdown extends PureComponent {
       }
     }
 
-    this._filter.focus();
+    if (this._filter) this._filter.focus();
```
gschier: Nitpick but this project doesn't use same-line `if` statements (should probably add a lint rule for it). 

Can you change this to

```js
if (this._filter) {
  this._filter.focus();
}
```

😄

packages/insomnia-app/app/ui/components/base/prompt-button.js
```diff
@@ -51,12 +68,12 @@ class PromptButton extends PureComponent {
     }, 2000);
   }
 
-  _handleClick (...args) {
+  _handleClick (e: T) {
```
gschier: This behavior is very hacky and messy, but this needs to be switched back to `...args`. 

This is because `<Button/>` can be passed a `value` attribute which will cause the callback to be called with `_handleClick(value, e)`. 

See `button.js` for the implementation.

https://github.com/getinsomnia/insomnia/blob/05d93aab8ffd90e134d09b507f56d91d9fc191fd/packages/insomnia-app/app/ui/components/base/button.js#L7-L16

packages/insomnia-app/flow-typed/npm/prop-types_v15.x.x.js
```diff
@@ -0,0 +1,35 @@
+// flow-typed signature: d9a983bb1ac458a256c31c139047bdbb
```
gschier: Hmm, defining `prop-type` shouldn't be necessary, should it? If a component is using the `prop-types` package that should mean it's not using Flow. If Flow is introduced in a component, the use of `prop-types` should be removed.

Thoughts?

bwlt: Because if a component A (which is not flow-typed) use a component B (which is flow-typed)
the type system does not helps A how to pass props to B; so it can fallback to PropTypes until component A (and the rest of the codebase) is fully covered with flow-type.
I know that is difficult to align all the codebase to a convention, but I think that this is the better experience for a developer in terms of safety of the code.
This is a philosophical issue :)
Let me know if you consider it a good way to proceed

packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js
bwlt: It's hard, because of some logic inside of the render method:
- [line 381](https://github.com/bwlt/insomnia/blob/7f2d72982f8a3d2a61e4c7aa36866a068cfab19e/packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js#L381) children are converted to an array (the typing here is additional and can be inferred)
- [line 382](https://github.com/bwlt/insomnia/blob/7f2d72982f8a3d2a61e4c7aa36866a068cfab19e/packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js#L382) children are flattened
- [line 385](https://github.com/bwlt/insomnia/blob/7f2d72982f8a3d2a61e4c7aa36866a068cfab19e/packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js#L385) children are treated as an object, so it does not match with a React.Node which is declared in this manner: 
screen shot 2018-06-19 at 14 41 22

gschier: Got it, that makes sense. I'm fine leaving it like this. That component is unlikely to change much in the future so it's fine if it's a bit, uh, different 😉

packages/insomnia-app/flow-typed/npm/prop-types_v15.x.x.js
gschier: Ah yes, that makes sense. I'm fine with it then 👍

packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js
bwlt: ok. The eslint rule to add is `"curly": ["error", "all"]`.
It produce various lint errors. better not to add it in this PR

PR #1036, opened 2018-07-05
Add Secure Text option for prompts

### LINKED ISSUE #962: [Feature Request] Prompt for secrets
- Insomnia Version: 5.16.6
- Operating System: macOS

## Details
Support prompts that mask user input when typing to better support prompts for username/password type fields. The prompts currently work great for username/password fields except that the user has to type the password in plain text on their screen. A "Secure Text" option when editing the prompt tag, that would then mask the input like a normal password field, would make the prompt safer to use for secure info.

flipxfx:
Add support for "Secure Text" option for prompts.

Closes #962

### DIFF
--- a/plugins/insomnia-plugin-prompt/index.js
+++ b/plugins/insomnia-plugin-prompt/index.js
@@ -40,7 +40,7 @@ module.exports.templateTags = [
           'something else.'
       },
       {
-        displayName: 'Secure Text',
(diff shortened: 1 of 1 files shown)


flipxfx:
Ok I can change that. Would "Mask Text" work?

gschier:
Ya, that sounds good!

PR #1039, opened 2018-07-09
Added plugin cookie jar for Issue #1028

### LINKED ISSUE #1028: [Feature Request] Allow access to cookie value from other domains
- Insomnia Version: 5.16.6
- Operating System: Windows 10 64-bit

## Details
Currently, there is no way to use a cookie value from a different domain. You are only able to use cookies from the URL of your request. 

One of our applications uses javascript from the site with the cookie and provides a different backend service with these cookie values via a header.

coolhome:
Closes #1028


plugins/insonmia-plugin-cookie-jar/index.js
```diff
@@ -0,0 +1,65 @@
...
+      const cookieJar = await context.util.models.cookieJar.getOrCreateForWorkspace(
+        workspace
+      );
+
+      const cookies = cookieJar.cookies;
+
+      // Validate cookie domain
+      let cookiesWithMatchedDomain = cookies.filter(cookie => cookie.domain === cookieDomain);
+
+      if(cookiesWithMatchedDomain.length === 0) {
```
gschier: What do you think of changing this to use the same matching logic as the Request => Cookie plugin? This would allow the user to also match on URL path and would keep the behavior consistent.

https://github.com/getinsomnia/insomnia/blob/528a8c3b2b8c9eeb368a1087aa74e76ec1d7e931/plugins/insomnia-plugin-request/index.js#L175-L199

gschier: Also note, this would mean changing the "Cookie Domain" parameter to accept a more flexible "URL"

coolhome: @gschier I did not realize the requested plugin did this. I am now using the same function. 

:)

PR #1042, opened 2018-07-12
Add bar color indicator for environments

### LINKED ISSUE #995: [Enhancement] Ability to better visually distinguish environments
Insomnia supports coloring the environment, however that creates a tiny dot that i think can be overlooked pretty easily.

Could you add an option for the color to be used more visibly?

tippl:
Add a bar color indicator along with new a setting to choose whether to use the original dot or the new bar.

Closes #995

### DIFF
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -26,7 +26,7 @@ type BaseSettings = {
   deviceId: string | null,
   updateChannel: string,
   updateAutomatically: boolean,
-  colorIndicatorType: string
+  environmentHighlightColorStyle: string
 };
 
 export type Settings = BaseModel & BaseSettings;
@@ -60,7 +60,7 @@ export function init(): BaseSettings {
     deviceId: null,
     updateChannel: UPDATE_CHANNEL_STABLE,
     updateAutomatically: true,
-    colorIndicatorType: 'dot'
+    environmentHighlightColorStyle: 'dot'
   };
 }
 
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/models/settings.js
```diff
@@ -25,7 +25,8 @@ type BaseSettings = {
   nunjucksPowerUserMode: boolean,
   deviceId: string | null,
   updateChannel: string,
-  updateAutomatically: boolean
+  updateAutomatically: boolean,
+  colorIndicatorType: string
```
gschier: Should be renamed to be more descriptive. Perhaps `environmentHighlightColorStyle`?

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -144,6 +144,19 @@ class General extends React.PureComponent<Props> {
           </label>
         </div>
 
+        <div className="form-control form-control--outlined pad-top-sm">
+          <label>
+            Color Indicator Type
```
gschier: Same here. This should be more descriptive. Like "Environment Highlight Color Style." Could also add a help tooltip (see Nunjucks setting for example) because many will not even know what environment colors are.

packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -423,6 +423,13 @@ class Wrapper extends React.PureComponent<Props, State> {
     const rows = `minmax(0, ${paneHeight}fr) 0 minmax(0, ${1 - paneHeight}fr)`;
 
     return [
+      activeEnvironment &&
+      activeEnvironment.color &&
+      settings.colorIndicatorType === 'bar' ? (
+        <div
```
gschier: This is probably better suited for the `#wrapper` element (below).

A style of `border-top: 5px solid <COLOR>` should do the trick.

PR #1054, opened 2018-07-20
Add bottom bar env highlight color style

tippl:
Had to add this one to the grid, because border appeared outside of the
viewport.

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper.js
+++ b/packages/insomnia-app/app/ui/components/wrapper.js
@@ -584,25 +584,25 @@ class Wrapper extends React.PureComponent<Props, State> {
           borderTop:
             activeEnvironment &&
             activeEnvironment.color &&
-            settings.environmentHighlightColorStyle === 'bar'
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -580,10 +580,29 @@ class Wrapper extends React.PureComponent<Props, State> {
...
           borderTop:
             activeEnvironment &&
             activeEnvironment.color &&
             settings.environmentHighlightColorStyle === 'bar'
+              ? '5px solid ' + activeEnvironment.color
+              : null,
+          borderBottom:
+            activeEnvironment &&
+            activeEnvironment.color &&
+            settings.environmentHighlightColorStyle === 'bar-bottom'
```
gschier: I think it might be a bit more future-proof to have a more specific naming scheme. Something like this maybe?

```bash
window-top
window-bottom
window-right
window-left
sidebar-edge
sidebar-indicator   # Best I could come up with for this one
```

PR #1130, opened 2018-08-22
Added support for file extension filtering as well as ability to select a directory to file picker dialog

nickhomme:
I've added support for the file input to have a file type filter applied as well as the ability to select directories if allowed. 

Reasoning for this is to provide plugins the ability to have more complex capabilities such as looping through files in a directory or only allowing certain files to be selected.

Possible improvements would be to also allow option of multiple files to be selected.


nickhomme:
Test Failed. Closing pull request. Will create another with fixed test, my apologies.

PR #1136, opened 2018-08-24
[Documentation] Change issues link on README

xeptore:
hi.
i think it should refer to `issues` page or it might be `](../../issues/new` in order to _open a new issue_

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ from the website.
 
 Have a bug or a feature request? First, read the 
 [issue guidelines](CONTRIBUTING.md#using-the-issue-tracker) and search for existing and 
-closed issues. If your problem or idea is not addressed yet, [please open a new issue](/issues).
+closed issues. If your problem or idea is not addressed yet, [please open a new issue](../../issues).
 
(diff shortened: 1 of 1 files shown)


gschier:
Where is the link you're referring to?

gschier:
A relative link seems less reliable because it depends which page you're viewing the rendered markdown from.

Example: If I'm viewing the README from here, the relative link will 404
https://github.com/getinsomnia/insomnia/blob/34894563a859b31b73926207ee518e384a72c6f5/README.md

What is the reason you don't want it to be absolute?

gschier:
My first thought was to make it the full URL `https://github.com/getinsomnia/insomnia` but I wanted it to still work if the repo name changed. Perhaps the full URL would be the most reliable though.

xeptore:
yes, youre right. but i thought because README.md file is always at the root level of project files it would be easier to fix where it refers with a `../` instead of making it an absolute link. but as you said its not a good idea when reading that file from commits or somewhere else, so now i think an absolute link is a better idea in total.

PR #1170, opened 2018-09-23
[Feature Request] - Make URLs in headers clickable #1050

### LINKED ISSUE #1050: [Feature Request] - Make URLs in headers clickable
- Insomnia Version:
>Version 5.16.6
>Shell 1.8.4
>Node 8.2.1
>V8 5.9.211.38
>Architecture x64
- Operating System: 
>Ubuntu 18.04

## Details

Hello. I'm using Symfony to build my API, and Symfony uses Headers to set a link [to the debug panel](https://symfony.com/blog/new-in-symfony-2-4-quicker-access-to-the-profiler-when-working-on-an-api).

So, it would be awesome to be able to click on such value.

image

KarthikSiddarth:
"Fixes #1050"
## Tasks
[x] Write a function that parses the given string and returns true if it is a valid URL or false.

[x] Place it in the insomnia-url package directory.

[x] Import the function to the response-header-viewer.

[x] Pass the string to the function, if the function returns True, wrap the value with an anchor tag and assign the value to the href attribute and text content property of the element.

### Specs of the implementation

1. Function considers only absolute URL
2. It should have a valid URL scheme, hostname, pathname

### Things considered to find the parsed URL is valid

1. Check if the protocol is one of the valid application layer protocol.
2. Check if the hostname property has a non-empty string.

### Validating specs:
1. A Hostname is should be a valid domain name, Ipv4 address, Ipv6 address.
2. A URI scheme must be one of the following (ftp, file, gopher, http, https, ws, wss)

### Reasons
1. Reason for not considering relative URL is, we can't be sure that the resource is available at the base URL of the request, it can be in any location.

Resources referred:
1. https://url.spec.whatwg.org/
2. https://softwareengineering.stackexchange.com/questions/294588/what-is-a-uri-scheme


KarthikSiddarth:
[X]  Wrote tests for the functions that validate URL, hostname, scheme
Things to be completed:
[ ] Validate shortened IPv6 URL
[ ] Validate file paths for "file:" scheme (Unix and windows)
[ ] Validate URL that comes with a port number attached.

Test Screenshot:
Response Header viewer

KarthikSiddarth:
Decided to re write the module, will just be using URL method of 'url' module in node. Also will write test for the functions implemented. Will be back soon.

Issue #1173, opened 2018-09-24
[Feature Request] EDN content-type in requests and responses


JoelSanchez:
I'm a Clojure developer, and I'd love to have EDN support in Insomnia. EDN is like JSON but it's a subset of the Clojure language.

![image](https://user-images.githubusercontent.com/8020017/45970358-95df3380-c036-11e8-9a14-94eefc416bea.png)
As you can see, I am suffering.

Thank you

gschier:
It looks like this might be possible with the EDN Codemirror plugin https://github.com/halohalospecial/codemirror-elixir-mode

JoelSanchez:
Hey, I got this. Turns out CodeMirror's support for Clojure is already there...
![image](https://user-images.githubusercontent.com/8020017/46002564-b6e36b00-c0ae-11e8-840b-32d1becc62f8.png)
Now I just need a prettifier to fix the whitespace and indentation.

JoelSanchez:
Can I be assigned to finish this?

JoelSanchez:
#1176 merged!

gschier:
WoooO!

PR #1176, opened 2018-09-25
EDN support

gacelita:
Solves #1173 
Adds EDN to the list of possible request types.
image
Enables EDN support in CodeMirror and prettifies the code
image
Adds the `application/edn` mime type
image

### DIFF
--- a/packages/insomnia-app/package-lock.json
+++ b/packages/insomnia-app/package-lock.json
@@ -1,31 +1,26 @@
 {
-  "name": "insomnia-app",
-  "version": "1.0.40",
-  "lockfileVersion": 1,
   "requires": true,
+  "lockfileVersion": 1,
   "dependencies": {
     "7zip-bin": {
       "version": "3.1.0",
       "resolved": "https://registry.npmjs.org/7zip-bin/-/7zip-bin-3.1.0.tgz",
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import CodeMirror from 'codemirror';
+import 'codemirror/mode/clojure/clojure';
```
gschier: This should be in `./base-imports` 😄

packages/insomnia-app/package.json
```diff
@@ -152,7 +152,8 @@
     "tough-cookie": "^2.3.1",
     "uuid": "^3.0.0",
     "vkbeautify": "^0.99.1",
-    "whatwg-fetch": "^2.0.1"
+    "whatwg-fetch": "^2.0.1",
+    "zprint-clj": "^0.2.0"
```
gschier: `package-lock.json` needs to be included in this PR as well. I think NPM 5+ generates `package-lock.json` when `npm install` is run 👍

JoelSanchez: So, I added the package-lock.json, but it seems to be doing more changes than needed. Should I just commit the line relevant to zprint?

JoelSanchez:
Did a rebase.

JoelSanchez:
Thank you. I think many Clojure devs will turn to Insomnia when/if they need to use an EDN api (starting by my coworkers...)

PR #1179, opened 2018-09-25
Ordered workspace list

### LINKED ISSUE #1161: [Feature Request] Ordered workspace list
- Insomnia Version: 6.0.2
- Operating System: Arch Linux

## Details

Hi there!

I would like to request a feature to always show the "Switch Workspace" list ordered. For each microservice we have a dedicated workspace, and since the workspace list got big, it became hard to find where is the workspace I need, because the list isn't ordered.

I think an option on preferences to enable this feature would be nice.

Best regards,
Rafael Pacheco.

Masov:
Closes #1161 

Added `Ordered Workspace List` checkbox in settings. When the checkbox is selected, workspaces on the list are sorted.

### DIFF
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -26,8 +26,7 @@ type BaseSettings = {
   deviceId: string | null,
   updateChannel: string,
   updateAutomatically: boolean,
-  environmentHighlightColorStyle: string,
-  isWorkspaceListOrdered: boolean
+  environmentHighlightColorStyle: string
 };
 
 export type Settings = BaseModel & BaseSettings;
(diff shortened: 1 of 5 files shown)


ryanprior:
Is there any practical reason to turn off workspace sorting? If not, maybe just make this the standard behavior with no checkbox.

Masov:
The only reason why I added the checkbox is that otherwise, the default behaviour would be changed and maybe someone is already accustomed to having this particular workspace on this specific position.

gschier:
I agree with @ryanprior. I think we can just turn this on by default and remove the setting (nice work on that though by the way 👍).

Further down the road, I think the ideal would be to allow the user to order them with drag-n-drop.

Masov:
It is now turned on by default.

PR #1186, opened 2018-09-30
Validate url

### LINKED ISSUE #1050: [Feature Request] - Make URLs in headers clickable

KarthikSiddarth:
Closes #1050

Used NodeJs's URL module, the module provides a class that is implemented by following WHATWG's URL Standard ([Link to the documentation](https://nodejs.org/dist/latest-v8.x/docs/api/url.html#url_the_whatwg_url_api))

When a string is passed as an argument to the constructor, if the stirng is a valid URL, it returns an object, that contains parsed URL (see example below). If the given string is not a valid URL string, then it throws an error stating the given string is an invalid URL.

#### Example for the value returned by WHATWG's URL API:
``` 
URL {
  href: 'http://www.example.com/',
  origin: 'http://www.example.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'www.example.com',
  hostname: 'www.example.com',
  port: '',
  pathname: '/',
  search: '',
  searchParams: URLSearchParams {},
  hash: '' }
```

#### Implementation details of making valid URL clickable in the response header
- Implemented a function in the name of **validateURL**  and placed into **Response-Header-viewer** module.
- When the response header arrives, the value of each header property is passed to the **validateURL** function.
- If the function returns true, then the value is wrapped with **Link** component, the component render the value as clickable link and make the webpage render in the default browser.
- If the function returns false, then the value is rendered as it is.

#### Implementation details of the validateURL module
- The module uses NodeJS's URL module, the URL module provide a class which is implemented by WHATWG's URL standard.
- The module has one function, which takes a string, the string is then passed to URL constructor, if it throws an error, then the function returns false.
- If URL constructor returns an object, then the function checks if the hostname property is a non-empty string, if it is an empty string, it returns false else it returns true.


packages/insomnia-url/src/validateURL.js
```diff
@@ -0,0 +1,13 @@
+const { URL } = require('url');
```
gschier: Since this function isn't doing very much and is not used anywhere else, I recommend just adding it as a method on the `ResponseHeadersViewer` component.

KarthikSiddarth: sure, will work it and update you ASAP.

KarthikSiddarth:
Hey Greg,
Removed the validateURL module from the insomnia-url package. Placed the function inside the response-header-viewer module. Checked by running the app, it worked fine.

PR #1197, opened 2018-10-03
filter "parameters method" from swagger imports

beardeddragon5:
In swagger files you can specify parameters for a path.
Insomnia added this parameters as methods. This simply filters this "parameters methods"


gschier:
Can you please provide more detail on what this is actually doing? I'm not familiar enough with Swagger to know what this means.

beardeddragon5:
In swagger you can write this to add parameters to all methods for a specific path.
At the moment if you import a swagger file the "parameters" will be interpretet as a method.
There might be a lot more wrong if the importer ignores this, but the most annoying are the extra useless requests that get "imported". 
```
paths:
    /test/:
        parameters:
           ...
        get:
```

PR #1200, opened 2018-10-06
Fix Python Requests Code generation

### LINKED ISSUE #1188: [Bug] Python Requests code generation drops query string parameters
- Insomnia Version: 6.0.2 (6.0.2.2311)
- Operating System: macOS Sierra 10.12.6

## Details

If you try to generate code using the "Python > Requests" method, it seems to ignore what's in the querystring, I checked all the other code generators (including Python http.client) and they don't seem to be affected by this issue.

To reproduce:

`https://test.null/test?bug=true`

Result:
```
import requests

url = "https://test.null/test"

payload = ""
response = requests.request("GET", url, data=payload)

print(response.text)
```

Expected result:
```
import requests

url = "https://test.null/test"

payload = {'bug': 'true'}
response = requests.request("GET", url, data=payload)

print(response.text)
```

thewheat:
Closes #1188

Currently the code generation with Python Requests is missing out query string variables specifed in the URL string

image

New code fixes this and confirm it works with both query strings specified in URL and in query section editor
image

### DIFF
--- a/packages/insomnia-httpsnippet/src/targets/python/requests.js
+++ b/packages/insomnia-httpsnippet/src/targets/python/requests.js
@@ -24,7 +24,7 @@ module.exports = function(source, options) {
   code.push('url = "%s"', source.url).blank();
 
   // Construct query string
-  if (source.queryString.length) {
+  if (Object.keys(source.queryObj).length) {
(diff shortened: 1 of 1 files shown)


ryanprior:
Thanks for this, @thewheat. Nice simple patch 😄

Issue #848, opened 2018-04-05
[Improvement] Show remaining time for the refresh token expiration


rafaelrenanpacheco:
- Insomnia Version: 5.15.0
- Operating System: Arch Linux

## Details

Hi there!

In the OAuth2 tab, Insomnia calculates the remaining time for the access token, but does not for the refresh token, as you can see here:

![image](https://user-images.githubusercontent.com/12160864/38342821-f4511b3a-3856-11e8-82ab-e765c4f11216.png)

Would be possible to calculate and show the remaining time for the refresh token as well? If the refresh token has exp claim, show remaining time, else, don't.

Also, looks like the expired label is wrong:

![image](https://user-images.githubusercontent.com/12160864/38342904-7ac11bac-3857-11e8-9253-acd11fdb2ee7.png)

Shouldn't it be "**Expired** a few seconds ago"?

Best regards,
Rafael Pacheco.

gschier:
This should remain open.

KarthikSiddarth:
Hi Greg,
I would like to contribute on implementing the mentioned improvement.

gschier:
I'm not actually sure if there is a way to know the refresh token expiry date. The OAuth 2.0 RFC states that the response looks like this:

```json
{
   "access_token":"2YotnFZFEjr1zCsicMWpAA",
   "token_type":"example",
   "expires_in":3600,
   "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA",
   "example_parameter":"example_value"
}
```

The `expires_in` key relates to the `access_token` and it doesn't look like there is anything else that would indicate the expiry of the refresh token.

Perhaps someone with more OAuth 2.0 experience could clarify this?

samovertonjr:
I'll do a more research as well. Thanks.

rafaelrenanpacheco:
The expiration time is actually in the refresh token. For example, we have the following refresh token:

```
[...]

PR #1209, opened 2018-10-11
Added refresh token expiration.

samovertonjr:
** What does the code do?**

`This feature adds the expiration to the Refresh Token. This was a requested feature in issue` #848.

`Link to gif on how it looks` https://cl.ly/ef236a0ddf75


gschier:
I really appreciate the contribution @samovertonjr! Unfortunately, I don't think this is correct.

The refresh token will have a different expiry than the access token and it looks like your code is reusing the same expiry date for both.

I did some digging and found that there may not be a way to know if the refresh token expires at all. See my comment on the issue for more details https://github.com/getinsomnia/insomnia/issues/848#issuecomment-429001529

samovertonjr:
Ok @gschier thanks for the feedback. I should have asked for clarification because I thought that might be the case.

PR #1214, opened 2018-10-15
Fixes #922: Clear Content-Type when selecting 'No Body'

### LINKED ISSUE #922: when switching a request to "No Body", the content-type header is not removed
i'm sending a `GET` request. If I set the body to `JSON`, a header is added for `content-type: application/json`. However, when I switch back to `No Body`, the `content-type` header is not removed.

thewheat:
Closes https://github.com/getinsomnia/insomnia/issues/922

**Summary**
- Removes `Content-Type` header when selecting `No Body` from the request dropdown

image

**Demo**
remove-content-type

**Implementation details and notes**
- Created `CONTENT_TYPE_NO_BODY` constant to make code a bit clearer (no longer needing to check for `typeof mimeType`
- Auto linting seems to cause most of the changes seen
- There was a `keeps content-type` test that I had to remove as this change removes that behaviour for null `mimeTypes` (unsure if this behaviour is something we'd like to keep, if we do want to keep this, I'll need to rethink this approach)


packages/insomnia-app/app/models/request.js
```diff
@@ -294,8 +282,9 @@ export function updateMimeType(
   // 1. Update Content-Type header //
   // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ //
 
-  const hasBody = typeof mimeType === 'string';
-  if (!hasBody || mimeType === CONTENT_TYPE_OTHER) {
+  if (mimeType === CONTENT_TYPE_NO_BODY) {
```
gschier: You pretty much got this right, although it's a bit simpler than what you did. Technically the `Request.body` attribute can be `{}`. This is why the `hasBody` variable was set to `typeof mimeType === 'string'`. 

That means that your change can be simplified to the following:

```suggestion
  if (!hasBody) {
    headers = headers.filter(h => h !== contentTypeHeader);
  if (mimeType === CONTENT_TYPE_OTHER) {
```

gschier: I think this is actually the only change needed for this PR besides updating the tests.

Issue #1220, opened 2018-10-17
[Feature] : Add Keyboard Mnemonics and Global Keyboard Shortcuts Binding


tostringtheory:
Insomnia should allow/utilize the system mnemonic style for Window menus in the header/navigation menu bar.  On windows, the menu bar mnemonics are denoted by preceding the character by an '&'.  This should be flexible to support mnemonics on linux systems as well.  Darwin architecture does not appear to have a similar concept as far as I can tell.

Related - the user should also be able to access the list of insomnia shortcuts by a single key combination, instead of having to first access a menu option (such as Options/Preferences or the Help Menu). 

This would greatly increase the usability of insomnia in non-mousing environments.

PR #1221, opened 2018-10-17
Related to #1220 - keyboard bindings/mnemonics

tostringtheory:
This pull request begins to address adding mnemonics to insomnia menu items for easy accessibility on the Windows platform, as well as adding a global shortcut listener for a hotkey to access the insomnia shortcuts list.  Most changes from this checkin appear to be from auto-linting.


gschier:
I think this PR looks pretty good so far! Just a few things.

1. What is the purpose of the `&`? Is this some sort of special thing that Windows/Linux understands? A link to some docs would be appreciated 😄 
2. Can you post before/after screenshots (or a recording if possible) to better showcase what this actually does?
3. What's left to do on this?

tostringtheory:
@gschier - thanks for looking over this!  Sorry for the ghost - dealing with sudden death in the family.  Happy to finally be getting back to normalcy though.  In response to your questions:

1 - The '&' character in a menu label on Windows (and Linux I believe) denotes a 'mnemonic' for that menu label.  It is used to indicate basically the ALT shortcut key for that menu item.  In Windows, this is what drives the function where you see the underlines appear on a character of menu items when you press ALT, and natively makes the menu items focus when pressing ALT+{letter after mnemonic}.  As for a link, you can check out this MSDN article on Mnemonics - https://docs.microsoft.com/en-us/cpp/windows/defining-mnemonics-access-keys?view=vs-2017

2 - I will definitely post a picture of this in the next revision of the PR (hopefully today)

3 - I only implemented mnemonics on the top level menu and the file menu as an example of mnemonic usage to see if you were OK with the method of implementation.  I'm going to implement the rest of the menu items now and submit a PR.

packages/insomnia-app/app/main/window-utils.js
```diff
@@ -234,14 +239,15 @@ export function createWindow() {
...
+        label: `Show App ${MNEMONIC_SYM}Data Folder`,
         click: (menuItem, window, e) => {
           const directory = misc.getDataDirectory();
           shell.showItemInFolder(directory);
         }
       },
       {
-        label: 'Insomnia Help',
+        label: `Insomnia ${MNEMONIC_SYM}Help`,
+        accelerator: 'F1',
```
gschier: Can we make this accelerator Windows/Linux only? `!isMac()`

packages/insomnia-app/app/ui/css/components/tabs.less
```diff
@@ -110,6 +110,10 @@
       opacity: @opacity-super-subtle;
     }
 
+    &.react-tabs__tab--selected:focus {
```
gschier: This change should not be included in this PR, as it is not related. Can you open another issue for this so we can discuss the change separately?

packages/insomnia-app/app/ui/components/settings/shortcuts.js
```diff
@@ -23,6 +23,7 @@ class Shortcuts extends PureComponent {
       <div>
         <table className="table--fancy">
           <tbody>
+            {this.renderHotkey(hotkeys.SHOW_KEYBOARD_SHORTCUTS)}
```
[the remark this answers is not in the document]
tostringtheory: On this one - this is actually non-pertinent to the mnemonics and is applicable to all OSes.  This is just adding a global keyboard shortcut to open the preferences menu and jump to the shortcut tab, that falls in line with other applications supporting some form of CmdOrCtrl+Shift+? (sometimes without shift or alt).  The actual spec is configurable via the Hotkey definition.

packages/insomnia-app/app/main/window-utils.js
tostringtheory: I can make this Windows/Linux only - but I did just double check and saw that Apple/Mac has the standard of Command+(sometimes shift)+? for an applications help menu.  Would you prefer to add that accelerator to the binding, or just drop mac support all together?  I don't have a Mac to be able to test what Command+(sometimes shift)+? for insomnia currently.

packages/insomnia-app/app/ui/css/components/tabs.less
tostringtheory: I can push this to another PR.  I had originally intended for this PR to be accessibility enhancements for keyboard usage (in this case highlighting focused elements), but at this point I see that I didn't name it as such.

gschier: Always best to keep PRs as small as possible 😄

packages/insomnia-app/app/ui/components/settings/shortcuts.js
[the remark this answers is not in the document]
tostringtheory: Did you need a change for this or is this good?

packages/insomnia-app/app/main/window-utils.js
tostringtheory: Applied filter to !isMac

packages/insomnia-app/app/ui/css/components/tabs.less
tostringtheory: Done

packages/insomnia-app/app/ui/components/settings/shortcuts.js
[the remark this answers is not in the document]
gschier: This is okay since I now know it's a built-in Mac shortcut

PR #1226, opened 2018-10-21
Fixes #994: Fix multipart code generation

### LINKED ISSUE #994: [Bug] Multipart generated code miss pieces of the payload
- Insomnia Version: ___
- Operating System: Win 7

## Details

I've a multipart with two fields, username and password. It works in Insomnia. 

image

Whenever I generate the code for any language, I've the following (here in python requests)

```python
import requests

url = "http://192.168.60.10/api/index.php"

payload = "-----011000010111000001101001\r\nContent-Disposition: form-data; name=\"username\"\r\n\r\nadmin"
headers = {'content-type': 'multipart/form-data; boundary=---011000010111000001101001'}

response = requests.request("POST", url, data=payload, headers=headers)

print(response.text)
```

The payload is missing its data past the first one. I've tried a bit with more data, changing data name just in case, swapping order between username and password : only the first data is in the generated code.

thewheat:
Closes: https://github.com/getinsomnia/insomnia/issues/994

- Multipart code generation with multiple parameters did not have all parameters in the payload due to the asynchronous nature of `pipe` and the code not waiting for it to complete
- Added a promise and made the code wait for it to be resolved before continue
- Try to look for locations of `snippet.convert` and only found 1 other location so modified that as well

Scenario mentioned [in reporting issue](https://github.com/getinsomnia/insomnia/issues/994) now correctly returns both parameters in the payload

Wasn't too familiar with async and await stuff so open for feedback if I got anything wrong 😄 

image

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/generate-code-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/generate-code-modal.js
@@ -86,7 +86,7 @@ class GenerateCodeModal extends PureComponent {
     const { environmentId } = this.props;
     const har = await exportHarRequest(request._id, environmentId, addContentLength);
     const snippet = new HTTPSnippet(har);
-    const cmd = snippet.convert(target.key, client.key);
+    const cmd = await snippet.convert(target.key, client.key);
 
     this.setState({ request, cmd, client, target });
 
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/ui/components/modals/generate-code-modal.js
```diff
@@ -95,9 +95,9 @@ class GenerateCodeModal extends PureComponent {
     window.localStorage.setItem('insomnia::generateCode::target', JSON.stringify(target));
   }
 
-  show(request) {
+  async show(request) {
     const { client, target } = this.state;
-    this._generateCode(request, target, client);
+    await this._generateCode(request, target, client);
```
thewheat: Without this, the modal could show first without any code (or code from the last run) before overwriting with latest code.

thewheat:
Thanks for the feedback @gschier 
Will see what I can do 👍 May take a bit of time as I've got a couple of things going on at the moment but still keen to fix this up

PR #1292, opened 2018-12-03
Add clean deleted responses task on startup, and add tests

### LINKED ISSUE #1201: [Feature Request] Limitation of responses history
Hello!
After half year of my working with Insomnia I noticed that Insomnia's app data folder has size of 2Gb. After opening it I saw that it has "responses" folder inside contained 12k files of all server responses for the whole period. Is there any limitation or automatic removing for these files to prevent clear history manually? For example, I don't want to store all responses, it's enough for me to store only the last 5 for every request.
I looked Preferences window carefully and didn't see any settings related to history.

rickychandra:
Closes #1201
I add a function in `models.response` that will get all response files from responses directory. Then check whether each of them exists in the db. If no, then delete the file.
I batch each checks to the db by `MAX_RESPONSES` (currently 20) responses, on the assumption that there could be many files in the responses directory to be checked.
I also add tests for this function, but reasons I don't know yet, the `jest.useFakeTimers()` used in the `migrate()` test suite in the same file causes the subsequent suites' tests to timeout when they are calling async/await functions. Adding `jest.useRealTimers()` seems to solve it :smile:.

### DIFF
--- a/packages/insomnia-app/app/models/__tests__/response.test.js
+++ b/packages/insomnia-app/app/models/__tests__/response.test.js
@@ -12,6 +12,11 @@ describe('migrate()', () => {
     jest.useFakeTimers();
   });
 
+  afterEach(async () => {
+    // Reset to real timers so that other test suites don't fail.
+    jest.useRealTimers();
+  });
+
   it('migrates utf8 body correctly', async () => {
     const initialModel = { body: 'hello world!', encoding: 'utf8' };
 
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/models/response.js
```diff
@@ -232,3 +232,36 @@ function migrateBodyCompression(doc: Object) {
...
+  mkdirp.sync(responsesDir);
+
+  let files = fs.readdirSync(responsesDir);
+  if (files.length === 0) {
+    return;
+  }
+
+  // ~~ is for division without remainder.
+  const totalBatch = ~~((files.length - 1) / MAX_RESPONSES) + 1;
+  for (let batch = 0; batch < totalBatch; batch++) {
```
gschier: Is there a reason you're doing these in batches? It's difficult for me to tell what's going on.

Since we're using an in-memory database, it should be performant enough to loop over each response file one-by-one. Either that or fetch all responses from the DB at once, in one query.

packages/insomnia-app/app/sync/index.js
```diff
@@ -106,6 +106,8 @@ export async function init() {
 
   _writeChangesInterval = setInterval(writePendingChanges, WRITE_PERIOD);
 
+  await models.response.cleanDeletedResponses();
```
gschier: This isn't a great place to put this since it doesn't have anything to do with the sync system. However, there was no other option so I don't blame you 😄 

I just pushed a commit to the `develop` branch to  support an option `hookDatabaseInit()` in each model that will be called when the DB is initially created. Please move this call to there:

https://github.com/getinsomnia/insomnia/blob/8eb494dfe86428fa10a50dd90fd8f75fd8b8d94f/packages/insomnia-app/app/models/response.js#L82-L84

packages/insomnia-app/app/models/__tests__/response.test.js
```diff
@@ -134,3 +139,80 @@ describe('migrate()', () => {
     ).toBe('zip');
   });
 });
+
+describe('cleanDeletedResponses()', function() {
```
gschier: Nice work on these tests 👍

packages/insomnia-app/app/sync/index.js
rickychandra: Sure!
Actually until just now I thought package `sync` means synchronous, instead of synchronize. Aside from that I also thought it was a place for random operations (it says "things that can wait" [here](https://github.com/getinsomnia/insomnia/blob/9902b6617e7426b8901b4fa9caf56cecfbfeecb7/packages/insomnia-app/app/ui/index.js#L53)) :laughing:.

packages/insomnia-app/app/models/response.js
rickychandra: Actually I have concern that loading all responses will take too much memory.
I did a little benchmark on the 3 approaches (fetch for each `bodyPath` from db, batch fetch, and fetch all from db) using non in-memory db with 2000 responses (~100 requests) in db and 4000 from filesystem.
- The fetch-for-each is the slowest, on average 10.5 secs.
- The batch-fetch on average 1.6 sec.
- The fetch-all is the fastest, on average 0.5 sec. But I go with this option now, because I don't see any significant memory usages when I ran the performance profiler in the DevTools. Or maybe we could wait until there is a report on this :grimacing:.

rickychandra: I ran the benchmark with the spec:

Issue #1031, opened 2018-07-04
[Bug] Result preview pane does not refresh when it does not have focus


mlambley:
- Insomnia Version:
![image](https://user-images.githubusercontent.com/17264636/42296354-3ca1d9a0-8038-11e8-996d-b892b0a5a7ac.png)
- Operating System: Windows 10

## Details
- Click Send on a url, see the response
- Click into Header to see the headers which were sent to the client
- Change the url to one which will produce a different response
- Press Send
- See the different headers (good!)
- See that the header pane still has focus (great!)
- Click on Preview
- See that it still has the previous response, and that it hasn't updated
- Click Send again while the preview pane has focus
- See that it now updates the response

The Preview pane needs to update even when Header, Cookie, Timeline or any of the other tabs are selected.

But please do not fix this by forcing the preview pane to receive focus every time you click send! I want to able to review the headers generated by different requests without having to click on the header tab after each send.

Thank you.

gschier:
Woah, I can't believe no one has noticed this before. I think this actually has to do with Codemirror not updating when it's not visible. I hope this is an easy one to fix...

rickychandra:
Hi, I would like to work on this. :smile:

PR #1294, opened 2018-12-08
Refresh CodeMirror editor when focused, and remove duplicate listener

rickychandra:
Closes: #1031
I added response pane tab select listener, then trigger CodeMirror editor refresh when it is selected. For the editor to be able to be refreshed it must be visible; but since the tab selection listener is triggered right before that, I used `process.nextTick()` to do the refresh.
I also removed a seemingly duplicate line that registers the same callback to CodeMirror, which causes the callback to be called twice everytime the editor's value is changed.

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
@@ -132,7 +132,12 @@ class CodeEditor extends React.Component {
 
   focus() {
     if (this.codeMirror) {
-      this.codeMirror.focus();
+      // Refresh must happen when the editor is visible,
+      // so use nextTick to give time for it to be visible.
+      process.nextTick(() => {
+        this.codeMirror.refresh();
+        this.codeMirror.focus();
+      });
     }
   }
 
@@ -268,7 +273,6 @@ class CodeEditor extends React.Component {
 
     // Set default listeners
     const debounceMillis = typeof ms === 'number' ? ms : DEBOUNCE_MILLIS;
-    this.codeMirror.on('changes', misc.debounce(this._codemirrorValueChanged, debounceMillis));
     this.codeMirror.on('changes', misc.debounce(this._codemirrorValueChanged, debounceMillis));
     this.codeMirror.on('beforeChange', this._codemirrorValueBeforeChange);
     this.codeMirror.on('keydown', this._codemirrorKeyDown);
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
@@ -132,12 +132,13 @@ class CodeEditor extends React.Component {
 
   focus() {
     if (this.codeMirror) {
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/ui/components/viewers/response-viewer.js
```diff
@@ -60,6 +60,12 @@ class ResponseViewer extends React.Component<Props, State> {
     };
   }
 
+  focus() {
+    if (this._isViewFocusable()) {
```
gschier: Is there a reason for calling `focus()` on the editor? I feel like a new method specifically for refreshing may be more appropriate and a more specific fix to the original bug.

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -269,7 +274,6 @@ class CodeEditor extends React.Component {
     // Set default listeners
     const debounceMillis = typeof ms === 'number' ? ms : DEBOUNCE_MILLIS;
     this.codeMirror.on('changes', misc.debounce(this._codemirrorValueChanged, debounceMillis));
```
gschier: Oh, good catch!

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -132,7 +132,12 @@ class CodeEditor extends React.Component {
 
   focus() {
     if (this.codeMirror) {
-      this.codeMirror.focus();
+      // Refresh must happen when the editor is visible,
```
gschier: I feel like this `nextTick` logic would be better located in the caller of `focus()`. Technically, the editor has no way of knowing that it will be visible on the next frame but the caller hopefully does.

packages/insomnia-app/app/ui/components/viewers/response-viewer.js
rickychandra: Actually I was thinking about reusing `focus()`, because it's already been a "contract" function implemented by selectable views under `ResponseViewer`.
But on second thought, calling `refresh()` would have clearer meaning.
I've refactored it.

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
rickychandra: You're right, `nextTick` should be called by the caller.
CMIIW, I think `nextTick` is not a good solution, because it's not clear whether the selected tab will be made visible in the same event loop or not. But I haven't found better solution yet.

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -133,6 +139,17 @@ class ResponsePane extends React.PureComponent<Props> {
     });
   }
 
+  _handleTabSelect(index: number, lastIndex: number) {
+    if (this._responseViewer != null && index === 0 && index !== lastIndex) {
+      // Fix for CodeMirror editor not updating its content.
```
gschier: 👍 thumbs up for the very descriptive comment

PR #1323, opened 2019-01-03
Option to configure autocomplete popup delay

### LINKED ISSUE #1032: [Feature Request] Setting to not automatically suggest values while typing a url
- Insomnia Version:
image

## Details
image
As I'm typing a url, the above appears. If I press enter in the ~800ms before it appears, it sends the request, which is what I want. Unfortunately if that box appears, pressing enter selects the first option and messes up my url.

Could there be an setting to make it only appear when I press Ctrl+Space?

rafaelrenanpacheco:
Option to configure autocomplete popup delay

- Closes #1032


antimodern:
Is it availiable? I cant see it.

PR #1338, opened 2019-01-16
Fetch new access and refresh tokens when refresh fails

### LINKED ISSUE #849: [Improvement] Authenticate again if refresh token has expired
- Insomnia Version: 5.15.0
- Operating System: Arch Linux

## Details

Hi there!

When the access token expires, Insomnia tries to use the refresh token to get a fresh access token. But, if the refresh token has expired as well, the backend will throw an exception and will not send a fresh access token. In order to solve this issue, I have to clear the expired refresh/access token to allow Insomnia to authenticate again and fetch new refresh and access tokens.

If I have lots of requests with expired refresh tokens, I have to expend some time clearing and authenticating all over again.

Would be possible to add a checkbox near the refresh token, or in the advanced options, to indicate that if the refresh token is expired, Insomnia shall send a fresh authentication request to fetch new tokens? This should be optional, since the developer may want to test expired refresh token.

Best regards,
Rafael Pacheco.

rafaelrenanpacheco:
This PR will allow Insomnia to re-authenticate again to get brand new access and refresh tokens when the refresh token request fails due an unauthorized request. This happens when the backend rejects the refresh token for some reason, like refresh token expiration.

This is an opinionated implementation, which means that this behavior is not configurable. @gschier let me know if an option to configure this behavior is necessary.

- Closes #849

### DIFF
--- a/packages/insomnia-app/app/network/o-auth-2/get-token.js
+++ b/packages/insomnia-app/app/network/o-auth-2/get-token.js
@@ -190,6 +190,12 @@ async function _getAccessToken(
     authentication.scope,
   );
 
+  // If we didn't receive an access token it means the refresh token didn't succed,
+  // so we tell caller to fetch brand new access and refresh tokens.
+  if (!refreshResults.access_token) {
+    return null;
+  }
+
   // ~~~~~~~~~~~~~ //
   // Update the DB //
   // ~~~~~~~~~~~~~ //
--- a/packages/insomnia-app/app/network/o-auth-2/get-token.js
+++ b/packages/insomnia-app/app/network/o-auth-2/get-token.js
@@ -190,7 +190,7 @@ async function _getAccessToken(
     authentication.scope,
   );
 
-  // If we didn't receive an access token it means the refresh token didn't succed,
+  // If we didn't receive an access token it means the refresh token didn't succeed,
   // so we tell caller to fetch brand new access and refresh tokens.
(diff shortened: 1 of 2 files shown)


gschier:
I haven't taken a look yet but I'm excited for this! Will try and get to it next week.

packages/insomnia-app/app/network/o-auth-2/refresh-token.js
```diff
@@ -48,7 +48,14 @@ export default async function(
...
 
   const statusCode = response.statusCode || 0;
-  if (statusCode < 200 || statusCode >= 300) {
+
+  if (statusCode === 401) {
+    // If the refresh token was rejected due an unauthorized request, we will
+    // return a null access_token to trigger an authentication request to fetch
+    // brand new refresh and access tokens.
+
+    return responseToObject(null, [c.P_ACCESS_TOKEN]);
```
gschier: Nice and simple. I love it 😊

packages/insomnia-app/app/network/o-auth-2/get-token.js
```diff
@@ -190,6 +190,12 @@ async function _getAccessToken(
     authentication.scope,
   );
 
+  // If we didn't receive an access token it means the refresh token didn't succeed,
```
gschier: Thanks for adding the detailed comments 😄

PR #1339, opened 2019-01-18
Add ability to optionally use Hawk payload validation.

### LINKED ISSUE #1321: [Feature Request] Add Hawk Payload Validation
- Insomnia Version: 6.3.2
- Operating System: macOS High Sierra 10.13.6

## Details
Hawk authentication supports [optional payload validation](https://github.com/hueniverse/hawk#payload-validation) for validating the payload that the client sends to the server. If a server has enabled payload validation, insomnia will not generate the necessary hash for the server to validate. I've constructed an [example server](https://gist.github.com/rpless/2f1f215f5a638a331c736ec71d4c4844). If you make a post request from Insomnia with a body to that server and use the Auth Id and Auth Key from the readme, the server will return a 401. The error reported by the Hawk authenticator on the server is `Missing required payload hash`.

Would there be interest in adding this functionality if I made a PR?

rpless:
Closes #1321 
This change let's a user select whether Hawk Auth should use Payload Validation or not ). In terms of changes to the UI, it adds an additional checkbox to the Hawk Auth tab (which is off by default). I was a little unsure about the change I made to the signature of `getAuthHeader` so if there's a better way to get at the body and the content type without passing the whole request in let me know and I can adjust this.

### DIFF
--- a/packages/insomnia-app/app/network/network.js
+++ b/packages/insomnia-app/app/network/network.js
@@ -596,8 +596,7 @@ export async function _actuallySend(
         } else {
           const authHeader = await getAuthHeader(
             renderedRequest._id,
-            finalUrl,
-            renderedRequest.method,
+            renderedRequest,
             renderedRequest.authentication,
           );
 
--- a/packages/insomnia-app/app/network/network.js
+++ b/packages/insomnia-app/app/network/network.js
@@ -594,11 +594,7 @@ export async function _actuallySend(
         } else if (renderedRequest.authentication.type === AUTH_NETRC) {
           setOpt(Curl.option.NETRC, Curl.netrc.REQUIRED);
         } else {
-          const authHeader = await getAuthHeader(
-            renderedRequest._id,
-            renderedRequest,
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/network/network.js
```diff
@@ -596,8 +596,7 @@ export async function _actuallySend(
         } else {
           const authHeader = await getAuthHeader(
```
gschier: `finalUrl` still needs to be passed here as it might not be the same as `renderedRequest.URL`.

I'm fine with passing the whole request in (along with `finalUrl`). However, your change makes the first and last arguments redundant now, so they could be removed.

rpless: Ok so just to confirm, I should change the definition of `getAuthHeader` to be something like:
```
export async function getAuthHeader(renderedRequest: RenderedRequest, url: string): Promise<Header | null> {
...
```
And then adjust the call-sites and tests accordingly.

Issue #1348, opened 2019-01-23
[Feature Request] Support Base64url encoding


aldaris:
- Insomnia Version: 6.3.2
- Operating System: OSX

## Details

The Base64 encode/decode feature does not appear to support Base64url encoding at the moment. It would be nice if URL segments could be easily encoded/decoded using the Base64url encoding format, see:
https://tools.ietf.org/html/rfc4648#section-5

aldaris:
The main difference between regular Base64 encoding and Base64 url encoding is that the url version:
* uses `-` instead of `+`
* uses `_` instead of `/`
* does not have trailing padding `=` characters

If there is an issue with the PR, please do let me know, I'm more than happy to rework it.

PR #1349, opened 2019-01-23
Add support for Base64url encoding (#1348)

aldaris:
This small and dirty change adds support for Base64url encoding & decoding.

I had only one concern though, when I intentionally broke the test, npm test was still returning successfully, so not really sure what's going on with that...

### DIFF
--- a/plugins/insomnia-plugin-base64/index.js
+++ b/plugins/insomnia-plugin-base64/index.js
@@ -13,20 +13,33 @@ module.exports.templateTags = [
         ],
       },
       {
+        displayName: 'Kind',
+        type: 'enum',
+        options: [{ displayName: 'Normal', value: 'normal' }, { displayName: 'URL', value: 'url' }],
+      },
+      {
         displayName: 'Value',
         type: 'string',
         placeholder: 'My text',
       },
     ],
-    run(context, op, text) {
+    run(context, action, kind, text) {
       text = text || '';
 
-      if (op === 'encode') {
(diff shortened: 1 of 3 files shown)


plugins/insomnia-plugin-base64/index.js
```diff
@@ -9,7 +9,9 @@ module.exports.templateTags = [
         type: 'enum',
         options: [
           { displayName: 'Encode', value: 'encode' },
```
gschier: What do you think about splitting this up into two parameters?

Param 1: Select between `encode` and `decode`
Param 2: Select between base64 types

This way we can easily support more types (there are many apparently) in the future.

aldaris: That should be doable, I'll have a look.

aldaris:
@gschier do you have any tips, why `npm test` doesn't fail when I intentionally break the tests I've added?

aldaris:
Never mind, I figured it out.

aldaris:
The review comments have been addressed, let me know if any other changes are needed. Thanks!

gschier:
Awesome, thanks!

valdetero:
For some reason, this broke all of my existing encoding Tags. I had to remove and re-add them to get them to work. First it through a 400 and errored about attribute value. When I realized it was this tag, I went in and tried to fix it, but the preview constantly failed. Removing and re-adding caused it to work fine again.

PR #1378, opened 2019-02-20
Option to configure number of saved responses to history

Grimones:
Adds and option to configure how many responses it should save to history.

Today I got in a situation where I sent a malformed request to 3rd-party API and it accidentally cleared crucial data. And I couldn't find the response\url because at the time we've noticed that something was wrong the history was already overridden.


Grimones:
The OAuth 2.0 are breaking. Not sure why  😕

gschier:
Don't worry, it's complicated 😭. It's because the tests use deterministic ID generation logic, so the addition of the `models.settings.getOrCreate()` throws it off by adding an additional ID into the mix. 

In order to decouple this a bit, I think it'd be wise to pass the max responses into the `create` function as a parameter instead of looking up the settings object from there. I would add it as a second, optional, parameter to keep things simpler. Note, only the calls within `app.js` would _need_ to be updated.

```js
export async function create(patch: Object = {}, maxResponses: number = 20) {
  // ...
}
```

packages/insomnia-app/app/models/response.js
```diff
@@ -130,13 +129,19 @@ export async function create(patch: Object = {}) {
 
   const { parentId } = patch;
 
+  const settings = await models.settings.getOrCreate();
```
gschier: In order to decouple this a bit, I think it'd be wise to pass the max responses into the `create` function as a parameter instead of looking up the settings object from there. I would add it as a second, optional, parameter to keep things simpler. Note, only the calls within `app.js` would _need_ to be updated.

```js
export async function create(patch: Object = {}, maxResponses: number = 20) {
  // ...
}
```

PR #1393, opened 2019-03-02
Add request authentication getter and setter to plugin request context.

samirnijenhuis:
This PR enables plugin developers to set and get the authentication configuration of a request. 

I built a [small plugin](https://github.com/samirnijenhuis/insomnia-plugin-env-authentication) that allows me to define my OAuth 2 settings on an environment and automate the authentication for every request I add to my workspace. This way I don't have to look / configure auth settings for every single request.

The following issues could benefit from this pr:
https://github.com/getinsomnia/insomnia/issues/1116
https://github.com/getinsomnia/insomnia/issues/790

### DIFF
--- a/packages/insomnia-app/app/plugins/context/request.js
+++ b/packages/insomnia-app/app/plugins/context/request.js
@@ -131,14 +131,15 @@ export function init(
       }
     },
 
-
-    setAuthentication(name : string, value: string) : void {
-      renderedRequest.authentication[name] = value;
+    setAuthentication(name: string, value: string): void {
+      Object.assign(renderedRequest.authentication, {
+        [name]: value,
+      });
     },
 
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/plugins/context/request.js
```diff
@@ -131,6 +131,16 @@ export function init(
       }
     },
 
+    setAuthentication(name: string, value: string): void {
```
gschier: I feel like this would be more aptly named `setAuthenticationParameter` since it's only setting a single value within the auth configuration. Thoughts?

samirnijenhuis: Yes, can't agree more. `setAuthentication` feels like you set the whole object. Not just a key/value.

PR #1409, opened 2019-03-16
making exports git friendly

everplays:
without this, body of a json request looks like the following in the export:

```json
"body": {
  "mimeType": "application/json",
  "text": "{\n\t\"user\": {\n\t\t\"expectation\": \"this is in human readable format\",\n\t\t\"answer\": 42\n\t}\n}"
}
```

with it:

```json
"body": {
  "mimeType": "application/json",
  "payload": {
    "user": {
      "expectation": "this is in human readable format",
      "answer": 42
    }
  }
}
```

## Why?

We have a couple of workspaces that we have checked into our git repository and we use them as documentation and examples of our APIs. Everything works well until two developers change the body of a request. Even if they do not change the same line in the body (something that git can handle without merge conflict), we still end up with conflicts because the entire body gets collapsed into one line in the export (something that git can NOT handle).

This change ensures that at least for JSON format, not only we will end up with less conflicts but they can be resolved by the developers a lot easier.

## What about other formats?

One solution could be to generalize this concept by splitting each line of body. Something like:

```json
"body": {
  "mimeType": "application/json",
  "text": [
    "{",
    "\t\"user\": {",
    "\t\t\"expectation\": \"this is in human readable format\",",
    "\t\t\"answer\": 42",
    "\t}",
    "}"
  ]
}
```

Even though this is definitely better than putting everything in one line as string, I personally prefer to keep the JSON as currently implemented in this PR (it's odd to modify it when we can just put it in the export).

## Backwards compatibility?

Yes, it is backwards compatible.

## next?

I'll wait for feedback from you to see which way is preferred then I'll finalize the PR by addressing them and adding tests.


everplays:
@gschier any chance I can get your thoughts on this?

gschier:
This solution seems a bit too specific for my liking since it only works for JSON bodies. What about other fields like request documentation, environments, etc?

I do agree, however, that making exports git-friendly is a good idea. I feel like moving from JSON to something that supports newlines (like YAML or TOML) would be a better way to go.

everplays:
@gschier FYI, I opted for YAML.

A few notes:

- I tried to use `js-yaml` (existing dependency) and `swagger-parser`'s YAML. However, both of them escaped multi-line outputs so I had to use `yaml` package which has the desired effect.
- The git hooks of project changed parts of `output.json` files that I had not modified. Let me know if you don't want to include them, I can remove the changes and simply commit with hooks disabled.
- Not sure why but `flow check` seem to fail with `Cannot resolve module 'yaml'` even though the dependency is added and app works when I use the interface.

everplays:
@gschier thanks for taking care of it.

Issue #1202, opened 2018-10-07
[Bug] npm error when trying to install plugin


BlueberryKing:
- Insomnia Version: 6.0.2
- Operating System: Windows 10

## Details

I'm trying to install a plugin, `insomnia-plugin-jwtdecode` to be specific, but I get the following error when trying to install via Preferences -> Plugins:

```
insomnia-plugin-jwtdecode npm error: Command failed: C:\Users\username\AppData\Local\insomnia\app-6.0.2\Insomnia.exe --no-deprecation C:\Users\username\AppData\Local\insomnia\app-6.0.2\resources\bin\yarn-standalone.js info insomnia-plugin-jwtdecode --json
```

I opened DevTools, but I don't see an error, I only see: `[plugins] Fetching module info from npm`

Please let me know if you want me to provide more information. Thank you.

gschier:
Hmm, I can't reproduce this on Mac. Maybe try running the command manually from the command line and see if you get more of an error message?

gschier:
Sorry, after `Command failed: ` it shows the command that it tried to run. Can you try running that manually?

```C:\Users\username\AppData\Local\insomnia\app-6.0.2\Insomnia.exe --no-deprecation C:\Users\username\AppData\Local\insomnia\app-6.0.2\resources\bin\yarn-standalone.js info insomnia-plugin-jwtdecode --json```

Insomnia does not use your system version of Node or NPM. It uses a bundled version of Yarn instead.

**Comment by BlueberryKing** [comment]
Gotcha. I actually tried that, but it just launches Insomnia. The console shows this output:

```
[fix] Running database repairs
[db] Initialized DB at C:\Users\valen\AppData\Roaming\Insomnia\insomnia.$TYPE.db
[localstorage] Initialized at C:\Users\valen\AppData\Roaming\Insomnia\localStorage
[main] Loading file://C:\Users\valen\AppData\Local\insomnia\app-6.0.2\resources\app.asar/renderer.html
[updater] Using url https://updates.insomnia.rest/updates/win?v=6.0.2&channel=stable
[updater] Checking for updates url=https://updates.insomnia.rest/updates/win?v=6.0.2&channel=stable
[updater] Not Available
```

aduvoisin13:
I'm seeing this exact same problem, same output and results. 6.0.2 / Windows 10. Workaround was to go into the plugins folder and `npm install insomnia-plugin-default-headers`.

gschier:
[...]

PR #1425, opened 2019-04-03
Fix issue #1202. Lenient error handling for Yarn output. Check if out…

senoctar:
As written in #1202 the problem with exit error codes on Windows seems to be in yarn or electron. The root cause is not likely to be solved, it seems to be a reoccurring issue randomly affecting projects. It's probably deep in memory management.

The fix primarily consists of first testing if Yarn standard output is OK (either json or success prompt depending on command) and checking the exit error code only if the standard output test fails.


senoctar:
PR is updated with explanatory comments on changed code.

PR #1433, opened 2019-04-12
feat: postman 2.1 collection export initial

### LINKED ISSUE #1156: [Improvement] Export collections using Postman format
Insomnia supports exporting collections in the Insomnia and HAR formats. I would like to add the ability to export using the Postman collection format(s).

I have spent a lot of time building an extensive collection in Insomnia that would be of great use to others in my organization, but most of them use Postman, and currently there is no way to export collections to a Postman-compatible format. I have no intention of making it more attractive for people to use Postman (quite the opposite in fact), but this would be a useful feature that would allow teams to use the REST client of their choice while still being able to share collections.

Documentation for the various Postman collection JSON schemas can be found at https://schema.getpostman.com

If this is something you think would be useful, I would be happy to implement it in my spare time and open a PR.

dmhalejr:
🚧 **WORK IN PROGRESS: Postman 2.1 Collection Exports** 🚧 

Closes #1156 [eventually]

This is the initial start of exporting Insomnia workspaces to the Postman 2.1 collection format for later importing into Postman.

## General Description
Add a new export ability to POSTMAN 2.1 from the Export modal to save as a Postman 2.1 collection format file. 

## Implementation Details
Following the same pattern that has been defined with other exports functionalities I just added another `else if` to run `exportLatestPostmanCollection()` that is just taking the insomnia schema and changing it into the postman 2.1 schema.

## Shortcomings
Not all permutations of Insomnia features have been represented in this PR. If an end user were to want to export: environment variables, authorization headers, sub-folders, et. al. this will not do that. Furthermore, end users wanting to export from Postman 2.0 or earlier have not been met. 

This was to start a discussion to continue the work as I'm interested in getting it to feature complete but wanted to make sure what I can offer here and claim #1156 until complete.


gschier:
Thanks for taking this on @dmhalejr! I think a lot of people will find this useful.

I think keeping it limited to 2.1 is fine. The only thing that worries me at the moment is what to do if there are Insomnia features being used that we _can't_ export. For example, Insomnia's template syntax is a lot more feature rich than Postman's and is extendable via plugins. 

Did you have any ideas for how to handle stuff like that?

dmhalejr:
@gschier Getting back to this since last we spoke. I have not thought about the extendability of the template syntax and maybe just adding some sort of warning on the exporting modal explaining the limitations would be enough to warn those knowing what they are getting exporting out? 

I'll be working on getting core functionality mapped properly between the export and the postman equivalent and will hopefully get some additional insight to answer your earlier question.

dmhalejr:
@gschier

In Postman's world there is an ability to create environment variables at the collection level but it is recommended those environment variables are not sensitive in nature. 

I am taking the stance of creating a completely different environments file that gets exported from Insomnia for all variables when selecting Postman. Trying to figure out how to stutter step a file export dialog between the main sources/collection file and the environment file. 

What are your thoughts on that? 

```
Collection variables allow you to define a set of variables, that are a part of the collection, as opposed to environments, which are separate entities.
Note: Collection variables must not contain any sensitive information.
```
[source](https://schema.getpostman.com/collection/json/v2.1.0/draft-07/docs/index.html)

gschier:
I will leave that decision up to you @dmhalejr. I'm not that familiar with Postman and what you said sounds like a good idea 👍

As for exporting multiple files. One idea could be to have the user select the file location, then generate the environment file name based on the other one.

For example `my-postman-export.json` could save the environment as `my-postman-export.env.json` automatically.

PR #1442, opened 2019-04-20
Move request to collapsed folder by opening on hover

### LINKED ISSUE #1329: [Feature] Move request inside collapsed folder
Currently cannot drag and drop request in left sidebar to collapsed folder.
Currently this works only for expand folder.

develohpanda:
Closes #1329.

## Goals
- Use existing behavior for moving `Request` or `RequestGroup` into an expanded folder.
- Should support moving `Request` to collapsed `RequestGroup`
- Should support moving `RequestGroup` to collapsed `RequestGroup`
- Should support moving any nested combination of `Request` and `RequestGroup`

## Investigations
Initial attempt was to expand the collapsed group on drag-hover. Works, but far from ideal because in the case of multiple `RequestGroup` rows, **all** would open when dragging something from the top of the list to the bottom.

I investigated a hover delay but there is no built-in functionality to `react-dnd` to achieve this. Modifying the package feels too clunky and would be another thing to maintain.

I investigated dropping a `Request` onto the collapsed `RequestGroup` and have it added to the top of the group by default. A user would then need to click to open the group, and order to the desired location. An unnecessary additional click.

## Solution
I settled on showing an unobtrusive `open` tag when drag-hovering over a request group. While dragging, hovering over a collapsed `RequestGroup` will show the `open` tag, and hovering over the tag will expand the group. This tag is only shown on a collapsed `RequestGroup`.

Collapsed:
image

Expanded:
image

## Improvements
Would love some guidance on whether using `ref` to identify the expand tag in this use case is appropriate or if another solution should be used.

### DIFF
--- a/packages/insomnia-app/app/ui/components/sidebar/sidebar-request-group-row.js
+++ b/packages/insomnia-app/app/ui/components/sidebar/sidebar-request-group-row.js
@@ -89,7 +89,7 @@ class SidebarRequestGroupRow extends PureComponent {
             <div
               ref={this._setExpandTagRef}
               className={classnames('sidebar__expand', {
-                'sidebar__expand-hint': isDraggingOver,
+                'sidebar__expand-hint': isDraggingOver && isCollapsed,
               })}>
               <div className="tag tag--no-bg tag--small">
                 <span className="tag__inner">OPEN</span>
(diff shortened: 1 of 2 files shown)


develohpanda:
> Wow, I love this solution. Very nicely done 😄 💯

Thanks @gschier! 😄

4n70w4:
@develohpanda your way is very complicated and convenient. An example of how this is done in the phpStorm IDE:

image

image

It is easier, clearer and more convenient.

jose920405:
This still not working in Versión `6.6.2`

PR #1471, opened 2019-04-27
Support for pinned requests

### LINKED ISSUE #1109: [Feature Request] - Star/Favorite/Pin requests
My work-space has numerous folders for all of the CRUD endpoints for various resources. We're sharing these endpoints among our team and each folder uses environment variables to set the base url. It would be nice if I could star/pin/bookmark or somehow make a private list of shortcuts to these requests that wouldn't sync to other users but would still allow environment variables to be maintained.

This could potentially be workable with something like #576 where a global history of requests would allow for clicking between recent requests irrespective of environment/workspace

Insomnia version 6.0.2

### LINKED ISSUE #1344: [Feature Request] Sticky request in the left menu.
##Add sticky request in top of the left menu.

- Insomnia Version: 6.3.1
- Operating System: Linux Mate (Ubuntu 18.04.1 LTS)

## Details

I think it should be a good idea to have some sticky request, for example, when you have a important request like Token generator for authentication but are you have more than 30 other request saved in the request menu (left one in the UI).

image

I could be store in the same menu but with a flag or somthing like that, something that make a difference look from the other ones.

develohpanda:
Closes #1344. Closes #1109.

### Features
- Only allows pinning of top level `request` and `request-group`
- Prevent drag/drop of pinned items
  - Dragging nested items into/out of a pinned `request-group` is possible
  - Manipulation of nested items remains unchanged
- Independent scrolling for pinned and unpinned items
  - Pinned items always remain at the top
- Pinned item list + separator shown when at least one pinned item exists
- Pinned item list height constrained via `fit-content(20%)`
  - Auto height up to 20%, after which scroll bar will show
  - 20% will allow for 6 items to be pinned before overflow scrollbar shows
- Pin/Unpin action items added to action drop-downs
- `fa-thumb-tack` shown on pinned items

`pinned` state is stored in `request-meta` and `request-group-meta`

### Screenshots
No pinned items (current behavior):
image

Pinned folder + request, with unpinned items:
image

Six pinned items:
  image

Seven pinned items:
image

Open tag on collapsed, pinned folder to allow dnd of nested items:
image

Action dropdowns:
image
image
image
image

No pin action on nested items:
image


packages/insomnia-app/app/ui/components/sidebar/sidebar-children.js
```diff
@@ -13,13 +14,16 @@ type Child = {
   children: Array<Child>,
   collapsed: boolean,
   hidden: boolean,
+  pinned: Boolean,
```
gschier: Should be `boolean` 😄

gschier:
I did a once over of the code and it looks great. Nice work!

One high-level thing though is that I think people will want to be able to pin any request, not just the top-level ones. I'm not sure pinning folders is useful, although maybe you can point out some use-cases that I haven't considered.

Overall great work on this. I'm super excited to get this one merged 💯

develohpanda:
Thank you - I appreciate it! 😄

> people will want to be able to pin any request, not just the top-level ones

- Enabled the pin action item in nested items
- Updated `parentId` to the active workspace id inside the pin handle methods

> I'm not sure pinning folders is useful

I see pinned items as those being truly global, which would really only be authentication requests. If all requests are pinned, then none of them are pinned so I agree with your sentiment. 😄 Only example I can think of is with larger solutions with multiple APIs and 3rd party integrations where there may be logical groupings for pinned items.

Your call - given it is probably a niche use-case I could remove the ability to pin folders and add if a request comes up.

gschier:
Yep that sounds good. Pinned requests for now and folders if people want those later. 

Just a little background. The original use case for pins that I remember is basically treating them as tabs (like in a text editor). So you might pin just the few requests you're working with at the moment. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

develohpanda:
Makes sense. Done deal, PR has been updated!

rickychandra:
@develohpanda Hi, just a suggestion. Would it be better if the pin icon is at the left side? Putting it at the right side will take up the dropdown's arrow, and will be moved aside when the request row is hovered over, which IMHO is a bit awkward :smile:.

develohpanda:
> Would it be better if the pin icon is at the left side?

Thanks for the suggestion @rickychandra! I have a few reasons why I decided against an icon on the left which I'll outline below, but my ears are open to anything I may have missed. 😄 (apologies in advance for the long message!)

1) Padding inconsistency. If I move it to the left had side, it shifts the folder icon/method tag + text to the right, but only for pinned items. This alignment is inconsistent with unpinned items, and reading at a glance becomes tricky (similar to reading a balance sheet where the decimal point is not aligned with every other line). This is a hindrance and will slow down navigation.

2) Unnecessary cluttering. Differentiating request from folder (ie. method tag vs folder icon), and one row from another (ie. name) is more important than differentiating a pinned row from an unpinned row. A pinned row will always be at the top regardless, so adding unnecessary cluttering to the first two items (tag/folder + name) only gives more for the user to ingest. This is also a hindrance that will slow down navigation.

3) Cost/benefit.
**Pin on left**
Cost: styling inconsistencies + unnecessary cluttering _that are always visible_
Benefit: no movement of pin when hovering over a row
**Pin on right**
Cost: movement of pin _only on hover_ to make way for dropdown arrow
Benefit: styling consistency, no UI hindrances introduced for primary tasks (navigating to and identifying a row).

The pin is only supplementary info necessary for certain rows, and is kept out of the way. The drop-down arrow is similar, in that it is supplementary info only for hovered rows, and is kept out of the way (or hidden) when not needed.

image image

rickychandra:
>The pin is only supplementary info necessary for certain rows

Ah I see. I thought the pin is also clickable, so it would be easier to click if it doesn't move on hovered.

I agree with point 1 and 2, putting the pin on the left can make it hard to skim the items, especially when the pinned status is no longer that important (all pinned items are at the top and are separated with a line with the rest of the items).

Good job anyway! :smile:

develohpanda:
Thank you! Appreciate the suggestion 🥂

gschier:
After testing out your branch locally, I have a few usability questions/suggestions:

1. I expected pinning a request keep showing the request in its original position (I couldn't think any apps that support pinning besides GitHub but this is how GitHub issue pinning behaves there)
2. I expected the pin icon on the right to unpin the request. Having it move out of the way on hover felt jarring. I think it would either feel better if the pin icon disappeared on hover or that the dropdown caret appeared to the left of the pin icon instead of to the right. Also, if we decide to show both icons, clicking the pin icon should unpin it too.
3. It would be super useful to have a keyboard shortcut to toggle the pin state of a request. Let me know if you need help with this one.

And a bug: When pinning a request that's in a folder, then unpinning, the request is no longer inside the folder.

Let me know if your thoughts on these or if you have any questions 😸

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -31,6 +32,16 @@ class RequestActionsDropdown extends PureComponent {
     this.props.handleCopyAsCurl(this.props.request);
   }
 
+  _canPin() {
+    return this.props.handleSetRequestPinned !== misc.nullFn;
+  }
+
+  _handleSetRequestPinned() {
+    if (this._canPin()) {
```
gschier: No need to check if it's `nullFn` here right? We can just call it and it will no-op.

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -53,6 +64,11 @@ class RequestActionsDropdown extends PureComponent {
         <DropdownButton>
           <i className="fa fa-caret-down" />
         </DropdownButton>
+        {this._canPin() && (
```
gschier: Do we still need this check if all requests are now pinnable?

packages/insomnia-app/app/ui/components/sidebar/sidebar-request-row.js
```diff
@@ -143,7 +148,9 @@ class SidebarRequestRow extends PureComponent {
       );
     }
 
-    if (!this.state.isEditing) {
+    if (isPinned) {
+      return node;
```
gschier: Nice, good catch on disabling drag-n-drop 👍

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -437,6 +437,11 @@ class App extends PureComponent {
     App._updateRequestGroupMetaByParentId(requestGroupId, { collapsed });
   }
 
+  async _handleSetRequestPinned(request, pinned) {
+    App._updateRequestMetaByParentId(request._id, { pinned });
+    models.request.update(request, { parentId: this.props.activeWorkspace._id });
```
gschier: Shouldn't be setting the `parentId` in the request update right? Doing this will move it out of any folders it was in.

packages/insomnia-app/app/ui/css/components/sidebar.less
```diff
@@ -4,7 +4,7 @@
 .sidebar {
   position: relative;
   display: grid;
-  grid-template-rows: @height-nav auto auto minmax(0, 1fr) auto;
+  grid-template-rows: @height-nav auto auto fit-content(20%) auto minmax(0, 1fr) auto;
```
gschier: I think something here broke the empty state when sync is turned on. Now the sync menu, which is supposed to be on the bottom, floats up to the top:

image

It should be like this:

image

gschier: Ah, it's because the sidebar separator is hidden in that case, causing one fewer grid elements. Perhaps making a `sidebar__list-separator--invisible` class that sets the height to `0px` or something would work. It's up to you how you want to go about that.

develohpanda: Gotcha - any particular reason for using `grid` here vs `flexbox`?

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
develohpanda: Can't argue with this logic 😅 an oversight.

Is using `nullFn` a fairly common pattern? This is the first project where I have encountered it. Feels so much cleaner than `null` or `undefined`!

develohpanda:
> I expected pinning a request keep showing the request in its original position

Do you mean for the pinned request to show in two places - nested section and its original position? 👍

I think I won't be able to get to these until the weekend. Thank you for the feedback!

gschier:
@develohpanda yes, show in both places 😄 

Don't worry, take your time!

develohpanda:
@gschier two questions:

- How should the filter behave with pinned requests? IMO pinned items should not be affected by filters and should always be visible.
- Given pinned items are to show in both places, should the pin icon also show in both places (or only in the pinned list at the top)?

packages/insomnia-app/app/ui/components/sidebar/sidebar-children.js
```diff
@@ -16,6 +16,11 @@ type Child = {
   pinned: boolean,
 };
 
+export type SidebarChildObjects = {
+  pinned: Array<Child>,
+  all: Array<Child>,
+};
+
```
develohpanda: Is there a more appropriate place to put this type definition?

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -215,6 +215,18 @@ class App extends PureComponent {
...
+        hotKeyRefs.REQUEST_TOGGLE_PIN,
+        async () => {
+          const metas = Object.values(this.props.entities.requestMetas).filter(
+            m => m.parentId === this.props.activeRequest._id,
+          );
+          await this._handleSetRequestPinned(
+            this.props.activeRequest,
+            !(metas[0] && metas[0].pinned),
+          );
+        },
```
develohpanda: Not a big fan of this solution, but the other solution would be to put the `pinned` on `Request` rather than `RequestMeta`. 🤔 

Are there any options that I am missing?

packages/insomnia-app/app/ui/redux/selectors.js
```diff
@@ -149,15 +149,23 @@ export const selectSidebarChildren = createSelector(
...
+        return children.map(c => {
+          const child = {
+            doc: c,
+            hidden: false,
+            collapsed: !!collapsed[c._id],
+            pinned: !!pinned[c._id],
+          };
+
+          if (child.pinned) {
+            pinnedChildren.push(child);
```
develohpanda: I did this for efficiency. Extracting all pinned requests (ie. nested requests) in `sidebar-children` would require an extra unnecessary traversal of `childrenTree`.

packages/insomnia-app/app/ui/containers/app.js
develohpanda: Just thinking, if we put `pinned` on `Request`, it would simplify showing a pin icon for pinned requests in the quick switcher too. Is this something I should add?

packages/insomnia-app/app/ui/redux/selectors.js
gschier: I think it actually fits well here within the context of "sidebar children" since the only place it's used is in the sidebar. If/when pinned requests are needed elsewhere, this would make less sense. For now, it's good though.

gschier:
> How should the filter behave with pinned requests? IMO pinned items should not be affected by filters and should always be visible.

I'll leave this one up to you. I think both ways make sense. We'll leave it to future feedback to tell us if we made the wrong choice

> Given pinned items are to show in both places, should the pin icon also show in both places (or only in the pinned list at the top)?

I think in both places makes sense 👍

packages/insomnia-app/app/common/hotkeys.js
```diff
@@ -288,6 +290,11 @@ const defaultRegistry: HotKeyRegistry = {
     keyComb(true, false, false, false, keyboardKeys.d.keyCode),
   ),
 
+  [hotKeyRefs.REQUEST_TOGGLE_PIN.id]: keyBinds(
+    keyComb(false, false, false, true, keyboardKeys.p.keyCode),
```
gschier: Mac key comb should probably be `false, false, true, true` to map to Cmd+Shift+P because Cmd+P already maps to the request switcher. Windows/Linux one is fine though with what you have (Ctrl+Shift+P)

packages/insomnia-app/app/ui/containers/app.js
develohpanda: > Note. One thing that would help here is using the `getOrCreateByParentId` method on `request-meta.js`.

`_handleSetRequestPinned` internally calls `App._updateRequestMetaByParentId`, which is the same one used by the other handlers and supports `getOrCreate`. If it is a new request (with no meta) then this handler also creates meta.

https://github.com/getinsomnia/insomnia/blob/d8066bc55887b865c359a8f8815c44c9f7bdf325/packages/insomnia-app/app/ui/containers/app.js#L388-L396

develohpanda:
Ready again!

packages/insomnia-app/app/ui/containers/app.js
gschier: Ya I saw that. I thought your first comment was about the complexity of finding the request meta in the entities list.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -218,12 +218,17 @@ class App extends PureComponent {
       [
         hotKeyRefs.REQUEST_TOGGLE_PIN,
         async () => {
+          if (!this.props.activeRequest) {
+            return;
+          }
+
           const metas = Object.values(this.props.entities.requestMetas).filter(
```
gschier: Should be using `.find()` instead of `.filter()`

develohpanda: [TIL](https://stackoverflow.com/a/33759573). Will update this evening!

PR #1497, opened 2019-05-12
Preserve property order in environment editor

### LINKED ISSUE #1046: [Feature Request] Sort environment variables
- Insomnia Version: 5.16.6
- Operating System:Ubuntu 18.04
## Details

The variables in "Manage Environments" are automatically ordered alphabetically. While this is good some of the time, sometimes I'd like to logically group certain variables together and in a certain order.

It would be nice to have this optional - perhaps the alphabetical sorting could be a (default on) option?

develohpanda:
Closes #1046.

image

Added an option to keep property order by consuming [json-order](https://github.com/develohpanda/json-order). This is opt-in, thus **disabled** by default. This behavior is available to both the workspace environment editor, and the request group override environment editor.

Field names cannot begin by `$` or contain a `.` (as part of the `nedb` [documentation](https://github.com/louischatriot/nedb/blob/aa3302e5dcefa471431901a8f3fe63e532a5a7b2/README.md#inserting-documents)). 

As a result, the map generation uses `&` as the prefix and `~|` as the separator. These two options are parameters in the `parse` and `stringify` methods. An example map is shown below.

```js
const map = {
  "&": ["a"],
  "&~|a": ["z", "y"],
  "&~|a~|z": ["b"],
  "&~|a~|y~|2": ["d", "c"]
}
```


packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
```diff
@@ -441,39 +451,39 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
...
                     value={activeEnvironment}
-                    onClick={this._handleUnsetColor}
-                    disabled={!activeEnvironment.color}>
-                    <i className="fa fa-minus-circle" />
-                    Unset Color
-                  </DropdownItem>
-                </Dropdown>
-              ) : null}
-
-              {activeEnvironment && rootEnvironment !== activeEnvironment ? (
```
develohpanda: This condition is identical to the one above, so I pull both components into a parent `React.Fragment`. Easier to see with a "no whitespace" diff.

develohpanda:
I'm not sure I added the dependency correctly though - should `json-order` be under `packedDependencies` or `dependencies`?

gschier:
> I'm not sure I added the dependency correctly - should `json-order` be under `packedDependencies` or `dependencies`?

It should be in both. `packedDependencies` is a special insomnia-specific property that the build script uses to figure out whether the dependency should be bundled in the Webpack output or left as an external node module. In general, most should be bundled, but some modules don't work like that so need to be left alone.

develohpanda:
Hmmm - I'm not sure why the CI builds are failing. I added to the packed dep but it still failed. Surely I've missed a step somewhere while adding the package 🤔 

image

gschier:
Ah, @develohpanda it's because you need to add a Flow definition to the `flow-typed` folder. I wish there were a way to have a better error message when that happens.

Here's a good example of one to take a look at: https://github.com/getinsomnia/insomnia/blob/639d0cc33b0d455f43f449da1cbbe42dffa443cc/packages/insomnia-app/flow-typed/mkdirp.js

packages/insomnia-app/app/models/environment.js
```diff
@@ -26,6 +27,7 @@ export function init() {
   return {
     name: 'New Environment',
     data: {},
+    propertyMap: null,
```
gschier: Can we rename this to `dataPropertyOrder` in order to better signify what it's for?

packages/insomnia-app/app/models/request-group.js
```diff
@@ -22,6 +23,7 @@ export function init() {
     name: 'New Folder',
     description: '',
     environment: {},
+    propertyMap: null,
```
gschier: Same with this one. `environmentPropertyOrder` seems more descriptive. I'm open to other suggestions though.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -107,16 +134,32 @@ class EnvironmentEditor extends React.PureComponent<Props, State> {
...
           render={render}
           getRenderContext={getRenderContext}
           mode="application/json"
           {...props}
         />
-        {error && <p className="notice error margin">{error}</p>}
-        {!error && warning && <p className="notice warning margin">{warning}</p>}
+        {error && <p className="notice error margin-x margin-y-sm">{error}</p>}
+        {!error && warning && <p className="notice warning margin-x margin-y-sm">{warning}</p>}
+        <div className="form-control margin-x margin-y-sm">
```
gschier: My feeling is that we probably don't need to make this an option. It seems like maintaining order is expected no matter what. Thoughts?

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -69,11 +78,29 @@ class EnvironmentEditor extends React.PureComponent<Props, State> {
     this._editor = n;
   }
 
-  getValue() {
+  _updateMaintainOrderBoolean(checked: boolean) {
+    this.setState({ maintainOrder: !this.state.maintainOrder }, () => {
+      this.props.didChange();
```
gschier: Nice! I always forget setState has a callback param 👍

packages/insomnia-app/app/ui/components/modals/environment-edit-modal.js
```diff
@@ -30,10 +30,20 @@ class EnvironmentEditModal extends PureComponent {
       return;
     }
 
-    const environment = this._envEditor.getValue();
+    let patch;
```
gschier: Curious why you included the try/catch when there wasn't one before. Did you find an error case that wasn't handled?

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
```diff
@@ -342,9 +343,13 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
       return;
     }
 
-    let data;
+    let patch;
     try {
```
gschier: Oh I see. Same thing here. I just checked the `getValue` function and didn't realize it was doing JSON parsing. Makes sense now 👍

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
gschier: Nice! I bet that old code was written before `React.Fragment` existed 👀

packages/insomnia-app/app/ui/css/layout/base.less
```diff
@@ -554,6 +554,16 @@ i.fa.fa--fixed-width {
   margin: @padding-md;
 }
 
+.margin-x {
```
gschier: Nice 👍 Exactly the names I would have chosen too

packages/insomnia-app/app/ui/components/editors/environment-editor.js
develohpanda: Yeah fair - initial feature request mentioned for it to be optional but I don't see a particularly compelling use-case for not keeping order.

This checkbox is powered by the fact that a propertyOrder object exists or not - so its not a lot of extra work to remove. :)

develohpanda:
@gschier ready for review again :)

PR #1521, opened 2019-05-23
[WIP] Tweak Request API behavior for multi-value parameters & headers

### LINKED ISSUE #1518: [Bug] Request API is inconsistent when dealing with multi-value parameters
Say that we have multiple parameters with the same name. `request.parameters` looks like:
```json
[
    { "name": "multi", "value": "a" },
    { "name": "multi", "value": "b" },
    { "name": "multi", "value": "c" }
]
```

There are these problems when accessing this multi-value parameter with the existing `context.request.*Parameter` methods:

* `getParameter("multi")` will return `c`, because it only returns the _last_ value of a multi-value parameter
* `setParameter("multi", "d")` will replace the _first_ value, so...
* another call to `getParameter("multi")` will still return `c`
* and `addParameter("multi", "e")` will silently do nothing at all.

Furthermore, there are no methods that allow one to create a multi-value parameter, or to fetch it by name. (However, `getParameters()` does return the entire parameter collection.)

I have the beginnings of a pull request that fixes some of these problems, but they'd either require creation of new `request` methods or changing the behaviour of existing ones (though only changing the behaviour when dealing with multi-value parameters, and if we consider the current behaviour buggy then maybe that's OK?). @gschier: What are your preferences here? Are you OK with changing existing behaviour, or would you prefer something else?

yozlet:
Closes #1518, with the changes proposed by @gschier (unless discussion takes us elsewhere):

> * Change getParameter to return a string for single matches and an array for multiple
> * Change setParameter to remove all instances of key and replace with a single new one
> * Leave addParameter(name: string) as is, since it's working as expected. The method name might be a bit confusing, but the purpose of this one is to only set the parameter if one by that name does not yet exist.

### DIFF
--- a/packages/insomnia-app/app/plugins/context/__tests__/request.test.js
+++ b/packages/insomnia-app/app/plugins/context/__tests__/request.test.js
@@ -97,7 +97,13 @@ describe('request.*', () => {
         { name: 'hello', value: 'world' },
         { name: 'Content-Type', value: 'application/json' },
       ],
-      parameters: [{ name: 'foo', value: 'bar' }, { name: 'message', value: 'Hello World!' }],
+      parameters: [
+        { name: 'foo', value: 'bar' },
+        { name: 'message', value: 'Hello World!' },
+        { name: 'multi', value: 'a' },
+        { name: 'multi', value: 'b' },
+        { name: 'multi', value: 'c' },
+      ],
     });
   });
 
@@ -118,6 +124,9 @@ describe('request.*', () => {
     expect(result.request.getParameters()).toEqual([
       { name: 'foo', value: 'bar' },
       { name: 'message', value: 'Hello World!' },
+      { name: 'multi', value: 'a' },
+      { name: 'multi', value: 'b' },
+      { name: 'multi', value: 'c' },
     ]);
 
     // getParameter()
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/plugins/context/__tests__/request.test.js
```diff
@@ -97,7 +97,13 @@ describe('request.*', () => {
         { name: 'hello', value: 'world' },
         { name: 'Content-Type', value: 'application/json' },
       ],
-      parameters: [{ name: 'foo', value: 'bar' }, { name: 'message', value: 'Hello World!' }],
+      parameters: [
+        { name: 'foo', value: 'bar' },
```
gschier: Can we change the test data to the following?

```js
        { name: 'foo', value: 'bar' },
        { name: 'multi', value: 'a' },
        { name: 'message', value: 'Hello World!' },
        { name: 'multi', value: 'b' },
        { name: 'multi', value: 'c' },
        { name: 'trailing', value: 't' },
```

At first glance, it looks like your code does not preserve the position when using `setParameter` when multiple exist. 

What do you think the behavior should be in this case? 

- Never preserve position
- Replace first occurrence and delete the rest
- Replace last occurrence and delete the rest

Let me know what you think 🤔

packages/insomnia-app/app/plugins/context/__tests__/request.test.js
```diff
@@ -118,25 +124,38 @@ describe('request.*', () => {
...
     // getParameter()
     expect(result.request.getParameter('foo')).toBe('bar');
     expect(result.request.getParameter('FOO')).toBe(null);
     expect(result.request.getParameter('does-not-exist')).toBe(null);
     expect(result.request.hasParameter('foo')).toBe(true);
+    expect(result.request.hasParameter('does-not-exist')).toBe(false);
+    expect(result.request.getParameter('multi')).toBe('c');
 
-    // setHeader()
+    // setParameter()
```
gschier: Nice catch!

packages/insomnia-app/app/plugins/context/__tests__/request.test.js
yozlet: Honestly, I have no idea, and I'd be interested to hear arguments for any of them. In the meantime, maybe it'd be best to **explicitly**:
1. not make a decision here
2. say so in the API docs (as in something like "The parameter's position in the ordering may change. This behaviour is currently undefined, so don't depend on it as it may change in future.")

yozlet:
Sorry that this is taking a while! I have a load of work on right now, but I'd like to get back to this when I can. In the meantime, anyone who's interested should feel free to continue without me.

yozlet:
@gschier Talking of documentation (as I do in one of the threads above): I'm happy to update the API docs to go with this change, but I don't see them anywhere. Are they open to editing in some way, or should I leave that to you?

PR #1526, opened 2019-05-28
[OSX] Set titleBarStyle to hiddenInset. It's so pretty now! 🌟

mdeboer:
Insomnia looks great but it has this default grey title bar that looks awful when using most themes. So, as I like things to look pretty, I took the liberty of making some minor modifications (see screenshot) to the title bar.

I hooked in to the full-screen events to add/remove a class from the document body as in full screen the "traffic lights" are hidden, hence the workspace title should shift back to the left.

Please let me know what you think! 🎉 

Screenshot 2019-05-28 at 00 58 10

### DIFF
--- a/packages/insomnia-app/app/main/window-utils.js
+++ b/packages/insomnia-app/app/main/window-utils.js
@@ -59,7 +59,7 @@ export function createWindow() {
     height: height || DEFAULT_HEIGHT,
     minHeight: MINIMUM_HEIGHT,
     minWidth: MINIMUM_WIDTH,
-    titleBarStyle: titleBarStyle,
+    titleBarStyle,
     acceptFirstMouse: true,
     icon: path.resolve(__dirname, 'static/icon.png'),
     webPreferences: {
--- a/packages/insomnia-app/app/main/window-utils.js
+++ b/packages/insomnia-app/app/main/window-utils.js
@@ -45,6 +45,9 @@ export function createWindow() {
   const finalX = Math.min(maxX, x);
   const finalY = Math.min(maxX, y);
 
+  // Window frame
+  const titleBarStyle = isMac() ? 'hiddenInset' : 'default';
+
   mainWindow = new BrowserWindow({
     // Make sure we don't initialize the window outside the bounds
     x: finalX,
@@ -56,6 +59,7 @@ export function createWindow() {
     height: height || DEFAULT_HEIGHT,
     minHeight: MINIMUM_HEIGHT,
     minWidth: MINIMUM_WIDTH,
+    titleBarStyle: titleBarStyle,
     acceptFirstMouse: true,
     icon: path.resolve(__dirname, 'static/icon.png'),
     webPreferences: {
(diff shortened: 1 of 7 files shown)


packages/insomnia-app/app/main/window-utils.js
```diff
@@ -56,6 +59,7 @@ export function createWindow() {
     height: height || DEFAULT_HEIGHT,
     minHeight: MINIMUM_HEIGHT,
     minWidth: MINIMUM_WIDTH,
+    titleBarStyle: titleBarStyle,
```
vladanpaunovic: Could you please favor ES6 features for this (sorry my OCD :) )

```suggestion
    titleBarStyle,
```

vladanpaunovic (approved):
Other than one small syntax request I am very happy with this going out!
LGTM.

mdeboer:
All done, hope it gets merged soon!

gschier:
I'm torn on this one. I totally see why you would like this change but there are a couple things that prevent me from being on board.

1. The main issue is that Insomnia does not have a concept of a "titlebar" with the OS-level one removed. Apps with custom titlebars usually have a similar UI element that fits the same need (showing a title, some additional info, and maybe a few actions). Insomnia _does_ have a visual bar that runs across the top but, in my opinion, it's not a great replacement for the OS-level titlebar.

2. Similar to no. 1, not having a titlebar means there is nothing to grab to drag the window

So, while I _can_ see the reasons for a change like this, I don't think the pros are strong enough to outweigh the cons.

---

I would be much more comfortable providing a themeable replacement titlebar instead of removing the system one. This approach would benefit both Windows and Mac users too. I actually tried this a while ago but it was more work than I expected. Check out the PR for more details: https://github.com/getinsomnia/insomnia/pull/310

If someone wants to pick up where this PR left off I'd be happy to discuss or answer any questions.

mdeboer:
Imho it only serves to minimize, maximize and close, is the title really used? However you are right, it is also for dragging but afaik that still works. If not, it is very easily implemented by making the whole top bar draggable for example.

You can make it themeable but you'll accomplish the same thing basically. Except you have to handle the buttons yourself or at the very least on Windows. 

Fyi look at apps like [Atom](https://atom.io/) which have an option to make similar adjustments to the titlebar. Not sure they have it on Windows though. But they fully replicate a titlebar.

With the addition of a draggable region, would you consider merging?

gschier:
I still think the best way to go about it is to recreate the titlebar in a themable way, like VSCode does, re-implementing the buttons and menus from scratch so it can be completely themed to fit the app.

Paw is actually a good example of a custom titlebar done correctly in my opinion. It still operates like a stock titlebar by displaying the buttons, a title, and some extra global actions. I respectfully disagree that Paw's titlebar is similar to how Insomnia exists currently. Things in the titlebar should be global to the application. Editing a URL, as one example, is not a global action, and thus does not belong in the titlebar.

image

wohckcin:
@rmedaer @gschier  

Hey guys, just wondering if it's possible to reopen this issue. The current title bar doesn't look good in MacOS Big Sur, or at least can we have the option to enable/disable it?

Screen Shot 2021-01-29 at 10 22 39 am

PR #1543, opened 2019-06-03
Move request group to target workspace instead of duplicating

udkl:
Fixes : https://github.com/getinsomnia/insomnia/issues/1398

My first PR here.  
Suggestions are appreciated.

Still trying to figure out how to record a screen-cast on linux.


Gabb1995:
if you mean a screen recorded on linux try this https://github.com/phw/peek never had any problems with it super simple to use.

PR #1544, opened 2019-06-03
Download path settings per request

### LINKED ISSUE #1511: [Feature Request] Store request sending settings locally (not globally)
- Insomnia Version: 6.5.3
- Operating System: Windows 10 (Build 1903)

## Details

Currently Insomnia stores request settings like ''send on interval' globally for all requests.
It would be better and also allow more use cases if this were per request.

Quoting @gschier 

> It wouldn't be difficult, no. You'd have to create a new property on the `RequestMeta` model to store the preference, then change the sending code to use that instead of the globally-stored value

See following files:
https://github.com/getinsomnia/insomnia/blob/92641f872a616a2a85b1c9e295197a57ac8d0842/packages/insomnia-app/app/models/request-meta.js#L20

https://github.com/getinsomnia/insomnia/blob/92641f872a616a2a85b1c9e295197a57ac8d0842/packages/insomnia-app/app/ui/components/request-url-bar.js#L128

develohpanda:
Closes #1511.

Saves request download path in `request-meta` instead of `request-url-bar` state. This allows the setting to be per request rather than global.

Ideally I would expect this `downloadPath` setting (and others in future) to exist in a 'requestSettings` object in meta and passed around as a whole rather than adding x props for each new setting. I feel that is out of scope for this PR though - thoughts?

Send + download
image

Send only
image

### DIFF
--- a/packages/insomnia-app/app/models/request-meta.js
+++ b/packages/insomnia-app/app/models/request-meta.js
@@ -18,6 +18,7 @@ type BaseRequestMeta = {
   savedRequestBody: Object,
   pinned: boolean,
   lastActive: number,
+  downloadPath: string | null,
 };
 
 export type RequestMeta = BaseModel & BaseRequestMeta;
@@ -32,6 +33,7 @@ export function init() {
     savedRequestBody: {},
     pinned: false,
     lastActive: 0,
+    downloadPath: null,
   };
 }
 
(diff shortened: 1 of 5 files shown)


gschier:
> Ideally I would expect this downloadPath setting (and others in future) to exist in a 'requestSettings` object in meta and passed around as a whole rather than adding x props for each new setting. I feel that is out of scope for this PR though - thoughts?

Yes, I agree. It would be nice for the future but out of scope for this PR.

develohpanda:
> 
> 
> Thanks for the pull-request @develohpanda - can't wait to test this out

No worries mate!

PR #1584, opened 2019-07-02
Fix file not being selected on content-type change

### LINKED ISSUE #1542: [Bug] File won't be selected when accepting Content-Type suggestion
- Insomnia Version: 6.5.4
- Operating System: win10 x64 1903

## Details

Video: https://streamable.com/4nl75

After I click "Yes" in the "Change Content-Type" dialog, the file won't be selected as expected.

brendano86:
Fix race condition on setting request body and headers to ensure the file is selected after content-type change

Closes #1542


brendano86:
@gschier thoughts?

PR #1587, opened 2019-07-06
[network] Use system CA on Linux

### LINKED ISSUE #1432: [improvement] Use OS CA certs when possible
- Insomnia Version: 6.3.2.1

## Details

Ref: 

captn3m0:
- Closes #1432 
- Uses https://www.npmjs.com/package/find-root-ca-cert, which is based on [the list that golang uses](https://golang.org/src/crypto/x509/root_linux.go)

### DIFF
--- a/packages/insomnia-app/app/network/network.js
+++ b/packages/insomnia-app/app/network/network.js
@@ -12,6 +12,7 @@ import {
 import mkdirp from 'mkdirp';
 import crypto from 'crypto';
 import clone from 'clone';
+import certFinder from 'find-root-ca-cert';
 import { parse as urlParse, resolve as urlResolve } from 'url';
 import { Curl } from 'insomnia-libcurl';
 import { join as pathJoin } from 'path';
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/network/network.js
```diff
@@ -353,7 +354,11 @@ export async function _actuallySend(
 
       // Setup CA Root Certificates if not on Mac. Thanks to libcurl, Mac will use
       // certificates form the OS.
-      if (process.platform !== 'darwin') {
+      if (process.platform === 'linux') {
+        const fullCAPath = certFinder.findCABundle();
```
gschier: Will we still need to fallback to the bundled CAs if `findCABundle` does not return a value?

gschier:
Oh, and for the tests, you just need to add a Flow definition for the new module in `/packages/insomnia-app/flow-typed/`

packages/insomnia-app/app/network/network.js
captn3m0: I think this should work across all desktop linux distros, but I can re-route to force the local cert as a fallback.

Can you point me to the debian/span recipes? Want to ensure that `ca-certificates` is installed in both of these, and the cert-finder works.

gschier: Can you explain what you mean by recipes?

captn3m0:
I meant the packaging scripts used to generate the Deb and snap packages.

packages/insomnia-app/app/network/network.js
captn3m0: I meant the packaging scripts used to generate the Deb and snap packages.

PR #1597, opened 2019-07-09
[Prompt] Optionally make the prompt plugin pre-fill the last entered value

mdeboer:
I have added an option to the prompt plugin to save the last entered value and have that set as default value the next time the prompt shows. This saves you from having to enter all details for each request while still being able to change them if needed.

This should be full backwards compatible as the old behaviour is used when a storage key is set explicitly.

### Old scenario 😢 
1. User executes a request with a few prompt tags
2. User enters details for each of them
3. User sips some coffee, fixes some bugs in his code.
4. User executes the request again
5. User has to enter all details again and make sure they are exactly the same
6. Repeat from step 3

### New scenario 😄 
1. User executes a request with a few prompt tags
2. User enters details for each of them
3. User sips some coffee, fixes some bugs in his code.
4. User executes the request again
5. Bashes enter a few times as all things are pre-filled
6. Repeat from step 3

Makes my life a lot easier 😅

### DIFF
--- a/plugins/insomnia-plugin-prompt/index.js
+++ b/plugins/insomnia-plugin-prompt/index.js
@@ -1,23 +1,5 @@
 const crypto = require('crypto');
 
-module.exports.responseHooks = [
-  async context => {
-    const items = await context.store.all();
-    const requestId = context.request.getId();
-    const toRemove = items.filter(v => {
-      if (v.key && typeof v.key === 'string') {
-        return v.key.indexOf(requestId) === 0;
-      }
-
-      return false;
-    });
-    // Delete cached values we prompt again on the next request
-    for (const { key } of toRemove) {
-      await context.store.removeItem(key);
-    }
-  },
-];
-
 module.exports.templateTags = [
   {
     displayName: 'Prompt',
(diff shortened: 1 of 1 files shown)


gschier:
Hmm, Insomnia already has the ability to store the last value using the "Storage Key" field. 

I like your PR but, since Insomnia already offers such similar functionality, I'm hesitant to merge this one. Can you describe a bit more how they are different or how they might work together?

mdeboer:
I tried as best as I could but okay. I know very well Insomnia has that ability but this is *not* the same.

When you use the "storage key" method, Insomnia will *not* prompt again until you change the key. I'm sure there are advanced use cases but to me it is no use.

This plugin offers an option to remember the last entered value and *will* prompt the user every time, just like a regular prompt instance. Except this plugin will pre-fill the value field with the last entered value. No need to change keys to change your value etc.

This plugin should be fully backwards compatible and all the old functionality is still there. I set it to enabled by default but I can change this to disabled if you want.

gschier:
Thanks for the explanation! I noticed you weren't actually using the variable you added inside the render method, so I did it on my end: https://github.com/getinsomnia/insomnia/commit/685c5228866d357ae91ad2b677245f31550ff432

PR #1637, opened 2019-08-11
[Bug Fix 1288] "Undefined" for nested variables

### LINKED ISSUE #1288: [Bug] Variable uncovering returns "undefined" for nested variables
- Insomnia Version: 6.2.3
- Operating System: Linux Mint

## Details

https://github.com/getinsomnia/insomnia/pull/1274 introduces variable uncovering (helpful), but it seems that nested variable references returns `{undefined}:<variable>`

For example use the following environment:

````
{
  "foo": {
    "bar": "foobar"
  },
  "fu": {
    "baz": "{{ foo.bar  }}"
  },
  "foobar": "foobar2",
  "fubaz": "{{foobar}}"
}
````
- Hover over the `fu.baz` value ( `{{ foo.bar }}` ), and notice it shows `{undefined}:foobar`
- Hover over the `fubaz` value ( `{{ foobar }}` ), and notice it shows `{root}:foobar2`

Expectations:

- Hovering over the value shows `{{root}}:foobar` like a non-nested value does.

Other thoughts:
 - Should the `root` reference instead say "Base Environment" to correspond to the Base Environment?

### PR DESCRIPTION
**Pull request by Gabrz**
(Fixes #1288) Extract parent from nunjucks tag when it is a nested property nested. The parent is then used to fetch the keyContext

### DIFF
--- a/packages/insomnia-app/app/common/render.js
+++ b/packages/insomnia-app/app/common/render.js
@@ -243,6 +243,11 @@ export async function getRenderContext(
       if (Object.prototype.toString.call(subObject[key]) === '[object Object]') {
         // Type is an Object, keep on going, recursively building the full key path
         getKeySource(subObject[key], inKey + key + '.', inSource);
+      } else if (Object.prototype.toString.call(subObject[key]) === '[object Array]') {
+        // Type is an Array, Loop and store the full Key and Source in keySource
+        for (let i = 0, length = subObject[key].length; i < length; i++) {
+          keySource[inKey + key + '[' + i + ']'] = inSource;
+        }
       } else {
         // For all other types, store the full Key and Source in keySource
         keySource[inKey + key] = inSource;
--- a/packages/insomnia-app/app/common/render.js
+++ b/packages/insomnia-app/app/common/render.js
@@ -237,18 +237,27 @@ export async function getRenderContext(
   const subEnvironment = await models.environment.getById(environmentId || 'n/a');
 
   let keySource = {};
-  for (let key in (rootEnvironment || {}).data) {
-    keySource[key] = 'root';
-  }
-
-  if (subEnvironment) {
-    for (const key of Object.keys(subEnvironment.data || {})) {
-      if (subEnvironment.name) {
-        keySource[key] = subEnvironment.name;
+  // Function that recursively gets keys
+  function getKeySource(subObject, inKey, inValue) {
+    for (let key in subObject) {
+      if (typeof subObject[key] === 'string') {
+        keySource[inKey + key] = inValue;
+      } else if (typeof subObject[key] === 'object') {
+        getKeySource(subObject[key], inKey + key + '.', inValue);
       }
     }
+    return keySource;
   }
 
+  // Get Keys from root environment
+  getKeySource((rootEnvironment || {}).data, '', 'root');
+
+  // Get Keys from sub environment
+  if (subEnvironment) {
+    getKeySource(subEnvironment.data || {}, '', subEnvironment.name || '');
+  }
+
+  // Get Keys from ancestors
   if (ancestors) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/insomnia-app/app/common/render.js [hunk 1]
```diff
@@ -237,18 +237,27 @@ export async function getRenderContext(
   const subEnvironment = await models.environment.getById(environmentId || 'n/a');
 
   let keySource = {};
-  for (let key in (rootEnvironment || {}).data) {
-    keySource[key] = 'root';
+  // Function that recursively gets keys
+  function getKeySource(subObject, inKey, inValue) {
+    for (let key in subObject) {
```
gschier: Please use `const key of Object.keys(subObject)` to prevent iterating over the object prototype.

packages/insomnia-app/app/common/render.js
```diff
@@ -237,18 +237,27 @@ export async function getRenderContext(
...
+      if (typeof subObject[key] === 'string') {
+        keySource[inKey + key] = inValue;
+      } else if (typeof subObject[key] === 'object') {
```
gschier: Unfortunately `typeof null === 'object'` is `true` as well. Similar to `common/render.js`, these checks should use a more reliable method. https://github.com/getinsomnia/insomnia/blob/5ce2825ea94ed4a59cc6b0011776b43b292839c8/packages/insomnia-app/app/common/render.js#L158

Specifically, they should be changed to the following:

```js
typeof subObject[key] === 'string'
typeof subObject[key] === 'object'

// Should use the following, more reliable, checks instead

Object.prototype.toString.call(subObject[key]) === '[object String]'
Object.prototype.toString.call(subObject[key]) === '[object Object]'
```

packages/insomnia-app/app/common/render.js
```diff
@@ -237,18 +237,27 @@ export async function getRenderContext(
   const subEnvironment = await models.environment.getById(environmentId || 'n/a');
 
   let keySource = {};
-  for (let key in (rootEnvironment || {}).data) {
-    keySource[key] = 'root';
+  // Function that recursively gets keys
+  function getKeySource(subObject, inKey, inValue) {
+    for (let key in subObject) {
+      if (typeof subObject[key] === 'string') {
```
gschier: What happens if it's a number, array, boolean, or null?

packages/insomnia-app/app/common/render.js
```diff
@@ -237,18 +237,27 @@ export async function getRenderContext(
...
   if (subEnvironment) {
-    for (const key of Object.keys(subEnvironment.data || {})) {
-      if (subEnvironment.name) {
-        keySource[key] = subEnvironment.name;
-      }
-    }
+    getKeySource(subEnvironment.data || {}, '', subEnvironment.name || '');
   }
 
+  // Get Keys from ancestors
```
gschier: Nice work leaving a comment! 😄

packages/insomnia-app/app/common/render.js
```diff
@@ -257,9 +266,7 @@ export async function getRenderContext(
         ancestor.hasOwnProperty('environment') &&
         ancestor.hasOwnProperty('name')
       ) {
-        for (const key of Object.keys(ancestor.environment || {})) {
-          keySource[key] = ancestor.name || '';
-        }
+        getKeySource(ancestor.environment || {}, '', ancestor.name || '');
```
gschier: Awesome. I like that this logic is grouped together in one place now.

packages/insomnia-app/app/common/render.js
Gabrz: Thank you for your feedback! I usually don't program in JavaScript, it helped a lot that you specifically stated what you would like to have changed.

I really like Insomnia, and it would be awesome if my contribution made it into the build!

PR #1688, opened 2019-09-27
Escape dirs in the yarn cmd

### LINKED ISSUE #1355: Space in installation directory breaks plugin installation on Windows
- Insomnia Version: 3.6.2

## Details

When trying to install a plugin from the preferences window you will be presented an error when the installation directory contains a space. This will probably happen often because Windows uses your `Firstname Lastname` tuple as directory name for your home-directory and the insomnia installer installs into your `AppData` directory. (But even if it did install to `Program Files` or `Program Files (x86)` the same would happen)

Screenshot:
grafik

Text of error message:
```
insomnia-plugin-default-headers npm error: Command failed: "C:\Users\Johannes Schriewer\AppData\Local\insomnia\app-6.3.2\Insomnia.exe" --no-deprecation C:\Users\Johannes Schriewer\AppData\Local\insomnia\app-6.3.2\resources\bin\yarn-standalone.js info insomnia-plugin-default-headers --json
internal/modules/cjs/loader.js:604
    throw err;
    ^

Error: Cannot find module 'C:\Users\Johannes'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:602:15)
    at Function.Module._load (internal/modules/cjs/loader.js:528:25)
    at Function.Module.runMain (internal/modules/cjs/loader.js:753:10)
    at startup (internal/bootstrap/node.js:241:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:579:3)
```

If you correctly quote the path to the `yarn-standalone.js` it seems to work.

v1ktor:
Fixes #1355 

The yarn doesn't understand spaces in path, so they have to be escaped.


v1ktor:
I have downloaded 7.0.0 update today and unfortunately It didn't fully fix the issue on windows. The `_getYarnPath()` also should be escaped, but I have missed that since yarn-standalone was installed in the path without space on my local env :(

I will verify and do another PR soon.

PR #1690, opened 2019-09-28
Fix transparent line-numbers background

### LINKED ISSUE #1647: Line numbers overlap JSON body
- Insomnia Version: **v6.6.2**
- Operating System: **MacOS High Sierra 10.13.2**

## Details

1. Turn off body text wrapping
1. Scroll long body horizontally, so that line numbers start to overlap it
1. Observe:

Screen Shot 2019-08-22 at 11 49 26 AM

## Proposal

Would be nice to have some 80% opaque background under line numbers, so that they're distinct.

ranjan-purbey:
Closes #1647

### DIFF
--- a/packages/insomnia-app/app/ui/css/editor/general.less
+++ b/packages/insomnia-app/app/ui/css/editor/general.less
@@ -71,7 +71,6 @@
 
   .CodeMirror,
   .cm-s-seti.CodeMirror, // Hack because seti theme is dumb
-  .CodeMirror-gutters,
   .CodeMirror-scrollbar-filler,
   .CodeMirror-gutter-filler {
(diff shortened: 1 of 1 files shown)


ranjan-purbey:
I know, but the thing is background color is stored in a CSS3 variable and not a less variable. So the fade function won't work, and CSS by itself doesn't provide a way to achieve this. Is there any specific reason to add transparency?

arbitrary-dev:
Just for coolness) Aren't there a way to assign CSS3 variable to some temporary LESS variable and then apply `fade()` on it?

ranjan-purbey:
No, that can't be done

PR #1723, opened 2019-10-11
[Feature] [SwaggerV2] Create Folders from implicit defined tags

beardeddragon5:
Swagger Tags can be defined in the document root 
```yaml
...
tags:
  - name: tagname
    description: tagdescription
...
```
or implicit in every method
```yaml
...
get:
  operationId: "getsomething"
  tags:
    - tagname
...
```
The implicit variant can not define an description.

With this little change the implicit tags get imported as folders as well


beardeddragon5:
Yes, i can add a few test fixtures. The tests didn't run correctly on my side and always returned "passed". Not ideal to write tests.

PR #1727, opened 2019-10-13
add feature of expand and collapse folder

### LINKED ISSUE #480: [Feature Request] Expand/collapse All for tree view
- Insomnia Version: 5.8.2
- Operating System: Mac OS X

## Detaills
A possibility to expand and collapse all (shown) folders in the tree view with one click (or shortcut) would be very helpful. It could also be automatically applied after filtering, as currently sometimes only closed folders with hidden requests are shown. The "Expand all" feature could directly show all found requests. And "Collapse all" would close all folders again. Small buttons could be placed obove the tree view or also in the menu bar ("View"). Maybe it could be configurable in the preferences whether filtered folders are expanded directly.

maximilianoh:
Hi!

I would like a feature that expand/collapse All folders. I saw that it has not had recent activity, so I did. I put  ctrl + uparrow like shortcut. I do not know if that is such a good idea, but it is intuitive. I wait the review.

Thanks in advance for this awesome App!

Closes #480

scene00001


packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -395,12 +401,20 @@ class App extends PureComponent {
   }
 
   static async _updateRequestGroupMetaByParentId(requestGroupId, patch) {
-    const requestGroupMeta = await models.requestGroupMeta.getByParentId(requestGroupId);
-    if (requestGroupMeta) {
-      await models.requestGroupMeta.update(requestGroupMeta, patch);
+    if (requestGroupId === 'all') {
+      const allData = await models.requestGroupMeta.all();
```
gschier: Note, this query will get _all_ request groups, even ones in other workspaces, which isn't desired.

gschier (changes requested):
I like where this is headed! A couple things regarding UI/UX though:

- Clicking expand/collapse within a folder's submenu should only expand/collapse the nested folders within it
- The expand/collapse icon should probably be different than the folder one. The closest fit I could find is `fa-expand`
- For the root-level, expand-collapse does not make sense in that menu, since the menu is for adding things (hence the "+" icon). I'm not sure where else it could go, though, so I think it'd be fine if you keep in there, but add a `<DropdownDivider>Actions</DropdownDivider>` above it

maximilianoh:
Hi!

- Clicking expand/collapse within a folder's submenu should only expand/collapse the nested folders within it.  

     **_I did not think about that before. It is a great idea._**

- The expand/collapse icon should probably be different than the folder one. The closest fit I could find is fa-expand

     **I am not sure if "fa-expand" is intuitive design. I did not see a specific icon for collapse button. I see a "not icon" for collapse button. For that I left the folder icon.**

- For the root-level, expand-collapse does not make sense in that menu, since the menu is for adding things (hence the "+" icon). I'm not sure where else it could go, though, so I think it'd be fine if you keep in there, but add a DropdownDivider Actions DropdownDivider above it.

     **Beautiful.** 


- Note, this query will get all request groups, even ones in other workspaces, which isn't desired.

     **It is true. I do not like that, but I did not see or I did not understand other way. Have you a method that can update all "patchs"? Like models.requestGroupMeta.updateAll or something else.**

packages/insomnia-app/app/ui/containers/app.js
gschier: You might be able to do something like this (have not tested).

1. Get all descendants of the request group
2. Filter descendants for only RequestGroup type
3. Update each RequestGroup

```js
// Untested code!

const requestGroup = await models.requestGroup.getById(requestGroupId);
const descendants = await db.withDescendants(); // will return RequestGroup, Request, etc models too
const onlyRequestGroups = descendants.filter(d => d.type === models.requestGroup.type);
onlyRequestGroups.map(async doc => {
  await models.requestGroupMeta.update(doc, patch);
});
```

gschier (changes requested):
What you mentioned about the icon is fine. Leaving as folder for now is okay.

I added a suggestion for solving the DB query problem.

maximilianoh:
scene00001
I fix the workspace problem, expand/collapse the nested folders and add the `<DropdownDivider>Actions</DropdownDivider>`. 

I used your code. Only change this:
`descendants.filter(d => d.type === models.requestGroup.type);`
For
`descendants.filter(d => d.type === models.requestGroupMeta.type);`

Because requestGroup not have collapse key- value.

nijikokun:
Thoughts about adding the word `All` to the action to provide a clearer expectation on what will happen when you click that option?

maximilianoh:
> Thoughts about adding the word `All` to the action to provide a clearer expectation at the root level?

Hi. Yes, maybe is not the best name that I could put, it can be confusing. But, that is the parameter for all folders. If `all` is false, this mean that, only the clicked folder  expand/collapse.
This is for the recomendation of gschier: 'Clicking expand/collapse within a folder's submenu should only expand/collapse the nested folders within it.'
Perhaps, I could change for other word that can manifest what it do, something like 'allFolder' or 'allFolderWorkspace' .

gschier (changes requested):
I think I'm happy with this PR in terms of functionality and UI. However, this section of code feels overly complex and doesn't really fit in this location (`_updateRequestGroupMetaByParentId` shouldn't be responsible for collapse/expand logic).

If you can figure out a way to simplify improve this, I'd be happy to take another look.

maximilianoh:
> I think I'm happy with this PR in terms of functionality and UI. However, this section of code feels overly complex and doesn't really fit in this location (`_updateRequestGroupMetaByParentId` shouldn't be responsible for collapse/expand logic).
> 
> If you can figure out a way to simplify improve this, I'd be happy to take another look.

Yes, I see this code, again, and it is very complex in comparison with others _(functions).
I don't remember all the filter, but I think that I need all filters, maybe I can remove 2 o 3 lines.

I can create a function that do the "logic" filter by type, meta and, last step of this process, update. _updateRequestGroupMetaByParentId could call this function and left _updateRequestGroupMetaByParentId  with a short if and else. 

Can you give me some suggestions about a good place for this (if you agree), please? 
Or,
Do you have a better idea? I don't remember all and you know the code better  than I do.
Thanks!

gschier:
We really appreciate the effort on this one, but we're going to close this one out, as we don't currently have the bandwidth to help get it over the line 💜

mrnasil:
Any update Windows or Linux App?

mrfy:
Please come back to this task. Thanks in advance :)

PR #1729, opened 2019-10-13
Fetch new access tokens if refreshing token returns "invalid_grant"

Lavoaster:
I know the PR says to open an issue first, but I figured this could be the discussion as I've already done part of the work.

At my current workplace we have an oauth server with access tokens on a 5 min TTL and refresh tokens on a 30 min TTL. So every so often I'll see the following message.

Screenshot 2019-10-13 23 48 56

It's rather annoying to have to go into the OAuth page to clear the tokens and refresh it again. Given Insomnia automatically handles fetching the access / refresh. I'm sure it should be able to ignore this and fetch new tokens if it encounters this error.

So, as per the oauth spec [here](https://tools.ietf.org/html/rfc6749#section-5.2), I've updated the refreshToken function to return the magic response that will trigger a new token to be fetched if the server sends back `400` status, and the `error` in the json body is `invalid_grant`. 

I'm not sure how to go about testing this, so if you want test point me in the right direction and I'll take a shot.

### DIFF
--- a/packages/insomnia-app/app/network/o-auth-2/refresh-token.js
+++ b/packages/insomnia-app/app/network/o-auth-2/refresh-token.js
@@ -48,6 +48,7 @@ export default async function(
   });
 
   const statusCode = response.statusCode || 0;
+  const bodyBuffer = models.response.getBodyBuffer(response);
 
   if (statusCode === 401) {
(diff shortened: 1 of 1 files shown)


Lavoaster:
@gschier Have you had a chance to look at this or consider if this is a feature you want in insomnia?

PR #1753, opened 2019-10-25
Add clickhandler for links and back/forward button

### LINKED ISSUE #874: [Feature Request] Allow opening links in response in Insomnia
I prefer Insomnia to Postman in most respects, but one feature I use a lot in Postman, and really miss, is the ability to follow links in the body of one request within Postmman.

I do a lot of work with an API that represents a tree structure with one node per request. The bodies of these requests contain URLs to their child node locations, so the ability to simply click one and have it followed immediately is extremely useful.

When I clink a link in the body of a request in Insomnia, it opens the link in my web browser. If I right click the link I just get an option to copy it. The only way I can see to open it in Insomnia is to copy the link, duplicate the current request (for headers, see below) and paste in the link I want to follow. This is fine to do occasionally, but its very frustrating if you regularly need to follow a lot of links.

For this to be really useful, there needs to be a way to propagate headers (mainly for auth, a custom API key header in our case). Postman handles this by allowing the current requests headers to be copied to the new request of the followed link. Another solution would be to allow setting common headers for a group of requests since not all headers are always required to be copied. I saw something along those lines suggested in another feature request.

I wouldn't want it to follow the link in the current request though. Opening it up in a new request is definitely the way to go. 👍

Thanks!

dschulten:
Closes #874 

This PR allows to follow links in responses.

Changes:
* Adds clickhandler for links that executes follow-up request within the context of the current request. I have implemented the actual handling in app.js _handleSendFollowUpRequestWithEnvironment, passed it down as _handleSend to response-viewer.js _handleOpenLink and use a separate followupRequest object parameter to network.js send to distinguish initial from followup requests.
* You get a history with back/forward buttons and a current address above the response viewer
* The history is stored together with the request definition and will be available across restarts
* If a historic response is a POST, clicking back will ask you if you want to repeat the POST
* If a historic response is the result of a redirect, the history stores the redirect target to support the [PRG ](https://en.wikipedia.org/wiki/Post/Redirect/Get)pattern
* If the user goes back in history and clicks a link there, the history will restart from there, i.e. the history will be truncated at that point and continue with the new clicked link ("restartMidHistory")
* The history descriptor has a list of requests and responses because the historic url to dereference is sometimes the one in the response, not the request itself. This allows us to take care of redirects.

Example: coffeeshop API:

image

After clicking the link to the product "Latte Macchiato":

image

# Decisions
1. Follow-Up Requests
We decided to implement a click on a link as a follow-up request to the current request, and execute it in the context of the current request.

    * One alternative was to create and store new request definitions in Insomnia for every click, and then open them, or to introduce interim tabs for every followup request.
Drawback: Doing this everytime would pollute the request list or create many tabs. Conceptually, the HATEOAS principle would be undermined - every followup request would become a new application state without connection to the previous state. Implementing back and forward would be difficult.
A context menu on links can easily be added later, which could allow to create and store a new request from an interesting url (effectively, a bookmark function)
    * Another alternative was to update the url of the current request definition and repeat the request.
Drawback: the original request would be lost.

2. Separate Browser Address Bar
We decided to implement a separate address bar in the response view with back/forward buttons to reflect the current application state. The response view becomes an actual browser. The request pane stays the request definition pane.
Drawback: Having two urls might confuse people. On the other hand, it should be clear even to an infrequent Insomnia user that the request pane is really the request *definition* pane. Changes to the url in the request pane are permanent. The request pane is not "like a browser" at all.
Currently, the address bar in the response view is read-only, to keep the UX simple.

# Future Enhancements
1. Support for UriTemplates
Support for UriTemplates is underway: https://github.com/dschulten/insomnia/tree/feature/follow-rfc6570-uritemplates. That feature will allow users to enter parameter values for UriTemplates if an API supports that:
image
It works already, but support for [composite values](https://tools.ietf.org/html/rfc6570#section-2.4.2) could be improved by allowing a multiline editor on composite params.

2. Address Bar
Make it editable, bookmarkable, add refresh button, back/forward context menu shows stack of historic requests

3. Create new request definitions from URLs
A context menu action on hyperlinks should use information from the current context to create a new request definition, including header information for media-type, authentication etc.

4. There are self-describing response media-types which could allow to improve the browsing experience considerably, e.g. Collection+JSON, Siren or Hal-Forms, furthermore ALPS. The correct way to approach this would probably be plugins for each kind of api description, which understand the descriptions of the possible interactions and would make it much easier to work with these apis.
An experimental feature branch demonstrates how this works with the Siren mediatype: https://github.com/dschulten/insomnia/tree/feature/add-affordance-plugins

5. Support rfc5988 link headers: make link headers accessible in response viewer


gschier:
Hey @dschulten!

I'm super happy that you've decided to work on this. It's been a long-requested feature.

Before I do a full review, can you update your PR description to include a description of the high-level changes you've done, decisions you made, and also screenshots of new UI/UX that you've added?

Thanks 🙂

gschier:
First of all, thanks for putting the time in to submit this PR! 

Before looking too closely at your code, I think it's worth first discussion things at a higher level so we can get to a good place. Since this is such a large change I suggest we keep this PR as a "proof of concept" and then, once we decide on a concrete direction, break it up into smaller pieces to review and merge separately.

## High-level Feedback

I played around with your PR on my machine and have some feedback. 

**UI Implementation**
Since most Insomnia users will never click a link within a response, we should not have this feature impact their workflow. I suggest only showing the extra history UI once a link has been clicked and a history exists.

**User Experience**
My main question here is around history navigation. Can you explain the use for resending requests when navigating backward/forward in the history? Can we get away with simply showing the existing response instead of resending it?

**Navbar Row Appearance**
I'm not a huge fan of the way the extra nav row looks. I did a small change on my end to try something different but it seems like the back/forward might get confused with the history dropdown on the right in this case since they're on the same visual level.

image

## Moving Forward

As mentioned, I think a good plan of attack here is to use this PR to implement a rough outline and implementation, then break it up into smaller PRs. I think this would be a good strategy:

1. Discuss these high-level points on this PR
2. Implement minor tweaks without worrying about code quality
3. Decide on high-level implementation architecture (DB models, logic, etc)
4. Decide how to break this up into smaller PRs if possible

dschulten:
> I suggest only showing the extra history UI once a link has been clicked and a history exists.

Fine with me. That most users will not use the links does not immediately convince me, after all many seem to have requested this, but your solution seems a good compromise. Actually, I had the navigation in the existing row initially, but back then you gave me the idea to introduce an extra row for the buttons and the URL: https://github.com/getinsomnia/insomnia/issues/874#issuecomment-394822941. The advantage of the extra row is that there is enough space for most links and additional ui elements to create bookmarks, repeat the request etc. An extra row might be easier to add visually after the user clicked a link, we just have to make it ease in.

> Can you explain the use for resending requests when navigating backward/forward in the history? Can we get away with simply showing the existing response instead of resending it?

It is what the browser does. The browser also employs its cache but it seems libcurl does not support caching. Using a stored response history means that the persisted data potentially can become large (when can we evict them?), also we would be in the business of caching, so we should also honor cache headers like Last-Modified  and No-Cache. My expectation when using the back-forward buttons is that the responses are not stale, but up to date. I am not against caching, but that seems to be a big feature in its own right.
If we repeat the request, the content of the timeline is obvious: it is the timeline of what you see.
So I would vote to store the minimum information to repeat the requests in the history and repeat them when browsing the history. I'd rather think about caching later.

> I did a small change on my end to try something different but it seems like the back/forward might get confused with the history dropdown on the right in this case since they're on the same visual level.

Back then I think the concern that users might mistake the timeline for a browsing history lead you to propose a second row, if I read you correctly: https://github.com/getinsomnia/insomnia/issues/874#issuecomment-394524129
I'm fine with both solutions.

> Discuss these high-level points on this PR
The branch https://github.com/dschulten/insomnia/tree/feature/add-affordance-plugins shows what I ultimately have in mind, along the line of future enhancements. It already works and it was fairly easy to implement, so kudos to your architecture :)
Maybe we should include these ideas in the discussion, in the sense of a possible roadmap.

> Implement minor tweaks without worrying about code quality
I would like to reduce the amount of data stored in the history and come up with a good way to test the linking and history feature. 
What would you change?

> Decide on high-level implementation architecture (DB models, logic, etc)
I sense that you have something in mind here. Tell me more ;)

> Decide how to break this up into smaller PRs if possible
Of course.

nijikokun:
> I'm not a huge fan of the way the extra nav row looks. I did a small change on my end to try something different but it seems like the back/forward might get confused with the history dropdown on the right in this case since they're on the same visual level.

What if the back/forward & url box goes under the Preview tab, between the tabbing component and the code preview. It could show once you navigate.

gschier:
Due to this being a rather large and impactful change, we don't have the bandwidth to help get this one over the line in a way we feel comfortable with. I'm going to close this PR but we appreciate the effort that went into it 💜

PR #1761, opened 2019-11-01
[Feature] Added response history grouping by time

### LINKED ISSUE #831: [Feature Request] Response history pop-up - group in dates.
- Insomnia Version: 5.14.9
- Operating System: Ubuntu 16.04

## Details

This is a feature request to improve "Response history" pop-up. At the moment it shows:
Past Responses list with: [Status Code] [URL] [Time] [Size] which is fine but I would like it to also show time when the response has been made.

Current view:
historypopup

So instead of flat list, it could be grouped in date ranges, e.g.:
[Today - Tuesday 27 Mar]
Entry
Entry
[Yesterday - Monday 26 Mar]
Entry
Entry
[Earlier this week]
Entry
Entry 

I use responses history a lot, and such an improvement would make it much easier to find which request from the history am I looking for.

NilayMajorwar:
Modifies the response history dropdown to group the responses into time groups, as shown:

res-history

Closes #831

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
 import autobind from 'autobind-decorator';
+import moment from 'moment';
 import { Dropdown, DropdownButton, DropdownDivider, DropdownItem } from '../base/dropdown';
 import StatusTag from '../tags/status-tag';
 import URLTag from '../tags/url-tag';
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.js
```diff
@@ -82,6 +83,41 @@ class ResponseHistoryDropdown extends React.PureComponent<Props> {
...
+        categories[2].push(r);
+      } else if (now.isSame(resTime, 'week')) {
+        // This week
+        categories[3].push(r);
+      }
+    });
+
+    return (
+      <React.Fragment>
+        <DropdownDivider>5 Minutes Ago</DropdownDivider>
```
gschier: Should we also make it so a divider doesn't show up if there are no responses under it?

NilayMajorwar: The divider wouldn't show up if there are no responses under it - the dropdown component does not render "empty" dividers ([dropdown.js:338](https://github.com/getinsomnia/insomnia/blob/414e35ea112e4df3bf47b0f8f7bc3229330cd2bd/packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js#L338))

Should I make empty time groups show up (with "No responses" text or something)?

gschier: Nope, this is perfect! I totally forgot dropdown dividers worked that way.

PR #1789, opened 2019-11-17
Add description field for body entries

### LINKED ISSUE #1673: [Feature Request] – Description field for body entries
Before coming to Insomnia I used Postman, which at some point got really really slow and thus wasted a lot of my time. After trying to fix it and reinstalling which all didn't help I decided to move on to a new REST client. Ever since I'm super happy with Insomnia!

However there's one feature I miss here, which I found really useful in Postman, which is a small description input field in the body section next to each entry, to save a small note for each body entry. 
This is especially useful for my case, as I need to send a lot (!) of different requests to the API I'm developing, e-mails to be exact, and I obviously always loose track of which entry in the body section contains which mail. A note at the side would therefore save me a lot of time!!

I'm sure this feature would be useful for a lot of other users / use cases too.

Let me know what you think of it. If this feature request gets accepted I am more than happy to contribute this feature myself.

pachisi456:
Closes #1673 

Besides key and value the user can provide a description to the
according body entry to better keep track of them. This is useful when
testing an API with multiple different bodies.

For now I added the description field only to the body form editor where I am thinking most people would like having it. However it could also be interesting for the url encoding editor and header editor, either by default or through an enable/disable option in the settings.


gschier:
I'm torn on this one. I love having it here as a feature but don't like how much it cramps the UI since most people will never use it.

What do you think about this?

- If any fields have a description, show it automatically
- Add ability to toggle description visibility in the gear icon menu
- Also do the same thing for urlencoded form, query, and header editors

image

gschier (changes requested):
I'm torn on this one. I love having it here as a feature but don't like how much it cramps the UI since most people will never use it.

What do you think about this?

- If any fields have a description, show it automatically
- Add ability to toggle description visibility in the gear icon menu
- Also do the same thing for urlencoded form, query, and header editors

image

pachisi456:
Thank you for your feedback!
Your suggestions make perfect sense to me. I will implement them as suggested. :+1:

pachisi456:
Done. Let me know what you think. :)

gschier (approved):
Awesome! Thanks for submitting this!

joshwilliard:
This is great! This was the feature I have been waiting for to replace finally replace PostMan entirely. It was really difficult to use Insomnia if you used encoded query parameters. I use a lot of base64 encoding in my query params so its nice to see the un-encoded versions in the description.

Issue #675, opened 2017-12-21
[Feature] Ability to provide custom CA Certificate


gschier:
`libcurl` provides a way to specify a CA bundle used for requests ([CURLOPT_CAINFO](https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html)). Currently, this is [being used](https://github.com/getinsomnia/insomnia/blob/31eb01f24d2dfbd07654a23950b9401bdbd7f050/packages/insomnia-app/app/network/network.js#L274-L288) to provide a base CA bundle for Windows and Linux, but a user-supplied bundle could be concatenated to this fairly easily.

## To-Do

- Add setting to workspace settings to import a custom CA bundle (PEM format)
- Add logic to concatenate user-supplied bundle to base bundle (if Windows or Linux). On Mac, simply provide user-supplied bundle

## Unknowns

Since Curl retrieve the base CA bundle from the OS on Mac, it's unclear whether or not setting `CURLOPT_CAINFO` will override this bundle entirely or add to it. This needs to be tested.

aaron-trout:
Providing a CA on a per environment or per request basis might provide better flexibility than per workspace. For example talking to Kubernetes API and providing the cert to validate with.

coderjoe:
This is something I'd be very interested in having. I'd love to give a PR a try.

gschier:
@aaron-trout I agree that it's more flexible but I don't think the use case for a global CA cert is much more common. If we see enough desire to add per-request CA certs later, we can come up with a mechanism to do so.

@coderjoe are you still wanting to give this a try? It might end up being quite a bit of work but I'm happy to point you in the right direction. I'm picturing a new tab in the workspace setting (beside Client Certificates) to upload additional CA certificates.
![image](https://user-images.githubusercontent.com/587576/46822618-f5894e80-cd59-11e8-8aba-a20dd53996f3.png)

coderjoe:
@gschier absolutely still interested. I'll likely need help with testing on Mac OSX as I only have Linux and Windows machines.

gschier:
Sure, I could definitely help test that out :smile:

coderjoe:
Sorry for the delay. I very much appreciate the guidance/suggestions. If I run into any snags or questions I'll be sure to ask here.

gschier:
Nice! 👍 😄

macbookandrew:
I’d love to see this feature too and can help test it out.

[...]

PR #1804, opened 2019-11-24
[Feature] Certificate Authority Settings

coolhome:
Related Issues: #675 and #1432
Status: In Progress

### [Completed] Allow Windows users to pull certificate authorities from the Windows Cert Store
With the help of the library [`win-ca`](https://github.com/ukoloff/win-ca) we are able to pull windows trusted certificates under the users profile. This package can be installed for all operating systems but will only do something on windows.

### Allow Windows/Linux/(Mac?) to provide custom ca bundle
Completed for windows/linux.

Upon research on [CURLOPT_CAINFO](https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html) we should be able to allow mac users to also provide a CA file. 

> If the option is not set, then curl will use the certificates in the system and user Keychain to verify the peer.

### Unit Tests
The whole thing needs unit tests still.

### Polish Code
Maybe move the logic outside of network.js into its own file. Maybe a function to return the file that should be used with curl?

## Ideas
* Combine CA bundles?
* Workspace CA files? Would only work with a user provided cert?
* When on Windows use Cert Store, when on Mac use Keychain, else use provided ca info. To overwrite you would add to workspace?

@gschier would love your feedback :)

### DIFF
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -70,6 +70,8 @@ export function init(): BaseSettings {
     proxyEnabled: false,
     timeout: 0,
     validateSSL: true,
+    caBundle: '',
+    caBundlePath: '',
     forceVerticalLayout: false,
     autoHideMenuBar: false,
     theme: packageJson.app.theme,
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -20,6 +20,8 @@ type BaseSettings = {
   proxyEnabled: boolean,
   timeout: number,
   validateSSL: boolean,
+  caBundle: string,
+  caBundlePath: String,
   forceVerticalLayout: boolean,
   autoHideMenuBar: boolean,
   theme: string,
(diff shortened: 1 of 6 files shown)


gschier (commented):
I like what you have here so far! Do you think it's worth splitting this PR into two separate ones? One for the setting to allow Windows (and no one else) users to use the OS bundle and one to let users specify their own bundle?

And yes, `CURLOPT_CAINFO` should work for specifying your own bundle since we're [already using it](https://github.com/getinsomnia/insomnia/blob/36059c848080a60b9d90683c0932a743866677f7/packages/insomnia-app/app/network/network.js#L362) to specify the Linux/Windows one.

coolhome:
@gschier I was thinking that as well. I will resubmit a PR with the ability to use a custom bundle. Later in another PR I will look into implementing the ability to take the OS specific versions.

gschier:
Awesome, that sounds good.

develohpanda:
Hi @coolhome! Thanks a bunch for doing this, awesome work! 💯

I've picked it up to get it across the line after some internal discussions. I have created a PR (#2020) with some refactoring and resolved merge conflicts, so I will close this PR from your fork for now. _Please reopen if you have changes in addition to those here_, and we can take it from there. 😄

PR #1829, opened 2019-12-06
[Feature] Prettify Json When Saving Response To File

### LINKED ISSUE #764: [Feature Request] Pretty print JSONs while saving response to files
[Feature Request] – Pretty print JSONs while saving response to files

- Insomnia Version: 5.14.7
- Operating System: Windows 7

## Details

Currently, if you save a JSON response to a file, the entire
JSON string gets saved as a single line in the file. Do you plan to
support pretty printing while saving to files ? Similar to the Visual Preview
feature but for files.

elemanhillary-zz:
Closes [#764](https://github.com/getinsomnia/insomnia/issues/764)

### DIFF
--- a/packages/insomnia-app/app/ui/components/response-pane.js
+++ b/packages/insomnia-app/app/ui/components/response-pane.js
@@ -28,6 +28,8 @@ import ErrorBoundary from './error-boundary';
 import type { HotKeyRegistry } from '../../common/hotkeys';
 import { hotKeyRefs } from '../../common/hotkeys';
 import type { RequestVersion } from '../../models/request-version';
+import { showError } from '../components/modals/index';
+import prettify from 'insomnia-prettify';
 
 type Props = {
   // Functions
@@ -96,11 +98,24 @@ class ResponsePane extends React.PureComponent<Props> {
       }
 
       const readStream = models.response.getBodyStream(response);
+      let dataBuffers = [];
       if (readStream) {
-        const to = fs.createWriteStream(outputPath);
-        readStream.pipe(to);
-        to.on('error', err => {
-          console.warn('Failed to save response body', err);
+        readStream.on('data', data => {
+          dataBuffers.push(data);
+        });
+        readStream.on('end', () => {
+          const to = fs.createWriteStream(outputPath);
+          if (extension === 'json') {
+            dataBuffers = prettify.json(dataBuffers);
+          }
+          to.write(dataBuffers);
+          to.on('error', err => {
+            showError({
+              title: 'Save Failed',
+              message: 'Failed to save response body',
+              error: err,
+            });
+          });
         });
       }
     });


packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -97,10 +97,14 @@ class ResponsePane extends React.PureComponent<Props> {
 
       const readStream = models.response.getBodyStream(response);
       if (readStream) {
-        const to = fs.createWriteStream(outputPath);
-        readStream.pipe(to);
-        to.on('error', err => {
-          console.warn('Failed to save response body', err);
+        readStream.on('data', data => {
```
gschier: You'll need to handle the case where the `data` event is called multiple times.

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -97,10 +97,14 @@ class ResponsePane extends React.PureComponent<Props> {
...
+          const data_ = data.toString();
```
gschier: We should avoid converting to a string unless we know it's JSON (can inspect the `Content-Type` header)

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -97,10 +97,14 @@ class ResponsePane extends React.PureComponent<Props> {
...
+          const readStreamContents = JSON.stringify(JSON.parse(data_), null, 2);
```
gschier: We should handle the error here when `parse` fails on invalid JSON.

We should also use Insomnia's custom JSON formatter, which is more flexible than the built-in and preserves more of the original data. You can find examples in the repo if you search for `'insomnia-prettify'`

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -97,10 +97,14 @@ class ResponsePane extends React.PureComponent<Props> {
...
+          const to = fs.createWriteStream(outputPath);
+          to.write(readStreamContents);
+          to.on('error', err => {
+            console.warn('Failed to save response body', err);
```
gschier: Should use the `showError()` helper to show the error to the user.

gschier (changes requested):
Hey @elemanhillary, thanks for taking a crack at this! I've left quite a few comments here so let me know if you have any questions.

I also think we should present an additional dropdown button to save "prettified" response because saving the original un-prettified response is still something that should be possible.

packages/insomnia-app/app/ui/components/response-pane.js
elemanhillary: on it

packages/insomnia-app/app/ui/components/response-pane.js
elemanhillary: on it

packages/insomnia-app/app/ui/components/response-pane.js
elemanhillary: on it

packages/insomnia-app/app/ui/components/response-pane.js
elemanhillary: on it

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -96,11 +98,21 @@ class ResponsePane extends React.PureComponent<Props> {
       }
 
       const readStream = models.response.getBodyStream(response);
+      let responseBody = '';
```
gschier: Since this is a string, your logic will force `data` to be converted from a Buffer a string. For binary data, we need to ensure that it remains a Buffer if it's not a string content-type (eg. PDF document, Image, or mp3).

There's some other code that does this over here that can be referenced: https://github.com/Kong/insomnia/blob/c03c52ef90c641143994d03bc2b5a97d67d86c6f/packages/insomnia-app/app/ui/components/viewers/response-multipart.js#L177-L180

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -96,11 +98,21 @@ class ResponsePane extends React.PureComponent<Props> {
...
-        const to = fs.createWriteStream(outputPath);
-        readStream.pipe(to);
-        to.on('error', err => {
-          console.warn('Failed to save response body', err);
+        readStream.on('data', data => {
+          responseBody += data;
+        });
+        readStream.on('end', () => {
+          const to = fs.createWriteStream(outputPath);
+          to.write(prettify.json(responseBody));
```
gschier: Nice! I think the only thing left to do here is add a check to see if the content type is actually JSON.

There is a `contentType` variable defined above that can be used to see if it contains the substring `json`. If it is not JSON, we should leave it unprocessed.

gschier (changes requested):
Just a couple more comments 🙂

packages/insomnia-app/app/ui/components/response-pane.js
elemanhillary: i don't know how i forgot that but let me work on it, it's  that i get to work on insomnia when am over worked at my work place

gschier: No rush @elemanhillary, we're just grateful you're contributing 😄 

Also, we're happy to help finish this one if you don't have the time 🤗

gschier (approved):
Nice, thanks for making those changes so quickly ❤️

robross0606:
This should be an option and not done by default.  Saving the response with whitespace can increase the file size considerably.

PR #1845, opened 2019-12-12
[Feature] Toggle URL Click Throughs

### LINKED ISSUE #1817: [Feature Request] Add option to disable URL clickthroughs in Preview results
It would be nice to have a feature to disable URL clickthroughs in Preview results section.

Reason: We use Insomnia to test our REST API, which provides cyber threat intel, and it frequently contains payloads with URIs pointing to malicious (C2) servers. If there was an option to disable URL clickthoughs, it would prevent the user from accidently clicking on a malicious link.

image

elemanhillary-zz:
Closes [#1817](https://github.com/Kong/insomnia/issues/1817)


packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -57,8 +57,15 @@ type Props = {
   response: ?Response,
 };
 
+type State = {
+  disable: boolean,
```
gschier: What do you think about making this setting global to the application instead of local to the response? Having it stored in the React state will not persist across application loads. 

I suggest adding a `disableResponseLinkHandling` property to the `Settings` model and storing it there.

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -97,10 +104,14 @@ class ResponsePane extends React.PureComponent<Props> {
...
-        to.on('error', err => {
-          console.warn('Failed to save response body', err);
+        readStream.on('data', data => {
+          const data_ = data.toString();
+          const readStreamContents = JSON.stringify(JSON.parse(data_), null, 2);
+          const to = fs.createWriteStream(outputPath);
+          to.write(readStreamContents);
+          to.on('error', err => {
+            console.warn('Failed to save response body', err);
+          });
```
gschier: Should this be removed? It looks like there changes are from your other PR.

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -155,6 +166,12 @@ class ResponsePane extends React.PureComponent<Props> {
     }
   }
 
+  _handleURLClickthrough() {
+    this.setState(prevState => ({
```
gschier: Nice React work here!

packages/insomnia-app/app/ui/components/viewers/response-viewer.js
```diff
@@ -206,8 +205,9 @@ class ResponseViewer extends React.Component<Props, State> {
       responseId,
       updateFilter,
       url,
+      urlClickable,
     } = this.props;
-
+    console.log('URLCLCIC', this.props.urlClickable);
```
gschier: Debug console logs should be removed.

packages/insomnia-app/app/ui/components/viewers/response-viewer.js
```diff
@@ -392,6 +392,7 @@ class ResponseViewer extends React.Component<Props, State> {
           uniquenessKey={responseId}
           ref={this._setSelectableViewRef}
           onClickLink={this._handleOpenLink}
+          urlClickable={urlClickable}
```
gschier: As mentioned above, the `urlClickable` property is not necessary. Links can be done by doing the following instead:

```
onClickLink={urlClickable ? this._handleOpenLink : null}
```

gschier (changes requested):
Left an initial rounds of comments. Nothing too major except the note about storing the setting in the Settings model instead of React state.

Thanks for taking care of this one! 🤗

packages/insomnia-app/app/ui/components/dropdowns/preview-mode-dropdown.js
```diff
@@ -70,4 +70,13 @@ class PreviewModeDropdown extends React.PureComponent<Props> {
   }
 }
 
+PreviewModeDropdown.propTypes = {
```
gschier: This isn't needed here since this component uses Flow to define it's props at the top of this file.

Sorry for the confusion. Flow was introduced to Insomnia quite far into it's development so many things still use `propTypes` but all the new stuff has migrated to Flow.

packages/insomnia-app/app/models/settings.js
```diff
@@ -57,6 +58,7 @@ export function init(): BaseSettings {
   return {
     showPasswords: false,
     useBulkHeaderEditor: false,
+    enableUrlClickThrough: false,
```
gschier: Can we change this to be `disableResponsePreviewLink` and default it to `false`? It's usually good to ensure that the default and `undefined` map to the same value.

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -171,6 +171,7 @@ class General extends React.PureComponent<Props, State> {
               'forceVerticalLayout',
               '',
             )}
+            {this.renderBooleanSetting('Enable URL click through', 'enableUrlClickThrough', '')}
```
gschier: Same here. Can we update the description to read "Disable Response Preview Links"?

packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -851,6 +851,7 @@ class Wrapper extends React.PureComponent<Props, State> {
             requestVersions={requestVersions}
             responses={activeRequestResponses}
             response={activeResponse}
+            settings={settings}
```
gschier: Since the other props here pull values out of settings, can we do the same here and instead pass a `disableResponsePreviewLinks` property to the component?

It makes sense what you did but it's better to be consistent 🙂

gschier (changes requested):
Just a few minor comments left. Getting close! 🚀🌟

packages/insomnia-app/app/models/settings.js
elemanhillary: which default are we talking about here @gschier

packages/insomnia-app/app/ui/components/dropdowns/preview-mode-dropdown.js
elemanhillary: i dont see flow's defined props in that file.

gschier (approved):
Thanks @elemanhillary!

tmursch:
Is this available in the current production version of insomnia (7.0.6)?

PR #1855, opened 2019-12-15
Enable and disable plugins!

### LINKED ISSUE #1610: [Feature Request] Enable / Disable Plugins
- Insomnia Version: v6.5.4
- Operating System: OSX

## Details

While having plugins is great, I would like the ability to enable / disable them either at the global level, at the workspace level, or even at the request / folder level. I have times when I don't want a plugin to run, but there's no way to disable the plugin, which will block my request from going through and error from the plugin either not running a supporting side car application or the request I'm making does not pertain to the plugin so it fails the request based on the plugin's requirements to make the request. The current way to "disable" a plugin is to delete it from the plugins directory in the application, and then to "enable" it would just be to install it again. I feel that this experience is high friction, and adding simple enable / disable logic could make this experience smoother for consumers of the plugin.

#### Highest Priority
The ability to enable / disable plugins at the global level, from the plugin management view.

#### Low Priority (Stretch)
* The ability to enable / disable plugins from the workspace level
* The ability to enable / disable plugins from the request level

develohpanda:
Closes #1610.

Added the ability to configure a plugin. The only configuration at the moment is to disable, but more can be added in the future. Plugins are installed as enabled by default.

I figured there would be fewer changes if I update the config settings and reload plugins using existing logic, instead of writing logic to unsubscribe a plugin.

I decided to use a toggle (using [react-switch](https://www.npmjs.com/package/react-switch) instead of checkbox for the UI because a toggle feels more intuitive here. Let me know if you have any suggestions for this!

Changing the state of a plugin involves reloading the plugins (to bind with the app). This means the `onChange` callback is async, and the UI waits for the logic to execute before visually showing the change, causing a ~1 sec UI delay. To overcome this, I extracted a stateful `ToggleSwitch` component which receives the `checked` prop (populated via settings) but also maintains a local `checked` state. The state drives the UI, while an effect runs to ensure the state and prop is kept in sync in case they fall out of sync.

Because changing the toggle triggers a plugin reload, and the current reload button disables while reloading, the toggles follow the same behavior. All toggles will disable when changed until the logic has executed. Champion mouse clickers will only be able to toggle one plugin at a time, no matter how fast they click. 😄 

image

Question: I can't quite figure out how Storybook generates prop types. Had a look at some of the other examples (like `Notice`) but am not sure.


develohpanda:
@gschier hmmm I've got an error with `React.FC` or `React.FunctionComponent` not being found in the react flow types. Where are these types defined? 😢 

image

nijikokun:
From a product perspective, this is looking great, amazing work so far 🎉

gschier:
> Hey @gschier, hmmm I've got an error with `React.FC` or `React.FunctionComponent` not being found in the react flow types. Where are these types defined? 😢
> 
> image

You may have to add a definition in the `flow-typed` folder

gschier:
On second thought. It might be worth converting that to a Class component. Insomnia doesn't currently have any Hooks related stuff so I propose that we favor consistency over niceness in this case. 

Perhaps we can have a talk later about whether or not introducing Hooks is a good idea and how we could go about it.

packages/insomnia-app/package.json
```diff
@@ -108,6 +108,7 @@
     "@hot-loader/react-dom",
     "react-redux",
     "react-sortable-hoc",
+    "react-switch",
```
gschier, line 111: Nice! Glad you figured this one out 😄

develohpanda:
> Insomnia doesn't currently have any Hooks-related stuff in the codebase and I think it makes sense to favor consistency for now.

Yep agreed, that makes sense. All ready to review now. 😄

I tried to type `react-switch` in `flow-typed` folder, similar to how `react-tabs` is typed, but when consuming in `toggle-switch.js` I kept facing tricky errors like the following. 

> Either  `React.Element` [2] is incompatible with  `React.Portal` [3]. Or property `@@iterator` is missing in  `React.Element` [2] but exists in  `$Iterable

packages/insomnia-app/app/plugins/index.js
```diff
@@ -200,21 +213,38 @@ export async function getResponseHooks(): Promise<Array<ResponseHook>> {
...
+  packageJSON: Object,
+  module: any,
+  allConfigs: PluginConfigMap,
+  path: ?string,
+): Plugin {
   const meta = packageJSON.insomnia || {};
+  const name = packageJSON.name || meta.name;
+
+  // Find config
+  const config: PluginConfig = allConfigs.hasOwnProperty(name)
```
gschier, line 263: Nice, I was wondering where you'd chose to setup the default config and this is a great choice.

packages/insomnia-app/app/ui/components/settings/toggle-switch.js
```diff
@@ -0,0 +1,71 @@
...
+  componentDidMount() {
+    this._mounted = true;
+  }
+
+  componentWillUnmount() {
+    this._mounted = false;
+  }
+
+  setChecked(checked: boolean) {
+    if (this._mounted) {
```
gschier: Was it necessary to have this check here? I know it's in a couple places throughout the codebase but only as a last resort hack.

packages/insomnia-app/app/ui/components/settings/toggle-switch.stories.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
gschier, line 1: Thanks for adding this! I was confused at first because I forgot Storybook was part of this project now 😆

packages/insomnia-app/app/ui/css/layout/base.less
```diff
@@ -115,6 +115,11 @@ table.table--fancy {
     padding: @padding-xs;
   }
 
+  &.table--valign-middle td,
```
gschier: Just checking, should this be:

```css
&.table--valign-middle td,
&.table--valign-middle th {
  ...
```

Could also simplify it with nesting:

```css
&.table--valign-middle {
  td, th {
    ...
```

gschier (approved):
Great execution on this one @develohpanda 😀

gschier (changes requested):
Woops, accidentally approved this PR before without realizing I had some comments worth addressing.

packages/insomnia-app/app/ui/components/settings/toggle-switch.stories.js
develohpanda: Storybook is great, I've used it a lot over the past year and was happy to see it added here 💯

OT but we could use [Zeit](https://zeit.co) to host it, as a dev resource that is always available. Useful for component libraries and not crucial for Insomnia but it would be nice!

packages/insomnia-app/app/ui/css/layout/base.less
develohpanda: Oh you're right, nice find! 😄 Still getting familiar with CSS selectors

packages/insomnia-app/app/ui/components/settings/toggle-switch.js
develohpanda: I tested a potential edge case to try and produce a memory leak error in console to no avail (which is great!), so I have removed this functionality 😄

gschier (dismissed):
Nice work @develohpanda! So excited to get this merged

PR #1863, opened 2019-12-22
Nunjuck resolution in order of property definitions

### LINKED ISSUE #1809: [Bug] Extracting substring to token value doesn't reference tag properly
I'd need to chain certain requests. In the middle of the chain I'd need to parse a substring from a response JSON field and use that value in the next step. I couldn't find out way to do this. Is it possible somehow?

Example:

Response body

```json
{
  "url": "https://redirect.to?code=abc123"
}
```

And I'd like to extract the `code` query param value from the `$.url` field.

develohpanda:
Putting up a draft PR that closes #1809, in-case this is the path forward. I was experimenting to see how far I would get and here is the result. 😄 

The key changes are within the `buildRenderContext` function in `render.js`, and two tests in `/common/__tests__/render.test.js`. The existing tests succeed as well as the two tests added pertaining to key order.

I see some inconsistent behavior in jest when compared to a manual test, so the assertion for the test titled `handles variables using tag before tag is defined as expected (incorrect order)` is a little odd. 😢 I have added a comment within the test where this happens.

This demo imports the dummy [environment](https://insomnia.rest/run/?label=Issue%201809&uri=https%3A%2F%2Fgist.githubusercontent.com%2Fgschier%2F21647324d9e89e8c98d1987830a380c5%2Fraw%2F870d6f4d19f235eb95e903be209d3ccfe3507e7b%2Finsomnia.1809.yaml) linked in PR 1809


packages/insomnia-app/app/common/__tests__/render.test.js
```diff
@@ -399,6 +399,54 @@ describe('buildRenderContext()', () => {
...
+        '&': ['value', 'replaced', 'hashed', 'consume'],
+      },
+    };
+
+    const context = await renderUtils.buildRenderContext([], rootEnvironment);
+
+    expect(context).toEqual({
+      value: 'ThisIsATopSecretValue',
+      hashed: 'f67565de946a899a534fd908e7eef872',
+      replaced: 'cat', // is cat at runtime, but is f67565de946a899a534fd908e7eef872 during the test, rest are the same :shrug:
```
develohpanda: Strange behavior 😢🤔

packages/insomnia-app/app/common/render.js
```diff
@@ -36,23 +37,41 @@ export async function buildRenderContext(
...
+  // Get root environment keys in correct order
+  // Then get sub environment keys in correct order
+  // Then get ancestor (folder) environment keys in correct order
   if (rootEnvironment) {
-    envObjects.push(rootEnvironment.data);
+    const ordered = orderedJSON.order(
+      rootEnvironment.data,
+      rootEnvironment.dataPropertyOrder,
+      JSON_ORDER_PREFIX,
+    );
```
develohpanda, line 48: Likely not the most efficient place to re-order, but this way the change is isolated. Open to suggestions! 😄

packages/insomnia-app/app/common/__tests__/render.test.js
```diff
@@ -399,6 +399,54 @@ describe('buildRenderContext()', () => {
...
+    const rootEnvironment = {
+      type: models.environment.type,
+      data: {
+        consume: '{{ replaced }}',
+        hashed: "{% hash 'md5', 'hex', value %}",
+        replaced: "{{ hashed | replace('f67565de946a899a534fd908e7eef872', 'cat') }}",
+        value: 'ThisIsATopSecretValue',
+      },
+      dataPropertyOrder: {
+        '&': ['value', 'replaced', 'hashed', 'consume'],
```
gschier: Looks like you just have this in the wrong order maybe? Your object is `consume, hashed, replaces, value` but your order is `value, replaced, hashed, consume`

develohpanda: Hmm I did this intentionally because `buildRenderContext` will order the env correctly (https://github.com/Kong/insomnia/pull/1863/files#r360679103). The `data` object here is in alphabetical order, representing how it would be stored in SQLite.

Possibly related but out of scope? It might be nicer to handle ordering at the DB read/write level, so that the retrieved object is in the correct order for all consumers.

gschier (approved):
Looks good @develohpanda! `reliability++` 🥳

PR #1874, opened 2020-01-06
Add option to download the response data in csv/excel format

Sandhya-nambrath:
Currently the response is downloadable only in json format. It would be good if we were able to download the response into csv / Excel format. 

The view response documentation does not mention the above format. 
https://support.insomnia.rest/article/63-viewing-responses


gschier:
> The view response documentation does not mention the above format.
https://support.insomnia.rest/article/63-viewing-responses

Thank you for pointing this out. In fact, Insomnia _does_ preview CSV files, but not XLS files. If there is an easy way to add XLS support, I would rather add that as an officially-supported format.

image

PR #1905, opened 2020-01-22
Scope responses to active environment

### LINKED ISSUE #1895: [Feature Request] Option to scope Request Chaining per Environment

Using Request Chaining to store the result of a previous request in an environment variable for use in future requests is very useful.  However, when I switch environments, I usually do not want to use the results of requests made in the previous environment, since that environment probably uses a different database which returns, for example, different user ids.  I would like to propose an enhancement that gives the user the option to scope a request chain to the environment in which it's made.

For example:

Suppose you have two environments, `localhost` and `staging`, and an environment variable `user_id` that's set from the result of a Create User action and used in the url for subsequent requests.

I am working in the `localhost` environment:
`POST users  # => returns a user_id of 42`
`GET users/:user_id/posts => returns the posts for user 42`
...etc

Then I push up my work and want to check it in `staging`, and I change environments to look at the user's posts:
`GET users/:user_id/posts => uses 42, the value from localhost`
[user 42 does not exist in staging]
[I must manually change my `user_id` to the correct id]

Then, it's back to development in `localhost`, but I've lost my original `user_id`
`GET users/:user_id/posts => uses NOT 42`

Current behavior forces me to update the `user_id` myself each time I change environments.

Request-Chained results used in environment variables should be scoped to the environment where it's set.  Changing environments would silently change the value to the last one set in that environment, or null if no value has been set.  The user may optionally elect to have values persist across environments, but this would not be default behavior.

**Additional context**
none

gschier:
Closes #1895 

This PR adds a new `environmentId` to responses and stores the current environment on every response that is sent. Only responses sent on the currently-active environment will be shown.

What does this mean?

- Response template tag will now only reference responses from the current environment
- Won't get confused by seeing responses from other environments when browsing (Dev vs Prod)

### DIFF
--- a/packages/insomnia-app/app/common/__tests__/har.test.js
+++ b/packages/insomnia-app/app/common/__tests__/har.test.js
@@ -137,6 +137,30 @@ describe('exportHar()', () => {
     req3.headers.push({ name: 'X-Request', value: '3' });
     await models.request.create(req3);
 
+    const envBase = await models.environment.getOrCreateForWorkspace(workspace);
+    await models.environment.update(envBase, {
+      data: {
+        envvalue: '',
+      },
+    });
+    const envPublic = await models.environment.create({
+      _id: 'env_1',
+      name: 'Public',
+      parentId: envBase._id,
+      data: {
+        envvalue: 'public',
+      },
+    });
+    const envPrivate = await models.environment.create({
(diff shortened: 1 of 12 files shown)


desmondmonster:
🎉🎉🎉

PR #1916, opened 2020-01-30
More control over the prompt plugin

superruzafa:
This PR extends the idea introduced in #1597 which allowed the possibility of reuse the previous entered value if you marked `Default to Last Value`.

### Background

I've found myself having a separate file with text strings to C&P them into the prompt dialogs because the behavior of that dialog wasn't as flexible as I wanted, so I tried to improve the current prompt dialog plugin with my own needs, but I thought they are general enough so they could be useful for others.

### Feature

imagen

This PR leaves the user to choose between a couple of options about how to show the prompt:

- on each request with an empty text box, ignoring any previously entered value (original behavior)
- on each request but defaults to the prompt's default value, ignoring any previously entered value (new behavior)
- on each request but defaults to the current stored value (aka `Default to Last Value`)
- once and storing the value for futher requests (the behavior used when a `Storage Key` was set).

As a new feature now the prompt text is autoselected in order to quickly edit it if you want to change the value.

### Remarks

In order to keep the backwards compatibility I transformed the current `Default to Last Value` booleans to the equivalents enums. 

A broken behavior is that the prompt now is shown regardless a custom storage key has been set if the user choosed any of the `on each request` options, but this could be useful because until now there was no way to change the stored message unless one reload the whole application.

### DIFF
--- a/plugins/insomnia-plugin-prompt/index.js
+++ b/plugins/insomnia-plugin-prompt/index.js
@@ -40,23 +40,35 @@ module.exports.templateTags = [
         defaultValue: false,
       },
       {
-        displayName: 'Default to Last Value',
-        type: 'boolean',
+        displayName: 'Prompt for a Value...',
+        type: 'enum',
         help:
-          'If this is enabled, the input field will be pre-filled with this value. This option is ' +
-          'ignored when the storage key is set.',
-        defaultValue: true,
+          'Controls when this prompt should be shown and what value should be used to '+
+          'pre-populate the prompt dialog',
+        options: [
+          { displayName: 'on each request', value: 'always-fresh' },
+          { displayName: 'on each request, using the default value as default',  value: 'always-default' },
+          { displayName: 'on each request, using the stored value as default',  value: 'always-storage' },
+          { displayName: 'at most once, storing the value for further requests',  value: 'once-storage' },
+        ]
       },
     ],
-    async run(context, title, label, defaultValue, explicitStorageKey, maskText, saveLastValue) {
+    async run(context, title, label, defaultValue, explicitStorageKey, maskText, showCondition) {
       if (!title) {
         throw new Error('Title attribute is required for prompt tag');
       }
 
-      // If we don't have a key, default to request ID.
+      // Backward compatibility with "Default to Last Value" #1597
+      if (showCondition === true) {
+          showCondition = 'always-storage'
+      } else if (showCondition === false) {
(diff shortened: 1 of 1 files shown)


nijikokun:
I'm wondering if this should be split into it's own select, and retain the previous functionality as-is. Thoughts?

superruzafa:
I choosed this approach because I think that the previous feature was a very specific case, while mine is more general and encompasses it (also it's backwards compatible)
Also I tried to prevent increasing the number of parameters of the prompt plugin.

This feature also allows the user to change the stored value **without** reload the whole app.

The label of the enum could be improved, though.

gschier (approved):
I'm okay with this change. Thanks for submitting it @superruzafa! I might play around with the label names after merging as it seems a bit confusing.

gschier:
@superruzafa After playing around with this, I can't actually find any new behavior that can't already be done with the old setup.

- _on each request with an empty text box, ignoring any previously entered value (original behavior)_
  - **Default to Last Value** unchecked
  - **Default** empty
  - **Storage Key** unchecked
- _on each request but defaults to the prompt's default value, ignoring any previously entered value (new behavior)_
  - **Default to Last Value** unchecked
  - **Default** set
  - **Storage Key** empty
- _on each request but defaults to the current stored value (aka Default to Last Value)_
  - **Default to Last Value** checked
  - **Default** empty or set
  - **Storage Key** empty
- _once and storing the value for futher requests (the behavior used when a Storage Key was set)._
  - **Default to Last Value** checked or unchecked
  - **Default** empty or set
  - **Storage Key** set

Going to keep the old behavior for now unless you can point out something I'm missing. I agree that having a dropdown does make it a bit nicer, but since we can't remove other params, it sends up complicating things too much.

superruzafa:
@gschier, it seems that the behavior I though it couldn't be reproduced can be done in fact using a combination of settings. Having said that it's ok for me if you ignore this PR :)

But maybe we can learn something good after all. What about let the user to set the exactly behavior of the prompt by choosing an option from a dropdown instead of think what chekboxes/textboxes should be checked/unchecked/filled/emptied to have the same behavior?
That, I guess, is what this PR would be useful for.

Anyway, thanks a lot for your fantastic application!

gschier:
I agree it would be useful to do that. However, I don't see an easy way to add this functionality without (1) duplicating options or (2) breaking backwards-compatibility. Perhaps a new prompt plugin v2 later on that implements all the things we've added since the beginning 😄

PR #1923, opened 2020-02-03
Convert Less vars to CSS vars

gschier:
In order to support the `styled-components` Storybook stuff, we need to use CSS vars instead of Less vars. Since the app already used a mixture of the two, it made sense to just convert everything to CSS vars. This both simplifies things and allows us to use the same config within Storybook components.


sonicyeti (approved):
Awesome man, zeroing in on a consistent source of truth for styling.

PR #1926, opened 2020-02-05
View open source licenses in app

gschier:
This PR adds a build step to generate a list of all included open source licenses, which can then be viewed within the app.

image

### DIFF
--- a/packages/insomnia-app/scripts/build.js
+++ b/packages/insomnia-app/scripts/build.js
@@ -124,7 +124,7 @@ async function buildLicenseList(relSource, relDest) {
           repository ? `REPOSITORY: ${repository}` : null,
           publisher ? `PUBLISHER: ${publisher}` : null,
           email ? `EMAIL: ${email}` : null,
-          txt ? '\n' + txt : null,
+          '\n' + txt,
         ]
           .filter(v => v !== null)
           .join('\n');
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/scripts/build.js
```diff
@@ -97,6 +100,52 @@ async function copyFiles(relSource, relDest) {
...
+        const txt = licenseFile ? fs.readFileSync(licenseFile) : '[no license file]';
+        const body = [
+          '-------------------------------------------------------------------------',
+          '',
+          `PACKAGE: ${pkgName}`,
+          licenses ? `LICENSES: ${licenses}` : null,
+          repository ? `REPOSITORY: ${repository}` : null,
+          publisher ? `PUBLISHER: ${publisher}` : null,
+          email ? `EMAIL: ${email}` : null,
+          txt ? '\n' + txt : null,
```
nijikokun: Interesting check given the result of line `118`

- No license file: `[no license file]`
- Empty license file: `null`

Should that second check be `[empty license file]` instead of `null`?

nijikokun (changes requested):
Woah, awesome 😍 

The code is solid, nice work, just have one question about the fallback value on empty license.

packages/insomnia-app/scripts/build.js
gschier: Good point!

PR #1936, opened 2020-02-12
Add support for live display name for template tags

### LINKED ISSUE #1935: [Feature Request] Live display name for template tags
When using template tags only the name of the tag is shown, without any other useful information:

imagen

It would be nice to show more relevant information.

For example, the Prompt plugin could be rendered by using the value of its Title instead of just the name of the tag:

imagen

Since this feature could be confusing then the tags could preserve part of the old look:

imagen

Also, an option to (de)activate this feature should be provided:

imagen

superruzafa:
This PR allows the template tag plugins to have an optional entry named `liveDisplayName` [supposed to be a function](https://github.com/Kong/insomnia/pull/1936/files#diff-23c147cb703706cbc7adbd0060ab1ed4) bound to the tag itself.
The template tag's current values are passed to the function each time the template tag needs to be rendered. The returned value is finally used as display name.

If the plugin doesn't define the `liveDisplayName` function then the fallback `displayName` is used instead.

The only tag included here that takes advantage of this feature is the Prompt template tag like this:

imagen

but other formats can be discussed.

Closes #1935


gschier:
I like this improvement! One question, though; It doesn't seem like an app setting for this should be necessary. If the plugin implements the new API, I feel like it's okay to use it. Did you have a specific reason for adding the setting?

superruzafa:
I added it because I don't know how users use the Title field.
Something like Title = "Please, enter here the username to be created" would broke completely the layout and because of that I set the feature disabled by default, leaving the user to enable it manually.

superruzafa:
@gschier, it has been a while since I created this PR. Are you still thinking this is a good feature to be added?

superruzafa:
@gschier, I released the code I was using in my local machine and created a plugin fully compatible with the current API:

https://github.com/superruzafa/insomnia-plugin-param

I'd like to use the the live display name feature requested in this PR.

Would you agree if I remove all the stuff from the Prompt plugin and the app settings from this PR but keep the code to allow plugins to use the `liveDisplayName` feature?

BTW, congratulations for launching the Plugin Hub!

superruzafa:
@gschier, any word about this?

gschier (approved):
I just played around this this locally and it's super useful/powerful! Thanks for getting this in 👍 🥳

PR #1978, opened 2020-03-03
feature: Generate Variable Name from SecuritySchemas name parameter.

pocket7878:
# Background
Some of my api requires multiple securitySchema simultaneously:

```openapi.yml
securitySchemes:
    AuthToken:
      type: apiKey
      description: JWT Token for authorizations.
      name: X-AuthToken
      in: header
    AppType:
      type: apiKey
      description: Application platform
      name: X-AppType
      in: header
    AppVersion:
      type: apiKey
      description: Application semantic version.
      name: X-AppVersion
      in: header
```

```yaml
/sample/endpoint.json:
    patch:
      security:
        - AuthToken: []
        - AppVersion: []
        - AppType: []
```

# Problem
Currently Insomnia merge this security schemas into same variable `apiKey`.
But I need to setup these variables separately.

# Feature
In this PR I'v changed variable name generation process.
And use security schemas `name` parameter to generate variable.

```yaml
components:
  securitySchemes:
    Key-Header-With-Variable:
      type: apiKey
      name: x-api_key
      in: header
paths:
  /key/header:
    get:
      security:
        - Key-Header-With-Variable: []
      responses:
        '200':
          description: OK
```

will generate

```json
{
      "_id": "env___BASE_ENVIRONMENT_ID___sub",
      "_type": "environment",
      "data": {
        "xApiKey": "xApiKey", // Automatically generated instead of apiKey variable.
        "base_path": "/v1",
        "host": "api.server.test",
        "httpPassword": "password",
        "httpUsername": "username",
        "scheme": "https"
      },
      "name": "OpenAPI env",
      "parentId": "__BASE_ENVIRONMENT_ID__"
 },
{
      "_id": "req___WORKSPACE_ID__48bba8a5",
      "_type": "request",
      "authentication": {},
      "body": {},
      "headers": [
        {
          "disabled": false,
          "name": "x-api_key",
          "value": "{{ xApiKey }}"
        }
      ],
      "method": "GET",
      "name": "/key/header",
      "parameters": [],
      "parentId": "__WORKSPACE_ID__",
      "url": "{{ base_url }}/key/header"
 },
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,5 +58,8 @@
     "prettier": "^1.16.4",
     "pretty-quick": "^1.10.0",
     "rimraf": "^2.6.3"
+  },
+  "dependencies": {
+    "change-case": "^4.1.1"
   }
 }
--- a/packages/insomnia-importers/src/importers/openapi3.js
+++ b/packages/insomnia-importers/src/importers/openapi3.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const crypto = require('crypto');
+const changeCase = require('change-case');
 
 const SwaggerParser = require('swagger-parser');
 const { parse: urlParse } = require('url');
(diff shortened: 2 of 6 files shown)


gschier (commented):
I would rather solve the naming conflict more generically and somehow base it off of information already available in the security schema. 

Since the security schema names are unique, could we perhaps generate the environment variable name based off that instead?

Another option could be to suffix the variable name with an integer number if we detect a conflict.

pocket7878:
@gschier 
Thank you for your review! and sorry for my late reply.
I just updated my codes to generate variable name from securitySchemas name property.

package.json
```diff
@@ -58,5 +58,8 @@
     "prettier": "^1.16.4",
     "pretty-quick": "^1.10.0",
     "rimraf": "^2.6.3"
+  },
+  "dependencies": {
+    "change-case": "^4.1.1"
```
gschier, line 63: Woah, I wouldn't have figured this to be a module! Nice one 👍

packages/insomnia-importers/src/importers/openapi3.js
```diff
@@ -290,23 +291,28 @@ function parseSecurity(security, securitySchemes) {
   const apiKeyHeaders = apiKeySchemes
     .filter(scheme => scheme.in === 'header')
     .map(scheme => {
+      let variableName = changeCase.camelCase(scheme.name);
```
gschier, line 294: Love this. Super clean

gschier (approved):
Looks good! thanks for taking care of this one 💜

PR #1980, opened 2020-03-03
Allow checkbox input type in prompt dialog

superruzafa:
This PR allows the Prompt modal dialog to show `<input type="checkbox">` elements.

You only have to call it by setting `inputType` to `checkbox`. You can also set `defaultValue` to `false`/`true` to uncheck/check the checkbox, respectively.

The modal will return the string value `"true"` if the the checkbox was checked and `"false"` otherwise.

imagen.

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/prompt-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/prompt-modal.js
@@ -235,21 +235,22 @@ class PromptModal extends React.PureComponent<Props, State> {
     let field;
     if (inputType === 'checkbox') {
       field = (
-        <div>
-          {input}
-          <label htmlFor="prompt-input" className="label--checkbox">
-            {label}
+        <div className="form-control form-control--wide">
+          <label htmlFor="prompt-input" className="inline-block">
+            {label} {input}
           </label>
         </div>
       );
     } else if (label) {
       field = (
-        <label>
-          {label} {input}
-        </label>
+        <div className="form-control form-control--outlined form-control--wide">
+          <label htmlFor="prompt-input">
+            {label} {input}
+          </label>
+        </div>
       );
     } else {
-      field = input;
+      field = <div className="form-control form-control--outlined form-control--wide">{input}</div>;
     }
 
     return (
@@ -257,7 +258,7 @@ class PromptModal extends React.PureComponent<Props, State> {
         <ModalHeader>{title}</ModalHeader>
         <ModalBody className="wide">
           <form onSubmit={this._handleSubmit} className="wide pad">
-            <div className="form-control form-control--outlined form-control--wide">{field}</div>
+            {field}
             {sanitizedHints}
           </form>
         </ModalBody>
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/ui/css/components/forms.less
```diff
@@ -78,6 +78,10 @@
     }
   }
 
+  .label--checkbox {
+    line-height: var(--line-height-xs);
+  }
+
```
superruzafa: I'm not a CSS expert.
This could be improved as I observed some baseline misalignment.

packages/insomnia-app/app/ui/components/modals/prompt-modal.js
```diff
@@ -226,21 +232,32 @@ class PromptModal extends React.PureComponent<Props, State> {
...
+    let field;
+    if (inputType === 'checkbox') {
+      field = (
+        <div>
+          {input}
+          <label htmlFor="prompt-input" className="label--checkbox">
+            {label}
+          </label>
+        </div>
+      );
```
develohpanda: There are existing checkbox inputs with labels, inside the preferences modal (for example), and it would be awesome to implement this checkbox in a similar fashion to have consistent styling.

image

Code snippet:
https://github.com/Kong/insomnia/blob/327eb5b692afd14e9d499691a4aa8da7f4ebc3d7/packages/insomnia-app/app/ui/components/settings/general.js#L118-L129

develohpanda (changes requested):
Hi @superruzafa! Thanks for putting up this PR; LGTM! Just one suggestion regarding spacing, and then good to go!

In the future it would be nice to update the `onComplete` callback to accept the true type (`boolean`, `integer`, `string`), but this is good for now. 😃

superruzafa (commented):
That's exactly what I was looking for :)

packages/insomnia-app/app/ui/components/modals/prompt-modal.js
```diff
@@ -226,21 +232,33 @@ class PromptModal extends React.PureComponent<Props, State> {
...
+    } else if (label) {
+      field = (
+        <div className="form-control form-control--outlined form-control--wide">
+          <label htmlFor="prompt-input">
+            {label} {input}
+          </label>
+        </div>
+      );
```
develohpanda: Looks good now! 🙌 

I see there's some duplication in the logic here and it would be nice to compress that down a bit.

Looks like the main difference is with the `className` attribute. If you're familiar with the `classnames` package, one way to consolidate would be to provide an object parameter where the key is your `className` and the value is a boolean.

Here's an example, would you like to give it a go? ⭐️ 

https://github.com/Kong/insomnia/blob/327eb5b692afd14e9d499691a4aa8da7f4ebc3d7/packages/insomnia-app/app/ui/components/base/dropdown/dropdown-divider.js#L9-L14

superruzafa: Done!

develohpanda (approved):
Nice one, thanks! 🎉

PR #1992, opened 2020-03-11
Set solarized background and foreground colors to spec

### LINKED ISSUE #1990: Solarized Dark theme does not use the correct colors
The Solarized Dark theme does not use the correct background color specified by the [official Solaized spec](https://ethanschoonover.com/solarized/). It is significantly lighter.

reynolek:
Fixes the font and background colors of the solarized themes to match the spec. Closes https://github.com/Kong/insomnia/issues/1990


gschier (approved):
Thanks for taking this one @reynolek! The Solarized fans will be happy once again 🥳

PR #1995, opened 2020-03-12
Mask template tag title popup when preview is disabled

superruzafa:
I found a bug that causes a template tag to display its rendered value even when the template tag disables the preview. For example, the Prompt template tag disables the preview when the user checks the Mask Text checkbox.

This works well in the template dialog's Live Preview, but you can still seeing the value if you move the mouse over the tag:

image

This PR fixes this by masking rendered values when the template tag disables the preview, as it's done with the Live Preview.

image


gschier (approved):
Awesome, nice catch! 😃

PR #2001, opened 2020-03-17
Add support for default values in graphql fix #2000

### LINKED ISSUE #2000: [Feature Request] Add support for GraphQL defaultValues
I was trying to browse the documentation of some input of my mutation,
and it seems that Insomnia does not display default values 
( looks like this `isGlobal: Boolean = false`)

It would be great if it could be displayed at least in the code pane

currently 

image

eMerzh:
Closes #2000

Here is an attempt to fix missing default values for graphql.


gschier:
Thanks for taking a crack at this one @eMerzh! Since there is UI involved in this change, can you post a screenshot of the final result?

eMerzh:
of course....
it looks like this
image

or 
image

gschier (approved):
Thanks for this! I will fix the Flow errors after merging

eMerzh:
thanks ... sorry didn't saw the flow issue... i can try to fix it in a few hours

PR #2002, opened 2020-03-18
Introduce flow `$Shape<T>` to db models

develohpanda:
[`$Shape<T>`](https://flow.org/en/docs/types/utilities/#toc-shape) in Flow allows autocomplete to work nicely and makes it safer to consume the data model functions 🎉

- I have reverted any changes to `models/response.js` because updating that was proving to be quite tricky and it would be better of to do that as a standalone change-set.
- `request-group-meta.js` had no typings so I introduced them as part of this PR.

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
@@ -40,18 +40,18 @@ class MoveRequestGroupModal extends React.PureComponent<Props, State> {
     e.preventDefault();
 
     const { requestGroup, selectedWorkspaceId } = this.state;
-    if (!requestGroup) {
+    if (!requestGroup || !selectedWorkspaceId) {
       return;
     }
 
-    const workspace = await models.workspace.getById(selectedWorkspaceId || 'n/a');
+    const workspace = await models.workspace.getById(selectedWorkspaceId);
     if (!workspace) {
       return;
     }
 
     const newRequestGroup = await models.requestGroup.duplicate(requestGroup);
     await models.requestGroup.update(newRequestGroup, {
-      sortKey: -1e9,
+      metaSortKey: -1e9,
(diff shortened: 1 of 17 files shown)


packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
```diff
@@ -40,18 +40,18 @@ class MoveRequestGroupModal extends React.PureComponent<Props, State> {
     e.preventDefault();
 
     const { requestGroup, selectedWorkspaceId } = this.state;
-    if (!requestGroup) {
+    if (!requestGroup || !selectedWorkspaceId) {
       return;
     }
 
-    const workspace = await models.workspace.getById(selectedWorkspaceId || 'n/a');
```
develohpanda, line 47: Is `n/a` a valid workspace id?

gschier: No, just a value I've been using to designate a non-existing db entry. I don't really have a super good reason why I started doing it but I kept it up for sake of consistency.

packages/insomnia-app/app/ui/components/modals/request-settings-modal.js
```diff
@@ -141,18 +141,18 @@ class RequestSettingsModal extends React.PureComponent<Props, State> {
...
-    const workspace = await models.workspace.getById(activeWorkspaceIdToCopyTo || 'n/a');
+    const workspace = await models.workspace.getById(activeWorkspaceIdToCopyTo);
     if (!workspace) {
       return;
     }
 
     const newRequest = await models.request.duplicate(request);
     await models.request.update(newRequest, {
-      sortKey: -1e9, // Move to top of sort order
+      metaSortKey: -1e9, // Move to top of sort order
```
gschier, line 155: Woah, already found a bug, nice!

gschier (approved):
Love this change 💜

PR #2003, opened 2020-03-18
Expose the request body for plugins

### LINKED ISSUE #423: [Feature Request] expose request/response body to modify or replace
Would insomnia plugin support to modify or replace request/response body?
It seems that only reponse buffer is exposed which only porivided to be modified now.

yittg:
Support the following api for plugins:

- get/set the request body;

And mark the old `(get/set)BodyText` api as deprecated.

Closes #423


yittg:
cc @gschier

parsadotsh:
Would it be possible to also expose `renderedRequest.body.fileName` in this PR, maybe with `getBodyFileName` and `setBodyFileName`. 
Together with `setMimeType`, plugins will then be able to send files.

yittg:
@lucashyper 😄 Actually, i missed it. This change should implement them.

yittg:
@gschier @develohpanda hi, PTAL or help to mention the right people, thanks.

packages/insomnia-app/app/plugins/context/request.js
```diff
@@ -147,6 +154,82 @@ export function init(
...
     },
+    getUploadFileName(): string {
+      return renderedRequest.body.fileName || '';
+    },
+    setUploadFileName(fileName: string): boolean {
+      if (!renderedRequest.body || renderedRequest.body.mimeType !== CONTENT_TYPE_FILE) {
+        return false;
+      }
+      renderedRequest.body.fileName = fileName;
+      return true;
```
develohpanda: This might be clearer to understand without the negative-or conditional 😄 

How about something like the following?

```suggestion
	  const isFile = renderedRequest.body && renderedRequest.body.mimeType === CONTENT_TYPE_FILE;
	  
	  if (isFile) {
	  	renderedRequest.body.fileName = fileName;
	  }
	  return isFile;
```

packages/insomnia-app/app/plugins/context/request.js
```diff
@@ -147,6 +154,82 @@ export function init(
...
+    },
+    getFileForm(): Array<any> {
+      return (renderedRequest.body.params || [])
+        .filter(p => p.type === 'file')
+        .map(({ name, fileName }) => ({ name, fileName }));
+    },
+    addFormItem(formItem: any): boolean {
+      if (
+        formItem.type === 'file' &&
+        (!renderedRequest.body || renderedRequest.body.mimeType !== CONTENT_TYPE_FORM_DATA)
```
develohpanda: Same as above, I'd suggest to avoid the negative-or (or in this case, compress it down to a positive condition, with a single `!` negation) 🎉

packages/insomnia-app/app/plugins/context/request.js
```diff
@@ -147,6 +154,82 @@ export function init(
...
+      ) {
+        return false;
+      }
+      const oldFormItem = misc.filterFormItem(renderedRequest.body.params || [], '', formItem.name);
+      if (oldFormItem.length === 0) {
+        if (!renderedRequest.body.params) {
+          renderedRequest.body.params = [formItem];
+        } else {
+          renderedRequest.body.params.push(formItem);
+        }
```
develohpanda: Another one to do a swap for the negative condition 😄

packages/insomnia-app/app/plugins/context/request.js
```diff
@@ -147,6 +154,82 @@ export function init(
...
+      return (renderedRequest.body.params || [])
+        .filter(p => !p.type || p.type === 'text')
+        .map(({ name, value }) => ({ name, value }));
+    },
+    getFileForm(): Array<any> {
+      return (renderedRequest.body.params || [])
+        .filter(p => p.type === 'file')
+        .map(({ name, fileName }) => ({ name, fileName }));
+    },
+    addFormItem(formItem: any): boolean {
```
develohpanda: I wonder if this can be typed, instead of `any`?

```suggestion
    addFormItem(formItem: FormItem): boolean {
```

packages/insomnia-app/app/common/misc.js
```diff
@@ -19,6 +19,13 @@ type Parameter = {
   value: string,
 };
 
+type FormItem = {
```
develohpanda: Good to see 💯 Typing is awesome!

packages/insomnia-app/app/plugins/context/request.js
```diff
@@ -147,6 +154,82 @@ export function init(
...
+    setTextFormItem(name: string, value: string): boolean {
+      return this.setFormItem({ type: 'text', name, value });
+    },
+    setFileFormItem(name: string, fileName: string): boolean {
+      return this.setFormItem({ type: 'file', name, fileName });
+    },
+    removeFormItem(name: string): void {
+      if (!renderedRequest.body.params) {
+        return;
+      }
```
develohpanda: A negative condition here is all good, another developer can understand it at a glace 👍

develohpanda (changes requested):
Hey @yittg! Sorry for the delay, thanks for putting this PR up. Awesome to see that you came back to build a feature, that _you_ requested in the early days! 🔥 

Overall, the code looks good to me. 😄 I've added a couple of comments to your use of conditionals that I feel are worth addressing.

An observation: For clarity and maintainability, conditions are typically easier to follow if they are kept positive (especially negative-or combinations). We could also alias a condition and set the result to a well named variable, to then consume in an if statement in order to make it easier to follow, test and debug.

Speaking of tests, I'd love to see some unit tests for the additional functions you've added! 💯

packages/insomnia-app/app/plugins/context/request.js
yittg: thanks, fixed.

yittg:
@develohpanda PTAnotherL, thanks

gschier:
Sorry to jump in here at the last minute but I think we should discuss the API design here before moving forward. 

And, just to be clear up front... you did a great job with this PR and we really appreciate your contribution! I just think it's worth putting in more thought up front since it's really hard to change a public API like this later. 💜

---

Okay, let's dive in. To get a feel for what was added in this PR, I'll list the proposed new methods here:

```text
addFileFormItem
addFormItem
addTextFormItem
getFileForm
getMimeType
getTextForm
getUploadFileName
removeFormItem
setFileFormItem
setFormItem
setMimeType
setTextFormItem
setUploadFileName
getFileForm
getMimeType
getTextForm
getUploadFileName
```

My first thought is "that's a lot of methods to handle body logic."  😅I think my biggest concern, however, is making it easy and safe to work with the various body types.

For example, with the proposed API, it's still possible to call `addTextFormItem()` even if the body is not currently a form. I can see this being difficult to debug, as the user will need to add additional logic to check the mimeType. 

Another pain point I see here if you want to switch body types. Currently there is no obvious way to do that with the proposed API. We could add more methods for doing that (eg. `setMultipartFormBody(parameters)` but now the API gets even larger.

You may be able to see where I'm going with this... Essentially, an API design should be clear, understandable, and prevent users from making errors. If this solution provided all of those things, I'd be willing to overlook the added surface area. However, if the user needs to add a bunch of checks/logic anyway, I propose we greatly simplify it to the following:

```js
getBody(): RequestBody
setBody(body: RequestBody): void
```

This essentially puts all responsibility on the plugin developer and also provides a flexible base for us to build on. Later, we can provide additional helper utilities to create and update various body types if needed.

I can imagine a set of body-specific helpers similar to the ones included in the [HTML5 fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Body).

For now, though, I vote to keep things as simple as possible.

Thoughts @develohpanda @yittg?

gschier (changes requested):
Sorry to jump in here at the last minute but I think we should discuss the API design here before moving forward. 

And, just to be clear up front... you did a great job with this PR and we really appreciate your contribution! I just think it's worth putting in more thought up front since it's really hard to change a public API like this later. 💜

---

Okay, let's dive in. To get a feel for what was added in this PR, I'll list the proposed new methods here:

```text
addFileFormItem
addFormItem
addTextFormItem
getFileForm
getMimeType
getTextForm
getUploadFileName
removeFormItem
setFileFormItem
setFormItem
setMimeType
setTextFormItem
setUploadFileName
getFileForm
getMimeType
getTextForm
getUploadFileName
```

My first thought is "that's a lot of methods to handle body logic."  😅I think my biggest concern, however, is making it easy and safe to work with the various body types.

For example, with the proposed API, it's still possible to call `addTextFormItem()` even if the body is not currently a form. I can see this being difficult to debug, as the user will need to add additional logic to check the mimeType. 

Another pain point I see here if you want to switch body types. Currently there is no obvious way to do that with the proposed API. We could add more methods for doing that (eg. `setMultipartFormBody(parameters)` but now the API gets even larger.

You may be able to see where I'm going with this... Essentially, an API design should be clear, understandable, and prevent users from making errors. If this solution provided all of those things, I'd be willing to overlook the added surface area. However, if the user needs to add a bunch of checks/logic anyway, I propose we greatly simplify it to the following:

```js
getBody(): RequestBody
setBody(body: RequestBody): void
```

This essentially puts all responsibility on the plugin developer and also provides a flexible base for us to build on. Later, we can provide additional helper utilities to create and update various body types if needed.

I can imagine a set of body-specific helpers similar to the ones included in the [HTML5 fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Body).

For now, though, I vote to keep things as simple as possible.

Thoughts @develohpanda @yittg?

yittg:
@gschier, I'm ok about the single simple API, just one thing should we mark the existed body text APIs as deprecated?

yittg:
Another question: should we update the content type header automatically if set body with different mimeType?

develohpanda:
Agree with the explanation Greg gave with what an API design should achieve and how this change to the API ensures simplicity, especially given the potential this has to grow. Thanks for your time and effort, @yittg! 💯 

> Should we mark the existing body text APIs as deprecated?

That one feels like a body-specific helper that Greg mentioned, so perhaps we need to find a better place to put it. Since helpers don't belong in the core API, my vote is to mark the existing one as deprecated and direct users to the `setBody` API.

Also an internal note, we should make sure to update the plugin API [docs](https://support.insomnia.rest/article/26-plugins#context.request).

> Should we update the content type header automatically if set body with different mimeType?

Kind of on the edge - I feel that we should given the Insomnia client already updates the header automatically, but that also means we need to be clear in docs about the additional automatic change this API may make, therefore adding complexity to the API. I'll defer to @gschier for this one! 👍

yittg:
> should we update the content type header automatically if set body with different mimeType?


@develohpanda  @gschier prefer to not do this implicitly, and just point out the potential issue in the document.

gschier:
I agree with not changing the header implicitly in this API. Again, it's easier to add functionality later than paint ourselves into an API corner before necessary 👍 

And, also yes on marking the old `(get|set)BodyText()` as deprecated. We don't have a great system for doing that currently, but we just deprecated our first method last week which can be used as an example (I'm also open to suggestions if you have a better way to mark something deprecated). https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/plugins/context/app.js#L103-L113

yittg:
That's great,  I'll fix it soon.

yittg:
Have marked the deprecated api.
However, i didn't find the source code of the document to update,    🤣

develohpanda (approved):
> I didn't find the source code of the document to update

Don't think that's in source code 😄

gschier:
Ya, the docs are hosted on HelpScout so we'll have to update them after the PR is merged 👍

yittg:
So then? Let's merge it? :)

yittg:
@gschier @develohpanda anything more needed?

gschier (approved):
Looks great @yittg! Thanks for taking care of this and being accepting to changes/feedback 🙂💜

yittg:
@gschier @develohpanda Thanks. Pleasure to contribute to the tool I use everyday, :)

PR #2020, opened 2020-04-01
Extend Certificates with Windows Cert Store + Custom CA bundle options

### LINKED ISSUE #675: [Feature] Ability to provide custom CA Certificate

### LINKED ISSUE #1432: [improvement] Use OS CA certs when possible

develohpanda:
Continuing with #1804 😄

- Default + Custom bundle options for Mac/Linux
- Default + Custom bundle + Windows Cert Store options for Windows

Tasks remaining from original PR:
- [ ] Unit tests (add/fix)
- [x] Investigate extracting some logic from `network.js` to clean-up code (seems okay to me)
- [ ] From #675: _Since Curl retrieve the base CA bundle from the OS on Mac, it's unclear whether or not setting CURLOPT_CAINFO will override this bundle entirely or add to it. This needs to be tested._
- [ ] Add screenshot from Windows showing the Windows Cert Store option

UX to be advised but here's an attempt. Alternatively, instead of hiding the custom file-picker we can always show it and only enable when relevant.

CA-cert

Credit to @coolhome for picking this up, I'm just getting it across the line! 🔥 

Closes #675, closes #1432.


packages/insomnia-app/app/models/settings.js
```diff
@@ -51,6 +51,8 @@ type BaseSettings = {
   useBulkHeaderEditor: boolean,
   useBulkParametersEditor: boolean,
   validateSSL: boolean,
+  caBundle: string,
```
gschier: Should this be a union type? `'' | 'windowsCertStore' | 'userProvided'`

gschier: Also, what are your thoughts on combining these into an Object property, since they're very closely related?

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -158,6 +163,53 @@ class General extends React.PureComponent<Props, State> {
     return this.renderTextSetting(label, name, help, { ...props, type: 'number' });
   }
 
+  renderCertificateBundleSettings() {
+    const { settings } = this.props;
+
+    const userProvidedValue = 'userProvided';
```
gschier: +1 for enum for this

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -347,6 +397,10 @@ class General extends React.PureComponent<Props, State> {
           )}
         </div>
 
+        <h3>Certificate Authority</h3>
```
gschier: This header seems a bit too specific to encompass the Validate Certificates setting. Can we rename it to Certificates instead?

Also, in the future note, it might be nice to separate the `SSL_VERIFYHOST` and `SSL_VERIFYPEER` options into their own settings for more control. Right now, we're setting both on libcurl from a single setting.

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -158,6 +163,53 @@ class General extends React.PureComponent<Props, State> {
...
+    const userOption = <option value={userProvidedValue}>Custom Bundle</option>;
+
+    const disabled = !settings.validateSSL;
+    return (
+      <React.Fragment>
+        {this.renderBooleanSetting('Validate certificates', 'validateSSL', '')}
+        <div className="form-row pad-top-sm">
+          <div className="form-control form-control--outlined">
+            <label>
+              Select Certificate Bundle
```
gschier: Can we remove the `Select` prefix from these two labels?  💄

gschier (changes requested):
Getting close! Looked at it with some fresh eyes and jotted down some thoughts.

packages/insomnia-app/app/network/network.js
```diff
@@ -349,9 +353,37 @@ export async function _actuallySend(
...
+        const fullCAPath = pathJoin(baseCAPath, 'windows.pem');
+        let caFile = '';
+
+        await new Promise(resolve => {
+          winCa({
+            format: winCa.der2.pem,
+            store: ['root'],
+            save: fullCAPath,
+            onsave: folder => {
+              caFile = folder + '/roots.pem';
```
coolhome: According to win-ca documentation `onsave` will have `undefined` passed if it failed to save certificate to disk.

I wonder if we should check this and reject the promise or maybe at least a improved log message?

packages/insomnia-app/app/models/settings.js
develohpanda: Hmm, I like the idea of it, also would want to pull `validateSSL` into that object property typed as `CertSettings` or something.

We would have to make the update settings functions a little smarter to work with updating nested properties (eg. `certSettings.caBundlePath`). I did something similar for the plugin settings (which right now only has enable/disable).

Do you reckon I should try it as part of this PR? Would want to do it cleanly. It _is_ something we can introduce after this PR with a migration 🤔

packages/insomnia-app/app/ui/components/settings/general.js
develohpanda: Would be good to do that, though I think it needs some more thought about when that user setting should kick in during consumption (in `network.js`).

packages/insomnia-app/app/network/network.js
develohpanda: Noted, I'll need to play with this on Windows to see what the UX is like. Thanks! 😄

packages/insomnia-app/app/models/settings.js
gschier: Ya, we could hold off for now as it's not a huge deal 👍

packages/insomnia-app/app/network/network.js
develohpanda: Rejecting the promise will stop the request as a whole.

image

I've resorted to providing a nicer message in console log and not setting CAINFO.

develohpanda:
Hi @coolhome! Would you have any capacity to help test and validate this PR? I didn't get very far with smoke-testing, as I'm unfamiliar with how to quickly setup a backend and CA bundle that would allow for testing this manually. 😄

Aside from some unit testing to lock down the new functionality, this is ready to roll.

coolhome:
@develohpanda I have been trying to create a working build to transfer to my work machine for testing. In the process I have discovered that `win-ca` with the current setup for insomnia are not compatible. It works when you run start but on a packaged version we get the following error:
image

Upon investigation I found the path `win-ca` is trying to execute a program to receive the windows certificate but fails because it has the path to the asar and not the unpacked location where the `roots.exe` exists.

image
ref in dependency: https://github.com/ukoloff/win-ca/blob/e48dfd163783d8be9949b0a764780f7d8d23fbcb/src/fallback.ls#L72

The file does exists here tho:
image


Switching the import statement to this at least gets the app to bootup but now the error displays in the http response box.
```js
import winCa from 'win-ca/api';
```

ref: https://github.com/ukoloff/win-ca#entry-points

Likely need to do more research. I found this issue that looks like our issue. https://github.com/ukoloff/win-ca/issues/23

---

EDIT: `win-ca@3.2.1` did not resolve the issue either..

coolhome:
In `win-ca` there is a line to get the `roots.exe` location using a relative path off `__dirname`. In this case the `__dirname` resolves to `resources/app.asar/roots.exe`. ASAR is like a zip to my understanding. We are obviously unable to extract the executable and run from a package. Good news is the package was put in the ASAR unpacked at `resources\app.asar.unpacked\node_modules\win-ca\lib\roots.exe`

Setting webpack `node.__dirname` to `false` (current value)
 -> Error: `. spawn .\roots.exe ENOENT`
 -> Code:
```js
// __dirname is ".\"
 bin = path.join(__dirname, 'roots.exe');
```

Setting webpack `node.__dirname` to `true` 
 -> Error: `. spawnSync ..\node_modules\win-ca\lib\roots.exe ENOENT`
 -> Code:
```js
// __dirname is "..\node_modules\win-ca\lib\"

 bin = path.join(__dirname, 'roots.exe');
```

When this is true we are pretty close. Unfortunately we are in ASAR when this codes executed so it does not know the path should have `resources\app.asar.unpacked\` prepended to the path.


So enabling [`node.__dirname`](https://webpack.js.org/configuration/node/#node__dirname) to preserve nodejs module paths gets us closer but we still have a problem. I think I will need to see the possibilities of switching `win-ca` to use [`electron-util`]([https://github.com/sindresorhus/electron-util](https://github.com/sindresorhus/electron-util)) over [`is-electron`]([https://github.com/cheton/is-electron](https://github.com/cheton/is-electron)). This new package will allow us to know if `win-ca` is in a electron, detect if its in a ASAR, and use the correct path.
-----
EDIT: I'm not sure if node.__dirname is the correct approach. I think false is better. I will post more if I figure out how to get win-ca and insomnia packaged together.

develohpanda:
Thanks for the help @coolhome, a couple of people have requested this so hoping to get it in soon! 🤗 

I found the `win-ca` documentation stating it has been adapted to work within an Electron app and support ASAR, with an example: https://github.com/ukoloff/win-ca#electron; suppose this doesn't work for our case.

I wonder if it could be related to our webpack configuration rather than `win-ca` specifically. @gschier does this behavior ring any bells with any other dependencies that have given issues during packaging?

gschier:
Unfortunately, I don't know off the top of my head why it _wouldn't_ be working. However, we _have_ worked-around a similar problem before.

@coolhome, we bundle `yarn` inside the app to download and install plugins, and had to list it in `extraResources` in the `electron-builder` config to force it to live outside the `asar` package. If you search the codebase for `yarn-standalone.js` you can perhaps follow a similar pattern to get it working. 

My guess is that you'd need to add a config option like this

```js
    {
      "from": "./node_modules/win-ca/../",
      "to": ".",
      "filter": "roots.exe"
    }
```

develohpanda:
Hi @coolhome! I've pushed a commit that marks `roots.exe` as per Greg mentioned, and was able to package, run the app and make a simple HTTP request (albeit on macOS).

The error you saw, did you see that on first launching the packaged app or when running a request?

Quick note, that you'll need to update the fallback [here](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/scripts/getBuildContext.js#L34) to the current build version (`2020.4.0-beta.4`), in order to build locally. (Working on this to improve it, slowly 😄)

develohpanda:
Closed in favor of #3203

Issue #2023, opened 2020-04-02
GraphQL query prettification doesn't handle a large number of arguments very well


chasingmaxwell:
When a query contains a field or named query with a large number of arguments, the "Prettify GraphQL" button condenses them all down to one line when a multi-line format would be easier to read. Here's a very contrived example:

![Selection_081](https://user-images.githubusercontent.com/3128659/78255378-3bddea80-74c5-11ea-8ee6-625bdf19884b.jpg)

I think something like this would be easier to read:

![Selection_082](https://user-images.githubusercontent.com/3128659/78255564-70ea3d00-74c5-11ea-888e-11b1c7838028.jpg)

Using prettier for the formatting of GraphQL queries would actually do this (and is what tools like playground use).

seanghay:
Love it

gschier:
While I agree that having prettier be part of the app would be awesome, it's a huge dependency to be adding just to format a GraphQL query.

I was talking to the team about this yesterday and we came to the conclusion that this would be great to expose as a plugin API behavior. A "Prettier Format" plugin would be self-contained and easy to install for those who want it.

gschier:
I was talking to the team about this yesterday and we came to the conclusion that this would be great to expose as a plugin API behavior. A "Prettier Format" plugin would be self-contained and easy to install for those who want it.

chasingmaxwell:
Sounds reasonable. I should be able to put that together shortly.

rocketraman:
Work seems to be happening under https://github.com/Kong/insomnia/issues/3242.

dimitropoulos:
this should be fixed by https://github.com/Kong/insomnia/pull/4022 which should be in the next release

PR #2024, opened 2020-04-02
Use prettier for GraphQL query prettification

chasingmaxwell:
This PR uses prettier to format GraphQL queries which addresses #2023.

There were some diffs in a couple `package-lock.json` files that seem to be related to my environment. Particularly the removal of `"dev": true` all over the place. I did some looking up on how to resolve this, but was having little luck. If this is undesirable, do you have any suggestions for what I can change locally to avoid it? I'm using nvm to manage the node version according to the `.nvmrc` file. Maybe npm version?


gschier (changes requested):
While I agree that having prettier be part of the app would be awesome, it's a huge dependency to be adding just to format a GraphQL query. Do you think there's a way to do this with the current `graphql` library?

chasingmaxwell:
I'll look into it a bit more. If not, I'm happy to go the plugin route.

chasingmaxwell:
There is not a way to do this currently with the graphql library. There is an [issue](https://github.com/graphql/graphql-js/issues/1799) to add pretty-printing and some related [pull requests](https://github.com/graphql/graphql-js/pull/1557), but for now `print` is the only option and its primary function is not for prettifying. It looks like the recommended approach for now is to use prettier so a plugin is the way to go here.

PR #2026, opened 2020-04-03
New plugin API for showing custom UI inside a dialog

gschier:
This is the first step towards providing the ability for plugins to show custom (and dynamic) UI.

- [x] Add new `context.app.dialog()` API
- [x] Deprecate old (undocumented) `context.app.showGenericDialog` method

This PR adds a new `dialog()` method which can accept a DOM node to show inside the dialog's body. Having this be a DOM node instead of static HTML allows plugins to mount a UI framework (eg. React) to the node before showing.

Here's a code example of a plugin that shows the classic "counter example" inside of a `dialog()`

```js
// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ //
// Sample Folder Action to show counter app //
// ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ //

import React, { useState } from 'react';
import ReactDOM from 'react-dom';

export const requestGroupActions = [{
  label: 'React Demo',
  icon: 'fa-star',
  action: (context) => {
    const root = document.createElement('div');
    ReactDOM.render(<Counter />, root);

    context.app.dialog('React Counter App', root, {
      onHide: () => ReactDOM.unmountComponentAtNode(root),
    });
  },
}];

function Counter() {
  const [count, setCount] = useState(0);
  return (
    <React.Fragment>
      <p>You clicked {count} times!</p>
      <button onClick={() => setCount(count + 1)}>Increment</button>
    </React.Fragment>
  );
}

```

### DIFF
--- a/packages/insomnia-app/app/plugins/context/__tests__/app.test.js
+++ b/packages/insomnia-app/app/plugins/context/__tests__/app.test.js
@@ -10,6 +10,7 @@ describe('init()', () => {
     expect(Object.keys(result)).toEqual(['app']);
     expect(Object.keys(result.app).sort()).toEqual([
       'alert',
+      'dialog',
       'getPath',
       'prompt',
       'showGenericModalDialog',
--- a/packages/insomnia-app/app/plugins/context/app.js
+++ b/packages/insomnia-app/app/plugins/context/app.js
@@ -100,11 +100,12 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): { a
 
       /** @deprecated as it was never officially supported */
       showGenericModalDialog(title: string, options?: { html: string } = {}): void {
-        console.warn(
-          'app.showGenericModalDialog() is a deprecated plugin API. Use app.dialog() instead.',
-        );
+        console.warn('app.showGenericModalDialog() is deprecated. Use app.dialog() instead.');
+
+        // Create DOM node so we can adapt to the new dialog() method
         const body = document.createElement('div');
         body.innerHTML = options.html;
+
         return this.dialog(title, { body });
       },
     },
--- a/packages/insomnia-app/app/plugins/context/app.js
+++ b/packages/insomnia-app/app/plugins/context/app.js
@@ -26,9 +26,10 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): { a
       },
       dialog(
         title,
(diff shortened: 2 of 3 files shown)


develohpanda (approved):
Looks awesome!

One question I have is around the example plugin in the description. Is unmounting a requirement (`ReactDOM.unmountComponentAtNode(root)`) for UI plugins to work, or is that specific for dialogs? We might need to update the parameters to reflect that 😄 

Also! What are your thoughts on separating the UI APIs into `context.ui`? Alternatively, how do you see the new UI APIs being grouped under the `context` object (given the current groupings of `request`, `response`, `store`, `app`, `data`, and `network`)?

gschier:
Good questions @develohpanda!

> One question I have is around the example plugin in the description. Is unmounting a requirement (`ReactDOM.unmountComponentAtNode(root)`) for UI plugins to work, or is that specific for dialogs? We might need to update the parameters to reflect that 😄

Unmounting is required for React, otherwise the component will remain in memory after the dialog closes. This logic will be different (or unnecessary) depending on what UI library (if any) is being used. You may also want to do other things on hide depending on your plugin.

> Also! What are your thoughts on separating the UI APIs into `context.ui`? Alternatively, how do you see the new UI APIs being grouped under the `context` object (given the current groupings of `request`, `response`, `store`, `app`, `data`, and `network`)?

The purpose of `app` _is_ essentially UI right now. It's purpose is currently for reserved for things that are specific to the user (showing alerts, save dialogs, etc). However, it includes things outside UI as well, but still specific to the user, like getting the app data folder (a user-specific setting based on desktop environment). I would generally agree, though, that a `ui` context could make sense but it would require deprecating/changing a bunch of the existing public API.

develohpanda:
> ...otherwise the component will remain in memory after the dialog closes. This logic will be different (or unnecessary)...

Fair enough! We should make sure to call out some of these gotchas when it comes to documentation. 👍 

> ...a `ui` context could make sense but it would require deprecating/changing a bunch of the existing public API

I quite like how the theme system it split into `sidebar`, `header` etc, and I feel like it would be intuitive if UI API could also be grouped that way. It looks like `App` serves the purpose for more _global_ UI (alerts and dialogs, since we don't have any other UI plugin functionality), so maybe future UI API can still be exposed, without necessarily replacing or deprecating the existing API. Future discussion 😄

PR #2030, opened 2020-04-08
Fix closing modal when pressed ESC while editing environment

### LINKED ISSUE #2029: [Bug] [User-Experience] Pressing ESC while renaming object closes whole modal dialog
**Describe the bug**
In some cases, if one's renaming an object, and presses ESC (to discard rename), the triggered action is exiting the current modal dialog, if any is open

**To Reproduce**
1. Go to 'Manage Environments'
2. Double click on an existing environment to start renaming it
3. Press ESC key
4. The entire environment management window closes

**Expected behavior**
(4.) The environment management window is still open, and renaming is discarded

**Screenshots**

 - OS: Linux Mint 19
 - Installation Method: tar.gz
 - App Version 7.1.1

**Additional context**

serpilacar:
While editing an environment variable, if user presses ESC key the modal closes and variable is saved. 

Modal shouldn't be closed and variable shouldn't be saved.

I've added `noEscape` prop as true to the Modal in `WorkspaceEnvironmentsEditModal` to prevent Modal from catching ESC key press. (`_handleKeyDown` method was catching ESC key.)

Also removed a TODO comment written 3 years ago by resetting environment variable to default if user hasn't saved while editing (presses ESC key).

Closes #2029

### DIFF
--- a/packages/insomnia-app/app/ui/components/base/editable.js
+++ b/packages/insomnia-app/app/ui/components/base/editable.js
@@ -55,8 +55,7 @@ class Editable extends PureComponent {
       this._handleEditEnd();
     } else if (e.keyCode === 27) {
       // Pressed Escape
-      // NOTE: This blur causes a save because we save on blur
-      // TODO: Make escape blur without saving
+      this._input.value = this.props.value;
       this._input && this._input.blur();
     }
   }
--- a/packages/insomnia-app/app/ui/components/base/editable.js
+++ b/packages/insomnia-app/app/ui/components/base/editable.js
@@ -6,13 +6,10 @@ import autobind from 'autobind-decorator';
 class Editable extends PureComponent {
   constructor(props) {
     super(props);
+    this._input = React.createRef();
     this.state = {
       editing: false,
     };
-  }
-
-  _handleSetInputRef(n) {
-    this._input = n;
   }
 
   _handleSingleClickEditStart() {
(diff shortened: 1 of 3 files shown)


baleeds:
Would this PR disable the ability to close the modal with escape?  I believe the "Manage Environments" modal should save and be closed on escape.  I personally use that feature on a daily basis.  The linked issue is regarding editing the name of an environment and hitting escape to discard the name.

serpilacar:
@baleeds Yes, turns out this would disable the close with escape key functionality on modal. I updated PR to set noEscape prop conditionally to prevent disabling this completely.

baleeds:
Thanks for looking into it! I haven't looked at the code so I could be wrong, but could we stop propagation on the escape key event during field edit mode? That would isolate the behavior of the field from its context.

serpilacar:
Stop event propagation while editing/finish editing was what came to my mind when I first check this issue actually but I couldn't able to stop bubbling.

baleeds:
I think I was able to figure it out: #2039.

My limited understanding is that the `KeydownBinder` is not respecting React's SyntheticEvents, based on the fact that the keydown event in the modal doesn't have a `persist()` function.  This would be why stopping propagation in the editable component didn't work, because it would stop the SyntheticEvent, which has no bearing on the actual DOM event (I think!).

In order to fix the issue I used the `KeydownBinder` to "line up" the events, so both are using raw DOM events (I think!).

That could all be wrong, but it does appear to work 😆

PR #2039, opened 2020-04-16
fix: Escape key in editable field closing modals

### LINKED ISSUE #2029: [Bug] [User-Experience] Pressing ESC while renaming object closes whole modal dialog

baleeds:
This PR will:
- Fix an issue where pressing escape in an editable field closed the containing modal
- Fix an issue where pressing escape in an editable field would save the current value, instead of discarding

Thanks to @serpilacar for spearheading this one!

This PR is an alternative (for better or worse) to https://github.com/Kong/insomnia/pull/2030.  The primary difference is that this PR attempts to squash the event at the source.  There are potential downsides to this approach since it could affect more than just the cases I tested.  The advantage is that this change would prevent similar bugs from occurring in the future.

Closes #2029

### DIFF
--- a/packages/insomnia-app/app/ui/components/base/editable.js
+++ b/packages/insomnia-app/app/ui/components/base/editable.js
@@ -1,6 +1,7 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
+import KeydownBinder from '../keydown-binder';
 
 @autobind
 class Editable extends PureComponent {
@@ -38,11 +39,14 @@ class Editable extends PureComponent {
     const value = this._input.value.trim();
 
     if (!value) {
-      // Don't do anything if it's empty
+      // Don't do anything if it's empty.
       return;
     }
 
-    this.props.onSubmit(value);
+    if (value !== this.props.value) {
+      // Don't run onSubmit for values that haven't been changed.
+      this.props.onSubmit(value);
+    }
 
     // This timeout prevents the UI from showing the old value after submit.
     // It should give the UI enough time to redraw the new value.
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/base/editable.js
```diff
@@ -55,9 +59,15 @@ class Editable extends PureComponent {
...
     } else if (e.keyCode === 27) {
       // Pressed Escape
-      // NOTE: This blur causes a save because we save on blur
-      // TODO: Make escape blur without saving
-      this._input && this._input.blur();
+
+      // Prevent bubbling to modals and other escape listeners.
+      e.stopPropagation();
+
+      if (this._input) {
```
gschier, line 66: I might be missing something but don't we want to have `this._input.blur()` here still?

gschier (changes requested):
Thanks so much for taking care of this! Just one comment of the removal of the `blur` logic.

packages/insomnia-app/app/ui/components/base/editable.js
baleeds: It's possible!  I removed it because the text input is unmounted when `editing` becomes false, so in my mind the blur state is irrelevant.

What would be the reason it's needed?

gschier: I think the blur is there so the `onBlur` handler gets called, which actually just calls `_handleEditEnd` anyway.

I think I see why you made this change now. Currently, pressing `escape` blurs the input but still saves the value. With your change, the value will be reverted to the original. I like it!

gschier (approved):
Thanks for taking care of this!

PR #2040, opened 2020-04-17
Prevent migrations from running for database.duplicate

develohpanda:
Once an entity has been loaded from the database, it is already in a valid state - migrations have run, and defaults during `init()` have been set.

Therefore, when duplicating an entity, migrations do **not** need to be run, and the entity does **not** need initialization.

It is seemingly harmless to run `migrate()` and `init()` during duplication, however if the `migrate()` function on the model does complex tasks such as creating other entities, then surplus children can be created.

A duplicate should be just that, a duplicate - it shouldn't have any side effects.

This PR simplifies the duplicate function to remove the `migration()` and `init()` calls from the process, plus adds unit tests to validate these. 😄


packages/insomnia-app/app/common/database.js
```diff
@@ -483,6 +489,7 @@ export async function docUpdate<T: BaseModel>(
   originalDoc: T,
   ...patches: Array<Object>
 ): Promise<T> {
+  // No need to re-initialize the model during update; originalDoc will be in a valid sate by virtue of loading
```
gschier: Typo: "sate" 😝

packages/insomnia-app/app/common/database.js
```diff
@@ -597,13 +604,20 @@ export const duplicate = (database.duplicate = async function<T: BaseModel>(
...
     // 1. Copy the doc
-    const newDoc = Object.assign({}, docToCopy, patch);
-    delete newDoc._id;
-    delete newDoc.created;
-    delete newDoc.modified;
+    const newDoc = Object.assign({}, docToCopy, patch, overrides);
 
-    const createdDoc = await docCreate(newDoc.type, newDoc);
+    // Don't initialize the model during insert, and simply duplicate
+    const createdDoc = await database.insert(newDoc, false, false);
```
gschier: Nice! Ended with pretty much the same number of lines 😄

gschier (dismissed):
LGTM if we can get the CI tests working 👍 😄

packages/insomnia-app/app/common/__tests__/database.test.js
```diff
@@ -300,3 +300,58 @@ describe('_fixThings()', () => {
...
+});
+
+describe('docCreate()', () => {
+  beforeEach(globalBeforeEach);
+  afterEach(() => jest.resetAllMocks());
+
+  it('should call migrate when creating', async () => {
+    const spy = jest.spyOn(models.workspace, 'migrate');
+
+    await db.docCreate(models.workspace.type, {
```
develohpanda, line 350: Fails during `npm run test` but passed when running in IDE. 😢 

image

gschier (approved):
⭐

PR #2043, opened 2020-04-22
Remove insomnia-components

develohpanda:
Remove insomnia-components in preparation for an updated package. Removed from `insomnia-app/package.json` as well, given it is not currently consumed anywhere in the code-base.


gschier (approved):
Looks good!

PR #2058, opened 2020-04-27
fix: correct curl import to support multi -d params

### LINKED ISSUE #2057: [Bug] Some post params are not imported from curl
When importing request from curl, it seems that we drop some --data params
 
**To Reproduce**

- Create a new request 
- select POST, Form URL Encoded
- add param key "Hello" value "World"
- add param key "Foo" value "Baz"
- click copy as curl

- then create a new request
- past the curl in the url to import it

Only Hello=World is reported in query params

query : 

```bash
curl --request POST \
  --url http://example.com/ \
  --header 'content-type: application/x-www-form-urlencoded' \
  --data Hello=world \
  --data foo=baz
```

after import

image

**Expected behavior**
Hello=World
and Foo=Baz  are imported

 - OS: 10.15.3 (19D76)
 - Installation Method: dmg
 - App Version:  7.1.1 (7.1.1.3350)

eMerzh:
when importing curl -d or --data params, take more than the first one.
fix #2057

```
curl --request POST \
  --url http://example.com/ \
  --header 'content-type: application/x-www-form-urlencoded' \
  --data Hello=world \
  --data foo=baz
```
is correctly imported

image

### DIFF
--- a/packages/insomnia-importers/src/__tests__/fixtures/curl/multi-data-output.json
+++ b/packages/insomnia-importers/src/__tests__/fixtures/curl/multi-data-output.json
@@ -1,0 +1,37 @@
+{
+  "_type": "export",
+  "__export_format": 4,
+  "__export_date": "2016-11-18T22:34:51.526Z",
+  "__export_source": "insomnia.importers:v0.1.0",
+  "resources": [
+    {
+      "_id": "__REQ_1__",
+      "_type": "request",
+      "parentId": "__WORKSPACE_ID__",
+      "url": "https://insomnia.rest",
+      "name": "https://insomnia.rest",
+      "method": "POST",
+      "body": {
+        "mimeType": "application/x-www-form-urlencoded",
+        "params": [
+          {
+            "name": "foo",
+            "value": "bar"
+          },
+          {
+            "name": "baz",
+            "value": "qux"
+          }
+        ]
+      },
+      "parameters": [],
+      "headers": [
+        {
+          "name": "Content-Type",
+          "value": "application/x-www-form-urlencoded"
+        }
+      ],
+      "authentication": {}
+    }
+  ]
+}
--- a/packages/insomnia-importers/src/importers/curl.js
+++ b/packages/insomnia-importers/src/importers/curl.js
@@ -174,16 +174,23 @@ function importArgs(args) {
 
   // Body (Text or Blob)
   const bodyAsGET = getPairValue(pairs, false, 'G', 'get');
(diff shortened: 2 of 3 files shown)


packages/insomnia-importers/src/importers/curl.js
```diff
@@ -174,16 +174,23 @@ function importArgs(args) {
...
     'data',
     'data-raw',
     'data-urlencode',
     'data-binary',
     'data-ascii',
-  );
+  ]) {
+    if (pairs[paramName] && pairs[paramName].length) {
+      textBodyParams = textBodyParams.concat(pairs[paramName]);
+    }
```
develohpanda, line 189: We can tidy this one up to reduce the number of operations against an array, and make it a little more performant by using `array.push` instead of `array.concat`. ~~We also recently enabled optional chaining!~~ Not in this package yet.

So with that, I suggest something like the following:

```suggestion
    const pair = pairs[paramName];
    if (pair && pair.length) {
      textBodyParams.push(pair);
    }
```

packages/insomnia-importers/src/__tests__/fixtures/curl/multi-data-output.json
```diff
@@ -0,0 +1,37 @@
+{
```
develohpanda, line 1: Nice, awesome to see tests 💯

develohpanda (changes requested):
Thanks for the PR, @eMerzh!

Overall looks good, one small suggestion. 😄

packages/insomnia-importers/src/importers/curl.js
develohpanda: Kidding, I totally misread this! `pairs` is a nested array. 👍

develohpanda (approved):
Misread during my first review 🔥

packages/insomnia-importers/src/importers/curl.js
```diff
@@ -174,16 +174,23 @@ function importArgs(args) {
 
   // Body (Text or Blob)
   const bodyAsGET = getPairValue(pairs, false, 'G', 'get');
-  const textBody = getPairValue(
-    pairs,
-    null,
+
+  let textBodyParams = [];
+  for (const paramName of [
```
nijikokun, line 179: This list should probably be broken out into its own variable (preferably hoisted outside of the function) so it's not instantiated every time this is ran

develohpanda:
Merged as #2070 😄

PR #2063, opened 2020-04-28
Add filename to apiSpec model and decouple from workspace name

develohpanda:
From the documents homepage, you are able to rename a document.

If you then open the document, and navigate from design to debug, it reverts the document name to the `info.title` field from the api spec.

This PR decouples the document name from the workspace name. As such, there is now a requirement for an `apiSpec` to exist for a `workspace`, so migrations have been added to ensure this. This relationship may change in the future, from 1:1 to 1:many.

rename

The gif is too slow to show this, but I don't like how the window title update is a little janky. That is to do with the order in which `activeWorkspace` and `activeApiSpec` are updated in redux (they don't happen together). Maybe a separate action that updates _both_ in the few cases where we want to update both (clicking a card on document home)?

Closes https://github.com/Kong/studio/issues/278


sonicyeti (approved):
Awesome

PR #2064, opened 2020-04-28
Control code-editor indenting by tab and shift-tab

develohpanda:
CodeMirror defaults `Shift-Tab` to `indentAuto` instead of `indentLess`, whereas most code editors use the latter.

This PR updates the keymap defaults used by Insomnia and designer. I know @nijikokun will like this one 😉  Worth pulling locally and testing to ensure this is desired behavior.

Any proposition for a custom keymap to `indentAuto`, now that we have overwritten `Shift-Tab`?

indenting

From CodeMirror [manual](https://codemirror.net/doc/manual.html)
image


sonicyeti (approved):
Woot woot!

PR #2069, opened 2020-04-28
Handle deletion of last document

develohpanda:
Similar to Insomnia Core, Designer will notify of deletion of the last document and create a new workspace before deleting.

For Designer, the new workspace will be defaulted to `scope = 'spec'`.

Cleanup: There is an additional change in the redux init method, that ensures the correct activity is started depending on the selected app id (during development or production). The same validation runs when changing an activity, however that check is disabled while in development (to allow for real-time toggle) and only runs in production.

delete


sonicyeti (dismissed):
Nice!

packages/insomnia-app/app/models/workspace.js
```diff
@@ -45,7 +46,7 @@ export async function all(): Promise<Array<Workspace>> {
   const workspaces = await db.all(type);
 
   if (workspaces.length === 0) {
-    await create({ name: getAppName() });
+    await create({ name: getAppName(), scope: getAppId() === APP_ID_DESIGNER ? 'spec' : null });
```
gschier, line 49: I forgot we added the `scope` attribute. Do we actually use it anywhere?

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -618,5 +608,8 @@ export function init() {
     // Nothing here...
   }
 
+  // If the default app id is insomnia, then default to the insomnia view at initialization
```
gschier, line 611: Nice 👍

gschier (approved):
Looks good!

packages/insomnia-app/app/models/workspace.js
develohpanda: Yah. It feels redundant, but was the easiest fix atm.

Used for rendering a card on the homepage: https://github.com/Kong/insomnia/blob/9037a9290642ac52407f6ace2e5fe6b8892d068e/packages/insomnia-app/app/ui/components/wrapper-home.js#L351-L363

PR #2074, opened 2020-04-29
Set filename of apiSpec using Workspace name during db repair

develohpanda:
Went onto develop, and reset all ApiSpec filenames to be an empty string
image

Went onto this branch, and after launching, ensure `apiSpec.fileName` = `workspace.name`.
image

This will recover the fileNames for any documents that are called 'Insomnia Designer', but that'll be updated in future to only recover names of those that are missing a fileName. 

Feel free to merge if there is pressing need for this.

### DIFF
--- a/packages/insomnia-app/app/common/__tests__/database.test.js
+++ b/packages/insomnia-app/app/common/__tests__/database.test.js
@@ -78,13 +78,19 @@ describe('bufferChanges()', () => {
     // Assert changes seen after flush
     await db.flushChanges();
     expect(changesSeen).toEqual([
-      [[db.CHANGE_INSERT, newDoc, false], [db.CHANGE_UPDATE, updatedDoc, false]],
+      [
+        [db.CHANGE_INSERT, newDoc, false],
+        [db.CHANGE_UPDATE, updatedDoc, false],
+      ],
     ]);
 
     // Assert no more changes seen after flush again
     await db.flushChanges();
     expect(changesSeen).toEqual([
-      [[db.CHANGE_INSERT, newDoc, false], [db.CHANGE_UPDATE, updatedDoc, false]],
+      [
+        [db.CHANGE_INSERT, newDoc, false],
+        [db.CHANGE_UPDATE, updatedDoc, false],
+      ],
     ]);
   });
 });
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/common/__tests__/database.test.js
```diff
@@ -247,23 +252,31 @@ describe('_fixThings()', () => {
...
     });
 
     await models.cookieJar.create({
       _id: 'j2',
       parentId: 'w1',
-      cookies: [{ id: '1', key: 'foo', value: '2' }, { id: 'j2_1', key: 'j2', value: '2' }],
+      cookies: [
+        { id: '1', key: 'foo', value: '2' },
+        { id: 'j2_1', key: 'j2', value: '2' },
+      ],
```
develohpanda, line 275: Linting

packages/insomnia-app/app/common/__tests__/database.test.js
```diff
@@ -274,13 +287,20 @@ describe('_fixThings()', () => {
...
+        ],
       },
       {
         _id: 'j2',
         parentId: 'w1',
-        cookies: [{ id: '1', key: 'foo', value: '2' }, { id: 'j2_1', key: 'j2', value: '2' }],
+        cookies: [
+          { id: '1', key: 'foo', value: '2' },
+          { id: 'j2_1', key: 'j2', value: '2' },
+        ],
```
develohpanda, line 301: Linting

packages/insomnia-app/app/common/__tests__/database.test.js
```diff
@@ -78,13 +78,19 @@ describe('bufferChanges()', () => {
...
     ]);
 
     // Assert no more changes seen after flush again
     await db.flushChanges();
     expect(changesSeen).toEqual([
-      [[db.CHANGE_INSERT, newDoc, false], [db.CHANGE_UPDATE, updatedDoc, false]],
+      [
+        [db.CHANGE_INSERT, newDoc, false],
+        [db.CHANGE_UPDATE, updatedDoc, false],
+      ],
```
develohpanda, line 93: Linting

packages/insomnia-app/app/models/api-spec.js
```diff
@@ -9,6 +9,7 @@ export const canDuplicate = true;
 export const canSync = false;
 
 type BaseApiSpec = {
+  name: string,
```
gschier: Don't need this, right?

develohpanda: Added it suspecting the git sync issue, but no not needed

gschier: That one turned out to be a different issue so this won't be necessary

gschier (approved):
Nice! I like this migration here much better :)

PR #2076, opened 2020-04-29
Fixing hit area for card menu

sonicyeti:
Addressing an issue where the hovered state of the card menu trigger was not accurately reflecting the click target visually.

Recording:
dd-hit-target


gschier (approved):
Nice!

PR #2077, opened 2020-04-29
Suppress lint on init when spec empty or default value

sonicyeti:
Fixes regression where if the spec contains nothing OR the default value, we suppress linting.

Preview (notice lint indicator on empty doc):

Fix:

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
@@ -760,7 +760,7 @@ class CodeEditor extends React.Component {
   _codemirrorValueBeforeChange(doc, change) {
     const value = this.codeMirror.getDoc().getValue();
     // Suppress lint on empty doc or single space exists (default value)
-    if (value === '' || value === ' ') {
+    if (value.trim() === '') {
       this.codeMirror.setOption('lint', false);
     } else {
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -759,8 +759,8 @@ class CodeEditor extends React.Component {
 
   _codemirrorValueBeforeChange(doc, change) {
     const value = this.codeMirror.getDoc().getValue();
-
-    if (value === '') {
+    // Suppress lint on empty doc or single space exists (default value)
+    if (value === '' || value === ' ') {
```
gschier: Sorry I broke this yesterday! Instead of checking default, can we use this to ignore any amount of whitespace instead?

```suggestion
    if (value.trim() === '') {
```

gschier (changes requested):
Just one suggestion here 👍

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
sonicyeti: Ahh nice nice, yes!

sonicyeti: Ahh nice nice, yes!

PR #2099, opened 2020-04-30
Handle non-string types in filterHeaders()

gschier:
This is a small fix for a bug report that was submitted to support. I'm not sure how it happened, but somehow the user ended up with a request header with a non-string name.

### DIFF
--- a/packages/insomnia-app/app/common/__tests__/misc.test.js
+++ b/packages/insomnia-app/app/common/__tests__/misc.test.js
@@ -43,6 +43,7 @@ describe('filterHeaders()', () => {
     expect(misc.filterHeaders(['bad'], null)).toEqual([]);
     expect(misc.filterHeaders(['bad'], 'good')).toEqual([]);
     expect(misc.filterHeaders(null, 'good')).toEqual([]);
+    expect(misc.filterHeaders([{ name: '', value: 'valid' }], '')).toEqual([]);
     expect(misc.filterHeaders([{ name: 123, value: 123 }], 123)).toEqual([]);
     expect(misc.filterHeaders([{ name: 'good', value: 'valid' }], 123)).toEqual([]);
     expect(misc.filterHeaders([{ name: 'good', value: 'valid' }], null)).toEqual([]);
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/common/misc.js
```diff
@@ -29,12 +29,12 @@ export function filterParameters<T: Parameter>(parameters: Array<T>, name: strin
...
 
 export function filterHeaders<T: Header>(headers: Array<T>, name: string): Array<T> {
-  if (!Array.isArray(headers) || !name) {
+  if (!Array.isArray(headers) || !(typeof name === 'string')) {
     return [];
   }
 
   return headers.filter(h => {
-    if (!h || !h.name) {
+    if (!h || !(typeof h.name === 'string')) {
```
develohpanda: Should the `!h.name` check still exist? That would explicitly exclude an empty string (considering the previous behavior looks like it excludes empty strings).

Also, TIL you can effectively filter and map at the same time. This implementation is cool although not immediately obvious (and it would also implicitly filter out a header if `h.name` is an empty string)

packages/insomnia-app/app/common/__tests__/misc.test.js
```diff
@@ -43,6 +43,8 @@ describe('filterHeaders()', () => {
     expect(misc.filterHeaders(['bad'], null)).toEqual([]);
     expect(misc.filterHeaders(['bad'], 'good')).toEqual([]);
     expect(misc.filterHeaders(null, 'good')).toEqual([]);
+    expect(misc.filterHeaders([{ name: 123, value: 123 }], 123)).toEqual([]);
+    expect(misc.filterHeaders([{ name: 'good', value: 'valid' }], 123)).toEqual([]);
```
develohpanda, line 48: Yay more tests

packages/insomnia-app/app/common/misc.js
```diff
@@ -29,12 +29,12 @@ export function filterParameters<T: Parameter>(parameters: Array<T>, name: strin
 }
 
 export function filterHeaders<T: Header>(headers: Array<T>, name: string): Array<T> {
-  if (!Array.isArray(headers) || !name) {
+  if (!Array.isArray(headers) || !(typeof name === 'string')) {
```
develohpanda: Should the `!name` check also still exist? 🤔 

image

develohpanda (changes requested):
Nice! 

Small note about empty string checks that have disappeared

packages/insomnia-app/app/common/misc.js
gschier: Hmm, ya I think so. I _did_ think about it a bit while changing it but I think I agree now that it should be kept. A header with no name is essentially a non-existing header.

develohpanda (approved):
LGTM! 😄

PR #2106, opened 2020-05-01
<Button> small variant and default color (gray)

gschier:
Was trying to use the `<Button>` component in part of the app and needed some additional styles.

https://deploy-preview-2106--insomnia-storybook.netlify.app/?path=/story/button--reference

image


develohpanda (approved):
LGTM



A lot of rules in that styled-component CSS - I wonder if it might benefit from some unit tests. Maybe even just snapshots? 🤔

gschier:
Ya, perhaps some visual diff tests using something like https://percy.io/

develohpanda:
> 
> 
> Ya, perhaps some visual diff tests using something like [percy.io](https://percy.io/)

Woah, this is cool! I was thinking Jest snapshots but this would be much nicer.

PR #2114, opened 2020-05-04
Bug in restoring code-folds

develohpanda:
When this section of code was changed from `marks && marks.forEach(...)`, it was converted to a `for...in` loop instead of `for...of`.

Currently, if you fold any code, navigate away, and come back, it causes a fatal failure. This has not been released into Core yet, however **Designer does have this bug**.

fatal

Note: unnecessary changes often get added in from linting. I changed `lint:fix` command in the pre-commit hook back to `lint`, because after fixing it does not stage the changes for the commit. Ideally, it _should_ stage those items. It would be cool to combine prettier, eslint, husky, and lint-staged, so that _all_ methods of fixing code use the exact same specification. Will do when we have downtime. 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "lint": "eslint \"**/*.js\" \"**/*.html\" \"**/*.json\"",
-    "lint:fix": "eslint \"**/*.js\" \"**/*.html\" \"**/*.json\" --fix",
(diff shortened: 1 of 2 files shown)


gschier (approved):
Oh, nice catch! I must not have tested that when making the refactor originally 🤦

PR #2115, opened 2020-05-04
Move toggle-switch to `insomnia-components`, attach to themes

develohpanda:
Move toggle-switch to `insomnia-components` and bind it to the theme system, so that the active color is `--color-surprise` instead of the default green.

image
https://deploy-preview-2115--insomnia-storybook.netlify.app/?path=/story/toggle-switch--checked


sonicyeti (approved):
Good stuff, glad this little wonder is an official component and provisioned for theming!

Woo hoo

sonicyeti (approved):
Very cool

PR #2125, opened 2020-05-06
Storybook Sidebar Provisioning & WIP Implementation

sonicyeti:
This PR encompasses the following:

1. Implementing the latest round of SVGs per Figma design specs
2. Rolling in updated & expanded base CSS vars in both core & storybook (preview head)
3. Adding Storybook > Figma design preview add-on (seen in screenshot)
4. Adding Framer animation package (see animation below)
5. Migrating **WIP** sidebar from Studio (see in screenshot)

Screen Shot 2020-05-06 at 8 28 17 AM


gschier:
Some interaction feedback:

When playing the Sidebar component, I kept clicking the + icon to try and expand the sections. I never even thought to click on the rest of the item. 

I think these might help:

- Add a hover state to each item
- Add `cursor: default` since it's currently showing the caret
- Make the whole thing clickable instead of just the label

Thoughts @sonicyeti @jitachi?

sonicyeti:
> Some interaction feedback:
> 
> When playing the Sidebar component, I kept clicking the + icon to try and expand the sections. I never even thought to click on the rest of the item.
> 
> I think these might help:
> 
> * Add a hover state to each item
> * Add `cursor: default` since it's currently showing the caret
> * Make the whole thing clickable instead of just the label
> 
> Thoughts @sonicyeti @jitachi?

The plus icon threw me for a loop as well. I would never expect a plus icon to do anything other than expand an accordion style piece of UI but low-and-behold Jitachi showed me how Figma's UI worked and its pretty slick. Same concept (check out the export panel). Title is hot for expand/contract while the plus sign adds an export preference line. This pattern can also be seen in VS Code but I forgot where.

packages/insomnia-app/app/ui/css/constants/dimensions.less
```diff
@@ -1,15 +1,38 @@
...
+
+  /* Spacing */
+  --spacing-xs: calc(var(--font-size) * 0.40);
+  --spacing-sm: calc(var(--font-size) * 0.60);
+  --spacing-md: calc(var(--font-size) * 1);
+  --spacing-lg: calc(var(--font-size) * 1.25);
+  --spacing-xl: calc(var(--font-size) * 1.5);
+  --spacing-xxl: calc(var(--font-size) * 2);
+  --spacing-xxxl: calc(var(--font-size) * 3);
+  --spacing-xxxxl: calc(var(--font-size) * 4);
```
gschier: We talked about this briefly in the past, but I think the new `spacing` should entirely replace `padding` vars since they will be used for the same purpose (padding and margin).

I think we have to options:

1. Replace and update padding vars with these values, and continue using `padding` as the name
2. Delete the `padding` vars and update all usages to reference the new `spacing` ones

For the purpose of keeping this PR _small_, my vote would be to No. 1

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,404 @@
...
+  h6 {
+    font-size: var(--font-size-xs);
+  }
+  h5 {
+    font-size: var(--font-size-sm);
+  }
+  /* END To constants */
+  width: 260px;
+  height: 100%;
+  min-height: 1100px;
```
gschier: Hmm, these styles seem like they should be controlled by the place the component is being consumed (App or Story), not by the component itself.

For example, `min-height: 1100px` will break if we render this in the app with a window shorter than `1100px`

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,404 @@
...
+
+  const [visible3, setVisible3] = useState(true);
+  const toggleVisible3 = () => setVisible3(!visible3);
+
+  const [visible4, setVisible4] = useState(true);
+  const toggleVisible4 = () => setVisible4(!visible4);
+
+  return (
+    <StyledSidebar>
+      {/*
```
gschier: Delete comment

gschier (changes requested):
A few comments! Looking pretty good though

packages/insomnia-components/.storybook/preview-head.html
```diff
@@ -179,16 +179,44 @@
...
+    --spacing-lg: calc(var(--font-size) * 1.25);
+    --spacing-xl: calc(var(--font-size) * 1.5);
+    --spacing-xxl: calc(var(--font-size) * 2);
+    --spacing-xxxl: calc(var(--font-size) * 3);
+    --spacing-xxxxl: calc(var(--font-size) * 4);
+
+    /* Colors */
+    --color-border: #d9d9d9;
+    --color-sidebar: #fafafa;
+    --color-text: #404040;
```
gschier: For now, can we use theme colors for these instead, until we update the theme system to support Jitachi's new variables?

With the current system, this would be the values:

- border: `var(--hl-md)`
- sidebar: `var(--color-bg)`
- text: `var(--color-font)`

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,404 @@
...
+  const [visible2, setVisible2] = useState(false);
+
+  const [visible3, setVisible3] = useState(true);
+  const toggleVisible3 = () => setVisible3(!visible3);
+
+  const [visible4, setVisible4] = useState(true);
+  const toggleVisible4 = () => setVisible4(!visible4);
+
+  return (
+    <StyledSidebar>
```
gschier: This should have `className="theme--sidebar"` so that `--color-bg` uses the sidebar scope from the theme.

packages/insomnia-app/app/ui/css/constants/dimensions.less
sonicyeti: Let me break this PR up. I agree 100% with your first point and don't feel like we should cut corners for the sake of the PR being too big. Given the actual sidebar in this PR is only located in SB, I'll remove these changes from "dimensions.less" all together since they aren't directly needed. I was trying to provision for the next step, in which case they will be needed but lets address this now before moving forward.

sonicyeti: Ok reverted dimension back.

packages/insomnia-components/components/sidebar.js
sonicyeti: Yep, for sure not driven by a fixed value like this. I literally just needed a way for forcibly increase the height given SB has no parent element with a fixed height. I'll work around this though now before re-review!

packages/insomnia-components/.storybook/preview-head.html
sonicyeti: Sounds good.

packages/insomnia-components/.storybook/preview-head.html
```diff
@@ -213,9 +214,9 @@
     --spacing-xxxxl: calc(var(--font-size) * 4);
 
     /* Colors */
-    --color-border: #d9d9d9;
-    --color-sidebar: #fafafa;
-    --color-text: #404040;
+    --color-border: var(--hl-md);
+    --color-sidebar: var(--color-bg);
+    --color-text: var(--color-font);
```
gschier, line 219: Oh, woops. I meant to use these vars directly in the component CSS. If we merge this, the component won't work in the app, since `--color-border` etc won't be defined there yet.

packages/insomnia-components/components/sidebar.js
```diff
@@ -67,6 +72,10 @@ const StyledSection: React.ComponentType<{}> = styled(motion.ul)`
     justify-content: space-between;
     align-items: center;
 
+    &:hover {
+    background-color: var(--hl-xs);
```
gschier: Indent?

packages/insomnia-components/.storybook/preview-head.html
gschier: Also just realized that the other news vars in this file also won't be present in the app yet.

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,380 @@
...
+import Tooltip from './tooltip';
+import SvgIcon, { IconEnum } from './svg-icon';
+
+type Props = {|
+  className?: string,
+|};
+
+const StyledSidebar: React.ComponentType<{}> = styled.div`
+  /* To constants */
+  background-color: #fafafa;
```
gschier: Should be `var(--color-bg)`, which will pull the bg color from the theme `sidebar: {}` scope.

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,380 @@
...
+  border: 1px solid var(--color-border);
+  color: var(--color-text);
+  position: relative;
+  svg {
+    font-size: var(--font-size-xl);
+    fill: #737373;
```
gschier: Should use a theme color here. `--hl-md` perhaps?

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,380 @@
...
+const StyledFilter: React.ComponentType<{}> = styled(motion.div)`
+text-align: left;
+padding: 0;
+width:100%;
+display:block;
+overflow:hidden;
+input {
+  box-sizing:border-box;
+  font-size: var(--font-size-md);
+  min-width: 88%;
```
gschier: I have a feeling this relative width might cause a problem when resizing the sidebar?

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,380 @@
...
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(var(--spacing-xxl), 1fr));
+  column-gap: var(--spacing-sm);
+  grid-template-rows: 1fr;
+  align-items: start;
+  white-space: nowrap;
+  font-size: var(--font-size-md);
+  line-height: var(--font-line-height-md);
+
+  .inline-icon {
```
gschier: Not used anywhere

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,380 @@
...
+              animate={{ opacity: visible2 ? 0.6 : 0 }}
+              transition={{ duration: 0.2, ease: 'easeInOut', delay: 0.2 }}>
+              <SvgIcon icon={IconEnum.trashcan} />
+            </motion.span>
+            <motion.span
+              onClick={toggleVisible3}
+              initial={{ opacity: visible2 ? 0.6 : 0 }}
+              animate={{ opacity: visible2 ? 0.6 : 0 }}
+              transition={{ duration: 0.2, ease: 'easeInOut', delay: 0.15 }}>
+              <SvgIcon icon={IconEnum.search} />
```
gschier, line 178: Hmm, the search seems to have a BG color for some reason
 
image

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,380 @@
...
+              transition={{ duration: 0.2, ease: 'easeInOut', delay: 0.1 }}>
+              <Tooltip message="Add" position="top">
+                  <SvgIcon icon={IconEnum.plus} />
+              </Tooltip>
+            </motion.span>
+          </div>
+        </li>
+
+        <StyledFilter
+            initial={{ height: visible3 ? '0px' : '45px' }}
```
gschier: This hardcoded px value makes me uneasy. Is there a way we can do this that doesn't involve fixed sizing? Perhaps by animating `transform: scaleY(0);` to `transform: scaleY(1);` and defining the height in CSS using a line-height var. Making this change will also prove useful once we have an <Input> component that controls its own height.

gschier (changes requested):
Second pass complete 🥳

packages/insomnia-components/components/sidebar.js
sonicyeti: Pulled it out for now, will refine once we get it in. Renders too skinny for and would rather fix once I get it in the actual environment.

packages/insomnia-components/components/sidebar.js
```diff
@@ -0,0 +1,351 @@
...
+  /* END To constants */
+  width: 260px;
+  height: 100%;
+`;
+
+const StyledSection: React.ComponentType<{}> = styled(motion.ul)`
+  overflow: hidden;
+  box-sizing: border-box;
+  border-bottom: 1px solid var(--hl-md);
+  }
```
gschier: Random bracket leftover

packages/insomnia-app/app/ui/css/constants/dimensions.less
```diff
@@ -18,6 +18,29 @@ html {
...
+
+  /* Font Weight */
+  --font-weight-regular: 400;
+  --font-weight-medium: 600;
+  --font-weight-bold: 700;
```
gschier: For the purposes of this PR, can we remove these new variables for now and stick with the currently-existing ones? 

ie. use `--padding-*` instead of spacing, `calc(font-size-*, x)` instead of line height, and hardcode font-weight. 

I think it makes sense to do a separate sweep to update all the variables in both the app and components library at once.

packages/insomnia-components/components/tooltip.js
```diff
@@ -0,0 +1,236 @@
...
+import ReactDOM from 'react-dom';
+import styled from 'styled-components';
+
+type Props = {
+  children: React.Node,
+  message: React.Node,
+  position?: 'bottom' | 'top' | 'right' | 'left',
+
+  // Optional
+  className?: string,
```
gschier, line 14: We talked about passing the color in as a prop like we do with <Button> here. What made you decide against that?

gschier (changes requested):
A few comments but we're really close now!

packages/insomnia-components/components/sidebar.js
sonicyeti: Thanks for the catch!

packages/insomnia-components/components/tooltip.js
gschier: Woops! Somehow I thought this was the <Notice> component.

gschier (approved):
We made it! 🥳

PR #2131, opened 2020-05-07
Wrap SwaggerUI component with an error boundary that can invalidate

### LINKED ISSUE #2094: Crashes when typing openAPI spec
**Describe the bug**
Program crashes

**To Reproduce**
1. Create new blank document, name it whatever
2. Type the following into the spec:
```json
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
```

3. Enter the following line after "servers:":
```
  - urls:
```
4. The program should crash

**Expected behavior**
The program does not crash

**Screenshots**
After step 2: https://allthepics.net/image/t1jsE
After step 3: https://allthepics.net/image/t1yeT

 - OS: Microsoft Windows 10 Home, Version	10.0.18362 Build 18362
 - App Version:
Version 2020.1.3
Shell 3.1.13
Node 10.2.0
V8 6.6.346.32
Architechture x64

libcurl/7.57.0-DEV OpenSSL/1.0.2o zlib/1.2.11 libssh2/1.7.0_DEV

**Additional**
When it crashes it says to email support, i did, but they responded to open an issue here. Would probably be better to have the app just link to opening an issue instead. Also, I should be able to copy the Insomnia Designer Help -> About info instead of manually typing all that out

develohpanda:
Current:
- Currently, the error boundary shows one error and doesn't re-attempt a render until the page is navigated to again, for instance. 
- Swagger UI can be inconsistent with its errors, for instance #2094. Because this is not a component controlled by us, there may be other unknown bugs that exist as well.

This PR introduces an `ErrorBoundary` around the SwaggerUI component, and it also allows an `ErrorBoundary` to be invalidated.

Therefore, an error is caught by the `ErrorBoundary` around `SwaggerUI`, but it is retried once a user changes the specification because the `ErrorBoundary` is invalidated by the edit.

Invalidation is controlled via a changing prop, which triggers a component update, and a life-cycle hook that decides whether the error should be cleared and a re-render attempted.

Demo:

Closes #2094


packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -235,7 +247,29 @@ class WrapperDesign extends React.PureComponent<Props, State> {
...
+                    'put',
+                    'post',
+                    'delete',
+                    'options',
+                    'head',
+                    'patch',
+                    'trace',
+                  ]}
+                />
+              </ErrorBoundary>
```
develohpanda: This is the primary change in this file... the rest are from linting 😢

packages/insomnia-app/app/ui/components/error-boundary.js
```diff
@@ -24,6 +27,17 @@ class SingleErrorBoundary extends React.PureComponent<Props, State> {
...
 
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps: $ReadOnly<Props>) {
+    const clearState = { error: null, info: null };
+
+    if (nextProps.invalidateWith === undefined || this.state === clearState) {
+      return;
+    }
+
+    this.setState(clearState);
```
develohpanda: There will not be a double render when `setState` is called from `componentWillRecieveProps` 😄

packages/insomnia-app/app/ui/components/error-boundary.js
```diff
@@ -7,6 +7,9 @@ type Props = {
   children: React.Node,
   errorClassName?: string,
   showAlert?: boolean,
+
+  // Avoid using invalidation with showAlert, otherwise an alert will be shown with every attempted re-render
+  invalidateWith?: object,
```
develohpanda: The value of this prop isn't used directly by the error boundary, but it triggers a component update when bound to a text field (such as the spec object in wrapper-design)

packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -235,7 +247,29 @@ class WrapperDesign extends React.PureComponent<Props, State> {
...
+                invalidateWith={swaggerUiSpec}
+                replaceWith={
+                  <div className="text-center margin">
+                    <h3>An error occurred while trying to render Swagger UI 😢</h3>
+                    This preview will automatically refresh, once you have a valid specification
+                    that can be previewed. If it doesn't, try re-opening the document.
+                  </div>
+                }>
+                <SwaggerUI
+                  spec={swaggerUiSpec || {}}
```
develohpanda: Is parsing the apiSpec necessary? We _can_ send the original contents (`activeApiSpec.contents`) directly to SwaggerUI, and it handles the error case fixed by this PR correctly, however we have no control over the error it shows.

With the following code change, SwaggerUI shows this in the preview
```suggestion
                  spec={activeApiSpec.contents}
```
image

packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -235,7 +247,29 @@ class WrapperDesign extends React.PureComponent<Props, State> {
...
             <div id="swagger-ui-wrapper">
-              <SwaggerUI spec={swaggerUiSpec || {}} supportedSubmitMethods={['get', 'put', 'post', 'delete', 'options', 'head', 'patch', 'trace']} />
+              <ErrorBoundary
+                invalidateWith={swaggerUiSpec}
+                replaceWith={
+                  <div className="text-center margin">
+                    <h3>An error occurred while trying to render Swagger UI 😢</h3>
+                    This preview will automatically refresh, once you have a valid specification
+                    that can be previewed. If it doesn't, try re-opening the document.
+                  </div>
```
develohpanda: @nijikokun how's this for an error message to show in place of the preview when SwaggerUI throws a fatal error?

packages/insomnia-app/app/ui/components/error-boundary.js
gschier: Can we force this to be a string? Object comparison in JS isn't great, as it matches by reference, not by shape.

packages/insomnia-app/app/ui/components/error-boundary.js
```diff
@@ -24,6 +27,17 @@ class SingleErrorBoundary extends React.PureComponent<Props, State> {
     };
   }
 
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps: $ReadOnly<Props>) {
+    const clearState = { error: null, info: null };
```
gschier: This object comparison with `this.state` seems like it could be brittle, as it won't compare the properties themselves. We should probably be checking both `error` and `info` explicitly to ensure our code is robust.

```
> { error: null, info: null } === { error: null, info: null }
false
```

packages/insomnia-app/app/ui/components/wrapper-design.js
gschier: What do you think about removing the "If it doesn't..." sentence?

gschier (changes requested):
A few comments but nothing too major! Will be nice to get this in 😄

packages/insomnia-app/app/ui/components/wrapper-design.js
gschier: If it works, I'm fine with passing it straight through

packages/insomnia-app/app/ui/components/settings/plugins.js
```diff
@@ -293,7 +293,7 @@ class Plugins extends React.PureComponent<Props, State> {
 
         <hr />
 
-        <div className="text-right">
+        <div key={} className="text-right">
```
gschier: 🤔

packages/insomnia-app/app/ui/components/error-boundary.js
```diff
@@ -24,6 +27,17 @@ class SingleErrorBoundary extends React.PureComponent<Props, State> {
...
 
+  // eslint-disable-next-line camelcase
+  UNSAFE_componentWillReceiveProps(nextProps: $ReadOnly<Props>) {
+    const { error, info } = this.state;
+
+    if (nextProps.invalidateWith === undefined || (error === null && info === null)) {
+      return;
+    }
+
+    this.setState({ error: null, info: null });
```
gschier: I feel like we should be checking if the invalidation key changed here. I know your current solution works, but it's hard to follow what's happening without adding a ton of comments. I suggest something a bit more descriptive and explicit.

```suggestion
    const invalidationKeyChanged = nextProps.invalidateWith !== this.props.invalidateWith;
    const isErrored = error !== null || info !== null;
    const shouldResetError = invalidationKeyChanged && isErrored;
    
    if (shouldResetError) {
        this.setState({ error: null, info: null });
    }
```

packages/insomnia-app/app/ui/components/settings/plugins.js
develohpanda: 🙈

packages/insomnia-app/app/ui/components/error-boundary.js
```diff
@@ -57,10 +73,14 @@ class SingleErrorBoundary extends React.PureComponent<Props, State> {
...
-    const { errorClassName, children } = this.props;
+    const { errorClassName, children, renderError } = this.props;
 
     if (error && info) {
-      return <div className={errorClassName || null}>Render Failure: {error.message}</div>;
+      return (
+        renderError?.(error) || (
+          <div className={errorClassName || null}>Render Failure: {error.message}</div>
+        )
+      );
```
gschier: Optional chaining is nice in many cases but this one seems like adds more complexity than it saves. This code will also render the fallback if `renderError(error)` returns a falsey value, which seems like confusing behavior. 

```suggestion
      return (
        renderError ? renderError(error) : (
          <div className={errorClassName || null}>Render Failure: {error.message}</div>
        )
      );
```

gschier (changes requested):
Almost there! Just one comment on `renderError` behavior.

sonicyeti (approved):
Nice!

gschier (approved):
Woo!

PR #2132, opened 2020-05-07
Prioritize plugins correctly if the same type of plugin exists at different levels in OAS

### LINKED ISSUE #2100: Apply priority order if the same kong plugin exists in different sections of OAS
Kong plugins `x-kong-plugin-*` can exist at the OpenAPI object, server, path, and operation levels. These must collate together, however if the _same_ plugin type (eg. `x-kong-plugin-key-auth`) exists at multiple levels, then the highest priority should be selected, and the others ignored.

The priority order is **Operation > Path > Server > Global** (OpenAPI spec object).

For example, with the following specification snippet

```yaml
openapi: 3.0.0
servers:
  - url: http://api.insomnia.rest
    x-kong-plugin-custom-one:
      ...
    x-kong-plugin-custom-two:
      name: 'on server'
      ...
paths:
  '/path':
    GET:
      x-kong-plugin-custom-two:
        name: 'on operation'
        ...
      x-kong-plugin-custom-three:
        ...     
```

The resultant plugins that are applied to the `GET` operation on `/path` for the server `http://api.insomnia.rest`, should be:
- `custom-one`
- `custom-two` (name `on operation`)
- `custom-three`

The current (incorrect) behavior applies:
- `custom-one`
- `custom-two` (name `on server`)
- `custom-two` (name `on operation`)
- `custom-three`

develohpanda:
This PR prioritizes which plugin instance to apply to an Ingress document when combining global, server, path and operation plugins. The rules are described in #2100

Closes #2100 😄

### DIFF
--- a/packages/openapi-2-kong/src/kubernetes/plugins.js
+++ b/packages/openapi-2-kong/src/kubernetes/plugins.js
@@ -197,3 +197,32 @@ export function normalizeOperationPlugins(operationPlugins: OperationPlugins): O
   const pluginsExist = operationPlugins.some(o => o.plugins.length);
   return pluginsExist ? operationPlugins : [blankOperation];
 }
+
+export function prioritizePlugins(
+  global: Array<KubernetesPluginConfig>,
+  server: Array<KubernetesPluginConfig>,
+  path: Array<KubernetesPluginConfig>,
+  operation: Array<KubernetesPluginConfig>,
+): Array<KubernetesPluginConfig> {
+  // Order in priority: operation > path > server > global
+  const plugins: Array<KubernetesPluginConfig> = [...operation, ...path, ...server, ...global];
+
+  // Select first of each type of plugin
+  return distinctByProperty(plugins, p => p.plugin);
+}
+
+export function distinctByProperty<T>(arr: Array<T>, propertySelector: (item: T) => any): Array<T> {
+  const result: Array<T> = [];
+  const map = new Map();
+
+  for (const item of arr.filter(i => i)) {
+    const selector = propertySelector(item);
+    if (map.has(selector)) {
+      continue;
+    }
+
+    map.set(selector, true);
+    result.push(item);
+  }
+  return result;
+}
--- a/packages/openapi-2-kong/src/kubernetes/plugins.js
+++ b/packages/openapi-2-kong/src/kubernetes/plugins.js
@@ -213,15 +213,15 @@ export function prioritizePlugins(
 
 export function distinctByProperty<T>(arr: Array<T>, propertySelector: (item: T) => any): Array<T> {
   const result: Array<T> = [];
-  const map = new Map();
+  const set = new Set();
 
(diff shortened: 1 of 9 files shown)


packages/openapi-2-kong/types/openapi3.flow.js
```diff
@@ -206,3 +206,16 @@ declare type OpenApi3Spec = {
...
+const HttpMethod = {
+  get: 'GET',
+  put: 'PUT',
+  post: 'POST',
+  delete: 'DELETE',
+  options: 'OPTIONS',
+  head: 'HEAD',
+  patch: 'PATCH',
+  trace: 'TRACE',
+};
```
develohpanda, line 219: Because this is a `.flow.js` file in the `types` folder (I'm guessing), this can't be _exported_ otherwise flow breaks in other areas.

gschier: Perhaps because there is a type with the same name in `src/common.js`?

packages/openapi-2-kong/src/kubernetes/plugins.js
```diff
@@ -197,3 +197,32 @@ export function normalizeOperationPlugins(operationPlugins: OperationPlugins): O
...
+  return distinctByProperty(plugins, p => p.plugin);
+}
+
+export function distinctByProperty<T>(arr: Array<T>, propertySelector: (item: T) => any): Array<T> {
+  const result: Array<T> = [];
+  const map = new Map();
+
+  for (const item of arr.filter(i => i)) {
+    const selector = propertySelector(item);
+    if (map.has(selector)) {
```
gschier: FYI, you can use a `Set` here instead, then you don't have to add the `boolean` value.

gschier (approved):
Looks good! Just a couple minor comments.

packages/openapi-2-kong/src/kubernetes/plugins.js
develohpanda: I just thought of another place that can use a `Set`

`Set` is my friend now

packages/openapi-2-kong/types/openapi3.flow.js
develohpanda: If I _import_ that _Object_ into this `flow.js` file, usages of every type defined in this file show a warning saying it is not exported. It seems you can't import or export an object from a flow type definition file (or I wasn't able to figure out how to do it).

gschier: Can't you `declare` it here and delete the other definition, though?

develohpanda: `HttpMethod` is a `const` and not `type`, so it can't be declared?

I did remove the derived type from `src/common.js` so there is only one definition for `HttpMethodType`. I wasn't able to figure out a nice way to reuse the `HttpMethod` object, while still allowing flow types to work implicitly (no need to import in each file)

image

PR #2134, opened 2020-05-07
Improve card filtering

gschier:
With the `ApiSpec` filename refactor #2063 we forgot to update the filter logic to use it, so it stopped working.

This PR fixes the filtering, but also adds the other things on the card as filterable stings (label, title, version, branch)

image

image


develohpanda (approved):
🔥 this is super cool

Issue #1847, opened 2019-12-14
[Build] Update Electron


gschier:
The version of Electron we're on is now quite out of date. We should update to the latest.

The OAuth 2.0 dialog is already starting to fail due to the outdated user-agent of our app's Chromium build https://github.com/Kong/insomnia/issues/1774

## Issues Discovered

- [x] compiling font manager breaks
  - [x] Try upgrading [`font-scanner`](https://github.com/axosoft/font-scanner) 
  - [ ] ~~Switch to a JS-only implementation like [`get-system-fonts`](https://github.com/princjef/get-system-fonts)~~
- [ ] compiling [`node-libcurl`](https://github.com/JCMais/node-libcurl) breaks
  - [ ] Try updating `node-libcurl`
- [ ] compiling `fsevents` fails

semoal:
Actually for me, with lts node version, i cant even start the project in my local machine.

I can look at this if you want :+1:

gschier:
Going to call this issue done as a first try. Will move progress to the newer https://github.com/Kong/insomnia/issues/2369

PR #1528, opened 2019-05-29
Switch from insomnia-node-libcurl fork back to node-libcurl

gschier:

Issue #997, opened 2018-06-22
[Feature Request] Support HTTP/2 requests


aldaris:
- Insomnia Version: 5.16.6 (5.16.6.2073)
- Operating System: OSX

## Details

Currently when Insomnia makes its requests to an HTTPS enabled site, it only uses HTTP/1.1. It would be great if Insomnia would utilize HTTP/2 requests via ALPN.

gschier:
This _should_ be easy to add, except the `node-libcurl` project doesn't look like it is built with HTTP/2 compatibility when targeting Windows. See https://github.com/peters/curl-for-windows

I still think this will be good to have eventually 👍

WingGithub:
I hope someone is working on this. http2 adoption is getting pretty high and this is a blocking issue for us. I have to disable http2 in our development server and I can't recommend this tool to devOps if it won't work on production.

gschier:
Just a note: Support for HTTP/2 is already included in the OSX and Linux versions of Insomnia, since they are linked with the `nghttp2` library. 

I think the only thing that needs to be done is to allow the user to supply the [CURLOPT_HTTP_VERSION](https://curl.haxx.se/libcurl/c/CURLOPT_HTTP_VERSION.html) option

FallingSnow:
@gschier HTTP/2 does not appear to be working for me.

```sh
$ uname -a
Linux laptop 5.0.5-arch1-1-ARCH #1 SMP PREEMPT Wed Mar 27 17:53:10 UTC 2019 x86_64 GNU/Linux
```

When using the firefox extension [RESTClient](https://addons.mozilla.org/en-US/firefox/addon/restclient/) it works as expected.
![Screenshot from 2019-04-01 16-08-33](https://user-images.githubusercontent.com/2512197/55365194-6f43fe80-5498-11e9-875f-bda875d13b0e.png)

However Insomnia is another story. It reacts the same way as Postman.

![Screenshot from 2019-04-01 16-08-18](https://user-images.githubusercontent.com/2512197/55365195-70752b80-5498-11e9-96de-78ffc8e7a07c.png)

![Screenshot from 2019-04-01 16-10-12](https://user-images.githubusercontent.com/2512197/55365243-a9ad9b80-5498-11e9-801a-3c4a9dd0c6ee.png)

WingGithub:
[...]

PR #2150, opened 2020-05-10
deps: bump node-libcurl to latest prerelease version

JCMais:
This is related to the following issue:
https://github.com/Kong/insomnia/issues/997

And should unblock this:
https://github.com/Kong/insomnia/issues/1847

Some code is based on the previous PR: https://github.com/Kong/insomnia/pull/1528

Some notes:

## Prebuilt binary
`node-libcurl` has been providing prebuilt binaries for Electron, however the current non-prerelease version, `2.1.1`, does not provide binaries for Electron `v3`. I've added it back starting with `2.1.2-1`.

Right now Insomnia is not using the prebuilt binaries, because the project uses `electron-rebuild` and it does not support `node-pre-gyp`, I've opened an issue there about adding support: https://github.com/electron/electron-rebuild/issues/367

If support is not added the prebuilt binaries can still be used by changing the `bootstrap` script on the root `package.json` to the following:
```
npm install && lerna bootstrap -- --runtime=electron --target=v3.1.13 --disturl=https://www.electronjs.org/headers && lerna run --parallel --stream bootstrap
```
And removing the `electron-rebuild` call from the `insomnia-app` `bootstrap` script.

This would simplify Insomnia build as there would be no need to build the native dependencies from source anymore. For the scripts used to build libcurl and their dependencies for `node-libcurl` itself, see this directory: https://github.com/JCMais/node-libcurl/tree/develop/scripts/ci

If Insomnia still needs to have control over the libcurl build, those environment variables:
https://github.com/Kong/insomnia/blob/8162772dae9ec11333783d9be5c9bd9caf50efd2/.github/workflows/release-core.yml#L49
https://github.com/Kong/insomnia/blob/8162772dae9ec11333783d9be5c9bd9caf50efd2/.github/workflows/release-designer.yml#L49

Must be changed to `npm_config_curl_static_build: true`, which is the environment variable that would cause the following gyp variable to become `true`: https://github.com/JCMais/node-libcurl/blob/9aa421d2f39306a6e42afe84afe528164edd44c3/binding.gyp#L11

## `Curl.getVersionInfoString`

`Curl` class now has a new static method `getVersionInfoString`, which returns more information than `getVersion`, it should be almost identical to the output of `curl -V`:  

Just to not introduce that many changes, I've kept the current implementation of the wrapper to use `getVersion`.

## Test `actuallySend() › sends a generic request`

This test was missing the `HTTPAUTH` option before, don't know why.

## Uncommited files not ignored
After running the project those files were generated:
```
packages/openapi-2-kong/dist/index.js
packages/openapi-2-kong/dist/index.js.map
```

And they are not on `.gitignore` list, should they be added?


develohpanda:
> And they are not on `.gitignore` list, should they be added?

I have added them into `.gitignore` now 😄 #2165

gschier:
Thanks for kicking off this effort @JCMais! I added a bunch of stuff to this PR and merged it in #2223 

Having prebuilt versions for Electron is a life saver 👍

JCMais:
@gschier glad it worked.

I will release a new version of `node-libcurl` sometime later this week with the changes made on the current pre-release version and a few other improvements.

PR #2151, opened 2020-05-11
Honor variable defaults from servers declaration during OpenAPI3 import

### LINKED ISSUE #1640: [Bug] OpenAPI import does not seem to honor default variables in servers declaration
- Insomnia Version: 6.6.2
- Operating System: MacOS mojave 10.14.6

## Details

I'm trying to import an OpenAPI 3 spec that uses variables in the server url. Default values are provided for these in the variables declaration. This results in an error when i try to import the file: Invalid URL: http://{host}:{port}/{basePath}

If I replace the url with the following string it works: http://localhost:8080/filemanagement

OpenAPI yaml file (parts of it - but it can be used to reproduce the bug and import works after replacing the servers/url part):

```
openapi: 3.0.2
info:
  description: OpenAPI definition for filemanagment services
  license:
    name: todo
    url: todo
  title: filemanagement
  version: 0.1.0
servers:
  - description: Devserver
    url: http://{host}:{port}/{basePath}
    variables:
      host:
        default: localhost
        description: Various servers (production, local development etc).
      port:
        default: "8080"
        description: Port to reach server on.
        enum:
          - "8080"
          - "9090"
      basePath:
        default: filemanagement
        description: API version, for information on other supported versions see those
          specifications.
tags:
  - description: Operations for managing file storages such as S3.
    name: storages
  - description: Operations for managing files connected to data sources.
    name: files

paths:
  /files:
    get:
      operationId: find_files
      parameters:
        - description: Search for files based on metadata.
          explode: true
          in: query
          name: metadata
          required: false
          schema:
            type: string
          style: form
      responses:
        200:
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/FileDescription'
          description: List of files matching the search criteria. List will be empty
            if no files match the criteria.
      tags:
        - files
      x-openapi-router-controller: filemanagement_service.controllers.files_controller
```

develohpanda:
This PR adds functionality to honor variable defaults from the servers declaration during an OpenAPI3 spec import.

The `openapi-2-kong` package already does this, the functionality is simply adapted for `insomnia-importers`.

The openapi3 importer honors variable defaults, but if a variable exists in the URL and is missing in `server.variables`, **an error is thrown and presented to the user**.

<details>
<summary>Example spec</summary>

```yaml
openapi: 3.0.2
info:
  description: This API is for administrating files and file storages.
  license:
    name: Some license
  title: File and Storage API
  version: 0.1.0
servers:
  - description: Devserver
    url: '{protocol}://{host}:{port}/{basePath}'
    variables:
      port:
        default: 8080
      host:
        default: localhost
      basePath:
        default: filemanagement
paths:
  /files:
    get:
      operationId: find_files
```
</details>

**Insomnia Core**
image

**Insomnia Designer**
Doesn't do anything during design -> debug navigation, silently swallows any errors caused by parsing an invalid spec. This is existing behavior, which we should improve.

Closes #1640


packages/insomnia-importers/src/importers/openapi3.js
```diff
@@ -97,6 +96,67 @@ module.exports.convert = async function(rawData) {
...
 };
 
+/**
+ * Gets a server to use as the default
+ * Either the first server defined in the specification, or an example if none are specified
+ *
+ * @param {Object} api - openapi3 object
+ * @returns {UrlWithStringQuery} the resolved server URL
+ */
+function getDefaultServerUrl(api) {
```
gschier, line 111: Curious why you used a different approach than `openapi-2-kong` does. Is it so you could replace vars with a default value if they don't have one defined? If so, can we make sure the fixture defined handles the case where there is no default?

https://github.com/Kong/insomnia/blob/d6998ad6843a73dd191f1142a997581c60c12ad0/packages/openapi-2-kong/src/common.js#L121

gschier (changes requested):

packages/insomnia-importers/src/importers/openapi3.js
develohpanda: There are two approaches (there should be a unified one) to fill server variables in `openapi-2-kong`:
1. iterates over `server.variables`, throws if default missing, and replaces items in the url
2. iterates over variables in the url, searches for a default in `server.variables`, defaults to `http` or `.*` if missing

I used a hybrid approach for the openapi3 importer, for several reasons:
- To successfully parse the URL, we need a valid, resolved URL, and smart defaults are much more involved,
- A truly valid OpenAPI spec should have variable defaults defined for a server url
- `o2k` (kubernetes) can handle missing defaults (as per the requirements) using `.*` fallback

The openapi3 importer honors variable defaults, but if a variable exists in the URL and is missing in `server.variables`, an error is thrown and presented to the user.

@gschier

develohpanda: > It would be nice to be able to reuse the resolveVariables implementation (and therefore, tests) between openapi-2-kong and insomnia-importers. How could we do that? Is a separate package needed to achieve this reuse?

Looks like there are three (not two) similar implementations spread between `insomnia-importers` and `openapi-2-kong` - is it possible to have one implementation be used across the packages?

develohpanda: There are two approaches to fill server variables in `openapi-2-kong`:
1. iterates over `server.variables`, throws if default missing, and replaces items in the url
2. iterates over variables in the url, searches for a default in `server.variables`, defaults if missing

I used a hybrid approach for the openapi3 importer, for several reasons:
- To successfully parse the URL, we need a valid, resolved URL, and smart defaults are much more involved,
- A truly valid OpenAPI spec should have variable defaults defined for a server url
- `o2k` (kubernetes) can handle missing defaults (as per the requirements) using `.*` fallback

The openapi3 importer honors variable defaults, but if a variable exists in the URL and is missing in `server.variables`, an error is thrown and presented to the user.

DMarby (approved):

gschier (approved):

PR #2164, opened 2020-05-13
 k4k8s - ingress document name must be unique

### LINKED ISSUE #2163: k4k8s - ingress document name must be unique
If multiple Ingress documents exist, they must have unique names for each run. Otherwise, on applying the configuration, the last ingress document will overwrite all previous ones.

develohpanda:
Closes #2163 

Index the ingress document name to ensure each document has a unique name.


packages/openapi-2-kong/README.md
```diff
@@ -223,29 +222,29 @@ paths:
...
+  name: insomnia-api-1
 spec:
   rules:
     - host: one.insomnia.rest
       http:
         paths:
           - path: /v1/.*
             backend:
-              serviceName: insomnia-api-s0
+              serviceName: service-0
```
gschier: I understand the `metadata.name` change by adding the index suffix, but this one now seems like you could potentially collide and have two services end up with the same name `service-0`. Is there a reason for changing the prefix from the ingress name to the generic `service-*`?

gschier (commented):
One question about the change from `${ingress-name}-${index}` to `service-${index}`

packages/openapi-2-kong/README.md
develohpanda: Ingress document name (and thus resource) cannot be reused, nor can service name. But the service name here is referencing a service, not declaring one.

There is a separate task to also declare a service document as part of the spec generation, where the name for each service should be unique, but the same service can be referred to in multiple ingress documents.

This is my understanding of what the service name is - it's not unique to an ingress document, it's unique to a service.

> Is there a reason for changing the prefix from the ingress name to the generic `service-*`?

Just to simplify the implementation. The service name can be anything, it just needs to be unique _per service_, and there's no reason for that to be coupled to anything.

nijikokun: What about a middle ground since I believe context is king. The document name slugified definitely makes sense as it will show up in UIs so by being present ops can discern differing services and adds to the uniqueness in the scenario where one imports two separate spec generations.

Something like `service-${name}-${index}`

develohpanda: Went ahead with `${specName}-service-${index}` 🙂

gschier (approved):
LGTM!

PR #2165, opened 2020-05-13
Remove built files from source control and update bootstrap steps

develohpanda:
Remove `dist` from `insomnia-components` and `openapi-2-kong` from source control, and update the bootstrap steps for each to build during bootstrap.

After a dry-run, with `.npmignore`:
image

image


packages/insomnia-components/.gitignore
```diff
@@ -1,3 +1,4 @@
 .cache
 storybook-static
 assets/svgr
+dist
```
gschier, line 4: I think we need to also create an `.npmignore` if we want git-ignored files to publish to NPM. http://npm.github.io/publishing-pkgs-docs/publishing/the-npmignore-file.html

Since we really only need to publish a few things, I think something like this could work:

```
# Ignore everything by default
# NOTE: NPM publish will never ignore package.json, package-lock.json, README, LICENSE, CHANGELOG
*

# Don't ignore dist folder
!dist/*
```

gschier (changes requested):
One comment about publishing to NPM

packages/insomnia-components/.gitignore
develohpanda: Looks like we're currently publishing everything, but I agree we should add `.npmignore` and just publish dist 👍 TIL that npm uses `.gitignore` in the absence of `.npmignore` though

image

gschier (approved):

PR #2170, opened 2020-05-14
Handle empty security object in openapi3

### LINKED ISSUE #2167: Openapi3 spec with empty item in security schema fails to import
**Describe the bug**
I can't import a Json OpenApi file.

**To Reproduce**
Hello,

I have an OpenApi Json file that generate an error when I try to import it in Insomnia Designer.
This file can be opened in Swgger Editor and this editor does not report any problem.
Here is the error message I have when I try to import the json file :
`
Import Failed
TypeError:
result.summary[_models_WEBPACK_IMPORTED_MODULE_11__.workspace.type]

Import failed
Cannot read property 'type' of undefined
`

For confidentiality reasons I cannot publish this JSON file here but I can send it to you by email.
**Expected behavior**
The file should be imported.

 - Windows 10 Professional Version 10.0.18363 Number 18363
 - Installation via your website
 - App Version [e.g. v7.0.3]

develohpanda:
Closes #2167, to handle the case where a security object under a path contains an empty item.

Also updates the import to show an error dialog _before_ trying to extract workspaces, if there is a failure during import. This should give more informative error messages.

image

### DIFF
--- a/packages/insomnia-app/app/ui/redux/modules/global.js
+++ b/packages/insomnia-app/app/ui/redux/modules/global.js
@@ -7,6 +7,7 @@ import path from 'path';
 import AskModal from '../../../ui/components/modals/ask-modal';
 import * as moment from 'moment';
 
+import type { ImportResult } from '../../../common/import';
 import * as importUtils from '../../../common/import';
 import AlertModal from '../../components/modals/alert-modal';
 import PaymentNotificationModal from '../../components/modals/payment-notification-modal';
@@ -29,7 +30,6 @@ import { reloadPlugins } from '../../../plugins';
 import { setTheme } from '../../../plugins/misc';
 import { setActivityAttribute } from '../../../common/misc';
 import { isDevelopment } from '../../../common/constants';
-import type { ImportResult } from '../../../common/import';
 import type { Workspace } from '../../../models/workspace';
 
 const LOCALSTORAGE_PREFIX = 'insomnia::meta';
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -277,15 +279,19 @@ export function importFile(workspaceId: string, forceToWorkspace?: ForceToWorksp
...
           const result = await importUtils.importUri(
             askToImportIntoWorkspace(workspaceId, forceToWorkspace),
             uri,
           );
-          importedWorkspaces = [...importedWorkspaces, ...result.summary[models.workspace.type]];
+          const workspaces = handleImportResult(
+            result,
+            'The file does not contain a valid specification.',
+          );
+          importedWorkspaces.push(workspaces);
```
gschier: Shouldn't this be `importedWorkspaces.push(...workspaces)`?

gschier (changes requested):
Looks good! Just a minor comment

packages/insomnia-app/app/ui/redux/modules/global.js
develohpanda: Simplified - `importedWorkspaces` is always initialized to `[]`, so we can simply reassign the variable rather than having to operate on the array. 👍 

But yep, using `.push` would require the spread operator. My bad!

gschier (approved):

PR #2188, opened 2020-05-18
WIP - git sync failures on Windows - change root directory from / to /repo

develohpanda:
Initial approach is to set `/repo` as the base dir. Windows cannot handle the double slash route (eg `//.insomnia/ApiSpec/`) that gets created by using `/` as the base dir.

Further notes in the linked issue.


develohpanda:
Closing in favor of #2194

PR #2197, opened 2020-05-19
Forms component migration

sonicyeti:
Migrating forms to new SB instance at /insomnia-components/ also styling component directly via styled components.

Screen Shot 2020-05-19 at 1 35 49 PM


develohpanda (changes requested):
While this is awesome as a demo it feels out of place 🤔 

In a consumer of this component library, how would I create a form that looks and behaves like the ones shown in the story? There are no components that `insomnia-components` exports to help us create a form like this, since all the logic and styling is in a styled-component within the story.

For example, I would expect components specific to each input type wrapped in appropriate, minimal styles to be exported so creating a form that follows Insomnia styles is trivial. Otherwise just one demo to demonstrate the `<IndeterminateCheckbox />`. Thoughts?

PR #2198, opened 2020-05-19
Notice Component Migration

sonicyeti:
Moving to new SB /insomnia-components/ based instance. Leveraging styled components.

Screen Shot 2020-05-19 at 1 34 52 PM


packages/insomnia-components/components/notice.js
```diff
@@ -0,0 +1,167 @@
...
+    position: absolute;
+    left: 0;
+    right: 0;
+    top: 0;
+    bottom: 0;
+    z-index: -1;
+    opacity: 0.05;
+  }
+
+  &.error {
```
gschier, line 65: We can use inline JS here like we do for <Button>

Something like...

```js
const StyledNotice: React.ComponentType<{color: string}> = styled.p`

  ...

  ${({ color }) => {
    if (color !== 'subtle') {
      return `
        border-color: var(--hl-lg);
        &::before { background-color: var(--hl-xs); }
      `;
    }

    // Colored borders inherit from button text color
    return `
      border-color: var(--color-${color});
      &::before { background-color: var(--color-${color}); }
    `;
  }}}

  ...

`;

// ...

<StyledNotice color={color}>
  // ...
```

This lines up more closely with styled-components practices (avoiding using classes).

packages/insomnia-components/components/notice.js
```diff
@@ -0,0 +1,167 @@
...
+    const { visible, localStorageKey } = this.state;
+
+    if (!visible) {
+      return null;
+    }
+
+    return (
+      <StyledNotice className={classnames('notice', color, className)}>
+        {localStorageKey && (
+          <button className="icon pull-right" onClick={this._dismissNotification}>
```
gschier, line 157: In order to keep all the styles isolated to this component, we should remove the references to `icon` and `pull-right` classes and add the necessary CSS here.

packages/insomnia-components/components/notice.stories.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import Notice from './notice';
+
+export default { title: 'Helpers | Notice' };
+
+export const _default = () => (
+  <Notice>This is a notice. It is a bright block of text, meant to stand out from the rest!</Notice>
+);
+
+export const colors = () => (
```
gschier, line 10: Can we add a story that sets the `dismissKey` too?

In order for it to not be hidden forever once clicked, we could generate a random `dismissKey` when the page is loaded (maybe set to `(new Date()).toString()`) so it will start fresh every time.

packages/insomnia-components/components/notice.js
```diff
@@ -0,0 +1,167 @@
...
+  render() {
+    const { children, color, className } = this.props;
+    const { visible, localStorageKey } = this.state;
+
+    if (!visible) {
+      return null;
+    }
+
+    return (
+      <StyledNotice className={classnames('notice', color, className)}>
```
gschier, line 155: No need for this `notice` class since we're now using StyledComponents

gschier (changes requested):
Lookin' good so far! just a few suggestions on areas to cleanup 👍

PR #2199, opened 2020-05-19
Prompt Button component migration

sonicyeti:
Migrating prompt button new SB instance at /insomnia-components/

### DIFF
--- a/packages/insomnia-components/components/prompt-button.js
+++ b/packages/insomnia-components/components/prompt-button.js
@@ -1,0 +1,129 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import autobind from 'autobind-decorator';
+import Button from './button';
+
+const STATE_DEFAULT = 'default';
+const STATE_ASK = 'ask';
+const STATE_DONE = 'done';
+
+@autobind
+class PromptButton extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      state: STATE_DEFAULT,
+    };
+  }
+
+  _confirm(...args) {
+    // Clear existing timeouts
+    clearTimeout(this._triggerTimeout);
+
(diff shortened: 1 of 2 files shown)


packages/insomnia-components/components/prompt-button.js
```diff
@@ -0,0 +1,129 @@
...
+    const finalConfirmMessage = (typeof confirmMessage === 'string'
+      ? confirmMessage
+      : 'Click to confirm'
+    ).trim();
+    const finalDoneMessage = doneMessage || 'Done';
+
+    let innerMsg;
+    if (state === STATE_ASK && addIcon) {
+      innerMsg = (
+        <span className="warning" title="Click again to confirm">
```
gschier: `warning` isn't defined anywhere in this component. We should add the style so the text correctly shows as the warning color.

packages/insomnia-components/components/prompt-button.js
```diff
@@ -0,0 +1,129 @@
...
+          <i className="fa fa-exclamation-circle" />
```
gschier, line 115: <SvgIcon> 👍

packages/insomnia-components/components/prompt-button.js
```diff
@@ -0,0 +1,129 @@
...
+
+    return (
+      <Button onClick={this._handleClick} disabled={disabled} tabIndex={tabIndex} {...other}>
+        {innerMsg}
+      </Button>
+    );
+  }
+}
+
+PromptButton.propTypes = {
```
gschier: If it's not too much trouble, can we convert this file to use Flow so we can remove PropTypes?

gschier (changes requested):
A few minor things. I also noticed the story is throwing errors on click. I think it's because the `onClick` method isn't being passed.

Issue #1666, opened 2019-09-10
Environmental Variable imports from postman evaluate to "NaN"


Spirarel:
## Bug

When you import environmental variables created in postman, any keys containing hyphens `-` evaluate to `NaN`
Ex:
```json
"api-key":"abc123"
```

### Versioning

- Insomnia Version: 6.6.2 (6.6.2.2932)
- Operating System: macOS 10.14.5

### Comments

This is probably related to #456

goodeye:
See issue  #456

Spirarel:
Hey goodeye,
Did you read the OP, where I referenced this very issue. An issue which I
commented on explaining there are common use-cases where this will be an
enduring problem?

goodeye:
Oh, sorry - I didn't see the link at the bottom of your post! It seemed like this was a duplicate, and commentary was going on with 456. I agree it's a problem - that's how I found your post. (I should probably have said "See also issue...")

Spirarel:
No worries, happens!

Hopefully, this gets some attention. It looked a little dire in that issue, but if Insomnia wants to be an alternative or replacement for Postman, it seems like a good thing to address.

veritascs:
Bump, I created #1960 which is a duplicate (but I've closed it out so we can use this existing ticket).

* doesn't matter if you import from postman or not, any environmental key with a dash will fail
* still happening on v7.1.1

DanAlexson90:
> See issue #456

Sadly closed.

andreassiegel:
I just heard about Insomnia (I'm actually using Postman) and wanted to give it a try... At first, I was amazed by the opportunity to import both collections and environments from Postman! 🎉 

Unfortunately, nothing worked after that - just because of `-` in the names of environment variables, and all variables of my environments have names with dashes. That's really sad! 😞

matheusmb:
Same issue here. Will it ever be fixed?

nciske:
Same issue.

A bit disappointed to see there's still no fix after [3 years](https://github.com/Kong/insomnia/issues/456).

rhukster:
Won't work with cloudflare API at all for this very reason.  "x-auth-email" won't allow `-` which is ludicrous. 

I was seriously considering this is a replacement for Postman, and one tiny regex fix would sort it.

nijikokun:
Hi @rhukster,

[...]

Issue #456, opened 2017-08-23
[Usability] Dash in environment config key resolves value to `NaN`


Moeriki:
- Insomnia Version: Version 5.7.14 (5.7.14.1021)
- Operating System: macOS 10.13 Beta (17A352a)

## Details

Environment keys containing dash `-` resolve their value to `NaN`.

**Base environment**

```
{
  "mykey": "myvalue",
  "my-key": "my-value"
}
```

**URL**

    http://{{ mykey  }}/{{ my-key  }}

**URL PREVIEW**

    http://myvalue/NaN

gschier:
Hmm, I swear there was an issue on this before but I can't find it. 

There is currently a warning that shows up after editing an environment if one of your variables contains dashes. This is because nunjucks (the rendering engine Insomnia uses) interprets this as subtraction (ie. `{{ my - key }}`). Maybe a more prominent message is needed?

![image](https://user-images.githubusercontent.com/587576/29635779-ea9c05fa-8802-11e7-88cc-e77980699ce3.png)

Moeriki:
After updating I have this visual message al well.

Good enough.

seangarner:
I ran into this today testing insomnia as a potential replacement for postman for my small team.  I feel like somewhat of a goon having now actually read the error message, but I did just spend 10 minutes trying to figure out what was going on and only found the solution when going to file a bug and finding this first.

Perhaps it's because I saw it as a warning message rather than an error?  I expect errors to be red, not orange, and it seems a little weird that it'd even let me save a config it knows it's broken.  Wouldn't it be better to disable the "Done" button while it's invalid?

gschier:
@seangarner it's a warning because technically there _are_ use cases where a dash is acceptable.

One way to make this better might be to treat the scenario as an error but provide an "I know what I'm doing" button to dismiss the error in the future. Thoughts?

seangarner:
Yes that'd work well @gschier.

I'm very curious what the use case for doing maths in a property name is. :thought_balloon:

gschier:
Well, it's only math if you use it as a variable like `{{ foo-bar }}`. You may want dashes in a key name for other reasons, like if you wanted to store free-form JSON in your environment that happens to use dashes. Plugins can also reference the environment, which may or may not use dashed key names for the use case.

There definitely aren't very many use cases, but even if there's only a small chance that it might be useful, I don't want to fully prevent it from happening.

Cricen:
[...]

PR #2212, opened 2020-05-23
Fix for warning not displaying when using an environment variable with '-'

### LINKED ISSUE #2207: Environment Keys does not allow "-" in their names
**Describe the bug**
If a key in environment overrides contains "-", it causes following error.
```
Error: Couldn't resolve host name
```

**To Reproduce**
1. Right-click on any of the folders on the left-hand side.
2. Click on **Environment**
3. Add following key-value pair to the window that opened:
```
  "GOOGLE-UK": "https://google.co.uk"

```
4. Click Done, and create a new request in that folder
5. Paste **{{ GOOGLE-UK }}** on Address Bar (i.e where URL is typed) and choose Environment Variable
6. Click Send

**Expected behavior**
You will see following error:

image

**Screenshots**

 - OS: macOS 10.14.6
 - App Version Version 7.1.1 (7.1.1.3350)

Vorashil:
This is a follow up to this bug report (closes #2207). I would suggest reading this bug report before reading the rest of this message.

I figured dash ( "-" )should not be allowed in environmental variables, as I found the error message saying that "_Keys must only contain letters, numbers, and underscores_".

image

Nevertheless, this ```if (!key.match(...))```  block never shows this error message when a user is using illegal characters. It was due to ```Object.keys(value)``` part.

```value``` is an object of EnvironmentInfo class. 
```
export type EnvironmentInfo = {
  object: Object,
  propertyOrder: Object | null,
};
```
So, ```Object.keys(value)``` returns ["object", "propertyOrder"]. However, Environment variables are stored as key-value pairs in the ```object``` field of  ```EnvironmentInfo``` class. So to check keys in environmental variables I needed to change 
``` 
Object.keys(value)
```
to
``` 
Object.keys(value.object)
```

### DIFF
--- a/packages/insomnia-app/app/ui/components/editors/environment-editor.js
+++ b/packages/insomnia-app/app/ui/components/editors/environment-editor.js
@@ -54,7 +54,7 @@ class EnvironmentEditor extends React.PureComponent<Props, State> {
 
     // Check for invalid key names
     if (value) {
-      for (const key of Object.keys(value)) {
+      for (const key of Object.keys(value.object)) {
         if (!key.match(/^[a-zA-Z_$][0-9a-zA-Z_$]*$/)) {
           warning = `"${key}" must only contain letters, numbers, and underscores`;
           break;
--- a/packages/insomnia-app/app/ui/components/editors/environment-editor.js
+++ b/packages/insomnia-app/app/ui/components/editors/environment-editor.js
@@ -53,7 +53,7 @@ class EnvironmentEditor extends React.PureComponent<Props, State> {
     }
 
     // Check for invalid key names
-    if (value) {
+    if (value && value.object) {
       for (const key of Object.keys(value.object)) {
         if (!key.match(/^[a-zA-Z_$][0-9a-zA-Z_$]*$/)) {
           warning = `"${key}" must only contain letters, numbers, and underscores`;

### CODE REVIEW

File: packages/insomnia-app/app/ui/components/editors/environment-editor.js [hunk 1]
```diff
@@ -54,7 +54,7 @@ class EnvironmentEditor extends React.PureComponent<Props, State> {
 
     // Check for invalid key names
     if (value) {
```
develohpanda: Looks like this condition should also be updated

```suggestion
    if (value && value.object) {
```

develohpanda (changes requested):
Thanks for looking into this! One small change, otherwise it's good.

Note: I'm going to update the PR title to reflect this change; there are similar issues (#1666 and #456) with more discussion around catering for `-` in an environment variable (and why it currently doesn't work).

Fixing the warning message is good for now, as support for `-` is a little more complex :)

packages/insomnia-app/app/ui/components/editors/environment-editor.js
Vorashil: done

develohpanda (approved):
Not sure why GH didn't run tests, but this passed locally and brought the warning behavior back. 💯

PR #2215, opened 2020-05-26
Git sync commit on Windows always shows all objects as unversioned

### LINKED ISSUE #2210: Git Sync commit always shows files as unversioned on Windows
**Describe the bug**
While committing changes with Git Sync on _Windows_, the items always come up as unversioned.

**To Reproduce**
1. Go to a new document
2. Setup git sync
3. Commit all items
4. Open the commit dialog again
5. Observer all items exist in the diff

**Expected behavior**
Should not show any changes (or unversioned items)

**Screenshots**
https://recordit.co/yq3YTeNcW3

 - OS: Windows 10
 - Installation Method: EXE
 - App Version: v2020.1.3

develohpanda:
Closes #2210 

The cause of files always showing as unversioned, is due to the difference in path separator between Windows (`\`) and POSIX (`/`) systems. Looking at code within isomorphic-git ([example](https://github.com/isomorphic-git/isomorphic-git/blob/c2cd5cb8c4f69163de535e026f9e6e09cef29dc6/src/api/status.js#L172)), it often uses a hard-coded POSIX separator (`/`), meaning the logic doesn't always work on Windows.

When running status checks on Windows to identify the git status of a file, isomorphic-git fails to operate on the filepath with a Windows separator, therefore fails to find the file in the git index. This works on macOS by default.

In this PR, any functions in `git-vcs.js` that call `isomorphic-git` with a filePath parameter, convert that filePath to POSIX separators. Additionally, the `listFiles` function converts the file paths read from the git index, to the OS specific separator. 

This way the the rest of the logic can work as normal using the OS specific separator, only converting to POSIX specifically for `isomorphic-git` at it's interface.

Breaking test that shows this behavior: 
image


gschier (approved):
I approve this change 👍

PR #2223, opened 2020-05-27
Upgrade node-libcurl

gschier:
This PR builds on top of #2150 

Test build located here: https://github.com/Kong/insomnia/releases/tag/designer%402020.1.4-alpha.1

## Summary

This PR makes the following changes:

- Update `node-libcurl` to latest release to prepare for Electron upgrade #1847 
- Switch to using `node-libcurl` static prebuilt binaries so we no longer have to build
  - Move TravisCI Mac builds back to GitHub Actions
  - Remove Docker scripts for Linux builds

## To-Do

- [x] Update CI release logic
  - [x] Add `NPM_CONFIG_...` env vars to CI because `.npmrc` was being ignored in CI
- [x] Now rely on new `node-libcurl` prebuilt binaries instead of building ourselves
  - [x] Remove Docker from build pipeline (now build directly on GH Actions)
  - [x] Remove `electron-rebuild` dependency
  - [x] Ditch TravisCI and go back to GH Actions for Mac builds (were on TravisCI before because they had the oldest version of macOS, which was required to get the most compatibility for linking to the OS version of libcurl – no longer necessary because we use a static version)
  - [x] Migrate secrets from TravisCI to GH Actions
- [x] Add `.npmrc` to project to tell NPM to install for Electron target instead of host OS
- [x] QA on macOS 14 and 15, Ubuntu 18, Windows 10
  - [x] Timeline tab
  - [x] Curl-specific auth formats
  - [x] Redirects
  - [x] SSL
  - [x] Client certificates
  - [x] Unix sockets
  - [x] `file://` URIs
- [x] **[MISC]** replace `font-scanner` with `font-manager` (`font-scanner` was an unmaintained fork)
- [x] Run unit tests
- [x] Verify reasonable binary sizes ~80MB


.github/workflows/release-designer.yml
```diff
@@ -5,49 +5,60 @@ on:
     tags:
       - designer@*
 
+env:
```
develohpanda, line 8: `release-designer.yml` and `release-core.yml` seem identical, other than the trigger tag filter. I hope GitHub Actions gets syntax similar to Azure DevOps to reference workflows instead of duplicate!

.github/workflows/release-core.yml
```diff
@@ -5,47 +5,60 @@ on:
...
         env:
-          GITHUB_REF: ${{ github.ref }}
-          GITHUB_SHA: ${{ github.sha }}
-          GH_TOKEN: ${{ secrets.GH_TOKEN }}
-          NODELIBCURL_BUILD_STATIC: yes
-          BUILD_TARGETS: AppImage,Snap,deb,tar.gz,rpm
-          SNAPCRAFT_LOGIN_FILE: ${{ secrets.SNAPCRAFT_LOGIN_FILE }}
+          BUILD_TARGETS: AppImage,deb,tar.gz,rpm
+          # BUILD_TARGETS: AppImage,Snap,deb,tar.gz,rpm
+          # SNAPCRAFT_LOGIN_FILE: ${{ secrets.SNAPCRAFT_LOGIN_FILE }}
```
develohpanda, line 62: Do these lines need a comment to explain why they are excluded?

packages/insomnia-app/app/network/network.js
```diff
@@ -267,32 +282,34 @@ export async function _actuallySend(
...
       // Setup debug handler
-      setOpt(Curl.option.DEBUGFUNCTION, (infoType: string, content: string) => {
-        const name = Object.keys(Curl.info.debug).find(k => Curl.info.debug[k] === infoType) || '';
+      setOpt(Curl.option.DEBUGFUNCTION, (infoType: string, contentBuffer: Buffer) => {
+        const content = contentBuffer.toString('utf8');
+        const rawName = Object.keys(Curl.info.debug).find(k => Curl.info.debug[k] === infoType);
+        const name = LIBCURL_DEBUG_MIGRATION_MAP[rawName] || rawName || '';
 
-        if (infoType === Curl.info.debug.SSL_DATA_IN || infoType === Curl.info.debug.SSL_DATA_OUT) {
+        if (infoType === Curl.info.debug.SslDataIn || infoType === Curl.info.debug.SslDataOut) {
```
develohpanda: Are these  being checked by Flow or were they replaced by doing a string search for `SSL_DATA_IN`?

packages/insomnia-app/scripts/build.js
```diff
@@ -136,85 +135,56 @@ async function buildLicenseList(relSource, relDest) {
...
+      },
+    );
   });
 }
 
 async function install(relDir) {
   return new Promise(resolve => {
     const prefix = path.resolve(__dirname, relDir);
 
-    // Link all plugins
```
gschier, line 183: This is no longer needed since all Lerna packages are now on NPM

packages/insomnia-app/app/node-libcurl/curl.js
```diff
@@ -14,24 +20,11 @@ export class Curl {
     }
 
     // Throw on deprecated options
```
develohpanda: Comment can be removed too 😄

packages/insomnia-app/app/node-libcurl/curl.js
```diff
@@ -79,9 +72,12 @@ export class Curl {
...
-Curl.info = nodeLibcurl.Curl.info;
+Curl.option = _Curl.option;
+Curl.info = _Curl.info;
+
+// These were removed from node-libcurl >= v2
+Curl.auth = CurlAuth;
+Curl.code = CurlCode;
+Curl.feature = CurlFeature;
+Curl.info.debug = CurlInfoDebug;
+Curl.netrc = CurlNetrc;
```
develohpanda: Two points:
- Out of ignorance, why are these being assigned outside the class declaration, instead of as static properties in the class itself?
- If we set them within the class, and convert the TS enums to Flow enums, it would be easier for flow to typecheck these values (and make them easier to consume).

packages/insomnia-app/scripts/build.js
```diff
@@ -74,7 +73,7 @@ module.exports.start = async function(forcedVersion = null) {
   await copyFiles('../app/static', '../build/static');
   await copyFiles(`../app/icons/${appConfig().appId}`, '../build/');
 
-  // Generate package.json
+  // Generate necessary files needed by `electron-builder`
```
develohpanda, line 76: 💯

packages/insomnia-app/scripts/build.js
develohpanda: Answered a question before it was asked 😄

packages/insomnia-app/app/network/__tests__/network.test.js
```diff
@@ -86,8 +86,7 @@ describe('actuallySend()', () => {
     expect(body).toEqual({
       meta: {},
       features: {
-        NO_HEADER_PARSING: true,
-        NO_DATA_PARSING: true,
```
develohpanda, line 90: I'm assuming this is removing old properties that aren't actually used? I couldn't see changes in `network.js` that would require this change to be made.

develohpanda (approved):
image

^Enough said? 😄 

LGTM, a couple of questions but not major.

Do you know if the requirement for Windows Build Tools was to allow libcurl to be built? This change might make it easier to dev on Windows too!

.github/workflows/release-designer.yml
gschier: I think there's probably a way to combine them but I didn't want to spend any unnecessary time on it. Can take a look later on to try and consolidate.

packages/insomnia-app/app/network/network.js
gschier: I'll add some Flow types

packages/insomnia-app/app/main.development.js
```diff
@@ -20,6 +21,15 @@ if (checkIfRestartNeeded()) {
 const { app, ipcMain, session } = electron;
 const commandLineArgs = process.argv.slice(1);
 
+// Explicitly set userData folder from config because it's sketchy to
+// rely on electron-builder to use productName, which could be changed
+// by accident.
+if (!isDevelopment()) {
```
gschier, line 27: I changed the product name in one of these commits without realizing `electron-builder` used it as the userData folder name. This adds a new option in the config and sets the userData folder explicitly so it's less likely we'll screw it up later.

packages/insomnia-app/app/network/network.js
gschier: Now Flow is checking it 😄

packages/insomnia-app/app/network/__tests__/network.test.js
gschier: Ya, they were able to be replaced by a single `Raw` option.

PR #2226, opened 2020-05-28
Rollback pending changes in git sync commit dialog

### LINKED ISSUE #2182: Ability to remove pending changes with git sync
I have an issue where I make some changes to saved endpoints but I wish to revert them as I've likely messed things up :trollface: 

It'd be nice to have an option similar to `git checkout .` or even `git reset --hard` with a warning dialog before executing 

Not really sure this is possible to do other than recloning the repo

**Additional context**
Would be super useful to work out (or hammer fix) potential merge conflicts

develohpanda:
Add rollback functionality; per object, and for all objects. Note: this is an initial implementation. The improvements identified a the end of this description will require some changes in the `git-staging-modal.js` implementation.

## Improvements for future PRs
- ~Remove the full modal refresh on rolling back a single file (it looks janky).~
- Only show actions on hover of row
- Update row order to be `checkbox | type | name | ... | actions (on-hover) | status icon`
- I suggest we move forward with a UX similar to VS Code, as it feels more intuitive.
  
  image

  - the sections are **changes** and **staged** (rather than _modified_ and _unversioned_, coupled with a checkbox to show the staged items). We should  either update the sections or improve our naming 👍 
  - The file specific functions (rollback, add to stage) only show on hover
  - The status colors (red, yellow, lime-green, green) are coupled with a character (D, M, U, A) for deleted, modified, untracked, and added, respectively. This might be easier to understand at a glance than the colored circles (a grey dot indicating modified doesn't feel intuitive).

Closes #2182


packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -146,10 +146,21 @@ export default class GitVCS {
     });
   }
 
-  async remove(relPath: string): Promise<void> {
+  remove(relPath: string): void {
     relPath = convertToPosixSep(relPath);
     console.log(`[git] Remove relPath=${relPath}`);
-    return git.remove({ ...this._baseOpts, filepath: relPath });
+    git.remove({ ...this._baseOpts, filepath: relPath });
```
gschier: Any reason for removing the `await` here? Without it, the caller won't bail if `git.remove` throws.

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -317,6 +328,18 @@ export default class GitVCS {
     }
   }
 
+  async undoPendingChanges(filepaths?: Array<String>): Promise<void> {
+    console.log('[git] Undo pending changes');
+
+    await git.fastCheckout({
+      ...this._baseOpts,
+      ref: await git.currentBranch(this._baseOpts),
```
gschier: Should probably be using `this.getBranch()`.

gschier:
I quite like the current UI for staging Git changes. It was based on the one included in IntelliJ

image

I agree the icons could better signify modified/deleted/etc but I feel like more validation is required to justify a larger change.

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -146,10 +146,21 @@ export default class GitVCS {
...
 
-  async remove(relPath: string): Promise<void> {
+  remove(relPath: string): void {
     relPath = convertToPosixSep(relPath);
     console.log(`[git] Remove relPath=${relPath}`);
-    return git.remove({ ...this._baseOpts, filepath: relPath });
+    git.remove({ ...this._baseOpts, filepath: relPath });
+  }
+
+  async removeUntracked(paths: Array<string>): Promise<void> {
```
gschier: It seems like this function would be better named `deleteObj` since it does not contain any logic to restrict it to untracked objects.

packages/insomnia-app/app/ui/components/modals/git-staging-modal.js
```diff
@@ -267,6 +274,31 @@ class GitStagingModal extends React.PureComponent<Props, State> {
...
 
+  async _handleRollback(item: Item) {
+    const { vcs } = this.props;
+    const { path: gitPath, status } = item;
+
+    if (status.includes('added')) {
+      await vcs.removeUntracked(gitPath);
+    } else {
+      await vcs.undoPendingChanges([gitPath]);
+    }
```
gschier: I feel like this logic could be contained within `GitVCS` inside a function titled something like `resetPath` or `checkoutPath`. It could lookup the current staging status itself and decide what to do.

packages/insomnia-app/app/ui/components/modals/git-staging-modal.js
```diff
@@ -267,6 +274,31 @@ class GitStagingModal extends React.PureComponent<Props, State> {
...
+  }
+
+  async _handleRollbackAll(items: Array<Item>, tracked?: boolean) {
+    const { vcs } = this.props;
+
+    if (tracked) {
+      await vcs.undoPendingChanges(items.map(t => t.path));
+    } else {
+      await vcs.removeUntracked(items.map(t => t.path));
+    }
```
gschier: Same comment here as above

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -317,6 +328,18 @@ export default class GitVCS {
     }
   }
 
+  async undoPendingChanges(filepaths?: Array<String>): Promise<void> {
```
gschier: Should we make this arg required since we specify it everywhere anyway (except in a test)?

packages/insomnia-app/app/sync/git/git-rollback.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import GitVCS from './git-vcs';
+
+export type FileWithStatus = { filePath: string, status: string };
+
+const isAdded = ({ status }: FileWithStatus) => status.includes('added');
+const isNotAdded = ({ status }: FileWithStatus) => !status.includes('added');
+
+export const gitRollback = async (vcs: GitVCS, files: Array<FileWithStatus>): Promise<void> => {
```
gschier, line 9: Nice, I like having this separate

gschier (approved):
Nice! Looks good to me. Just a few minor comments

packages/insomnia-app/app/sync/git/git-vcs.js
develohpanda: I fixed this hehe, noticed after during my self review.

For some reason I removed it because the source types told me it was not returning a promise, but on double checking now I was mistaken.

packages/insomnia-app/app/ui/components/modals/git-staging-modal.js
develohpanda: Extracted out of `GitStagingModal` but didn't move it directly into `GitVCS`; as per our conversation it also does not belong with the `GitVCS` git wrapper. Introduced a separate `gitRollback` utility 👍

packages/insomnia-app/app/sync/git/git-vcs.js
develohpanda: I had it there to indicate that specifying `filepaths` will filter the undo to _just_ those files (similar to how isomorphic-git has filepaths as optional).

I've improved the naming of it. I want to leave it in there to make sure that when we do want to undo _everything_ in the future, we don't need to go through the process of figuring it all out again.

PR #2227, opened 2020-05-28
Fix windows tests

develohpanda:


.github/workflows/test.yml
```diff
@@ -1,6 +1,14 @@
...
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - develop
+      - master
+  pull_request:
+    types:
+      - opened
+      - synchronize
```
gschier, line 11: Nice! This seems like it should work

gschier (approved):

PR #2230, opened 2020-06-01
Default to rendered request URL when name is not present

### LINKED ISSUE #1505: [Improvement] Allow for unnamed requests and display the url as the name
- Insomnia Version: 6.5.1
- Operating System: macOS 10.14.3

## Details

Every time I create a new request from scratch or by duplicating an existing request, I have to create a name for it.  Most of the time I'd rather that the name displayed in the sidebar just be the url or the request.  It requires an extra step to create the request, and if I it also allows for a   mismatch between the request name and the url by pasting the url as the name of the request, then changing the actual url of the request without updating the name.

I have two suggestions: 
1. Make it possible to leave the name blank, and if the name is blank, the url will show instead.
2. Don't even ask for the name upfront, thus defaulting to using the url as the name, while still allowing someone to explicitly name the request after creation.

The first is possible without the second, but I believe implementing both would significantly speed up the workflow.

jodyheavener:
Closes #1505

This PR implements the ability to use a Request's URL as its sidebar "name" in lieu of a defined value.

When creating a new request the initial value remains "My Request", but the placeholder is now "Your request URL":

Screen Shot 2020-05-31 at 11 55 51 PM

Similarly, the Request settings displays the current URL (when available) as the placeholder:

Screen Shot 2020-06-01 at 12 04 00 AM

The name in the sidebar automatically updates as you change the URL in the Request pane:

Update as you type

It supports environment variables:

Screen Shot 2020-06-01 at 12 00 03 AM

And template tags (within reason)...

A Prompt template tag does not have a value until request time, so it yields nothing:

Screen Shot 2020-06-01 at 12 02 03 AM

Dynamically generated string template tags, such as Hashes, Timestamps, and UUIDs, render once and don't change (so a timestamp, for example, may not accurately represent the request's name):

Screen Shot 2020-06-01 at 12 02 23 AM

And finally, if there's an error with a template tag or environment variable we fall back to the raw, un-rendered URL:

Screen Shot 2020-06-01 at 12 00 31 AM

---

This could probably use some tests, but I can't seem to find any integration or component-level tests (maybe I'm just blind) - if you have suggestions for testing I'll be happy to add them.


packages/insomnia-app/app/ui/components/base/editable.js
```diff
@@ -36,7 +36,7 @@ class Editable extends PureComponent {
   }
 
   _handleEditEnd() {
-    const value = this._input.value.trim();
+    const value = this._input.value.trim() || this.props.fallback;
```
jodyheavener: I think this should be undone. This is assigning the fallback value _to_ the name instead of allowing the fallback to be used at render time.

packages/insomnia-app/app/ui/components/sidebar/sidebar-request-row.js
```diff
@@ -78,12 +79,41 @@ class SidebarRequestRow extends PureComponent {
     return null;
   }
 
+  async _setRenderedUrl(props) {
```
jodyheavener: Consider throttling this. We probably don't need to update for _every_ keystroke, and it might reduce the likelihood of seeing the unparsed URL rendered in the sidebar if you're hand-typing an env var.

packages/insomnia-app/app/ui/components/modals/request-create-modal.js
```diff
@@ -36,6 +36,7 @@ class RequestCreateModal extends PureComponent {
     this._input = n;
     if (this._input) {
       this._input.value = 'My Request';
+      this._input.placeholder = 'Your request URL';
```
jodyheavener: Hmmmm. Mixing "My" and "Your"...

jodyheavener (commented):
Little self-review.

packages/insomnia-app/app/ui/components/modals/request-create-modal.js
gschier: It's usually recommended to set placeholder as an example of a valid value. I would prefer using "My Request" for both placeholder and value.

packages/insomnia-app/app/ui/components/sidebar/sidebar-request-row.js
```diff
@@ -141,6 +171,7 @@ class SidebarRequestRow extends PureComponent {
                 />
                 <Editable
                   value={request.name}
+                  fallback={this.state.renderedUrl}
```
gschier: We should probably show something when both name and URL are blank

packages/insomnia-app/app/ui/components/sidebar/sidebar-request-row.js
gschier: Yes, debouncing it seems like a good idea. 

Here's a good example to follow:
https://github.com/Kong/insomnia/blob/819f4b9cf57f433378cb9620ca671e51db365067/packages/insomnia-app/app/ui/components/rendered-query-string.js#L17-L22

gschier (changes requested):
This is a really awesome improvement, thanks for doing it! I just have a few minor comments before we can merge 😁

packages/insomnia-app/app/ui/components/modals/request-create-modal.js
develohpanda: A thought I had was to somehow mark the name as optional as well, now that it _is_ optional. eg the placeholder set as "My Request (optional)"?

jodyheavener:
@gschier updated to reflect your feedback. Notably:

A _truly_ empty request name now displays an "Empty" label, which looks slightly visually different than a valid name/URL value:

Screen Shot 2020-06-04 at 7 58 31 PM

And, the New Request modal now has a helper label next to "Name" that indicates the default will be your request URL (cc @develohpanda):

Screen Shot 2020-06-04 at 8 05 32 PM

gschier (approved):
Nice work on the empty state! I'm excited to get this in

PR #2231, opened 2020-06-01
Persist tab selection per document

### LINKED ISSUE #2229: Persist tab selection per document
**Describe the bug**
If a document was previously opened, when I reopen it I expect to be defaulted to the view I previously left off at (design or debug).

**To Reproduce**
1. Go to home
2. Click on a document
3. Navigate to debug
4. Go back to home
5. Reopen the document

**Expected behavior**
Navigate to debug by default

**Current behavior**
Always navigates to design

**Screenshots**

 - OS: macOS 10.15.3
 - Installation Method: DMNG
 - App Version: 2020.2.0

**Additional context**
This functionality was previously introduced by https://github.com/Kong/studio/issues/226. From my testing, I believe it broke as part of https://github.com/Kong/studio/pull/348. The `activeActivity` on `workspaceMeta` that is set during `wrapper > _handleWorkspaceActivityChange`, gets cleared in `git-sync-dropdown > componentDidMount > _refreshState()`

develohpanda:
Closes #2229.

`wrapper > _handleWorkspaceActivityChange` sets `workspaceMeta.activeActivity` when changing between tabs.

https://github.com/Kong/insomnia/blob/819f4b9cf57f433378cb9620ca671e51db365067/packages/insomnia-app/app/ui/components/wrapper.js#L295-L298

`git-sync-dropdown > componentDidMount > _refreshState()` is currently updating `workspaceMeta`, but with a stale object where it resets `workspaceMeta.activeActivity` back to `null`.

https://github.com/Kong/insomnia/blob/819f4b9cf57f433378cb9620ca671e51db365067/packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js#L62-L75

There is a sequencing issue in which both of these run at approx. the same time. Changing the logic from `workspaceMeta.update` to `workspaceMeta.updateByParentId` fixes this, ~but I do not know _why_, given the unfolded logic is all very similar~. `updateByParentId` does a read immediately before updating, and the `activeActivity` change is included in that.

OT: I tried to set up Spectron for a UI test but couldn't get it working in my timebox. 😢

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
@@ -66,7 +66,7 @@ class GitSyncDropdown extends React.PureComponent<Props, State> {
 
     // Clear cached items and return if no state
     if (!vcs.isInitialized() || !workspaceMeta.gitRepositoryId) {
-      await models.workspaceMeta.update(workspaceMeta, {
+      await models.workspaceMeta.updateByParentId(workspace._id, {
         cachedGitRepositoryBranch: null,
         cachedGitLastAuthor: null,
         cachedGitLastCommitTime: null,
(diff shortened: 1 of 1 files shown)


gschier (commented):
I think we should try and get to the bottom of this one. This fix maybe prevent the race condition for now, but it seems like it could break again later, down the road. Thoughts?

develohpanda:
> I think we should try and get to the bottom of this one. This fix maybe prevent the race condition for now, but it seems like it could break again later, down the road. Thoughts?

This particular one shouldn't break again I think. Because both calls that are updating meta, now read from the db immediately prior to updating. There is less time for a race condition to kick in.

It would be nice to either prioritize a more robust fix (which we need to think deeper about), or merge this in to fix this functionality and have a separate task item to make a more permanent fix. 🤔

gschier (approved):

PR #2232, opened 2020-06-01
Unit Test Generation and Running PoC

### LINKED ISSUE #2179: Unit Testing

gschier:
This is a WIP POC to add introductory unit testing functionality in a standalone library.

- [x] Mocha test suite generation
- [x] Run Mocha test suite and output to JS object
- [x] Experimental `insomnia.send()` utility using Axios
- [x] Tests

**Next Up:** Focus on the utilities that will be used for actually writing the tests (assertion library, sending requests, rendering environments, etc)

```js
// Example Suite
const { expect } = chai;

describe('Demo', () => {
  describe('Networking', () => {
    it('should send successful request', async () => {
      const resp = await insomnia.send({ url: '200.insomnia.rest' });
      expect(resp.status).toBe(200);
      expect(resp.headers['content-type']).toBe('application/json');
      expect(resp.data.foo).toBe('bar');
    });

    it('should send successful request by ID', async () => {
      const resp = await insomnia.send('req_123');
      expect(resp.status).toBe(301);
    });
  });

  describe('Random Stuff', () => {
    it('should be true', async () => {
      expect(true).toBe(true);
    });
  });
});
```


packages/insomnia-testing/src/generate/__tests__/index.test.js
```diff
@@ -0,0 +1,36 @@
+// @flow
```
develohpanda: Should this be called `fixtures.test.js` to match with the other fixture tests?

packages/insomnia-testing/src/generate/__tests__/index.test.js
```diff
@@ -0,0 +1,36 @@
...
+describe('fixtures', () => {
+  for (const input of fixtures) {
+    if (input.match(/\.output\.js$/)) {
+      continue;
+    }
+
+    const prefix = input.replace(/\.input\.json$/, '');
+    const output = `${prefix}.output.js`;
+
+    if (prefix.startsWith('skip')) {
```
develohpanda: Nice

packages/insomnia-testing/src/generate/index.js
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Generate test to a temporary file and return path. This is useful for testing
+ * @param suites
+ * @returns {Promise<void>}
+ */
+export async function generateToTmpFile(suites: Array<Suite>): Promise<string> {
+  const p = path.join(os.tmpdir(), `${Math.random()}.test.js`);
+  await generateToFile(p, suites);
+  return p;
+}
```
develohpanda: This feels better placed as a testing utility rather than exported from this package / production code

packages/insomnia-testing/src/generate/index.js
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  const lines = [];
+
+  lines.push(indent(n, `it('${escapeJsStr(test.name)}', async () => {`));
+  test.code && lines.push(indent(n + 1, test.code));
+  lines.push(indent(n, `});`));
+
+  return lines;
+}
```
develohpanda, line 77: I think the logic could be greatly simplified by replacing the indenting logic with prettier (or something more lightweight). This way it just becomes a few `.map` calls, plus the formatting can be made optional given it is typically going to be read by a machine.

Thoughts?

packages/insomnia-testing/src/generate/index.js
```diff
@@ -0,0 +1,90 @@
...
+  code: string,
+};
+
+type Suite = {
+  name: string,
+  suites: Array<Suite>,
+  tests?: Array<Test>,
+};
+
+export async function generate(suites: Array<Suite>): Promise<string> {
```
develohpanda: Don't think this needs to be `async`

```suggestion
export function generate(suites: Array<Suite>): string {
```

packages/insomnia-testing/src/run/index.js
```diff
@@ -0,0 +1,74 @@
...
+  filename: string | Array<string>,
+  options: { requests?: { [string]: Request } } = {},
+): Promise<TestResults> {
+  return new Promise(resolve => {
+    // Add global `insomnia` helper.
+    // This is the only way to add new globals to the Mocha environment as far
+    // as I can tell
+    global.insomnia = new Insomnia(options.requests);
+
+    const m = new Mocha({
```
develohpanda: Naming 😛 When reviewing a change to this file where the assignment `const m = new Mocha` is not shown, it'll be a tricky diff.

```suggestion
    const mocha = new Mocha({
```

packages/insomnia-testing/src/run/insomnia.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+export type Response = {
+  status: number,
+  statusText: string,
+  data: Object,
+  headers: {
+    [string]: string,
+  },
+};
```
develohpanda, line 19: I'm a little torn about re-declaring these types, when we already have the types defined in the core application.

In future once we have extracted logic out of `insomnia-app`, it would be nice to not redefine the types here and maintain the single source of truth.

Same with the types in `openapi-2-kong`

packages/insomnia-testing/src/run/reporter.js
```diff
@@ -0,0 +1,106 @@
+/**
+ * NOTE: This is a straight copy of the default Mocha JSON reporter, except
```
develohpanda: Is that source on github? Useful to add a permalink to this comment.

Also maybe mark what has been changed in this file explicitly. The comment here says stdout logging has been removed, _but where from and what replaced it_? 😄

packages/insomnia-testing/webpack.config.js
```diff
@@ -0,0 +1,32 @@
...
+          loader: 'babel-loader',
+        },
+      },
+    ],
+  },
+  externals: {
+    // Don't bundle Mocha because it needs to use require() to
+    // load tests. If it's bundled in the Webpack build, it will
+    // try to use Webpack's require() function and fail to
+    // import the test file because it lives outside the bundle.
```
develohpanda, line 29: 👍

Could even go to the extent of writing a unit test to check this because it is a fairly critical issue if this setting is changed.

develohpanda (commented):
Looks great! I've added a couple of comments that may be worth addressing now, but most are improvements for future.

develohpanda (changes requested):
Oops, clicked the wrong button before 😞

packages/insomnia-testing/flow-typed/axios.js
```diff
@@ -0,0 +1,5 @@
+// @flow
```
sonicyeti, line 1: Can we centralize our 3rd party module export declarations? I feel like we have batches of these sprinkled around.

sonicyeti (approved):
Excited that you've broken ground on this so quickly, can't wait to see how the UI manifests around this.

packages/insomnia-testing/flow-typed/axios.js
develohpanda: Maybe we could Script our Type management

😉

packages/insomnia-testing/src/generate/index.js
gschier: I'm inclined to take the simple route for now. Prettier is a fairly large dependency and users may not like their code being reformatted from under them.

gschier: If we needed to do something more complicated than indent, I might agree. However, the indent code here seems simple enough, has solid tests, and doesn't seem like it'll need to change much over time. I usually prefer in-house code over external dependencies for things of this nature.

packages/insomnia-testing/src/run/insomnia.js
gschier: I'm not quite sure what the best thing for this is at the moment. Sometimes it's good to have libraries define their own types because it reduces coupling. I think, if it ends up that the testing package needs to know that Insomnia exists, sharing types could make sense. However, if we can have the testing package only have to care about things inside itself, I think it makes more sense to define its own types (they may even be slightly different than the ones in Insomnia).

packages/insomnia-testing/webpack.config.js
gschier: Do you have a suggestion for this? It would require building the Webpack library for the test to use. Perhaps we could have some integration tests that run the production bundle.

packages/insomnia-testing/src/generate/index.js
develohpanda: > I usually prefer in-house code over external dependencies for things of this nature

Curious to dig into this one. But sounds good to leave as-is for this PR, for the reasons you gave. 👍

packages/insomnia-testing/webpack.config.js
develohpanda: A MVP test that will _break_ when `mocha` is removed from externals. Something as simple as:

```js
import config from 'webpack.config.js'
...
expect(config.externals.mocha).toBe('mocha');
```

develohpanda (approved):

packages/insomnia-testing/webpack.config.js
develohpanda: Added https://github.com/Kong/insomnia/blob/b1f30cdb20e79c183aff1fde463f7f4ea6d02926/packages/insomnia-testing/src/__tests__/integration.test.js#L12-L16

PR #2233, opened 2020-06-01
Force a view refresh on changing branches

### LINKED ISSUE #2224: Changing branches does not update contents of the spec editor
**Describe the bug**
Changing branches does not change the contents of the code editor in design

**To Reproduce**
1. Commit a specification while on the master branch
2. Create a new branch
3. Commit changes to the specification on the new branch
4. Switch branches (from the drop-down or branches modal)
5. Observe no change to the editor
6. Reopen the document to see the new contents

**Expected behavior**
The code editor should update to show the updated specification

**Screenshots**

 - OS: macOS 10.15.3
 - Installation Method: Dev
 - App Version: https://github.com/Kong/insomnia/commit/877a2a17254ff0d6457221a01e5cd63cfe04f1d5

develohpanda:
Closes #2224 

On changing branches, we should force an update to the views (excluding modals), because there could be some unknown areas that do not update but _should_ update. As per the linked issue, the initial observation was where the code editors were not updating correctly on changing branches.

On changing branches, `entities.initialize` is called. This is an async function, however it is not awaited, therefore allowing the UI to refresh before all of the data has been pushed through.

> Note: the delay on switching back to master (with a full spec) is to do with linting I believe.

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
@@ -79,7 +79,6 @@ class GitSyncDropdown extends React.PureComponent<Props, State> {
     const branches = await vcs.listBranches();
     const log = (await vcs.log()) || [];
     this.setState({ ...(otherState || {}), log, branch, branches });
-    handleGitBranchChanged(branch);
 
     const author = log[0] ? log[0].author : null;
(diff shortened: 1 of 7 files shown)


gschier (approved):
Looks good! Note, this fixed the problem for branch changes but we'll likely need a more robust solution for `git pull` and `git reset` functionality.

PR #2243, opened 2020-06-03
Conditionally disable legacy sync for Insomnia Core

### LINKED ISSUE #2242: Legacy sync is disabled for current session
Latest core release has the beta sync `if else` check commented out, [it appears to be a leftover task](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/index.js#L58) from the merge that went unnoticed. 

This means that legacy sync is always disabled and you always see `Sync Pending`:

image

develohpanda:
Closes #2242


gschier (approved):
Nice catch here. Let's get this out

PR #2258, opened 2020-06-08
Headless CLI POC

develohpanda:
This introduces an Insomnia CLI, with an initial command to generate configuration using the `openapi-2-kong` package. Currently the input is a file, but this should come from the git data directory, export file format, or electron appdata directory in a future PR.

The executable is called `inso` (screenshots below show `insomnia`)

- [x] Bootstrap [commander](https://github.com/tj/commander.js) with `insomnia generate config` command
- [x] Unit tests

help
image

missing required flag
image

output to console
image

output to file
image

PS: ignore the commit log lol, tests were failing _only on Windows CI_.


packages/insomnia-cli/flow-typed/openapi-2-kong.js
```diff
@@ -0,0 +1,38 @@
+// @flow
```
gschier, line 1: I wonder if it'd be useful to expose our types from the `openapi-2-kong` package.

They won't be in the compiled build, but we could probably reference them with something like 

```js
import type { DeclarativeConfigResult } from 'openapi-2-kong/types`;
```

As long as the type files are not in the `.npmignore` for `openapi-2-kong` it should work.

gschier (approved):
Just one comment about using potentially referencing types directly from `openapi-2-kong`.

Looks good beyond that!

packages/insomnia-cli/flow-typed/openapi-2-kong.js
develohpanda: Potentially too much clutter? This way we would be pulling types from the package, while also having to manually add the exports definition to `flow-typed` here.

I'm all for using types bundled within the package, but only if that removes the need for defining them in `flow-typed` altogether

PR #2265, opened 2020-06-08
Fix responses not being found if deleted active environment

### LINKED ISSUE #2237: No response found - No responses show up but request goes through
**Describe the bug**
A "No response found" log is printed in the console whenever I send any request (Error "No responses for request"). The response section still flashes "Loading..." but it says the response is never received or updated. I know the request is valid though, plus the console shows response status codes.

**To Reproduce**
Things broke after I enabled the plugin system and installed four plugins (default headers, faker, response eval, and save access token). I'm not sure at what point it happened after or during that because I tried out the plugins for a couple of requests and it was fine. 

Out of nowhere, requests stopped working. After deleting the plugins and restarting this issue persisted. I updated to version 2020.2.0 today which didn't help.

Steps to reproduce the behavior (from what I remember):
1. Install the plugins mentioned above and use them (set an environment variable to a Response=BodyAttribute using a function from plugin response-eval.
2. Open the developer console. Send a request that you know works.
3. See the response and console logs.

**Expected behavior**
A regular response.

**Screenshots**
This is when using an invalid bearer token (status code = 401)
Screen Shot 2020-06-02 at 12 35 33 PM

This is with a valid bearer token (200)
Screen Shot 2020-06-02 at 12 40 58 PM

 - OS: [e.g. macOS 10.15.1]
 - Installation Method: homebrew (original) and dmg (when updated)
 - App Version v7.1.1 (original) and 2020.2.0 (after update)

gschier:
Closes #2237

This PR handles the case where an active environment ID is set, but the environment no longer exists.

- [x] Ensure the active environment actually exists in the Redux selector
- [x] Set active environment to `null` if we're deleting it

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
@@ -24,6 +24,7 @@ import Tooltip from '../tooltip';
 const ROOT_ENVIRONMENT_NAME = 'Base Environment';
 
 type Props = {
+  handleChangeEnvironment: (id: string | null) => Promise<void>,
   activeEnvironmentId: string | null,
   editorFontSize: number,
   editorIndentSize: number,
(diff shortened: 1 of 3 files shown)


develohpanda (approved):
Nice fix 😄

PR #2267, opened 2020-06-08
Prevent linting from breaking code-editor persistence

### LINKED ISSUE #2234: GraphQL editor not saving properly
If you type in the query manually, it get saved. But if you click on the suggested pop up, it doesn't get saved.

Peek 2020-06-02 11-57

 - OS: Manjaro 20.0.2
 - Installation Method: AUR http://aur.archlinux.org/packages/insomnia
 - App Version

image

gschier:
Fixes #2234

This PR modifies our strategy for setting codemirror options. Since setting an option can throw an exception (eg. `setOption('lint', {schema: invalidGraphQLSchema})`), we need to do so with caution.

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
@@ -310,7 +310,8 @@ class CodeEditor extends React.Component {
         try {
           const parsed = JSON.parse(toParse);
           count = Object.keys(parsed).length;
-        } catch (e) {}
+        } catch (e) {
+        }
         return count ? `\u21A4 ${count} \u21A6` : '\u2194';
       },
     };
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -649,25 +650,41 @@ class CodeEditor extends React.Component {
...
+      return;
+    }
+
+    // Set the option safely. When setting "lint", for example, it can throw an exception
+    // and cause the editor to break.
+    try {
+      cm.setOption(key, value);
+    } catch (err) {
+      console.log('Failed to set CodeMirror option', err.message, { key, value });
+    }
```
develohpanda, line 684: It's strange that this wasn't happening before... there must have been a new CodeMirror release since the released bundled in Core@7.1.1 that caused this?

In any case, awesome fix, this is much safer now. 🎉

develohpanda (approved):

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
gschier: I think it was fine until we added some logic around enabling/disabling linting for the Spec editor stuff. We were setting the lint option in the onChange, which prevented the change from triggering the callback.

PR #2275, opened 2020-06-10
Add fallback status messages for HTTP/2

### LINKED ISSUE #2273: Response status code
previously, it used to be possible to see the description of each status code in the response. something like: `200 OK` or `404 Not Found` or whatever. now, i think since a recent update, it only shows the code and not the text. is this an intended feature? is there any way that i can get the text back?

i am on the latest version (v2020.2.2) on macos 10.15.4.

gschier:
Closes #2273

HTTP/2 removes the status message field for responses so Insomnia only shows the numeric status code in that case. This PR defines a set of default status messages that will be shown if one is not returned.

image


sonicyeti (approved):
This will be super helpful, sweet!

PR #2284, opened 2020-06-11
Add "When Expired" setting for request resend trigger condition.

### LINKED ISSUE #1972: [Feature Request] Allow variables to expire after a set time.
# Problem
Our server provides 5 minute session tokens. Currently I can either (1) manually send a login request to save the `session` environment variable when I want a new token or (2) configure insomnia to login every time I send any request.

There are occasional other problems with using a new session token every time, which precludes always using option 2 for me.

# Desired solution
I would like to configure the `session` environment variable to:
1. re-send when necessary
2. expire after X minutes

This would avoid spamming the server with login requests but also avoid the annoying manual session refresh. The feature seems general enough that it would be useful for other authentication mechanisms or dependent request triggering.

augustoroman:
This adds a new trigger condition for re-sending a request for the
`response` plugin called "When Expired". This condition allows
specifying a maximum age of a response to use. This is very useful when
configuring a hierarchy of environment variables that ultimately depend
on an auth token that needs to be refreshed periodically.

Previous workarounds are:
1. Use "always refresh". This fails in cases where you must use the same
   auth token since it is refreshed for every request.
2. Use "no history" or "never" and manually refresh the token. This is
   inconvenient and requires pinning the token.

Alternative solutions are:
1. Add more sophisticated logic to determine when the token needs to be
   refreshed (e.g. refresh if the current request returns a `401` status
   code)
2. Write a custom response plugin to time out the token. I believe this
   functionality is small enough and general enough to be contributed
   back generally.

Fixes #1972

### DIFF
--- a/plugins/insomnia-plugin-response/__tests__/index.test.js
+++ b/plugins/insomnia-plugin-response/__tests__/index.test.js
@@ -470,6 +470,45 @@ describe('Response tag', () => {
       );
     });
 
+    it('sends when behavior=when-expired and no responses', async () => {
+      const requests = [{ _id: 'req_1', parentId: 'wrk_1' }];
+      const responses = [];
+      const context = _genTestContext(requests, responses);
+
+      expect(await tag.run(context, 'raw', 'req_1', '', 'when-expired', 60)).toBe('Response res_1');
+    });
+
+    it('sends when behavior=when-expired and response is old', async () => {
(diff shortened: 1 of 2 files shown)


gschier (approved):
This is awesome! Thanks for the improvement @augustoroman

misterpickypants:
Any chance you could make a point release with this?

PR #2286, opened 2020-06-12
The One With Hooks

develohpanda:
[We finally did it.](https://www.youtube.com/watch?v=dQw4w9WgXcQ) 

> For reference, the original error was that insomnia-components was unable to locate the react-dom module. Changing the way we alias resolves that issue.

This error would surface as the common invalid react hook error seen [here](https://reactjs.org/warnings/invalid-hook-call-warning.html).


packages/insomnia-app/package.json
```diff
@@ -74,7 +74,6 @@
     "nunjucks",
     "pdfjs-dist",
     "prop-types",
-    "react",
```
develohpanda: This was the culprit

packages/insomnia-components/webpack.config.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
     library: 'insomniaComponents',
     libraryTarget: 'commonjs2',
   },
+  externals: ['react'],
```
develohpanda: Tell `webpack` to not bundle react, when packing `insomnia-components`

gschier (commented):
I _think_ setting React as an `external` in Webpack (by removing it from `packedDependencies`) means that we'd be using the development version in production (since it won't be processed) which could have a significant performance hit.

It would be nice to figure out the root of the problem.

packages/insomnia-components/webpack.config.js
gschier: Should also have `react-dom` in here

develohpanda:
My understanding is setting mode to production achieves that result, not externals. See the last paragraph [here](https://webpack.js.org/guides/production/#specify-the-mode).

But you're right, setting as external in the app doesn't seem correct.

gschier:
Production mode is definitely required, but I don't think it will do anything to externals since externals are not included in the outputted bundle. Specifying `module-name` as external causes webpack to not process it at all and will simply include `require('module-name')` in the bundle instead.

On a positive note, I stumbled upon a solution to the `react` external testing out this fix. See #2292 which uses this branch as a base.

gschier (approved):
Yay, it feels so good to have gotten to the bottom of this

PR #2288, opened 2020-06-12
[insomnia-importers]: Update programmatic usage example in README

sebastianjonasson:
The example wasn't working out-of-the-box as the `importers.convert` returns a promise.


gschier (approved):
Thanks for the correction 😄

PR #2291, opened 2020-06-12
Prevent insomnia-components from bundling entire crypto module

gschier:
I was wondering why the bundle for `insomnia-components` was over 500KB. Upon investigation, it was because the `<GravatarImg>` component was pulling in the entire `crypto` module to do an `md5` hash. This PR includes a dedicated `md5` module to do it instead, lowering the bundle size to ~230Kb


develohpanda (approved):
Nice 😄 

Is there a significant difference in the bundle size with/without `md5`? If so, we could set it as a peer dependency as well and have webpack not bundle it in.

gschier:
Looks like md5 is only 5kb. I did read that it's recommended to set styled-components as a peer dependency though

PR #2292, opened 2020-06-12
Fix for react-dom alias to @hot-loader/react-dom

gschier:
As I was testing out the fix for #2286 I stumbled upon this section of the `react-hot-loader` docs:

https://github.com/gaearon/react-hot-loader/tree/92961be0b44260d3d3f1b8864aa699766572a67c#linking

It specifies to use a different linking technique when using `npm link`. Since we're using Lerna, we need to follow this practice because Lerna uses linking heavily. 

For reference, the original error was that `insomnia-components` was unable to locate the `react-dom` module. Changing the way we alias resolves that issue.

### DIFF
--- a/packages/insomnia-app/webpack/webpack.config.base.babel.js
+++ b/packages/insomnia-app/webpack/webpack.config.base.babel.js
@@ -46,7 +46,11 @@ module.exports = {
   },
   resolve: {
     alias: {
-      'react-dom': '@hot-loader/react-dom',
+      // Create aliases for react-hot-loader
+      // https://github.com/gaearon/react-hot-loader/tree/92961be0b44260d3d3f1b8864aa699766572a67c#linking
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/webpack/webpack.config.base.babel.js
```diff
@@ -46,7 +46,11 @@ module.exports = {
   },
   resolve: {
     alias: {
-      'react-dom': '@hot-loader/react-dom',
+      // Create aliases for react-hot-loader
+      // https://github.com/gaearon/react-hot-loader/tree/92961be0b44260d3d3f1b8864aa699766572a67c#linking
+      'react-hot-loader': path.resolve(path.join(__dirname, '../node_modules/react-hot-loader')),
```
develohpanda, line 51: Stumbled across this syntax in other docs but only for `react` and `react-dom` and it didn't work. I didn't think to alias `react-hot-loader` :(

PR #2294, opened 2020-06-15
Support CLI operations against the `.insomnia` git data directory

### LINKED ISSUE #2293: Read from `.insomnia` git data directory
`inso` currently works with an API spec on the filesystem itself. It should also allow reading from the `.insomnia` git data directory, in order for commands to run seamlessly. You should be able to access a model via an id-ish, ie: `inso generate config spc_123`.

In the future we also want to use other data sources, like the Insomnia export file format `insomnia-export.yaml`, and the Insomnia Electron AppData directory `NeDB`. If the implementation of supporting the `.insomnia` git data directory allows for supporting the above items, 💯.

There is likely some logic that needs to be extracted from `insomnia-app` into a shared library such as `insomnia-stdlib`. We need to ensure that:
- interactive debugging capability is not lost after extraction,
- the code is still type-safe, and insomnia db model types are shared

develohpanda:
image

Approach:
- `db.init` at the start (simplified init logic extracted from insomnia-app)
- `db.seedGitDataDir` as part of the logic in a command. If `.insomnia` is not found, etc, no data is seeded and db remains empty
- search for `<identifier>`:
  - if in db, extract `apiSpec.contents` and send to `openapi-2-kong.generateFromString`
  - if not in db, then it may be a file; send `<identifier>` to `openapi-2-kong.generate`
  - if not in file, o2k throws error, which is returned to console

I envision more `db.seedFrom*` functions, that are smarter, and know if the db should seed from the git data dir, insomnia appdata, or export format, etc.

I also expect the db from the CLI to be read-only.

Key changes are in the following files, I suggest starting there.
image

TODO:
- [x] Simplify `git-repo` fixture (it doesn't need to have that many files)
- [x] Tidy up `git-nedb.js` (no extraction to `insomnia-stdlib` yet)
- [x] Proper error handling
- [x] Improve `GlobalOptions<T>`
- [ ] Docs (after finalizing the flag names and aliases)

Next:
- [ ] Support id-ish
- [ ] Extract db model information to a separate package, in order to share types between app and CLI
- [ ] Unless the user clones the repo (and sets the `workingDir`). this can't be run 'locally' because the `.insomnia` directory doesn't technically exist

Closes #2293


gschier (commented):
It might be better to have the NeDB adapter relationship "inverted." For example, if the CLI code operated against a simpler storage format like an in-memory object of `{ [id: string]: Model }`, it could have an NeDB adapter to convert to that simpler format. I think it makes sense to map to the format that's best for the library, instead of mapping to the most common (potentially) format.

packages/insomnia-cli/src/db/types.js
```diff
@@ -0,0 +1,23 @@
...
+
+export type BaseModel = {
+  _id: string,
+  type: string,
+  parentId: string,
+  modified: number,
+  created: number,
+};
+
+type BaseApiSpec = {
```
gschier, line 15: Just a note here that it may not actually be necessary for this module to know the full types of all the models. For example, right now it's only referencing the `contents` property of an ApiSpec. It doesn't really even need to know or care that there are other properties.

Not suggesting to make any changes but something to think about going forward. Usually, the less a module needs to know about the outside world, the better.

gschier (approved):
This looks good @develohpanda! 😁

Just a note that I wasn't suggestion to remove the NeDB stuff entirely in my last review. Only to change the base DB for this module from NeDB to an in-memory store. It could still make sense to use the work you already did to add an `NeDBDb()` function that also returns a `Database`, similar to `gitDataDirDb()`.

Up to you, though. We can also add that later if we end up needing the ability to read from NeDB.

PR #2295, opened 2020-06-15
Edit variable modal sort environment variables in dropdown 

### LINKED ISSUE #2277: Edit Variable dialog's Environment Variable dropdown should be sorted.
## Description
In a request body, when I click on the widget that represents my variable, an "Edit Variable" dialog box pops up that has a drop-down/combo-box for switching the Environment Variable. The list in the drop-down is not alphabetically sorted, so it's really hard to find the variable I want to switch to. The list should be (alphabetically) sorted according to what names are displayed (omitting the "-- Custom --" one).

## To Reproduce
1. Define some variables in the Base Environment for a Workspace. 
1. Make a new request that has a request body and insert one of the variables.
1. Click that variable to pop-up the dialog
1. Click the drop-down for the "Environment Variable"
1. Note that the list (omitting "-- Custom --") is not sorted alphabetically.

## E.g. Screenshot
image

## Additional wrinkle
If you **only** define variables in the "Base Environment", the drop-down tends to preserve the as-defined order. However, I have found that this is not always the case (though I don't have steps to reproduce). When you define overrides in Sub Environments and you keep adding variables to the base, the drop-down order appears to change randomly after each edit.

anyxem:
Closes #2277

This PR adds sort to variable dropdown in Edit Variable

Screenshot 2020-06-15 at 18 28 32
Screenshot 2020-06-15 at 18 27 38

### DIFF
--- a/packages/insomnia-app/app/ui/components/templating/variable-editor.js
+++ b/packages/insomnia-app/app/ui/components/templating/variable-editor.js
@@ -66,7 +66,7 @@ class VariableEditor extends PureComponent {
     }
 
     const context = await this.props.handleGetRenderContext();
-    const variables = context.keys;
+    const variables = context.keys.sort((a, b) => (a.name < b.name ? -1 : 1));
     const variableSource = context.context.getKeysContext().keyContext[cleanedValue] || '';
     // Hack to skip updating if we unmounted for some reason
     if (this._select) {


gschier (approved):
Thanks for this QoL improvement 😄

PR #2305, opened 2020-06-17
Support toggling of auto beautify in editor.

l0lcat:
This feature adds a button to toggle beautify in the editor. When the checkbox is toggled, the text is 'beautified' immediately. While the checkbox remains toggled, text is prettified when:
- Text is pasted into the editor.
- When the editor is blurred.

The following screenshots illustrates the added button and it's orientation.
Disabled:
image

Enabled:
image

Of course I'm open for suggestions of changes in UI / functionality in order to complete the PR!

Kind regards,

Erik


anyxem:
Hello, look great, however there are some room for improvement, what do you think about hiding checkbox. that toggle takes place, but I believe that it's state will change rarely, so it can be hidden.
What if add `...` or `arrow` near Beautify JSON and hide this checkbox in menu

Screenshot 2020-06-17 at 12 32 39

just IMO

l0lcat:
That is a great idea! The checkbox was already bugging me and I like your suggestion. I will change it tonight.

sonicyeti:
@erickrico

l0lcat:
So, I've made a few changes according to the suggestion. Next to the button, there is nu a 'caret-up' button to show the dropdown. In the dropdown the desired mode can be selected. The text of the button changes based on the selected mode. When in 'manual' mode, the button can be clicked to trigger the beautify process. In 'auto' mode, the button cannot be clicked and the process is triggered automatically (as described in the original post). 

See the screenshots below for illustration of the new situation.

Dropdown:
image

Hovering the button in 'manual' mode:
image

Hovering the button in 'auto' mode:
image

I could move the 'caret-up' further to the left, but then it will overlap with the button, that didn't look nice. Is this what you had in mind @anyxem? IMO it already looks much cleaner this way. 

I'm just thinking, maybe the term 'manual' can be removed from the button, so it will show "Beautify {contentType}" or "Auto beautify {conentType}". What d'you think?

Also (final note), I inserted the 'caret-up' as DropdownButton next to the existing 'beautify' Button. Is this a proper implementation?

anyxem:
@l0lcat yes, much better! Good job
There is still space for discussion about labels, I think that Manual Beautify JSON should stay the same: "Beautify JSON", so caret will catch attention to the available options.
As for Auto I can't suggest the better option for now.

l0lcat:
@anyxem So, just to verify your suggestion: 
- 'Manual beautify JSON' should be changed to 'Beautify JSON'. 
- 'Auto beautify JSON' will not change.

anyxem:
"Beautify JSON" - is action
"Auto beautify JSON" - in this case is a _status_ because we choosed auto beautify in menu, so I think it would be better to type "Auto Beautify JSON is ON"

l0lcat:
Okay, here are some screenshots:

Beautify JSON:
image

Auto beautify JSON enabled:
image

Auto beautify JSON is ON:
image

'Beautify JSON' is a no-brainer IMO.

My personal preference for 'auto' would be 'Auto beautify JSON enabled'.

anyxem:
I see that caret behaviour is differ from other UI elements, everywhere it stick to label and menu appear on click to label, so caret just show that label have some menu, sorry for confusion with caret. 

It might be better to show this menu on click to some settings icon (gear or `...`), it may be in the same way like we have history icon in response window. In request we can have gear Icon (or dots) to place request view options, first will be Beautify JSON options, then it may be a room for other options

Screenshot 2020-06-17 at 23 36 51

l0lcat:
Good point, UI behaviour should always be uniform. Also, the gear will be more verbose to the user telling that there are more options.

Screenshots:
image

image

Big picture:
image

Gear on the left side of the button:
image

I think the gear should be on the left side, looking at the larger screenshots. The gear on the right side of the button seems a bit 'lost'.

nijikokun:
I'd opt to place the setting in the preferences pane since the cog button can be mistaken as for the preferences / settings panel of the request itself rather than that specific option

l0lcat:
I'm not sure, the 'cog' is not only used for preferences, i.e. in the query parameters tab:

image

l0lcat:
I've looked at the CSS and found some other options, all are visualised in the screenshot:

image

sonicyeti:
> I've looked at the CSS and found some other options, all are visualised in the screenshot:
> 
> image

The wrench may be a good fit for now. @nijikokun thoughts?

l0lcat:
> > I've looked at the CSS and found some other options, all are visualised in the screenshot:
> > image
> 
> The wrench may be a good fit for now. @nijikokun thoughts?

For what it's worth, this is also my preference!

sonicyeti (approved):
Thank you for the contribution, we've got a release coming up and are excited to have this in!

PR #2307, opened 2020-06-18
CLI: input and output flags should respect workingDir

develohpanda:
As per title, currently setting `workingDir` does not change the directory used when reading an input and writing an output.


gschier (approved):
LGTM!

PR #2309, opened 2020-06-18
Feature: Show Dev Tools with hotkeys

### LINKED ISSUE #2157: Show Dev Tools with ⌘⌥I  (⌃⇧I)
Show Dev Tools with `⌘⌥I` (or `⌃⇧I`) like all the browsers do, and many Electron apps.

Also, it would be a nice if some level of smarter completion occurred when editing environment JSON - tabbing through the entries, auto-insert quotes, etc. Enter could append a new entry like:
```
{    ...
    "aKey": "aVal",
    "⎽": ""
}
```

nicholasraphael:
Closes #2157 

We want to allow users to show Dev Tools with ⌘⌥I (or ⌃⇧I) like all the browsers do, and many Electron apps.


gschier (changes requested):
Thanks for taking this one on @nicholasraphael! 

The solution here is actually much simpler than you might expect. In `window-utils.js` you will notice the menu item for toggling devtools. There are actually two in there: one in the Developer menu (which has an "accelerator" and is only shown when building from source) and one in the View menu (which doesn't have a hotkey). We should be able to add the Accelerator to the item in the View menu and remove the menu item from the Develop menu.

Let me know if that makes sense! ⚡

nicholasraphael:
@gschier Thanks for the help 😄! Let me know if this is looking good.

packages/insomnia-app/app/common/hotkeys.js
```diff
@@ -163,6 +163,8 @@ export const hotKeyRefs: { [string]: HotKeyDefinition } = {
 
   ENVIRONMENT_UNCOVER_VARIABLES: defineHotKey('environment_uncoverVariables', 'Uncover Variables'),
 
+  SHOW_DEVTOOLS: defineHotKey('showDevTools', 'Show Developer Tools'),
```
gschier: We don't actually need the edits you made to this file, since we don't currently have a way to configure hotkeys for actions in the app titlebar.

gschier (changes requested):
One comment but looks good after that's addressed!

packages/insomnia-app/app/common/hotkeys.js
nicholasraphael: Removed the edits

gschier (approved):
Nice, thanks for this!

PR #2311, opened 2020-06-19
Add unit testing to CLI, configurable Mocha reporter

### LINKED ISSUE #2179: Unit Testing

### LINKED ISSUE #2312: Implement unit testing capability for CLI

develohpanda:
New command: `inso run test --reporter progress --bail` (screenshot is out of date)

- Add support for various Mocha reporters (only supports single reporter for now, can support multiple in future with [mocha-multi-reporters](https://www.npmjs.com/package/mocha-multi-reporters)).
- Add support for `--bail` Mocha flag

Currently uses a hard-coded test suite.

image

> Note: currently fails with the reference error, resolved in a separate branch.

Todo:
- [x] Exit code 1 if any test failures (completed in #2318)
- [x] Unit tests

Closes #2312


packages/insomnia-cli/src/commands/run-tests.js
```diff
@@ -0,0 +1,67 @@
...
+    console.log(`Reporter "${reporter}" not unrecognized. Options are [${reporterTypes}].`);
+    return false;
+  }
+
+  return true;
+}
+
+export async function runInsomniaTests(options: RunTestsOptions): Promise<void> {
+  if (!validateOptions(options)) {
+    return;
```
gschier, line 31: Should be exiting with error code here right?

packages/insomnia-testing/src/run/index.js
```diff
@@ -45,27 +45,53 @@ type TestResults = {
...
 
     const mocha = new Mocha({
       global: ['insomnia'],
+      bail,
     });
 
-    mocha.reporter(JavaScriptReporter);
+    switch (reporter) {
+      case 'dot':
+        mocha.reporter(Mocha.reporters.Dot);
```
gschier: `reporter()` accepts a string too instead of a constructor so this switch may not be necessary. I'm fine with the explicitness if you want to leave it, though.

packages/insomnia-testing/src/run/index.js
```diff
@@ -45,27 +45,53 @@ type TestResults = {
...
+    }
 
     const filenames = Array.isArray(filename) ? filename : [filename];
     for (const f of filenames) {
       mocha.addFile(f);
     }
 
     const runner = mocha.run(() => {
-      resolve(runner.testResults);
+      resolve(runner.testResults || { stats: runner.stats });
```
gschier: I wonder if we should have two separate `runTest` functions. One for running tests that return a JS object and one that outputs to `stdout`. I think that would clean of some of the logic in here like this part. If it's not the `js` runner, we don't really need to return anything.

gschier (changes requested):
A few minor comments but it looks pretty much good to go!

packages/insomnia-testing/src/run/index.js
develohpanda: We _do_ need to return stats (or a true/false result), so that `inso` can exit with an appropriate exit code (#2318 does this). We could also use https://www.npmjs.com/package/mocha-multi-reporters to always use the JS reporter, in addition to whatever the user specifies (although I didn't test it).

packages/insomnia-testing/src/run/index.js
develohpanda: Oh 😆 removing the switch would be ideal, so I will clean this up

packages/insomnia-cli/src/commands/run-tests.js
develohpanda: Yah, done so in #2318 ❤️

packages/insomnia-testing/src/run/index.js
gschier: Oh I see. So the reporter prints to the console as it runs but the stats are the only way to detect whether it passed or failed? 

My hesitation here is that `runTests()` returns an object with more than just a stats field. Since these two cases return different shapes, it may make sense to have another function that simply returns stats (or even a boolean if that's all we need).

packages/insomnia-cli/src/commands/run-tests.js
gschier: Nice!

packages/insomnia-testing/src/run/index.js
develohpanda: > The stats are the only way to detect whether it passed or failed?

To my knowledge, that's correct. The `stats` property is populated by each reporter.

I agree about separating the functions. One is solely for the JS reporter (and does not allow for overriding the reporter, used by insomnia-app), while the other function has a configurable mocha reporter. Will make the change.

packages/insomnia-testing/src/run/index.js
```diff
@@ -61,40 +65,39 @@ export async function runTests(
...
     const runner = mocha.run(() => {
-      resolve(runner.testResults || { stats: runner.stats });
+      resolve(extractResult(runner));
 
       // Remove global since we don't need it anymore
       delete global.insomnia;
     });
   });
 }
+
```
gschier: Nice refactor!

gschier (approved):
Looks good after we get those conflicts resolved 🥳

PR #2324, opened 2020-06-25
Add multi-switch component, refactor radio group, and add success icon

gschier:
This PR adds a new <MultiSwitch> component, which is a styled version of <RadioButtonGroup> to act as the activity toggle when we have more than two options.

https://deploy-preview-2324--insomnia-storybook.netlify.app/?path=/story/1st-party-multi-switch--default

image

This PR also includes a green checkmark that will be used for unit test results

image

And finally, also fixes the dark theme in storybook (re-imported themes from app)


sonicyeti (approved):
Looks great, glad we were able to get this component in play!

alpaca_gifanime

PR #2318, opened 2020-06-24
CLI linting + exit codes

develohpanda:
New command: `inso lint spec <identifier>`

image

- Very basic UI for presenting lint results for now. Uses Spectral, same as the app itself.
- Added `util.exit()` and `util.logErrorExit1()` functions to wrap the command result and call `process.exit()` with appropriate error codes.
- Each command returns true/false, or throws an error, which controls the exit codes and allows for easier unit testing.

Updates to `inso run test`:
- Automatically delete generated test file after a test run
- Update unit tests to delete any test files that are generated on the filesystem in `os.tmpdir()`
- Add `--keep-file` option to `inso run test` command, to not delete the generated test file, and print to console the file location. (These tests generate at `os.tmpDir()/insomnia-cli/*.test.js`)
image

TODO in future PRs:
- Mock the db layer, to make testing easier
- Improve linting output

Closes #2321

PR #2325, opened 2020-06-26
Update test fixtures and support sendRequest() callback

gschier:
This PR adds a `sendRequest` callback to the `insomnia-testing` package to enable the ability to use the app's native networking library to handle requests. It's still rough but should be a good starting point until we have a better understanding of the requirements.

This PR also changes `runTests()` to accept stringified source code instead of a file path.


packages/insomnia-testing/src/generate/index.js
```diff
@@ -17,6 +18,12 @@ export type TestSuite = {
 export function generate(suites: Array<TestSuite>): string {
   const lines = [];
 
+  lines.push(`const { expect } = chai;`);
```
gschier, line 21: This is not cool. Since the tests are being run inside a Jest environment, it wasn't complaining that `expect()` did not exist as a global. Once I tried using it in the app, it didn't exist. In the future, I think it would be better to somehow run the tests in a sub-process or something, so no global state is shared.

develohpanda: Explains why tests also failed when running from CLI. Worth nothing this as a TODO in the code itself and/or creating a card on the kanban board

packages/insomnia-testing/src/generate/index.js
```diff
@@ -25,8 +32,16 @@ export function generate(suites: Array<TestSuite>): string {
...
+  return new Promise((resolve, reject) => {
+    const js = generate(suites);
+    return fs.writeFile(filepath, js, err => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
```
develohpanda, line 44: What was the reason for this change? 🤔

packages/insomnia-testing/src/run/index.js
```diff
@@ -44,31 +53,43 @@ type TestResults = {
...
+
+/**
+ * Copy test to tmp dir and return the file path
+ * @param src - source code to write to file
+ */
+function writeTempFile(src: string): string {
+  const p = path.join(os.tmpdir(), `${Math.random()}-test.js`);
+  fs.writeFileSync(p, src);
+  return p;
+}
```
develohpanda: Two notes about this:
1. we should generate to a folder in the tempdir (like most other applications),
2. we should make sure to delete the file after generation. It might be good to have the option to keep the file as well, maybe? 👇 From CLI #2318 
https://github.com/Kong/insomnia/blob/c593b396a8c9a9526ae1f4d0c82788c3b91478ef/packages/insomnia-cli/src/cli.js#L43 https://github.com/Kong/insomnia/blob/c593b396a8c9a9526ae1f4d0c82788c3b91478ef/packages/insomnia-cli/src/commands/run-tests.js#L33-L44

packages/insomnia-testing/src/run/index.js
```diff
@@ -1,9 +1,13 @@
 // @flow
```
develohpanda, line 1: We're going to have some conflicts with this file 😢 lets cross that bridge when we get to it

packages/insomnia-testing/src/run/insomnia.js
```diff
@@ -18,48 +25,88 @@ export type Response = {
   },
 };
 
+export type InsomniaOptions = {
+  requests?: Array<$Shape<Request>>,
+  sendRequest?: (requestId: string) => Promise<$Shape<Response>>,
```
develohpanda, line 30: Yay `$Shape<T>` 🎉

develohpanda (changes requested):
> This PR also changes runTests() to accept stringified source code instead of a file path.

I agree with this, but I've made two points further down that I think are worth discussing before merging.

Nice work pulling in `insomnia` networking magic. Is there anything extra we need to think about when trying to get these tests to run through CLI?

packages/insomnia-testing/src/generate/index.js
gschier: Ya, you get a big angry warning when using the `fs.promises` API in our current version of Node 
image

packages/insomnia-testing/src/run/index.js
gschier: I like keeping the temp file around because it provides the ability to debug if you get an exception that includes a stack trace. Since the files are so small, it doesn't seem like a big deal to keep the files around until the OS cleans them up. 

I would side on not deleting them unless there is a good reason to do so but am not super adamant on it. Thoughts?

gschier: A agree with the folder thing, though. Will address that one.

packages/insomnia-testing/src/generate/index.js
develohpanda: Oh I see 🤔 I thought git-sync was using `fs.promises`, but I now see the elusive workaround (and future proofing 😉)

https://github.com/Kong/insomnia/blob/1c46b95ac90b16cc099ad8657ace91398eba149c/packages/insomnia-app/app/sync/git/fs-plugin.js#L61

`mem-db.js` in the CLI currently uses `fs.promise`, I suppose that needs to be updated? Thought I haven't seen this warning.

packages/insomnia-testing/src/run/index.js
develohpanda: My thoughts are to:
1. Delete by default,
2. Keep file if there are any test failures (this can be configurable if need be)
3. Allow an override to keep the file _always_ if you so desire (eg. a CLI `--keep-file` flag)

But mainly, point 1 - deleting by default. Only in exceptional cases should a user have a reason to look at the file (failures, a stacktrace pointing to the filepath). We have no reason to keep it around if those exceptional cases don't arise.

gschier: Added the deletion logic and sub folder 😄

packages/insomnia-testing/src/generate/index.js
gschier: Perhaps Node doesn't show those types of warnings. I'm not sure

PR #2328, opened 2020-06-26
Design sidebar

sonicyeti:
Initial PR covering the integration of the latest sidebar design into both SB and the application.


packages/insomnia-components/components/sidebar.js
```diff
@@ -140,204 +213,637 @@ const StyledPanel: React.ComponentType<{}> = styled(motion.div)`
...
-  const [visible3, setVisible3] = useState(true);
-  const [visible4, setVisible4] = useState(true);
-  const toggleVisible = () => setVisible(!visible);
-  const toggleVisible2 = () => setVisible2(!visible2);
-  const toggleVisible3 = () => setVisible3(!visible3);
-  const toggleVisible4 = () => setVisible4(!visible4);
+  // Section Expansion & Filtering
+  // Info
+  const [infoSec, setInfoSec] = useState(false);
+  const toggleInfoSec = () => setInfoSec(!infoSec);
```
develohpanda: Quick note: these callbacks will always create a new instance every time, causing the child components to update unnecessarily due to a prop change, even though most of the `toggle` callbacks will remain unchanged.

I suggest the `useCallback` [hook ](https://reactjs.org/docs/hooks-reference.html#usecallback) to set these, which returns a memoized callback if the dependencies (in this case, `infoSec` and `setInfoSec`) don't change. It's a little more verbose, but you can write your own hook to wrap that logic, eg:

```js
const useToggle = (state, set) => React.useCallback(() => set(!state), [set, state]);
```

Resulting in 
```suggestion
  const toggleInfoSec = useToggle(infoSec, setInfoSec);
```

packages/insomnia-components/components/sidebar.js
```diff
@@ -140,204 +213,637 @@ const StyledPanel: React.ComponentType<{}> = styled(motion.div)`
...
+                  </StyledItem>
+                )}
+              </React.Fragment>
+            ))}
+          </StyledPanel>
+        </StyledSection>
+      )}
     </StyledSidebar>
   );
 }
```
develohpanda: Is suggest to break this component down in several ways:
1. By each logical section. Components like `<SidebarServers servers={props.spec.servers} />`. Each component can then contain it's respective state and dispatch values, isolated.
2. By the styled components, and the children you expect inside each styled component. A good example would be `StyledFilter` (see comment below).
3. As a nice to have, basic stories for each `<Sidebar* />` component in addition to the all-in-one.

I think you've done an amazing job with how it visually looks. With this breakdown, it will be much easier to unit test, debug and review, and will remove a lot of the code duplication.

Thoughts?

packages/insomnia-components/components/sidebar.js
```diff
@@ -140,204 +213,637 @@ const StyledPanel: React.ComponentType<{}> = styled(motion.div)`
...
+              initial={{ height: schemasFilter ? '100%' : '0px' }}
+              animate={{ height: schemasFilter ? '100%' : '0px' }}
+              transition={{ duration: 0.2, ease: 'easeInOut', delay: 0 }}>
+              <input
+                type="text"
+                value={schemaFilter}
+                onChange={e => setSchemaFilter(e.target.value.toLowerCase())}
+                placeholder="Filter..."
+              />
+            </StyledFilter>
```
develohpanda: Sections like this are duplicated a lot, with the only difference being the value (`schemasFilter` vs `headerFilter`) and the `onChange` handle. Perfect candidate for a component.

The `StyledFilter` is a filter component with only styling. A `SidebarFilter` can be made, which contains a `StyledFilter`, as well as the actual filtering logic and children.

This way, when composing the sidebar itself, adding a filter will be a one-liner:
```
<SidebarFilter filter={headerFilter} onChange={setHeaderFilter} />
<SidebarFilter filter={schemasFilter} onChange={setSchemaFilter} />
```

packages/insomnia-components/components/sidebar.js
```diff
@@ -140,204 +213,637 @@ const StyledPanel: React.ComponentType<{}> = styled(motion.div)`
...
+              transition={{ duration: 0.2, ease: 'easeInOut', delay: 0 }}>
+              <input
+                type="text"
+                value={serverFilter}
+                onChange={e => setServerFilter(e.target.value)}
+                placeholder="Filter..."
+              />
+            </StyledFilter>
+            {servers.map((server, index) => (
+              <React.Fragment key={index}>
```
develohpanda: Index as key is a [cause for concern]( https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318), we should definitely look to fix this. Some other instances of it in this components too.

ES Lint should have picked it up, [this rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md) may not be enabled.

packages/insomnia-components/components/sidebar.js
```diff
@@ -140,204 +213,637 @@ const StyledPanel: React.ComponentType<{}> = styled(motion.div)`
...
-            </motion.span>
-            <motion.span
-              initial={{ opacity: visible2 ? 0.6 : 0.4 }}
-              animate={{ opacity: visible2 ? 0.6 : 0.4 }}
-              transition={{ duration: 0.2, ease: 'easeInOut', delay: 0.1 }}>
-              <Tooltip message="Add" position="top">
-                <SvgIcon icon={IconEnum.plus} />
-              </Tooltip>
-            </motion.span>
+            <Dropdown renderButton={({ open }) => <SvgIcon icon={IconEnum.ellipsesCircle} />}>
```
develohpanda: I believe this syntax will result in this drop-down component and it's children updating on every render. The inline function will mean a new instance is sent every time. Definitely a candidate to declare the a button component outside the component, and consume it here. Also, looks like the `open` prop isn't being used so I have removed it. Eg:
```
const DropdownElipsis = () => <SvgIcon icon={IconEnum.ellipsesCircle} />;
```

```suggestion
            <Dropdown renderButton={DropdownElipsis}>
```

packages/insomnia-components/components/sidebar.js
```diff
@@ -140,204 +213,637 @@ const StyledPanel: React.ComponentType<{}> = styled(motion.div)`
...
+  // Info
+  const [infoSec, setInfoSec] = useState(false);
+  const toggleInfoSec = () => setInfoSec(!infoSec);
+
+  // Servers
+  const [serversSec, setServersSec] = useState(false);
+  const toggleServersSec = () => setServersSec(!serversSec);
+  const [serversFilter, setServersFilter] = useState(false);
+  const toggleServersFilter = () => setServersFilter(!serversFilter);
+  const [serverFilter, setServerFilter] = useState('');
```
develohpanda: I wonder if the state declarations might benefit from `useReducer` instead of `useState`. Easier to test, and less verbose.

develohpanda (changes requested):
Had a play around in the storybook and it looks amazing. I have added some notes regarding the implementation side of it. The rest are open for discussion, however we should address index as keys in _this_ PR.

packages/insomnia-components/components/sidebar.js
```diff
@@ -4,16 +4,23 @@ import { useState } from 'react';
...
 import styled from 'styled-components';
 import Tooltip from './tooltip';
+import Dropdown from './dropdown/dropdown';
+import DropdownItem from './dropdown/dropdown-item';
+import DropdownDivider from './dropdown/dropdown-divider';
 import SvgIcon, { IconEnum } from './svg-icon';
 
 type Props = {|
   className?: string,
+  jsonData: Object,
```
gschier: We're going to need some sort of event handlers here. I suggest an `onClick` to start that returns an array containing the path to the object that was clicked. 

```js
onClick: (specPath: Array<string | number>) => void,
```

For example, clicking on an operation (request) might call the `onClick` handler with `[ 'paths', '/foo', 'get' ]`

gschier (changes requested):
@develohpanda did a great job reviewing this! I just had one other comment about adding a click handler so we can hook up the click-to-scroll later.

packages/insomnia-components/components/sidebar.js
sonicyeti: This sounds great, will start breaking these down now. Originally the plan was one item type to rule them all but we switched it mid-way where we have things like methods grouped/tucked under paths & results. The above will be much easier to work with.

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -2,9 +2,11 @@
 import * as React from 'react';
 import autobind from 'autobind-decorator';
 import YAML from 'yaml';
+import SpecEditorSidebarItem from './spec-editor-sidebar-item';
+import { Sidebar } from 'insomnia-components';
 import type { ApiSpec } from '../../../models/api-spec';
 import { trackEvent } from '../../../common/analytics';
-import SpecEditorSidebarItem from './spec-editor-sidebar-item';
+const jsyaml = require('js-yaml');
```
develohpanda: Interesting, is this not able to use ES6 imports?

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -36,9 +38,14 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
     handleSetSelection(pos.start.col - 1, pos.end.col - 1, pos.start.line - 1, pos.end.line - 1);
   }
 
+  componentDidMount() {
+    // const { apiSpec } = this.props;
+  }
+
```
develohpanda: Can be removed

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -118,13 +125,22 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
     const { apiSpec } = this.props;
     const specCst = YAML.parseCST(apiSpec.contents);
 
+    let specJSON = {};
+
+    if (apiSpec.contentType === 'yaml') {
+      specJSON = YAML.parse(apiSpec.contents);
+    } else {
+      specJSON = jsyaml.safeLoad(apiSpec.contents);
+    }
```
develohpanda: Curious about this - could you please elaborate on why we need jsyaml given we only call it if the content type is _not_ yaml?

packages/insomnia-app/app/ui/css/components/spec-editor-sidebar.less
```diff
@@ -1,4 +1,5 @@
 .spec-editor-sidebar {
+  /*
```
develohpanda: Should the commented lines be removed altogether?

packages/insomnia-components/components/sidebar/sidebar-filter.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+`;
+
+const SidebarFilter = ({ filter, children, onChange }: Props) => (
+  <StyledFilter
+    initial={{ height: filter ? '100%' : '0px' }}
+    animate={{ height: filter ? '100%' : '0px' }}>
+    <input type="text" placeholder="Filter..." onChange={onChange} />
+    {children}
```
develohpanda: Do any consumers of `SidebarFilter` currently set `children`? If not, we should remove this prop.

packages/insomnia-components/components/sidebar/sidebar-header.js
```diff
@@ -0,0 +1,66 @@
...
+import styled from 'styled-components';
+import { motion } from 'framer-motion';
+import SvgIcon, { IconEnum } from '../svg-icon';
+
+type Props = {
+  headerTitle: string,
+  toggleSection: Function,
+  toggleFilter?: Function,
+  section: boolean,
+  children?: React.Node,
```
develohpanda: - Given we are in a header component, the `headerTitle` prop can simply be `title`
- Is it possible to type the `toggleSection` and `toggleFilter` props more explicitly than `Function`? How about something like `toggleSection: () => void`
- A prop simply named `section` doesn't convey what it represents - I suggest something more explicit like bodyVisible/sectionVisible/sectionExpanded etc.

packages/insomnia-components/components/sidebar/sidebar.js
```diff
@@ -0,0 +1,224 @@
...
+const DropdownEllipsis = () => <SvgIcon icon={IconEnum.ellipsesCircle} />;
+// Section Expansion & Filtering
+const useToggle = (state, set) =>
+  React.useCallback(
+    e => {
+      e.stopPropagation();
+      set(!state);
+    },
+    [set, state],
+  );
```
develohpanda: Can be removed in favor of `useToggle` in `react-use`. All of the state declarations in this file can simplify, see usage in `SidebarSection`.

packages/insomnia-components/components/sidebar/sidebar.js
```diff
@@ -0,0 +1,224 @@
...
+    responses,
+    parameters,
+    headers,
+    schemas,
+    securitySchemes,
+  } = props.jsonData.components;
+  const paths = Object.entries(props.jsonData.paths);
+
+  return (
+    <StyledSidebar className="theme--sidebar">
```
develohpanda: Feels a little strange to use a styled component and also a theme class - what is the reason for using both?

packages/insomnia-components/components/sidebar/sidebar.js
```diff
@@ -0,0 +1,224 @@
+// @flow
```
develohpanda: Worth renaming this file to `index.js` given it's in the `sidebar` folder

package.json
```diff
@@ -68,5 +68,7 @@
     "prettier": "^1.16.4",
     "rimraf": "^2.6.3"
   },
-  "dependencies": {}
+  "dependencies": {
+    "react-use": "^15.3.2"
```
develohpanda: Doesn't need to be in the root `package.json`

packages/insomnia-components/components/sidebar/sidebar-header.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+`;
+
+const SidebarHeader = ({ headerTitle, toggleSection, toggleFilter, section, children }: Props) => (
+  <StyledHeader>
+    <h6 onClick={toggleSection}>{headerTitle}</h6>
+    <div>
+      {children}
+      {!children && (
```
develohpanda:
```suggestion
      {children || (
```

packages/insomnia-components/components/sidebar/sidebar.js
```diff
@@ -0,0 +1,224 @@
...
+import SidebarHeaders from './sidebar-headers';
+import SidebarSchemas from './sidebar-schemas';
+import SidebarSecurity from './sidebar-security';
+import Dropdown from '../dropdown/dropdown';
+import DropdownItem from '../dropdown/dropdown-item';
+import DropdownDivider from '../dropdown/dropdown-divider';
+import SvgIcon, { IconEnum } from '../svg-icon';
+
+type Props = {|
+  className?: string,
```
develohpanda: `className` prop is not used

packages/insomnia-components/components/sidebar/sidebar.js
```diff
@@ -0,0 +1,224 @@
...
+    font-size: var(--font-size-xs);
+    display: flex;
+    flex-grow: 1;
+    &:hover {
+      cursor: default;
+    }
+  }
+  h5 {
+    font-size: var(--font-size-sm);
+  }
```
develohpanda: None of these elements exist directly in this component; they exist in child components - I suggest these styles be moved closer to where they are used.

packages/insomnia-components/components/sidebar/sidebar-servers.js
```diff
@@ -0,0 +1,42 @@
...
+  servers: Array<any>,
+};
+
+let itemPath = [];
+const handleClick = items => {
+  itemPath.push('server');
+  itemPath.push.apply(itemPath, items);
+  console.log(itemPath);
+  itemPath = [];
+};
```
develohpanda: This on click handler, and the element it is linked to, is duplicated through each sidebar section component. Did you want to make a stab at extracting it to a component and reusing? 😄

packages/insomnia-components/components/sidebar/sidebar-servers.js
```diff
@@ -0,0 +1,42 @@
...
+// Implemented as a class component because of a caveat with render props
+// https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent
+export default class SidebarServers extends React.Component<Props> {
+  renderBody = (filter: string) => (
+    <div>
+      {this.props.servers.map(server => (
+        <React.Fragment key={server.url}>
+          {server.url.includes(filter) && (
```
develohpanda: It would be nice to have a placeholder in the empty state (eg. "no results"), which would require a small refactor here. The empty state could also be handled by the `SidebarSection`, so lets discuss some ideas on how to go about doing that.

Currently, this is what the empty state looks like. 

image

packages/insomnia-components/components/sidebar/sidebar-servers.js
```diff
@@ -0,0 +1,42 @@
...
+            <SidebarItem>
+              <div></div>
```
develohpanda: Rogue empty div? Same with the other components

packages/insomnia-components/components/sidebar/sidebar-security.js
```diff
@@ -0,0 +1,42 @@
...
+  itemPath = [];
+};
+// Implemented as a class component because of a caveat with render props
+// https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent
+export default class SidebarSecurity extends React.Component<Props> {
+  renderBody = (filter: string) => (
+    <div>
+      {Object.keys(this.props.security).map(scheme => (
+        <React.Fragment key={scheme}>
+          {scheme.toLowerCase().includes(filter) && (
```
develohpanda: Filtering should be case-insensitive (same with the other components)

image

image

image
(no results)

packages/insomnia-components/components/sidebar/sidebar-info.js
```diff
@@ -0,0 +1,60 @@
...
+let itemPath = [];
+const handleClick = items => {
+  itemPath.push('info');
+  itemPath.push.apply(itemPath, items);
+  console.log(itemPath);
+  itemPath = [];
+};
+// Implemented as a class component because of a caveat with render props
+// https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent
+export default class SidebarInfo extends React.Component<Props> {
```
develohpanda: Can be a function component, the code comment does not apply to this component

packages/insomnia-components/components/sidebar/sidebar-item.js
```diff
@@ -0,0 +1,43 @@
...
+    cursor: default;
+  }
+  &:last-child {
+    margin-bottom: var(--padding-md);
+  }
+`;
+
+const SidebarItem = ({ children }: Props) => <StyledItem>{children}</StyledItem>;
+
+export default SidebarItem;
```
develohpanda, line 84: This component is just a proxy into `StyledItem`, but that means places `SidebarItem` is consumed, there is an empty div to consume the first grid column.

In order to consume `SidebarItem`, one shouldn't need to consider how the `SidebarItem` is styled, so with that in mind, lets discus some ways to abstract that away. 😄 

Same goes for the `SidebarBlockItem`.

packages/insomnia-components/components/sidebar/sidebar-info.js
```diff
@@ -0,0 +1,60 @@
...
+          </SidebarBlockItem>
+        )}
+        {this.props.info.description && (
+          <SidebarBlockItem>
+            <strong>Description:</strong>
+            <span onClick={() => handleClick([this.props.info.description])}>
+              {this.props.info.description}
+            </span>
+          </SidebarBlockItem>
+        )}
```
develohpanda: Destructuring props, and extracting a reusable component in this file would make this really concise 😄

packages/insomnia-components/components/sidebar/sidebar-panel.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+import * as React from 'react';
+import styled from 'styled-components';
+import { motion } from 'framer-motion';
+
+type Props = {
+  parent: boolean,
```
develohpanda: What does `parent` represent? I suggest to name the prop according to what it controls in _this_ component. What this prop is linked to, is the responsibility of the consumer. 😄

packages/insomnia-components/components/sidebar/sidebar-requests.js
```diff
@@ -0,0 +1,75 @@
...
+  console.log(itemPath);
+  itemPath = [];
+};
+
+// Implemented as a class component because of a caveat with render props
+// https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent
+export default class SidebarRequests extends React.Component<Props> {
+  renderBody = (filter: string) => (
+    <div>
+      {Object.keys(this.props.requests).map((requestName, value) => (
```
develohpanda: This component can again be broken down, for readability if not for reuse. 👍

packages/insomnia-components/components/sidebar/sidebar-requests.js
```diff
@@ -0,0 +1,75 @@
...
+                    </div>
+                    <span onClick={() => handleClick([requestFormat])}>{requestFormat}</span>
+                  </SidebarItem>
+                  <SidebarBlockItem>
+                    {Object.keys(
+                      this.props.requests[requestName].content[requestFormat].examples,
+                    ).map(requestExample => (
+                      <div
+                        onClick={() => handleClick([requestExample])}
+                        className="method-post"
```
develohpanda: Should this className be hard-coded? Feels like it should be linked to the request method 🤔

packages/insomnia-components/components/sidebar/sidebar-paths.js
```diff
@@ -0,0 +1,57 @@
...
+};
+
+// Implemented as a class component because of a caveat with render props
+// https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent
+export default class SidebarPaths extends React.Component<Props> {
+  renderBody = (filter: string) => (
+    <div>
+      {this.props.paths.map(path => (
+        <React.Fragment key={path[0]}>
+          {path[0].includes(filter) && (
```
develohpanda: `path[0]` and `path[1]` are difficult to understand, given they are typed as `any`.

I suggest setting the required array items to named variables and consuming those.

`const [foo, bar] = path;`

develohpanda (changes requested):
Nice one, it's looking at lot more concise, and it's nice to see some separation of concerns where the logic for rendering the requests is scoped, and the logic for rendering servers is scoped, etc.

A little more to go, many of the review comments added to one component apply to other components as well. There is a fair bit of functionality we can still abstract away and extract into reusable components, for readability and making the consumption of each component easier.

packages/insomnia-components/components/sidebar/sidebar-panel.js
sonicyeti: Parent wrapper visibility, yeah this could be clearer for sure.

packages/insomnia-components/components/sidebar/sidebar.js
sonicyeti: This is to pickup our inline styling back on preview-head.html, where we essentially dump all CSS variables from core theming. Not an ideal solution and temporary until we centralize all shared styling variables.

packages/insomnia-components/components/sidebar/sidebar.js
sonicyeti: I was able to move them all except svg, which spans header, section, panels and items. Header & section are flat from a DOM perspective due to the header being the target to hide/show panels while persisting visibly.

packages/insomnia-components/components/sidebar/sidebar-servers.js
sonicyeti: Yeah, this was the start of the separate path reveal to the app's code editor wrapper class. This will not be present next round!

packages/insomnia-components/components/sidebar/sidebar-info.js
```diff
@@ -1,49 +1,37 @@
...
   const { title, description, version, license } = props.info;
   return (
-    <SidebarPanel parent={props.parent}>
-      {props.info.title && (
-        <SidebarBlockItem>
-          <br />
-          <strong>Title:</strong>
-          <span onClick={() => handleClick([title])}>{title}</span>
-        </SidebarBlockItem>
+    <SidebarPanel childrenVisible={props.childrenVisible}>
```
develohpanda, line 15: Since all of the children in `SidebarPanel` (title, description, version, license.name) are conditional, if _none_ of them are present an empty panel would render.

What happens/should happen in that case? Should we be rendering `null` (ie. nothing) or an empty `SidebarPanel`?

packages/insomnia-components/components/sidebar/sidebar-panel.js
```diff
@@ -4,8 +4,8 @@ import styled from 'styled-components';
 import { motion } from 'framer-motion';
 
 type Props = {
-  parent: boolean,
   children: React.Node,
+  childrenVisible: boolean,
```
develohpanda, line 8: 🔥

packages/insomnia-components/components/sidebar/sidebar.stories.js
```diff
@@ -22,7 +22,7 @@ const apiSpec = {
   },
   servers: [
     {
-      url: 'https://dev.konghq.com/v1',
+      url: 'https://dev.kongonghq.com/v1',
```
develohpanda, line 25: 😆

packages/insomnia-components/components/sidebar/sidebar-requests.js
```diff
@@ -18,56 +18,69 @@ const handleClick = items => {
...
+    );
+
+    if (!filteredValues.length) {
+      return null;
+    }
+
+    return (
+      <div>
+        {filteredValues.map(requestName => (
+          <React.Fragment key={requestName}>
```
develohpanda: I suggest destructing the request data up here.

```js
const {description, content} = this.props.requests[requestName];
...
<Tooltip message={description} ... />
...
{Object.keys(content).map(format => (
    ...
    {Object.keys(content[format].examples).map(requestExample => (...))}
    ...
  )
}
```

packages/insomnia-components/components/sidebar/sidebar-responses.js
```diff
@@ -20,27 +20,36 @@ const handleClick = items => {
...
+    if (!filteredValues.length) {
+      return null;
+    }
+
+    return (
+      <div>
+        {filteredValues.map(response => (
+          <React.Fragment key={response}>
+            {response.toLowerCase().includes(filter.toLocaleLowerCase()) && (
+              <SidebarItem>
```
develohpanda: This can be simplified now

```suggestion
        {filteredValues.map(response => (
          <SidebarItem key={response}>
```

packages/insomnia-components/components/sidebar/sidebar-schemas.js
```diff
@@ -19,23 +19,32 @@ const handleClick = items => {
...
+    if (!filteredValues.length) {
+      return null;
+    }
+
+    return (
+      <div>
+        {filteredValues.map(schema => (
+          <React.Fragment key={schema}>
+            {schema.toLowerCase().includes(filter.toLocaleLowerCase()) && (
+              <SidebarItem>
```
develohpanda:
```suggestion
        {filteredValues.map(schema => (
          <SidebarItem key={schema}>
```

packages/insomnia-components/components/sidebar/sidebar-paths.js
```diff
@@ -17,39 +17,53 @@ const handleClick = items => {
...
+
+    if (!filteredValues.length) {
+      return null;
+    }
+
+    return (
+      <div>
+        {filteredValues.map(([route, method]) => (
+          <React.Fragment key={route}>
+            {route.includes(filter.toLocaleLowerCase()) && (
```
develohpanda: This check should be in the filter above on line 29 and not here, I think?

packages/insomnia-components/components/sidebar/sidebar-section.js
```diff
@@ -35,9 +40,11 @@ const SidebarSection = ({ title, renderBody }: SectionProps) => {
         toggleSection={toggleBodyVisible}
         toggleFilter={toggleFilterVisible}
       />
-      <SidebarPanel parent={bodyVisible}>
+      <SidebarPanel childrenVisible={bodyVisible}>
         <SidebarFilter filter={filterVisible} onChange={handleFilterChange} />
-        {renderBody(filterValue)}
+        {renderBody(filterValue) || (
+          <StyledNoResults>No results found for "{filterValue}"...</StyledNoResults>
```
develohpanda, line 52: 👍

develohpanda (approved):
Couple of things worth addressing, either in this PR or a separate one, otherwise good to go. 👍

packages/insomnia-components/components/sidebar/sidebar-info.js
sonicyeti: Great point, I'm going to back some overarching checks for top level sections into sidebar itself to we don't have empty panels.

packages/insomnia-components/components/sidebar/sidebar-requests.js
sonicyeti: So much cleaner

packages/insomnia-components/components/sidebar/sidebar-section.js
```diff
@@ -32,6 +33,11 @@ const SidebarSection = ({ title, renderBody }: SectionProps) => {
     setFilterValue(e.target.value);
   }, []);
 
+  useLayoutEffect(() => {
+    toggleFilterVisible(false);
+    setFilterValue('');
+  }, [bodyVisible, toggleFilterVisible]);
```
develohpanda, line 39: Nice! I had to research about the difference between `useEffect` and `useLayoutEffect`, [TIL](https://daveceddia.com/useeffect-vs-uselayouteffect/)! 🤗

packages/insomnia-components/components/sidebar/sidebar-filter.js
```diff
@@ -35,12 +36,22 @@ const StyledFilter: React.ComponentType<{}> = styled(motion.div)`
...
 
-const SidebarFilter = ({ filter, onChange }: Props) => (
-  <StyledFilter
-    initial={{ height: filter ? '100%' : '0px' }}
-    animate={{ height: filter ? '100%' : '0px' }}>
-    <input type="text" placeholder="Filter..." onChange={onChange} />
-  </StyledFilter>
-);
+const SidebarFilter = ({ filter, onChange }: Props) => {
+  const filterField = React.createRef();
```
develohpanda, line 40: There is a `useRef` hook to use here in place of `React.createRef()`. Although I don't know the difference between them, would favor `useRef` to stay with hooks.

It would also be nice to achieve this without using a ref altogether, but I'm not totally sure what the solution might be right now (maintaining the value as state), so not going to suggest removing the ref yet. 😄

develohpanda (approved):
Nice one 👍

gschier (approved):

PR #2329, opened 2020-06-29
POC to get Insomnia networking available outside of app

gschier:


packages/insomnia-testing/package.json
```diff
@@ -28,7 +28,21 @@
...
+    "insomnia-plugin-cookie-jar": "^2.2.10",
+    "insomnia-plugin-core-themes": "^2.2.10",
+    "insomnia-plugin-file": "^2.2.10",
+    "insomnia-plugin-hash": "^2.2.10",
+    "insomnia-plugin-jsonpath": "^2.2.10",
+    "insomnia-plugin-now": "^2.2.10",
+    "insomnia-plugin-prompt": "^2.2.10",
+    "insomnia-plugin-request": "^2.2.10",
+    "insomnia-plugin-response": "^2.2.10",
+    "insomnia-plugin-uuid": "^2.2.10",
```
develohpanda: 🤔 should all of these be here?

packages/insomnia-testing/webpack.config.js
```diff
@@ -22,11 +26,23 @@ module.exports = {
...
+    'fsevents',
+  ],
+  resolve: {
+    alias: {
+      electron: path.resolve(path.join(__dirname, './lib/electron')),
+    },
   },
+  plugins: [
+    new webpack.DefinePlugin({
+      'process.env.APP_ID': JSON.stringify('com.insomnia.app'),
```
develohpanda: Should the APP_ID be designer? (Does this matter for the purposes of the lib?)

develohpanda (commented):
Good thing with this method is we don't need to test networking in the app since no code has been moved. Only exposed in a different way! Nice one 💯

packages/insomnia-app/app/common/analytics.js
```diff
@@ -3,7 +3,7 @@ import { buildQueryStringFromParams, joinUrlAndQueryString } from 'insomnia-url'
 import * as electron from 'electron';
 import * as models from '../models/index';
 import * as db from '../common/database';
-import uuid from 'uuid';
+import * as uuid from 'uuid';
```
gschier, line 6: The UUID module default export wasn't working for some reason in the CLI bundle, so changing all the imports to use `*` notation is a workaround.

packages/insomnia-app/package.json
```diff
@@ -63,15 +64,13 @@
     "graphql",
     "insomnia-components",
     "insomnia-cookies",
-    "insomnia-importers",
```
gschier, line 66: Removed these because they also didn't seem to work inside the bundle.

packages/insomnia-testing/src/run/insomnia.js
```diff
@@ -59,28 +62,28 @@ export default class Insomnia {
 
   /**
    *
-   * @param reqOrId - raw request object or an ID to reference a request
+   * @param reqId - request ID to send. Specifying nothing will send the active request
```
gschier, line 65: Realized that `reqId` was never being used as an object, so removed that ability for now.

packages/insomnia-cli/package.json
```diff
@@ -28,18 +28,36 @@
...
+    "webpack-node-externals": "^1.7.2"
   },
   "dependencies": {
     "commander": "^5.1.0",
+    "insomnia-send-request": "^2.2.10",
     "insomnia-testing": "^2.2.9",
     "openapi-2-kong": "^2.2.10",
-    "yaml": "^1.10.0"
+    "yaml": "^1.10.0",
+    "insomnia-plugin-base64": "^2.2.10",
```
gschier, line 49: Plugins need to be here because the plugin system adds core plugins with `global.require('plugin-name')`. I originally tried putting them in `insomnia-send-request/package.json` but the plugin system wasn't able to require them from the nested `node_modules/insomnia-send-request/node_modules/insomnia-plugin-*`.

packages/insomnia-testing/webpack.config.js
gschier: It doesn't matter but I will change it anyway.

packages/insomnia-app/app/common/send-request.js
```diff
@@ -0,0 +1,32 @@
...
+import { send } from '../network/network';
+import { getBodyBuffer } from '../models/response';
+
+export function getSendRequestCallback(environmentId, memDB) {
+  return async function sendRequest(requestId) {
+    // Initialize the DB in-memory and fill it with data
+    await db.init(modelTypes(), { inMemoryOnly: true });
+    for (const type of Object.keys(memDB)) {
+      for (const doc of Object.values(memDB[type])) {
+        await db.insert(doc);
```
develohpanda: Just an observation, no need to change right now. 

Feels like doing some maps to return an array of promises, and then Promise.all() might be good here. Also potentially moving it one level  up, and out of the sendRequest function? In order to avoid initialising a new db with _every_ request.

packages/insomnia-cli/package.json
```diff
@@ -28,19 +28,37 @@
     "@babel/cli": "^7.10.1",
```
develohpanda: @babel/cli and some of the new babel dependencies no longer need to exist since we've switched to webpack, yay!

packages/insomnia-cli/src/commands/run-tests.js
```diff
@@ -34,6 +36,13 @@ export async function runInsomniaTests(options: RunTestsOptions): Promise<void>
 
   const { reporter, bail, keepFile } = options;
 
+  const workingDir = options.workingDir || '.';
+
+  const db = await gitDataDirDb({
```
develohpanda: mem-db no longer exists, if you have a look at one of the other commands it has a loadDb() function to use here. Copy-paste 😊 (or leave it for me to sort out tomorrow, might be quicker)

packages/insomnia-send-request/package.json
```diff
@@ -0,0 +1,46 @@
+{
```
develohpanda, line 1: Also need to add a gitignore to ignore the dist directory. Bundled files have been added to git 😊

packages/insomnia-testing/src/run/insomnia.js
```diff
@@ -25,9 +25,11 @@ export type Response = {
   },
 };
 
+type SendRequestCallback = (requestId: string) => Promise<$Shape<Response>>;
+
 export type InsomniaOptions = {
   requests?: Array<$Shape<Request>>,
-  sendRequest?: (requestId: string) => Promise<$Shape<Response>>,
+  sendRequest?: SendRequestCallback,
```
develohpanda, line 32: 😍

develohpanda (approved):
Looks good! Epic 💯

PR #2330, opened 2020-06-30
Prompt for API Spec and support NeDB database

develohpanda:
- Prompt for a spec if no identifier provided
- Try load the git data db, then try load the designer NeDB
- New `--app-data-dir` global option (I believe linux requires this?)


packages/insomnia-cli/src/db/index.js
```diff
@@ -0,0 +1,59 @@
...
+  return {
+    ApiSpec: [],
+    Environment: [],
+    Request: [],
+    RequestGroup: [],
+    Workspace: [],
+  };
+}
+
+export type DbAdapter = (
```
gschier, line 25: 😍

gschier (approved):
Beautiful! Loving the new DB adapter interface

PR #2333, opened 2020-06-30
Add support for snippets in autocomplete

gschier:
This PR enables the ability to add snippets to the autocomplete, similar how we can pass constants.

```js
<CodeEditor
  // ...
  getAutocompleteSnippets={() => [{
    name: 'Snippet', 
    displayValue: '', 
    value: async () => 'Can do anything here'
  }]}
/>
```

image

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
@@ -511,6 +511,7 @@ class CodeEditor extends React.Component {
       autoCloseBrackets,
       dynamicHeight,
       getAutocompleteConstants,
+      getAutocompleteSnippets,
       getRenderContext,
       hideGutters,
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -779,7 +781,7 @@ class CodeEditor extends React.Component {
     if (value.trim() === '') {
       this._codemirrorSmartSetOption('lint', false);
     } else {
-      this._codemirrorSmartSetOption('lint', true);
+      this._codemirrorSmartSetOption('lint', this.props.lintOptions || true);
```
gschier, line 784: This was a fix because `lintOptions` can be an object instead of `true`

develohpanda (approved):

PR #2334, opened 2020-06-30
Alternate send request callback for app to use

gschier:
This PR splits the `getSendRequestCallback` into two functions. One that accepts an in-memory DB, and one that doesn't. The one that doesn't, will be useful for the app code since it will not need an in-memory DB.

### DIFF
--- a/packages/insomnia-app/app/common/send-request.js
+++ b/packages/insomnia-app/app/common/send-request.js
@@ -3,8 +3,8 @@ import { types as modelTypes } from '../models';
 import { send } from '../network/network';
 import { getBodyBuffer } from '../models/response';
 
-export async function getSendRequestCallback(environmentId, memDB) {
-  // Initialize the DB in-memory and fill it with data
+export async function getSendRequestCallbackMemDb(environmentId, memDB) {
+  // Initialize the DB in-memory and fill it with data if we're given one
   await db.init(modelTypes(), { inMemoryOnly: true }, true);
 
   const docs = [];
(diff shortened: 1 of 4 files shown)


packages/insomnia-cli/flow-typed/insomnia-send-request.js
```diff
@@ -2,7 +2,8 @@
 
 declare module 'insomnia-send-request' {
   declare module.exports: {
-    getSendRequestCallback: (
+    getSendRequestCallback: ( environmentId: string ) => Promise<Object>,
+    getSendRequestCallbackMemDb: (
       environmentId: string,
       memDb: { [string]: Array<Object> },
     ) => Promise<Object>,
```
develohpanda, line 9: This is typed in the consumer, but not in the implementation (in the same code-base). Really feels backwards, but more a conversation about typing across the code-base rather than this particular case.

packages/insomnia-app/app/common/send-request.js
```diff
@@ -3,8 +3,8 @@ import { types as modelTypes } from '../models';
 import { send } from '../network/network';
 import { getBodyBuffer } from '../models/response';
 
-export async function getSendRequestCallback(environmentId, memDB) {
-  // Initialize the DB in-memory and fill it with data
+export async function getSendRequestCallbackMemDb(environmentId, memDB) {
```
develohpanda, line 6: Curious about decision process -  why not make the db argument optional instead of creating a new function? There is clear separation between what should and shouldn't happen if a db is defined.

develohpanda (approved):
Looks good 👍

packages/insomnia-cli/flow-typed/insomnia-send-request.js
gschier: Ya, will be nice to have unified types in the future. I had this untyped at first but thought this was at least better than nothing.

packages/insomnia-app/app/common/send-request.js
gschier: It's harder to screw it up if you have two functions, imo. As a consumer of this API, it will likely be unclear that `memDB` is not required unless I looked at the code or docs. In this case, the two functions make it easier to discover.

gschier: In other words, I prefer multiple functions over optional args, unless they are more for configuration. Initializing a DB seems like a large enough difference to have two clearly-separate functions.

PR #2335, opened 2020-06-30
Swap out switch for new activity toggle component

gschier:
This PR swaps out the old activity toggle for the new `<ActivityToggle>` component, which will support three items for unit testing in the future.

This PR also removes the old `<ActivityBar>` (left side-bar) component, which was no longer in use.


packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -294,32 +297,46 @@ class Wrapper extends React.PureComponent<WrapperProps, State> {
     return null;
   }
 
-  async _handleWorkspaceActivityChange(workspaceId: string, activeActivity: GlobalActivity) {
-    const { activity: updatedActivity } = this.props.handleSetActiveActivity(activeActivity);
-    await models.workspaceMeta.updateByParentId(workspaceId, { activeActivity: updatedActivity });
-  }
+  async _handleWorkspaceActivityChange(workspaceId: string, nextActivity: GlobalActivity) {
```
gschier, line 300: Moved all activity switching into a single method and deleted others

packages/insomnia-app/app/ui/components/activity-toggle.js
```diff
@@ -0,0 +1,40 @@
...
+import React from 'react';
+import { MultiSwitch } from 'insomnia-components';
+import type { GlobalActivity } from '../../common/constants';
+import { ACTIVITY_DEBUG, ACTIVITY_SPEC, ACTIVITY_UNIT_TEST } from '../../common/constants';
+import type { Settings } from '../../models/settings';
+import type { Workspace } from '../../models/workspace';
+
+type Props = {
+  activity: GlobalActivity,
+  handleActivityChange: (workspaceId: string, activity: GlobalActivity) => any,
```
develohpanda: Would prefer to not type this as `any`. Should it be `Promise<void>`, in which case we should also await the promise?

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -273,8 +273,13 @@ class WrapperHome extends React.PureComponent<Props, State> {
   async _handleSetActiveWorkspace(id: string, defaultActivity: GlobalActivity) {
     const { handleSetActiveWorkspace, handleSetActiveActivity } = this.props.wrapperProps;
 
-    const selectedWorkspaceMeta = await models.workspaceMeta.getOrCreateByParentId(id);
-    handleSetActiveActivity(selectedWorkspaceMeta.activeActivity || defaultActivity);
+    const { activeActivity } = await models.workspaceMeta.getOrCreateByParentId(id);
+
+    if (!activeActivity || activeActivity === ACTIVITY_HOME) {
```
develohpanda, line 278: What is the reason for the `activeActivity === ACTIVITY_HOME` check? That's different to the previous implementation.

packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -294,32 +297,46 @@ class Wrapper extends React.PureComponent<WrapperProps, State> {
...
+      await models.workspaceMeta.updateByParentId(workspaceId, { activeActivity: nextActivity });
+    }
 
-  async _handleSetDebugActivity(apiSpec: ApiSpec): Promise<void> {
-    const workspaceId = apiSpec.parentId;
-    await this._handleWorkspaceActivityChange(workspaceId, ACTIVITY_DEBUG);
+    if (activity !== ACTIVITY_SPEC) {
+      handleSetActiveActivity(nextActivity);
+      return;
+    }
```
develohpanda, line 311: Does this mean that if the current activity is `ACTIVITY_SPEC`, you can't navigate away from it? 🤔 I must be understanding this wrong

packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -294,32 +297,46 @@ class Wrapper extends React.PureComponent<WrapperProps, State> {
...
+    const results = await spectral.run(activeApiSpec.contents);
+    if (activeApiSpec.contents && results && results.length) {
+      showModal(AlertModal, {
+        title: 'Error Generating Configuration',
+        message:
+          'Some requests may not be available due to errors found in the ' +
+          'specification. We recommend fixing errors before proceeding. 🤗',
+        okLabel: 'Proceed',
+        addCancel: true,
+        onConfirm: async () => {
```
develohpanda: Doesn't need to be an async callback (I think?)

packages/insomnia-app/app/ui/components/wrapper.js
develohpanda, line 300: There is now functionality in this function (since it is smarter about what to do), which would benefit from some unit tests to make sure it does what it should

packages/insomnia-app/app/common/constants.js
```diff
@@ -153,6 +151,14 @@ export const DEFAULT_PANE_WIDTH = 0.5;
 export const DEFAULT_PANE_HEIGHT = 0.5;
 export const DEFAULT_SIDEBAR_WIDTH = 19;
 
+// Activities
+export type GlobalActivity = 'spec' | 'debug' | 'monitor' | 'home';
+export const ACTIVITY_SPEC: GlobalActivity = 'spec';
+export const ACTIVITY_DEBUG: GlobalActivity = 'debug';
+export const ACTIVITY_UNIT_TEST: GlobalActivity = 'unittest';
+export const ACTIVITY_HOME: GlobalActivity = 'home';
+export const ACTIVITY_INSOMNIA: GlobalActivity = 'insomnia';
```
develohpanda, line 160: Good candidate for an enum (don't need to in this PR)

develohpanda (changes requested):
Overall in favor of this refactor, a couple of comments/questions about the functionality. 🤗

packages/insomnia-app/app/ui/components/wrapper-home.js
gschier: I got into a state where the `home` activity was stored on the workspace meta so it would try to go home when it was already there. Not sure of a better way to prevent that than doing this.

packages/insomnia-app/app/ui/components/wrapper.js
gschier: No, `handleSetActiveActivity` is called further down as well.

PR #2336, opened 2020-06-30
Rename emptyValue props to blankValue

gschier:
It turns out `emptyValue` is a reserved thing in React, so this PR renames usages of it to `blankValue` instead.

Here's the error from React that prompted this change.

image


develohpanda (approved):
TIL 💯

PR #2337, opened 2020-07-01
Unit test improvements

### LINKED ISSUE #2179: Unit Testing

gschier:
- [x] Add confirm dialog on test deletion
- [x] Catch case when tests throw and show error
- [x] Add header to unit test panes
- [x] Support editing test suite names
- [x] Some other minor tweaks

image


packages/insomnia-app/app/ui/components/base/editable.js
```diff
@@ -101,7 +101,6 @@ class Editable extends PureComponent {
         title: singleClick ? 'Click to edit' : 'Double click to edit',
         onClick: this._handleSingleClickEditStart,
         onDoubleClick: this._handleEditStart,
-        blankValue,
```
gschier, line 104: Turned out the `emptyValue` PR I did yesterday wasn't actually necessary. It was just complaining about passing an attribute it didn't recognize into `<input>`

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -165,7 +167,18 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
   }
 
   async _handleDeleteTest(unitTest: UnitTest): Promise<void> {
-    await models.unitTest.remove(unitTest);
+    showAlert({
```
develohpanda, line 170: Looking through the code path, `showAlert` returns a promise, which we should return. And since we return the promise, no need to mark `_handleDeleteTest` as `async`.

develohpanda (approved):
Looking good!

A note about the cog drop-down for each test. Running an individual test is currently a two-click process, and ideally it would be one-click (given we have a UI). So I propose two options:
1. Remove the cog, and replace with two buttons (run, delete)
2. Keep the cog, but add a keyboard shortcut to run the "current" test

We also have an awesome opportunity to show test results (tick or cross) anywhere the test name appears (sidebar, editor pane), and, for example, clicking on a test result on the results pane scrolling to the test. This is awesome as MVP though.

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
gschier: The alert modal promise is actually quite a hack if you take a look at the code.

Essentially, it was originally written without a callback method. Instead, the promise would be resolved if the user clicked OK. If they cancelled, the promise would just never be called. I don't think there's actually any code that uses the promise method anymore so we should probably deprecate it. 

So, the reason it's not awaited there is because it's really not obvious that it may never get past that line.

develohpanda: Oh! Noted, thanks for the explanation 😄

PR #2344, opened 2020-07-04
fix: prettify graphql based on user indentation size

### LINKED ISSUE #2301: "Prettify GraphQL" Incorect Indentations
**Describe the bug**
In my preferences, I have "Indent with tabs" enabled and my "Editor Indent Size" is set to 4. However, whenever I click "Prettify GraphQL" in a GraphQL request body, it reformats my request to 2 spaces.

**To Reproduce**
Steps to reproduce the behaviour:
1. Enabled: "Indent with tabs"
1. Set "Editor Indent Size" to 4
1. Entered GraphQL request
1. Clicked "Prettify GraphQL"

**Expected behaviour**
Reformat my request with tabs of an indentation size of 4.

 - OS: Ubuntu 20.0.4
 - Installation Method: deb from Website
 - App Version v2020.2.2

arjunkesava:
Closes #2301 

By default, graphql comes with a default 2 space indentation.
If the user had a different tab space setting, then we should apply the indentation


packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.js
```diff
@@ -360,10 +360,29 @@ class GraphQLEditor extends React.PureComponent<Props, State> {
...
+      if (indentSize !== 2) {
+        const indentQuery = this._applyIndentation(prettifyQuery, indentSize);
+        return indentQuery;
+      }
+    }
+    return '';
+  }
+
+  _applyIndentation(query: string, indentSize: number): string {
+    const indentationSpace = Array(indentSize).join(' ');
```
develohpanda, line 378: This means the indent doesn't respect the `indent with tabs` setting and always applies space instead of tabs.

image

develohpanda (changes requested):
Thanks for the PR! I cloned locally and did some testing, and unfortunately it's not quite working as expected.

- [ ] Indent is only applied to the first level of nesting, not all levels
- [ ] If indent is set to 2, it clears the textbox
- [ ] Indent does not respect the `indent with tab` setting

🎉

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.js
```diff
@@ -360,10 +360,29 @@ class GraphQLEditor extends React.PureComponent<Props, State> {
...
+    return query.replace(/\n\s\s/g, `\n${indentationSpace}`);
+  }
```
gschier, line 380: This indentation strategy seems like there could be some edge-cases. For example, what happens if the user has `\n  ` in the query itself?

```graphql
mutation {
  addUser(bio: "this could have a newline \\n  some more stuff") {
    ...
  }
}
```

I think, ideally, we swap out for a different formatting strategy rather than try to patch it after the fact. There's also another downside of the `print(parse(` strategy we're using now, which is that any comments are discarded. Switching the formatter we use would also fix this problem.

gschier:
Thanks for the effort on this, but we're going to close this one as there isn't a reliable way to implement this using the existing formatter.

PR #2351, opened 2020-07-06
Stub console log during db initialization

develohpanda:
insomnia-app db initialization prints to console, and this clutters when running in the CLI. This PR injects the `console.log` function as a parameter, allowing you to supply a no-op function instead of `console.log`. This feels like a cleaner solution than using a "enableConsoleLog" boolean flag.


gschier (approved):
It would be nice to add a more robust/configurable logging system in the future, but this is great for now! 👍 😄

PR #2352, opened 2020-07-06
Get app data dir from process.env, update webpack

develohpanda:
Slight bundling differences between production and development - namely the app data directory to use when bundling the app for production.


packages/insomnia-inso/src/db/index.js
```diff
@@ -47,9 +47,9 @@ export async function loadDb({
 
   // try load from nedb
   if (!db) {
-    // TODO: Note, unit tests will also try to access the Insomnia Designer app data directory. We should configure this depending on development or production.
     db = await neDbAdapter(
-      appDataDir || envPaths('Insomnia Designer', { suffix: '' }).data,
+      appDataDir ||
+        envPaths(process.env.DEFAULT_APP_DATA_DIR || 'insomnia-app', { suffix: '' }).data,
```
gschier: Should we instead throw an error here instead of providing a fallback value for the env? There's no reason it shouldn't exist now, and it seems a bit weird to have a fallback _and_ set the same value in `webpack.config.development.js`.

packages/insomnia-inso/webpack/webpack.config.development.js
```diff
@@ -0,0 +1,13 @@
+const webpack = require('webpack');
+const {merge} = require('webpack-merge');
+const base = require('./webpack.config.base');
+
+module.exports = merge(base, {
```
gschier, line 5: Wish I knew about `webpack-merge` sooner 💯

gschier (approved):
Just a small comment but it looks good!

packages/insomnia-inso/package.json
```diff
@@ -12,11 +12,11 @@
...
     "test:snapshots": "npm run build && jest -u",
-    "start": "npm run build:watch",
-    "build": "webpack --config webpack.config.js --display errors-only",
-    "build:watch": "npm run build -- --watch",
+    "build": "webpack --config webpack/webpack.config.development.js --display errors-only",
+    "build:production": "webpack --config webpack/webpack.config.production.js --display errors-only",
+    "start": "npm run build -- --watch",
     "bootstrap": "npm run build",
-    "prepublish": "npm run build"
+    "prepare": "npm run build:production"
```
gschier, line 19: Woah, did not know about `prepare`! Amazing

packages/insomnia-inso/src/db/index.js
develohpanda: Oh ya, good point. I agree, we should throw an error here. Will update!

packages/insomnia-inso/package.json
develohpanda: `prepublish` has been deprecated in favor of `prepare` 😄 https://docs.npmjs.com/misc/scripts#deprecation-note

packages/insomnia-inso/webpack/webpack.config.development.js
develohpanda: Was going to add it to the app when I was working through the bundling/react hooks PR 😆

PR #2359, opened 2020-07-08
Resuable sidebar section

develohpanda:
This PR introduces a `SidebarSection` that contains the `SidebarHeader`, the `SidebarFilter`, and all state management with handling visibility of a section, filter, and filter value. This is on the premise that the only thing unique for each section is the actual section body.

I only did this for the servers, other sections would follow a similar pattern.

CI failure is with the old component in `components/sidebar.js`


packages/insomnia-components/components/sidebar/sidebar-section.js
```diff
@@ -0,0 +1,51 @@
...
+`;
+
+const SidebarSection = ({ title, renderBody }: SectionProps) => {
+  const [bodyVisible, toggleBodyVisible] = useToggle(false);
+  const [filterVisible, toggleFilterVisible] = useToggle(false);
+  const [filterValue, setFilterValue] = React.useState('');
+
+  const handleFilterChange = React.useCallback((e: SyntheticInputEvent<HTMLInputElement>) => {
+    setFilterValue(e.target.value);
+  }, []);
```
develohpanda, line 29: image[ref](https://reactjs.org/docs/hooks-reference.html#usestate)

Yay, the `setState` function does not need to be in the dependency array. This will only ever run once. 🎉

PR #2375, opened 2020-07-10
Add `inso export spec` command

### LINKED ISSUE #2368: CLI command for exporting specification
Currently the specifications are stored under the `.insomnia` folder and are wrapped with insomnia meta information. We need a way to export the raw specification for use with third-party tools like `portal-cli` and `kong mock plugin`.

Expected usage:

`inso export spec`

Expected features

- Raw output
- Output to file

develohpanda:
`inso export spec [-o, --output] [identifier]`

### Prompt if no identifier

image

### Print to console

image

### Output to file, file in directory, or a directory

image

TODO:
- [x] Unit tests

Mocking of the db isn't easily possible for unit tests right now, hoping to improve that in the future.

Closes #2368


DMarby:
I really like the thorough testing 🎉

packages/insomnia-inso/src/__tests__/write-file.test.js
```diff
@@ -0,0 +1,81 @@
...
+  sync: jest.fn().mockResolvedValue(),
+}));
+
+jest.mock('fs', () => ({
+  promises: {
+    writeFile: jest.fn().mockResolvedValue(),
+  },
+}));
+
+describe('writeFileWithCliOptions', () => {
```
DMarby, line 17: Perhaps it could be nice to add some tests for what happens when errors are encountered here as well?

packages/insomnia-inso/src/write-file.js
```diff
@@ -0,0 +1,21 @@
...
+  fallbackFileName: string,
+  workingDir?: string,
+): Promise<string> {
+  // if output ends with a filename, use it, otherwise use the fallback filename
+  const fullPath = path.join(
+    workingDir || '.',
+    output,
+    path.extname(output) ? '' : fallbackFileName,
+  );
+  await mkdirp.sync(path.dirname(fullPath));
```
DMarby: Do we need/should we have any error handling here, or will exceptions get propagated up the stack in a way that's handled gracefully/exposed to the user in a good way?

packages/insomnia-inso/src/write-file.js
```diff
@@ -0,0 +1,21 @@
...
+import mkdirp from 'mkdirp';
+import fs from 'fs';
+
+export async function writeFileWithCliOptions(
+  output: string,
+  contents: string,
+  fallbackFileName: string,
+  workingDir?: string,
+): Promise<string> {
+  // if output ends with a filename, use it, otherwise use the fallback filename
```
DMarby: As far as I can tell, the code below uses the fallback if the output doesn't end with a file `extension` not a filename?

I think it probably makes sense to allow people to export to files without a file extension if they want to?
In terms of UX, when specifying an absolute path, I think I would expect a program to write to the exact path I specify, without modifications.

DMarby (changes requested):

packages/insomnia-inso/src/write-file.js
develohpanda: Good note, I've removed the fallback for now and we can add it should people ask.

We'll need to rethink how exports work when we have multi-file support. In that case we would likely need to export multiple files to a directory - not needed right now though.

packages/insomnia-inso/src/write-file.js
develohpanda: Guessing we should have them, if filenames use illegal characters or the directory has no write permission (for example), these commands would fail.

Exceptions will get propagated up and handled by the root exception handler _but_ for this functionality we should catch errors sooner and present to the user. 👍

DMarby (approved):

PR #2394, opened 2020-07-10
Fix autocomplete and make the code a bit better

### LINKED ISSUE #2378: Environment variable is not correctly selected after updating to v2020.3.0
**Describe the bug**
 When trying to enter an environment variable the variable has an error and it has to be manually updated. 

**To Reproduce**

1. Try to enter a variable in the URL
2. Press `(Control + Space bar)` (mac)
3. Press `Enter` or select from the dropdown list

**Expected behavior**
The variable to be added without need to manually fix it.

**Screenshots**
Screenshot 2020-07-10 11 25 33
Screenshot 2020-07-10 11 25 49

 - OS: macOS 10.15.4
 - Installation Method: update and also download and replace
 - App Version 2020.3.0

gschier:
Closes #2378 

This PR fixes variable autocompletion, which was broken with the addition of snippet autocomplete. The issue was that variables were being auto-completed to their value, not their name.

For example.

- For environment `{ "foo": "bar"}`
- Autocompleting `foo` would yield `{{ bar }}` instead of `{{ foo }}`

For now, I have made the autocomplete code a bit less "magic" and made sure to QA all autocompletion types manually (constants, variables, tags, snippets). After this, we should probably consolidate some of this logic and add tests.

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
@@ -184,10 +184,16 @@ CodeMirror.defineOption('environmentAutocomplete', null, (cm, options) => {
  * @returns {Promise.<{list: Array, from, to}>}
  */
 function hint(cm, options) {
-  const variablesToMatch = options.variables || [];
-  const constantsToMatch = options.constants || [];
-  const snippetsToMatch = options.snippets || [];
-  const tagsToMatch = options.tags || [];
+  // Add type to all things (except constants, which need to convert to an object)
+  const variablesToMatch = (options.variables || []).map(v => ({ ...v, type: TYPE_VARIABLE }));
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
```diff
@@ -372,12 +378,19 @@ function matchSegments(listOfThings, segment, type, limit = -1) {
...
-    } else if (t.value) {
+    if (t.type === TYPE_CONSTANT) {
+      defaultFill = t.value;
+    } else if (t.type === TYPE_SNIPPET) {
       defaultFill = t.value;
+    } else if (t.type === TYPE_VARIABLE) {
+      // Variables fill with variable name, not value (eg. {{ foo }})
+      // TODO: This is extremely confusing and does not make any sense so let's
+      //  refactor this to a single unified format for all types
+      defaultFill = t.name;
```
gschier, line 389: This is where the old code went wrong. It was using `t.name` instead of `t.value` for variables

Issue #1373, opened 2019-02-13
[Bug] Hangs on moving a request to a folder


mertvetsky:
- Insomnia Version: 6.32
- Operating System: Arch Linux 4.20.5, Xfce/Xfwm

## Details

On moving any request to any folder app fully hangs.
It may depends on electron 3.* bug: https://github.com/electron/electron/issues/14430

gschier:
That's really strange. It doesn't seem to happen on my end (Mac). Maybe an Electron update will fix it?

mertvetsky:
Yeah, this may solve problem. If needs, I can try build with next Electron, but if updating already in your nearest plans, I'll just waiting.

Btw, in other linux distros (like Ubuntu/Mint) problem doesn't appear.

gschier:
So I think the only way forward on this one is to wait for Electron to fix it. There might be a fix in 4.x already but there were some build issues caused by the upgrade and I didn't have the time to spend to fix it. I'll try and get an upgrade into the next release though.

nkkollaw:
@gschier : are you able to reproduce the problem?

gschier:
I haven't been able to reproduce it no. Perhaps it depends on something external to the application like mouse/monitor/etc settings?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

nkkollaw:
No idea, but I doubt it. The same thing happened in VS Code, and they fixed it (I guess by upgrading Electron). 

This is most probably the bug in Electron that it's causing the crash: https://github.com/electron/electron/issues/12820/

Unfortunately I have no idea how Electron apps work or I'd create a pull request. I'm just here to complain :-D

gschier:
😢

nkkollaw:
Mh..?

gschier:
Just shedding a tear because bugs like this are so difficult 🙃

nkkollaw:
HA!
Actually, it's even worse now. Now it's working :-/

I have no idea. Maybe it's related to memory usage..? For instance, if I'm low in memory it doesn't work..?

Baffling.

mertvetsky:
Seems to be fixed by updating xfce to 4.14 (with gtk3)

wdawson:
Sorry folks! This was incorrectly tagged as a feature request along with other bugs. I'm going back through and cleaning up, sorry for the inconvenience.

However, it looks like this could be resolved now? Let us know if we should reopen.

PR #2403, opened 2020-07-14
Upgrade Electron 

### LINKED ISSUE #2185: OAuth2 Implicit with Google broken
**Describe the bug**
When using the implicit OAuth2 flow for our API I can not authenticated with my Google Account anymore. The following message appears after submitting the credentials:

> This browser or app may not be secure.[ Learn more](https://support.google.com/accounts/answer/7675428)
> Try using a different browser. If you’re already using a supported browser, you can refresh your screen and try again to sign in.

**To Reproduce**
1. Use a request using oAuth2
1. Use Google as the identity provider for sign in

**Expected behavior**
I think doing some adjustments in the included "browser" of Insomnia could help so that Google is accepting it as a trusted browser again. Or implement [browser-based OAuth](https://developers.google.com/identity/protocols/oauth2/native-app).

**Screenshots**
image

 - OS: macOS 10.15.4
 - App Version v7.1.1

### LINKED ISSUE #2369: Upgrade Electron
Now that we're finally able to update `node-libcurl` (the main blocker for this task) we should be able to have a better chance at this. Note, there is a previous issue for the first attempt at upgrading Electron, which is now closed in favor of this: #1847

## Discoveries So Far

- `font-manager` does not build for Node 12 yet, however there is an [open PR](https://github.com/Kong/insomnia/issues/2369) that fixes it that will (hopefully) be merged soon. If not, apparently `font-scanner` also contains the fix which we were actually using up until a couple months ago
- `node-libcurl` only supports prebuild binaries for Electron <= 8 so I've submitted an issue to add the latest version (9) https://github.com/JCMais/node-libcurl/issues/238

gschier:
Closes #2369 
Closes #2185 
Related #1373

## To-Do

- [x] Bump Electron version to 9
- [x] Update NPM config (`.npmrc`, etc) to target builds for new Electron version
- [x] Identify and fix breaking Electron API changes (see below)
- [x] Switch from `font-manager` back to `font-scanner` as it was first to support Node 12
- [x] Ensure plugins installation still works
- [x] Refactor usage of Electron APIs that switched from callback-style to Promises in v6
- [x] Remove user-agent forcing for OAuth 2 login windows

## Things Electron Broke

- [x] Application launches and renders main app
  - Renders white window with error `module is not defined`
  - [x] Specify `nodeIntegration: true` when creating `BrowserWindow`
- [x] Use of `<webview>` to render HTML responses visually
  - `<webview>` doesn't render anything
  - [x] Enable webview tag with new `webviewTag: true` setting
  - rendering webview now crashes the app
    - ~~It looks like `<webview>` no longer supports Data URIs 😢~~
    - This is an open bug ([electron#20700](https://github.com/electron/electron/issues/20700)) regarding usage of `baseURLForDataURL` setting
    - [x] **Find a workaround for broken webview** 🚨
    - Moved back to how we did it before, which is to inject the `<base>` tag into `<head>`
- [x] ~~**Move native module access to `main` process** 🚨 (no longer allowed in renderer) [link](https://github.com/electron/electron/issues/18397)~~
  - Set `app.allowRendererProcessReuse = false` for now as workaround, but will need to do this for Electron 11 which will require it
- [x] Check if `net` module is being used for anything https://github.com/electron/electron/pull/17517
- [x] This seems alarming but is probably nothing: https://github.com/electron/electron/pull/18577
- [x] Auto-hiding of menubar setting and hotkey (alt) on Linux/Windows
- [x] Exec path for running `yarn-standalone.js` now includes `(Renderer)` in the file path
  - Parentheses need to be escaped

## Reference

- [Electron 4 Changelog](https://www.electronjs.org/releases/stable?version=4&page=6#release-notes-for-v400)
- [Electron 5 Changelog](https://www.electronjs.org/releases/stable?version=5&page=3#release-notes-for-v500)
- [Electron 6 Changelog](https://www.electronjs.org/releases/stable?version=6&page=6#release-notes-for-600)
- [Electron 7 Changelog](https://www.electronjs.org/releases/stable?version=7&page=5#release-notes-for-v700)
- [Electron 8 Changelog](https://www.electronjs.org/releases/stable?version=8&page=4#release-notes-for-v800)
- [Electron 9 Changelog](https://www.electronjs.org/releases/stable?version=9&page=2#release-notes-for-900)


.github/workflows/release-core.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: 10
+          node-version: 12
```
develohpanda: I suggest we do something like this, to ensure workflows and local dev all depend on the exact same version of node (patch as well), defined in `.nvmrc`. Could do the same for `.npmrc` as well.

https://github.com/actions/setup-node/issues/32#issuecomment-525791142

Would help with reducing the number of `package-lock.json` changes.

packages/insomnia-app/app/network/o-auth-2/misc.js
```diff
@@ -99,18 +99,8 @@ export function authorizeUserInWindow(
       _parseUrl(url);
     });
 
-    const options = {};
-
-    // Force user-agent for GitHub until we update Chromium version. Note, we don't do this for
```
develohpanda, line 104: 🎉

packages/insomnia-app/app/ui/components/base/file-input-button.js
```diff
@@ -65,15 +66,15 @@ class FileInputButton extends React.PureComponent<Props> {
...
 
-    remote.dialog.showOpenDialog(options, async paths => {
-      // Only change the file if a new file was selected
-      if (!paths || paths.length === 0) {
-        return;
-      }
+    const { canceled, filePaths } = await remote.dialog.showOpenDialog(options);
+
+    // Only change the file if a new file was selected
+    if (canceled && filePaths.length) {
```
develohpanda: Should this be the following?

```suggestion
    if (canceled || !filePaths.length) {
```

The two are not the same 🤔

packages/insomnia-app/app/ui/components/modals/request-create-modal.js
```diff
@@ -115,7 +115,7 @@ class RequestCreateModal extends PureComponent {
               <div className="form-control form-control--outlined">
                 <label>
                   Name
-                  <span class="txt-sm faint italic space-left">
+                  <span className="txt-sm faint italic space-left">
```
develohpanda, line 118: Woah! Sneaked in during a recent PR.

packages/insomnia-app/app/ui/components/request-url-bar.js
```diff
@@ -134,13 +134,12 @@ class RequestUrlBar extends React.PureComponent<Props, State> {
       properties: ['openDirectory'],
     };
 
-    remote.dialog.showOpenDialog(options, paths => {
-      if (!paths || paths.length === 0) {
-        return;
-      }
+    const { canceled, filePaths } = await remote.dialog.showOpenDialog(options);
+    if (canceled) {
```
develohpanda, line 138: Should we be checking for `filePaths.length` here as well, similar to a previous comment?

```suggestion
    if (canceled || !filePaths.length) {
```

packages/insomnia-app/app/ui/components/viewers/response-multipart.js
```diff
@@ -139,20 +139,19 @@ class ResponseMultipart extends React.PureComponent<Props, State> {
...
+    window.localStorage.setItem('insomnia.lastExportPath', path.dirname(filename));
 
-      // Save the file
-      fs.writeFile(outputPath, part.value, err => {
-        if (err) {
-          console.warn('Failed to save multipart to file', err);
-        }
-      });
+    // Save the file
+    fs.writeFile(filePath, part.value, err => {
```
develohpanda: Are we able to use the `fs.promises` API for this now?

packages/insomnia-app/app/ui/components/viewers/response-web-view.js
```diff
@@ -41,9 +41,19 @@ class ResponseWebView extends React.PureComponent<Props> {
     }
 
     const { body, contentType, url } = this.props;
-    webview.loadURL(`data:${contentType},${encodeURIComponent(body)}`, {
-      baseURLForDataURL: url,
-    });
+    const bodyWithBase = body.replace('<head>', `<head><base href="${url}">`);
+    webview.loadURL(`data:${contentType},${encodeURIComponent(bodyWithBase)}`);
+
+    // NOTE: We *should* be setting the base URL like this, but there is a bug
```
develohpanda:
```suggestion
    // NOTE: We *should* be setting the base URL by specifying the baseURLForDataURL option, but there is a bug
```

I had to read a few times to understand "this" was referring to the example at the end of the comment, not referring to the code above.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -611,18 +611,20 @@ class App extends PureComponent {
...
-      });
-    });
+    const options = {
+      title: 'Select Download Location',
+      buttonLabel: 'Send and Save',
+    };
+
+    const defaultPath = window.localStorage.getItem('insomnia.sendAndDownloadLocation');
+    if (defaultPath) {
+      // NOTE: An error will be thrown if defaultPath is supplied but not a String
```
develohpanda, line 621: Should we check for `typeof defaultPath === 'string'`, in that case?

packages/insomnia-app/scripts/build.js
```diff
@@ -50,8 +50,8 @@ module.exports.start = async function(forcedVersion = null) {
   console.log(`[build] npm: ${childProcess.spawnSync('npm', ['--version']).stdout}`.trim());
   console.log(`[build] node: ${childProcess.spawnSync('node', ['--version']).stdout}`.trim());
 
-  if (process.version.indexOf('v10.') !== 0) {
-    console.log('[build] Node v10.x.x is required to build');
+  if (process.version.indexOf('v12.') !== 0) {
+    console.log('[build] Node v12.x.x is required to build');
```
develohpanda, line 54: 👍 Nice, I would have missed this.

packages/insomnia-app/webpack/webpack.config.development.babel.js
```diff
@@ -25,10 +25,6 @@ module.exports = {
       },
     ],
   },
-  stats: {
-    // https://webpack.js.org/loaders/source-map-loader/#ignoring-warnings
-    warningsFilter: [/Failed to parse source map/],
-  },
```
develohpanda, line 31: Ahhh, `contentBase: '/'` is a fix for this warning suppression.

I remember facing these warnings, but the source-maps still loaded, so it was very confusing. Awesome to get to the bottom of it. 👍

develohpanda (changes requested):
Looking great 👍 Big effort!

Question regarding the file selection dialog responses, which now return a `{canceled, <result>}` object. Should we be validating the `result`? All appropriate uses check for the `canceled` property, but only some check for the `result`.

One really minor (but important) comment change suggested, otherwise good to go.

packages/insomnia-app/scripts/build.js
gschier: Hehe, hard to miss when it prevents the build from working 😆

packages/insomnia-app/app/ui/containers/app.js
gschier: LocalStorage always returns strings or `null`
https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem

packages/insomnia-app/app/ui/components/viewers/response-multipart.js
gschier: Yep! Looks like no errors are logged when using it now 😄

packages/insomnia-app/app/ui/components/base/file-input-button.js
gschier: The Electron docs say the array will be empty if `canceled` is `true` so I don't think there's a need to check the length. I'll just remove it

EDIT: Sorry, this reply should have been to your below comment.

gschier: Also, good catch! I must not have tested the cancel functionality on that one.

develohpanda (approved):
Awesome, nice to be consistent with the file checking 😄 and yay for using `.nvmrc` in github actions. That's one less manual step (of remembering to change the node version in actions!)

PR #2406, opened 2020-07-14
Attach codemirror fold marker to themes

### LINKED ISSUE #2404: [Improvement] [Accessibility] Allow changing the folding "heading" font style
Is there a way to change this bold/glowing of the purple text that appears inside the { } when a fold is closed in editor? Check the image below..

Explanation:
I don't like this glowing style as it lowers the contrast and makes it harder to see.

image

develohpanda:
Closes #2404 

Existing UI:
image

Updated UI:
image

Attached to themes:

### DIFF
--- a/packages/insomnia-app/app/ui/css/editor/theme.less
+++ b/packages/insomnia-app/app/ui/css/editor/theme.less
@@ -29,6 +29,15 @@
     border-left: 1px solid var(--color-font);
   }
 
+  .CodeMirror-foldmarker {
+    color: var(--color-surprise);
+    text-shadow: none;
+    font-family: inherit;
+    line-height: .3;
+    cursor: pointer;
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/css/editor/theme.less
```diff
@@ -29,6 +29,15 @@
     border-left: 1px solid var(--color-font);
   }
 
+  .CodeMirror-foldmarker {
```
develohpanda, line 32: Taken from https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2ed25b5dd1f1eccc7b757ca6dbd118d/addon/fold/foldgutter.css#L1-L7 and modified

gschier (approved):
Nice! I love it 😄 

Much easier on the eyes now

PR #2420, opened 2020-07-16
Support config file for inso

### LINKED ISSUE #2370: Configuration file for CLI

develohpanda:
image

Features:
- Uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) for config file management, meaning any of the following items found in the working tree are valid:
	- a `inso` property in `package.json`
	- a `.insorc` file in JSON or YAML format
	- a `.insorc.json` file
	- a `.insorc.yaml`, `.insorc.yml`, or `.insorc.js` file
	- a `inso.config.js` file exporting a JS object
- Settings from the config file are combined with any predefined defaults and any explicit overrides in commands
	This combination is in priority order: `commandOptions` > `configFileOptions` > `defaultOptions`
- A `--config <file>` option to specify an exact file to use, if it exists outside the directory tree
	`inso --config ../../configDir/.insorc gen-conf`
	- An error is only presented to the user if a config file is expected (due to the `--config` flag being specified), and not found
- Run scripts from within the config file either by default `inso gen-conf` or explicitly `inso script gen-conf`, including script nesting
	```yml
	  gen-conf: generate config "Designer Demo" --type declarative
  	  gen-conf:k8s: gen-conf --type kubernetes
	```

Closes #2370


packages/insomnia-inso/src/util.js
```diff
@@ -1,41 +1,10 @@
...
-  }
-
-  return command;
-}
-
 export function getVersion() {
   return packageJson.version;
 }
 
-export function getAllOptions<T>(cmd: Object): T {
```
develohpanda, line 26: Extracted to `get-options.js` because it is a little more complex now

packages/insomnia-inso/src/__tests__/get-options.test.js
```diff
@@ -0,0 +1,116 @@
+// @flow
+import commander from 'commander';
+import getOptions, { extractCommandOptions, loadCosmiConfig } from '../get-options';
+import path from 'path';
+
+jest.unmock('cosmiconfig');
```
develohpanda, line 6: If a manual mock for a node module is present, jest will automatically use it. So, we need to explicitly unmock `cosmiconfig` to use the real module in this test, in order to test loading and merging.

Using fixtures feels more understandable, than mocking the responses, and the `cosmiconfig` implementation itself moved away from mocks for the same reason.

packages/insomnia-inso/src/get-options.js
```diff
@@ -0,0 +1,73 @@
...
+  appDataDir?: string,
+  workingDir?: string,
+  ci?: boolean,
+} & ConfigFileOptions;
+
+export const loadCosmiConfig = (configFile?: string): ConfigFileOptions => {
+  try {
+    const explorer = cosmiconfigSync('inso');
+
+    const results = configFile ? explorer.load(configFile) : explorer.search();
```
gschier, line 29: This is neat

gschier (dismissed):
Implementation looks good!

My only concern is potential config collisions, since everything in `settings` is flat. What happens if you have two commands with a `type` flag? To me, it seems like the only things in `settings` should be truly global.

packages/insomnia-inso/src/__tests__/__snapshots__/inso-snapshot.test.js.snap
```diff
@@ -6,18 +6,20 @@ exports[`Snapshot for "inso --help" 1`] = `
 A CLI for Insomnia!
 
 Options:
-  -v, --version             output the version number
-  -w, --working-dir <dir>   set working directory
-  -a, --app-data-dir <dir>  set the app data directory
-  --ci                      run in CI, disables all prompts
-  -h, --help                display help for command
+  -v, --version           output the version number
+  -w, --workingDir <dir>  set working directory
```
DMarby, line 10: Is the reason these went from kebab-case to camelCase so that they work well as keys in yaml/json, since the config keys has to match the CLI flags?
I don't really mind either way, just curious :)

I guess we need to update the documentation in the README and such as well, since I think this would be a breaking change?

packages/insomnia-inso/src/cli.js
```diff
@@ -1,100 +1,162 @@
...
+        return exit(new Promise(resolve => resolve(false)));
+      }
+
+      // Collect args
+      const scriptArgs: Array<string> = parseArgsStringToArgv(
+        `${scriptTask} ${passThroughArgs.join(' ')}`,
+      );
+
+      // Print command
+      // console.log(`> > inso ${scriptArgs.join(' ')}`);
```
DMarby: Is this meant to be commented out?

DMarby (changes requested):
This is great 🎉 
Just a few minor nitpicks

packages/insomnia-inso/src/__tests__/__snapshots__/inso-snapshot.test.js.snap
develohpanda: Yep, that's the reason for switching from kebab to camel. I prefer kebab for a CLI though :(

Good note regarding docs, I'll publish a README update in this PR documenting the new commands as well.

packages/insomnia-inso/src/cli.js
develohpanda: Yep, I've left it out because piping commands seem to include this text 🤔

Also I feel like printing the command being run is misleading because it doesn't add in any flags extracted from the config file. Thoughts?

develohpanda:
> My only concern is potential config collisions, since everything in `settings` is flat. What happens if you have two commands with a `type` flag?

They will use the global, or use any overrides given in the command/script.

> To me, it seems like the only things in `settings` should be truly global.

Yeah, potentially. I would explicitly be ignoring any non-global options, rather than adding functionality to _support_ non-global options. It feels like it's okay to leave that up to the user but I'm not too pressed either way. Would you suggest to explicitly exclude any non-global options?

packages/insomnia-inso/src/cli.js
DMarby: Sounds fair enough!

DMarby (approved):

gschier:
> Would you suggest to explicitly exclude any non-global options?

That would be my preference, at least to start – especially since we don't know for sure that it will be useful. We can always add support for non-global options later if there's demand.

If we do end up adding command-specific settings, this would be my first approach:

```yaml
settings:
  ci: false
  command1:
    foo: bar
  command2:
    foo: baz
scripts:
  example1: inso command1 ...
  example2: inso command2 ...
  # ...
```

It's a bit more verbose but also a bit more readable, as it's clear which settings relate to which commands.

develohpanda:
> That would be my preference, at least to start – especially since we don't know for sure that it will be useful. We can always add support for non-global options later if there's demand.

That's fair, I agree - will update.

I considered command specific settings (similar to `.babelrc` and their NODE_ENV specific settings), but that was definitely scope creep. The flat structure we get for free, but I'll limit it to global only.

packages/insomnia-inso/src/cli.js
develohpanda: I've added the console log back, given flags specific to a command don't exist in the config file any more :)

DMarby (approved):

gschier (approved):

PR #2432, opened 2020-07-23
Handle silent failures for git push operation

### LINKED ISSUE #2371: Silent failure when pushing to protected branch with insufficient permissions
Trying to push to a protected branch, when a user does not have permission to push to that branch, silently fails. Insomnia does not prompt of any failure to push to a protected branch, but the commits never end up in the repository. Needs further investigation working with protected branches on other VCS providers too.

Found while testing #2240

develohpanda:
Updated error UI:

image

Closes #2371


DMarby (approved):
Very nice 🎉

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -36,6 +37,12 @@ export type GitLogEntry = {|
   },
 |};
 
+export type PushResponse = {
+  ok?: string[],
+  errors?: string[],
```
gschier: Nit: we usually use `Array<string>` in this codebase 🔍

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -60,6 +67,12 @@ export default class GitVCS {
     this._git = git;
     git.plugins.set('fs', fsPlugin);
     git.plugins.set('http', httpPlugin);
+    const emitter = new EventEmitter();
+    git.plugins.set('emitter', emitter);
```
gschier, line 71: Woah, that's a convenient hook.

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
```diff
@@ -150,7 +150,15 @@ class GitSyncDropdown extends React.PureComponent<Props, State> {
 
     const bufferId = await db.bufferChanges();
     try {
-      await vcs.push(gitRepository.credentials, force);
+      const response = await vcs.push(gitRepository.credentials, force);
+      if (response.errors?.length) {
```
gschier: So only git-specific errors throw exceptions? Network errors return error responses?

gschier: I wonder if it would simplify things if we checked `response.errors` inside `vcs.push()` and threw an error from there. Then the caller would only have to `try/catch` and not forget to check the errors array.

gschier (changes requested):
Just a minor comment on simplifying the error handling logic.

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
develohpanda: Git hooks that fail (on the server) are returned in the responses array, because the network request succeeded, but the git operation was rejected. Authentication issues are thrown as exceptions, because the network request is rejected.

This is how `isomorphic-git` is choosing to return the various responses anyway. I agree, abstracting the handling of `isomorphic-git` responses into the GitVCS class is better than outside. I have updated this to make it nicer to consume.

packages/insomnia-app/app/sync/git/git-vcs.js
develohpanda: Oops, this was a straight copy-paste from the bundled Typescript declaration file in the `isomorphic-git` module.

I expect code-style preferences to be governed by ES Lint rules (because no developer or reviewer should need to think about them), so for this I will add:
- [array-style-simple-type](https://github.com/gajus/eslint-plugin-flowtype#array-style-simple-type)
- [array-style-complex-type](https://github.com/gajus/eslint-plugin-flowtype#array-style-complex-type)

These have an auto-fix, and it looks like they picked up an existing discrepancy as well.

packages/insomnia-app/app/ui/components/modals/error-modal.js
```diff
@@ -62,12 +62,12 @@ class ErrorModal extends PureComponent<{}, ErrorModalOptions> {
...
         <ModalHeader>{title || 'Uh Oh!'}</ModalHeader>
         <ModalBody className="wide pad">
-          {message ? <div className="notice error">{message}</div> : null}
+          {message ? <div className="notice error pre">{message}</div> : null}
           {error && (
             <details>
               <summary>Stack trace</summary>
               <pre className="pad-top-sm force-wrap selectable">
-                <code>{error.stack || error}</code>
+                <code className="wide">{error.stack || error}</code>
```
develohpanda, line 70: I have a growing preference for `styled-components`. The changes in this file are meaningless unless the reviewer knows what CSS classes exist in the code-base and what they do.

DMarby (approved):

packages/insomnia-app/app/sync/git/git-vcs.js
gschier: Oh nice! That's a great addition 👍 🌟

packages/insomnia-app/app/ui/components/modals/error-modal.js
gschier: I'm curious how `styled-components` might help. I haven't yet come across a solution for CSS utility classes for styled-components but I'm picturing something like MaterialUI's [`<Box>`](https://material-ui.com/components/box/#box) component that can modify (or wrap) the styles of its children. It would be quite a bit more verbose, however, and still requires us to learn/remember utility names.

I'm still a big fan of utility classes but you're right that they can be messy. Especially ours, which are home-grown and inconsistently named. If the main problem is documentation/standardization, perhaps [TailwindCSS](https://tailwindcss.com/docs/width/#app) would be a good substitute.

---

Zooming back out a bit, for components that use a lot of utility helpers like this, it might make sense to factor the styles out into a component-specific `.less` file. Or, is this maybe what you mean when you mention using `styled-components`, essentially giving the whole component it's own stylesheet?

gschier (approved):

packages/insomnia-app/app/ui/components/modals/error-modal.js
develohpanda: > If the main problem is documentation/standardization

And discoverability. To figure out I need to use `wide`, I did a global search for "width: 100%", and found an appropriate class that has that, that I can use. Definitely not a nice approach, and much, much less manageable for anything more complex. Tailwind, Bootstrap, etc - they all solve this case.

> It might make sense to factor the styles out into a component-specific `.less` file

Instead of this, I'd favor styled-components. With the `.less` approach, if a style is changed, as a reviewer (and sometimes as a developer) you don't know where it's used. And if an element has a new class added to it (like in this PR), a reviewer doesn't know what _css_ styles are being added. styled-components would mean the CSS exists close to where it is consumed and there is more context when a change is seen in a PR.

My preference for this particular fix would be to _not_ used styled-components, but use a utility from Tailwind or Bootstrap (documented, discoverable), but I think the `styled-component` route is faster in this codebase. I don't think `styled-components` is an outright fix to all issues, but it's an approach.

Also, speaking of MaterialUI's `<Box>` component - bootstrap (and I'm sure Tailwind) will have something similar. Bootstrap itself is CSS, while [`react-bootstrap`](https://react-bootstrap.github.io/components/buttons/) introduces components with props that wrap that CSS.

develohpanda: Also note - there's nothing stopping us from using the best of both worlds.

Any really specific/custom CSS can be handled with styled-components, while utilities used for layout.

Just need to think carefully about how to integrate insomnia's theming system with tailwind/boostrap.

```js
const AnimatedFilter = styled.div`...`;

<AnimatedFilter className="d-flex w-100" />
```

sonicyeti: I think we could benefit from hugely from simplifying the theme system's variables to cover colors, fonts and base unit values. We then leverage bootstrap for layout and utility classes. From there, we leverage styled components to compose each individual component.

The styled component would extend bootstrap's utility classes as needed via SC's attrs capability while theming & branding cohesion would be driven via variables similar to what we have now, only greatly pruned down as mentioned above. Bootstrap would essentially be picking up a lot of what is currently being handled by our theming system while exposing a ton of other helpful utility classes.

An example component might look like:

    const MyCustomThing: React.Component<{}> = 
     styled.div.attrs({
        // Extending From Bootstrap core CSS utils
        className='d-inline-flex p2'
      })`
      
      // From Theming System
      color: var(--color-surprise);
      background-color: var(--color-bg);
      font-size: var(--font-size-sm);
      
      // From Hi-Fi design specs > figma
      Custom...
    }`;
 

In short:
- Bootstrap for utils only
- Themes as vars for core branding and dimensional units only
- Styled components to compose UI leveraging above

Tailwind, while epic looking, would impose more unknowns than solutions at this point IMO.

develohpanda: This is a nice approach, I didn't know about `styled.div.attrs`. I think we can also fairly easily configure bootstrap to be aware of our theme variables without having to explicitly set them.

PR #2433, opened 2020-07-23
Sidebar code mirror interactions

sonicyeti:
This PR covers:

- When a user clicks on an item in the sidebar, the editor will select and scroll to the position within the spec.
- Fixed overflow scrolling issue
- Incorporation of yaml-source-map to expose line and column numbers for specs

Example:

Section Validation:


package.json
```diff
@@ -69,5 +69,7 @@
     "prettier": "^1.16.4",
     "rimraf": "^2.6.3"
   },
-  "dependencies": {}
+  "dependencies": {
+    "yaml-source-map": "^2.1.1"
+  }
```
gschier: Should be in `packages/insomnia-app/package.json`

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -144,12 +149,43 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
+        scrollPosition.start.line = itemPosition.start.line;
+      } else {
+        scrollPosition.start.line = itemPosition.start.line - 1;
+      }
+      scrollPosition.end.line = scrollPosition.start.line;
+      scrollPosition.end.col = 200;
+
+      // Scroll to selection
+      this._handleScrollEditor(scrollPosition);
+    };
```
gschier: Would be good to move this all into a component method (since this component is a class)

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -160,6 +160,7 @@ class CodeEditor extends React.Component {
   setSelection(chStart, chEnd, lineStart, lineEnd) {
     if (this.codeMirror) {
       this.codeMirror.setSelection({ line: lineStart, ch: chStart }, { line: lineEnd, ch: chEnd });
+      this.codeMirror.scrollIntoView({ line: lineStart, char: chStart }, 400);
```
gschier: Is this supposed to be here?

packages/insomnia-app/app/ui/css/components/sidebar.less
```diff
@@ -11,6 +11,11 @@
   width: 100%;
   height: 100%;
 
+  .spec-editor-sidebar {
+    overflow: hidden;
+    overflow-y: auto;
+  }
+
```
gschier: This Less file is for the old sidebar. I think this style would better fit in `spec-editor.less`.

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -2,13 +2,15 @@
...
 import YAML from 'yaml';
+import YAMLSourceMap from 'yaml-source-map';
 import SpecEditorSidebarItem from './spec-editor-sidebar-item';
 import { Sidebar } from 'insomnia-components';
 import type { ApiSpec } from '../../../models/api-spec';
 import { trackEvent } from '../../../common/analytics';
 
 type Props = {|
   apiSpec: ApiSpec,
+  onClick: (path: any) => any,
```
gschier: It looks like this prop isn't being used by anything.

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -2,13 +2,15 @@
 import * as React from 'react';
```
gschier, line 2: It looks like there is a bunch of methods in this class that can now be deleted 💣

packages/insomnia-components/components/sidebar/index.js
```diff
@@ -21,6 +21,7 @@ import { useToggle } from 'react-use';
 type Props = {|
   className?: string,
   jsonData: Object,
+  onClick: Function,
```
gschier: Can we increase the specificity on these `onClick` type definitions? Should be something like `onClick: (section: string, path: Array<string>) => void`

gschier (dismissed):
A few higher-level comments right off the bat! Nothing too major, though.

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
sonicyeti: Yeah, scrollIntoView allows for the passing of the margin as a visual buffer (in this case 400) but alone, it doesn't set selection. setSelection has no option for margin but does achieve the actual base positioning.

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -158,8 +159,14 @@ class CodeEditor extends React.Component {
   }
 
   setSelection(chStart, chEnd, lineStart, lineEnd) {
+    const selectionFocusPos = remote.getCurrentWindow().getSize()[1] / 2 - 100;
```
gschier: Seems a bit strange to pull in `electron` as a dependency for this purpose. Should be able to get away with `window.innerHeight`.

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -158,8 +159,14 @@ class CodeEditor extends React.Component {
...
     if (this.codeMirror) {
       this.codeMirror.setSelection({ line: lineStart, ch: chStart }, { line: lineEnd, ch: chEnd });
+      this.codeMirror.scrollIntoView(
```
gschier, line 171: I think a some other parts of the app call this method as well. It might be safer (and easier to understand) to add a new method to scroll to the position (`scrollToSelection(...)`), then call them both from outside.

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -126,30 +45,43 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
+      end: {
+        line: 0,
+        col: 0,
+      },
+    };
 
-    // const navigationEl = document.contents.map(v => this.renderNext(v, '$'));
+    const _handleItemClick = (section, item) => {
+      itemPath = [section];
+      itemPath.push.apply(itemPath, item);
```
gschier: Can also do this with `const itemPath = [section, item...]`

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -126,30 +45,43 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
+    const sourceMap = new YAMLSourceMap();
+    const specMap = sourceMap.index(
+      YAML.parseDocument(apiSpec.contents, { keepCstNodes: true /* must specify this */ }),
+    );
 
-    if (!document.contents) {
-      return null;
-    }
+    let itemPath = [];
+    const scrollPosition = {
```
develohpanda: `sourceMap`, `specMap`, `itemPath` and `scrollPosition` are all, only used by the `_handleItemClick` function. This means we're regenerating the YAML sourcemap on _every_ render, but only consuming it when an item is clicked. We're also re-declaring the `_handleItemClick` function on every render (every change of the spec).

Is this necessary? I think it's worth exploring solutions that make the `render` function leaner and more efficient.

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -126,30 +45,43 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
+        col: 0,
+      },
+      end: {
+        line: 0,
+        col: 0,
+      },
+    };
 
-    // const navigationEl = document.contents.map(v => this.renderNext(v, '$'));
+    const _handleItemClick = (section, item) => {
```
develohpanda: Please add typing to this function definition. For example:

`const _handleItemClick = (section: string, item: Array<string>): void => {`

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -126,30 +45,43 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
+        scrollPosition.start.line = itemPosition.start.line - 1;
+      }
+      scrollPosition.end.line = scrollPosition.start.line;
+      scrollPosition.end.col = 200;
+      // Scroll to selection
+      this._handleScrollEditor(scrollPosition);
+    };
 
     return (
       <div className="spec-editor-sidebar">
```
develohpanda: I suggest using a `styled-component` here and keeping the styles close to where the are being consumed, especially given the CSS for this is simply
```less
.spec-editor-sidebar {
  overflow: hidden;
  overflow-y: auto;
}
```

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -126,30 +45,43 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
+      itemPath.push.apply(itemPath, item);
+      const itemPosition = sourceMap.lookup(itemPath, specMap);
+      // Hack for servers array due to offest in YAML CST mapping
+      if (section === 'servers') {
+        scrollPosition.start.line = itemPosition.start.line;
+      } else {
+        scrollPosition.start.line = itemPosition.start.line - 1;
+      }
+      scrollPosition.end.line = scrollPosition.start.line;
+      scrollPosition.end.col = 200;
```
develohpanda: Any specific reason for this being a hard-coded 200?

packages/insomnia-components/components/sidebar/sidebar-headers.js
```diff
@@ -35,7 +30,7 @@ export default class SidebarHeaders extends React.Component<Props> {
               <div>
                 <SvgIcon icon={IconEnum.indentation} />
               </div>
-              <span onClick={() => handleClick([header])}>
+              <span onClick={() => this.props.onClick('components', ['headers', header])}>
```
develohpanda: I highly encourage destructing `this.props` where-ever appropriate, because React components are verbose enough as it is.

```js
const {headers, onClick} = this.props;
```

packages/insomnia-components/components/sidebar/sidebar-info.js
```diff
@@ -28,7 +28,7 @@ function SidebarInfo(props: Props) {
           <SidebarTextItem label={'Version:'} headline={version} />
         </SidebarItem>
       )}
-      {license.name && (
+      {license && license.name && (
```
develohpanda, line 31:
```suggestion
      {license?.name && (
```

[Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) is a friend 🤗

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
develohpanda: Looking at the usages of this, I feel like it might be worthwhile using  js [`rest parameters`](https://scotch.io/bar-talk/javascripts-three-dots-spread-vs-rest-operators543#toc-rest-parameters) here.

```js
const _handleItemClick = (...itemPath: Array<string>): void => {
	const isServersSection = itemPath[0] === 'servers;
	...
}
```

Fewer array operations make it more efficient - this replaces three array creations and one `push.apply`, in favor of a single `...itemPath` and indexed access.

packages/insomnia-components/components/sidebar/sidebar.stories.js
```diff
@@ -2944,11 +2945,217 @@ const apiSpec = {
...
+  const _handleItemClick = (section, item) => {
+    itemPath = [];
+    itemPath.push(section);
+    itemPath.push.apply(itemPath, item);
+    setTargetPath(itemPath.join(' > '));
+  };
+
+  return (
+    <div style={{ width: '350px' }}>
+      <div style={{ display: 'block', margin: '0px 0px 20px 0px' }}>Target: {targetPath}</div>
```
develohpanda, line 3146: This is cool 👍

develohpanda (changes requested):
I love the changes in storybook! 🔥

Couple of comments, but generally looking good. 😄

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
sonicyeti: On the yaml source map we get back the entire node's start and ending column position rather than just the intended single line item's ending so I'm manually passing 200 which it will shrink from when rendering the selection.

packages/insomnia-components/components/sidebar/sidebar-info.js
sonicyeti: Tried enabling in flow via esproposal.optional_chaining=enable within components and restarted flow server but no luck. It has already been enabled in app. Not sure why it isn't happy here, I'll circle back on this.

develohpanda: You might need to install `@babel/plugin-proposal-optional-chaining` and add that to `.babelrc`. didn't realize it wasn't enabled in components! We can do that outside this PR.

https://github.com/Kong/insomnia/blob/93c91ebdbed9dc635056c920686913722fd19965/packages/insomnia-app/.babelrc#L14

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -126,31 +74,18 @@ class SpecEditorSidebar extends React.PureComponent<Props, State> {
...
-    const [document] = specCst;
-
-    if (!document.contents) {
-      return null;
-    }
-
-    // const navigationEl = document.contents.map(v => this.renderNext(v, '$'));
+    const _handleItemClick = (...itemPath): void => {
+      // Buid up path (no arr.flat() support)
+      const mappedPosition = this._mapPosition([itemPath[0]].concat(...itemPath[1]));
```
develohpanda: Just the comment here we discussed about triggering this callback with `onClick('components', 'parameters', parameter)` instead of `onClick('components', ['parameters', parameter])`, and moving the this._scrollToPosition inside the class method so we can remove this callback being created during render 😄

develohpanda (changes requested):
🔥

develohpanda (approved):
Nice! Excited for this one 😁🔥

packages/insomnia-components/components/sidebar/index.js
```diff
@@ -67,16 +68,10 @@ function Sidebar(props: Props) {
...
-    parameters,
-    headers,
-    schemas,
-    securitySchemes,
-  } = props.jsonData.components;
-  const paths = Object.entries(props.jsonData.paths);
+  const { servers, info } = props.jsonData || {};
+  const { requestBodies, responses, parameters, headers, schemas, securitySchemes } =
+    props.jsonData.components || {};
+  const paths = Object.entries(props.jsonData.paths || {});
```
gschier: Should probably be checking to see if the type of `paths` is an object here. You'll get some weird results if `paths` is a string, for example.

gschier: Same with `components` and the rest

gschier (commented):
A few comments

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
develohpanda: I think you may have been reviewing an old version of this file? This logic has been switched in favor of rest parameters.

https://github.com/Kong/insomnia/blob/cccad719c73be4060f5073b6d8a9e0218a1e8823/packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js#L46-L72

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
sonicyeti: Hah, I forget sometimes that we are in a browser ; ) thanks!

packages/insomnia-components/components/sidebar/index.js
```diff
@@ -68,10 +68,24 @@ function Sidebar(props: Props) {
...
+    schemas,
+    securitySchemes,
+  }: {
+    requestBodies: Object,
+    responses: Object,
+    parameters: Object,
+    headers: Object,
+    schemas: Object,
+    securitySchemes: Object,
+  } = props.jsonData.components || {};
```
develohpanda: Hmm, not a fan of typing like this, it makes the code unnecessarily complex and difficult to understand.

These types should be defined with the `Props` definition.

```js
type Props = {
	jsonData: {
		servers?: Object,
		info?: Object,
		paths?: Object,
		components?: {
			requestBodies?: Object,
		    responses?: Object,
		    parameters?: Object,
		    headers?: Object,
		    schemas?: Object,
		    securitySchemes?: Object,
		}
	}
}
```

develohpanda (changes requested):
Comment regarding type definitions

develohpanda (approved):
Yay! 😊🤗

packages/insomnia-components/components/sidebar/sidebar-headers.js
```diff
@@ -16,6 +17,10 @@ export default class SidebarHeaders extends React.Component<Props> {
   renderBody = (filter: string): null | React.Node => {
     const { headers, onClick } = this.props;
 
+    if (Object.prototype.toString.call(headers) !== '[object Object]') {
```
develohpanda, line 20: 🤔 Is there a particular reason we didn't just use `typeof` here?

packages/insomnia-components/components/sidebar/sidebar-paths.js
```diff
@@ -19,7 +20,9 @@ const StyledMethods: React.ComponentType<{}> = styled.span`
 export default class SidebarPaths extends React.Component<Props> {
   renderBody = (filter: string): null | React.Node => {
     const { paths, onClick } = this.props;
-
+    if (Object.prototype.toString.call(paths) !== '[object Array]') {
```
develohpanda, line 23: Same here, is there a particular reason for not using `Array.isArray(paths)`?

Strange that the type definition in Props was generalized to `Object` as well 🤔

PR #2462, opened 2020-07-29
Return 'dev' as the version if running in development

develohpanda:
It gets difficult to identify which version of inso is installed locally, so the `inso -v` option will not identify whether the version is live (packageJson.version) or development (dev).

image

### DIFF
--- a/packages/insomnia-inso/src/util.js
+++ b/packages/insomnia-inso/src/util.js
@@ -2,7 +2,11 @@
 import * as packageJson from '../package.json';
 
 export function getVersion() {
-  return packageJson.version;
+  return isDevelopment() ? 'dev' : packageJson.version;
+}
+
+export function isDevelopment() {
+  return process.env.NODE_ENV === 'development';
 }
 
 export function logErrorExit1(err: Error) {
--- a/packages/insomnia-inso/src/__tests__/cli.test.js
+++ b/packages/insomnia-inso/src/__tests__/cli.test.js
@@ -5,6 +5,7 @@ import { lintSpecification } from '../commands/lint-specification';
 import { runInsomniaTests } from '../commands/run-tests';
 import { exportSpecification } from '../commands/export-specification';
 import { parseArgsStringToArgv } from 'string-argv';
+import * as packageJson from '../../package.json';
 
 jest.mock('../commands/generate-config');
 jest.mock('../commands/lint-specification');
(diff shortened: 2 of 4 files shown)


packages/insomnia-inso/src/util.js
```diff
@@ -2,7 +2,11 @@
 import * as packageJson from '../package.json';
 
 export function getVersion() {
-  return packageJson.version;
+  return isDevelopment() ? 'dev' : packageJson.version;
+}
+
+export function isDevelopment() {
+  return process.env.NODE_ENV === 'development';
```
DMarby, line 9: I think this approach might give unintended results (in particular if used for gating anything else in the future), since it's probably not uncommon for NodeJS devs to have `NODE_ENV` set to `development` in their local environment variables, and they might also be using `inso`?

Maybe I'm misunderstanding how this is set however, do we overwrite to be hardcoded to production somewhere in our build process for our release builds?

DMarby (changes requested):

packages/insomnia-inso/src/util.js
develohpanda: image

Setting the webpack [mode](https://webpack.js.org/configuration/mode/#usage) to production, configures the NODE_ENV before it builds.

develohpanda: We set it to production here: https://github.com/Kong/insomnia/blob/c73566f12ad1ce9e51c94bd802ee5626046c2295/packages/insomnia-inso/webpack/webpack.config.production.js#L6 and consume it here: https://github.com/Kong/insomnia/blob/c73566f12ad1ce9e51c94bd802ee5626046c2295/packages/insomnia-inso/package.json#L21

develohpanda: You'll notice in the example linked in the first comment, that `mode: 'production'` configures the following plugin:
`new webpack.DefinePlugin({ "process.env.NODE_ENV": JSON.stringify("production") })`

`DefinePlugin`, [TIL](https://webpack.js.org/plugins/define-plugin/#usage), actually inlines each key into code before the build 
> The values will be inlined into the code allowing a minification pass to remove the redundant conditional.

DMarby: Ah, sweet, should be all good then!

I was looking in our webpack configs, but didn't spot any direct usage of DefinePlugin or similar to set it, so I was unsure 😅 
Didn't know webpack automagically does that for you nowadays when you set the mode, wonder if that's always been there 😂

DMarby (approved):
🎉

PR #2466, opened 2020-07-30
Inso error handling and verbosity setting for logging

### LINKED ISSUE #2438: Better error messaging around duplicate base environments in the CLI
Currently it exits with a stack trace and semi-cryptic message, we should provide a more specialized and searchable message for users to be able to take action on.

develohpanda:
- Added `--verbose` and `--printOptions` global flags.
- Converted all `console.*` calls to be `logger.*`, which route through [consola](https://github.com/nuxt-contrib/consola). This allows for mocking the logger, as well as control verbosity during running.
- By default, only `logger.[fatal|error|log|info]` will show. With the `--verbose` flag, `logger.[trace|debug]` will also be shown.

image

image

image

image

image

image

For the purpose of unit testing (in the app), `logger.__getLogs()` extracts all of the calls made to each method and their parameters.

image

Closes #2438


packages/insomnia-inso/src/__tests__/cli.test.js
```diff
@@ -46,18 +55,31 @@ describe('cli', () => {
     });
 
     it.each(['-v', '--version'])('inso %s should print version from package.json', args => {
-      jest.spyOn(console, 'log').mockImplementation(() => {});
+      logger.wrapAll();
       expect(() => inso(args)).toThrowError(packageJson.version);
+      logger.restoreAll();
```
develohpanda, line 60: Wrapping and unwrapping, mainly to keep the test output clean. Commander internally handles the version flag.

packages/insomnia-inso/src/__tests__/cli.test.js
```diff
@@ -192,13 +214,11 @@ describe('cli', () => {
...
   describe('script', () => {
-    let consoleLogSpy;
-    beforeEach(() => {
-      consoleLogSpy = jest.spyOn(console, 'log').mockImplementation(() => {});
-    });
-
     const insorcFilePath = '--config src/__fixtures__/.insorc-with-scripts.yaml';
 
+    const expectExitWith = async (result: boolean): Promise<void> =>
+      expect((exit: Object).mock.calls[0][0]).resolves.toBe(result);
```
develohpanda, line 220: `.calls[0][0]` represents the first argument of the first time this mock was called.

packages/insomnia-inso/src/cli.js
```diff
@@ -81,7 +100,11 @@ function makeExportCommand(createCommand: CreateCommandType) {
     .command('spec [identifier]')
     .description('Export an API Specification to a file')
     .option('-o, --output <path>', 'save the generated config to a file')
-    .action((identifier, cmd) => exit(exportSpecification(identifier, getOptions(cmd))));
+    .action((identifier, cmd) => {
+      const options = getOptions<ExportSpecificationOptions>(cmd);
+      prepareCommand(options);
+      return exit(exportSpecification(identifier, options));
```
develohpanda, line 106: `exit(command())` returns a promise, which is used by commander, so we must return it here.

packages/insomnia-inso/src/cli.js
```diff
@@ -118,7 +142,7 @@ function addScriptCommand(originalCommand: Object) {
       );
 
       // Print command
-      console.log(`>> ${scriptArgs.slice(1).join(' ')}`);
+      logger.debug(`>> ${scriptArgs.slice(1).join(' ')}`);
```
develohpanda, line 145: Yay for verbosity :) This will only show with the `--verbose` flag.

packages/insomnia-inso/src/logger.js
```diff
@@ -9,3 +11,14 @@ export const noConsoleLog = async <T>(callback: () => Promise<T>): Promise<T> =>
...
   }
 };
+
+const logger = consola.create({
+  reporters: [new consola.FancyReporter({ dateFormat: ' ' })],
+});
+
+export const configureLogger = (verbose: boolean = false, ci: boolean = false) => {
+  logger.level = verbose ? consola.LogLevel.Verbose : consola.LogLevel.Info;
+  ci && logger.setReporters([new consola.BasicReporter()]);
```
develohpanda, line 21: I would like to remove `--ci` as an option, in favor of something like [std-env](https://www.npmjs.com/package/std-env) or [is-ci](https://www.npmjs.com/package/is-ci), in the future.

DMarby: I think while having auto-detection with `std-env` or `is-ci` or similar would be great, there should still be a way to explicitly set it (via a flag or env var or similar), because those detection methods doesn't necessarily work with all CI setups, generally only the large common ones.

packages/insomnia-inso/src/cli.js
DMarby: Awesome! :tada:

packages/insomnia-inso/src/logger.js
```diff
@@ -9,3 +11,14 @@ export const noConsoleLog = async <T>(callback: () => Promise<T>): Promise<T> =>
     (console: Object).log = oldConsoleLog;
   }
 };
+
+const logger = consola.create({
+  reporters: [new consola.FancyReporter({ dateFormat: ' ' })],
```
DMarby: Is this to turn off the date formatting until the `date` option that you submitted upstream is released?

DMarby (approved):

packages/insomnia-inso/src/logger.js
develohpanda: Yeah. I'm pretty sure it prints a blank space, instead of printing nothing at all. In any case, not very intuitive. 😅 The new setting will be consumed in this exact constructor once released.

packages/insomnia-inso/src/logger.js
develohpanda: Good point, that should be simple enough to combine 👍

packages/insomnia-inso/src/logger.js
DMarby: Fair enough!

develohpanda: Updated!

PR #2467, opened 2020-07-30
Checking format to offset line number if spec string is JSON parsable.

### LINKED ISSUE #2435: Add support for scroll to click in sidebar for JSON specifications

sonicyeti:
Spec type is YAML regardless of content format but when mapping, if the content format is JSON the scroll line number is offset. This fixes that by checking if the spec contents are parsable as JSON.

Closes #2435


develohpanda (commented):
Going to wait on this review until https://github.com/Kong/insomnia/pull/2433 is merged I think; there are changes here that are also in the other PR?

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -71,6 +86,17 @@ class SpecEditorSidebar extends React.Component<Props, State> {
     this._mapPosition(itemPath);
   };
 
+  componentDidMount() {
```
develohpanda, line 80: Since this only runs on mount, if I start with a blank spec (which will be treated as YAML), but then start typing in JSON, will it be treated correctly?

I don't think we need to change it, given the immediate impact would be the click to scroll is one line off, and it'll be fixed for every future navigation to the design tab. Just want to understand the impact of it treating a spec as YAML when it is, in-fact JSON.

I don't think changing to `componentDidUpdate` is the solution either, since `JSON.parse` will run and likely throw an exception on _every_ update 😬

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -28,9 +28,21 @@ class SpecEditorSidebar extends React.Component<Props, State> {
     super(props);
     this.state = {
       error: '',
+      specContentJSON: false,
     };
   }
 
+  _IsSpecContentJson(str: string) {
```
develohpanda: I don't think this method is used anymore 😄

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -60,9 +68,16 @@ class SpecEditorSidebar extends React.Component<Props, State> {
...
+      scrollPosition.start.line = 1;
+    }
+    const specMap = sourceMap.index(YAML.parseDocument(contents, { keepCstNodes: true }));
+    const itemMappedPosition = sourceMap.lookup(itemPath, specMap);
+    const isServersSection = itemPath[0] === 'servers';
     isServersSection
-      ? (scrollPosition.start.line = itemMappedPosition.start.line)
-      : (scrollPosition.start.line = itemMappedPosition.start.line - 1);
+      ? (scrollPosition.start.line += itemMappedPosition.start.line)
+      : (scrollPosition.start.line += itemMappedPosition.start.line - 1);
```
develohpanda: Just to make this one a little less jarring

```js
scrollPosition.start.line += itemMappedPosition.start.line;

if (isServersSection) {
	scrollPosition.start.line -= 1;
}
```

develohpanda (changes requested):
Couple of minor comments 😄 Looks good though!

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
sonicyeti: Great point. In testing, it is thrown off by 1 line if starting fresh via default YAML and you author in JSON. This brings up another question. What if I decided to shift from JSON to YAML at any time on any existing spec, the same occurs. I'm wondering if we start detecting format as a standard, in general, before rendering anything.

develohpanda: In both cases, if you close the document and reopen, it will be fixed - so I don't think we need to do it any more frequently than on mount, but just something to keep in mind.

The cost of checking format on _every render_ isn't worth it, especially since the alternative is just being one line out, and only until the document is reopened.

@nijikokun thoughts, from a product perspective?

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
```diff
@@ -75,10 +64,12 @@ class SpecEditorSidebar extends React.Component<Props, State> {
     const specMap = sourceMap.index(YAML.parseDocument(contents, { keepCstNodes: true }));
     const itemMappedPosition = sourceMap.lookup(itemPath, specMap);
     const isServersSection = itemPath[0] === 'servers';
-    isServersSection
-      ? (scrollPosition.start.line += itemMappedPosition.start.line)
-      : (scrollPosition.start.line += itemMappedPosition.start.line - 1);
+    scrollPosition.start.line += itemMappedPosition.start.line;
+    if (!isServersSection) {
```
develohpanda, line 68: 🤣 see how I was confused by the previous implementation and suggested `if(isServersSection)`, when it should have been `if(!isServersSection)`. This is much clearer.

develohpanda (approved):

packages/insomnia-app/app/ui/components/spec-editor/spec-editor-sidebar.js
nijikokun: Good point, let's punt on it, in the future we should have the user declare types.

PR #2469, opened 2020-07-31
Missing dot at last line.

luanldt:
It is uncomfortable for perfect eyes.


develohpanda (approved):
Sharp eye! You're right, once seen it's difficult to unsee. 👀 Thanks for the PR!

PR #2474, opened 2020-07-31
Port same deterministic logic from OpenAPI 3 to Swagger 2 importer

### LINKED ISSUE #2437: Determine and resolve duplication issues with auto-generation from Swagger 2 specs
Currently when generating debug requests, sometimes, the base environment can be duplicated causing errors in the CLI. Additionally, folders can be duplicated as well, causing user frustration.

gschier:
Fixes #2437

When clicking the Debug tab in Designer (or importing files), workspaces/environments/folders/requests are created. Before this PR, the Swagger 2 importer would generate new IDs every time, causing things to be "duplicated." We had a solution for this for OpenAPI 3 already, so I ported this logic to the Swagger 2 importer. 

Note: There aren't any workflows I can think of that would be broken by this change, but it's something to keep in mind.


develohpanda (approved):
LGTM 💯💥

DMarby (approved):

sonicyeti (approved):
Nice, glad you circled back on this for Swagger 2!

Screen Shot 2020-08-03 at 8 08 28 AM

PR #2475, opened 2020-08-01
Replace textarea with code editor for response raw viewer

### LINKED ISSUE #2397: When the user switches to "RAW DATA" viewing mode in a binary file, insomnia crash.
1. Create a request for an api that responds with a binary file (PDF, ZIP).
2. Switch from "visual preview" mode to "row data" mode.

Screenshots
image

**All functions of the Insomnia crashes, nothing works**
image

Description
When the answer files are deleted, the insomnia will work again

Info
 - OS: Windows 10
 - App Version: v8 6.6.346.32

aszx87410:
Fixes #2397 

After look into this issue, I found that it's not actually "crash" but freeze. When we set large text to textarea, somehow it blocks the main thread so nothing works until text is loaded. 

I think we should not use textarea in this scenario because it's not for large text. We can replace textarea with code editor to avoid this issue because loading large chunk of text already handled by codemirror.

Below are some pdf files to reproduce the issue:

1. https://github.com/tpn/pdfs/raw/master/A%20Comparison%20of%20Programming%20Languages%20in%20Economics%20(16-Jun-2014).pdf 206kb
2. https://eloquentjavascript.net/Eloquent_JavaScript.pdf 2.2MB ( be careful to use this because it will freeze the current app for a long time)

After switching to code editor it works fine. Please let me know if there is any problems, thanks!


develohpanda (changes requested):
Nice, I like this. Code-mirror doing what it's good at, efficiently.

It seems like code-mirror is doing something to render raw JSON response nicely. This is only a rendering difference though, as I don't see any white-space when copy-pasting the content into VS Code.

Top is v2020.3.3, bottom is this PR.

image

So two things to try and address for this PR:
- [ ] Are we able to disable anything extra that code-mirror (or our code-editor component) is doing, in order to allow the raw response to be truly raw? 💥 
- [ ] It would be awesome to bring the CSS spacing from the text-area implementation to the code-mirror implementation. You'll notice in the screenshot above they are slightly different on all four sides. 💄

develohpanda:
Raised https://github.com/Kong/insomnia/issues/2478 as a related issue

aszx87410:
Fortunately it seems that it's all styling issue. I made some changes in this commit: https://github.com/Kong/insomnia/pull/2475/commits/887de23f36d62673cdcebfd94deb96d258147749

We need to restyle codemirrow so I added a new class called `raw-editor`.  Although we can pass `raw-editor` as className to CodeEditor, I still introduce a new props called `raw` because we may need this raw editor somewhere else in the future. Moreover, we can also disable some functionalities in the CodeEditor component by looking at `raw` props. Passing this props is like: "Hey, I want this editor behave like a real raw viewer", the meaning is quite different compare to just passing a className.

Left one is PR while right is the original textarea, it looks pretty much the same!

螢幕快照 2020-08-04 上午12 17 23

packages/insomnia-app/app/ui/css/editor/raw-editor.less
```diff
@@ -0,0 +1,10 @@
+/*
+  We need to restyle CodeMirrow for raw editor
+*/
+.raw-editor {
+  .CodeMirror-wrap pre.CodeMirror-line, .CodeMirror-wrap pre.CodeMirror-line-like {
+    word-break: break-all;
+    padding: var(--padding-md);
```
develohpanda: I'm on the fence about specifying padding here, versus sending a `className="pad"` prop into `CodeEditor` from `ResponseRaw`.

Spacing is a concern of the consumer, while `word-break`/`line-height` are a concern of the `CodeEditor` showing raw content correctly, as controlled by the `raw` prop.

For this reason I suggest removing padding from the `raw-editor` style and sending it as a prop from `ResponseRaw`. What are your thoughts?

Side note: could also merge this file and place the style within `css/editor/general.less`

develohpanda (approved):
> Although we can pass `raw-editor` as className to CodeEditor, I still introduce a new props called `raw`

I'm in favor of `raw` being a prop rather than passing the `raw-editor` class as a prop, however I don't think padding should be a part of `raw-editor`; see comment below.

Other than this minor change, looks good to me!

develohpanda (changes requested):
Oops, one comment that is worth addressing. Clicked the wrong button! 😄

packages/insomnia-app/app/ui/css/editor/raw-editor.less
aszx87410: At first I tried to send className prop to `CodeEditor` but it doesn't work, so I have to restyle this way. The problem is, the HTML structure of the code mirrow. We have limited control by adding class class in the wrapper because of the structure under the hood. Below is the screenshot for sending `className="pad"` prop:

截圖 2020-08-04 下午12 17 53

The UI is different from original textarea, for example, the scrollbar is not at the right position, and also an unexpected gap on the top of the container. I have looked into this and found that we need to add padding in CSS to restyle, instead of sending a class name to the container.

aszx87410: I pushed a new commit to merge the style into `general.less`

develohpanda (approved):
Nice one, thanks for working through this! I see what you mean regarding padding, and why it needs to exist on the `CodeMirror-*` classes rather than higher up. 🎉

LGTM 💯

DMarby (approved):
This is awesome, thanks for taking the time to implement this! 🎉

sonicyeti (approved):
Super solid, LGTM!

PR #2476, opened 2020-08-01
Consistent export dropdown verbiage

### LINKED ISSUE #2098: Designer export dropdown says "workspace" instead of "document"
**Describe the bug**
Designer verbiage says "Workspace" instead of "Document"

**To Reproduce**
1. `Cmd` + `,` to open the preferences tab
2. Click on data
3. Click Export data

**Expected behavior**
Should use the "document" verbiage from https://github.com/Kong/insomnia/blob/6fc394891d575931a0ca7f9babde3adb6a7452ef/packages/insomnia-app/app/common/strings.js#L7

**Screenshots**
image

 - OS: macOS 10.15.3
 - Installation Method: DMG install
 - App Version: v2020.1.0-beta.1

aszx87410:
Fixes #2098

Thanks for the instructions in the PR, it's quite clear and helpful.

I follow the naming convention here: [insomnia/packages/insomnia-app/app/ui/components/dropdowns/document-card-dropdown.js](https://github.com/Kong/insomnia/blob/443d6df08d50c0912c11b039d46739d70790d9be/packages/insomnia-app/app/ui/components/dropdowns/document-card-dropdown.js#L12), use `import Strings from '../../../common/strings';` 

It works as expected now, display "document" for Designger.

螢幕快照 2020-08-01 下午4 01 36

### DIFF
--- a/packages/insomnia-app/app/ui/components/settings/import-export.js
+++ b/packages/insomnia-app/app/ui/components/settings/import-export.js
@@ -4,6 +4,7 @@ import autobind from 'autobind-decorator';
 import { Dropdown, DropdownButton, DropdownDivider, DropdownItem } from '../base/dropdown';
 import Link from '../base/link';
 import { showPrompt } from '../modals/index';
+import Strings from '../../../common/strings';
 
 @autobind
 class ImportExport extends PureComponent {
@@ -48,11 +49,11 @@ class ImportExport extends PureComponent {
             <DropdownDivider>Choose Export Type</DropdownDivider>
             <DropdownItem onClick={handleShowExportRequestsModal}>
               <i className="fa fa-home" />
-              Current Workspace
+              Current {Strings.workspace}
             </DropdownItem>
             <DropdownItem onClick={handleExportAll}>
               <i className="fa fa-empty" />
-              All Workspaces
+              All {Strings.workspace}
             </DropdownItem>
           </Dropdown>
           &nbsp;&nbsp;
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/ui/components/settings/import-export.js
```diff
@@ -48,11 +49,11 @@ class ImportExport extends PureComponent {
             <DropdownDivider>Choose Export Type</DropdownDivider>
             <DropdownItem onClick={handleShowExportRequestsModal}>
               <i className="fa fa-home" />
-              Current Workspace
+              Current {Strings.workspace}
             </DropdownItem>
             <DropdownItem onClick={handleExportAll}>
               <i className="fa fa-empty" />
-              All Workspaces
+              All {Strings.workspace}
```
develohpanda: Interesting, I hadn't thought about singular vs plural. This should resolve to `All Documents` in Designer, but it resolves to `All Document`. 😅

I'd say for now, just make another entry in `common/strings` for the plural, `workspaces`, and set it correctly. Once more things are moved into the strings file, we can look at how to do plurals in a nicer way.

develohpanda (changes requested):
Thanks for the PR! I overlooked singular vs plural when raising the issue, so a small comment regarding that. 😄

develohpanda (approved):
Nice! 🎉 💯

packages/insomnia-app/app/ui/components/settings/import-export.js
aszx87410: Nice catch! Thanks for pointing out, I didn't notice this. I pushed a new commit to add `workspaces` string.

Issue #760, opened 2018-02-14
[Improvement] Display version number more obviously


Mardoxx:
- Insomnia Version: Hold on while I try and figure it out (5.14.7)
- Operating System:  w10

## Details

Would be nice to have Help->About to show about page with version number.

Wasn't very intuitive to go Application->Preferences and then find the version number at the top of the modal.

gschier:
I'm not a Windows user so maybe you can answer this. Are About screens common in Windows? If so, can you provide some screenshots examples of about screens from some popular apps?

Mardoxx:
Not sure if it's just a windows thing! It's certainly a common theme in windows and xplat apps.

I tried to find out if this sort of UI feature had a name, but I came up short :(

Here are some examples!

All found `Help->About`

Chrome just links to `chrome://settings/help`

VS Code's is simple like this 
![image](https://user-images.githubusercontent.com/239665/36262094-640e2fc2-125e-11e8-85c6-689232f3dec8.png)

Slack:
![image](https://user-images.githubusercontent.com/239665/36262152-7fdd3dec-125e-11e8-8f6c-02fa2adda050.png)

Visual Studio:
![image](https://user-images.githubusercontent.com/239665/36262110-6aaaa310-125e-11e8-89e0-7e19e3a8989c.png)

Photoshop:
![image](https://user-images.githubusercontent.com/239665/36262283-e311cbd0-125e-11e8-80c7-56513133402e.png)

Mardoxx:
Something like VSCode's would be easiest to implement I suppose https://github.com/Microsoft/vscode/blob/c869fef06cd7edcc0c26eee143bea4695a1a8a67/src/vs/platform/windows/electron-main/windowsService.ts#L416-L449

gschier:
Ah yes, it's definitely a cross-platform thing but Electron provides it for free on Mac using the native OS APIs. So I guess my question was "Is there a _specific_ way to do it in Windows", which it seems like there isn't. 

Both Slack and VS Code are Electron too, so it's probably worth looking at how they implement it for non-mac platforms. Thanks for the response! 😄 

![image](https://user-images.githubusercontent.com/587576/36262689-240b4438-12a3-11e8-84ea-a7ecba8b067a.png)

Mardoxx:
Haha! Completely misunderstood what you were asking! I had no idea macOS did that 😄 

Can take a look at this after I've gone through that file-plugin bug, unless you get round to it before me 😄

[...]

PR #2480, opened 2020-08-03
Copy info to clipboard

### LINKED ISSUE #2097: Menu item for copying app info to clipboard
See **additional info** [here](https://github.com/Kong/insomnia/issues/2094#issue-610168314), and maybe there are more instances of this.

Idea is that there is a menu item that when clicked will copy application details such as the version, operating system, etc... that one could paste into an issue body or comment body instead of having to manually type it out.

Innoxious:
Closes #2097 

Adds a button in the message box which will write system + insomnia details into the users clipboard.
image
copyInfoToClipboard


packages/insomnia-app/app/main/window-utils.js
```diff
@@ -295,23 +295,32 @@ export function createWindow() {
...
+          'Shell ' + process.versions.electron,
+          'Node ' + process.versions.node,
+          'V8 ' + process.versions.v8,
+          'Architecture ' + process.arch,
+          Curl.getVersion(),
+        ].join('\n');
+
+        const msgBox = await dialog.showMessageBox({
           type: 'info',
+          buttons: ['Copy to clipboard', 'OK'],
```
develohpanda: Ooooo this is cool, TIL 😮

develohpanda (approved):
Looking awesome 🔥

packages/insomnia-app/app/main/window-utils.js
```diff
@@ -295,23 +295,32 @@ export function createWindow() {
...
   if (!isMac()) {
+    const os = require('os');
     helpMenu.submenu.unshift({
       label: `${MNEMONIC_SYM}About`,
       click: async () => {
-        await dialog.showMessageBox({
+        const operatingSystem = [os.type(), os.release()].join(' ');
+        const detail = [
+          'OS ' + operatingSystem,
+          'Version ' + getAppVersion(),
```
develohpanda: Might be good to add the long app name here, so it shows in the snippet pasted into GH. This will either be `Insomnia Core` or `Insomnia Designer`.

```suggestion
          `Version ${getAppLongName()} ${getAppVersion()}`,
```

develohpanda (changes requested):
Small change I just noticed! 🤗

packages/insomnia-app/app/main/window-utils.js
```diff
@@ -295,23 +295,32 @@ export function createWindow() {
   };
 
   if (!isMac()) {
+    const os = require('os');
```
develohpanda: Oh this can also move to the imports at the top of the file 😄

develohpanda:
I looked back at #760, seems that the About screen doesn't exist for macOS because Electron gives it for free under `Insomnia Designer > About Insomnia Designer`.

image

VS Code also being electron, pops open a similar dialog with the `About Visual Studio Code` menu item, and has copy logic in it, so likely we need to extend whatever is setting the about page for macOS.

image

Innoxious:
@john-vuong pls help for mac

john-vuong:
There is no way to configure the [macOS provided `About`](https://developer.apple.com/documentation/appkit/nsapplication/aboutpaneloptionkey) menu item to how we want.

[vscode](https://github.com/microsoft/vscode/blob/d890a22dc21fed13cf5a9eed81780d700f6e4f32/src/vs/workbench/services/dialogs/electron-browser/dialogService.ts#L247) uses the `showMessageBox` to show the About dialog like we have.


image

packages/insomnia-app/app/main/window-utils.js
```diff
@@ -294,24 +292,52 @@ export function createWindow() {
...
+    `OS ${os.type()} ${os.arch()} ${os.release()}`,
+    `Version ${getAppLongName()} ${getAppVersion()}`,
+    `Shell ${process.versions.electron}`,
+    `Node ${process.versions.node}`,
+    `V8 ${process.versions.v8}`,
+    `Architecture ${process.arch}`,
+    Curl.getVersion(),
+  ].join('\n');
+
+  const aboutMsgOptions = {
```
develohpanda, line 305: It would be nice to add `cancelId` and `defaultId` - see the [electron docs](https://www.electronjs.org/docs/api/dialog#dialogshowmessageboxbrowserwindow-options) about what they represent and see the [VS Code implementation](https://github.com/microsoft/vscode/blob/d890a22dc21fed13cf5a9eed81780d700f6e4f32/src/vs/workbench/services/dialogs/electron-browser/dialogService.ts#L254-L255) as inspiration.

packages/insomnia-app/app/main/window-utils.js
```diff
@@ -294,24 +292,52 @@ export function createWindow() {
...
-            'Shell ' + process.versions.electron,
-            'Node ' + process.versions.node,
-            'V8 ' + process.versions.v8,
-            'Architecture ' + process.arch,
-            '', // Blank line before libcurl
-            Curl.getVersion(),
-          ].join('\n'),
+        const msgBox = await dialog.showMessageBox({
+          ...aboutMsgOptions,
+          buttons: ['Copy to clipboard', 'OK'],
```
develohpanda: Looking at the [electron docs](https://www.electronjs.org/docs/api/dialog#dialogshowmessageboxbrowserwindow-options), setting `noLink: true` will stop the "Copy to clipboard" showing up as a full width button on Windows. Can also rename to just "Copy", like the macOS one above.

I noticed this while looking at how VS Code is [handling their buttons](https://github.com/microsoft/vscode/blob/d890a22dc21fed13cf5a9eed81780d700f6e4f32/src/vs/workbench/services/dialogs/electron-browser/dialogService.ts#L252-L253).

Just handle macOS and Windows for now, once it works for those I'll clone and test on Linux to see how the buttons array should be configured for that. AFAIK there are slight variations in ordering.

develohpanda (changes requested):
LGTM, almost there! A couple of minor comments. 🎉

nijikokun:
Amazing addition, thank you for the PR @Innoxious  👏

develohpanda (approved):
Solid 👍 I'll make a small change later on to make things less verbose and reuse the click handler but the functionality looks good! 💯

PR #2493, opened 2020-08-06
Fix app reloading and insomnia-components in watch mode

### LINKED ISSUE #2491: App reloading not working after Node12 upgrade
After `npm run app-start`, making a change in code is not reflected in the app, even after a full reload.

Currently, a developer needs to run `npm run app-start` again, which is not ideal.

develohpanda:
This is caused by a nested dependency: `webpack > watchpack > chokidar > fsevents`. `watchpack` uses a function in `chokidar` that _requires_ fsevents@~2.1.2 to exist. It is marked as an optional dependency, so lerna/npm doesn't detect or complain.

It also wasn't possible to run `insomnia-components` in watch mode at the same time as `insomnia-app`, because two instances of `styled-components` would be running (similar to the react issue).

This has been a frustrating fix - plenty of digging through lockfiles and writing debug statements into `node_modules` - but **more** frustrating is the need to restart the app with _every_ change.

# Tests
Validate that `npm run clean && npm run bootstrap && npm run app-start` works on
- [x] macOS
- [ ] Windows
- [ ] Linux


develohpanda:
I think intsead of setting styled-components as an external it should be hoisted via lerna. Needs testing.

netlify.toml
```diff
@@ -3,3 +3,4 @@
   base = "packages/insomnia-components/"
   command = "npm run convert-svg && npm run build-storybook"
   ignore = "git diff --quiet HEAD^ HEAD packages/insomnia-components/"
+  environment = { NODE_ENV = "production" }
```
DMarby, line 6: Very nice!

DMarby (approved):

sonicyeti (approved):
Epic sleuthing!!!!

PR #2495, opened 2020-08-06
Paths validation tweak, utils setup.

sonicyeti:
Setting up a utils resource for components, cleaning up object validation.


develohpanda (approved):

packages/insomnia-components/components/sidebar/sidebar-parameters.js
```diff
@@ -17,7 +18,7 @@ export default class SidebarParameters extends React.Component<Props> {
   renderBody = (filter: string): null | React.Node => {
     const { parameters, onClick } = this.props;
 
-    if (Object.prototype.toString.call(parameters) !== '[object Object]') {
+    if (!isObject(parameters)) {
```
develohpanda, line 21: If `parameters` is set to an array in the spec, it will not exit with this conditional because of the `typeof` check within `isObject`, so I see why `Object.prototype.toString.call(parameters) !== '[object Object]'` was used.

If `isObject` needs to use the `Object.prototype...` check, that's fine as well, as long as it's reused and commented as to why it is checked in this way. 👍

dimitropoulos: yep: the behavior of `isObject` (as implemented in this PR) is different:

Screenshot_20210604_124035

dimitropoulos:
this is now more than a year behind.  I'm gonna close it, but keep it in mind for the future once we're in TypeScript (phase 4, I mean).  Most of these checks will not be necessary at that point because the type system will tell us what kinds of objects we are dealing with.

A note for the future:
there appears to be no value of `x` for which `(Object.prototype.toString.call(x) === '[object Array]') === Array.isArray(x)` will return `false`.

PR #2497, opened 2020-08-06
Fix webpack file watching, remove `styled-components` from `insomnia-components` bundle

### LINKED ISSUE #2491: App reloading not working after Node12 upgrade

develohpanda:
Closes #2491; this issue is caused by a nested dependency: `webpack > watchpack > chokidar > fsevents`.

`watchpack` automatically uses a version of `chokidar` depending on the Node version; on going from Node 8 to Node 12, `watchpack` began to use `chokidar` v3 instead of v2.

Chokidar v3 requires `fsevents@~2.1.2` to exist, however, for an unconfirmed reason, `fsevents` was previously manually removed during the bootstrap script in `insomnia-app` - `rimraf ./**/node_modules/fsevents`.

This caused issues with watchpack from running correctly, resulting in code files not being watched and caused changes to not be reflected in the app.

Additionally, it recently became impossible to run `insomnia-components` in watch mode at the same time as `insomnia-app`, because two instances of `styled-components` were found to exist - this is a similar to the requirement of React and react hooks. Just like `react`, [`styled-components` should _not_ be bundled with a library](https://styled-components.com/docs/faqs#with-webpack). This PR also marks `styled-components` as a webpack external.

# Tests
Validate that `npm run clean && npm run bootstrap && npm run app-start` works on
- [x] macOS
- [x] Windows
- [x] Linux


DMarby (approved):

sonicyeti (approved):
Stoked that we'll get proper watch on the components and that this issue, in general, is resolved!

packages/insomnia-components/webpack/webpack.common.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     library: 'insomniaComponents',
     libraryTarget: 'commonjs2',
   },
-  externals: ['react', 'react-dom'],
+  externals: ['react', 'react-dom', 'styled-components'],
```
develohpanda, line 13: As per [guidelines](https://styled-components.com/docs/faqs#with-webpack)

packages/insomnia-app/package.json
```diff
@@ -20,7 +20,7 @@
     "build:sr": "webpack --config send-request/webpack.config.babel.js",
     "package": "node ./scripts/package.js",
     "release": "node ./scripts/release.js",
-    "bootstrap": "rimraf ./**/node_modules/fsevents && rimraf node_modules/graphql-language-service-interface/dist/*.flow && npm run build:sr"
+    "bootstrap": "rimraf node_modules/graphql-language-service-interface/dist/*.flow && npm run build:sr"
```
gschier, line 23: Hopefully this works. The reason it was removed before was because it failed to build in certain environments. I can't remember the exact scenarios but I know I've run into issues in Mac dev and on CI. Perhaps it's resolved now with the Node upgrade and other tweaks 🤞

gschier (approved):
This looks like a great fix if it works! I'll be glad to finally get away from the hack of removing `fsevents` after install.

sonicyeti (approved):
Right on time!

PR #2505, opened 2020-08-11
Add release date to about menu

develohpanda:
- Refactor about menu click handler
- Adjust the order of buttons using `isLinux` flag, to ensure visually they are in the same order.
- Move release date from the settings modal, to the about modal

Linux:
image

macOS:
image


DMarby (approved):
Nice 🎉

sonicyeti (approved):

PR #2507, opened 2020-08-12
Add preinstall step to check node version

develohpanda:
This will ensure users are on the expected version of node, during the `npm run bootstrap` step.

image

### DIFF
--- a/scripts/check-version.js
+++ b/scripts/check-version.js
@@ -8,7 +8,7 @@ const expectedNodeVersion = fs
 
 if (foundNodeVersion !== expectedNodeVersion) {
   console.log(`Current node version: ${foundNodeVersion}`);
-  console.log(`Expect node version: ${expectedNodeVersion}`);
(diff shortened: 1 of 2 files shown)


DMarby (approved):
Awesome! This should with some of the noise from changes to the package.json and lock files we've been seeing due to differing node versions as well :)

scripts/check-version.js
```diff
@@ -0,0 +1,17 @@
...
+
+const foundNodeVersion = process.versions.node;
+const expectedNodeVersion = fs
+  .readFileSync('.nvmrc')
+  .toString()
+  .trim();
+
+if (foundNodeVersion !== expectedNodeVersion) {
+  console.log(`Current node version: ${foundNodeVersion}`);
+  console.log(`Expect node version: ${expectedNodeVersion}`);
```
DMarby: Nitpick, but maybe this should be `Expected` rather then `Expect`?

sonicyeti (approved):
Nice!

PR #2517, opened 2020-08-15
Add tooltip with docs for git authentication key generation

### LINKED ISSUE #2172: Lack of feedback for connecting Insomnia Designer to github
**Describe the bug**
Lack of feedback for connecting Insomnia Designer to github, mainly where does one get the token for github? Which token is it?

Also, if you hit Done, nothing happens. No error, or feedback that you forgot the token or such. 

**To Reproduce**
1. Click on 'Create" on the top right
2. Click on 'Git Clone'
3. What does the token input represent?

**Expected behavior**
a help marker for the authentication token, perhaps linking to githubs documentation. A notification or any kind of text indicating what was incorrectly filled out in the "Configure Repository" form.

 - OS: macOS 10.15.3
 - Installation Method: Download from website
 - App Version: Version 2020.1.3 (2020.1.3.3486)

bartolomej:
I added a tooltip message for the Authentication Token field. Here is a screenshot preview:

Screenshot 2020-08-15 at 13 52 55

It would be useful to add some helpful hint like messages on error dialogs as also mentioned in issue #2172. For example, I forgot to allow permission for the private repositories when creating an access token. The message I got was just "Error 404: Not found", which could be a little bit confusing if you've never created access tokens before.

Closes #2172

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal.js
@@ -7,6 +7,7 @@ import ModalHeader from '../base/modal-header';
 import type { GitRepository } from '../../../models/git-repository';
 import ModalFooter from '../base/modal-footer';
 import * as models from '../../../models';
+import HelpTooltip from '../help-tooltip';
 
 type Props = {||};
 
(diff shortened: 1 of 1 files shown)


develohpanda (approved):
This is an awesome addition, thanks for taking the time to put up this PR! 💯 Congrats on your first PR 🎉  

The Bitbucket documentation linked here is specific to Bitbucket _Server_, post-merge I'll add an additional item for Bitbucket Cloud. We also have a file with [documentation links](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/common/documentation.js) in one central place, so I'll move these items across. 😄

PR #2518, opened 2020-08-16
Add bitbucket cloud docs, add documentation tooltip to git settings modal header

develohpanda:
Follow up to #2517

image

image

### DIFF
--- a/packages/insomnia-app/app/common/documentation.js
+++ b/packages/insomnia-app/app/common/documentation.js
@@ -7,7 +7,7 @@ export const docsBase = insomniaDocs('/');
 export const docsGitSync = insomniaDocs('/article/96-git-sync');
 export const docsGitAccessToken = {
   github:
-    'https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token',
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/common/documentation.js
```diff
@@ -5,3 +5,11 @@ function insomniaDocs(slug: string): string {
 
 export const docsBase = insomniaDocs('/');
 export const docsGitSync = insomniaDocs('/article/96-git-sync');
+export const docsGitAccessToken = {
+  github:
+    'https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token',
```
DMarby: Might be good to not include the `/en/` part of the URL so that we let their localization pick based on browser/user setting

DMarby (approved):

packages/insomnia-app/app/common/documentation.js
develohpanda: Oh, you're right! I've never even thought of that. 💯 Will update.

PR #2519, opened 2020-08-16
Don't try to render swaggerui if the preview is hidden

develohpanda:
What it says on the tin.

Errors in preview are still shown in console if the preview is hidden. Unnecessary.


DMarby (approved):

sonicyeti (approved):
Nice!

PR #2527, opened 2020-08-18
Fixing hover icon positioning, overflow and core layout of header elements

sonicyeti:
Updating the layout for both test and test suit header groupings. Re-positioning hover hint for more flexibility. Demo below:

### DIFF
--- a/packages/insomnia-app/app/ui/css/components/unit-tests.less
+++ b/packages/insomnia-app/app/ui/css/components/unit-tests.less
@@ -39,15 +39,19 @@
         display: flex;
         flex-direction: row;
         align-items: center;
-        padding:0px var(--padding-sm);
+        padding:0px var(--padding-sm) 0px 0px;
 
         button {
           padding: 0px var(--padding-sm);
+          flex: 0 0 auto;
         }
 
         h2 {
           font-size: var(--font-size-lg);
+          flex: 1 1 auto;
           white-space: nowrap;
+          overflow: hidden;
+          text-overflow: ellipsis;
         }
 
         & > * {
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/ui/css/components/unit-tests.less
```diff
@@ -148,8 +161,10 @@
...
     h2 {
-      margin: 0 auto 0 0;
-      font-size: var(--font-size-xl);
+      flex: 1 1 auto;
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
     }
   }
 }
```
develohpanda, line 170: As a reviewer, I have no context about what impact these CSS changes have, being unfamiliar with the actual component make up of `unit-tests`, and this is only a 3 file review.

This is why I'm so in favor of a framework + styled-components. 🙃

develohpanda (approved):
Looks alright. I'm not sure I'm a fan of having the editable hint on the _left_ or the text field, it feels a little off. Would you have any examples of other UIs that do it like this? @erickrico any view on positioning as well as the icon itself (boxy)?

Also, where does the hint show, in the environment modal?

packages/insomnia-app/app/ui/css/components/unit-tests.less
sonicyeti: @develohpanda @erickrico We’re limited on positioning due to the fact that we also can’t have fixed widths or max characters all while still needing to overflow > ellipsis. I’d prefer it on the right :  ) Opted for the box icon over the pencil, on the left the pencil seemed to just float. Something contained seemed more intentional. Not set on this, just providing some context.

Let’s move on bootstrap or tailwind, I’ll start today on unit tests, it’s the perfect entry point for it being the least complex from a UI standpoint!

PR #2528, opened 2020-08-19
Expose styled-components on the window object for UI plugins

develohpanda:
Not setting `styled-components` as an external in a plugin results in the following behavior: 
image

There is no error about `styled-component` duplicates, but uncommenting line 15 being the only change allows for plugins that need `insomnia-components` and/or `styled-components` to not break the app.

image

Curiously, I only see this behavior, right now, during development. 🤷


plugins/insomnia-plugin-kong-portal/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "build": "webpack --config webpack.config.js --display errors-only",
     "watch": "webpack --config webpack.config.js --watch",
-    "bootstrap": "npm run build"
+    "bootstrap": "npm run build",
+    "prepack": "npm run build"
```
DMarby, line 8: Is this used by lerna or some other part of the build process?

packages/insomnia-app/app/ui/index.js
```diff
@@ -76,6 +77,8 @@ if (isDevelopment()) {
   window.db = db;
 }
 
+window['styled-components'] = styledComponents;
```
DMarby, line 84: Perhaps it would be good with a comment here explaining why we are exporting this to the window, for future reference?

plugins/insomnia-plugin-kong-portal/webpack.config.js
```diff
@@ -11,6 +11,9 @@ module.exports = {
     library: 'kongDeploy',
     libraryTarget: 'commonjs2',
   },
+  externals: {
+    'styled-components': 'window styled-components',
```
DMarby, line 20: Would be good with a comment here as well, explaining why we're pulling this in the way we do/that it is exported by the main app.

DMarby (approved):

plugins/insomnia-plugin-kong-portal/package.json
develohpanda: Lerna calls bootstrap as a general "set up this project", but I'll remove `bootstrap` and keep `prepack`.

`prepack` is called after `npm install` anyway, with the added benefit that it will build right before publishing to npm.

develohpanda: For reference, this is what lerna does during the 'npm run bootstrap' step

https://github.com/Kong/insomnia/blob/c5d36a5223efe11f71d2a5a8cee0fea546349a4b/package.json#L14

sonicyeti (approved):
Digging it!

plugins/insomnia-plugin-kong-portal/package.json
DMarby: Fair enough, thanks for explaining! 👍

plugins/insomnia-plugin-kong-portal/package.json
```diff
@@ -18,14 +18,12 @@
     "@babel/preset-env": "^7.9.0",
     "@babel/preset-flow": "^7.9.0",
     "@babel/preset-react": "^7.9.4",
-    "webpack": "^4.42.1",
-    "webpack-cli": "^3.3.11"
-  },
-  "dependencies": {
```
develohpanda, line 24: No more dependencies - all moved to dev dependencies, because webpack bundles what it needs.

packages/insomnia-app/app/ui/index.js
```diff
@@ -76,6 +77,12 @@ if (isDevelopment()) {
   window.db = db;
 }
 
+// Styled components is added to the window object here, for plugins to use.
+// UI plugins built with webpack (such as insomnia-plugin-kong-portal) define styled-components as an external resolved
+// from the window object. This is to ensure there is only one instance of styled-components on the page.
+// Because styled-components are loaded at runtime, they don't have direct access to modules in the electron bundle
```
gschier, line 83: Loading at runtime is one reason (can't reach inside the Webpack bundle), but even if nothing was bundled, the plugin directory is not a descendent of the app, so `require()` will never find anything app-related.

gschier (approved):
Nice work on this @develohpanda. I thought about this a bit, and this seems like this is the best solution going forward. Obviously we'd love for plugin developers to use whatever `styled-component` dependency they want but, since UI-based plugins run in the same DOM as the app, it doesn't seem like it's possible unless we figure out a way to run multiple instances of `styled-components` at once.

Issue #2464, opened 2020-07-29
Unit Testing Design Implementation @sonicyeti


sonicyeti:
We will be performing the following visual and functional tweaks to the existing unit testing MVP.

![unit_testing](https://user-images.githubusercontent.com/52717970/89420367-99e33e80-d700-11ea-9a28-4ed281ed5b35.png)

### Left Rail
**1.**  Integrate new sidebar component in the following format: 
TEST SUITE NAME
	- Test Name
	- Test Name
	- Test Name
TEST SUITE NAME
	- Test Name
**2.** Leverage new sidebar component’s first item contextual menu to replace “New Test Suite” button.

### Center Stage
**3.** ~~Dropdown icon padding (address at component level for global fix)~~
**4.** ~~Horizontally center the lint hint iconography (style like design mode)~~ - **NOT FIXING**
**5.** ~~Max height on testing > overflow scroll 20 lines~~
**6.** ~~Remove dropdown > Play Icon button for Run (NO LABEL)~~
**7.** ~~Remove dropdown > Trash Icon button for Delete (NO LABEL)~~
**8.** ~~Convert existing method dropdown to styled one (leverage method color styling)~~  - **STYLED EXISTING**
**9.** ~~Consistently style the buttons leveraging primary and secondary visual treatment. This issue is more apparent on themes with a dark BG but essentially we have 4 different button styles going currently.~~ 

### Right Rail
**10.** ~~Address padding & line-height issues~~

### Global
**11.** ~~Add a subtle pencil hint on the hover state for all editable input fields. (Environment name via CMD + E, Test & Test Suite Name etc.)~~
**12.** Address issue where long labels on editable fields blows out the layout.
**13.** ~~Address issue where clearing (then saving) and empty label results in the user being unable to see or edit it.~~

### Progress
3, 6, 7, 9: https://github.com/Kong/insomnia/pull/2534
10: https://github.com/Kong/insomnia/pull/2533

### Default Empty State
@erickrico Erick provide me with empty state, mirror Debug

filfreire:
@wdawson, I'm currently cataloguing through all unit test feature related issues.

Marking this one as closed, don't think it's applicable right now.

PR #2533, opened 2020-08-20
Unit testing test output

sonicyeti:
This PR covers the creation of a list group & supporting list group item specifically for the the purpose of our unit testing right rail which is responsible for test result display. Styling provisioning for an expand & collapse chevron has been implemented for next iteration.

Context, see item 10: https://github.com/Kong/insomnia/issues/2464

**Before:**
Screen Shot 2020-08-20 at 8 18 50 AM

**After:**
Screen Shot 2020-08-20 at 8 23 44 AM

**Supporting Story:**
Screen Shot 2020-08-20 at 8 26 13 AM


packages/insomnia-components/components/list-group-item.js
```diff
@@ -0,0 +1,115 @@
...
+
+  p {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    flex-grow: 1;
+  }
+`;
+
+// TODO: Create a collection of visual badges within our component system.
```
develohpanda: We have tags (like packages/insomnia-app/app/ui/components/tags/method-tag.js). Is that similar?

packages/insomnia-components/components/list-group-item.js
```diff
@@ -0,0 +1,115 @@
...
+    border-color: var(--color-success);
+    color: var(--color-success);
+  }
+  &.fail {
+    border-color: var(--color-danger);
+    color: var(--color-danger);
+  }
+`;
+
+class ListGroupItem extends React.PureComponent<Props> {
```
develohpanda: This is named as a really generic component, but it's actually _very_ specific to test results. It almost doesn't need to be within `insomnia-components`, but at the very least I would name it more specifically and/or separate what is needed for a `ListGroupItem` (eg expand/collapse) vs what is needed for a `TestResultItem`

packages/insomnia-components/components/list-group-item.js
```diff
@@ -0,0 +1,115 @@
...
+    flex-direction: row;
+    align-items: center;
+    justify-content: space-between;
+
+    > *:not(:first-child) {
+      margin: var(--padding-xs) var(--padding-xs) var(--padding-xs) var(--padding-xs);
+    }
+  }
+
+  .itemToggler {
```
develohpanda: Could we please remove any direct classes from a styled-component in favor of a standalone styled-component? See `notice-table` for an example.

You can also [extend styles](https://styled-components.com/docs/basics#extending-styles) if need be

packages/insomnia-components/components/list-group-item.js
```diff
@@ -0,0 +1,115 @@
...
+  &.fail {
+    border-color: var(--color-danger);
+    color: var(--color-danger);
+  }
+`;
+
+class ListGroupItem extends React.PureComponent<Props> {
+  render() {
+    const { errorMsg, result, name, time } = this.props;
+    return (
```
develohpanda: Great candidate for a functional component 😄

packages/insomnia-components/components/list-group.js
```diff
@@ -0,0 +1,23 @@
...
+  children?: React.Node,
+|};
+
+const StyledListGroup: React.ComponentType<{}> = styled.ul`
+  list-style-type: none;
+  margin: 0px;
+  padding: 0px var(--padding-xs) 0px 0px;
+`;
+
+class ListGroup extends React.PureComponent<Props> {
```
develohpanda: Great candidate for a functional component 😄

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -327,19 +336,25 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
...
+                      time={t.duration}
+                      result="fail"
+                      errorMsg={t.err.message}
+                    />
                   )}
-                </li>
+
+                  {!t.err.message && (
+                    <ListGroupItem key={i} name={t.title} time={t.duration} result="pass" />
+                  )}
```
develohpanda: This logic can (and should, IMO), be inside a `TestResultListItem` component.

develohpanda (commented):
Couple of notes :)

packages/insomnia-components/components/list-group-item.js
sonicyeti: Similar in theory but the design is very far off and more appropriate for accent type usages (currently). My hope is that we can craft a nice array of badges similar to other frameworks.

packages/insomnia-components/components/list-group.stories.js
```diff
@@ -0,0 +1,26 @@
...
+import ListGroupItem from './list-group-item';
+
+export default { title: 'Lists | List Group' };
+
+export const _default = () => (
+  <div style={{ width: '400px' }}>
+    <ListGroup>
+      <ListGroupItem
+        name="My test name one"
+        time="602ms"
```
DMarby: Nitpick, but imo it looks kind of strange to have the unit directly after the number, I think adding a space to make it `602 ms` would make more sense, same for all the other items.

packages/insomnia-components/components/list-group-item.js
```diff
@@ -0,0 +1,115 @@
...
+      <StyledListItem>
+        <div>
+          <div className="itemToggler">
+            {/* 
+            TODO: Add show/hide toggle support in the future.
```
DMarby: Imo it's nice to avoid leaving comments/commented code like this as much as possible, if we want something to be done in the future, we should open an issue for it to have it properly tracked instead.

Not sure what the common approach is in this codebase however, but that's my two cents.

packages/insomnia-components/components/list-group-item.js
```diff
@@ -0,0 +1,115 @@
...
+            errorMsg && <SvgIcon icon={IconEnum.chevronDown}></SvgIcon>}
+            */}
+          </div>
+          <StyledBadge className={result}>{result}</StyledBadge>
+          <p>{name}</p>
+          <div className="timestamp">
+            <SvgIcon icon={IconEnum.clock}></SvgIcon>
+            <span>{time}</span>
```
DMarby: Since this is just outputting the raw input, currently, we already have two different results here:
- The storybook has strings in the format of `100ms`
- The actual UI has integers, resulting in just `100`

I think it would be better to just always pass in milliseconds or a duration type or such to the component, and have the presentation formatting be done by the component itself, this way it's always consistent.
With presentation formatting I mean either doing something like `%d ms` which is equivalent to what we had before this change, or by using a library that rounds up to seconds/minutes/etc.

DMarby (changes requested):

packages/insomnia-components/components/list-group-item.js
develohpanda: To add to that, I suggest naming the prop as `timeMilliseconds` or `timeMs`. Almost everything else in JS operates with ms, so lets stick with that convention, both with naming and with logic.

DMarby (approved):

packages/insomnia-components/components/list-group/list-group.stories.js
```diff
@@ -0,0 +1,58 @@
...
+
+export const _unitTestResults = () => (
+  <div style={{ width: '350px' }}>
+    <ListGroup>
+      {tests.map((test, i) => (
+        <UnitTestResultItem item={test} key={i} />
+      ))}
+    </ListGroup>
+  </div>
+);
```
develohpanda, line 58: Epic!

packages/insomnia-components/components/list-group/unit-test-result-badge.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import * as React from 'react';
+import styled from 'styled-components';
+
+type Props = {|
+  result?: String,
```
develohpanda: I suggest changing this prop to be more explicit

```suggestion
  failed?: boolean,
```

packages/insomnia-components/components/list-group/unit-test-result-item.js
```diff
@@ -0,0 +1,54 @@
...
+    text-overflow: ellipsis;
+    flex-grow: 1;
+  }
+`;
+
+const UnitTestResultItem = ({ item }: Props) => {
+  return (
+    <StyledResultListItem>
+      <div>
+        <UnitTestResultBadge result={item.err.message} />
```
develohpanda: Then updating this usage:

```suggestion
        <UnitTestResultBadge failed={!!item.err.message} />
```

packages/insomnia-components/components/list-group/unit-test-result-item.js
```diff
@@ -0,0 +1,54 @@
...
+        <p>{item.title}</p>
+        <UnitTestResultTimeStamp timeMs={item.duration} />
+      </div>
+      {item.err.message && <code>{item.err.message}</code>}
+    </StyledResultListItem>
+  );
+};
+
+export { UnitTestResultItem, UnitTestResultBadge, UnitTestResultTimeStamp, SvgIcon };
```
develohpanda: Do all of these still need to be exported? I would think we only need to export `UnitTestResultItem` now.

If all do need to be exported, careful about re-exporting `SvgIcon`

packages/insomnia-components/components/list-group/unit-test-result-timestamp.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+import * as React from 'react';
+import styled from 'styled-components';
+import SvgIcon from '../svg-icon';
+
+type Props = {|
+  timeMs: String,
+  children?: React.Node,
```
develohpanda: `children` prop can be removed now 😄

packages/insomnia-components/index.js
```diff
@@ -7,13 +7,16 @@ import _DropdownDivider from './components/dropdown/dropdown-divider';
...
+import _ListGroup from './components/list-group/list-group';
+import _ListGroupItem from './components/list-group/list-group-item';
 import _MultiSwitch from './components/multi-switch';
 import _NoticeTable from './components/notice-table';
 import _SvgIcon from './components/svg-icon';
 import _Sidebar from './components/sidebar/';
 import _RadioButtonGroup from './components/radio-button-group';
 import _Switch from './components/switch';
 import _ToggleSwitch from './components/toggle-switch';
+import * as unitTestResultItem from './components/list-group/unit-test-result-item';
```
develohpanda: After updating the exports from `unit-test-result-item`, we should be able to change this back to follow the same pattern as the other exports.

develohpanda (changes requested):
Almost there!! A couple of minor notes 😄 looking really good now.

packages/insomnia-components/components/list-group/unit-test-result-item.js
sonicyeti: Thanks for the catch!

develohpanda (approved):

PR #2534, opened 2020-08-20
Unit tests buttons & select input styling

sonicyeti:
This PR covers the stylistic tweaks to various buttons found in the upper portions of all major portions in addition to the styling of the operation selection menu.

For context, [please see 3, 6, 7 & 9 here](https://github.com/Kong/insomnia/issues/2464).   

**Before:**
Screen Shot 2020-08-20 at 9 47 29 AM

**After:** 
Screen Shot 2020-08-20 at 9 46 52 AM

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper-unit-test.js
+++ b/packages/insomnia-app/app/ui/components/wrapper-unit-test.js
@@ -442,6 +442,7 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
               size="default"
               disabled={testsRunning}>
               {testsRunning ? 'Running... ' : 'Run Tests'}
+              <i className="fa fa-play space-left"></i>
             </Button>
           </div>
           {activeUnitTests.map(this.renderUnitTest)}
@@ -460,7 +461,7 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
       <ErrorBoundary showAlert>
         <div className="unit-tests__sidebar">
           <div className="pad-sm">
-            <Button variant="outlined" bg="surprise" onClick={this._handleCreateTestSuite}>
+            <Button variant="outlined" onClick={this._handleCreateTestSuite}>
               New Test Suite
             </Button>
           </div>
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -446,6 +442,7 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
               size="default"
               disabled={testsRunning}>
               {testsRunning ? 'Running... ' : 'Run Tests'}
+              <i className="fa fa-play space-left"></i>
```
develohpanda, line 445: Strange that we're using font-awesome here but an SVG further up. Is it possible for both to use the same thing? You should be able to add `<SvgIcon>` here, because the SVG, when passed through svgr, should respect font sizes.

sonicyeti: Agreed however in this button, theming color changes aren't picked up. Similar with the hover state on editable, can't use SVG in a content pseudo-class while supporting themes. We can probably go back and provision all the themes to support it (I think).

develohpanda: Confused - if theming isn't linked with SVGs, does that mean the "run test" button doesn't change color, while the "run all tests" button, does? Or vice-versa. 🤔 

In any case, I'd expect all run buttons to (or not to) change colors on theme changes, consistently.

sonicyeti: Vice-versa. The end result is as expected, the issue here is that without opening back up theming for extension, in order to get color changing iconography inside a primary themed button, FA is used. Similar to the git branch icon found in the upper right.

Example:
Left play icon is FA, right is SVG.

Screen Shot 2020-08-20 at 5 50 03 PM

Screen Shot 2020-08-20 at 5 53 15 PM

DMarby (approved):

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
develohpanda: Aha, thanks for explaining that. Makes sense. That's okay for now, since most of our icons that need to exist both in FA and in SVG format are pretty much identical.

We can look to tackle this when we need to have a custom non-fa icon used in this manner! 👍

develohpanda (approved):
🔥

Issue #2535, opened 2020-08-20
Add spectron for automated e2e testing


develohpanda:
- [x] Add spectron with a basic test #2531
- [x] Run with installed app #2531
- [x] Run with packaged app (after `npm run app-package`) #2531
- [x] Run with built app (after `npm run app-build`) #2531 #2537
- [x] Mock local server to query against #2544 
- [x] Add Core to PR pipeline #2536 
- [x] Run tests specific to Core vs Designer #2545 
- [x] Add Designer to PR pipeline #2549
- [x] Add to Linux PR pipeline #2536
- [ ] Add to Release pipeline - package and test
- [ ] Convert to ES6 style imports (investigate TS for the smoke-testing project)
- [ ] Update `userData` to set the app data directory - https://github.com/electron-userland/spectron/blob/master/test/commands-test.js#L214-L217

PR #2536, opened 2020-08-20
Add e2e Core tests to PR workflow

develohpanda:
As per title.

Related to #2535

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,12 +11,11 @@ on:
       - synchronize
 
 jobs:
-  Windows:
-    name: Test Windows
-    runs-on: windows-latest
+  Mac:
+    name: Mac
+    runs-on: macos-latest
     steps:
-      - name: Checkout branch
-        uses: actions/checkout@v1
+      - uses: actions/checkout@v1
       - name: Read Node version from .nvmrc
         run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
         id: nvm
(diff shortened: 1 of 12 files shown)


packages/insomnia-app/scripts/getBuildContext.js
```diff
@@ -0,0 +1,48 @@
...
+  const channel = tagMatch ? tagMatch[4] : 'stable';
+
+  return {
+    app,
+    channel,
+    version,
+    gitRef,
+    gitCommit,
+  };
+}
```
develohpanda, line 48: Taken from `packages/insomnia-app/scripts/build.js`

.github/workflows/test.yml
```diff
@@ -11,6 +11,25 @@ on:
       - synchronize
 
 jobs:
+  Mac:
+    name: Test Mac
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v1
```
DMarby: Nitpick, just to make it consistent with windows/linux
```suggestion
      - name: Checkout branch
        uses: actions/checkout@v1
```

packages/insomnia-smoke-test/spec.js
```diff
@@ -3,6 +3,22 @@ const Application = require('spectron').Application;
 const electronPath = require('../insomnia-app/node_modules/electron');
 const path = require('path');
 
+// export function isMac() {
```
DMarby: Are these meant to be used for anything, or can we remove this commented code?

packages/insomnia-app/scripts/release.js
```diff
@@ -15,11 +15,15 @@ const octokit = new Octokit({
 if (require.main === module) {
   process.nextTick(async () => {
     try {
-      const buildContext = await buildTask.start();
+      const buildContext = await buildTask.start(true);
+      if (buildContext.smokeTest) {
```
DMarby: As far as I can tell, this can never happen, since we're setting the forceFromGitRef argument to true on the line above?

develohpanda: Whoops, yes! I initially didn't have that param, hence put the check in here. I think through the frustration of working with Linux I forgot to self-review this. 😂

Great catch though, the condition needs to 🔥!

packages/insomnia-smoke-test/spec.js
develohpanda: Will be needed for testing packages. 

Build artifacts are common across OS, but package artifacts are unique per OS, so these checks will be needed to determine the correct directories. I just left them in there since it's code we'll need fairly soon, but will remove. 😅 Copy-pasta from insomnia-app.

DMarby (approved):

sonicyeti (approved):
LGTM!

PR #2537, opened 2020-08-21
Build and package to a folder prefixed by the application id

develohpanda:
This will allow us to concurrently build and smoke test both applications, as their artifacts will end up in different directories instead of overwriting each other.

image

- I was able to `npm run app-start` - development worked fine
- I was able to `npm run app-build` - see the artifacts in the expected directory and run smoke tests on them
- I was able to `npm run app-package` - see the artifacts in the expected directory and run smoke tests on them

Related to #2535

### DIFF
--- a/packages/insomnia-app/config/electronbuilder.core.json
+++ b/packages/insomnia-app/config/electronbuilder.core.json
@@ -9,7 +9,7 @@
       "filter": "yarn-standalone.js"
     },
     {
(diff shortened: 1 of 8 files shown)


daelid237:
how tips and trick secrete to get apps pinjaman online from  goole play?

DMarby (approved):

PR #2557, opened 2020-08-26
Handle server-side OAuth2 redirects to non-existent protocols

GyroJoe:
The Microsoft OAuth2 service (i.e. https://login.microsoftonline.com/common/oauth2/authorize) handles transitioning to the final redirect URI via a 302 redirect (in some cases at least). When using a non-existent protocol as the redirect URI like `urn:ietf:wg:oauth:2.0:oob`, none of the existing events were being triggered.

In the dev tools I could clearly see a load failing, but for some reason, `did-fail-load` isn't being triggered.

Added a listener for the [`will-redirect` event on webContents](https://www.electronjs.org/docs/api/web-contents#event-will-redirect), which is able to catch the redirect and complete the OAuth2 flow.


packages/insomnia-app/app/network/o-auth-2/misc.js
```diff
@@ -99,6 +99,10 @@ export function authorizeUserInWindow(
       _parseUrl(url);
     });
 
+    child.webContents.on('will-redirect', (e, url) => {
```
develohpanda: I'm not too familiar with the specific variations between different OAuth2 providers, but adding an additional event hook could potentially cause conflicts with the `did-navigate` event, for redirects that have _valid_ protocols and do redirect correctly.

If we were to **replace** `did-navigate` with `will-redirect`, rather than the two event hooks co-existing, in theory that should cater for other OAuth2 providers as well. Is this something you are able to help validate/test?

GyroJoe: I can do some testing with a few providers and see what the behavior is with that setup.

My original thinking with this approach was that hooking more events is preferable since this code doesn’t change the behavior of the frame (it’s not cancelling or rejecting any operation) - it’s just looking for a match to the regex. It could cause a problem though, if the regex catches something that isn’t actually the final OAuth code.

develohpanda: > I can do some testing with a few providers and see what the behavior is with that setup.

Thanks! 🤗 

If things work as expected with just the single event (`will-redirect`), I think I would favor that for the sake of one source of truth for redirect handling.

If we definitely need both events, again worth testing if there are any race conditions in that setup, and adding sufficient code comments to explain why there are two hooks for handling redirects.

GyroJoe: Did some further testing with various events and providers. Unfortunately I couldn't find an event that worked universally.

`will-redirect` doesn't get triggered for GitHub's flow and `did-navigate` doesn't get triggered for Microsoft's. I also tried `did-start-navigation`, but that didn't work universally either.

Added some explanatory comments to detail the reasoning behind the multiple events, and added the event source to logging to make things easier to diagnose in the future.

develohpanda: Thanks for testing! Good idea with adding the event source, happier with this approach as it can help debug if things don't quite work with other providers. 👍

develohpanda (approved):
LGTM!

DMarby (approved):

nijikokun (approved):

packages/insomnia-app/app/network/o-auth-2/misc.js
```diff
@@ -91,12 +91,19 @@ export function authorizeUserInWindow(
...
       const currentUrl = child.webContents.getURL();
-      _parseUrl(currentUrl);
+      _parseUrl(currentUrl, 'did-navigate');
+    });
+
+    child.webContents.on('will-redirect', (e, url) => {
+      // Also listen for will-redirect, as some redirections do not trigger 'did-navigate'
+      // 'will-redirect' does not cover all cases that 'did-navigate' does, so both events are required
+      // GitHub's flow triggers only 'did-navigate', while Microsoft's only 'will-redirect'
+      _parseUrl(url, 'will-redirect');
```
nijikokun, line 101: The only comment I have is that the previous event pulls the `url` using `getURL` even though the [event passes the url](https://www.electronjs.org/docs/api/web-contents#event-did-navigate). Should we do the same in these as well?

GyroJoe: It was like that way back when this code was originally written also, so I'm not sure why this was done:
https://github.com/electron/electron/blob/v1.6.2/docs/api/web-contents.md#event-did-navigate

For `will-redirect`, the URL won't change at the point the event fires (and may not change at all, if the webContents can't load the URL). 

Since the old code has been working as-is for a while now, I'm inclined to leave it as-is for now.

develohpanda:
Quick note that this looks good, I will merge it in early next week 😄

PR #2559, opened 2020-08-26
fixes #2552:addded a (slightly) more verbose output in check-version.js

### LINKED ISSUE #2552: Improve the output of the node version check script
image

Currently the response at the top is overpowered by the npm error lines. It would be nice to increase the visibility of the output through console log text colors: https://stackoverflow.com/a/41407246.

Note we cannot use an npm module like chalk for this, as the script runs and errors _before_ node_modules are installed.

yashshah1:
* updated version script to show more verbose errors in case of a mismatch node version

Closes #2552

### DIFF
--- a/scripts/check-version.js
+++ b/scripts/check-version.js
@@ -1,16 +1,28 @@
 const fs = require('fs');
 
+const GREEN_FG = '\x1b[32m';
+const RED_FG = '\x1b[31m';
+const UNDERSCORE = '\x1b[4m';
+const BRIGHT = '\x1b[1m';
+const RESET = '\x1b[0m';
+
 const foundNodeVersion = process.versions.node;
 const expectedNodeVersion = fs
   .readFileSync('.nvmrc')
   .toString()
   .trim();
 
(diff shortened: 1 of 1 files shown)


develohpanda (commented):
Nicely done! Could you also please add a screenshot showing what the new output looks like?

yashshah1:
Hey, here you go!

image

develohpanda:
Awesome! Nicely done with adding a link to nvm👌

Two minor points:
- in the case a user does have the correct version installed, we don't want to print anything. So I suggest changing the opening line to something like "Error - incorrect Node version installed!" when a mismatched version is found. Colour this as you see fit!
- the closing line still seems to blend in, and arguably it is the most important - could we please change that to a yellow, for example? Any colour will do to really make it stand out. 😊

yashshah1:
Hey, hope this is a little better!


image

develohpanda (approved):
Looking really good now, thanks for working on this @yashshah1 🎉 And congrats on your first PR into Insomnia! 😄

PR #2566, opened 2020-08-28
Mark kong plugins as private

develohpanda:
We don't want to publish these until the next version of the app has been released to production.


DMarby (approved):
@develohpanda Maybe we should put a card in the top of the backlog or something to publish these after the next release, so that we don't forget?

develohpanda:
> @develohpanda Maybe we should put a card in the top of the backlog or something to publish these after the next release, so that we don't forget?

Added https://github.com/Kong/insomnia/issues/2570

PR #2568, opened 2020-08-29
fix: trigger linting with existing value on mount

### LINKED ISSUE #2456: Lint errors don't show in editor gutter after reopening document
**Describe the bug**
I have a specification which contains errors. When I navigate to home and reopen the document, or navigate to debug/test and come back to design, the gutter does not show lint errors. It refreshes the errors, after I start typing.

**To Reproduce**
1. Introduce lint error into spec
2. Observe errors in gutter
3. Click on "Documents" to go 
4. Click on the document again to open it
5. Observe no errors in gutter
5. Start typing
5. Observe errors in gutter

**Expected behavior**
Should always show errors in gutter if they exist

**Screenshots**

 - OS: 10.15.6
 - Installation Method: dev
 - App Version 2020.3.3

rbarbazz:
Aug-28-2020 22-36-04

Closes #2456


develohpanda (commented):
Thanks for working on this! I did some testing locally and it looks great. 💯

I won't merge it _just_ yet however, as I need to do some more research about potential side-effects of this change to other areas of the app, because the code-editor is such a central component. INS-119.

rbarbazz:
@develohpanda  Sounds good, thanks for reviewing!

develohpanda:
Hi @rbarbazz! Finally had a chance to look through this one, and my goodness I've never spent so much time reviewing a two line PR 😅 

Here's what my findings are:
- Linting happens as part of `_codemirrorValueChanged()`
- The only time linting is _skipped_, is when the codemirror value is changed and `this._ignoreNextChange` is set to `true`
- This is set to true during `_codemirrorSetValue(defaultValue || '')`, which happens during initialization in `_handleInitTextarea`

I think the root cause may be that we should not ignore _linting_ on change, we should only ignore calling `this.props.onChange`. In the following method, I achieved the same result by moving the exit conditional from the beginning, to just before `this.props.onChange(value)` is called. By doing this, I could even remove the `defaultValue` attribute from the `<textarea />` and saw the same behavior.

https://github.com/Kong/insomnia/blob/4885d00e0f71baf74e54eb75f1984f1d299576a2/packages/insomnia-app/app/ui/components/codemirror/code-editor.js#L831-L854

I dug deeper into this because I can't quite figure out why the `defaultValue` attribute in the `textarea` was previously a single space, and can't quite figure out why your setting defaultValue there was working.

Am I missing something that you found during your investigation?

rbarbazz:
Hi @develohpanda,

Thanks for taking the time to investigate.

You definitely dug far deeper than I did. Initially I thought it was odd that the default value wasn't set when it was provided in the props. I just noticed that the first value that was linted was a single space.

It looks like your solution is a lot more solid than mine, considering there is a proper demonstration behind it.

> I dug deeper into this because I can't quite figure out why the defaultValue attribute in the textarea was previously a single space, and can't quite figure out why your setting defaultValue there was working.

I have spent some time tonight trying to figure this one out but couldn't 😬

Thanks again for the detailed explanation and apologies for the time you had to spend investigating this.

develohpanda:
No stress! 🤗 We recently had a breaking bug through a seemingly small change in the `code-editor` component, so just wary about exploring all avenues with changes here. Ideally it would be broken down a bit more, with some degree of unit tests - that's definitely a work in progress. My time is dedicated to ensuring a stable solution, and if an extensive review is what that requires, that's no issue _at all_. 😄 

All things considered, I would feel a bit more comfortable with moving the exit condition in `_codemirrorValueChanged` to check and exit after linting, as there isn't really a case to skip linting anytime the value changes (or is set programmatically). Or, to make things even clearer, new functions to execute linting and conditionally trigger the `props.onChange` callback.

After that, I think it would be best to explore removal of the `textarea.defaultValue` property entirely - it is quite misleading as setting the `defaultValue` is already handled by `_handleInitTextarea`.

Admittedly I hadn't investigated the root cause on initially creating the issue. I really appreciate you picking this up and working through it!

rbarbazz:
I like the project and the codebase so I'm happy to help!

> All things considered, I would feel a bit more comfortable with moving the exit condition in `_codemirrorValueChanged` to check and exit after linting, as there isn't really a case to skip linting anytime the value changes (or is set programmatically). Or, to make things even clearer, new functions to execute linting and conditionally trigger the `props.onChange` callback.
> 
> After that, I think it would be best to explore removal of the `textarea.defaultValue` property entirely - it is quite misleading as setting the `defaultValue` is already handled by `_handleInitTextarea`.

Do you want me to update the PR with those changes or were you planning on making a separate PR yourself?

develohpanda:
I'm glad you like it! It's a bit of a beast; you'll notice we recently added e2e integration tests (in `packages/insomnia-smoke-test`) as a way to get quick, broad test coverage. I'll be adding some more component specific tests in the future so we can make changes with more confidence.

> Do you want me to update the PR with those changes or were you planning on making a separate PR yourself?

Feel free to update this PR! I don't plan on doing it separately, and I probably won't be able to get to it soon either 🤗

rbarbazz:
I'll update the PR when I have a chance!

develohpanda:
> I'll update the PR when I have a chance!

👍 No rush! Take your time

rbarbazz:
Sorry for the late update, I won't have time to follow up on this.

I checked `develop` locally today and couldn't reproduce the bug. It might have been fixed here https://github.com/Kong/insomnia/pull/2733/files#diff-111d80360a042255b5795cbbd547911cd7ffe089a551617a57ba92e2a4daaa9aR1271?

develohpanda:
Thank you for updating the PR @rbarbazz! Have a wonderful holiday period 🎉

rbarbazz:
You too! 🎄

PR #2572, opened 2020-08-31
Fix netlify/storybook deployment issues

develohpanda:
See comments further down


netlify.toml
```diff
@@ -2,5 +2,4 @@
   publish = "storybook-static"
   base = "packages/insomnia-components/"
   command = "npm run convert-svg && npm run build-storybook"
-  ignore = "git diff --quiet HEAD^ HEAD packages/insomnia-components/"
```
develohpanda, line 5: From netlify logs:

image

Trying a commit with the `--` separator still doesn't produce the expected result; netlify _always_ skips a deployment regardless of if `insomnia-components` changed.

Removing this until we have an option that works as expected.

packages/insomnia-components/.nvmrc
```diff
@@ -0,0 +1 @@
+12.18.3
```
develohpanda, line 1: Netlify wasn't picking up `.nvmrc` in the root of the repo because there is an override for the `base` directory in `netlify.toml`.

Instead of setting the repo root to be netlify's root, and having it install unnecessary dependencies, it feels faster and more logical to just duplicate `.nvmrc`. **I am open to other suggestions.**

Netlify has failed to install node 10 for recent builds, but I'm not debugging this because it shouldn't be using node 10 anyway - https://app.netlify.com/sites/insomnia-storybook/deploys/5f4b8abf98d2860008bb91b6
image

netlify.toml
```diff
@@ -2,5 +2,5 @@
...
+  ignore = "git diff --quiet HEAD^ HEAD ./"
```
develohpanda, line 5: Because we are already in `packages/insomnia-components`

Ignoring commits wasn't working as it should have previously. This should reduce our Netlify build minutes

packages/insomnia-components/.nvmrc
DMarby: I think this should be fine, can't really come up with a better solution either.

DMarby (approved):

PR #2586, opened 2020-09-04
Save and pre-select last used export format

### LINKED ISSUE #1592: [Feature Request] Remembering the format for the exported file
- Insomnia Version: 6.5.4 (6.5.4.2876)
- Operating System: macOS 10.14.5

## Details

I often export the workspaces. And I feel a little discomfort. Can you, please, to make memorizing the last selected format to export to this place?

For example, I prefer YAML. And I always have to do extra activities. No problem, when use this functionality rarely. But when often...

Снимок экрана 2019-07-08 в 17 16 59

tobiasr:
Closes #1592 

This is a very small fix. Please let me know if something should be changed.

I think placing the `.setItem` call before calling onDone is ok, but it might be that this call shouldn't happen until a file have been successfully exported?


develohpanda (approved):
Looks good! Thank you for doing this, apologies for the delay in review 🤗

> I think placing the .setItem call before calling onDone is ok

I think this is fine as well, since it is only storing the preference

PR #2598, opened 2020-09-10
Editor toggling, visual design spec. integration, various stylistic tweaks

### LINKED ISSUE #2562: Unit Testing > UI Lag
**Describe the bug**
When multiple test suites exist, each containing multiple tests the Unit Testing UI is slow to initially render.

**To Reproduce**
See replication steps [here](https://github.com/Kong/insomnia/issues/2562#issuecomment-681013144).

**Expected behavior**
Near zero lag.

image

sonicyeti:
This PR covers the incorporation of accordion style functionality for unit tests including: 

- Unit tests can now be toggled to expand or collapse
- A single unit test's max height caps at around the 20 line mark, where scrolling takes over to preserve real estate
- The unit test header has been brought closer to alignment with the original design specification
- Unit Test is now a list group item type from a component standpoint
- Swapping play and delete iconography positioning
- Resolution for multiple Code Mirror instances introducing lag

**Screenshot:**
Screen Shot 2020-09-10 at 10 00 43 AM

**Demo:**

Closes #2562
Fixes INS-85
Fixes INS-48
Fixes INS-52


DMarby (approved):

packages/insomnia-app/app/ui/components/unit-test-selection.js
```diff
@@ -0,0 +1,29 @@
+// @flow
+import React from 'react';
+
+type Props = {
+  onChange: Function,
+  unitTest: Object,
+  selectableRequests: Object,
```
develohpanda: It would be great if we could type these more explicitly.

For example, just from looking at the usage in the component
```suggestion
  onChange: () => void,
  selectedRequestId?: string,
  selectableRequests: Array<{name: string, request: { _id: string }}>,
```

packages/insomnia-app/app/ui/components/unit-test-selection.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export default function UnitTestSelection({ onChange, unitTest, selectableRequests }: Props) {
```
develohpanda: Suggestion to rename component as `UnitTestRequestSelector`, as this is specific to selecting requests 😄

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -351,45 +352,28 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
...
+      <UnitTestItem
+        item={unitTest}
+        key={unitTest._id}
+        testSelection={
+          <UnitTestSelection
+            onChange={this._handleSetActiveRequest.bind(this, unitTest)}
+            unitTest={unitTest}
+            selectableRequests={selectableRequests}
+          />
+        }
```
develohpanda: I like that you extracted `UnitTestItem` and `UnitTestSelection` out into standalone components 👍 

We can avoid having to pass `<UnitTestSelection />` as a prop into `UnitTestItem`; instead, we can simply pass the onChange handler as a prop, for instance, `onRequestSelectionChanged={this._handleSetActiveRequest.bind(this, unitTest)}`.

Then, `UnitTestItem` becomes

```js
function UnitTestItem(props) {
  ...
  <UnitTestSelection onChange={props.onRequestSelectionChanged} ... />
  ...
}
```
instead of
```js
function UnitTestItem(props) {
  ...
  {props.testSelection}
  ...
}
```

This way UnitTestSelection can be moved into insomnia-components!

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -351,45 +352,28 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
...
+        onDeleteTest={() => this._handleDeleteTest(unitTest)}
+        onRunTest={() => this._handleRunTest(unitTest)}
```
develohpanda: Passing arrow functions as props mean a new function is instantiated on each render. A newly instantiated function, even if it has the same implementation, is unique. Because it is unique, React will tell `UnitTestItem` to update, because a prop has changed, when in-fact it hasn't. Try use the `.bind` syntax within class components, or `React.useCallback` within function components 😄

packages/insomnia-components/components/list-group/list-group.stories.js
```diff
@@ -47,10 +48,47 @@ const tests = [
...
+    code:
+      'const response1 = await insomnia.send();\nexpect(response1.status).to.equal(404);\nconst response1 = await insomnia.send();\nexpect(response1.status).to.equal(404);\nconst response1 = await insomnia.send();\nexpect(response1.status).to.equal(404);\nconst response1 = await insomnia.send();\nexpect(response1.status).to.equal(404);\nconst response1 = await insomnia.send();\nexpect(response1.status).to.equal(404);',
+  },
+];
+
+export const _unitTests = () => (
+  <div style={{ width: '100%' }}>
+    <ListGroup>
+      {unitTests.map((test, i) => (
+        <UnitTestItem item={test} key={i}>
```
develohpanda: Index as keys hehe. Doesn't really matter here, but good to get into the practice of avoiding it. I suggest to use `test._id` as the key

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -351,45 +352,28 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
...
+        testsRunning={testsRunning}
+        testNameEditable={
+          <UnitTestEditable
+            onSubmit={this._handleChangeTestName.bind(this, unitTest)}
+            value={unitTest.name}
+          />
+        }>
```
develohpanda, line 371: Same as above, but less trivial because `Editable` isn't in `insomnia-components`. _But_, it would be a great idea to pull `Editable` into the component library, and other usages can be updated later on.

packages/insomnia-components/components/list-group/unit-test-item.js
```diff
@@ -0,0 +1,108 @@
...
+const UnitTestItem = ({
+  item,
+  children,
+  onDeleteTest,
+  onRunTest,
+  testSelection,
+  testNameEditable,
+  testsRunning,
+}: Props) => {
+  const [collapse, toggleCollapse] = React.useState(false);
```
develohpanda: We can use the `useToggle` hook from `react-use`, like in the new sidebar hehe

packages/insomnia-components/components/list-group/unit-test-item.js
```diff
@@ -0,0 +1,108 @@
...
+      <div>
+        <Button
+          onClick={() => toggleCollapse(!collapse)}
+          variant="text"
+          style={!collapse ? { transform: `rotate(${toggleRotation}deg)` } : {}}>
+          <SvgIcon icon="chevron-down" />
+        </Button>
+        <Button variant="text" disabled>
+          <SvgIcon icon="file" />
+        </Button>
```
develohpanda, line 79: Does this icon need to exist inside a disabled `Button`? I'm guessing it's for consistent spacing

develohpanda (changes requested):
Nicely done 😄 

I've made a couple of comments, but one overarching one is to do with defining Props for components. Because props are the interface between a component and it's consumer, we should try to be as explicit as possible with the prop type definitions, rather than generic `Function`, `Object`, etc. It will only help us when we come to look at or consume the same component in a few months time.

I will add that there are many components currently that do not define their prop types clearly; those components were likely created before a typing methodology was introduced. Even though we are currently using Flow right now, when we eventually look to move towards TS, the flow type definitions will be useful. 😄

packages/insomnia-components/components/list-group/unit-test-item.js
sonicyeti: Exactly. Disabled happened to match the design spec while reducing custom styling overhead.

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -354,20 +353,18 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
     const { testsRunning } = this.state;
     const selectableRequests = this.buildSelectableRequests();
 
+    console.log(selectableRequests);
```
DMarby: I think this might be a leftover debug statement that should be removed?

sonicyeti: Thanks for the catch, pulling it now!

packages/insomnia-components/components/list-group/list-group.stories.js
```diff
@@ -73,12 +77,119 @@ const unitTests = [
...
-        <UnitTestItem item={test} key={i}>
-          <div>Toggle Content...</div>
+        <UnitTestItem
+          testNameEditable={test.name}
+          item={test}
+          key={test._id}
+          onChange={_handleRequestSelected}
+          selectableRequests={unitTestRequests}>
+          <div style={{ padding: '10px 15px', border: '1px solid #eee', margin: '10px' }}>
+            CodeMirror instance.
```
develohpanda, line 191: 👌

packages/insomnia-components/components/list-group/list-group.stories.js
```diff
@@ -89,7 +200,7 @@ export const _unitTestResults = () => (
   <div style={{ width: '350px' }}>
     <ListGroup>
       {unitTestResults.map((test, i) => (
-        <UnitTestResultItem item={test} key={i} />
+        <UnitTestResultItem item={test} key={test._id} />
```
develohpanda, line 203: 👌

packages/insomnia-components/components/list-group/unit-test-item.js
```diff
@@ -63,27 +49,36 @@ const UnitTestItem = ({
...
+  selectedRequestId,
+  selectableRequests,
 }: Props) => {
-  const [collapse, toggleCollapse] = React.useState(false);
-  const toggleRotation = -90;
+  const [isToggled, toggle] = useToggle(false);
+  const toggleIconRotation = -90;
+  const _handleItemToggle = React.useCallback(() => {
+    toggle();
+  }, [toggle]);
```
develohpanda: I think this can be removed in favor of just using `onClick={toggle}`

The `React.useCallback` would be useful if you were to do `onClick={() => { toggle() }}`, because the `() => {}` syntax creates a new function. 🙂

packages/insomnia-components/components/list-group/unit-test-item.js
```diff
@@ -1,41 +1,27 @@
...
   item: Object,
   children?: React.Node,
-  onDeleteTest?: Function,
-  onRunTest?: Function,
-  testSelection?: React.Node,
+  onDeleteTest?: () => void,
+  onRunTest?: () => void,
   testNameEditable?: React.Node,
   testsRunning?: Object,
+  onChange: () => void,
```
develohpanda: It's not immediately clear what this change handler is for - changing the test body, title, request, etc?

Similar to the other event handles (`onDeleteTest`, `onRunTest`), I suggest naming this explicitly like `onSetActiveRequest`, for example.

`onChange` on it's own is fine for a generic component like `Editable`, though. Contextual 😄

packages/insomnia-components/components/list-group/unit-test-item.js
```diff
@@ -1,41 +1,27 @@
...
+import UnitTestRequestSelector from './unit-test-request-selector';
 
 type Props = {|
   item: Object,
   children?: React.Node,
-  onDeleteTest?: Function,
-  onRunTest?: Function,
-  testSelection?: React.Node,
+  onDeleteTest?: () => void,
+  onRunTest?: () => void,
```
develohpanda, line 15: I'd like to double check but I don't think these callbacks should be optional, right?

packages/insomnia-components/components/list-group/unit-test-item.js
```diff
@@ -1,41 +1,27 @@
...
   testsRunning?: Object,
+  onChange: () => void,
+  selectedRequestId?: string,
+  selectableRequests: Array<{ name: string, request: { _id: string } }>,
 |};
 
 const StyledResultListItem: React.ComponentType<{}> = styled(ListGroupItem)`
   padding: 0 var(--padding-sm);
 
-  .form-control {
```
develohpanda: Yay, happy to see this move into what uses it specifically. 😄

packages/insomnia-components/components/list-group/unit-test-item.js
```diff
@@ -94,13 +89,7 @@ const UnitTestItem = ({
           <SvgIcon icon="play" />
         </Button>
       </div>
-      {collapse && (
-        <div
-          initial={{ height: collapse ? '100%' : '0px' }}
-          animate={{ height: collapse ? '100%' : '0px' }}>
```
develohpanda: Was the removal of animation intentional?

packages/insomnia-components/index.js
```diff
@@ -33,6 +34,7 @@ export const ListGroup = _ListGroup;
 export const ListGroupItem = _ListGroupItem;
 export const UnitTestItem = _UnitTestItem;
 export const UnitTestResultItem = _UnitTestResultItem;
+export const UnitTestRequestSelector = _UnitTestRequestSelector;
```
develohpanda: We can get away with not exporting this selector from insomnia-components since it's more of an internal component. Keen for your thoughts on this however - what stuff is exported from the library and what is not.

develohpanda (approved):
A couple of minor notes, good to merge after they are addressed. 👍

packages/insomnia-components/index.js
sonicyeti: 100% agreed, I didn't realize that this export serves the sole purpose of exposing it for usage in app. I thought it was necessary for bundling deps which the UT result item needs. I'll yank it!

packages/insomnia-components/components/list-group/unit-test-item.js
sonicyeti: ~~I pulled this because while we can slide-reveal the CM instance, when collapsed, it no longer exists in the DOM and looks quite janky when attempting to slide-hide and wanted to avoid having a toggle collapsed callback tied to the tween duration just for the sake of animating the exit of the CM instance.~~

Decided to keep and enable it. It just looks so much better sliding into view despite it immediately disappearing on collapse.

develohpanda: Fair enough, we can alter it depending on feedback. Either way seems fine

Issue #2051, opened 2020-04-24
[Bug] [Improvement] Improve the import of Postman variables


LamineMbn:
**Describe the bug**
In my new job they're using Postman and personally I prefer Insomnia. So I tried to import postman environnement variables to Insomnia and it didn't work for because the variables were either *nested* (`openapi.tokenId`) or with a *dash*

**To Reproduce**
1. Go to 'Preferences'
2. Click on 'Data'
3. Import data using this example :

```json
{
  "id": "65851a98-bb7b-4132-a0b4-5175edc63f01",
  "name": "env",
  "values": [
    {
      "key": "dev_or_prod_vpc",
      "value": "dev",
      "enabled": true
    },
    {
      "key": "color",
      "value": "abidjan",
      "enabled": true
    },
    {
      "key": "service-gateway-host",
      "value": "service-gateway-{{dev_or_prod_vpc}}.eu-west-1.elasticbeanstalk.com:8080",
      "enabled": true
    },
    {
      "key": "openAM.tokenId",
      "value": "AQIC5wM2LY4SfcztVCmR1P5o2QlyQQpJuapG3IZgT6E3AZc.*AAJTSQACMDIAAlNLABQtNDIwNjIxMDAxNTIyODQyMDAxNwACUzEAAjAx*",
      "enabled": true
    }
  ],
  "_postman_variable_scope": "environment",
  "_postman_exported_at": "2020-04-20T09:40:11.697Z",
  "_postman_exported_using": "Postman/7.22.1"
}

```
You'll get an error.

**Solution**
So I wrote some code to handle this for my own purpose and then I decided to add it to the Insomnia source code and it work like a charm.

So I can make a PR if this issue is relevant.

Thank you for your great work.

develohpanda:
Hi @LamineMbn! Thanks for using Insomnia! ♥️ 

This is a commonly reported issue: #1666 and #456, and the limitations for why it's a tricky one are outlined in the linked there too.

However! If you have something working, feel free to submit a PR and we'll take a look. We were talking about the _dash_ restriction just yesterday!

I'll leave this ticket open for the nesting failure, but this is a duplicate of #1666. 😊

develohpanda:
Closing this issue to keep discussing in #1666 :)

wongstein:
I'm going to close this issue in favor of other issues reported in the comment above.

PR #2601, opened 2020-09-11
Support hyphen `-` and other special characters in variable names

### LINKED ISSUE #456: [Usability] Dash in environment config key resolves value to `NaN`

### LINKED ISSUE #1666: Environmental Variable imports from postman evaluate to "NaN"

### LINKED ISSUE #1960: [Bug] accessing environmental variables that contain hyphens
**Describe the bug**
Environmental variables with a dash in them can't be referenced correctly (ie 'x-client-id' or something).

**To Reproduce**
1. go to env vars `control+e`
2. add a variable with a dash in it, ie `x-client-id`
3. click done
4. go to a request, add a header
5. press `control+space`, select one the new variable
6. hover mouse over the var, see NaN in live preview

**Expected behavior**
See the value you put in your environmental variable.  Additionally, the value is null on actual REST calls.

**Screenshots**
image

image

**Desktop:**
 - OS: Clear Linux
 - Installation Method: appimage (https://support.insomnia.rest/article/23-installation#linux)
 - App Version, see screen shot below (v7.1.1):
image

develohpanda:
Updating variable access logic to work with `{{ regular }}` (which is current behavior) as well as `{{ _['base-url'] }}`, which caters for dashes and special characters in the key.

Nunjucks templates support JS square bracket notation out of the box, but not as the first node. For example, `{{ ['base-url'] }}` is invalid (as it is in JS) while `{{ variable['base-url'] }}` is valid. For this reason, we have exposed the template context object (which is provided to Nunjucks), on the variable `_`, allowing you to access root keys that contain dashes or special characters, like `{{ _['base-url'] }}`.

The context object is exposed under `_` in _addition_ to the current implementation, so the old syntax (without `_`) will still work, when importing workspaces, or upgrading versions of Insomnia, for example.

**I selected `_` somewhat arbitrarily. If we would like to change that to something like `_ctx`, `_env`, `$`, or a different character, now is the best time to chime in. `_` feels least invasive.**

The autocomplete drop-down has been updated to always prefix an environment variable with `_`, to keep things consistent. It will also only use square bracket notation where necessary. If there are no special characters, regular dot notation will be used, now prefixed with `_`.

Old left, new right, both work
image

Old returns `NaN`, new returns the correct variable
image

> Hiding the `_.` prefix or the bracket syntax from the user in the nice rendered tag view is doable, and shouldn't be too tricky, but it would be somewhat misleading. Open to thoughts on that, but _I will not change in this PR, it will be changed in a future PR._

Additionally, NeDB has rules to prevent a key from starting with `$` or contain a `.`, so I have updated the regex key check for the environment editor. In the current implementation, errors are presented in console, and the changes do not persist (semi-silent failure), if you try to save something without following these rules which the existing regex did not check for.

image

### Screenshots

Environment
image

<details>
<summary>Autocomplete dropdown</summary>

image

</details>

<details>
<summary>Tags</summary>

image

Raw syntax
image

Not found errors
image

</details>

<details>
<summary>Variable editor modal</summary>

image

</details>

<details>
<summary>Regex Checks</summary>

image
image
image

</details>

<details>
<summary>Most importantly, this doesn't impact any existing tags</summary>

image
image
image

</details>

TODO:
- [x] Extract reusable regex and hard-coded strings into constants
- [x] Ensure folder environment editor handles hyphen the expected way
- [x] Double check all regex related changes
- [ ] Add unit tests where possible

To test:
- [x] Ensure export/import works
- [x] Check importers (postman)
- [x] Check openapi import
- [x] Ensure environment editor presents appropriate errors for invalid keys
- [x] Ensure folder environment editor presents appropriate errors for invalid keys

Fixes #1666, Fixes #456, Fixes #1960, Related to #2051, Fixes INS-118


packages/insomnia-app/app/templating/index.js
```diff
@@ -40,6 +40,7 @@ export function render(
 
   return new Promise(async (resolve, reject) => {
     const nj = await getNunjucks(renderMode);
+    nj.addFilter('getVarFromString', o => context[o]);
```
develohpanda: Not the best place to put this, ideally it should be near line 161, where the `debug` filter is added.

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -350,7 +350,7 @@ class TagEditor extends React.PureComponent<Props, State> {
           -- Select Variable --
         </option>
         {variables.map((v, i) => (
-          <option key={`${i}::${v.name}`} value={v.name}>
+          <option key={`${i}::${v.name}`} value={`'${v.name}' | getVarFromString`}>
```
develohpanda: Extract this value creation into a function and reuse

rhukster:
👍 as a developer who works with Twig on a daily basis, I was going to suggest a custom filter to handle the dashes and it seems exactly the approach you have already gone with. I think your solution is solid.  Retains template capabilities, UI mostly hides the filter so it does not impact usability.  Really looking forward to this.

packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.js
```diff
@@ -277,11 +277,13 @@ async function _updateElementText(render, mark, text, renderContext, isVariableU
       }
     } else {
       // Render if it's a variable
+      const matches = /^'([a-zA-Z_\-$][0-9a-zA-Z_\-$]*)' \| getVarFromString/.exec(cleanedStr);
+      const variableName = matches && matches[1];
```
gschier: Should this be `const variableName = matches ? matches[1] : cleanedStr` to support the old syntax (`{{ foo }}`) too?

packages/insomnia-app/app/templating/index.js
gschier: Might make sense to use `jsonpath` for this lookup so nested variables work: `{{ 'foo.bar' | ... }}`

gschier:
Another option I've thought of in the past is to expose the root environment under `$` (or something similar) just before rendering, then use square bracket syntax like `{{ $['foo']['bar-haha+blah  woo!']['baz'] }}` instead of dot notation. You wouldn't need an extra filter and it works out of the box with nested properties.

As @rhukster mentioned, though, the UI hides it all so it doesn't really matter what it looks like as long as the implementation is solid and it's not a breaking change 👍

rhukster:
> Another option I've thought of in the past is to expose the root environment under `$` (or something similar) just before rendering, then use square bracket syntax like `{{ $['foo']['bar-haha+blah woo!']['baz'] }}` instead of dot notation. You wouldn't need an extra filter and it works out of the box with nested properties.
> 

Yah in Twig the option is there to use `_context['foo']['bar-haha+blah woo!']['baz']` in the same way, I just didn't realize how similar Nunjucks was. The whole `_context` and square brackets is slightly uglier though, but if it's hidden, it really it comes down to what fits best for Insomnia.

develohpanda:
I will explore this option as well 👍

Either way, the template and the UI will need to be different, so if the feature set is identical between the two approaches I would favour the custom filter still, as it feels a bit clearer/easier to explain. Although context provided by a `$` also feels good. Will ponder 🤔

packages/insomnia-app/app/templating/index.js
develohpanda: Ya agreed 👏

packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.js
develohpanda: 🤔 Yes it absolutely should be

packages/insomnia-app/app/templating/index.js
develohpanda: Nested variable with hyphens and an array
image

Resolved 😄 
image

develohpanda:
Added support for nesting:

image

image

develohpanda:
Further testing - looks like nesting can kind of work without the filter altogether, [square bracket syntax](https://mozilla.github.io/nunjucks/templating.html#variables) being supported OOTB

image

Doesn't quite work if the first item has illegal characters though, so it would require exposing the entire resolved context under a `_ctx` or `$` variable, and then (likely) some regex to convert from the bracket syntax to display nicely in the tag.

packages/insomnia-app/app/templating/index.js
```diff
@@ -16,6 +16,7 @@ export class RenderError extends Error {
 export const RENDER_ALL = 'all';
 export const RENDER_VARS = 'variables';
 export const RENDER_TAGS = 'tags';
+export const NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME = '_';
```
develohpanda, line 19: In theory, now the only regex rule that needs to exist against an environment editor is that you cannot have a variable be solely `_`, it being a reserved key. All other characters would be allowed... 🤯

develohpanda:
All of these syntax (syntaii? :trollface:) now work. 1 & 2 are new, 3, 4 and 5 are existing, ensuring backwards compatibility

Note that bracket notation is only used where it's necessary, otherwise dot notation is used. If you have no special characters, dot notation will be used.

image

packages/insomnia-app/app/common/render.js
```diff
@@ -276,28 +277,30 @@ export async function getRenderContext(
   function getKeySource(subObject, inKey, inSource) {
     // Add key to map if it's not root
     if (inKey) {
-      keySource[inKey] = inSource;
+      keySource[templatingUtils.normalizeToDotAndBracketNotation(inKey)] = inSource;
```
develohpanda, line 280: The changes in this file relate to building the key source lookup, so we can know if data is coming from the base environment, a sub environment, or a folder. The notation needs to be the same as what the template tag itself contains, hence uses the forceBracket and normalize functions used by the key indexing logic.

packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.js
```diff
@@ -280,7 +280,10 @@ async function _updateElementText(render, mark, text, renderContext, isVariableU
       title = await render(str);
       const context = await renderContext();
       const con = context.context.getKeysContext();
-      title = '{' + con.keyContext[cleanedStr] + '}: ' + title;
+      const contextForKey = con.keyContext[cleanedStr];
+
+      // Only prefix the title with context, if context is found
+      title = contextForKey ? `{${contextForKey}}: ${title}` : title;
```
develohpanda, line 286: `contextForKey` is root, or the sub environment, or the folder name, etc

packages/insomnia-app/app/ui/components/editors/__tests__/environment-editor.test.js
```diff
@@ -0,0 +1,27 @@
...
+  const name = NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME;
+
+  it(`${name} as a key should be invalid`, () => {
+    expect(ensureKeyIsValid(name)).toBe(`"${name}" is a reserved key`);
+  });
+
+  it.each(['a', 'ab', 'a$', 'a$b', 'a-b', `a${name}b`, `${name}ab`])('%s should be valid', key => {
+    expect(ensureKeyIsValid(key)).toBe(null);
+  });
+});
```
develohpanda, line 27: Tests to ensure a key cannot be `_`, or contain `.`, or start with `$`.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -430,7 +431,7 @@ class App extends PureComponent {
 
   async _handleGetRenderContext() {
     const context = await this._fetchRenderContext();
-    const keys = getKeys(context);
+    const keys = getKeys(context, NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME);
```
develohpanda, line 434: `getKeys` is recursive, and is defined in `templating/utils.js`. The second argument reflects the key prefix, as the algorithm traverses all nodes. We now start the traversal with `_`, so all keys are prefixed by `_`.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -53,18 +69,19 @@ class EnvironmentEditor extends React.PureComponent<Props, State> {
...
     if (value && value.object) {
       for (const key of Object.keys(value.object)) {
-        if (!key.match(/^[a-zA-Z_$][0-9a-zA-Z_$]*$/)) {
-          warning = `"${key}" must only contain letters, numbers, and underscores`;
+        error = ensureKeyIsValid(key);
+        if (error) {
           break;
         }
       }
     }
```
develohpanda, line 80: There is a bug here in that both previously and now, only the root keys are checked. A recursive algorithm needs to be created to validate nested keys appropriately.

develohpanda (commented):
Some comments for reviewers.

DMarby (approved):

sonicyeti (approved):
Ultra comprehensive, great work!

koldoon:
IMO there is nothing wrong to simply remove `_.` prefix. In most cases it looks dirty and doesn't really help, so it should be removed everywhere when possible.

PR #2617, opened 2020-09-15
Auto focus filter and improve click targets in design sidebar

sonicyeti:
This PR covers the default cursor focusing on the filter's input field upon click. See demo animation below:

FIXES INS-66

### DIFF
--- a/packages/insomnia-components/components/sidebar/sidebar-filter.js
+++ b/packages/insomnia-components/components/sidebar/sidebar-filter.js
@@ -42,6 +42,8 @@ const SidebarFilter = ({ filter, onChange }: Props) => {
   useLayoutEffect(() => {
     if (filterField.current && !filter) {
       filterField.current.value = '';
+    } else if (filterField.current) {
+      filterField.current.focus();
     }
   }, [filter, filterField]);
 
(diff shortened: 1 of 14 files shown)


DMarby (approved):

packages/insomnia-components/components/sidebar/sidebar-filter.js
```diff
@@ -42,6 +42,8 @@ const SidebarFilter = ({ filter, onChange }: Props) => {
   useLayoutEffect(() => {
     if (filterField.current && !filter) {
       filterField.current.value = '';
+    } else if (filterField.current) {
```
develohpanda, line 45: Don't think this needs to be an `else if`, is that intentional?

It needs to be in focus every time it is newly opened IMO. If achieving that behavior requires a standalone react effect, that's fine as well.

packages/insomnia-components/components/sidebar/sidebar-info.js
```diff
@@ -7,29 +7,31 @@ import SidebarTextItem from './sidebar-text-item';
 type Props = {
   info: Object,
   childrenVisible: boolean,
+  onClick: (section: string, ...args: any) => void,
```
develohpanda:
```suggestion
  onClick: (section: string, ...args: Array<string>) => void,
```

🔥`any`:fire:

develohpanda (approved):
Looks good, however this PR does change more than is necessary for resolving INS-66.

There is a separate issue fixed ad-hoc - click targets for a sidebar section and item (a part of INS-31), which can be be standalone PRs, or at the very least called out in the PR description. Please take care for future. 🤗 

Couple of comments on the code but none blocking

PR #2622, opened 2020-09-16
Ignore x- keys for methods in the design sidebar paths section

develohpanda:
What it says on the tin

The same logic is used to ignore `x-*` keys while importing an OpenAPI3 specification.

https://github.com/Kong/insomnia/blob/51d3a5de32043dcf429f66afab955e66b6ac7e76/packages/insomnia-importers/src/importers/openapi3.js#L191-L193

x- keys can be used, for example, to provide a unique name to a path for Kong

```yaml
...
paths:
  "/api/orders":
    x-kong-name: All Orders
    get:
      operationId: getAllOrders
```

Previously, `x-kong-name` would appear in white next to `get` in the sidebar

image

### DIFF
--- a/packages/insomnia-components/components/sidebar/sidebar-paths.js
+++ b/packages/insomnia-components/components/sidebar/sidebar-paths.js
@@ -15,6 +15,8 @@ const StyledMethods: React.ComponentType<{}> = styled.span`
   padding-left: var(--padding-lg);
 `;
 
+const isNotXDashKey = key => key.indexOf('x-') !== 0;
+
 // Implemented as a class component because of a caveat with render props
 // https://reactjs.org/docs/render-props.html#be-careful-when-using-render-props-with-reactpurecomponent
 export default class SidebarPaths extends React.Component<Props> {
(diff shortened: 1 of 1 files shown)


DMarby (approved):

packages/insomnia-components/components/sidebar/sidebar-paths.js
```diff
@@ -47,14 +49,16 @@ export default class SidebarPaths extends React.Component<Props> {
...
+                {Object.keys((routeBody: any))
+                  .filter(isNotXDashKey)
+                  .map(method => (
+                    <span
+                      key={method}
+                      className={`method-${method}`}
+                      onClick={() => onClick('paths', route, method)}>
+                      {method}
+                    </span>
+                  ))}
```
develohpanda, line 61: Review with whitespace hidden
image

ronyfhebrian:
Is there anyway that I can view x-byte-read-ops in headers?

develohpanda:
Hi @ronyfhebrian, if I understand correctly I think your question is unrelated to this PR? Do you have an OpenAPI spec defined like the following?

```yaml
...
paths:
  "/api/orders":
    x-byte-read-ops: Some name
```

PR #2626, opened 2020-09-16
Add shortcuts to foldAll and unfoldAll JSON nodes

### LINKED ISSUE #2252: Feature Request: ability to expand or collapse all JSON nodes
When you want to look at the last object in a JSON response, for example, it can be tedious to collapse 50+ objects by hand to get to the last one. Having the ability to collapse the entire response and navigate to where you want to go by expanding specific nodes one at a time can be much faster than collapsing a dozen nodes to get there, or scrolling and reading.

At the least, I would like to have shortcuts added for "expand/collapse all" in the response. Ideally, I'd also like to be able to right click on any node and have an "expand/collapse children" option available as well.

Alternatives considered: I could scroll to the bottom, but shortcuts or options for collapsing would be much faster, and scrolling to the bottom is only feasible when I happen to be looking at the last object in the response. Auto-collapsing is also useful if you want to look at a particular object in the middle of the response or something.

This is not my screenshot, but may be helpful in case what I was referring to was unclear:
Screenshot

jgiovaresco:
Use `Shift Cmd -` or `Shift Ctrl -` to fold all nodes
Use `Shift Cmd =` or `Shift Ctrl =` to unfold all nodes

I've chosen my usual shortcuts I'm using with JetBrains tools. 

I didn't found if there were tests covering shortcuts declaration 🤔 

Closes #2252

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.js
@@ -66,6 +66,9 @@ const BASE_CODEMIRROR_OPTIONS = {
     // search box stays open after pressing Enter
     [isMac() ? 'Cmd-F' : 'Ctrl-F']: 'findPersistent',
 
+    [isMac() ? 'Shift-Cmd--' : 'Shift-Ctrl--']: 'foldAll',
+    [isMac() ? 'Shift-Cmd-=' : 'Shift-Ctrl-=']: 'unfoldAll',
+
     'Shift-Tab': 'indentLess',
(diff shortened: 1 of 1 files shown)


develohpanda (approved):
Looks good, we can change the keymap if there is any feedback later on. Thanks for working on this! 🎉 

> I didn't found if there were tests covering shortcuts declaration

Testing coverage of components (and logic such as this inside them) is something we are working towards. No coverage for these areas at the moment. 😞

NSExceptional:
Amazing! Thank you!

PR #2632, opened 2020-09-17
Sidebar method styling & click regions

sonicyeti:
This PR covers the method badge design integration into the application along with the fixing of sidebar item click region inconsistencies.

- Theme support
- Subtle hover animation and alpha shift
- Fixed legacy alignment & spacing issues
- Address less than intuitive click target inconsistencies with the rest of the sidebar

Static screenshot and hover animation below:

Screen Shot 2020-09-17 at 10 21 21 AM

FIXES INS-31
FIXES INS-76
FIXES INS-34


packages/insomnia-components/components/sidebar/sidebar-panel.js
```diff
@@ -10,33 +10,6 @@ type Props = {
 
 const StyledPanel: React.ComponentType<{}> = styled(motion.div)`
   height: 0;
-  .method {
```
develohpanda, line 13: 👍

packages/insomnia-components/components/sidebar/sidebar-method-badge.js
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+    &.put {
+      background-color: rgba(var(--color-warning-rgb), 0.8);
+      &:hover {
+        background-color: var(--color-warning);
+      }
+    }
+  }
+`;
```
develohpanda: Looking at the two usages below, I propose an update to this component.

```js
// sidebar-badge.js

type Props = {	
	onClick: () => void,
	label?: string,
	method?: 'get' | 'post' | ... | 'put',
}

export const SidebarBadge = (props: Props) => { ... };
```

Where:
- if `method` _is not_ provided, some default color is applied
- if `label` _is_ provided, it is overwritten, otherwise is the `method`
- if both `method` and `label` are not provided, returns `null` because there is not enough information to render (to avoid this case, this can be split into two components as well, for example)

Consumed as:
```js
	<SidebarBadge onClick={...} method="get" />
	<SidebarBadge onClick={...} label="cat" />
```

Doesn't have to be _exactly_ this - this can also be split into two components, but notice how this approach abstracts the need for a user to know the inner styling of `SidebarBadge` in order to use it.

After redoing this component, you can also create stories in Storybook.

develohpanda (changes requested):
Visually this looks really good 🙌 definitely a step forwards for usability. One important note about the implementation of the new component - `SidebarMethodBadge` - and the way it is consumed. I encourage you to design towards exporting _components_ instead of _stylesheets as components_. 🤗

DMarby (approved):

packages/insomnia-components/components/sidebar/sidebar-badge.js
```diff
@@ -1,7 +1,22 @@
...
-export default styled.span`
+type Props = {
+  label?: string,
+  onClick?: () => void,
+  method?:
+    | 'get'
+    | 'post'
+    | 'delete'
+    | 'parameters'
+    | 'parameters'
```
DMarby: Should "parameters" be duplicated here?

DMarby (changes requested):

DMarby (approved):

packages/insomnia-components/components/sidebar/sidebar-method-badge.js
sonicyeti: Got it, coming along although I opted to make the onClick optional as well in the event we need a visual only badge for something. Pinged you on a flow related issue.

packages/insomnia-components/components/sidebar/sidebar-requests.js
```diff
@@ -86,10 +88,7 @@ export default class SidebarRequests extends React.Component<Props> {
                               'examples',
                               requestExample,
                             )
-                          }
-                          key={requestExample}>
-                          <span className={'post'}>{requestExample}</span>
-                        </SidebarMethodBadge>
+                          }></SidebarBadge>
```
develohpanda: Can be self closing

```suggestion
                          } />
```

packages/insomnia-components/components/sidebar/sidebar-requests.js
```diff
@@ -75,7 +75,9 @@ export default class SidebarRequests extends React.Component<Props> {
                   {content[requestFormat].examples && (
                     <SidebarItem>
                       {Object.keys(content[requestFormat].examples).map(requestExample => (
-                        <SidebarMethodBadge
+                        <SidebarBadge
+                          key={requestExample}
```
develohpanda, line 79: 👍

packages/insomnia-components/components/sidebar/sidebar-badge.js
```diff
@@ -63,3 +70,13 @@ export default styled.span`
     }
   }
 `;
+
+const SidebarBadge = ({ onClick = () => {}, method = 'post', label = method }: Props) => {
```
develohpanda, line 77: The `onClick` prop isn't used anywhere - it needs to be sent to the `<StyledBadge>` right?

packages/insomnia-components/components/sidebar/sidebar-badge.js
```diff
@@ -63,3 +70,13 @@ export default styled.span`
...
+  return (
+    <StyledBadge>
+      <span className={method}>{label}</span>
+    </StyledBadge>
+  );
+};
+
+export default SidebarBadge;
```
develohpanda, line 85: Keen to see some Storybook stories for this new component 😄

develohpanda (commented):
Nicely done 👏 I've made a couple of comments, but you're on the right path!

DMarby (approved):

develohpanda (approved):
Nice! 💯

PR #2634, opened 2020-09-17
Section visibility toggle

sonicyeti:
This PR fixes the dropdown list item row not being clickable when setting section visibility preferences. Also addressed is the misalignment of the label with the checkbox itself.

Demo:

FIXES INS-67


develohpanda (approved):
👍

DMarby (approved):

PR #2637, opened 2020-09-18
Upgrade apiconnect-wsdl to remove yargs-parser dependency

develohpanda:
The recommendation for `yargs-parser` is to upgrade to versions 13.1.2, 15.0.1, 18.1.1 or later.

There are two deps, whose nested dependencies used versions below one of those specified - `apiconnect-wsdl` and `prompt-run`.

`apiconnect-wsdl` is used for importing wsdl documents, and a newer patch release was available. I upgraded, and the latest version removes yargs-parser as a dependency altogether. This is the dependency that caused a dependabot alert.

`prompt-run` has the following tree to `yargs-parser`, and while a new major version of `meow` is available, `prompt-run` needs to consume it and publish a new version, in order to remove the dependency on `yargs-parser@10`. `prompt-run` is only used when starting the app during development, so I am inclined to leave it as-is.

image


packages/insomnia-importers/src/__tests__/fixtures/wsdl/addition-output.json
```diff
@@ -35,7 +35,7 @@
...
+        "text": "\n<soapenv:Envelope xmlns:soapenv=\"http://schemas.xmlsoap.org/soap/envelope/\">\n <soapenv:Header>\n  <!-- The Security element should be removed if WS-Security is not enabled on the SOAP target-url -->\n  <wsse:Security xmlns:wsse=\"http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd\" xmlns:wsu=\"http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd\">\n   <wsse:UsernameToken>\n    <wsse:Username>string</wsse:Username>\n    <wsse:Password>string</wsse:Password>\n    <wsse:Nonce EncodingType=\"string\">string</wsse:Nonce>\n    <wsu:Created>string</wsu:Created>\n   </wsse:UsernameToken>\n   <wsu:Timestamp wsu:Id=\"string\">\n    <wsu:Created>string</wsu:Created>\n    <wsu:Expires>string</wsu:Expires>\n   </wsu:Timestamp>\n  </wsse:Security>\n </soapenv:Header>\n <soapenv:Body>\n  <tns:Add xmlns:tns=\"http://tempuri.org/\"><!-- mandatory -->\n   <tns:intA><!-- mandatory -->3</tns:intA>\n   <tns:intB><!-- mandatory -->3</tns:intB>\n  </tns:Add>\n </soapenv:Body>\n</soapenv:Envelope>"
```
develohpanda, line 38: Note, the only change to this single line (in each file), is the following comment which is a part of the output after upgrading apiconnect-wsdl

```xml
  <!-- The Security element should be removed if WS-Security is not enabled on the SOAP target-url -->
```

develohpanda: There were two fixture tests, which still pass after adding that comment into the output

image

DMarby: 👍

DMarby (approved):
I think it sounds reasonable to leave the import caused by `prompt-run` as-is as well.

PR #2642, opened 2020-09-19
Introduce a checkbox in Basic Auth form to choose ISO-8859-1 encoding for credentials.

### LINKED ISSUE #1558: Basic authentication doesn't handle special French characters in password
- Insomnia Version: 6.5.4

## Details

While investigating an authentication issue with Jira API (see https://bitbucket.org/farmas/atlassian.net-sdk/issues/460/cant-authenticate-when-password-contains), I tested the same requests with Insomnia.

Sadly Insomnia had the same issue and I couldn't authenticate as soon as my password contains a é.

It's probably related to the encoding too.

jgiovaresco:
Introduce a checkbox in Basic Auth form to choose ISO-8859-1 encoding for credentials.

Screenshot 2020-09-19 at 19 57 37
Screenshot 2020-09-19 at 19 57 46

I had to update network's "Handle Authorization header" to use `getAuthHeader` instead of libcurl options for Basic Http
because it seems that lib-curl does not let us choose the encoding for Basic Auth credentials

Closes #1558


nijikokun:
~I'm actually wondering if we should [follow suit and simply apply UTF-8 encoding](https://www.fxsitecompat.dev/en-CA/docs/2018/basic-auth-credentials-are-now-encoded-in-utf-8-instead-of-iso-8859-1/) by default, and have a checkbox to fallback to the previous encoding.~

Actually, after looking through the PR, it appears that the default is `utf8` and this is a fallback to `latin1`. Interesting that `utf8` isn't properly encoding french accents as it should (see link above for Firefox reasoning). I suspect the server expects the encoding format from `latin1` like the listed exception jboss.

Looks good from a product perspective, with one small suggestion that we put `Use` in front of the encoding.

packages/insomnia-app/app/models/request.js
```diff
@@ -127,6 +127,7 @@ export function newAuth(type: string, oldAuth: RequestAuthentication = {}): Requ
     case AUTH_NTLM:
       return {
         type,
+        useISO88591: oldAuth.useISO88591 || false,
```
develohpanda: It looks like the rest of the code now treats AUTH_BASIC differently, and because this new option is only applicable for AUTH_BASIC, I would favour keeping the DIGEST and NTLM cases as previous, and create a standalone switch case for BASIC that adds this property.

packages/insomnia-app/app/ui/components/editors/auth/basic-auth.js
```diff
@@ -126,6 +135,32 @@ class BasicAuth extends React.PureComponent<Props> {
                 </div>
               </td>
             </tr>
+            <tr>
+              <td className="pad-right no-wrap valign-middle">
+                <label htmlFor="use-iso-8859-1" className="label--small no-pad">
+                  ISO 8859-1
```
develohpanda:
```suggestion
                  Use ISO 8859-1
```

develohpanda (approved):
Looks good to me, thanks for doing this!

Small change I think is worth addressing, otherwise good to go.

packages/insomnia-app/app/models/request.js
jgiovaresco: Indeed I should have done that. Nice catch :)

jgiovaresco:
@develohpanda I've rebased the PR and amended the 1st commit (#fef8cd3) with your suggestions. I let you take a look ;)

develohpanda (approved):
Nice! 💯 

I'm going to merge this in, and will explore adding an end-to-end test for Core in the `packages/insomnia-smoke-test` project. Feel free to give that a go separately, but it's a little new so there isn't much documentation or instructions as yet.

PR #2644, opened 2020-09-21
CodeMirror debounce wait for smoke tests

develohpanda:
Typing into a code-editor field is a little flaky, and I realized that the code-editor has a configurable debounce that is applied, which by default is 100ms.

I have added an artificial 150ms pause, in order to allow the debounce timeout to elapse before proceeding to the next step. Hopefully this makes typing into a code-editor a little less flaky, but I will monitor it.

### DIFF
--- a/packages/insomnia-smoke-test/modules/debug.js
+++ b/packages/insomnia-smoke-test/modules/debug.js
@@ -29,14 +29,17 @@ const createNewRequest = async (app, name = undefined) => {
 
 const typeUrl = async (app, url) => {
   const urlEditor = await app.client.$('.urlbar .editor .input');
-  await typeCodeMirror(urlEditor, url);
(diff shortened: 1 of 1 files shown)


DMarby (approved):
Perhaps it would be good to heighten the "wait for app window" timeout a bit from 5 seconds as well? 
I've noticed that has failed a few times

develohpanda:
> Perhaps it would be good to heighten the "wait for app window" timeout a bit from 5 seconds as well?
> I've noticed that has failed a few times

I'll keep an eye out (pls link when you see it again too) and then figure out what might be causing it. I've noticed that a test doesn't _exit_ in certain conditions, instead the GitHub Action itself stops after a timeout. Trying to narrow that down, as it doesn't happen very frequently

PR #2645, opened 2020-09-21
Update pdfjs-dist

### LINKED ISSUE #2623: PDF Preview broken in Insomnia 2020.4.0
**Describe the bug**
The response preview is empty and a error is thrown in the console.

**To Reproduce**
Send a GET request with a response of content type `application/pdf`.

**Screenshots**
insomnia-broken-pdf-preview

**Desktop:**
 - OS: tested on Linux and macOS
 - App Version: Insomnia Core 2020.4.0

develohpanda:
This PR updates pdfjs-dist and fixes our usage to bring it in line with the interfaces in the latest version.

This PR also adds an E2E test, which currently loads [this](https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf) PDF. Adding in a mock PDF response through Prism wasn't working/isn't supported yet, and we may need to spin up our own local server to present mock data. We can remove this test if it becomes flaky.

image

Closes #2623

### DIFF
--- a/packages/insomnia-app/package.json
+++ b/packages/insomnia-app/package.json
@@ -152,7 +152,7 @@
     "oauth-1.0a": "^2.2.2",
     "openapi-2-kong": "^2.2.22",
     "papaparse": "^5.2.0",
-    "pdfjs-dist": "^2.2.228",
+    "pdfjs-dist": "^2.5.207",
     "prop-types": "^15.6.1",
     "react": "^16.8.6",
     "react-animated-tree": "^1.0.10",
--- a/packages/insomnia-smoke-test/core/app.test.js
+++ b/packages/insomnia-smoke-test/core/app.test.js
@@ -79,7 +79,7 @@ describe('Application launch', function() {
     await expect(csvViewer.getText()).resolves.toBe('a b c\n1 2 3');
   });
 
-  fit('sends PDF request and shows rich response', async () => {
+  it('sends PDF request and shows rich response', async () => {
     const pdfUrl = 'https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf';
     await debug.workspaceDropdownExists(app);
     await debug.createNewRequest(app);
--- a/packages/insomnia-smoke-test/core/app.test.js
+++ b/packages/insomnia-smoke-test/core/app.test.js
@@ -78,4 +78,17 @@ describe('Application launch', function() {
     const csvViewer = await debug.getCsvViewer(app);
     await expect(csvViewer.getText()).resolves.toBe('a b c\n1 2 3');
   });
+
+  fit('sends PDF request and shows rich response', async () => {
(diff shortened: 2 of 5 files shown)


packages/insomnia-smoke-test/core/app.test.js
```diff
@@ -78,4 +78,17 @@ describe('Application launch', function() {
     const csvViewer = await debug.getCsvViewer(app);
     await expect(csvViewer.getText()).resolves.toBe('a b c\n1 2 3');
   });
+
+  it('sends PDF request and shows rich response', async () => {
+    const pdfUrl = 'https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf';
```
DMarby, line 85: Would be good to leave a comment mentioning why this external URL is used over Prism, perhaps referencing https://github.com/stoplightio/prism/issues/1248#issuecomment-646056440

packages/insomnia-smoke-test/core/app.test.js
```diff
@@ -78,4 +78,17 @@ describe('Application launch', function() {
...
+    await debug.workspaceDropdownExists(app);
+    await debug.createNewRequest(app);
+    await debug.typeUrl(app, pdfUrl);
+    await debug.clickSendRequest(app);
+
+    await debug.expect200(app);
+    const pdfCanvas = await debug.getPdfCanvas(app);
+    await expect(pdfCanvas.getAttribute('width')).resolves.toBe('1018');
```
DMarby: Is it possible to get the contents of the canvas, to ensure that it contains "Dummy PDF File"?

packages/insomnia-app/package.json
```diff
@@ -152,7 +152,7 @@
     "oauth-1.0a": "^2.2.2",
     "openapi-2-kong": "^2.2.22",
     "papaparse": "^5.2.0",
-    "pdfjs-dist": "^2.2.228",
+    "pdfjs-dist": "^2.5.207",
```
DMarby, line 155: As far as I can tell this downgrades pdfjs rather then upgrades it, is there any particular reason for this?

DMarby (changes requested):

packages/insomnia-app/package.json
DMarby: Never mind, I can't read, completely missed the .5 major version

packages/insomnia-smoke-test/core/app.test.js
develohpanda: The general approach I've seen is to take and compare screenshots, or use an image processing library to extract text. I'll add that as a note within the test for now and add a card to come back and remove a) the dependency on an external resource, and b) try and extract some text

DMarby: Sounds fair enough 👍

DMarby (approved):
🎉Awesome

packages/insomnia-smoke-test/core/app.test.js
develohpanda: INS-141 & INS-142

PR #2652, opened 2020-09-22
Add PKCE support

### LINKED ISSUE #1653: Feature Request: Support Proof Key for Code Exchange (PKCE)
A lot of APIs are now supporting PKCE authentication flow for end-users (for mobile, native and SPA applications).

To explore and test those APIs, obtaining a token with this flow is complex to do manually. 

It would be a great addition to Insomnia if a user/tester could authenticate itself using PKCE flow.

### LINKED ISSUE #2154: Support "code_challenge" with OAuth2 Auth Code
**Support of "code_challenge" with OAuth2**
Hi, does Insomia support the Pkce "Code Challenge" Machanism with OAuth2 and the Auth Code Flow?

An error is returned, when Pkce is activated, without Pkce insomia can get the token.

Greetings
sdc4

DASPRiD:
This is a small change to support PKCE with authentication code authorization.

Closes #1653
Closes #2154


develohpanda:
Thanks for the PR @DASPRiD! Could you please take a look at the breaking tests? Please mark as ready to review again once fixed up. Thanks!

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
```diff
@@ -235,6 +239,36 @@ class OAuth2Auth extends React.PureComponent<Props, State> {
...
+          </label>
+        </td>
+        <td className="wide">
+          <div className="form-control form-control--underlined no-margin">
+            <Button
+              className="btn btn--super-duper-compact"
+              id="use-pkce"
+              onClick={onChange}
+              value={authentication.usePkce}
+              title={authentication.usePkce ? 'Disable PKCE' : 'Enable PKCE'}>
```
develohpanda, line 259: Strange, it seems the `Button` component doesn't use the `title` prop for anything.

It might have been removed accidentally because there are a lot of instances that provide the `title` prop. No need to change this line, we should fix button to consume it. 😄

This is more a note, no need to action it in this PR.

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.js
```diff
@@ -182,3 +207,11 @@ async function _getToken(
 
   return results;
 }
+
+function _base64UrlEncode(string: string): string {
+  return string
+    .toString('base64')
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_')
+    .replace(/=/g, '');
```
develohpanda, line 216: Could you please explain why these replacements need to happen - is it something specific with OAuth 2, or something else that I'm missing?

If there are any docs, it would be great to add as a comment in the code, for anyone looking at it later on.

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.js
```diff
@@ -82,6 +100,11 @@ async function _authorize(
   audience && params.push({ name: c.P_AUDIENCE, value: audience });
   resource && params.push({ name: c.P_RESOURCE, value: resource });
 
+  if (codeChallenge) {
+    params.push({ name: c.P_CODE_CHALLENGE, value: codeChallenge });
+    params.push({ name: c.P_CODE_CHALLENGE_METHOD, value: 'S256' });
```
develohpanda, line 105: Does `S256` need to be a hard-coded constant in `o-auth-2/constants.js`, or be defined closer to the hash type (`crypto.createHash('sha256')`) to indicate that the two are related?

(I assume they are related)

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.js
```diff
@@ -182,3 +207,11 @@ async function _getToken(
 
   return results;
 }
+
+function _base64UrlEncode(string: string): string {
+  return string
```
develohpanda: We absolutely should not be using reserved keywords as variable names, very strange that ES Lint didn't complain 🤔 Even if it works, it can be a little confusing.

```suggestion
function _base64UrlEncode(str: string): string {
  return str
```

develohpanda (changes requested):
Apologies for the delayed review! Thank you for working through this feature, I'm looking forward to get it in. 😄

A couple of questions and one minor change, but otherwise LGTM. 👏

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.js
DASPRiD: Base64-URL encoding is defined by the PKCE RFC:
https://tools.ietf.org/html/rfc7636#section-4.2

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.js
DASPRiD: Not really. There's only S256 and plain, and plain **must not** be used when the client does support sha256 hashing, according to the RFC. See also:

https://tools.ietf.org/html/rfc7636#section-4.2

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.js
develohpanda: 👍 thanks for that

```suggestion
    .replace(/\+/g, '-')
    .replace(/\//g, '_')
    .replace(/=/g, '');
    
    // The characters + / = are reserved for PKCE as per the RFC,
    // so we replace them with unreserved characters
    // Docs: https://tools.ietf.org/html/rfc7636#section-4.2
```

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.js
develohpanda: Noted, thanks!

develohpanda (approved):
LGTM

DASPRiD:
> LGTM

Your suggested change broke prettier D:

DASPRiD:
Fixed the prettier report. For some reason the MacOS build is failing now, but that seems to be build server related… *shrugs*

develohpanda:
> Fixed the prettier report. For some reason the MacOS build is failing now, but that seems to be build server related… *shrugs*

No probs, I'll have a look at what's breaking there, should be fine though.

DMarby (approved):

et1975:
Can anyone confirm if this impl works with Azure AD?

EDIT: Confirmed myself, make sure to specify `Credentials: In Request Body` though!

PR #2658, opened 2020-09-24
Add babel to smoke tests

develohpanda:
Convert to ES6 imports and exports

### DIFF
--- a/packages/insomnia-smoke-test/.babelrc
+++ b/packages/insomnia-smoke-test/.babelrc
@@ -4,7 +4,7 @@
       "@babel/preset-env",
       {
         "targets": {
-          "node": "10"
+          "node": "12"
         }
       }
     ]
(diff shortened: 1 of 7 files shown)


packages/insomnia-smoke-test/.babelrc
```diff
@@ -0,0 +1,15 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "10"
```
DMarby: Might make more sense to set this to 12, to match the babelrc in `insomnia-app`, and since we're running Node 12?

DMarby (changes requested):

packages/insomnia-smoke-test/.babelrc
develohpanda: Yes 😅 following that, I should change the other projects that have `.babelrc` set to node 10 as well

DMarby (approved):

PR #2659, opened 2020-09-24
Refactor and reuse app launch logic, launch app with a fresh data directory, prepare for tests in release pipeline

develohpanda:
What it says on the tin. See the `app.test.js` files for usage.

### DIFF
--- a/packages/insomnia-smoke-test/modules/application.js
+++ b/packages/insomnia-smoke-test/modules/application.js
@@ -1,0 +1,75 @@
+import { Application } from 'spectron';
+import path from 'path';
+import os from 'os';
+import electronPath from '../../insomnia-app/node_modules/electron';
+import { APP_ID_INSOMNIA, APP_ID_DESIGNER } from '../../insomnia-app/config';
+
+const getAppPlatform = () => process.platform;
+const isMac = () => getAppPlatform() === 'darwin';
+const isLinux = () => getAppPlatform() === 'linux';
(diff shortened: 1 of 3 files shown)


DMarby (approved):

packages/insomnia-smoke-test/modules/application.js
```diff
@@ -0,0 +1,75 @@
...
+const isWindows = () => getAppPlatform() === 'win32';
+
+const spectronConfig = appId => {
+  let packagePathSuffix = '';
+  if (isWindows()) {
+    packagePathSuffix = 'win-unpacked/Insomnia.exe';
+  } else if (isMac()) {
+    packagePathSuffix = 'mac/Insomnia.app/Contents/MacOS/Insomnia';
+  } else if (isLinux()) {
+    packagePathSuffix = ''; // TODO: find out what this is
```
DMarby, line 19: Probably good to fix this before merging

develohpanda: Because we don't run package tests right now, and my Linux dev env doesn't actually work right now, I'm going to do it in a separate PR.

If we try to run package tests before fixing this, the ubuntu test will just fail, so I don't think it's too harmful, but to keep the other PR unblocked. Does that sound good?

DMarby: Sounds reasonable 👍

PR #2662, opened 2020-09-24
Expose network context to requestHook & responseHook scopes

### LINKED ISSUE #2012: [Improvement] Plugin API to send new request in response hook
An external API does not always respond successfully. Server errors for example 50x or `429 Too Many Requests` may occur. Manually pressing the send button while waiting for a successful response can be tedious.

Be able to set the number of retries and the expected/unexpected status codes of the response. As well as the interval between requests.

jgiovaresco:
The issue #2012 was talking about exposing the `network` context to the `responseHook` scope to be able to send new request from there.

I took the opportunity to expose the `network` context to the `requestHook` also to be able to send request in request hook. Personally I am interested by this to create a plugin to login automatically to a private api. If you prefer I can move the commit in another PR ;)

Closes #2012

### DIFF
--- a/packages/insomnia-app/app/network/network.js
+++ b/packages/insomnia-app/app/network/network.js
@@ -997,11 +997,11 @@ async function _applyRequestPluginHooks(
 
 async function _applyResponsePluginHooks(
   response: ResponsePatch,
-  request: RenderedRequest,
-  renderContext: Object,
+  renderedRequest: RenderedRequest,
+  renderedContext: Object,
 ): Promise<ResponsePatch> {
   const newResponse = clone(response);
-  const newRequest = clone(request);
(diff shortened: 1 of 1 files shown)


jgiovaresco:
Documentation should be updated (https://support.insomnia.rest/article/26-plugins#context.response / https://support.insomnia.rest/article/26-plugins#context.request) but I don't know how to update it 😅

develohpanda (approved):
Looks good to me 👍 

What part of the documentation needs to be updated? Docs are hosted on HelpScout for various reasons, but if you post a diff/changes into a comment here I can go ahead and update HelpScout with it. (I don't think there are any changes required but I might be missing something)

PR #2668, opened 2020-09-25
Bump httpsnippet to fix Copy as Curl not respecting header case

### LINKED ISSUE #2615: Copy as Curl not respecting header case
When doing "Copy as Curl" on a request - the header is copied in all lower case. 
E.g. I have a query with a header called "X-Application" - when doing a "Copy as Curl" it is copied as "x-application"

 - OS: macOS 10.15.6
 - Installation Method: dmg 
Release date: 10/09/2020
OS: Darwin x64 19.6.0
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64

jgiovaresco:
Closes #2615

### DIFF
--- a/packages/insomnia-app/package-lock.json
+++ b/packages/insomnia-app/package-lock.json
@@ -10893,9 +10893,9 @@
 			}
 		},
 		"httpsnippet": {
-			"version": "1.22.0",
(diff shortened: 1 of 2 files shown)


develohpanda (approved):
💯

PR #2671, opened 2020-09-28
Basic auth e2e test

develohpanda:
This is a follow-up to https://github.com/Kong/insomnia/pull/2642#pullrequestreview-495945646.

I've replaced `tinyhttp` with `express` because express has a larger library of existing middleware to help speed up and simplify development. For example, with this test I add the `express-basic-auth` middleware.

20 successful runs with the latest version

image


packages/insomnia-app/app/ui/components/request-pane.js
```diff
@@ -288,7 +288,7 @@ class RequestPane extends React.PureComponent<Props> {
                 <i className="fa fa-caret-down space-left" />
               </ContentTypeDropdown>
             </Tab>
-            <Tab tabIndex="-1">
+            <Tab id="auth-dropdown-tab" tabIndex="-1">
```
jgiovaresco: I don't understand why we had an id here ? I was expecting to see a selector using it in tests but I didn't find it ?
Is it a standard rule ?

jgiovaresco (commented):
👏 
It is a nice way to test the app, quite easy to review one everything is in place.

We may need to split `debug.js` file in the future because it will become huge as we will write more tests. Maybe we could extract authentication related functions (`clickBasicAuth`, `clickRequestAuthDropdown`...) in a dedicated file ?

develohpanda:
> We may need to split `debug.js` file in the future because it will become huge as we will write more tests

Yep, I think it's worth breaking down by the three panes - sidebar, request, response - which can then be broken down further for specific things (eg auth). Step by step. 😄

packages/insomnia-app/app/ui/components/request-pane.js
develohpanda: Ah, nice catch. This isn't used because `id` is not a property that the react-tabs Tab component uses. There isn't a way (from what I could see) to set a unique id, so the test is just using the generated one for now. Will remove.

I think it's worth creating a tab selection helper, to allow you to click according to the tab _contents_.

packages/insomnia-smoke-test/modules/debug.js
```diff
@@ -74,3 +87,77 @@ export const getPdfCanvas = async app => {
...
+
+  // Allow username changes to persist
+  await app.client.pause(100);
+
+  await passwordEditor.keys(password);
+
+  // Allow password changes and encoding checkbox to persist
+  await app.client.pause(100);
+
+  await app.client.$('button#use-iso-8859-1').then(e => e.click());
```
DMarby: It doesn't feel super obvious to me that a helper method for entering basic auth would select a non-standard encoding option.

If we want to test this encoding, perhaps we should make this parameter and test both cases?

DMarby (changes requested):

DMarby (approved):

packages/insomnia-smoke-test/server/index.js
```diff
@@ -0,0 +1,49 @@
...
+import basicAuth from 'express-basic-auth';
+import { basicAuthCreds } from '../fixtures/constants.js';
+
+const app = express();
+const basicAuthRouter = express.Router();
+const port = 4010;
+
+// Artificallly slow each request down
+app.use((req, res, next) => {
+  setTimeout(next, 500);
```
develohpanda, line 11: We can experiment with reducing this `500ms` delay until it is no longer than necessary. As it stands, I think it's fine. Longest test (this one) is `16s` and it tests every aspect of basic auth

DMarby (approved):

PR #2676, opened 2020-09-29
Toast based logo fixes, consistent names, fixing positioning

sonicyeti:
This PR covers the following:

- Shift away from gravatar based logos, sourcing locally instead
- Updated core and designer logos per latest designs
- Address padding and general spacing inconsistencies

**Previous Version:**
Screen Shot 2020-09-29 at 3 23 14 PM

**Updated version in both core and designer:**
Screen Shot 2020-09-29 at 1 40 18 PM
Screen Shot 2020-09-29 at 1 37 11 PM
 
Fixes INS-129


packages/insomnia-app/app/ui/components/toast.js
```diff
@@ -179,10 +199,10 @@ class Toast extends React.PureComponent<Props, State> {
         className={classnames('toast theme--dialog', {
           'toast--show': visible,
         })}>
-        <div className="toast__image">
-          <GravatarImg email={notification.email} fallback={appIconSrc} size={100} rounded />
-        </div>
-        <div className="toast__content">
+        <StyledLogo>
+          <img src={getAppName() === 'Insomnia' ? imgSrcCore : imgSrcDesigner} />
```
develohpanda: One minor suggestion is to use `getDefaultAppId() === APP_ID_INSOMNIA` instead of the application name check, because the ID will basically _never_ change. We use appId to check for similar things in other areas as well. 😄 

Also, I think this `img` needs the `alt` attribute.

packages/insomnia-app/app/ui/css/components/toast.less
```diff
@@ -23,18 +23,6 @@
     right: var(--padding-md);
   }
 
-  .toast__image,
-  .toast__content {
-    height: 100%;
-    width: 100%;
-  }
-
-  .toast__message {
```
develohpanda, line 32: It looks like this class is still used in `toast.js` - does the CSS need to remain, or does the usage of the class need to be removed in the component?

develohpanda (approved):
Spacing looks so much cleaner now 😍 

Code looks good! If it's not a lot of work it would be awesome to remove `toast.less` entirely and move that stuff to styled-components, rather than having half-half. Doesn't seem too involved at first sight.

packages/insomnia-app/app/ui/components/toast.js
sonicyeti: Gotcha, ok sounds good and I'll roll in alt tags.

DMarby: Agreed that using getDefaultAppId sounds like a good idea 👍

DMarby (commented):
Nice work! 🎉

DMarby (approved):

packages/insomnia-app/app/ui/components/toast.js
sonicyeti: Right on, I have it swapped out although I ended up having to use getAppName for the alt tags anyhow, bubbled that up to state so we aren't pulling it excessively though at least. PR coming shortly, pulling gravatar.

DMarby:
@sonicyeti @develohpanda Do you think it makes sense to remove the `email` property from the ToastNotification component that was previously used for Gravatar in this PR as well, or would that be better done in a separate one?

sonicyeti:
> Code looks good! If it's not a lot of work it would be awesome to remove `toast.less` entirely and move that stuff to styled-components, rather than having half-half. Doesn't seem too involved at first sight.

Pulled footer actions out as well but decided to leave the wrapper as-is given its leveraging the classnames module and controls theming.

DMarby (approved):

develohpanda (approved):
Nice! Looks awesome

PR #2678, opened 2020-09-29
Add support for request action plugins

### LINKED ISSUE #2119: Request/Respone actions for plugins
Currently, there are actions support for workspaces and folders, but not for individual requests itself when building plugins. It would be awesome to be able to write plugins that can use used at a request/response context also. 

Often times, I am using quicktype to convert json, graphql, or json schema responses to various types depending on the lang I am using. Quicktype supports various languages that are similar to code generator, while providing a [core interface for js](https://www.npmjs.com/package/quicktype#calling-quicktype-from-javascript)

Based on the description above, it would be great if plugins could be developed per request response context. Based on the above problem, maybe there there could be a core plugin around quicktype that could convert various response types to code similar to how `Generate Code` works? If not, at the least, it would be nice to be able to build plugins for each request context and then handle the response. This can hopefully support both response and request interfaces based on the above example.

Would be great also if there was an output window similar to `Generate Code`. 

**Additional context**
This feature exists in the `Generate Code` option of a a request, but would be nice to be able to extend on that. 

image

nijikokun:
# Goal

This PR provides plugin developers the same type of action hook available to workspaces and request folders for individual requests. See the documentation and examples below.

# Backstory

While looking through a [recently submitted plugin](https://insomnia.rest/plugins/insomnia-plugin-swagerize-response) I realized that this individual was using the existence of a header property to create an action for a request. While it is an [ingenious work-around](https://user-images.githubusercontent.com/198276/94623922-28d88700-026a-11eb-8242-f253920d0f63.png), it showed that there is an obvious gap.

You can create actions on request folders, and on the workspace, but not the actual requests themselves.

# Documentation

## Request Actions

Actions can be added to the bottom of the request dropdown by defining a request action plugin.

```ts
type RequestAction = {
    label: string,
    action: (context: Context, { 
        requestGroup: RequestGroup, 
        request: Request
    }): void | Promise<void>,
    label: string,
    icon?: string,
};

// Request actions are exported as an array of objects
module.exports.requestActions = Array<RequestAction>
```

<details>
 <summary>Example: Plugin to get request details in a modal</summary>

> Screen Shot
> Screen Shot 2020-09-29 at 2 50 13 PM
> 
> ```ts
> module.exports.requestActions = [
>   {
>     label: "See request data",
>     action: async (context, data) => {
>       const { request } = data;
>       const html = `<code>${JSON.stringify(request)}</code>`;
>       context.app.showGenericModalDialog("Results", { html });
>     },
>   },
> ];
> ```
</details>

<details>
 <summary>Example: Send request</summary>

> ```ts
> module.exports.requestActions = [
>   {
>     label: "Send request",
>     action: async (context, data) => {
>       const { request } = data;
>.      const response = await context.network.sendRequest(request);
>       const html = `<code>${request.name}: ${response.statusCode}</code>`;
>       context.app.showGenericModalDialog("Results", { html });
>     },
>   },
> ];
> ```
</details>

# Additional Details

- Replicates implementation of already existing feature `RequestGroupActions`
- Converts `RequestActionsDropdown` to be stateful component to support plugin actions
- Adds support for `activeEnvironment` to the `RequestActionsDropdown` for network requests

# Open questions

- What do you think about putting plugins above `settings` instead of below, this way we can slowly move actions over to being plugins, and their position does not jump around (Generate Code for example is an example).

Closes https://github.com/Kong/insomnia/issues/2119

### DIFF
--- a/packages/insomnia-app/app/plugins/index.js
+++ b/packages/insomnia-app/app/plugins/index.js
@@ -41,6 +41,19 @@ export type RequestGroupAction = {
   icon?: string,
 };
 
+export type RequestAction = {
+  plugin: Plugin,
+  action: (
+    context: Object,
+    models: {
+      requestGroup: RequestGroup,
+      request: Array<Request>,
+    },
+  ) => void | Promise<void>,
+  label: string,
+  icon?: string,
+};
+
 export type WorkspaceAction = {
   plugin: Plugin,
   action: (
@@ -212,6 +225,16 @@ export async function getRequestGroupActions(): Promise<Array<RequestGroupAction
   let extensions = [];
   for (const plugin of await getActivePlugins()) {
     const actions = plugin.module.requestGroupActions || [];
+    extensions = [...extensions, ...actions.map(p => ({ plugin, ...p }))];
+  }
+
+  return extensions;
+}
+
+export async function getRequestActions(): Promise<Array<RequestAction>> {
+  let extensions = [];
+  for (const plugin of await getActivePlugins()) {
+    const actions = plugin.module.requestActions || [];
     extensions = [...extensions, ...actions.map(p => ({ plugin, ...p }))];
   }
 
--- a/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
@@ -99,23 +99,24 @@ class RequestActionsDropdown extends React.PureComponent<Props, State> {
   async _handlePluginClick(p: RequestAction) {
     this.setState(state => ({ loadingActions: { ...state.loadingActions, [p.label]: true } }));
 
-    try {
-      const { activeEnvironment, request, requestGroup } = this.props;
-      const activeEnvironmentId = activeEnvironment ? activeEnvironment._id : null;
+    if (p.action) {
+      try {
+        const { activeEnvironment, request, requestGroup } = this.props;
+        const activeEnvironmentId = activeEnvironment ? activeEnvironment._id : null;
+        const context = {
+          ...(pluginContexts.app.init(RENDER_PURPOSE_NO_RENDER): Object),
+          ...(pluginContexts.data.init(): Object),
+          ...(pluginContexts.store.init(p.plugin): Object),
+          ...(pluginContexts.network.init(activeEnvironmentId): Object),
+        };
 
-      const context = {
-        ...(pluginContexts.app.init(RENDER_PURPOSE_NO_RENDER): Object),
-        ...(pluginContexts.data.init(): Object),
-        ...(pluginContexts.store.init(p.plugin): Object),
-        ...(pluginContexts.network.init(activeEnvironmentId): Object),
-      };
-
-      await p.action(context, { request, requestGroup });
-    } catch (err) {
-      showError({
-        title: 'Plugin Action Failed',
-        error: err,
-      });
+        await p.action(context, { request, requestGroup });
+      } catch (err) {
+        showError({
+          title: 'Plugin Action Failed',
+          error: err,
+        });
+      }
     }
 
     this.setState(state => ({ loadingActions: { ...state.loadingActions, [p.label]: false } }));
@@ -173,7 +174,10 @@ class RequestActionsDropdown extends React.PureComponent<Props, State> {
 
         {actionPlugins.length > 0 && <DropdownDivider>Plugins</DropdownDivider>}
         {actionPlugins.map((p: RequestAction) => (
-          <DropdownItem key={p.label} onClick={() => this._handlePluginClick(p)} stayOpenAfterClick>
+          <DropdownItem
+            key={`${p.plugin.name}::${p.label}`}
+            onClick={() => this._handlePluginClick(p)}
+            stayOpenAfterClick>
             {loadingActions[p.label] ? (
               <i className="fa fa-refresh fa-spin" />
             ) : (
--- a/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
@@ -1,5 +1,6 @@
 // @flow
-import React from 'react';
+import React from 'react'; // @flow
+import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import PromptButton from '../base/prompt-button';
 import {
@@ -87,7 +88,7 @@ class RequestActionsDropdown extends React.PureComponent<Props, State> {
     return requestOperations.remove(request);
   }
 
-  async onOpen() {
+  async _onOpen() {
     const plugins = await getRequestActions();
     this.setState({ actionPlugins: plugins });
   }
@@ -135,7 +136,7 @@ class RequestActionsDropdown extends React.PureComponent<Props, State> {
     const canGenerateCode = isRequest(request);
 
     return (
-      <Dropdown ref={this._setDropdownRef} onOpen={this.onOpen} {...other}>
+      <Dropdown ref={this._setDropdownRef} onOpen={this._onOpen} {...other}>
         <DropdownButton>
           <i className="fa fa-caret-down" />
         </DropdownButton>
--- a/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
@@ -97,7 +97,7 @@ class RequestActionsDropdown extends React.PureComponent<Props, State> {
   }
 
   async _handlePluginClick(p: RequestAction) {
-    this.setState((state) => ({ loadingActions: { ...state.loadingActions, [p.label]: true } }));
+    this.setState(state => ({ loadingActions: { ...state.loadingActions, [p.label]: true } }));
 
     try {
       const { activeEnvironment, request, requestGroup } = this.props;
@@ -117,7 +117,7 @@ class RequestActionsDropdown extends React.PureComponent<Props, State> {
       });
     }
 
-    this.setState((state) => ({ loadingActions: { ...state.loadingActions, [p.label]: false } }));
+    this.setState(state => ({ loadingActions: { ...state.loadingActions, [p.label]: false } }));
     this._dropdown && this._dropdown.hide();
   }
 
--- a/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
@@ -174,7 +174,8 @@ class RequestActionsDropdown extends React.PureComponent<Props, State> {
         {actionPlugins.map((p: RequestAction) => (
           <DropdownItem
             key={`${p.plugin.name}::${p.label}`}
-            onClick={() => this._handlePluginClick(p)}
+            value={p}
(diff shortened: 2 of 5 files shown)


packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -95,20 +166,21 @@ class RequestActionsDropdown extends PureComponent {
           <i className="fa fa-wrench" /> Settings
           <DropdownHint keyBindings={hotKeyRegistry[hotKeyRefs.REQUEST_SHOW_SETTINGS.id]} />
         </DropdownItem>
+
+        {actionPlugins.length > 0 && <DropdownDivider>Plugins</DropdownDivider>}
+        {actionPlugins.map((p: RequestAction) => (
+          <DropdownItem key={p.label} onClick={() => this._handlePluginClick(p)} stayOpenAfterClick>
+            {loadingActions[p.label] ? (
```
develohpanda: I think it might be good practice to prefix the component key and loadingActions key with the plugin name, something like `{p.plugin.name}::{p.label}`, if for whatever reason two plugin actions have the same label.

We don't have any control over ensuring two plugins _don't_ have the same label, but can be fairly confident that plugin name + label would be unique.

packages/insomnia-app/app/ui/components/sidebar/sidebar-children.js
```diff
@@ -122,6 +122,8 @@ class SidebarChildren extends React.PureComponent<Props> {
             request={child.doc}
             workspace={workspace}
             hotKeyRegistry={hotKeyRegistry}
+            // Necessary for plugin actions on requests
```
develohpanda, line 125: 👍

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -13,8 +13,46 @@ import * as models from '../../../models';
...
+import type { Request } from '../../../models/request';
+import type { RequestGroup } from '../../../models/request-group';
+import type { Environment } from '../../../models/environment';
+import { getRequestActions } from '../../../plugins';
+import * as pluginContexts from '../../../plugins/context/index';
+import { showError } from '../modals';
+import { RENDER_PURPOSE_NO_RENDER } from '../../../common/render';
+import classnames from 'classnames';
+
+type Props = {
```
develohpanda, line 31: Normally I would suggest typing (especially the callbacks) more explicitly than `Function`, but seeing as this is a copy-paste-ish from `prop-types` that's fine, since you didn't actually work with those callbacks. 👍

packages/insomnia-app/app/plugins/index.js
```diff
@@ -41,6 +41,19 @@ export type RequestGroupAction = {
...
+  plugin: Plugin,
+  action: (
+    context: Object,
+    models: {
+      requestGroup: RequestGroup,
+      request: Array<Request>,
+    },
+  ) => void | Promise<void>,
+  label: string,
+  icon?: string,
```
develohpanda, line 54: I think `icon` is missing in the documentation update you have in the description, just a note to add it there as well so it can be copied across to HelpScout.

packages/insomnia-app/app/plugins/index.js
```diff
@@ -41,6 +41,19 @@ export type RequestGroupAction = {
...
 
+export type RequestAction = {
+  plugin: Plugin,
+  action: (
+    context: Object,
+    models: {
+      requestGroup: RequestGroup,
+      request: Array<Request>,
+    },
+  ) => void | Promise<void>,
```
develohpanda, line 52: Docs in the description need to be updated as they have this typed as returning only `Promise<void>`.

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -45,8 +83,39 @@ class RequestActionsDropdown extends PureComponent {
...
+      const activeEnvironmentId = activeEnvironment ? activeEnvironment._id : null;
+
+      const context = {
+        ...(pluginContexts.app.init(RENDER_PURPOSE_NO_RENDER): Object),
+        ...(pluginContexts.data.init(): Object),
+        ...(pluginContexts.store.init(p.plugin): Object),
+        ...(pluginContexts.network.init(activeEnvironmentId): Object),
+      };
+
+      await p.action(context, { request, requestGroup });
```
develohpanda, line 113: `action` has a return type of `void | Promise<void>` - I believe this would throw an error if we try to await and action doesn't return a Promise.

develohpanda (commented):
Nicely done, I appreciate the time you put into the PR description 💯

A couple of notes worth addressing but generally LGTM

> What do you think about putting plugins above settings instead of below, this way we can slowly move actions over to being plugins, and their position does not jump around (Generate Code for example is an example).

I'd say to leave plugins below settings for now. I don't believe there is currently any control over the ordering of plugins, so some thought will need to be put towards that, in order to ensure that the other actions converted to plugins (eg pin, delete, duplicate, generate code) stay towards the top, while new/other plugins have their own order as well. The ordering will need to be applicable for request group and workspace plugins too.

develohpanda:
Fixes INS-168

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
nijikokun: I'd say we do this change in a separate PR, since this should also be done for workspaces / request groups if we decide to go this route.

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
nijikokun: Yeah, would need more debugging to infer the actual arguments passed so kept it the same as before, agree it should be addressed later

packages/insomnia-app/app/plugins/index.js
nijikokun: Updated the description, should also do the same for the other actions :) since it's from them

packages/insomnia-app/app/plugins/index.js
nijikokun: Done

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
nijikokun: This is a good catch, and I assume why it's typed that way on the site, but in the codebase, all actions use `void | ...` we should simply remove the `void | ` from this action and investigate the others before doing the same

nijikokun:
> I'd say to leave plugins below settings for now. I don't believe there is currently any control over the ordering of plugins, so some thought will need to be put towards that, in order to ensure that the other actions converted to plugins (eg pin, delete, duplicate, generate code) stay towards the top, while new/other plugins have their own order as well. The ordering will need to be applicable for request group and workspace plugins too.

Ah, perhaps my intent wasn't clear, but I do agree about ordering, however, it was more placing plugins _above_ settings and _below_ the application items. It looks awkward in its current form, imo. But, I think I'm okay with this for now.

develohpanda:
> it was more placing plugins _above_ settings and _below_ the application items

Oh I get ya, misunderstood before. That'd look nice I agree. The slightly different dropdown divider UI does look kind of strange as well 😅

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
develohpanda: I think it's just looking at the usages, most of the callbacks boil up to `wrapper-*.js` or `app.js`, but typing is lost along the way. I tend to fix up components as I work on them.

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
develohpanda: Fair enough, keen to see them updated as well. Please create a separate PR to update workspace/request group IDs, and change the request one in this PR as part of this feature.

jgiovaresco:
>  it was more placing plugins above settings and below the application items.

"application items" are `Duplicate`, `Generate Code`, `Pin`, `Copy as Curl` and `Delete` ?

I would prefer putting plugins actions above Settings. But I can't find any argument I guess it is just a personal preference 🤷‍♂️

nijikokun:
> "application items" are `Duplicate`, `Generate Code`, `Pin`, `Copy as Curl` and `Delete` ?

Yes, 

> I would prefer putting plugins actions above Settings.

I agree, it seems better, will update

sonicyeti (approved):
Very cool!

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
nijikokun: After further review, gong to mark this as resolved since the above code is fine. This is because `action` *must be a function* according to its type declaration and calling `await` on a `function` works without issue regardless of the function's _return value_.

See the example below:

```
async function exampleOne() {
  return undefined;
}

async function exampleTwo() {
  return Promise.resolve(true);
}

(async function () {
  await exampleOne();
  await exampleTwo();
})();
```

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -45,8 +83,38 @@ class RequestActionsDropdown extends PureComponent {
     models.request.remove(request);
   }
 
-  show() {
-    this._dropdown.show();
+  async onOpen() {
```
develohpanda: _nit_ should this be `_onOpen` following the general pattern of private methods prefixed by `_`? (It seems private to me, LMK if I'm missing something)

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -89,6 +159,21 @@ class RequestActionsDropdown extends PureComponent {
           <DropdownHint keyBindings={hotKeyRegistry[hotKeyRefs.REQUEST_SHOW_DELETE.id]} />
         </DropdownItem>
 
+        {actionPlugins.length > 0 && <DropdownDivider>Plugins</DropdownDivider>}
+        {actionPlugins.map((p: RequestAction) => (
+          <DropdownItem
+            key={`${p.plugin.name}::${p.label}`}
+            onClick={() => this._handlePluginClick(p)}
```
develohpanda: I have recently learned that if you specify the `value` prop, `DropdownItem` will feed that back as the first argument to `onClick`, so we can avoid creating an inline function here 🎉 

https://github.com/Kong/insomnia/blob/aeafe4a7c92b1d5e93f4dfe7ac9a699b0804bf45/packages/insomnia-app/app/ui/components/base/dropdown/dropdown-item.js#L19-L23

```suggestion
            value={p}
            onClick={this._handlePluginClick}
```

develohpanda (approved):
LGTM! Minor comments 🎉

packages/insomnia-app/app/ui/components/dropdowns/request-actions-dropdown.js
```diff
@@ -1,4 +1,5 @@
-import React, { PureComponent } from 'react';
+// @flow
+import React from 'react'; // @flow
```
develohpanda:
```suggestion
import React from 'react';
```

PR #2685, opened 2020-10-02
Added nested keys check after root keys check for environment editor.

### LINKED ISSUE #2684: Validate environment variables against nested keys
There is a bug here in that both previously and now, only the root keys are checked. A recursive algorithm needs to be created to validate nested keys appropriately.

_Originally posted by @develohpanda in https://github.com/Kong/insomnia/pull/2601#discussion_r498003061_

Root keys must follow these rules (as currently implemented):
- A key cannot be simply `_`
- A key cannot start with `$`
- A key cannot contain `.`

Nested keys must follow these rules:
- A key cannot start with `$`
- A key cannot contain `.`

Note, these rules do not apply for array _items_, but a complex object inside an array, must have keys that follow the nested rules.

<details>
<summary>Context about $ and . rules</summary>

NeDB has rules to prevent a key from starting with $ or contain a ., so I have updated the regex key check for the environment editor. In the current implementation, errors are presented in console, and the changes do not persist (semi-silent failure), if you try to save something without following these rules which the existing regex did not check for.

</details>

There is some example code [here](https://github.com/Kong/insomnia/blob/1bbbf38a9653792fd137662d015846f4334b1989/packages/insomnia-app/app/templating/utils.js#L33-L60)  that does a similar recursive process.

johnhorsema:
Closes #2684


develohpanda (commented):
Nice first attempt 👏 

After #2601, keys themselves can contain spaces, so after calling `getNestedKeys`, splitting on a single space won't quite work as intended. Additionally, `getNestedKeys` will iterate over all keys first, and then check each, one by one. This needs to be made much more efficient in that as soon as a key that does not follow the regex rule is detected, the algorithm stops.

Remember, the example linked in the issue is an _example_, from which a new solution needs to be derived for the `environment-editor` specifically. 😄 Let me know if you'd like further direction here, but I'll let you improve the approach with this feedback in mind.

Lastly, please add some screenshots for what the user will see when adding a nested key which does not follow the rules, as well as some unit tests in `environment-editor.test.js`.

johnhorsema:
> Lastly, please add some screenshots for what the user will see when adding a nested key which does not follow the rules, as well as some unit tests in `environment-editor.test.js`.

Screenshots

johnhorsema:
@develohpanda please have a look at my latest PR, thanks

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -22,6 +22,48 @@ export const ensureKeyIsValid = (key: string): string | null => {
   return null;
 };
 
+export const ensureRootKeyIsValid = (key: string): string | null => {
+  if (key === '_') {
+    return `"${key}" cannot be '_'`;
+  }
```
develohpanda: The check for root keys (and subsequent error message) needs to be removed from the `ensureKeyIsValid` function and moved into here.

https://github.com/Kong/insomnia/blob/0d522a69a7efe8d5296b4d91791d1aedd851619b/packages/insomnia-app/app/ui/components/editors/environment-editor.js#L18-L20

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -22,6 +22,48 @@ export const ensureKeyIsValid = (key: string): string | null => {
...
+        result = ensureRootKeyIsValid(key);
+        break;
+      } else if (ensureKeyIsValid(key) !== null) {
+        result = ensureKeyIsValid(key);
+        break;
+      }
+    }
+    // Case: Nested keys
+    if (typeof obj[key] === 'object' && obj.hasOwnProperty(key)) {
+      result = checkNestedKeys(obj[key], result, false);
```
develohpanda: I don't quite understand why result is passed into the recursive call here. Does it need to be a parameter of `checkNestedKeys` at all?

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -22,6 +22,48 @@ export const ensureKeyIsValid = (key: string): string | null => {
...
+/**
+ * Recursively check keys of nested object and immediately return when found
+ * @param {object} obj - object to analyse
+ * @returns {boolean} - if any invalid key is found
+ */
+export function checkNestedKeys(
+  obj: any,
+  result: string | null = '',
+  isRoot: boolean = true,
+): boolean {
```
develohpanda: The return type is incorrect, here and in the comment above. It should be `string | null`.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -22,6 +22,48 @@ export const ensureKeyIsValid = (key: string): string | null => {
...
+  for (const key in obj) {
+    // Case: Root keys
+    if (isRoot) {
+      if (ensureRootKeyIsValid(key) !== null) {
+        result = ensureRootKeyIsValid(key);
+        break;
+      } else if (ensureKeyIsValid(key) !== null) {
+        result = ensureKeyIsValid(key);
+        break;
+      }
```
develohpanda: Running a function, checking the result, and then running the function again is quite an inefficient workflow.

But, you can simplify it using [short-circuit evaluation](https://codeburst.io/javascript-what-is-short-circuit-evaluation-ff22b2f5608c):

```js
result = ensureRootKeyIsValid(key) || ensureKeyIsValid(key);
```

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -22,6 +22,48 @@ export const ensureKeyIsValid = (key: string): string | null => {
...
+      result = checkNestedKeys(obj[key], result, false);
+    }
+    if (ensureKeyIsValid(key) !== null) {
+      result = ensureKeyIsValid(key);
+      break;
+    }
+  }
+
+  return result;
+}
```
develohpanda: I think this overall function (`checkNestedKeys`) can be simplified a fair bit. Root vs non-root keys aren't two different cases.

For example, the only conditional statement inside `if (isRoot)` needs to be `ensureRootKeyIsValid`, because `ensureKeyIsValid` needs to be applied to _all_ keys.

Also, we should only move further into the recursion, if we are absolutely sure all keys in the current level are valid, exiting as early as possible.

See if you can simplify this a little bit. 😄

develohpanda (changes requested):
Making some progress 👏 a couple of comments, particularly with the efficiency and clarity of the recursive function.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -22,6 +22,48 @@ export const ensureKeyIsValid = (key: string): string | null => {
...
+      if (ensureRootKeyIsValid(key) !== null) {
+        result = ensureRootKeyIsValid(key);
+        break;
+      } else if (ensureKeyIsValid(key) !== null) {
+        result = ensureKeyIsValid(key);
+        break;
+      }
+    }
+    // Case: Nested keys
+    if (typeof obj[key] === 'object' && obj.hasOwnProperty(key)) {
```
develohpanda: Do objects and arrays need to be treated differently? Note that in the following environment, `$val.id` is valid, but `$in.valid` is invalid.

```json
{
  "arr": [
    "$val.id",
    {
      "$in.valid": 1
    }
  ]
}
```

You might need to use `const typeOfObj = Object.prototype.toString.call(obj);` to differentiate between an array and an object.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
johnhorsema: You're right.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
johnhorsema: It's refactored to just have `ensureKeyIsValid`.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
johnhorsema: Removed and replaced with a local variable in the function.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
johnhorsema: Part of refactoring to single `ensureKeyIsValid`.

packages/insomnia-app/app/ui/components/editors/environment-editor.js
johnhorsema: No changes needed here. The check `typeof obj[key]` will return `object` for both objects and arrays.
So `environmentObj["arr"]` will still pass the condition and be checked for subsequent valid keys in its nested members.

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

packages/insomnia-app/app/ui/components/editors/environment-editor.js
johnhorsema: Sure. I've refactored the `ensureKeyIsValid` to incorporate root key checking.

johnhorsema:
@develohpanda please have a look at my latest PR, thanks

develohpanda:
@johnhorsema Thanks! Don't worry, I haven't missed it, just been quite busy. I will re-review when I get a chance to. 🤗

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -10,18 +10,40 @@ import { NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME } from '../../../templating';
...
+      result = checkNestedKeys(obj[key], false);
+    } else {
+      result = ensureKeyIsValid(key, isRoot);
+    }
+    if (result !== null) {
+      break;
+    }
+  }
+
+  return result;
```
develohpanda: It should be safe to simply exit without a null check, since `checkNestedKeys` has a return type that includes `null`. In addition, further down, there is a check to only set an error if an error is returned, so we don't need to double-up checks.

Also, it feels as though the `obj.hasOwnProperty(key)` check is obsolete, because `typeof obj[key]` returns `undefined` if the key is not present. The second condition will never be reached.

```suggestion
  for (const key in obj) {
    // Nested keys
    if (typeof obj[key] === 'object') {
      return checkNestedKeys(obj[key], false);
    } else {
      return ensureKeyIsValid(key, isRoot);
    }
  }
```

packages/insomnia-app/app/ui/components/editors/environment-editor.js
```diff
@@ -10,18 +10,40 @@ import { NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME } from '../../../templating';
 // Docs: https://github.com/DeNA/nedb#inserting-documents
 const INVALID_NEDB_KEY_REGEX = /^\$|\./;
 
-export const ensureKeyIsValid = (key: string): string | null => {
+export const ensureKeyIsValid = (key: string, isRoot: boolean = false): string | null => {
```
develohpanda: Is there a particular reason you chose to set `isRoot` to default false here, but default true in `checkNestedKeys`?

It feels unnecessarily complex and I would prefer to remove the defaults altogether.

packages/insomnia-app/app/ui/components/editors/__tests__/environment-editor.test.js
```diff
@@ -15,13 +16,17 @@ describe('ensureKeyIsValid', () => {
...
+  it.each(['_'])('%s should be invalid when key is _', key => {
+    expect(ensureKeyIsValid(key, true)).toBe(`"${key}" is a reserved key`);
   });
 
-  it.each(['a', 'ab', 'a$', 'a$b', 'a-b', `a${name}b`, `${name}ab`])('%s should be valid', key => {
-    expect(ensureKeyIsValid(key)).toBe(null);
+  it.each(['a', 'ab', 'a$', 'a$b', 'a-b', `a b`])('%s should be valid', key => {
+    expect(ensureKeyIsValid(key, true)).toBe(null);
   });
 });
```
develohpanda: I noticed there are no tests for `checkNestedKeys`; could you please add some permutations for that?

develohpanda (changes requested):
Sorry about the delay, I finally had a chance to review this. I'm generally happy with it, did some testing locally and it seems to be working nicely.

I've added a couple of comments that are worth addressing, please! 🙂

develohpanda (approved):
QA'd with the following environment (and by changing it round): https://www.loom.com/share/1fa103f7ce0141c286da62dad146e147.

@gatzjames @dimitropoulos the rules for root and nested keys are defined in https://github.com/Kong/insomnia/issues/2684.

```json
{
  "base-url": "https://api.insomnia.rest",
  "nested": {
    "path-with-hyphens": "/path-with-hyphen"
  },
  "ar-ray": [
    "/first",
    {
      "second": "second"
    },
    {
      "third": "third"
    }
  ]
}
```

Thanks for working through this, apologies for the mega delays; we're working through the older PRs.

dimitropoulos (approved):
works great.  I really appreciate all the tests, too!

PR #2694, opened 2020-10-06
Fix select all with large response

### LINKED ISSUE #2478: Select all text operation is painfully slow with a large response
**Describe the bug**
Code-mirror is great at efficiently rendering a very large amount of text, however both the response focus hotkey (`cmd + '` by default, which selects all text) and the select all (`cmd + a`) operations are _painfully_ slow, and also cause the application to become unresponsive.

Documents:
- 206 KB `https://github.com/tpn/pdfs/raw/master/A%20Comparison%20of%20Programming%20Languages%20in%20Economics%20(16-Jun-2014).pdf`
- 2.2 MB `https://eloquentjavascript.net/Eloquent_JavaScript.pdf`

**To Reproduce**
1. Create a new get request for one of the PDF documents above 
2. Focus the response with the hotkey `cmd + '` OR click on the response and select all `cmd + a`
3. Observe delay, and that the application becomes unresponsive

**Expected behavior**
1. Responsiveness
    - Don't select all if large response? the `response-viewer` has a `blockingBecauseTooLarge` state, which we could use to ignore the select-all step during the response focus hotkey handler https://github.com/Kong/insomnia/blob/e3f9e3c1d4854474053e4ef5b037f97a4fb1aa51/packages/insomnia-app/app/ui/components/viewers/response-viewer.js#L182-L195
2. A faster way to copy all contents
    - Potentially add a `CopyButton` to all appropriate response viewers? image

 - OS: macOS 10.15.6
 - Installation Method: DEV
 - App Version: v2020.3.3

Loosely related to https://github.com/Kong/insomnia/issues/2397

jgiovaresco:
Closes #2478, Closes INS-921


jgiovaresco:
I've tried 2 simples things for the CopyButton position:

Absolute top right:
position-absolute-right


Float right
float-right


As you can see the absolute position is not a good choice, it will hide part of the response...
Using a simple `float: right` would do the job, but we "loose" some space before the response body.

I will try to find something else...

develohpanda:
Hmm, thanks for trying this. Both aren't ideal, one covers text and the other adds a lot of blank space. I'm not quite sure what the best option for this is, yet.

jgiovaresco:
I will try to add an action in the Preview dropdown to copy the response data in the clipboard.

develohpanda:
> 
> 
> I will try to add an action in the Preview dropdown to copy the response data in the clipboard.

👍 

@nijikokun / @erickrico if you have any feedback here that'd be awesome. Some context in #2478. tl;dr is, the `ctrl + a` select-all action is quite slow when selecting a large response (due to CodeMirror), so an option is to put a copy-to-clipboard button _somewhere_, but we are unsure on UI placement. There are some options suggested in the issue as well.

jgiovaresco:
I've added an action `Copy raw response` in the Preview dropdown

image

nijikokun:
The above, inside the dropdown, seems like a good solution for now. That said, I do fear that it might be hidden and a user will encounter the same issue that was originally raised inside of #2478.

My personal opinion on the placement (to avoid overlap and gaps) is to place it here:

Screen Shot 2020-10-19 at 6 10 28 PM

Or here:

Screen Shot 2020-10-19 at 6 11 20 PM

cc @erickrico

jgiovaresco:
It looks like this when the button in the Response pane header

image

develohpanda:
**My two cents:** 

Placing it in the _top_ header disconnects it from what it relates to. Additionally, the copy button is also quite closely tied to which _response viewer_ is currently shown. If the PDF viewer is shown, the copy button doesn't apply, but it does apply for JSON and raw viewers, for instance.

I don't have high hopes for this, but if it's at all possible to _wrap_ text around the button, _should_ it ever be in a position to overlap, that'd be awesome (see gif [here](https://github.com/Kong/insomnia/issues/2478#issuecomment-668366640)). It will need investigation into CodeMirror to determine whether it is achievable.

erickrico:
> **My two cents:**
> 
> Placing it in the _top_ header disconnects it from what it relates to. Additionally, the copy button is also quite closely tied to which _response viewer_ is currently shown. If the PDF viewer is shown, the copy button doesn't apply, but it does apply for JSON and raw viewers, for instance.
> 
> I don't have high hopes for this, but if it's at all possible to _wrap_ text around the button, _should_ it ever be in a position to overlap, that'd be awesome (see gif [here](https://github.com/Kong/insomnia/issues/2478#issuecomment-668366640)). It will need investigation into CodeMirror to determine whether it is achievable.

i agree, especially if the copy button disappear one you switch to a different tab (header cookies, timeline) i would suggest to keep it under the "preview dropdown"

nijikokun:
I agree with the discussions above, let's keep in the dropdown for now. Given that we are also on an older version of codemirror this might be resolved when we upgrade in the future.

develohpanda (approved):
We are making an effort to work through our open-source PRs, sorry about the big delays! I updated this branch with develop and QA'd with the test files available here: https://code.google.com/archive/p/jquery-speedtest/downloads.

LGTM! Works as intended. The fact that we can handle/work with 100mb responses with a similar responsiveness (pun intended) to VS Code is promising, and the copy raw response appears to be working well. https://www.loom.com/share/bce854e46d734e98a5ea25d325e50c5a

Thank you for your hard work 🙌🏽

gatzjames (approved):
🚀

PR #2696, opened 2020-10-06
Cancel the active request instead of the most recent request

### LINKED ISSUE #2690: Starting two requests and cancelling the first one does not behave as expected
**Describe the bug**
I started two long running requests, the first one takes 15 seconds and the second one takes 10 seconds. Then I go to the first request and press cancel, but nothing happens. Then I look at the second one, and see that it has been cancelled.

The _cancel request_ button cancels the last started request, not the active request. I see the correct request duration as I switch between the two, so there is _some_ logic to handle multiple active requests.

<details>
<summary>Here is an example server using express to replicate this</summary>

```js
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
    const delaySec = Number.parseInt(req.query['delaySec'] || '2');
    setTimeout(function () {
        res.send(`Delayed by ${delaySec} seconds`);
    }, delaySec * 1000);
})

app.listen(port, () => {
    console.log(`Example app listening at http://localhost:${port}`)
})
```

</details>

**To Reproduce**
1. Start request A
2. Start request B
5. Go to request A
3. Cancel request A
4. **Nothing happens**
5. Go to request B
6. **Notice it is cancelled**

**Expected behavior**
Cancel only the active request, not the last started request.

**Screenshots**

```
Version: Insomnia Core 2020.4.1
Release date: 30/09/2020
OS: Darwin x64 19.6.0
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
```

**Additional context**
It is likely that the `cancelCurrentRequest` function needs to be made smarter - perhaps storing/clearing the `cancelRequestFunction` in a dictionary indexed by the request id instead of a single function variable.

https://github.com/Kong/insomnia/blob/1bbbf38a9653792fd137662d015846f4334b1989/packages/insomnia-app/app/network/network.js#L112-L119

The timing is updated correctly when switching between two active requests, because the request start time is stored in a dictionary indexed by the request id.

https://github.com/Kong/insomnia/blob/1bbbf38a9653792fd137662d015846f4334b1989/packages/insomnia-app/app/ui/containers/app.js#L1426

selamanse:
Closes #2690

I replaced the inner network.js variable cancelRequestFunction with a Map that remembers the requestId of the given cancelFunction.

Also the ui now uses cancelRequestById with the actual requestId to cancel the specific request.

I tested package insomnia-app locally via ui with the express based test-server that was proposed by the issue description (very helpful and precise by the way).

I also ran all tests with `npm test` locally from repository root...all seem fine.

I am unsure about ui or module tests. Is there something I should add?


packages/insomnia-app/app/network/network.js
```diff
@@ -109,12 +109,15 @@ const LIBCURL_DEBUG_MIGRATION_MAP = {
...
 let lastUserInteraction = Date.now();
 
-export async function cancelCurrentRequest() {
-  if (typeof cancelRequestFunction === 'function') {
-    return cancelRequestFunction();
+export async function cancelRequestById(requestId) {
+  if (cancelRequestFunctionMap.hasOwnProperty(requestId)) {
+    const cancelRequestFunction = cancelRequestFunctionMap[requestId];
+    if (typeof cancelRequestFunction === 'function') {
+      return cancelRequestFunction();
```
develohpanda, line 119: It's surprising there is nothing that clears the `cancelRequestFunction`, it just relies on the UI to hide the button so the function cannot be called.

After invoking the function, or after a request completes, we should clear it from the map (set it back to null) so residual data doesn't stick around.

Someone from the team rightly identified that it would introduce a memory leak if we don't clear it, although it looks like there is a similar memory leak already existing with the request timing logic. We can fix that in a separate PR: https://github.com/Kong/insomnia/blob/1bbbf38a9653792fd137662d015846f4334b1989/packages/insomnia-app/app/ui/redux/modules/global.js#L84-L93

develohpanda (approved):
I'm really impressed, brilliant work 👏 

> I am unsure about ui or module tests. Is there something I should add?

Currently there are no UI tests (this is WIP as I'd like to add them too).

Some tests could potentially be added for the `cancelRequestById` function in `network.test.js` though. Looks like there are only tests for `_actuallySend`, which sets up the cancellation logic, so we might be able to add a few test cases to also cancel.

Give that a go, and if it's non-trivial to add tests for cancellation then I'll merge this in as-is. Looks pretty good to me!

develohpanda (changes requested):
Small change about clearing the cancel function once a request completes or is cancelled, to avoid a memory leak 😄

packages/insomnia-app/app/network/network.js
selamanse: I agree, I fixed the newly introduced memory leak (by this PR) and added a simple test for the cancellation routine.
I did not quite understand if the tests use a local server to mock, and if this mock server can be delayed (as in your example. So if this is possible the test could be extended to simulate the behaviour shown in the issue.

develohpanda: Just had a look at the change; it will only clear on canceling a request but we should also clear when a request completes or fails. Looking at the code, all the exist paths lead to the [`respond()`](https://github.com/Kong/insomnia/blob/fa64e6b5e896c2d043499b6b937e1a8f616d5346/packages/insomnia-app/app/network/network.js#L147-L151) helper function (including cancellation), so you might be able to clear in one convenient location.

Also, I think the tests in the current state are a false positive - they're only checking whether the property has been cleared rather than the request cancelled. When a request is cancelled, the promise returned from the network call to `_actuallySend()` [should resolve with a statusMessage of `Cancelled`](https://github.com/Kong/insomnia/blob/fa64e6b5e896c2d043499b6b937e1a8f616d5346/packages/insomnia-app/app/network/network.js#L228-L238).

`node-libcurl` is being automatically mocked [here](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/__mocks__/node-libcurl.js). You should be able to extend that, adding things necessary to configuring your unit test. Note: any changes you make to the mock should only impact your test, not the other tests. [Here's](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/sync/git/__mocks__/path.js) an example that extends the node `path` module for testing.

selamanse: @develohpanda  ...sorry took me a while, I struggled a little with the understanding of the response() helper and the promises but now I think I have a solution that cleans up corresponding cancellation Function of a request at every respond invocation.
Also changed the test expectancy as you suggested.

packages/insomnia-app/app/network/network.js
```diff
@@ -182,6 +186,14 @@ export async function _actuallySend(
...
         );
         return;
+      } finally {
+        if (cancelRequestFunctionMap.hasOwnProperty(responsePatchBeforeHooks.parentId)) {
+          delete cancelRequestFunctionMap[responsePatchBeforeHooks.parentId];
+        } else {
+          console.log(
+            `WARN: cancelFunction for requestId ${responsePatchBeforeHooks.parentId} could not be found.`,
+          );
+        }
```
develohpanda: This `finally` block won't be reached when the respond function exits on line 174, as part of the `if (noPlugins)` condition. Is that excluded intentionally?

If unintentional, I'd be inclined to suggest running this clearing logic straight after storing the timeline, on line 157, using `renderedRequest._id` as the request id.

A few more quick notes about this code-block:
- Could you please extract this clearing logic in a standalone function, like `clearCancelFunctionForId` (doesn't need to be exported, though)
- I think it's safe to remove the log message from here while clearing (as long as we keep the one you added when actually trying to cancel)

packages/insomnia-app/app/network/network.js
```diff
@@ -109,13 +109,17 @@ const LIBCURL_DEBUG_MIGRATION_MAP = {
...
-  if (typeof cancelRequestFunction === 'function') {
-    return cancelRequestFunction();
+export async function cancelRequestById(requestId) {
+  if (cancelRequestFunctionMap.hasOwnProperty(requestId)) {
+    const cancelRequestFunction = cancelRequestFunctionMap[requestId];
+    if (typeof cancelRequestFunction === 'function') {
+      return cancelRequestFunction();
+    }
   }
+  console.log(`WARN: cancelFunction for ${requestId} not found.`);
```
develohpanda: Small suggestion to update the error message here, nice observation to add it in 😄 

```suggestion
  console.log(`[network] Failed to cancel req=${requestId} because cancel function not found`);
```

packages/insomnia-app/app/network/__tests__/network.test.js
```diff
@@ -612,6 +612,51 @@ describe('actuallySend()', () => {
...
+      settings,
+    );
+
+    await networkUtils.cancelRequestById(renderedRequest1._id);
+    const response1 = await response1Promise;
+    const response2 = await response2Promise;
+
+    // THEN
+    expect(response1.statusMessage).toBe('Cancelled');
+    expect(response2.statusMessage).toBe('OK');
```
develohpanda, line 658: Awesome update to this test, it looks great 🎉

How do you feel about adding a `hasCancelFunctionForId` function?
```js
export function hasCancelFunctionForId(requestId) {
	return cancelRequestFunctionMap.hasOwnProperty(requestId);
}
``` 

We can then use it to validate at the end of this test that the cancel functions have been cleared for both requests.

develohpanda (commented):
Your update to the unit test was exactly what I had in mind, nice work!

A couple of comments, almost there 🤗

packages/insomnia-app/app/network/__tests__/network.test.js
selamanse: Had the same idea, but wasn't sure about it until you mentioned it. I think its helpful.

develohpanda: We're on the same page 👏

develohpanda (approved):
Nicely done 💯

develohpanda:
I did some manual testing:
- cancelled requests
- successful requests
- failed requests (invalid url)

Cancel functions didn't persist longer than necessary. So excited to merge this!

selamanse:
@develohpanda thanks for reviewing and advising. I appreciate this.

develohpanda:
Credit where credit is due, @DMarby noticed the memory leak, and you did a great job with fixing it up.

I was certainly impressed with your refined unit test as well! 🥇

Issue #168, opened 2017-05-08
[Improvement] Provide a way to clear OAuth window cookies/cache


gschier:
Right now, the window that prompts the user to log into the OAuth provider persists its cookies forever. This makes it hard to switch accounts if you need to test multiple OAuth users. 

A simple solution to this would be to start a new session every time the app is launched but there could also be a button somewhere to clear the current session.

At the moment, the session is defined here: https://github.com/getinsomnia/insomnia/blob/develop/app/network/o-auth-2/misc.js#L35

PR #2701, opened 2020-10-08
Don't clear oAuth2 session on restart

### LINKED ISSUE #1550: [Improvement] Let users turn off Deletion of OAuth2 Session on restart
In this issue: https://github.com/getinsomnia/insomnia/issues/168 the "feature" that oauth2 cookies are automatically deleted on restart was added. I really don't like that since it is super annoying to have to type in the Credentials every time again. I don't mind the popup opening again but usually (for example with Postman or when logging in via browser) at least the linked windows account is still stored and I don't need to type in the password again. It would be wonderful if I could keep my oauth2 cookies in insomnia too.

Or maybe there could be a button that clears oauth2 cookies so a restart is not needed for people that want to login with a different account?

In any case it would ease my workflow a lot if oauth2 cookies are not always cleared on restart.

KarolineKarkosch:
Offer button to user where it can be cleared without restart

Closes #1550


packages/insomnia-app/app/network/o-auth-2/misc.js
```diff
@@ -2,7 +2,11 @@ import electron from 'electron';
 import * as uuid from 'uuid';
 import querystring from 'querystring';
 
-const AUTH_WINDOW_SESSION_ID = uuid.v4();
+let authWindowSessionId = `persist:oauth2`;
+
+export function clearOAuthSession() {
+  authWindowSessionId = `oauth2_${uuid.v4()}`;
+}
```
develohpanda: There is some interesting behavior here, and I'm not sure how I feel about it _just_ yet. 

Here's what I think is going to happen.
- Authenticate with Account A
- Restart
- **Account A persists** ✔️ 
- Clear token
- Authenticate with Account B
- Restart
- **Account A persists**, or no auth persists ❌ 

I would expect that if I clear the token and re-authenticate with account B, when I restart, my session persists with Account B.

Is my understanding of the current behaviour correct?

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -477,6 +483,11 @@ class General extends React.PureComponent<Props, State> {
           { placeholder: '~/.insomnia:/other/path' },
         )}
 
+        <hr className="pad-top" />
+        <h2>Clear OAuth2 Session</h2>
+        <button className="btn btn--clicky" onClick={clearSession}>
+          Clear session
+        </button>
```
develohpanda: In addition to this, it might be good to have two options:
- A checkbox to `Clear session on restart` (defaults to true because that is current behavior)
- A button to explicitly clear the session (keeping in mind my previous comment, depending on the state of the first checkbox)

What do you think?

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -185,6 +186,11 @@ class General extends React.PureComponent<Props, State> {
   render() {
     const { settings } = this.props;
     const { fonts, fontsMono } = this.state;
+
+    function clearSession() {
+      clearOAuthSession();
+    }
```
develohpanda: Could you please move this function out of the `render()` method, and create a method alongside `render()`? In the current implementation it will create a new function on every render, which can be inefficient. 😄

EDIT: what was I thinking, you don't even need to create a new function. The button can simply recieve `clearOAuthSession` in the `onClick` prop.

develohpanda (commented):
Nice! I generally like this, I think it would be beneficial for folks who have different workflows.

I have a couple of observations from the implementation though, could you please check the behavior? 😄

packages/insomnia-app/app/network/o-auth-2/misc.js
KarolineKarkosch: @develohpanda  you are right, that would be the current behaviour. What would you consider the best way to store the currentSessionId so that the last set session can be used again after restart? Could I store that in localstorage?

packages/insomnia-app/app/ui/components/settings/general.js
KarolineKarkosch: I personally do not see a use case for the checkbox: Either I want to always keep my oauth-session then I do nothing or I sometimes want to switch the session in which case using the button would probably be easier than a restart? I wouldn't mind adding it though :)

packages/insomnia-app/app/network/o-auth-2/misc.js
KarolineKarkosch: I updated the logic so that the current sessionId is persisted in localstorage and used again on reboot. I think that should solve the problem.

packages/insomnia-app/app/ui/components/settings/general.js
KarolineKarkosch: fixed

packages/insomnia-app/app/ui/components/settings/general.js
KarolineKarkosch: If you think this feature is needed, could you tell me how I could get to the settings in the misc.js of oauth2 when I use renderBooleanSetting to display the checkbox? I am not too familiar with react, it seems like the other services inject the settings but I don't think I can use that in the misc.js easily, can I?

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -477,6 +478,11 @@ class General extends React.PureComponent<Props, State> {
           { placeholder: '~/.insomnia:/other/path' },
         )}
 
+        <hr className="pad-top" />
+        <h2>Clear OAuth2 Session</h2>
+        <button className="btn btn--clicky" onClick={clearOAuthSession}>
+          Clear session
+        </button>
```
develohpanda: A few notes here:
- Please update the header to be "Security"
- There will be one row containing the checkbox on the left "Clear OAuth 2 session on start", and this "Clear OAuth 2 session" button on the right. Have a look at the two column layout that the Request/Response settings uses

Once you do, could you also please add a screenshot for what it looks like in settings? Thank you!

packages/insomnia-app/app/network/o-auth-2/misc.js
```diff
@@ -2,7 +2,19 @@ import electron from 'electron';
 import * as uuid from 'uuid';
 import querystring from 'querystring';
 
-const AUTH_WINDOW_SESSION_ID = uuid.v4();
+const LOCALSTORAGE_KEY_SESSION_ID = 'insomnia::current-oauth-session-id';
```
develohpanda, line 5: 👍 

Adding to local storage seems a good approach

packages/insomnia-app/app/ui/components/settings/general.js
develohpanda: Need to think about this one. There are three options, and I can think of use cases for each:
1. always reset on restart (current behavior)
2. never reset on restart
3. manually reset

Because all three can be valid, and because we currently reset, I'd like to potentially preserve that behavior and make it opt-in to change the current default behavior, while always providing an option to manually reset as well if the user workflow requires it. 

@nijikokun would you have any thoughts here? Worth getting some context from https://github.com/Kong/insomnia/issues/168, https://github.com/Kong/insomnia/pull/258 and https://github.com/Kong/insomnia/issues/1550, regarding past decisions.

(Most) Settings are stored in the database, so you won't be able to access it directly in `misc.js`, but it should be possible to invoke the `clearOAuthSession` function you have added after the database has been initialized and we have user settings. I don't anticipate it to be too challenging, should we need to add it.

develohpanda (commented):
Progress! 😄

packages/insomnia-app/app/ui/components/settings/general.js
nijikokun: I think the thoughts captured here:

> A checkbox to `Clear session on restart` (defaults to true because that is current behavior)

Retaining current behavioral support for users who rely on it today, _is a good thing_ and we should definitely ensure it is the default behavior as it is the current behavior, _and with this_ checkbox and button _we provide an alternative for those who do not want this behavior_. :+1:

develohpanda:
Given we want to make it opt-in for people to disable the clearing of the OAuth session on startup, here's what you'll need:
1. a new property in the settings (use "Follow Redirects" as an example for linking the checkbox to a [database property](https://github.com/Kong/insomnia/blob/03d83a8f52f9d9f4d4f38e014d406a95ef1d2611/packages/insomnia-app/app/models/settings.js#L86)
2. conditionally clear the token on launch

For the second item: looking at `main.development.js` (app entry point), there is logic to initialize the database, and then initialize the app in `_launchApp()`.

https://github.com/Kong/insomnia/blob/03d83a8f52f9d9f4d4f38e014d406a95ef1d2611/packages/insomnia-app/app/main.development.js#L36-L50

Before/during/after the `_launchApp()` invocation, you should be able to load the setting from the database and clear if necessary.

```js
const settings = await models.settings.getOrCreate();

if (settings.clearOAuthSessionOnStart) {
	clearOAuthSession();
}
```

Please reach out if you need any help! Thanks for working through this 🤗

KarolineKarkosch:
> Given we want to make it opt-in for people to disable the clearing of the OAuth session on startup, here's what you'll need:
> 
>     1. a new property in the settings (use "Follow Redirects" as an example for linking the checkbox to a [database property](https://github.com/Kong/insomnia/blob/03d83a8f52f9d9f4d4f38e014d406a95ef1d2611/packages/insomnia-app/app/models/settings.js#L86)
> 
>     2. conditionally clear the token on launch
> 
> 
> For the second item: looking at `main.development.js` (app entry point), there is logic to initialize the database, and then initialize the app in `_launchApp()`.
> 
> https://github.com/Kong/insomnia/blob/03d83a8f52f9d9f4d4f38e014d406a95ef1d2611/packages/insomnia-app/app/main.development.js#L36-L50
> 
> Before/during/after the `_launchApp()` invocation, you should be able to load the setting from the database and clear if necessary.
> 
> ```js
> const settings = await models.settings.getOrCreate();
> 
> if (settings.clearOAuthSessionOnStart) {
> 	clearOAuthSession();
> }
> ```
> 
> Please reach out if you need any help! Thanks for working through this hugs


Thanks for your advice. I tried to implement it as you described but it seems to me as if the task in main.development runs in a different thread than where I am trying to use oauth2/misc.js. When I set the currentSessionId in _launchApp it is still undefined when the OAuth2 window is opened. Also console.logs inside misc.js go into the terminal console when calling functions from launchApp and to the browser console when I later try to access the value which furthers my suspicion that they are running in different environments. Additionally window is not yet defined at that point, so localstorage is not usable at all.
So I moved the initialization from settings to ui/index.js where I saw that for example themes or fonts are initialized from settings as well. I hope that is ok too. 
It now works as expected: Depending on the clearOnRestart-Setting the session is deleted or not after starting and when using the sessionClear button the last session is persisted for after restart.

Also I fixed the styling per your suggestion, this is what it currently looks like:

Screenshot from 2020-10-08 21-03-08

packages/insomnia-app/app/ui/components/settings/general.js
KarolineKarkosch: Header was changed, styling was fixed and screenshot added

packages/insomnia-app/app/ui/components/settings/general.js
KarolineKarkosch: Setting was added with default true to keep the current behaviour as described

develohpanda:
> Also console.logs inside misc.js go into the terminal console when calling functions from launchApp and to the browser console

Nice sleuthing, that seems to make sense. When I was looking through that code, I noticed that `windowUtils.init()` was also initializing local storage, so thought it should be available at that point. I think `main.development.js` is the electron entry point and `ui/index.js` is the entry point for the web application being run in electron, so that might explain the differences.

It looks like you've put it in the right place now! 😃

packages/insomnia-app/app/network/o-auth-2/misc.js
```diff
@@ -2,7 +2,20 @@ import electron from 'electron';
...
+let authWindowSessionId;
+if (window.localStorage.getItem(LOCALSTORAGE_KEY_SESSION_ID)) {
+  authWindowSessionId = window.localStorage.getItem(LOCALSTORAGE_KEY_SESSION_ID);
+} else {
+  initNewOAuthSession();
+}
+
+export function initNewOAuthSession() {
+  // the value of this variable needs to start with 'persist:'
+  // otherwise sessions won't be persisted over application-restarts
```
develohpanda, line 15: Nice comment, some additional context [here](https://www.electronjs.org/docs/api/session#methods)

develohpanda (approved):
Code looks good to me, thank you for working through this!

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -390,6 +391,23 @@ class General extends React.PureComponent<Props, State> {
 
         <hr className="pad-top" />
 
+        <h2>Security</h2>
+        <div className="form-row pad-top-sm">
+          {this.renderBooleanSetting(
+            'Clear OAuth 2 session on start',
+            'clearOAuth2SessionOnRestart',
+            'Clears the session of the OAuth2 popup window every time insomnia is launched',
```
DMarby: Minor nitpick, but Insomnia should probably be capitalized
```suggestion
            'Clears the session of the OAuth2 popup window every time Insomnia is launched',
```

DMarby (approved):
Very nice, thanks for your contribution! 👍

sonicyeti (approved):

PR #2704, opened 2020-10-08
Prevent newlines for singleline Code editor

### LINKED ISSUE #2569: Request broken if line breaks inside request auth options fields
**Describe the bug**
The input fields in **Auth** tab allow new lines to be entered if a user types `ENTER` before inputting any text. This new lines are sent in the header of the request, breaking the request.

I inadvertently typed a `ENTER` in the `prefix` field of the Bearer auth option, while the cursor was there, and took a while to figure that there as a line break in the field after the requests stopped working, because the line break was the only character in the field.

On the other hand, the field does **not** accept a `ENTER` after some text has been entered.

This also applies to other auth modes, such as basic/digest/etc.

**To Reproduce**
1. Create a new request
2. Click on 'Auth' and select any mode (e.g. Bearer)
3. Put the cursor in any field (such as token or prefix) and press `ENTER` BEFORE entering any text
4. Check that `ENTER` is not allowed after entering some text
5. Execute the request to see line breaks sent in the header

**Expected behavior**
Line breaks should not be allowed in auth options fields.

 - OS: Windows 10
 - Installation Method: exe
 - App Version: 2020.3.3

thewheat:
Aims to fix https://github.com/Kong/insomnia/issues/2569 by detecting the enter key being pressed and returning early so that it doesn't register as a key press

**Behaviour before the change**
before

**Behaviour after the change**
after

Note: There were some other changes when I saved the file/ran the app that I did not include. Am unsure if that was done based on the editor configurations or if I haven't fully configured `.editorconfig` correctly 😅


develohpanda (dismissed):
Seems good to me 👍

What happens when you paste something with a linebreak? Can look at that in a separate PR though.

> Note: There were some other changes when I saved the file/ran the app that I did not include. Am unsure if that was done based on the editor configurations or if I haven't fully configured .editorconfig correctly 😅

What were the additional files? With the change you have made I don't anticipate any other changes, but it's possible there were `package-lock.json` changes (which are okay to leave out).

thewheat:
> What were the additional files? 

The changes were in the same file I was editing. Digging a bit more I think I got some mix up with the old `master` branch 😅 Using the lastest code it seems that my change doesn't work for the current `develop` branch. I'll go back to the drawing board with this 🙈 Sorry!

develohpanda:
No worries! I'll mark this PR as draft, feel free to mark as ready, once you're ready for review again. Take your time, no rush. 😄

sonicyeti (dismissed):
Nice UX enhancement here!

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -795,18 +795,20 @@ class CodeEditor extends React.Component {
...
+    // If we're in single-line mode, merge all changed lines into one
+    if (this.props.singleLine && change.text && change.text.length > 1) {
+      const text = change.text
+        .join('') // join all changed lines into one
+        .replace(/\n/g, ' '); // Convert all whitespace to spaces
+      change.update(change.from, change.to, [text]);
+    }
+
     // Suppress lint on empty doc or single space exists (default value)
     if (value.trim() === '') {
```
thewheat, line 808: A newline value when trimmed will be an empty string so old code to detect multilines is skipped. So just moved code up to allow processing so no net new code 💪
image

thewheat:
@develohpanda Managed to clean up my branch and updated code with a version that works with latest `develop` branch 🎉
It doesn't add any code but actually just moves existing code around

DMarby (approved):

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -795,18 +795,20 @@ class CodeEditor extends React.Component {
...
-      // If we're in single-line mode, merge all changed lines into one
-      if (this.props.singleLine && change.text && change.text.length > 1) {
-        const text = change.text
-          .join('') // join all changed lines into one
-          .replace(/\n/g, ' '); // Convert all whitespace to spaces
-        change.update(change.from, change.to, [text]);
-      }
 
       // Don't allow non-breaking spaces because they break the GraphQL syntax
       if (doc.options.mode === 'graphql' && change.text && change.text.length > 1) {
```
develohpanda, line 814: I think this condition also needs to be moved out, for the same reason? The `change.text.length > 1` condition is the same, and looking through the logic the graphql editor seems like it might exhibit the same buggy behavior when first entering text.

It doesn't relate directly to the issue this PR fixes so I would suggest doing it in a separate PR if it does need to be moved.

develohpanda (approved):
LGTM, thanks for working through this!

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
thewheat: Hmmm I'm not familiar with the graphql side of things so it wasn't something I tested. But definitely happy to test it and submit a PR if you know how and when that `doc.options.mode == 'graphql'` would get triggered 😃

develohpanda: Here's an example workspace with a request containing the special character after the "title" text (see screenshot below).

[Run in Insomnia}](https://insomnia.rest/run/?label=GQL%20with%20NBSP&uri=https%3A%2F%2Fgist.githubusercontent.com%2Fdevelohpanda%2F1da454049c24cb9111571571f8692161%2Fraw%2F4fa05e91f9c5c3953d945842836b4f6aa63fa9c8%2Fexport.yaml)

_I think_ the intended behavior of this code is to replace the special space with a regular space. If you change the text in the request body, I'm guessing it _should_ clear the error by replacing the character 🤔 

image

Also, on Windows you can add that special character with `Alt 255` (hold alt, and type 255 on the numpad)

thewheat: Managed to replicate the problem and PR to fix it is here https://github.com/Kong/insomnia/pull/2770 💪

PR #2708, opened 2020-10-10
Fix tag editor preview issue due to blackslash

### LINKED ISSUE #757: [Bug] Can't load file for file token
- Insomnia Version: 5.14.3
- Operating System: win10

## Details

1. Create request
2. Add a file token and click it
3. Choose file click Done
4. File token is in warning state

Quick look at error message says:
```ENOENT: no such file or directory, open 'C:\Users\Owner\AppData\Local\insomnia\app-5.14.6\UsersOwnerDesktopimage.png'```

So looks like it's resolving file path wrongly!

Thanks 😄

### LINKED ISSUE #2563: File paths not parsed on Windows
**Describe the bug**
Seems like slashes are not parsed correctly in the "File" JSON field type.

**To Reproduce**
1. Create a request with JSON body
2. Press Ctrl+Space
3. Choose "File"
4. Choose a file
5. See the error.

**Expected behavior**
File being parsed.

**Screenshots**
image

 - OS: Windows 10
 - Installation Method: exe
 - App Version 2020.3.3

**Additional context**
None.

thewheat:
**PR details**
- Currently in the Tag Editor modal we can have incorrect "Live Preview" data and this PR aims to fix it
image
- Closes #757, Closes #2563 (dupe of #757)
- Feels that there is some complexity to this so if I went down the wrong path let me know and I'll see how I can modify this 👍

### Before
file_problem

### After
file_fixed

### Before
base_problem

### After
base_fixed

* * * 

**Context**
- When a string with a backslash is used (e.g. in a file name or in input value of a function)
- It will work correctly on first use
image
- But closing the modal and reopening it will show that the "Live Preview" with no data / incorrect data (illustrated in the first image in this PR)

**Other details**
- It looks like there is escaping of strings when an input value is changed https://github.com/Kong/insomnia/blob/03d83a8f52f9d9f4d4f38e014d406a95ef1d2611/packages/insomnia-app/app/ui/components/templating/tag-editor.js#L142-L145
- This requires `activeTagData.args` values which are strings needing to be escaped on first load (which is the core part of this PR)
- While this PR fixes the current problem I feel that there could be something related to the escaping of strings which could be the root issue (I couldn't figure out the full context of why it needed to be escaped)


dimitropoulos (approved):
Works!

(before on left, after on right)

Screenshot_20210909_091637
Screenshot_20210909_091859

PR #2711, opened 2020-10-11
Fix color picker display when updating an environment color

### LINKED ISSUE #811: [BUG] Changing environment colour affects multiple environments
- Insomnia Version:  5.14.9 (5.14.9.1895)
- Operating System: macOS 10.13.3

Changing the colour of an environment seems to affect multiple environments. As you can see in the gif changing the colour of the `production` environment also affects the `testing` environment (it also nukes the content).

https://imgur.com/a/isEWC

Edit: It appears that environments get affected across workspaces as well.

### LINKED ISSUE #2647: Environment color selection on MacOS
**Describe the bug**
This bug is two-fold:

1. When trying to set an environment color on an environment with no color set, the only option for picking a color is RGB. This was not the case on earlier releases, where you could use the "pencil picker" (sorry, I don't know what this is actually called!!) or any of the usual MacOS options to pick.
2. For environments where you want to change the color, you have to unset, then you're back to the previous issue. Trying to change without brings the menu up for a second, then it disappears.

I am able to work around this for now by reverting back to 2020.3.0

**To Reproduce**

1. Go to manage environments
2. Pick an environment with a color
3. Go to change color on the environment
4. Note no color panel comes up

For the second issue:
Perform steps 1-3
- Note that the color pane that comes up is not the one that is the normal / native MacOS panel for color things; it's a whole other thing that only allows for RGB selection.

**Expected behavior**
I expect to see the color options as captured under screenshots!!

**Screenshots**
This is missing:
image

OS: Darwin x64 19.6.0
Electron: 9.1.1
Node: 12.14.1
Installed either from brew or downloading direct from the website

jgiovaresco:
We were removing the color picker to reset event handlers easily (fix #811).
It is possible to remove the listener quite easily by listening to the `change` event triggered when the color picker modal is closed.

There was still an issue with the 1st click on `Change color`. We were adding the `input` dynamically on the 1st click and we were having some race condition between the `click` and the `appendChild` 🤔 
To fix that, I've added the `input` element just before the `Color` dropdown. Besides, now we can open the color picker modal near the dropdown by making the `input` element invisible.

Before
image

After
color-picker

Closes #2647

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
@@ -101,6 +101,7 @@ const SidebarList = SortableContainer(
 @autobind
 class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
   environmentEditorRef: ?EnvironmentEditor;
+  environmentColorInputRef: HTMLInputElement;
   colorChangeTimeout: any;
   saveTimeout: any;
   modal: Modal;
@@ -126,6 +127,10 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
 
   _setEditorRef(n: ?EnvironmentEditor) {
     this.environmentEditorRef = n;
+  }
+
+  _setInputColorRef(ref: HTMLInputElement | null) {
+    this.environmentColorInputRef = ref;
   }
 
   _setModalRef(n: Modal | null) {
@@ -329,33 +334,19 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
   }
 
   async _handleClickColorChange(environment: Environment) {
-    let el = document.querySelector('#env-color-picker');
-
-    // Remove existing child so we reset the event handlers. This
-    // was easier than trying to clean them up later.
-    if (el && el.parentNode) {
-      el.parentNode.removeChild(el);
-    }
-
-    el = document.createElement('input');
-    el.id = 'env-color-picker';
-    el.type = 'color';
-    document.body && document.body.appendChild(el);
-
     const color = environment.color || '#7d69cb';
-
     if (!environment.color) {
       await this._handleChangeEnvironmentColor(environment, color);
     }
 
-    el.setAttribute('value', color);
-    el.addEventListener('input', (e: Event) => {
-      if (e.target instanceof HTMLInputElement) {
-        this._handleChangeEnvironmentColor(environment, e.target && e.target.value);
-      }
-    });
+    this.environmentColorInputRef?.click();
+  }
 
-    el.click();
+  _handleInputColorChange(event: SyntheticEvent<HTMLInputElement>) {
+    this._handleChangeEnvironmentColor(
+      this._getActiveEnvironment(),
+      event.target && event.target.value,
+    );
   }
 
   _saveChanges() {
@@ -473,6 +464,14 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
 
               {activeEnvironment && rootEnvironment !== activeEnvironment ? (
                 <React.Fragment>
+                  <input
+                    className="hidden"
+                    type="color"
+                    value={activeEnvironment.color}
+                    ref={this._setInputColorRef}
+                    onInput={this._handleInputColorChange}
+                  />
+
                   <Dropdown className="space-right" right>
                     <DropdownButton className="btn btn--clicky">
                       {activeEnvironment.color && (
--- a/packages/insomnia-app/app/ui/css/layout/base.less
+++ b/packages/insomnia-app/app/ui/css/layout/base.less
@@ -501,6 +501,10 @@ blockquote {
   display: none;
 }
 
+.hidden {
+  visibility: hidden;
+}
+
 .relative {
   position: relative;
 }


packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
```diff
@@ -344,9 +344,7 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
 
     const color = environment.color || '#7d69cb';
 
-    if (!environment.color) {
-      await this._handleChangeEnvironmentColor(environment, color);
-    }
```
develohpanda: How was this previously working at all? This logic says that once the environment color is set, it can't be set again until it's cleared, right?

develohpanda (dismissed):
Seems alright but I'd like to know how it was working before the Electron update or if the logic here changed recently.

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
develohpanda: Ah, this line is only for defaulting the color if one doesn't exist - there is an event handler further down that handles the actual input change.

develohpanda (commented):
> I've noticed that for any reason calling `await this._load(this.state.workspace)` just after appending the input element to the body is fixing the display of the color picker.

I'm guessing that adding an await to _something_ after appending the input element, is allowing some background process to complete.

I doesn't feel like forcibly updating the environment color each time we prompt for a new color is fixing the root cause.

I was able to get the same fix working by delaying the element click using `setTimeout(() => el.click(), 50);`, so needs further investigation to determine and fix the root cause.

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
```diff
@@ -329,30 +329,22 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
...
       }
+    };
+
+    el.value = color;
+    el.addEventListener('input', changeColor);
+    el.addEventListener('change', (e: Event) => {
+      el.removeEventListener('input', changeColor);
     });
 
     el.click();
```
develohpanda: I hope I'm not missing something, but we shouldn't need to DOM manipulation this way, if the `input` tag is being rendered directly as part of this component. It should be achievable just in React itself, right?

jgiovaresco: Good catch... I didn't think about that... 🤦 

Using React helps because we don't need to clean the listener to keep #811 fixed.

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
```diff
@@ -329,35 +329,22 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
...
 
-    el.setAttribute('value', color);
-    el.addEventListener('input', (e: Event) => {
-      if (e.target instanceof HTMLInputElement) {
-        this._handleChangeEnvironmentColor(environment, e.target && e.target.value);
-      }
-    });
-
+    const el = document.querySelector('#env-color-picker');
+    el.value = color;
```
develohpanda: Is there anything stopping you from setting the value on the input itself, using `activeEnvironment.color`?

You can also refer to the input using `ref` in order to programmatically click it.

There should be no need for `document.querySelector` with the changes you've made, just need to take it one step further.

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
```diff
@@ -329,35 +329,22 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
...
     el.click();
   }
 
+  _handleInputColorChange(environment: Environment, event: Event) {
+    if (event.target instanceof HTMLInputElement) {
+      this._handleChangeEnvironmentColor(environment, event.target && event.target.value);
+    }
+  }
```
develohpanda: This event handler can be simplified...

```suggestion
  _handleInputColorChange(e: SyntheticEvent<HTMLInputElement>) {
      this._handleChangeEnvironmentColor(this._getActiveEnvironment(), event.target && event.target.value);
  }
```

Then on the input itself, consumed as:
```js
<input
  ...
  type="color"
  onInput={this._handleInputColorChange}
/>
```

Is `onInput` still the best event to use?

packages/insomnia-app/app/ui/css/layout/base.less
```diff
@@ -501,6 +501,10 @@ blockquote {
   display: none;
 }
 
+.hidden-item {
```
develohpanda: Can this just be `hidden`, or does that clash with something else?

develohpanda (changes requested):
Thanks for updating the description and adding in a gif. Almost there!

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
jgiovaresco: > input is fired on the <input> element every time the color changes. The change event is fired when the user dismisses the color picker. In both cases, you can determine the new value of the element by looking at its value.
[source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color)

I've tried to use `onChange` and we keep the same behavior 🤔 
2647-onChange-event

Therefore we can change the event 🤷

packages/insomnia-app/app/ui/css/layout/base.less
jgiovaresco: My IDE has found a `hidden` class. But after a closer look, this class belongs to `swagger-ui-react` dependency. Therefore I should be able to rename it ;)

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.js
```diff
@@ -329,33 +334,19 @@ class WorkspaceEnvironmentsEditModal extends React.PureComponent<Props, State> {
...
       await this._handleChangeEnvironmentColor(environment, color);
     }
 
-    el.setAttribute('value', color);
-    el.addEventListener('input', (e: Event) => {
-      if (e.target instanceof HTMLInputElement) {
-        this._handleChangeEnvironmentColor(environment, e.target && e.target.value);
-      }
-    });
+    this.environmentColorInputRef.click();
```
develohpanda: Last one! It's usually good to check that the ref is not falsey before using it

```suggestion
    this.environmentColorInputRef && this.environmentColorInputRef.click();
```

I don't know if optional chaining is currently enabled, but you could also do something like

```suggestion
    this.environmentColorInputRef?.click();
```

develohpanda (approved):
Nicely done 👏

DMarby (approved):

PR #274, opened 2017-05-31
Add Bearer Authorization Support

mai7star:
I wrote a patch to add Bearer Authorization Scheme (RFC6750) Support.

PR #2712, opened 2020-10-11
Add resizing for Design Preview and Testing Results panes

### LINKED ISSUE #2583: Design Preview and Testing Results pane should be resizable
**Describe the bug**
In the design and test tabs, the right-most pane (preview and test results) are not resizable. In the debug tab, we can see that the three panels can be resized independently.

**Expected behavior**
All three panels on each tab should be resizable independently.

**Screenshots**

```
Version: Insomnia Designer 2020.4.0-beta.6
Release date: 02/09/2020
OS: Darwin x64 19.6.0
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
```

INS-54

Related: https://github.com/Kong/insomnia/issues/2584

simplywondrous:
## Context
This PR closes #2583 

## Changes
- Introduced `renderPaneOne` and `renderPaneTwo` in `PageLayout`
  - This rendered the wrapper grid div that wrapped the middle and right panes unnecessary, so that was removed
  - Consolidated CSS styling - removing styles now unused after these changes
  - For cases where panes aren't desired, `renderPageBody` is still available (such as for `WrapperHome`)
- Added ability to resize middle / right panes to Design and Test tabs

## GIF
Resize


simplywondrous:
There are three ways I'm thinking of for cleaning this up -
1. Forwarding the refs with the wrapper (that wraps the two panes) with `forwardRef` instead of creating classes. Doesn't seem like `forwardRef` is used anywhere, and React has a [note to component library maintainers](https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers) about introducing it. This is an app and not a component library I guess, but asking to be sure!

2. Keep the grid wrapper, slap the refs directly onto the dom elements themselves, and modify the resize logic to be
```
      const requestPane =
        this._requestPane instanceof HTMLElement
          ? this._requestPane
          : ReactDOM.findDOMNode(this._requestPane);
      const responsePane =
        this._responsePane instanceof HTMLElement
          ? this._responsePane
          : ReactDOM.findDOMNode(this._responsePane);
```
so that we get correct `offsetWidth` values from request and response panes, regardless of if the ref is an element or component
(I am currently stuck with this one. The ref element is correct, but the offsetWidth value isn't updated as I drag - if you have any ideas for what might be the issue?)

3. Change the wrapper to a flexbox, which handles resizing easier than a grid layout


For question / option #2 - here's what `renderPageBody` looks like:

```js
renderPageBody() {
  ...
  return (
    <div className="unit-tests layout-body--sidebar">
      <section ref={handleSetRequestPaneRef} className="unit-tests__tests">...</section>
      <div className="drag drag--pane-test">...</div>
      <section className="unit-tests__results results-pane" ref={handleSetResponsePaneRef}>
         {this.renderResults()}
      </section>
    </div>
  );
}
```

develohpanda:
All in all, a nice approach you've followed 👏 Code looks good!

Note: if none of the following options are viable in the short term, the current changes (and those anticipated for `WrapperDesign`) are still great! I have some really minor notes but we can address those once the PR is ready with the desired direction.

**To answer your questions**

> Forwarding the refs with the wrapper (that wraps the two panes) with `forwardRef` instead of creating classes.

I'm not too concerned about the warnings for library maintainers with this particular case as _we_ are the only consumers of the components. How do you envision `forwardRef` being set up? To my understanding, `forwardRef` works with a single ref, but we have multiple to pass around (sidebar, request, response panes).

Alternatively the new class components can also be simple(r) function components. New components might be unnecessary altogether too, if you get refs on dom elements working.

> slap the refs directly onto the dom elements themselves

I'm not sure why this wouldn't be working... could it be that you need to use `this._requestPane.current` when accessing a ref in the absence of `findDOMNode`?

If you can't get it working for this PR, it can be scoped to a standalone PR given it might need some additional work, but it would be good to remove `findDOMNode` (unless absolutely necessary) as it is now discouraged. Looks like the pane dragging logic was written in 2016, so is probably well outdated with current React practices. 🤣 

> Change the wrapper to a flexbox, which handles resizing easier than a grid layout

While this would be pretty awesome, my gut feel is that it might be a lot of work. CSS Grid is quite central to the layout and it's variations. Flexbox would definitely be a step forwards but might be too much to tackle in this PR?
 
**Some additional observations**

I was looking at how the drag pane was set up for the sidebar, because it seems to be working across all views (design, debug, test) in the current state. It looks like `PageLayout` has `renderPageSidebar` and `renderPageBody` as props, and sets up the sidebar dragging logic in there.

I wonder if it's worth investigating whether `renderPageBody` can be updated to `renderPaneOne` and `renderPaneTwo`, and any className props, and the dragging logic for the entire layout all be set up in one common location.

This doesn't really _change_ the approach, just prevents the common dragging logic from being scattered and duplicated throughout. It might be slightly more work than just duplicating but feels a little more maintainable. What do you think?

simplywondrous:
It's only a guess, but after thinking about it some more, maybe if the ref is a dom element since the element itself isn't changing on drag no update is triggered, whereas when the ref is on a class component (that then finds a dom element) when the pane gets resized something is updated and the changes propagate down to the component, which then updates?

Looking it up, seems the best way to forward multiple refs is to make the ref an object with refs. That being said, if we had `PaneOne` and `PaneTwo` then we wouldn't need to forward multiple, so that seems to be the way to go. Class components can be assigned refs because an instance exists whereas function components require `forwardRef`, so I'll just make them class components.

It'd make the most sense if the dragging logic were all in one place like `PageLayout`. I haven't looked at how the functions in `app.js` (including the drag logic) make it into `PageLayout`, but I think for this PR I'll try to separate `renderPageBody` into `renderPaneOne` and `renderPaneTwo` and consolidate the styling, along with adding resizing between the two panes. Then removing `findDOMNode` and seeing if the drag logic can be moved to PageLayout can be follow-up issues.

develohpanda:
My understanding of `ReactDOM.findDOMNode` is that if invoked with a ref attached to a _component_, it returns the first dom element rendered by the component, and so should be the same as just putting a ref on the dom element itself (as recommended). Neither the element nor the component is updated (as far as I can tell) because `paneWidth` and `paneHeight` are consumed higher up the tree in `PageLayout`. Needs some more research but I wonder if `findDOMNode` is loading the _latest_ dom before finding the node, which would have been updated with the row/column sizing applied in [`PageLayout`](https://github.com/Kong/insomnia/blob/81ad8ba2c89196ec2fefe088a9d2e860279d12a0/packages/insomnia-app/app/ui/components/page-layout.js#L60-L63). What a pickle!

> for this PR I'll try to separate `renderPageBody` into `renderPaneOne` and `renderPaneTwo` and consolidate the styling, along with adding resizing between the two panes.

This sounds good 👍 I can't think of any (great) reasons _not_ to split `renderPageBody` - in the case where only one of the panes is used (preview hidden in the design tab), one of the render props can just be optional, and drag handlers can be applied as required within `PageLayout`.

> Then removing `findDOMNode` and seeing if the drag logic can be moved to PageLayout can be follow-up issues.

I agree, it seems non-trivial, but also non-critical so it can be tackled as a standalone PR. Another thing to be tackled separately is persisting the unique sidebar/pane widths per workspace _and_ per tab. Previously we only had one tab (Debug) so it wasn't necessary.

simplywondrous:
Ohhh I see - so if `findDOMNode` is loading the latest dom and then finding the node, it'd get the updated offset, whereas when we're simply looking at the current ref element's offset, that doesn't get updated so stays the same.

Right now I'm working on hiding `PaneTwo` when hiding the preview in the design tab, and have a question.
Currently, `render` in `PageLayout` includes (with some code omitted)
```js
return {renderPaneTwo && <div> ... renderPaneTwo()</div>}
```
and in `WrapperDesign`, `renderPaneTwo={this.renderPreview}` is passed to `PageLayout`, with
```js
renderPreview = () => { return !previewHidden && <div /> }
``` 

Since we're passing it a function, `renderPaneTwo` is always `true` even when `renderPaneTwo()` returns `false`. To fix this, my first thought is to call `renderPreview` in `WrapperDesign` and pass the returned value to `renderPaneTwo`:

```js
const preview = renderPreview()
<PageLayout renderPaneTwo={preview} />
```

This would prevent dynamically generating the content of `PaneTwo` in `PageLayout`, though. Then `preview` from the example above would be statically passed in. I'm _pretty sure_ this would cause edge case issues of some sort, but am not quite skilled enough to give examples, so could be I have concepts mixed up :pensive:. Maybe if `renderPreview` called a handler that relied on variables set elsewhere which later changed?

If there are no issues, would it be a good idea to change just `renderPaneTwo` to take a JSX element instead of a function, while keeping the rest, such as `renderPaneOne`, the same?

Second thought is, in `PageLayout` I could do
```js
const preview = renderPaneTwo()
return {preview && <div />}
```
and I think then the content would still be dynamically generated upon component render. Would this be okay or is there a better way to go about hiding the second pane?

**Update:** Even after implementing the last suggestion, I'm still running into an issue where PaneOne's grid space is calculated by the parent `Wrapper` and updated when PaneTwo is dragged, but not when PaneTwo doesn't render, similar to what happened when I didn't use `findDOMNode`. I'll have to think on it more, but to get hiding the preview (in `WrapperDesign`) to trigger a style change in its parent component (`PageLayout`) would require the parent passing down a show/hide handler of some sort to PaneTwo to then call, right?

develohpanda:
There are a few different options I can think of, and no one is better than the other at first sight. The options you've given are both good too. 

I'll pull this code later today and see what's up, AFK for the next few hours! 😊

simplywondrous:
I very much appreciate your responsiveness, but it says your status is "on vacation" so please enjoy it! This is certainly not something that can't wait a few days!

develohpanda:
> 
> 
> I very much appreciate your responsiveness, but it says your status is "on vacation" so please enjoy it! This is certainly not something that can't wait a few days

I picked a good week to take off, middle of the busiest open-source month 😅 just making sure the wheels keep turning, but I'll dip away for the last few days; thanks for noting it!

I noticed `WrapperHome` is yet another variation of the grid which will need to be considered here. This is the documents home page when you hit the "Documents" link on the top left. Note: with this PR in the current state, the homepage doesn't show anything, because it's still using the old `renderPageBody` prop.

> If there are no issues, would it be a good idea to change just `renderPaneTwo` to take a JSX element instead of a function, while keeping the rest, such as `renderPaneOne`, the same?

Because the render functions don't actually have any arguments, I don't think there's a specific case for them to be functions over just `React.Element` or `React.Node`. Might be worth doing some testing here to validate this, but if you change them it _should_ behave like for like. I'd suggest `paneOne` and `paneTwo` be `React.Node` or `React.Element` (since they are directly impacted by this PR), but the other ones remain as functions as-is for now.

FYI the `Header` component has `React.Node` props (used in all Wrapper components as well): https://github.com/Kong/insomnia/blob/4fc8b21827b7b0222b67ef0f2756458d7007a6b0/packages/insomnia-components/components/header.js#L5-L10

> my first thought is to call `renderPreview` in `WrapperDesign` and pass the returned value to `renderPaneTwo`

Lastly, as far as I know, this is alright. It's a fairly common pattern. React components are, after-all, standard JS objects, and they can be passed around as props just fine.

<details>
<summary>One gotcha</summary>

```js
const preview = renderPreview()
<PageLayout renderPaneTwo={preview} />
```

Assuming the above snippet runs in each render of `WrapperDesign`, depending on the implementation `preview` would be a brand new JS object, and it will cause `PageLayout` to update because props will be shown as changed. Is this the concept you are thinking about, regarding the edge case?

It does but also doesn't matter (here) because there are other props using inline functions `() => {}` as props, which causes the same behavior (brand new function each render), causing the `PageLayout` to update.

</details>

If you get rid of the render functions, does the drag resizing work as expected?

packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -214,60 +285,8 @@ class WrapperDesign extends React.PureComponent<Props, State> {
             }
           />
         )}
-        renderPageBody={() => (
-          <div
-            className={classnames('spec-editor layout-body--sidebar theme--pane', {
-              'preview-hidden': previewHidden,
-            })}>
```
develohpanda, line 221: > To get hiding the preview (in `WrapperDesign`) to trigger a style change in its parent component (`PageLayout`) would require the parent passing down a show/hide handler of some sort to `PaneTwo` to then call, right?

Don't think so - looks like the size changes from hiding the preview are all just controlled by the `grid-template-columns` CSS property.

I think you're missing this div and classes, though.

Previously in `WrapperDesign`, the parent div in the body contains these classes: `spec-editor layout-body--sidebar theme--pane`, but these no-longer exist on the DOM after splitting the body into the panes. You will probably need to send a `pageBodyClassName` (for example) prop into `PageLayout`, and wrap `paneOne` and `paneTwo` in a div, applying those classes.

```js
// in page-layout.js
<div className={pageBodyClassName}>
	{paneOne && <ErrorBoundary>...</ErrorBoundary>}
	{/* horizontal drag pane */}
	{/* vertical drag pane */}
	{paneTwo && <ErrorBoundary>...</ErrorBoundary>}
</div>
```

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -413,35 +411,32 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
     }
 
     return (
-      <div className="unit-tests layout-body--sidebar theme--pane">
```
develohpanda, line 416: `<div className="unit-tests layout-body--sidebar theme--pane">` is missing as well (see previous comment) 😄

packages/insomnia-app/app/ui/components/wrapper-design.js
simplywondrous: Thank you, yes it is `grid-template-columns` and it required only a slight change to `const gridColumns` in `PageLayout` to accomplish hiding the preview!

While both design and test tabs had a parent div around both panes, debug doesn't, and by changing from `renderPageBody` to `renderPaneX` a lot of the styling in `unit-tests.less` and `spec-editor.less` can be removed (currently unused, has been duplicated elsewhere, or is now covered by pane one / two styles).

simplywondrous:
Pushed a couple more changes and updated the PR body! Still have a couple todos, but please let me know if the logic looks okay, and if I'm missing something not in the todos!

packages/insomnia-app/app/ui/css/components/unit-tests.less
```diff
@@ -1,62 +1,17 @@
...
+  background: var(--color-bg);
+  color: var(--color-font);
+  overflow-y: auto;
+  box-sizing: border-box;
 
   .editor {
     overflow-y: hidden !important;
     max-height: 25em;
-    overflow-y: auto;
+    overflow-y: auto; // TODO - Jasmine do we want it auto or hidden?
```
simplywondrous: Which `overflow-y` is correct?

packages/insomnia-app/app/ui/components/page-layout.js
```diff
@@ -140,11 +157,53 @@ class PageLayout extends React.PureComponent<Props, State> {
...
+              <>
+                <div className="drag drag--pane-horizontal">
+                  <div
+                    onMouseDown={handleStartDragPaneHorizontal}
+                    onDoubleClick={handleResetDragPaneHorizontal}
+                  />
+                </div>
 
-        {renderPageBody && <ErrorBoundary showAlert>{renderPageBody()}</ErrorBoundary>}
+                <div className="drag drag--pane-vertical">
```
simplywondrous, line 180: Is this in the correct position? Visually, where is the vertical drag bar in the app?

packages/insomnia-app/app/ui/css/components/spec-editor.less
```diff
@@ -1,63 +1,10 @@
...
-        background: none;
-        border: none !important;
-      }
-    }
-
-    h1 {
-      padding: 0 var(--padding-md);
-      margin: 0;
-    }
-  }
```
simplywondrous: I didn't find anywhere where these styles were used - not missing anything, I hope?

packages/insomnia-app/app/ui/css/components/unit-tests.less
```diff
@@ -1,62 +1,17 @@
...
-      border-bottom: 1px dashed var(--hl-md);
-      padding-bottom: var(--padding-md);
-
-      &__header {
-        display: flex;
-        flex-direction: row;
-        align-items: center;
-        padding:0px var(--padding-sm) 0px 0px;
-      }
-    }
```
simplywondrous: Same for these styles here - not missing anything, I hope?

packages/insomnia-app/app/ui/components/wrapper-design.js
develohpanda: Oooh interesting, I'm all for removing redundant code 🔥

packages/insomnia-app/app/ui/components/page-layout.js
develohpanda: The vertical drag bar shows up in between the two panes when displayed vertically. The layout automatically changes when resizing. There is another PR open to enable this behavior for Design and Test, so don't worry if it doesn't immediately work, as long as it works for the Debug tab.

image

Does that answer your question?

packages/insomnia-app/app/ui/css/components/spec-editor.less
develohpanda: Entirely possible they are redundant!

packages/insomnia-app/app/ui/css/components/unit-tests.less
develohpanda: Oh good find! `auto` will always be overridden by `hidden` because of `!important`...

image

packages/insomnia-app/app/ui/css/components/unit-tests.less
develohpanda: A lot of this styling was moved into the component library (`insomnia-components`), using `styled-components`. This are most likely residual styles, nice find!

packages/insomnia-app/app/ui/css/components/pane.less
```diff
@@ -26,7 +25,6 @@
   }
 
   .pane__body {
-    border-left: 1px solid var(--hl-md);
```
develohpanda, line 29: I'm guessing this was moved from here into `.pane-one` and `.pane-two`?

packages/insomnia-app/app/ui/components/response-pane.js
```diff
@@ -210,16 +210,16 @@ class ResponsePane extends React.PureComponent<Props> {
 
     if (!request) {
       return (
-        <section className={paneClasses}>
+        <div className={paneClasses}>
```
develohpanda: Curious, is there a specific reason for switching from `section` to `div` tags?

develohpanda (commented):
> Consolidated CSS styling, mostly removing styles previously unused or now unused after these changes

One note about styles, in favor of keeping the scope of this PR small. If styles were previously unused, move to a separate PR. If styles are now unused after these changes, keep in this PR. 😄 

Looks to be on the right path! I cloned locally and ran it again, and am quite impressed with what you have achieved. 👏 

I'm not so familiar with the CSS classes and where they are used upfront, but as long as the end result is correct, I'm not _too_ concerned about the changes made there.

Also, it is possible to create [custom themes](https://support.insomnia.rest/article/26-plugins#themes) via Insomnia plugins. It would be worth comparing the live application with your PR, and researching any of the removed classes, to ensure theme support hasn't changed.

For example, the [Core Default](https://github.com/Kong/insomnia/blob/develop/plugins/insomnia-plugin-core-themes/themes/default.js) theme in Insomnia _is_ fairly broken for Designer, but it would be good to understand why there are differences between this PR and 2020.4.1. It will be something to do with CSS classes and hierarchy.

<details>
<summary>Screenshots</summary>

image

image

</details>

packages/insomnia-app/app/ui/components/response-pane.js
simplywondrous: Before there were more changes to this file, but I ended up reverting them afterwards. I kept these 'cause they match the other tabs, where PaneOne and PaneTwo render their children in `section` tags and the children themselves are `div` elements. Should I change these back?

packages/insomnia-app/app/ui/css/components/pane.less
simplywondrous: Yep!

packages/insomnia-app/app/ui/components/page-layout.js
simplywondrous: Yes, and seems vertical dragging for Debug is working fine so I'll leave it where it is now!

simplywondrous:
Aww shucks, I even tried out a bunch of different themes but I guess not the first one! It was missing two properties - it should be good now!

I'll keep the CSS changes as is in this PR and open another new PR with just the previously unused styles, if that one can be merged in first?

And thanks for the compliment, it means a lot!

packages/insomnia-app/app/ui/components/response-pane.js
develohpanda: Hmm I don't immediately see any issue with them, would you mind investigating whether there are any CSS styles impacted by this change? If not, I'm okay with leaving them as is, for consistency. Good catch!

packages/insomnia-app/app/ui/components/page-layout.js
```diff
@@ -140,11 +157,53 @@ class PageLayout extends React.PureComponent<Props, State> {
...
         )}
+        {renderPageBody ? (
+          <ErrorBoundary showAlert>{renderPageBody()}</ErrorBoundary>
+        ) : (
+          <>
+            {renderPaneOne && (
+              <ErrorBoundary showAlert>
+                <PaneOne ref={handleSetRequestPaneRef}>{renderPaneOne()}</PaneOne>
+              </ErrorBoundary>
+            )}
```
develohpanda: > For cases where panes aren't desired, `renderPageBody` is still available (such as for `WrapperHome`)

These two elements look very similar. If the `sidebar` and `paneTwo` don't exist (you already cater for it), isn't `paneOne` the same as `pageBody`, given `paneOne` will take up the space that is available? 😄

It looks like `WrapperHome` _does_ apply `theme--pane` as well, so it feels fine to use `PaneOne`, if I'm not missing anything.

I think what you've done for _this_ PR, by keeping `renderPageBody` for `WrapperHome`, is good, keeping in scope for the issue being resolved! 🎉 Although, it feels like `renderPageBody` is now also redundant and can be removed in a subsequent PR.

packages/insomnia-app/app/ui/components/page-layout.js
```diff
@@ -140,11 +157,53 @@ class PageLayout extends React.PureComponent<Props, State> {
...
+  render() {
+    return <section className="pane-one theme--pane">{this.props.children}</section>;
+  }
+}
+
+class PaneTwo extends React.PureComponent {
+  render() {
+    return <section className="pane-two theme--pane">{this.props.children}</section>;
+  }
+}
```
develohpanda: This can just be a single `Pane` component that takes in a prop to switch between the `pane-one` and `pane-two` classes. The `ref` for each pane will still be unique, because two instances of the same class component would end up on the tree. 🤓

packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -155,12 +154,84 @@ class WrapperDesign extends React.PureComponent<Props, State> {
     }
   }
 
+  renderEditor = (): React.Node => {
```
develohpanda: Would prefer regular function syntax (prefixed with `_` for private functions) here to keep consistent with the rest of the class methods 😄 same with `renderPreview` below and some other instances in this PR. There's probably an ES Lint rule to govern this that we should enable...

```suggestion
  _renderEditor(): React.Node {
```

packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -155,12 +154,84 @@ class WrapperDesign extends React.PureComponent<Props, State> {
...
+      swaggerUiSpec = parseApiSpec(activeApiSpec.contents).contents;
+    } catch (err) {}
+
+    if (!swaggerUiSpec) {
+      swaggerUiSpec = {};
+    }
+
+    return (
+      !previewHidden && (
+        <div id="swagger-ui-wrapper">
```
develohpanda: Always keep in mind to exit logic as early as possible (the previous implementation should have been better too).

While this is compact and correct, if the preview is hidden, we can avoid the `parseApiSpec` step altogether, because we will never show the result! With the refactoring you've done, this is now possible (easily). 😄 

`null` is a valid `React.Node`, so with that, we can do the following as the first step of `renderPreview()`

```js
if (previewHidden) {
  return null;
}
// everything else
```

packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -214,60 +285,8 @@ class WrapperDesign extends React.PureComponent<Props, State> {
...
-                uniquenessKey={activeApiSpec._id}
-              />
-              {lintMessages.length > 0 && (
-                <NoticeTable notices={lintMessages} onClick={this._handleLintClick} />
-              )}
-            </div>
-          </div>
-        )}
+        renderPaneOne={this.renderEditor}
+        renderPaneTwo={this.renderPreview}
```
develohpanda: So much cleaner. So much easier to understand the layout at a glance. 😍

packages/insomnia-app/app/ui/css/components/spec-editor.less
```diff
@@ -1,32 +1,10 @@
 .spec-editor {
```
develohpanda: I think you can get rid of this class altogether and just use `tall`.

(We're exploring tailwind to use some standard CSS helpers, rather than using custom ones like this. 😎)

https://github.com/Kong/insomnia/blob/bd4ed35dab63a91b55384fe7c36029589e6cc7ae/packages/insomnia-app/app/ui/css/layout/base.less#L352-L354

develohpanda (changes requested):
Looks really good! Just some minor feedback that would be worth addressing, but nothing blocking. 🤗 

> Aww shucks, I even tried out a bunch of different themes but I guess not the first one! It was missing two properties - it should be good now!

I cloned and tested again, and the already broken theme is slightly less broken now (and matches the live application) so that's good to see!

Lastly, could you also please add a gif to the PR description of the final feature, for documentation purposes because this is a UI/UX related change? Thanks!

develohpanda:
...also, a net change of +1 so far? Epic! This is bound to hit negatives after some of the suggested changes! 😎 

image

packages/insomnia-app/app/ui/components/page-layout.js
simplywondrous: I used `PaneOne` at first, but that has `pane-one` styles that account for a sidebar, with `grid-column-start: 4` and `border-left: 1px solid var(--hl-md)`. My thought is, it could be passed in with `position="0"` with `pane-zero` styles in a follow-up PR?

develohpanda: Hmmm I see... that could work! To me, the render props in `PageLayout` are the source of truth, and depending on their presence or absence, the layout should adjust, and that it's invisible to any _consumers_ of `PageLayout`.

We can probably remove the `pane-*` CSS classes altogether by using `styled-components`, where the styles can [change depending on props](https://styled-components.com/docs/basics#adapting-based-on-props), but I'm not too stubborn on which solution to use. Feel free to give that a go in another PR if you'd like!

develohpanda (approved):
QA:
- [x] horizontal resizing panes work on all tabs
- [x] vertical resizing panes work on all tabs
- [x] automatic/forced vertical layout works 🤯 
- [x] designer homepage correctly wraps and scrolls with multiple documents
- [x] view -> toggle sidebar (to hide sidebar) works on all tabs
- [x] after navigating around, I don't see any additional errors or warnings in the console window
- [x] can send requests (E2E tests)
- [ ] double check the on-boarding flow (E2E tests do cover this)

Nicely done! I think there ended up being some overlap with #2713, because this PR handles vertical layout automatically, but this PR would be my preferred direction. I did find another bug (#2787) that might already be fixed by #2713, though.

LGTM, thank you for working through the various iterations, and putting your time into it! I'm looking forward to merge it in. 💯

packages/insomnia-app/app/ui/components/page-layout.js
simplywondrous: I wasn't quite sure how to create a PR that goes against this repo when the branch itself is against my forked repo's branch - opening a PR against `Kong:develop` from `simplywondrous:fix/pane-update` meant "Files changed" included all of the changes from `simplywondrous:develop`. Do you know of a good way to do so? In the future, instead of forking should I clone and branch instead?

In the meantime, the PR is [here](https://github.com/simplywondrous/insomnia/pull/1). I figure, if this PR is merged, I can then create another one in this repo after.

develohpanda: Nested branches and forks can become tricky, so what you've done is fine. Once this PR is merged, you can create another PR into `Kong:develop` from `simplywondrous:fix/pane-update` for the additional changes, and I'll review those changes in that PR to keep conversation in one place.

> instead of forking should I clone and branch instead?

Clone and branch won't work as OSS contributors don't have push permissions to this repo, so working through forks is the best approach.

packages/insomnia-app/app/ui/components/page-layout.js
```diff
@@ -140,11 +157,53 @@ class PageLayout extends React.PureComponent<Props, State> {
             </div>
           </ErrorBoundary>
         )}
+        {renderPageBody ? (
+          <ErrorBoundary showAlert>{renderPageBody()}</ErrorBoundary>
+        ) : (
+          <>
+            {renderPaneOne && (
+              <ErrorBoundary showAlert>
+                <Pane position="one" ref={handleSetRequestPaneRef}>
```
DMarby, line 166: Since we set the ref here, but also set it on `RequestPane` in wrapper-debug as we did previously, I think the ref will be overwritten?
It should probably only be set in one of the two places?

packages/insomnia-app/app/ui/components/page-layout.js
```diff
@@ -140,11 +157,53 @@ class PageLayout extends React.PureComponent<Props, State> {
...
-        {renderPageBody && <ErrorBoundary showAlert>{renderPageBody()}</ErrorBoundary>}
+                <div className="drag drag--pane-vertical">
+                  <div
+                    onMouseDown={handleStartDragPaneVertical}
+                    onDoubleClick={handleResetDragPaneVertical}
+                  />
+                </div>
+
+                <ErrorBoundary showAlert>
+                  <Pane position="two" ref={handleSetResponsePaneRef}>
```
DMarby, line 188: Same here regarding setting the ref both here and in wrapper-debug

packages/insomnia-app/app/ui/components/wrapper-design.js
```diff
@@ -155,12 +154,86 @@ class WrapperDesign extends React.PureComponent<Props, State> {
...
   render() {
     const { gitSyncDropdown, wrapperProps, handleActivityChange } = this.props;
 
-    const { activeApiSpec, settings, activity, activeWorkspace } = wrapperProps;
+    const { activeApiSpec, activity, activeWorkspace } = wrapperProps;
 
-    const { lintMessages, previewHidden, hasConfigPlugins } = this.state;
+    const { previewHidden, hasConfigPlugins } = this.state;
 
     let swaggerUiSpec;
```
DMarby: I think this can be removed now, since it was moved up into _renderPreview?

DMarby (changes requested):

develohpanda:
Apologies for the merge conflicts! There was some refactoring in #2799, and looking at that PR might help. The changes in the conflicting files were mostly copy-paste so it should not be too tricky. 🤞 Let me know if you need any help resolving them.

packages/insomnia-app/app/ui/components/wrapper-design.js
simplywondrous: It's still used on line #274 of this file, for `{previewHidden ? 'Preview: Off' : 'Preview: On'}` in the header

packages/insomnia-app/app/ui/components/page-layout.js
simplywondrous: Removed that ref in wrapper-debug!

packages/insomnia-app/app/ui/components/page-layout.js
simplywondrous: Removed that ref in wrapper-debug!

simplywondrous:
Merged and updated :)

develohpanda:
Merge looks good! Any future conflicts should only be prop changes to `GrpcRequestPane` and `GrpcResponsePane`. Unfortunately there is another piece of work in progress that touches this code, but all of the foundational work is in place and I don't anticipate any more major conflicts.

packages/insomnia-app/app/ui/components/wrapper-design.js
DMarby: Sorry, to clarify, I meant that the `swaggerUiSpec` part here in `render()` can be removed unless I'm mistaken:
```
    let swaggerUiSpec;
    try {
      swaggerUiSpec = parseApiSpec(activeApiSpec.contents).contents;
    } catch (err) {}

    if (!swaggerUiSpec) {
      swaggerUiSpec = {};
    }
```

You moved that up to `_renderPreview` function where it's now used, but it's still defined here in `render()` as well, even tho it's not used there anymore.

DMarby (changes requested):
There's still some dead code in the `render()` function of WrapperDesign, but as long as that gets removed this looks good to go: https://github.com/Kong/insomnia/pull/2712#discussion_r519712832

Thanks for implementing this @simplywondrous!

packages/insomnia-app/app/ui/components/wrapper-design.js
simplywondrous: Ahh gotcha, removed!

packages/insomnia-app/app/ui/components/wrapper-debug.js
```diff
@@ -227,153 +219,139 @@ class WrapperDebug extends React.PureComponent<Props> {
...
-            </ResizablePaneWrapper>
-          </ErrorBoundary>
-          {dragPanes}
-          <ErrorBoundary showAlert>
-            <ResizablePaneWrapper ref={handleSetResponsePaneRef}>
-              <GrpcResponsePane
-                activeRequest={activeRequest}
-                forceRefreshCounter={forceRefreshKey}
-                settings={settings}
-              />
```
develohpanda, line 298: I think this snippet was lost after the last merge 😄 The `<GrpcResponsePane />` tag has been moved across to `_renderResponsePane()`, just need to put these three props on it!

develohpanda (changes requested):
Just keeping an eye on merge conflict resolution and noticed some props need to be sent through 👀 should be a quick one

gRPC work is 90% complete so there will not be any more conflicts 🤞

packages/insomnia-app/app/ui/components/wrapper-debug.js
simplywondrous: Oh! I looked at the last four commits but can't find where I accidentally merged that out... would you happen to know what happened? In any case, adding it back in now!

develohpanda: No stress, it was bound to be a tricky one and very evident during QA. Here's what _I think_ happened.

This particular change with the props was lost as part of the last merge commit [`0332327` (#2712)](https://github.com/Kong/insomnia/pull/2712/commits/0332327837bf978f6b8a0325819849d2cbeffce7), but you won't see it when viewing that individual commit.

Each commit in git is a _change since the previous commit_, right? For this branch, the previous commit is [`9037d8d` (#2712)](https://github.com/Kong/insomnia/pull/2712/commits/9037d8d93f146484e987ec7065cbdd579708bfca), and that commit didn't have these props, because these props were introduced in a very recent commit on develop.

However! When comparing the whole PR against develop we can see that the props are lost.

It's a bit easier to draw a commit tree to explain, but I hope that makes sense? 😆 When you last merged with develop and resolved conflicts, I think it should have shown the conflict, but because `wrapper-debug` had a **lot** of lines moved around, it would have been hard to spot. Even looking at this PR alone, `wrapper-debug` has a lot of stuff moved around, but few functional changes. It would have been difficult to spot while resolving conflicts, and your IDE probably wouldn't have helped either...

develohpanda: You could probably experiment as well if you want 🤓 create a new branch from [`9037d8d` (#2712)](https://github.com/Kong/insomnia/pull/2712/commits/9037d8d93f146484e987ec7065cbdd579708bfca) and try merge develop into it, you can double check the conflicts to see if that change was there.

develohpanda (approved):
🎉

DMarby (approved):
:shipit:

packages/insomnia-app/app/ui/components/wrapper-debug.js
simplywondrous: Ohh I see I see. Thanks for the explanation! Checking out that commit and then trying to merge, I see where the conflict was. 

Hmm... other than good tests and QA, is there a good way to detect merge errors like that one? I guess while it seems hard to trace within the branch, it'd be more trivial to trace the PR as the culprit if it were merged in with the error.

develohpanda: That's a great question! Unit and integration tests would definitely help, but unless there's really high test coverage in the impacted areas it's hard to rely entirely on them. Unfortunately I don't think there's one great solution for safely resolving large conflicts, and it feels more difficult in an open-source project.

In my experience, with any large refactor (in terms of lines/files in particular), I've found that it's best to do it in a tightly scoped piece of work, and get that merged into the main branch fairly quickly. For example, if a feature requires some groundwork refactoring, do the refactoring first and ensure there is _no functional change_ to the existing code, merge that in, and then build the feature on top. This way it's easier to review and validate the refactor and feature separately. The longer a valid refactor stays in a child branch, the more likely it is there will be a difficult conflict to resolve. This isn't always possible, but I opt for it when it _is_ possible.

Additionally, when you as a developer know that you work is likely to conflict with someone else's work, and it's going to be difficult to avoid a conflict, it's good to communicate that early and stay on the lookout. In previous roles I've paired up with a team-mate when it comes to resolving the conflict, as each party has a good understanding of each side of the conflict, so it's easier to resolve correctly. This is definitely more challenging to co-ordinate in the open-source world, where as a maintainer I'm trying to keep up with OS PRs that might conflict, and also internal work that might benefit from or conflict with OS PRs, but in a completely asynchronous manner.

My goal with this PR was to hopefully merge it in _before_ starting with the gRPC work as I was anticipating some conflicts, but timing/deadlines didn't quite align. Even viewing the diff of this PR, because of the number of lines changed in `wrapper-debug`, I would have missed that the props on the `GrpcResponePane` were lost... I only picked it up because I was explicitly looking for it, given that was my side of the conflict. Pseudo pair-programming! 😄

We try to QA each PR before merging it, but even if this slipped through the cracks, like you said tracing it back to this PR would have been quite easy as a reactive measure (have a look at [`git bisect`](https://www.metaltoad.com/blog/beginners-guide-git-bisect-process-elimination)). What I've described above are some preventative measures.

simplywondrous:
Is this PR waiting for another review? Otherwise could someone merge it in?

develohpanda:
We've held off on merging OS PRs for the last week or so, pushing for a big release. Now that 2020.5 is on the horizon, we'll start merging OS PRs again. No more reviews needed, this PR is good to go. 😄

roqueemir (commented):
> @

PR #2714, opened 2020-10-12
Show Client Certificate paths

### LINKED ISSUE #2300: Display Paths of Client Certs in Workspace Settings -> Client Certificates
When dealing with multiple client certs, it'd be good to conveniently look up the paths where the client certs are stored, especially as the edit feature is absent.

T

The straightforward approach would be to just add a bunch of rows (1 for pfx and/or 2 for key-crt) for each host entry, displaying the relevant paths.

Code-wise, seems this would need change: https://github.com/Kong/insomnia/blob/0a616fba6b12e3e51e3643f5f85f9fd27471f762/packages/insomnia-app/app/ui/components/modals/workspace-settings-modal.js ... renderCertificate(certificate: ClientCertificate)

For now, I can look up the data in %appdata%\Insomnia\insomnia.ClientCertificate.db

thewheat:
Closes: #2300

- Creates a new div that lists one file per line with some padding to help indicate line is grouped to upper row
- There is an extra gap between the lines that I'm not sure how to get rid (not to familiar with flex box 😅 but a negative margin seems to be a possible option) 
- Happy for any other comments or feedback especially in terms of how we'd want to display this 👍

**Before**
image

**After**
image

I

### DIFF
--- a/packages/insomnia-app/app/ui/components/viewers/password-viewer.tsx
+++ b/packages/insomnia-app/app/ui/components/viewers/password-viewer.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useCallback, useState } from 'react';
+import React, { FC, useCallback, useEffect, useState } from 'react';
 import { useSelector } from 'react-redux';
 import styled from 'styled-components';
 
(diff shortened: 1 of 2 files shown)


dimitropoulos:
I updated to do a few things:

- no duplication anymore (i.e. the thing with the checkbox _as well as_ a row with the path)
- network hierarchy is represented now (see https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/network/network.ts#L547, but basically if there's a cert _and_ a pfx, the pfx will win and the UI will now represent that).
- optional fields are not represented (e.g. passphrase, key)
- the passphrase is not now always visible in a tooltip, but instead is in a password viewer eyeball thingy (that respects the global setting for password reveal!)

Screenshot_20210909_142806

packages/insomnia-app/app/ui/components/viewers/password-viewer.tsx
```diff
@@ -0,0 +1,37 @@
...
+    setTextVisible(!textVisible);
+  }, [textVisible, onShow]);
+
+  return (
+    <span className="monospace">
+      <EyeIcon
+        className={`fa ${textVisible ? 'fa-eye' : 'fa-eye-slash'}`}
+        onClick={toggleVisible}
+      />
+      {textVisible ? <span className="selectable">{text}</span> : null}
```
gatzjames: Note for a followup on this: we could display a masked form of the text instead of null.

gatzjames (approved):
💯

packages/insomnia-app/app/ui/components/viewers/password-viewer.tsx
dimitropoulos: great suggestion: added in f4d269517dd71e17e69e856acb6c766d756622b0

localhost_3334_renderer html

PR #2715, opened 2020-10-12
Add export as HAR format for single requests

### LINKED ISSUE #650: [Feature Request] Way to dump request with response
frequently when debugging someone else's API i need to dump the request and response and send it over email.

Can we have a dump/export as text button of both request and response together?

### LINKED ISSUE #2611: [FEATURE REQUEST] Save OpenAPI as file in local file system
Insomnia needs to work with OpenAPI specs in the local file system. Currently it keeps them in an internal format only.
This has several problems:

1. The specs are not useable outside of Insomnia Designer.
1. One cannot reference definitions from other specs. Each spec has to be entirely self-contained.

It would be much simpler and useful to work with spec yaml's in local file system instead.

beardeddragon5:
Closes #650

### DIFF
--- a/packages/insomnia-app/app/common/har.ts
+++ b/packages/insomnia-app/app/common/har.ts
@@ -9,6 +9,7 @@ import type { Cookie } from '../models/cookie-jar';
 import type { Request } from '../models/request';
 import { newBodyRaw } from '../models/request';
 import type { Response, Response as ResponseModel } from '../models/response';
+import { isWorkspace } from '../models/workspace';
 import { getAuthHeader } from '../network/authentication';
 import * as plugins from '../plugins';
 import * as pluginContexts from '../plugins/context/index';
@@ -181,7 +182,7 @@ export async function exportHarCurrentRequest(request: Request, response: Respon
     models.workspace.type,
     models.requestGroup.type,
   ]);
-  const workspace = ancestors.find(ancestor => ancestor.type === models.workspace.type);
+  const workspace = ancestors.find(isWorkspace);
   if (workspace === null || workspace === undefined) {
     throw new TypeError('no workspace found for request');
   }
--- a/packages/insomnia-app/app/common/har.ts
+++ b/packages/insomnia-app/app/common/har.ts
@@ -8,7 +8,7 @@ import * as models from '../models';
 import type { Cookie } from '../models/cookie-jar';
 import type { Request } from '../models/request';
 import { newBodyRaw } from '../models/request';
-import type { Response as ResponseModel } from '../models/response';
+import type { Response, Response as ResponseModel } from '../models/response';
 import { getAuthHeader } from '../network/authentication';
 import * as plugins from '../plugins';
 import * as pluginContexts from '../plugins/context/index';
@@ -182,7 +182,7 @@ export async function exportHarCurrentRequest(request: Request, response: Respon
     models.requestGroup.type,
   ]);
   const workspace = ancestors.find(ancestor => ancestor.type === models.workspace.type);
-  if (workspace === null) {
+  if (workspace === null || workspace === undefined) {
     throw new TypeError('no workspace found for request');
   }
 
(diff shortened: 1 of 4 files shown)


packages/insomnia-app/app/common/har.ts
```diff
@@ -172,6 +173,33 @@ export interface Har {
...
   environmentId: string | null;
+  responseId?: string;
+}
+
+export async function exportHarCurrentRequest(request: Request, response: Response): Promise<Har> {
+  const ancestors = await database.withAncestors(request, [
+    models.workspace.type,
+    models.requestGroup.type,
+  ]);
+  const workspace = ancestors.find(ancestor => ancestor.type === models.workspace.type);
```
develohpanda: We can use `isWorkspace` here

```suggestion
  const workspace = ancestors.find(isWorkspace);
```

packages/insomnia-app/app/ui/components/panes/response-pane.tsx
```diff
@@ -178,6 +178,41 @@ class ResponsePane extends PureComponent<Props> {
...
   }
 
+  async _handleExportAsHAR() {
+    const { response, request } = this.props;
+
+    if (!response) {
+      // Should never happen
+      console.warn('No response to download');
+      return;
+    }
```
develohpanda, line 188: It's possible to export _without_ a response according to the `exportHar` implementation (although I know a response pane doesn't show when there's no response)

I think this is fine but it also seems fitting to add exporting a single request as HAR in the request context menu

image

packages/insomnia-app/app/common/har.ts
```diff
@@ -8,12 +8,13 @@ import * as models from '../models';
 import type { Cookie } from '../models/cookie-jar';
 import type { Request } from '../models/request';
 import { newBodyRaw } from '../models/request';
-import type { Response as ResponseModel } from '../models/response';
+import type { Response, Response as ResponseModel } from '../models/response';
```
develohpanda: Intentional aliased import for the same thing?

develohpanda (approved):
Seems to be working as intended!

image

packages/insomnia-app/app/common/har.ts
dimitropoulos: good one 7747354e58cfc50131dbf581228aaf4eac945a89

packages/insomnia-app/app/ui/components/panes/response-pane.tsx
dimitropoulos: @wdawson take a think about the above suggestion - we can always iterate.  seems ok to me, but I'm not sure how often people use HAR

packages/insomnia-app/app/common/har.ts
dimitropoulos: nope, that alias was there from before, but we can remove it now.  I wasn't aware of the alias because I wasn't looking at the code that used it.

good find b8eae1f95c49a3dfa45a147b219a32b83a23ca3d

packages/insomnia-app/app/common/har.ts
```diff
@@ -172,6 +174,33 @@ export interface Har {
...
+    models.requestGroup.type,
+  ]);
+  const workspace = ancestors.find(isWorkspace);
+  if (workspace === null || workspace === undefined) {
+    throw new TypeError('no workspace found for request');
+  }
+
+  const workspaceMeta = await models.workspaceMeta.getByParentId(workspace._id);
+  let environmentId = workspaceMeta ? workspaceMeta.activeEnvironmentId : null;
+  const environment = await models.environment.getById(environmentId || 'n/a');
```
gatzjames, line 192: Curious why is `n/a` needed here? Is it a pattern we use often or specific to this method?

gatzjames (approved):
🚀 Works great!

packages/insomnia-app/app/ui/components/panes/response-pane.tsx
wdawson: From my experience, HAR is used when there is some sequence of requests/responses. I would rather not crowd this menu unless we find that there's a big demand for it there.

packages/insomnia-app/app/common/har.ts
develohpanda: It's just a remnant of the past, especially when flow was a thing

PR #2716, opened 2020-10-12
Add file logging

DMarby:
This PR consolidates logs from both the renderer and the main process using [electron-log](https://github.com/megahertz/electron-log) and prints them to the console + logs them to file when the app is running in production mode.

It hijacks the `console.log/warn/etc` methods in order to handle all our current logs, but when adding new logs we should use the `./common/log` util instead moving forward.

For now, all log levels are printed and written to file, but we might want to change that in the future (for example, by setting different levels in development vs production builds, or have a setting in the app for more verbose logging)

### DIFF
--- a/packages/insomnia-app/app/common/log.js
+++ b/packages/insomnia-app/app/common/log.js
@@ -1,16 +1,21 @@
 import log from 'electron-log';
+import { isDevelopment } from './constants';
 
 export const initializeLogging = () => {
-  const fileTransport = log.transports.file;
-  const logFile = fileTransport.getFile();
+  if (isDevelopment()) {
+    // Disable file logging during development
+    log.transports.file.level = false;
+  } else {
+    const fileTransport = log.transports.file;
+    const logFile = fileTransport.getFile();
+    // Set the max log file size to 10mb
+    // When the log file exceeds this limit, it will be rotated to {file name}.old.log file.
+    fileTransport.maxSize = 1024 * 1024 * 10;
 
-  // Set the max log file size to 10mb
-  // When the log file exceeds this limit, it will be rotated to {file name}.old.log file.
-  fileTransport.maxSize = 1024 * 1024 * 10;
-
-  // Rotate the log file every time we start the app
-  fileTransport.archiveLog(logFile);
-  logFile.clear();
+    // Rotate the log file every time we start the app
+    fileTransport.archiveLog(logFile);
+    logFile.clear();
+  }
 
   // Overwrite the console.log/warn/etc methods
   Object.assign(console, log.functions);
--- a/packages/insomnia-app/app/common/log.js
+++ b/packages/insomnia-app/app/common/log.js
@@ -1,0 +1,19 @@
+import log from 'electron-log';
+
+export const initializeLogging = () => {
+  const fileTransport = log.transports.file;
+  const logFile = fileTransport.getFile();
+
+  // Set the max log file size to 10mb
+  // When the log file exceeds this limit, it will be rotated to {file name}.old.log file.
+  fileTransport.maxSize = 1024 * 1024 * 10;
+
+  // Rotate the log file every time we start the app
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/ui/index.js
```diff
@@ -94,7 +97,7 @@ if (window && !isDevelopment()) {
     trackEvent('Error', 'Uncaught Error');
   });
 
-  window.addEventListener('unhandledRejection', e => {
+  window.addEventListener('unhandledrejection', e => {
```
DMarby, line 100: In difference to Node, Chrome throws the event `unhandledrejection` in all lowercase, so this now works correctly.

jgiovaresco (approved):
Nice add, it should help bug investigation 👏 

I'm wondering if we should update the documentation to inform where the logs can be found? 
Like in the [FAQ](https://support.insomnia.rest/article/48-frequently-asked-questions)

DMarby:
That is the plan, will be adding a help article about it shortly.
> 
> I'm wondering if we should update the documentation to inform where the logs can be found? 
>

packages/insomnia-app/app/common/log.js
```diff
@@ -0,0 +1,24 @@
...
+  } else {
+    const fileTransport = log.transports.file;
+    const logFile = fileTransport.getFile();
+    // Set the max log file size to 10mb
+    // When the log file exceeds this limit, it will be rotated to {file name}.old.log file.
+    fileTransport.maxSize = 1024 * 1024 * 10;
+
+    // Rotate the log file every time we start the app
+    fileTransport.archiveLog(logFile);
+    logFile.clear();
```
develohpanda, line 17: Would you think it's worth printing to console the location of the log file or folder?

develohpanda (commented):
I like it 👏 a couple of questions. 

I couldn't see whether we provide any prefix for the log file. Does it automatically use the app name, and where is the file stored?

Do log files ever get cleared (not sure what is best practice here).

It might be worth adding an item to the Help menu item for "Show Log File/folder" or something similar, if it's not stored in the app data folder.

packages/insomnia-app/app/common/log.js
DMarby: I was considering it, but I don’t think it adds much value, most people won’t see the console logs, let alone need to also see the same logs but as a file.

packages/insomnia-app/app/ui/index.js
nijikokun: Nice catch 👍

DMarby:
> I couldn't see whether we provide any prefix for the log file. Does it automatically use the app name, and where is the file stored?

As per the `electron-log` documentation, they're stored in the following locations:
```
on Linux: ~/.config/{app name}/logs/{process type}.log
on macOS: ~/Library/Logs/{app name}/{process type}.log
on Windows: %USERPROFILE%\AppData\Roaming\{app name}\logs\{process type}.log
```
Eg `~/Library/Logs/Insomnia/main.log` and `~/Library/Logs/Insomnia/renderer.log` for Core on macOS.


> Do log files ever get cleared (not sure what is best practice here).

The log file gets rotated upon every startup (and thus overwrites the previously rotated file if present), so we only store the log for the current run + the log of the previous app run.


> It might be worth adding an item to the Help menu item for "Show Log File/folder" or something similar, if it's not stored in the app data folder.

We're using fairly standard logging locations, and those locations will be documented, so while I think this is worthwhile doing in the future (as I mentioned in the Linear issue a while back), I'm opting for it to be done in a future PR instead of this one, as there are more urgent priorities currently.
Will create an issue for it.

develohpanda (approved):
Thanks for the filename examples, akin to screenshots with a UI change! 

I missed the note in the Linear issue 🙏 please create a good first issue on GH for it as I do feel it is worthwhile, but not a blocker.

PR #2713, opened 2020-10-12
Improve visibilty of the drag pane when viewed vertically

rmsy:
This PR fixes #2584, by making `forceVerticalLayout` also apply to the "Design" and "Test" tabs. Here's an overview of the changes that were made:

-  Rename the settings options (for consistency) as described below. (I realize the requirement listed in #2584 was to change the text to read "Force vertical layout", but "force" feels a little weird to me, as both of the settings below are supported features, not something that needs to be "forced". If that is still the desired verbiage, let me know and I'll change the PR appropriately).
    - ~~"Force bulk header editor" -> "Use bulk header editor"~~ -- *This has been moved to #2717, see comments below*
    - "Vertical request/response layout" -> "Use vertical layout"
 - Update CSS to apply appropriate layout for design/test in vertical mode (this was the primary issue)

One additional note: there is [some custom logic in `app.js`](https://github.com/Kong/insomnia/blob/1bbbf38a9653792fd137662d015846f4334b1989/packages/insomnia-app/app/ui/containers/app.js#L822-L879) to support the draggable resize bar in the Debug tab, both in the horizontal and the vertical layouts. The Design/Test tabs do not have this resize bar implemented in the horizontal layout, so I didn't bring it over for the vertical layout, either. (I couldn't think of a simple and clean way to implement it, and I didn't want to clutter `app.js` any further). Maybe implementing the resize bar for Design/Test could be a future improvement.  

**Screenshots**:

image
image
image

Closes INS-923

PR #2717, opened 2020-10-13
Change bulk header editor setting verbiage

rmsy:
This changes the "Force bulk header editor" setting text to read "Use bulk header editor" instead, as it is less forceful, and to be consistent with #2713.


develohpanda (approved):

develohpanda:
Merging this in as there were no wording other changes suggested for this particular item as part of the related PR.

PR #2721, opened 2020-10-14
Fix infinite request spinner when cancelling "Send and Download"

### LINKED ISSUE #2718: Infinite Request Spinner when cancelling "Send and Download"
**Describe the bug**
Using the option "Send and download", if you cancel the save dialog without saving a file, the UI is stuck on the request infinitely. The "Cancel Request" button does not work. The only way to leave this state is to start another request.

**To Reproduce**
1. Create a new GET request to any URL (ie. https://insomnia.rest/)
2. Right click "Send" and select "Send And Download" from the drop down
3. When the Save File menu appears, cancel it via the "ESC" key or pressing the "Cancel" button

**Expected behavior**
Show the response and let the user know that the file has not been saved.

 - OS: Windows 10 x64 10.0.19041
 - Installation Method: Windows Installer
 - App Version: 2020.4.1

jgiovaresco:
We were hiding the spinner at the end of the `_handleSendAndDownloadRequestWithEnvironment` method. However, we
were exiting early we the user cancel the save dialog (L672).

To prevent that, I've moved the code in a finally block, therefore we make sure that the spinner will be hidden

Closes #2718

Before
BEFORE

After
AFTER

### DIFF
--- a/packages/insomnia-app/app/ui/containers/app.js
+++ b/packages/insomnia-app/app/ui/containers/app.js
@@ -713,15 +713,15 @@ class App extends PureComponent {
           </div>
         ),
       });
+    } finally {
+      // Unset active response because we just made a new one
+      await App._updateRequestMetaByParentId(requestId, {
+        activeResponseId: null,
+      });
+
+      // Stop loading
+      handleStopLoading(requestId);
     }
-
-    // Unset active response because we just made a new one
-    await App._updateRequestMetaByParentId(requestId, {
-      activeResponseId: null,
-    });
-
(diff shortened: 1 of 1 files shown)


develohpanda (commented):
Looks good to me! Nice, clean fix.

Could you please:
- move the button rename to a separate PR (with screenshots of what changes will be seen by the user)
- add a gif to this PR to show the cancellation logic working

develohpanda (approved):

DMarby (approved):

PR #2724, opened 2020-10-15
Fix sync issues when switching between workspaces

DMarby:
When using beta sync and switching between multiple workspaces that has sync enabled, there is a race condition, causing the wrong workspace to be selected as active in the VCS engine.

In order to solve this, I added a lock to `_updateVCS()`, to ensure that we only update the component state to the result of the most recent call. 
This means that if `_updateVCS` gets called while it's already running, the result of the first call will be discarded.
This should be a sufficient guard against the race conditions we were seeing, as running multiple of these vcs operations in parallel should be safe, so the only concern is the component state.

This is not an ideal solution, it would be better to move this state management into Redux instead, but that is a much larger change, which is why I opted for this approach. 
Open to suggestions for alternative solutions.

I also stopped triggering `_updateVCS()` in `UNSAFE_componentWillReceiveProps`, as this should not be needed, as we trigger it on both `componentDidMount` and `componentDidUpdate`.

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper-debug.js
+++ b/packages/insomnia-app/app/ui/components/wrapper-debug.js
@@ -124,7 +124,6 @@ class WrapperDebug extends React.PureComponent<Props> {
       activeUnitTestResult,
       activeWorkspace,
       environments,
-      enableSyncBeta,
       handleActivateRequest,
       handleSetActiveWorkspace,
       handleCopyAsCurl,
@@ -188,7 +187,7 @@ class WrapperDebug extends React.PureComponent<Props> {
                 unseenWorkspaces={unseenWorkspaces}
                 hotKeyRegistry={settings.hotKeyRegistry}
                 handleSetActiveWorkspace={handleSetActiveWorkspace}
-                enableSyncBeta={enableSyncBeta}
+                enableSyncBeta={settings.enableSyncBeta}
                 isLoading={isLoading}
                 vcs={vcs}
               />
--- a/packages/insomnia-app/app/ui/containers/app.js
+++ b/packages/insomnia-app/app/ui/containers/app.js
@@ -62,7 +62,12 @@ import WorkspaceSettingsModal from '../components/modals/workspace-settings-moda
 import RequestSettingsModal from '../components/modals/request-settings-modal';
 import RequestRenderErrorModal from '../components/modals/request-render-error-modal';
 import * as network from '../../network/network';
-import { debounce, getContentDispositionHeader, getDataDirectory } from '../../common/misc';
+import {
+  debounce,
+  getContentDispositionHeader,
+  getDataDirectory,
+  generateId,
+} from '../../common/misc';
 import * as mime from 'mime-types';
 import * as path from 'path';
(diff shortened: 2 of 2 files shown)


packages/insomnia-app/app/ui/components/wrapper-debug.js
```diff
@@ -188,7 +187,7 @@ class WrapperDebug extends React.PureComponent<Props> {
                 unseenWorkspaces={unseenWorkspaces}
                 hotKeyRegistry={settings.hotKeyRegistry}
                 handleSetActiveWorkspace={handleSetActiveWorkspace}
-                enableSyncBeta={enableSyncBeta}
+                enableSyncBeta={settings.enableSyncBeta}
```
DMarby, line 190: `enableSyncBeta` doesn't exist on WrapperProps, so we were always showing the legacy share dialog, even when beta sync was enabled.

Since we already pass all of the settings, I switched to using that instead, as there's no need to add an extra property to WrapperProps.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1075,7 +1085,10 @@ class App extends PureComponent {
 
     await vcs.switchProject(activeWorkspace._id);
 
-    this.setState({ vcs });
+    // Prevent a potential race-condition when _updateVCS() gets called for different projects in rapid succession
```
develohpanda: By different projects, you mean rapidly switching between different workspaces?

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1240,12 +1253,6 @@ class App extends PureComponent {
   // eslint-disable-next-line camelcase
   UNSAFE_componentWillReceiveProps(nextProps) {
     this._ensureWorkspaceChildren(nextProps);
-
-    // Update VCS if needed
-    const { activeWorkspace } = this.props;
-    if (nextProps.activeWorkspace._id !== activeWorkspace._id) {
-      this._updateVCS(nextProps.activeWorkspace);
```
develohpanda, line 1247: All good to remove from this life cycle hook. 👍

I'm sure you did this but could you please validate that this call is awaited in all other places in this file as well?

develohpanda (approved):
LGTM. Locking seems alright, I think I can see the edge case for where it might be needed.

As long as the updateVCS function is correctly awaited everywhere it should be, I'm happy with the locking approach.

packages/insomnia-app/app/ui/containers/app.js
DMarby: Yep, they're called projects within the sync code to make things more confusing 😅
I'll amend the comment

packages/insomnia-app/app/ui/containers/app.js
DMarby: Yep

packages/insomnia-app/app/ui/containers/app.js
develohpanda: I had an inkling that was the case 😅 thank you!

PR #2726, opened 2020-10-15
Allow Autocomplete to be properly registered as a Shortcut

### LINKED ISSUE #1633: [BUG] Can't change keyboard binding for "Show Autocomplete"
- Insomnia Version: 6.6.2
- Operating System: macOS Mojave 10.14.5

## Details

I use Alfred bound to ^-Space and it's too ingrained in my workflow to give up. I would like to change the the "Show Autocomplete" shortcut to something else, but no other shortcuts will trigger it. Interestingly it seems this key was the initial inspiration for customizable keybindings - #1162

### Steps to replicate

* Open keyboard preferences.
* Delete the existing "Show Autocomplete" shortcut.
* Deleted shortcut does not trigger autocomplete, as expected.
* Add Shift-^-Space, or Command-Shift-A, or anything else as the shortcut.
* **New shortcut does not trigger autocomplete.**
* Add ^-Space as the shortcut, or use reset to accomplish the same.
* Original shortcut triggers autocomplete, as expected.

thewheat:
- Closes #1633
- Currently we display "Show Autocomplete" under the list of customizable keyboard shortcuts **but in actual testing it does not work**
image
- The shortcut is currently hardcoded in the `autocomplete.js`
- Added a way to pass defined shortcuts into `autocomplete.js`

**Demo**
autocomplete mov

**Current issue**
~- If you modify the shortcut you need to reload Insomnia for it to register (still trying to find a way for this but open for guidance on this 😃)~ Fixed in https://github.com/Kong/insomnia/pull/2726/commits/15e9e4fa632d1c944043b5c0f2655873466dad72 with some details here https://github.com/Kong/insomnia/pull/2726#issuecomment-778721686


packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
```diff
@@ -129,6 +130,32 @@ CodeMirror.defineOption('environmentAutocomplete', null, (cm, options) => {
...
+    cm.removeKeyMap('autocomplete-keymap');
+
+    const settings = await models.settings.getOrCreate();
+    const keyBindings = settings.hotKeyRegistry[hotKeyRefs.SHOW_AUTOCOMPLETE.id];
+    const keyCombs = getPlatformKeyCombinations(keyBindings);
+
+    const keymap = {
+      name: 'autocomplete-keymap',
+      "' '": functionMap.completeIfAfterTagOrVarOpen,
+    };
```
thewheat: Docs on CodeMirror's keymap https://codemirror.net/doc/manual.html#keymaps

nijikokun:
We might need a little piece of text or notice that a reload is required. cc @erickrico

thewheat:
@nijikokun I could definitely look into that route in this (or in another PR if this PR is something we'd want to merge first) 👍
I did some digging and wasn't able to find anything helpful to trigger a shortcut refresh in CodeMirror. So having a message may be an easier and quicker win for now

thewheat:
I've dug around a bit but haven't been able to find a good way to refresh the shortcuts on update but in testing further it looks like a full reload isn't necessary, we just need to select a new request and it works

image

If we were to display a message, how best should we do it? Tried mocking up this up in the settings section and it looks a bit odd if just add to the description
image

Perhaps the displaying of the message could be worked on in another PR? I'll set this PR as ready for review and if we want all the changes in this PR I'll be happy to make the change with some guidance on how best proceed 👍

develohpanda:
Hi @thewheat! Sorry about the delay. There are a lot of PRs to get through and I haven't had a chance to work through them all yet. 😢 

> If you modify the shortcut you need to reload Insomnia for it to register (still trying to find a way for this but open for guidance on this 😃)

There is a `uniquenessKey` applied to the root component of the app to trigger a force refresh of the UI without the need to reload the app. You should be able to do something with altering this at an appropriate time. See the usages of `forceRefreshCounter` inside the file below.

https://github.com/Kong/insomnia/blob/543dc2df99ed602b002f532da7a937c92a3ffc61/packages/insomnia-app/app/ui/containers/app.js#L1283

By doing this, no messages will need to be shown to the user. I am _anticipating_ that you might face an issue where the insomnia preferences modal automatically closes (because all the components will refresh). If that is the case, then I have another very similar solution to propose, but try re-use this for now.

thewheat:
No worries @develohpanda ! Thanks for the tip there, I'll try do some digging this weekend 👍

thewheat:
Took a long while for me to get back to this 😅 Digging around I saw that there was a method to refresh the wrapper

https://github.com/Kong/insomnia/blob/e9fb34490ac4d118c0374d356133f41d3e498cd6/packages/insomnia-app/app/ui/components/wrapper.js#L495-L497

But using that didn't work. There was another method (shown below) that adds a 100ms delay before the refresh and that successfully got it working! 🎉  So dediced to refactor that logic into its own method to be called in the existing location and in the new code when shortcuts are updated

https://github.com/Kong/insomnia/blob/e9fb34490ac4d118c0374d356133f41d3e498cd6/packages/insomnia-app/app/ui/components/wrapper.js#L227-L236

If there was another way to approach this do let me know!

**Demo**
gif2

packages/insomnia-app/app/ui/components/modals/settings-modal.js
```diff
@@ -75,6 +75,7 @@ class SettingsModal extends PureComponent {
 
   async _handleUpdateKeyBindings(hotKeyRegistry) {
     models.settings.update(this.props.settings, { hotKeyRegistry });
+    this.props.handleUpdateKeyBindings();
```
develohpanda: Model updates are asynchronous and because the callback depends on keybindings having updated correctly, I think we need to `await` the model update now to avoid race conditions.

packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -682,6 +686,7 @@ class Wrapper extends React.PureComponent<WrapperProps, State> {
 
             <SettingsModal
               ref={registerModal}
+              handleUpdateKeyBindings={this._forceRequestPaneRefreshAfterDelay}
```
develohpanda: Seems like an OK solution to me given the current state 👍🏽

packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
```diff
@@ -133,6 +134,32 @@ CodeMirror.defineOption('environmentAutocomplete', null, (cm, options) => {
...
+
+    for (const keyComb of keyCombs) {
```
develohpanda: This is some awesome voodoo!!

Could you please add a small comment here about what this for-loop is doing to the `keymap` object, for anybody venturing here in the future? It was a little tricky to decipher when I was looking at it 😅

packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
```diff
@@ -133,6 +134,32 @@ CodeMirror.defineOption('environmentAutocomplete', null, (cm, options) => {
...
+      const alt = keyComb.alt ? 'Alt-' : '';
+      const ctrl = keyComb.ctrl ? 'Ctrl-' : '';
+      const meta = keyComb.meta ? 'Cmd-' : '';
```
develohpanda: I think this should only happen if we're on a mac, otherwise default back to ctrl 🤔 thoughts? https://github.com/Kong/insomnia/blob/c4babfc129c737eb46f1f92e2992c72dfae34bfa/packages/insomnia-app/app/ui/components/keydown-binder.js#L25

packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
```diff
@@ -133,6 +134,32 @@ CodeMirror.defineOption('environmentAutocomplete', null, (cm, options) => {
...
+      const shift = keyComb.shift ? 'Shift-' : '';
+      const keyname = CodeMirror.keyNames[keyComb.keyCode];
+
+      const key = `${shift}${meta}${ctrl}${alt}${keyname}`;
```
develohpanda: There's a snippet in the documentation above where it talks about the keymap being in the expected order, otherwise `normalizeKeyMap` should be used.

I think it might be worth adding a comment that the key order used here is significant (and to not change it), and maybe link to the docs?

> When a map contains multi-stoke bindings **or** keys with modifiers that are not specified in the default order (`Shift-Cmd-Ctrl-Alt`), you must call `CodeMirror.normalizeKeyMap` on it before it can be used. This function takes a keymap and modifies it to normalize modifier order and properly recognize multi-stroke bindings. It will return the keymap itself.

develohpanda (changes requested):
Really great work, and no worries about the delay 🙂 

Looks good to me overall! There's only one blocker in my review, and that is to `await` the model update before triggering a refresh. The remaining items are just observations where I think it might be useful to add some code comments. 🤗

packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
thewheat: This is actually on purpose
- We shouldn't map it to `Ctrl` as there is already `keyComb.ctrl`. The `keyComb` object indicates the full raw shortcut. So for Windows the meta key should map to `Win` and I guess on Linux it maps to `Meta`
- The `meta` variable is used to create the `key` value (line 157). `key` is then fed into CodeMirror whose docs says supports `Ctrl`, `Shift`, `Alt`, `Cmd`

image

Tested a couple in Windows and registering `Cmd` in Codemirror does work to recognize the `Win` key. [Test file here](https://gist.github.com/thewheat/d852b807a3b92cb7c37f6ab82f6d1e6d)
image

packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
thewheat: Sure thing!

packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
thewheat: Ah good call. Definitely will add it!

packages/insomnia-app/app/ui/components/modals/settings-modal.js
thewheat: Thanks for this. I just realised after updating to the latest code base, that the refreshing mechanism I used doesn't work correctly anymore (even with using `await models.settings.update(this.props.settings, { hotKeyRegistry });` 😞

- CodeEditor instances (the URL and the response body) do not get rerendered when `forceRefreshKey` is set to a new value. 
- They will only get rerendred when there are any changes in them

Any idea how to force them to rerender to pick up changes? 

image

thewheat: Hopefully these screenshots and GIF can help demonstrate the current issue

**Comparing old code base and new code base behaviours of CodeEditor rendering**
- [test code for old code base](https://github.com/thewheat/insomnia/tree/fix-autocomplete-shortcut-old-version-oct-27)
- [test code for new code base](https://github.com/thewheat/insomnia/tree/fix-autocomplete-shortcut-debug)
image

**New code base only rerenders CodeEditor when updating contents**
change_content_rerenders

**Demo of shortcuts not working until contents are changed**
Shortcuts dont work until contents change

develohpanda: Wow, amazing investigation! I'm not totally surprised at this because a _lot_ has changed since Oct 27th (we introduced gRPC, for starters!). I can think of a few changes which may have caused this behavior (have a look inside `page-layout.js`).

When I get a chance to code-review in detail I'll dig deeper and suggest some ideas for forcing them to refresh - might just be that we need to bind those two components with `forceRefreshKey` but I'd like to keep that as the last option.

develohpanda (approved):
I pushed a few commits to replace the `_forceRequestPaneRefreshAfterDelay` approach. While testing I had noticed that the code-editor would briefly lose focus while typing immediately after changing a keyboard shortcut.

Instead I replaced the delay and force refresh approach by connecting the code-editor to `settings.hotKeyRegistry` in redux. I tested this a bit and it seems to be working well! Please have a look at the last few commits that add this!

PR #2733, opened 2020-10-16
Add a keyboard shortcut for beautify request body

eimanip:
I set the new shortcut to `Shift+Ctrl+I`. It's the same keyboard shortcut for "Format Document" in VSCode, so I guess it could be a good candidate. Also, most of the current project's hotkeys follow the `Shift+Ctrl+[Char]` pattern. Feel free to suggest another default shortcut for this feature.

This pull request closes #1851 (partially), closes #1915 .

Screenshot from 2020-10-17 01-53-38

PR #2734, opened 2020-10-16
Add the ability to set "Automatically beautify request body" in settings

eimanip:
This features adds a new option in the Preferences -> General tab named "Automatically beautify request body". This option is disabled by default to preserve backward compatibility for the users. If it's enabled, future request body imports will be automatically beautified. It's based on the existing property `autoPrettify` in `CodeEditor` class.

This feature could also be an alternative way of what pull request #2305 was looking for (which is then reverted in #2387 due to some side effects). However, it includes smaller changes and uses existing logic to achieve the same goal.

Screenshot from 2020-10-17 01-56-42

### DIFF
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -16,6 +16,7 @@ export type PluginConfigMap = {
 type BaseSettings = {
   autoHideMenuBar: boolean,
   autocompleteDelay: number,
+  autoPrettify: boolean,
   deviceId: string | null,
   disableHtmlPreviewJs: boolean,
   disableResponsePreviewLinks: boolean,
@@ -72,6 +73,7 @@ export function init(): BaseSettings {
   return {
     autoHideMenuBar: false,
     autocompleteDelay: 1200,
+    autoPrettify: false,
     deviceId: null,
     disableHtmlPreviewJs: false,
     disableResponsePreviewLinks: false,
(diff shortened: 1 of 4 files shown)


nijikokun:
Looks good from a product perspective and is definitely a good middle-ground :+1:

dimitropoulos:
Thank you for this contribution!  I can see that a lot of conversation happened in the past about this and the team's reading on it at the moment is we need to come up with a more ergonomic solution.  For example, in many code editors (e.g. vscode, jetbrains, sublime, etc.) there are specific events that take place that will cause the formatting, but that's less obvious in the context of insomnia because in all of these examples one of the primary mechanisms is on file save, but there's no such concept in insomnia where you're typing these specs.  Therefore, we end up trying to find other things to hook into, e.g. blur and paste.

The other reason I have reservations about this is that it looks like it'll work for any kind of file, but in the end it only works for response body types that can be manually prettified.  For example, I would expect checking an "auto beautify response body" setting to prettify yaml, but it doesn't.

In addition, this didn't seem to work on paste.  I was able to get it to work when switching between requests, but overall I didn't see the functionality.  I think this is because the prettifying mainly occurs on componentDidUpdate which doesn't seem to trigger because the uniquenessKey doesn't change when prettifying.

_____

Some good news, though!  We just merged https://github.com/Kong/insomnia/pull/2733 which is very similar in effect to what this does (i.e. "make prettifying easier").

PR #2735, opened 2020-10-17
Request History: Set dropdown value to the response instead of null

### LINKED ISSUE #1985: [Bug] Can't switch to previous item in response history
**Describe the bug**
I'm unable to open previous response in history. I can go two steps back, but can't go one step back.

**To Reproduce**
1. Open history menu in top right corner
2. Click on previous item to go one step back.
3. **Nothing happens.**
4. Click on one of the earlier items. 
5. See that response.
6. Click on the item second from the top (same as in step 2).
7. **Last response opens.** 

**Expected behavior**
3. Opens previous response (second from the top).
7. Opens previous response (second from the top).

**Screenshots**
https://yadi.sk/i/he8mbtNQFPzGXg

 - OS: [e.g. macOS 10.15.3 (19D76)]
 - Installation Method: .dmg from site
 - App Version: 7.1.1.3350

**Additional context**
I noticed this issue after updating to Catalina.

### LINKED ISSUE #2675: History – cannot navigate to item (1st in 2nd group)
**Description**
1st item in second group in history cannot be navigated as long as it is 1st one in 2nd group.

**How to reproduce**
1. Execute 2 request.
2. Wait for about 8 mins.
3. Execute new request.
4. Navigate to history.
5. Open last request (i.e. 1st of step 1).
6. Try open 2nd request from step 1.

**Expected behavior**
History is opened.

**Screenshots**
History

Version: Insomnia Core 2020.4.1
Release date: 9/23/2020
OS: Windows_NT x64 10.0.19041
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.69.1-DEV OpenSSL/1.1.1d zlib/1.2.11 WinIDN libssh2/1.9.0_DEV nghttp2/1.40.0

**Additional context**

- Video can be provided in case of misunderstanding.
- Instead of 1st item from 2nd group is chosen 1st item from1st group.

thewheat:
- Closes: #2675, Closes #1985, Closes INS-913
- Currently there can be cases where selecting the request history drop down returns an incorrect request
image
- This is due to code returning a `null` instead of the `response` when `i == 0`
- As the code uses `response.id` as the key it seems like we will always have a `response` object.

**Before (current broken behaviour)**
before

**After (which this fix)**
after

**Context of the bug**
- So the bug is due to this section of code
https://github.com/Kong/insomnia/blob/5a128ed25be4bc0d253507199e31eb8a861dc532/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.js#L125-L137
- Each `.map` will start at index of `0` so the first item in the new grouping would cause the issue

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.tsx
+++ b/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.tsx
@@ -55,7 +55,7 @@ class ResponseHistoryDropdown extends PureComponent<Props> {
     });
   }
 
-  renderDropdownItem(response: Response, i: number) {
+  renderDropdownItem(response: Response) {
     const { activeResponse, requestVersions } = this.props;
     const activeResponseId = activeResponse ? activeResponse._id : 'n/a';
     const active = response._id === activeResponseId;
--- a/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.tsx
+++ b/packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.tsx
@@ -68,7 +68,7 @@ class ResponseHistoryDropdown extends PureComponent<Props> {
       <DropdownItem
         key={response._id}
         disabled={active}
-        value={i === 0 ? null : response}
+        value={response}
         onClick={this._handleSetActiveResponse}
       >
         {active ? <i className="fa fa-thumb-tack" /> : <i className="fa fa-empty" />}{' '}


develohpanda (approved):
Thank you for fixing this and my apologies for the delay in review, it's been quite challenging to keep up with open-source PRs, but the team is working on it. We really appreciate your time! 🙌🏽

Verified as fixed in this PR, and broken in the current stable release.

PR #2736, opened 2020-10-17
Allow prompt values to be clear + add actions to template tag plugin API

### LINKED ISSUE #1456: [Feature Request] Allow prompt values to be cleared
- Insomnia Version: 6.4.1
- Operating System: Ubuntu 16.04 LTS

## Details

The Prompt tag is a really useful way of getting a user to provide a piece of data that can't be derived from a previous request. 

It'd be good if the user could choose to clear the value stored against a Prompt Tag. As it stands, there only seem to be two ways of doing this :

1. Close the application and restart it.
2. Change the Storage Key (the help tooltip even suggests this, but it's not a very nice way to do it).

## Use Case
I want to use the prompts at the start of a series of chained requests which make up a typical usage scenario. E.g. my API allows a user to navigate through a document page by page via multiple requests. At the start of my chain I want to provide the ID of the document to be traversed in a prompt.

I want the same document ID to be used on every request in the chain as I work my way through the document. Once I've finished I need to be able to execute the scenario again - sometimes with a different document ID, sometimes with the same one!

At the moment I'd have to either restart the tool at the end of the chain, or edit the Prompt Tag and change the Storage Key to achieve a "re-prompt".

## Solution ideas
Perhaps the "Edit Tag" screen could have a "Clear value" button on it somewhere? Or maybe there could be some other way of clearing these stored values.

jgiovaresco:
Closes #1456

Implementation of what Greg describes in this [comment](https://github.com/Kong/insomnia/issues/1456#issuecomment-524422705): add a way to define actions in template tag.

Some screenshot of the TagEditor

With 1 action
Screenshot 2020-11-27 at 09 12 14

With more actions
Screenshot 2020-10-17 at 18 47 24
image

Clear prompt values
clear-prompt-value


develohpanda:
> However, I'm blocking on how we can provide some context to the action callbacks.

My guess is to investigate how the `helperContext` is built (when needed) for running the template tag, and do something similar for the button, in `base-extension`. Note that the context should be built _when required_. For example, once the button is clicked, generate the context and then send it to the plugin.

https://github.com/Kong/insomnia/blob/27d588168457679f9e28463d43c7119b3c2f58b6/packages/insomnia-app/app/templating/base-extension.js#L130

I suggest only exposing the `store` context for now, but it should be built in the standard way that allows other contexts' to be added, should they be required.

jgiovaresco:
@develohpanda

I've managed to provide a context to the action callback. I had to look deeper on how plugins & nunjunks tags are working but it was interesting :)
I'm wondering if I should create a PR that will add only the feature "Define actions buttons on a template tag" and another that will fix the original issue? Or I keep this one 🤔

packages/insomnia-app/app/plugins/context/store.js
```diff
@@ -2,7 +2,16 @@
 import type { Plugin } from '../index';
 import * as models from '../../models';
 
-export function init(plugin: Plugin) {
+export type PluginStore = {
```
develohpanda, line 5: 👍

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -252,6 +258,19 @@ class TagEditor extends React.PureComponent<Props, State> {
...
+    const templateTags = await getTemplateTags();
+    const activeTemplateTag = templateTags.find(({ plugin, templateTag }) => {
+      return templateTag.name === this.state.activeTagData.name;
+    });
+
+    const helperContext = {
+      ...pluginContexts.store.init(activeTemplateTag.plugin),
+    };
+
+    return action.run(helperContext);
```
develohpanda: I suggest after running an action we also refresh the preview. In your original gif, after you clear the prompt, the preview still shows the old value and it looks as though nothing happened.

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -626,6 +645,30 @@ class TagEditor extends React.PureComponent<Props, State> {
...
+    );
+  }
+
+  renderAction(action: NunjucksActionTag, index: number) {
+    const name = action.name;
+    const icon = action.icon ? <i className={action.icon} /> : undefined;
+
+    return (
+      <button
+        key={index}
```
develohpanda: We need to [avoid index as keys](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). You should be able to use the `name` as the key, given we can safely assume actions in a particular plugin will always have unique names.

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -626,6 +645,30 @@ class TagEditor extends React.PureComponent<Props, State> {
     );
   }
 
+  renderActions(actions = []) {
+    return (
+      <div className="form-row">
+        {actions.map((action, index) => this.renderAction(action, index))}
```
develohpanda: Can be simplified 😄 

```suggestion
        {actions.map(this.renderAction)}
```

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -671,6 +714,9 @@ class TagEditor extends React.PureComponent<Props, State> {
           activeTagDefinition.args.map((argDefinition: NunjucksParsedTagArg, index) =>
             this.renderArg(argDefinition, activeTagData.args, index),
           )}
+
+        {activeTagDefinition && this.renderActions(activeTagDefinition.actions)}
```
develohpanda: I think this should avoid rendering an empty `form-row` if no actions exist, as such you can either do it here or inside `this.renderActions`.

```suggestion
        {activeTagDefinition?.actions?.length && this.renderActions(activeTagDefinition.actions)}
```

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -252,6 +258,19 @@ class TagEditor extends React.PureComponent<Props, State> {
     this._update(this.state.tagDefinitions, tagDefinition, null, false);
   }
 
+  async _handleActionClick(action: NunjucksActionTag) {
+    const templateTags = await getTemplateTags();
+    const activeTemplateTag = templateTags.find(({ plugin, templateTag }) => {
+      return templateTag.name === this.state.activeTagData.name;
+    });
```
develohpanda: Doesn't look like plugin is used any longer?

```suggestion
    const activeTemplateTag = templateTags.find(({ templateTag }) => {
      return templateTag.name === this.state.activeTagData.name;
    });
```

develohpanda (changes requested):
> I've managed to provide a context to the action callback. I had to look deeper on how plugins & nunjunks tags are working but it was interesting :)

💯 

> I'm wondering if I should create a PR that will add only the feature "Define actions buttons on a template tag" and another that will fix the original issue? Or I keep this one 🤔

Am I understanding correctly that in order to create another PR to fix the original issue, you will only need to move the changes in `plugins/insomnia-plugin-prompt/index.js`? If that's the case, I think what you've done here is fine. In order to fix the original issue, the plugin API needed to be extended so this is all related.

Sorry about the delay in getting to this, I've been really bogged down!

One request: could you please create _new commits_ when addressing review comments, rather than amending existing commits? It's difficult to see what has changed between iterations, and requires us to review the entire PR again. It can become quite time consuming with even a moderately sized PR. Because we squash and merge, a clean commit history in a PR is not _so_ important for us. Thank you! 🤗

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -626,6 +645,30 @@ class TagEditor extends React.PureComponent<Props, State> {
     );
   }
 
+  renderActions(actions = []) {
```
develohpanda, line 649: It might be good to label the buttons row with "Actions", similar to how  "Title" and "Label" appear, so that we have some context for the section. Looking at the screenshots, the actions section looks a little out of place.

packages/insomnia-app/app/ui/components/templating/tag-editor.js
jgiovaresco: 🤦  
A great developer explained that to me during a tech interview I had recently :p

develohpanda: You did this PR beforehand, and I just noticed now while reviewing 😆

develohpanda (approved):
LGTM 👍

packages/insomnia-app/app/plugins/context/store.js
```diff
@@ -2,7 +2,16 @@
...
+  hasItem(string): Promise<boolean>,
+  setItem(string, string): Promise<void>,
+  getItem(string): Promise<boolean | null>,
```
DMarby: This signature doesn't seem to match the implementation

```suggestion
  getItem(string): Promise<string | null>,
```

DMarby (approved):

DMarby (changes requested):
Looks good other then the mismatching type definition 👍

packages/insomnia-app/app/plugins/context/store.js
jgiovaresco: good catch 😅

DMarby (approved):

develohpanda:
Docs have been updated: https://support.insomnia.rest/article/26-plugins#template-tags

smacktoid:
Amazing! My issue got fixed!

develohpanda:
Thanks for reporting it initially, @scottmmmm! We do have a release planned for this week, unfortunately this PR won't make it in. There will likely be a patch release early 2021 with this included!

PR #2732, opened 2020-10-16
Adds sorting to GQL Documentation Browser

tomprats:
Really simple addition to sort list of graphql queries. Really helps in apps with a large number. Right now they're shown in a seemingly random order. Still might be better to have a full search (since ctrl-f doesn't seem to work), but this is a good start either way.

My original change was just:

```
+     {Object.keys(fields).map(key => {
-     {Object.keys(fields).sort().map(key => {
```

But eslint didn't like that. It's auto-fix did something like:

```
{Object.keys(fields)
  .sort()
  .map(key => {
    ...
```

Which indented most of the file a whole lot, so I wasn't sure if it was worth the large change log or if a variable was the better option. Let me know what you'd prefer and I can update it!

PR #2738, opened 2020-10-18
Add sort options for sidebar rows

### LINKED ISSUE #2011: Enhancement request; Provide option to sort Requests on left panel alphabetically.
It would be very useful if an option was provide to sort the list of Request items (left panel) alphabetically. Thanks!

### LINKED ISSUE #2594: Add sort options for folders and requests
Hard to find Folder / Request

Sort the saved Folder / Request by either create date / name / anything...

Nil

**Additional context**
Add more filed to describe the Folder /  Request name , easy for hashtag or find out the resource.

cobwebsonsale:
Closes #2594, Closes #2011
**Implementation details:**
1. Sending sortHandler function and currentSortOrder from App, all the way down to sidebarFilter
1. Created a new dropdown on the filter bar to display sort options
1. Added 7 options to sort the request rows

**Assumptions:**
1. For all sort orders, I'm not mixing the request rows inside the requestGroup to those outside
1. For sorting by HTTP method, since it does not apply to RequestGroups, I move them to the bottom and bring all Requests in a sorted order to the top.
1. Whenever the sort criteria is the same for two rows, I fall back to `metaSortKey` to keep the old order preserved. This allows multiple sorts to be applied. For example: Sort by 'name ascending' and then 'folders first' causes RequestGroups to come to top of the list in alphabetical order.
1. Offered some orders that I thought were useful. Let me know if should edit the list.

**I had some questions about the repo/feature:**
1. I wanted to reset sort order to 'CUSTOM' if a document is moved, so that the currentSortOrder is de-selected and can then be re-applied, but we can't access `this.state` in `App._moveDoc()`. Is there a reason why this method is defined outside the App class unlike the other methods in the file?
1. I can't seem to render the new sort-order-dropdown and the create-request-dropdown in the same line (added image below). Can you guys help me with that?
1. How/where can I add tests?
1. [not so important] I tried to add icons like [fa-sort-alpha-down](https://fontawesome.com/icons/sort-alpha-down?style=solid) to the rendering of sort-order display names, but a blank was rendered. Do I have to follow some other steps here?
1. [not so important] I thought we could save the ordering across different sessions. Should we? If so, what would be the best place to keep that.
1. [not so important] I wanted to reuse a method `_recalculateMetaSortKey` inside the old `_moveDoc` method as well, but the `__updateDoc` method inside it uses `getModel(docToMove.type)`. I wondered why it was this, and not `getModel(doc.type)`.

Looks like this so far:
sort


nijikokun:
Great stuff, the one thing that I think is obvious is that this pushes the new request to a new line which we definitely don't want, so I'm going to loop in @erickrico to see if he can take a look from a UI/UX perspective

conema:
@cobwebsonsale my two cents: you can try editing the 63th row of  `packages\insomnia-app\app\ui\css\components\sidebar.less` with `grid-template-columns: minmax(0, 1fr) auto auto;` to make the grid with three colums instead of two. With this edit all the icons should be in the same line, but I feel like that the textarea is too little.

develohpanda:
> this pushes the new request to a new line which we definitely don't want, so I'm going to loop in erickrico to see if he can take a look from a UI/UX perspective

@nijikokun there are notes in the original issue (with feedback from Erick) about UX

> With this edit all the icons should be in the same line, but I feel like that the textarea is too little.

Are you concerned about the `filter` text area being too little? How does it look with the default sidebar width? It's potentially okay, otherwise we can maybe reduce the padding on each button.

conema:
> Are you concerned about the `filter` text area being too little? How does it look with the default sidebar width? It's potentially okay, otherwise we can maybe reduce the padding on each button.

Yep, It's a little smaller than the one on your screenshot on the issue (I cannot post a screenshot because I resized the sidebar and I don't know how to reset the sizes). Reducing the paddings would be perfect.

nijikokun:
> @nijikokun there are notes in the original issue (with feedback from Erick) about UX

Yep, I feel we might need a better solution since it seems like it's cascading into other things (reducing padding, smaller input, responsiveness, etc). Starting to feel crowded and cramped.

packages/insomnia-app/app/common/constants.js
```diff
@@ -257,6 +257,37 @@ const authTypesMap = {
   [AUTH_NETRC]: ['Netrc', 'Netrc File'],
 };
 
+// Sort Orders
+export const SORT_CUSTOM = 'custom';
+export const SORT_NAME_ASC = 'name-asc';
```
develohpanda: You should be able to type this using a union - some duplication but it's safer to consume. See `GlobalActivity` for an example.

```suggestion
export type SortOption = 'custom' | 'name-asc' | ...;

export const SORT_CUSTOM: SortOption = 'custom';
export const SORT_NAME_ASC: SortOption = 'name-asc';
```

Later on, you can consume it like:
```js
export function getSortOrderName(sortOrder: SortOption): string {
	...
}
```

I selected `SortOption` arbitrarily, feel free to rename as you see fit.

packages/insomnia-app/app/common/constants.js
```diff
@@ -257,6 +257,37 @@ const authTypesMap = {
...
+    case SORT_METHOD:
+      return 'Method';
+    case SORT_TYPE_ASC:
+      return 'Folder First';
+    case SORT_TYPE_DESC:
+      return 'Folder Last';
+    default:
+      return '';
+  }
+}
```
develohpanda: This might be more efficient as a regular map 😄 

```js
const sortOrderName: { [SortOption]: string } = {
	[SORT_NAME_ASC]: 'Name Ascending',
	[SORT_NAME_DESC]: 'Name Descending',
	...
}
```

packages/insomnia-app/app/ui/components/sidebar/sidebar-filter.js
```diff
@@ -7,11 +7,14 @@ import type { HotKeyRegistry } from '../../../common/hotkeys';
...
 import { executeHotKey } from '../../../common/hotkeys-listener';
 import SidebarCreateDropdown from './sidebar-create-dropdown';
+import SidebarSortDropdown from './sidebar-sort-dropdown';
 
 type Props = {
   onChange: string => void,
   requestCreate: () => void,
   requestGroupCreate: () => void,
+  sidebarSort: Function,
+  sortOrder: string,
```
develohpanda: Ideally, props are typed more explicitly.

- What parameters are expected by the `sidebarSort` function, and what does it return?
- After introducing a union type (see my previous comment), this can be updated to `sortOrder: SortOption`.

I've made the comment only here, but please update all other instances of this as well. Thank you!

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
```diff
@@ -0,0 +1,59 @@
...
+} from '../../../common/constants';
+import { Dropdown, DropdownButton, DropdownItem } from '../base/dropdown';
+
+type Props = {
+  handleSort: Function,
+  sortOrder: string,
+};
+
+@autobind
+class SidebarSortDropdown extends React.PureComponent<Props> {
```
develohpanda: Is it possible to use a regular function component here? React Hooks are available in this project, for new components such as this.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
```diff
@@ -0,0 +1,59 @@
...
+  _handleSort(order: string) {
+    if (order !== this.props.sortOrder) {
```
develohpanda: What happens if I sort by a particular order, then rename a request or create a new one - does it move the request to the correct location?

If it doesn't move the request, but I want to order again by the same selection, this conditional will prevent me from doing so.

Is there a case for adding this condition that I am overlooking? Otherwise it might be good to remove the conditional and sort regardless of current and previous selection.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
```diff
@@ -0,0 +1,59 @@
...
+  renderSortOrder(order: string) {
+    const currentSortOrder = this.props.sortOrder || SORT_CUSTOM;
+
+    return (
+      <DropdownItem onClick={this._handleSort} value={order}>
+        {currentSortOrder === order ? <i className="fa fa-check" /> : <i className="fa fa-empty" />}{' '}
+        {getSortOrderName(order)}
+      </DropdownItem>
+    );
+  }
```
develohpanda: 👏 Nicely done.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -115,6 +124,7 @@ class App extends PureComponent {
       forceRefreshCounter: 0,
       forceRefreshHeaderCounter: 0,
       isMigratingChildren: false,
+      activeSortOrder: props.activeSortOrder || SORT_CUSTOM,
```
develohpanda: See an earlier comment - this should be removed from app state (here) and placed in the `workspace-meta` model, next to [`sidebarFilter`](https://github.com/Kong/insomnia/blob/b0b7613f2636b14f9cb700aea3b596ea5e3a3fb5/packages/insomnia-app/app/models/workspace-meta.js#L29).

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -335,6 +345,80 @@ class App extends PureComponent {
     });
   }
 
+  _getSortMethod(order) {
```
develohpanda: I believe this entire method can (and should) exist outside of the `app.js` class.

Create a standalone sorting helper file. You can then also write unit tests for each switch case to ensure items are ordered correctly.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -335,6 +345,80 @@ class App extends PureComponent {
...
+    }
+  }
+
+  async _recalculateMetaSortKey(docs) {
+    function __updateDoc(doc, metaSortKey) {
+      models.getModel(doc.type).update(doc, { metaSortKey });
+    }
+
+    await db.bufferChanges(300);
+    docs.map((doc, i) => __updateDoc(doc, i * 100));
```
develohpanda: The model update method is asynchronous, so all of these updates will return a promise. Do you need to await all promises? Is that what the `bufferChanges` line is for?

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -335,6 +345,80 @@ class App extends PureComponent {
...
+
+    const docs = [
+      ...(await models.requestGroup.findByParentId(parentId)),
+      ...(await models.request.findByParentId(parentId)),
+    ].sort(this._getSortMethod(order));
+
+    this._recalculateMetaSortKey(docs);
+
+    // sort RequestGroups recursively
+    docs.filter(d => d.type === models.requestGroup.type).map(g => this._sortSidebar(order, g._id));
```
develohpanda: Nice use of recursion, and reusing `docs` rather than loading `requestGroup` from the database again.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1335,6 +1419,8 @@ class App extends PureComponent {
               handleUpdateDownloadPath={this._handleUpdateDownloadPath}
               isVariableUncovered={isVariableUncovered}
               headerEditorKey={forceRefreshHeaderCounter + ''}
+              handleSidebarSort={this._sortSidebar}
+              sortOrder={this.state.activeSortOrder}
```
develohpanda: Once saved in `workspaceMeta`, you should be able to load it from `activeWorkspaceMeta`, so no need to send the selected sort order as a prop.

develohpanda (commented):
Thanks for the PR, and welcome to the project! Nice work. ⭐

I'll answer your questions below.

**Assumptions**

> 1. For all sort orders, I'm not mixing the request rows inside the requestGroup to those outside

Sounds good, we shouldn't alter the parent-child hierarchy 👍 

> 2. For sorting by HTTP method, since it does not apply to RequestGroups, I move them to the bottom and bring all Requests in a sorted order to the top.

I am assuming this will still sort the requests by method _within_ the request group, though?

> 3. Whenever the sort criteria is the same for two rows, I fall back to metaSortKey to keep the old order preserved.

Seems like logical behavior, this is what I would probably expect as a user, but it can be changed with user feedback.

> 4. Offered some orders that I thought were useful. Let me know if should edit the list.

They all seem fine to me 👍 

**Questions**

> 1. I wanted to reset sort order to 'CUSTOM' if a document is moved, so that the currentSortOrder is de-selected and can then be re-applied, but we can't access this.state in App._moveDoc(). Is there a reason why this method is defined outside the App class unlike the other methods in the file?

> 5. [not so important] I thought we could save the ordering across different sessions. Should we? If so, what would be the best place to keep that.

I suggest persisting the sort order in the database under the `workspace-meta` model, parallel to the [sidebarFilter](https://github.com/Kong/insomnia/blob/b0b7613f2636b14f9cb700aea3b596ea5e3a3fb5/packages/insomnia-app/app/models/workspace-meta.js#L29). This way, you can set it to 'CUSTOM' when required, and it will persist correctly between different workspaces.

> 2. I can't seem to render the new sort-order-dropdown and the create-request-dropdown in the same line (added image below). Can you guys help me with that?

See the [comment above](https://github.com/Kong/insomnia/pull/2738#issuecomment-712436472) for putting it on the correct line. **Spacing, and any positioning changes can be addressed independent of the feature implementation**, so try and get it looking _nice_ for now, without expending too much effort on it until confirmation from @erickrico on UX.

> 3. How/where can I add tests?

No UI tests exist at the moment, but see my review comments on how to break down the code such that some tests can be added.

> 4. [not so important] I tried to add icons like fa-sort-alpha-down to the rendering of sort-order display names, but a blank was rendered. Do I have to follow some other steps here?

It's possible that we are using an older version of FA that doesn't contain this icon. What exactly did you try - `<i className="fa fa-sort-alpha-down"/>`? We can also add custom icons if necessary ([for example](https://storybook.insomnia.rest/?path=/story/iconography-core-icons--reference)).

> 6. [not so important] I wanted to reuse a method _recalculateMetaSortKey inside the old _moveDoc method as well, but the __updateDoc method inside it uses getModel(docToMove.type). I wondered why it was this, and not getModel(doc.type).

I'm not sure off the top of my head and it seems a little confusing. Since you've marked it lower priority, I'll defer investigating and answering until the PR has been updated and if this is still a question then.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
cobwebsonsale: Good catch! The entire idea of storing the activeSortOrder in state, sending it in props, adding this condition, and having a SORT_CUSTOM (which gets selected on moving a request-row around, or on renaming a resource in your example) brought this feature out as a one-time-option-selection in a list that adds a check-mark next to its name to show what's the current order. According to this view, upon movement/renaming, the list _should_ get sorted automatically.

Instead, I think I can simplify this by presenting this feature as just a modification that is applied to request-rows (and remove the check-mark-next-to-the-name UI). This way, clicking an already applied modification can be expected to have no effect, and I will not have to store the sort order or send it in props. Does that make sense?

develohpanda: I think I get you - what you're saying is to never persist the sort order, it's more of an action taken when selecting a particular ordering, and creating a new request or renaming would not put it in the right location, until you chose to reorder again.

If that's correct, seems pretty decent, as it would eliminate the need for custom ordering too, because it would, effectively always be custom.

develohpanda: I'll get back to you about persisting the sorting vs making it on-click so don't worry about making any changes here at the moment, but it shouldn't block you from doing some of the other refactoring from my review. 🤗

cobwebsonsale: I'm done with all other refactors except moving activeSortOrder to workspaceMeta, which I'll hold off on, until this discussion concludes. I've pushed the remaining changes in just now.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
cobwebsonsale: Yep, definitely!

packages/insomnia-app/app/ui/components/sidebar/sidebar-filter.js
cobwebsonsale: Added explicit types to both these props in all places.

packages/insomnia-app/app/common/constants.js
cobwebsonsale: Changed to a map. Thanks for the suggestion!

packages/insomnia-app/app/common/constants.js
cobwebsonsale: Cool. Refactored to a Union. Thanks for the reference.

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: I've moved this to a helper file. (added a new folder under ui/. let me know if that's not the right location). Will add tests next.

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: I copied these lines from how `_moveDoc` in `app.js` updates the docs. I'll investigate what `bufferChanges` does and get back to you.

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: Thanks!

cobwebsonsale:
> Thanks for the PR, and welcome to the project! Nice work. ⭐

Thanks! :)

> > For sorting by HTTP method, since it does not apply to RequestGroups, I move them to the bottom and bring all Requests in a sorted order to the top.
>
> I am assuming this will still sort the requests by method within the request group, though?

Yep!

> > [not so important] I tried to add icons like fa-sort-alpha-down to the rendering of sort-order display names, but a blank was rendered. Do I have to follow some other steps here?
> 
> It's possible that we are using an older version of FA that doesn't contain this icon. What exactly did you try - `<i className="fa fa-sort-alpha-down"/>`? We can also add custom icons if necessary (for example).

Yeah, that's what I tried. I think I can keep the order names as text for now, and we can improve the UI in a subsequent PR.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
cobwebsonsale: > I think I get you - what you're saying is to never persist the sort order, it's more of an action taken when selecting a particular ordering, and creating a new request or renaming would not put it in the right location, until you chose to reorder again.

Yep, exactly.

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: `bufferChanges` only prevents components/listeners from being notified of db changes for a set time (here 300 - and now 200 - ms). Added an `await` for promises returned by the model `update`s.

Also, now I'm only buffering DB for changes once, in the first call to _sortSidebar.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1530,9 +1560,12 @@ async function _moveDoc(docToMove, parentId, targetId, targetOffset) {
   }
 
   function __updateDoc(doc, patch) {
-    models.getModel(docToMove.type).update(doc, patch);
+    return models.getModel(docToMove.type).update(doc, patch);
```
cobwebsonsale, line 1578: Line `1571` below 

``` 
await __updateDoc(docToMove, { parentId });
```

does not use its `await` if this method does not return a promise.

develohpanda:
Hi @cobwebsonsale! We had a good discussion about this feature, and have a solution now.

> > I think I get you - what you're saying is to never persist the sort order, it's more of an action taken when selecting a particular ordering, and creating a new request or renaming would not put it in the right location, until you chose to reorder again.
>
> Yep, exactly.

We'll go ahead with this - no persistence of the sort order, it will be an on-click action. With user feedback, we can look to add persistence (and all the related behaviors) in the future. This also removes the need for a checkbox in the sorting dropdown.

I call on @erickrico to share a screenshot of the updated design. There is a higher-level rework we are thinking about for that top section of the sidebar in the future, so no radical changes there, mainly some spacing improvements.

packages/insomnia-app/app/ui/containers/app.js
develohpanda: Nice catch. Does anything else call `__updateDoc`, and if so, how?

packages/insomnia-app/app/ui/containers/app.js
develohpanda: Nice! I recently came across an approach that doesn't use a fixed timeout, and it feels a bit nicer.

`bufferChanges` returns an id, which can be used to explicitly flush by that id later on. I'm now wondering whether the support for flushing via an id was added after the code in `_moveDoc` was written. If it's possible to use an approach similar to the one in the snippet below, I suggest updating it.

https://github.com/Kong/insomnia/blob/5852d27d3f11a51474519132e881b13da715e994/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js#L279-L283

packages/insomnia-app/app/ui/helpers/sorting.js
```diff
@@ -0,0 +1,62 @@
+// @flow
```
develohpanda: Could you please move this file under `app/common/sorting.js`?

packages/insomnia-app/app/ui/helpers/sorting.js
```diff
@@ -0,0 +1,62 @@
...
+  SORT_CREATED_LAST,
+  SORT_METHOD,
+  SORT_NAME_ASC,
+  SORT_NAME_DESC,
+  SORT_TYPE_ASC,
+  SORT_TYPE_DESC,
+} from '../../common/constants';
+import { request, requestGroup } from '../../models';
+
+export function _getSortMethod(
```
develohpanda: Is there a particular reason you chose to prefix by `_`? Generally the exports in this code-base aren't prefixed, so just curious.

packages/insomnia-app/app/ui/helpers/sorting.js
```diff
@@ -0,0 +1,62 @@
...
+  SortOrder,
+  SORT_CREATED_FIRST,
+  SORT_CREATED_LAST,
+  SORT_METHOD,
+  SORT_NAME_ASC,
+  SORT_NAME_DESC,
+  SORT_TYPE_ASC,
+  SORT_TYPE_DESC,
+} from '../../common/constants';
+import { request, requestGroup } from '../../models';
```
develohpanda: I suggest importing the types directly, in addition to the models themselves if necessary

```suggestion
import { request, requestGroup } from '../../models';
import type { Request } from '../../models/request';
import type { RequestGroup } from '../../models/request-group';
```

packages/insomnia-app/app/ui/helpers/sorting.js
```diff
@@ -0,0 +1,62 @@
...
+          const bIndex = HTTP_METHODS.indexOf(b.method);
+          if (aIndex !== bIndex) {
+            return aIndex < bIndex ? -1 : 1;
+          } else {
+            return a.metaSortKey < b.metaSortKey ? -1 : 1;
+          }
+        } else {
+          return a.metaSortKey < b.metaSortKey ? -1 : 1;
+        }
+      };
```
develohpanda: This is purely a stylistic suggestion, but it might be worth extracting each switch case into a standalone function in this file, so the switch statement is easier to understand at a glance, and the unit tests are clearer (import and test each function directly).

develohpanda (commented):
Making progress! I did a quick review 😄

packages/insomnia-app/app/ui/helpers/sorting.js
```diff
@@ -0,0 +1,62 @@
...
+  SORT_TYPE_DESC,
+} from '../../common/constants';
+import { request, requestGroup } from '../../models';
+
+export function _getSortMethod(
+  order: SortOrder,
+): (a: request | requestGroup, b: request | requestGroup) => -1 | 1 {
+  switch (order) {
+    case SORT_NAME_ASC:
+      return (a, b) => (a.name.toLowerCase() < b.name.toLowerCase() ? -1 : 1);
```
develohpanda: Taking note from [here](https://github.com/Kong/insomnia/pull/2732#discussion_r513407240), it might be worth using `localCompare` for user-populated data (such as `name`). TIL!

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: Yep, it's called in two other places and they don't wait for the promise to resolve. If you're okay with me making that change in this PR, I can add await to them!

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: Yep, I had tried for some time to do it the flushId way, because otherwise 200ms or 300ms is just a guess so it's not the best, but I couldn't find a way. Let me commit a not-so-nice way and you can tell me if it's okay.

packages/insomnia-app/app/ui/helpers/sorting.js
cobwebsonsale: Done! Thanks for the pointer.

packages/insomnia-app/app/ui/helpers/sorting.js
cobwebsonsale: No, I'd moved it from inside `app.js` and forgot to remove the `_` that it had as it was an internal method. Fixed!

packages/insomnia-app/app/ui/helpers/sorting.js
cobwebsonsale: Thanks! I couldn't find a way to do this when I tried.
Also, using these `type`s in the if checks inside sorting methods does not work, so I kept those the same.

packages/insomnia-app/app/ui/helpers/sorting.js
cobwebsonsale: It'll definitely be more readable. Will do!

packages/insomnia-app/app/ui/helpers/sorting.js
cobwebsonsale: That makes sense! Changed this.

cobwebsonsale:
Updated:
new_sort

packages/insomnia-app/app/ui/helpers/sorting.js
develohpanda: Yep that's fine. The types aren't available at runtime so you did that check correctly.

cobwebsonsale:
I do not have permission to view the logs of the steps that failed in the `develop` merge commit. Please let me know if I can check them locally somewhere.

develohpanda:
> 
> 
> I do not have permission to view the logs of the steps that failed in the `develop` merge commit. Please let me know if I can check them locally somewhere.


Netlify is not critical for PRs (only a storybook deployment) so I'll likely ignore it for now. As long as the "Test" GitHub Actions pass, you're all good!

(Netlify is failing to find/install `rimraf`..., which I've seen with other PRs recently as well)

packages/insomnia-app/app/common/constants.js
```diff
@@ -272,20 +272,20 @@ export type SortOrder =
...
-  [SORT_CREATED_FIRST]: 'Created First',
-  [SORT_CREATED_LAST]: 'Created Last',
-  [SORT_METHOD]: 'Method',
-  [SORT_TYPE_ASC]: 'Folder First',
-  [SORT_TYPE_DESC]: 'Folder Last',
+  [SORT_CREATED_ASC]: 'Oldest First',
+  [SORT_CREATED_DESC]: 'Newest First',
+  [SORT_METHOD]: 'HTTP Method',
+  [SORT_TYPE_DESC]: 'Folders First',
+  [SORT_TYPE_ASC]: 'Requests First',
```
develohpanda: I like this rewording 🎉

packages/insomnia-app/app/common/sorting.js
```diff
@@ -37,54 +37,66 @@ export function getSortMethod(
   }
 }
 
-function ascendingNameSort(a: Request | RequestGroup, b: Request | RequestGroup): 1 | -1 {
+export function ascendingNameSort(a: Request | RequestGroup, b: Request | RequestGroup): number {
```
develohpanda: Since every function in this file has the same signature, you can do something like:

```js
// Union type of all models that are sortable (more on this below)
type SortableModel = Request | RequestGroup;

// Type definition of a function
type SortFunc = (a: SortableModel, b: SortableModel) => number;

// a and b should be detected correctly to be a SortableModel
export const ascendingNameSort: SortFunc = (a, b) => {...}
```

There is a reason behind this, because I'm going to throw a spanner in the works 😅 I've been churning through gRPC work over the past week, and gRPC requests now also exist in the sidebar alongside folders and HTTP requests. The model for a gRPC request is `GrpcRequest`, and while 90% of the logic you've written should JustWork™️, `SortableModel` will need to accept a another type in it's union. Defining the types like I've done in the snippet above, will require you to only change that in one location, and makes the file slightly less verbose.

packages/insomnia-app/app/common/__tests__/sorting.test.js
```diff
@@ -0,0 +1,229 @@
...
+  METHOD_GET,
+  METHOD_HEAD,
+  METHOD_OPTIONS,
+  METHOD_PATCH,
+  METHOD_POST,
+  METHOD_PUT,
+} from '../constants';
+
+describe('Sorting methods', () => {
+  beforeEach(globalBeforeEach);
```
develohpanda: This is only needed if you're hitting the database, which this test suite is not, so you should be able to safely omit it. 😄

packages/insomnia-app/app/common/__tests__/sorting.test.js
```diff
@@ -0,0 +1,229 @@
...
+
+  it('sorts by metaSortKey', () => {
+    expect(metaSortKeySort({ metaSortKey: 1 }, { metaSortKey: 2 })).toBe(-1);
+    expect(metaSortKeySort({ metaSortKey: 2 }, { metaSortKey: 1 })).toBe(1);
+    expect(metaSortKeySort({ metaSortKey: -2 }, { metaSortKey: 1 })).toBe(-1);
+    expect(metaSortKeySort({ metaSortKey: 1 }, { metaSortKey: -2 })).toBe(1);
+    expect(metaSortKeySort({ metaSortKey: 1, _id: 2 }, { metaSortKey: 1, _id: 1 })).toBe(-1);
+    expect(metaSortKeySort({ metaSortKey: 1, _id: 1 }, { metaSortKey: 1, _id: 2 })).toBe(1);
+  });
+});
```
develohpanda, line 343: Nice, thorough tests 👏 You should be able to run this test suites with coverage in IntelliJ, and see if there are any cases you might have missed, but I'd be surprised if it doesn't return a really high test coverage!

packages/insomnia-app/app/ui/components/wrapper-debug.js
```diff
@@ -267,15 +268,14 @@ class WrapperDebug extends React.PureComponent<Props> {
...
-              hotKeyRegistry={settings.hotKeyRegistry}
-            />
+        <div className="drag drag--pane-vertical">
+          <div
+            onMouseDown={handleStartDragPaneVertical}
+            onDoubleClick={handleResetDragPaneVertical}
+          />
+        </div>
+      </>
+    );
```
develohpanda: Interesting, is it possible that this is a faulty merge, or was this change intentional?

develohpanda (commented):
Great work :) I'm definitely a fan of the unit tests.

As I mentioned in [this comment](https://github.com/Kong/insomnia/pull/2738#discussion_r524074739), I'm throwing a spanner in the works with gRPC support. In the last week or so, we've introduced a new model to the sidebar (!), so now the sidebar can show `Request`, `RequestGroup`, or `GrpcRequest`. The good news is that these models are quite similar for your use case, so **most of your sorting functions should work out of the box**. 😄 

The only difference is that `GrpcRequest` does not have a typical HTTP Method; it's just `GRPC`, so the HTTP Method sort function will need a small update. I've added some helpers to detect the type of a model in a readable way, so pull from here if you need it! https://github.com/Kong/insomnia/blob/ccd270a9fb6c86863fe5e5003a791ffe3485a0d0/packages/insomnia-app/app/models/helpers/is-model.js#L5-L7

If you import [this workspace](https://gist.github.com/develohpanda/57f96270875fba0478f9dc2ffeaee738), everything should be set up with an HTTP request and a gRPC request in the sidebar to test with. (Be sure to update with develop first, I only added import support for gRPC on Friday 😅)

packages/insomnia-app/app/ui/components/wrapper-debug.js
develohpanda: Ah, ignore this 😄

packages/insomnia-app/app/ui/containers/app.js
develohpanda: I'd say it's okay to leave them as is. I'm not a fan of the "if it ain't broke, don't fix it" mantra, but I just want to avoid any adverse side-effects, at least as part of this PR. `app.js` is very central to the entire app as you've probably noticed. 😄

packages/insomnia-app/app/ui/containers/app.js
develohpanda: I just had a look at your implementation in [`df4ac8e` (#2738)](https://github.com/Kong/insomnia/pull/2738/commits/df4ac8e0d6bdbca7d1ab83adc81c81018036961d); seems alright to me. What did you feel was not-so-nice about it?

packages/insomnia-app/app/common/constants.js
cobwebsonsale: Thanks!

packages/insomnia-app/app/common/sorting.js
cobwebsonsale: Nice. Makes things more readable. Made the suggested changes. Added `GrpcRequest` handling for http sort (GrpcRequests appear at the bottom of the request list) and type sort (GrpcRequests are treated as normal requests for folder first/request first sort orders).

packages/insomnia-app/app/common/__tests__/sorting.test.js
cobwebsonsale: Nice. Added some more tests for the changes I made and also using `getSortMethod` function now. Coverage at 100%.

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: The `if` condition to filter out the function run from the stack of recursive calls had sounded a little off. I think I was going to try the `parentId` check for the bottom condition too, but being able to use `flushId` itself made it look a little better. I think I'm okay with the implementation now, should have updated this thread. 😅

packages/insomnia-app/app/ui/containers/app.js
cobwebsonsale: yep yep. sounds good~

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
```diff
@@ -0,0 +1,49 @@
...
+import { Dropdown, DropdownButton, DropdownItem } from '../base/dropdown';
+
+type Props = {
+  handleSort: (sortOrder: SortOrder) => void,
+};
+
+const SidebarSortDropdown = (props: Props) => {
+  const _handleSort = (order: SortOrder) => {
+    props.handleSort(order);
+  };
```
develohpanda: `_handleSort` seems like an unnecessary wrapper for `props.handleSort` 🤔 Might be a remnant from a previous refactor 🙂

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const _renderSortOrder = (order: SortOrder) => {
+    return (
+      <DropdownItem onClick={_handleSort} value={order}>
+        {sortOrderName[order]}
+      </DropdownItem>
+    );
+  };
```
develohpanda: `_renderSortOrder` is, effectively, a component which receives `handleSort` and `order` as props. This is a similar model to `AbcList` and `AbcListItem`  components. Is there a particular reason you chose to make `_renderSortOrder` a function inside the parent component?

If you create a standalone component for this, something like `SortItem`, you can avoid re-creating this function on every render. This is more of a nit / readability / performance suggestion.

Note, you can extend the original props (to not re-define the sort handler).

```js
type SortItemProps = Props & { order: SortOrder };
```

packages/insomnia-app/app/common/sorting.js
```diff
@@ -0,0 +1,114 @@
...
+
+const createdLastSort: SortFunction = (a, b) => {
+  if (a.created === b.created) {
+    return 0;
+  }
+
+  return a.created > b.created ? -1 : 1;
+};
+
+const httpMethodSort: SortFunction = (a, b) => {
```
develohpanda, line 45: This function is a little confusing at first sight so some plain comments to explain what is happening would be awesome. In any case, this logic is unit tested so there is some safety around it still. 🎉

develohpanda (approved):
I'm happy with this now, and SO excited to get it merged! Thank you for the time you put into it and the various iterations. I did some QA locally and it worked great. 🙂 

There are a couple of comments I've added, but none of them are blocking. 💯

develohpanda:
> Updated:
> new_sort

@nijikokun @erickrico please have a look at this final UI. IMO this is complete, any UI changes can be standalone.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
cobwebsonsale: Oh yes, thanks for pointing out! Remnant of the version that stopped sorting by same order more than once.

packages/insomnia-app/app/common/sorting.js
cobwebsonsale: Added some comments.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
cobwebsonsale: Okay, added a component but the dropdown stopped functioning. Moving it to a separate file did not help either. It may be because of how the `Dropdown` classes are defined? 🤔 Looked at other uses of `DropdownItem` but none of them used a component either.
Figured, this is a simple enough component, so changed the render to a `map` of a list (instead of moving away from a functional component). Added the list in constants, similar to other things.

packages/insomnia-app/app/common/constants.js
```diff
@@ -260,6 +260,42 @@ const authTypesMap = {
...
+  | 'created-first'
+  | 'created-last'
+  | 'method'
+  | 'type-desc'
+  | 'type-asc';
+export const SORT_NAME_ASC: SortOrder = 'name-asc';
+export const SORT_NAME_DESC: SortOrder = 'name-desc';
+export const SORT_CREATED_ASC: SortOrder = 'created-asc';
+export const SORT_CREATED_DESC: SortOrder = 'created-desc';
+export const SORT_HTTP_METHOD: SortOrder = 'http-method';
```
cobwebsonsale, line 276: Refactored to disambiguate `SORT_METHOD` (http-method) from `sortMethod`s (functions).

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
```diff
@@ -0,0 +1,27 @@
...
+const SidebarSortDropdown = (props: Props) => {
+  const { handleSort } = props;
+
+  return (
+    <Dropdown>
+      <DropdownButton className="btn btn--compact">
+        <i className="fa fa-sort" />
+      </DropdownButton>
+      {SORT_ORDERS.map(order => (
+        <DropdownItem onClick={() => handleSort(order)} value={order} key={order}>
```
cobwebsonsale: Couldn't get this to work with `onClick={handleSort}`. Selecting a dropdown option throws [this](https://github.com/electron/electron/issues/23743) error. I'm unsure of differences between this use of dropdown and that in `grpc-method-dropdown`.

packages/insomnia-app/app/common/sorting.js
```diff
@@ -0,0 +1,108 @@
...
+
+export const metaSortKeySort: SortFunction = (a, b) => {
+  if (a.metaSortKey === b.metaSortKey) {
+    return a._id > b._id ? -1 : 1;
+  }
+
+  return a.metaSortKey < b.metaSortKey ? -1 : 1;
+};
+
+export const sortMethodMap: { [SortOrder]: SortFunction } = {
```
cobwebsonsale, line 100: Remembered your [suggestion](https://github.com/Kong/insomnia/pull/2738/files/e42fd1d0721613c92e0cab0cf259f1db2406b1ac#r508187547), changed this to a map too.

cobwebsonsale:
@develohpanda, thanks for patiently reviewing this so many times. 🙇‍♂️

packages/insomnia-app/app/common/sorting.js
develohpanda: Perfect!

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
develohpanda: Nice refactor! 😄

That might be an entirely unrelated error 🤔 What you've got is essentially identical to the usage in `grpc-method-dropdown`...

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
develohpanda: Yep, I like what you've done! It's strange that using a different component _didn't_ work, but that doesn't matter anymore 😄

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
develohpanda: I cloned locally and have found the culprit. TS would have helped here, because it's much better at type-checking than Flow. 😄 (We will eventually update, it's just a large undertaking).

It boils down to the fact that the signature for `App._sortSidebar` is actually `(order: SortOrder, parentId?: string): Promise<void>`, but the props in `sidebar-sort-dropdown` define it as `(sortOrder: SortOrder) => void`. Since `parentId` is only needed for the recursion this makes sense, but it hides some detail. I'm not sure which is the lesser of two evils. 😆 

The `Promise` doesn't matter so much as the second argument, and the fact that `dropdown-item` actually sends the click event as it's second argument. https://github.com/Kong/insomnia/blob/d5dde51b85d239e5a37e9a56244ef6bda11639a9/packages/insomnia-app/app/ui/components/base/dropdown/dropdown-item.js#L19-L20

This means that when `App._sortSidebar` is called, it's called with `parentId` set to a click event. We then try to load from the database with `parentId` set to a click event, which naturally fails and results in the error thrown from db internals.

All in all, you won't be able to use `onClick={handleSort}` directly but can either use the first argument, or leave it as-is and remove the `value` prop. 😄 That was fun!

develohpanda:
> thanks for patiently reviewing this so many times. 🙇‍♂️

And thank you for working through it! We're on the home-stretch now.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -342,6 +343,40 @@ class App extends PureComponent {
...
+      ...(await models.requestGroup.findByParentId(parentId)),
+      ...(await models.request.findByParentId(parentId)),
+      ...(await models.grpcRequest.findByParentId(parentId)),
+    ].sort(sortMethodMap[order]);
+    await this._recalculateMetaSortKey(docs);
+
+    // sort RequestGroups recursively
+    await Promise.all(
+      docs
+        .filter(d => d.type === models.requestGroup.type)
```
develohpanda: This can just become `.filter(isRequestGroup)` now 👀 so much easier on the eyes haha

cobwebsonsale: Wrapped up neatly in a line with this.

packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.js
cobwebsonsale: oooooooo! It was puzzling me so much, thanks for the find! Was why the last update took so much time. Everything makes sense now. 🕊️ 

Using `value` was longer lines, so didn't, but you're right, that would mean I can remove the prop altogether, nice!

develohpanda: Sounds good! 🤗

DMarby (approved):

nijikokun (approved):
This turned out _awesome_. Huge props to everyone involved, excited to see it get merged 🤗  🎉

PR #2740, opened 2020-10-19
Fix Copy as CURL with Multipart

### LINKED ISSUE #2282: Copy as CURL with Multipart
If you make an Copy with CURL in an POST with an Multipart Body with an choosen file, the file is not include.

1. Get in POST request
2-Use MULTIPART as body type and
3. Add a field
4. Change the type to "File"
5. Then -> Choose File (what you want)
6. right click on the post
7. Copy as CURL
8. Paste IT to an editor
9. Look on the result and see that the file is missing

**Expected behavior**
Include the file as multipart in the curl

**Additional context**
image

jgiovaresco:
The issue was located in the `exportHar` function. It was not handled `file` param correctly.
For this type of file, the `value` attribute is empty. Therefore we should take instead of the `fileName`

The issue also impacted the `Generate code` feature.

Closes #2282

**Before**

```
curl --request POST \
  --url http://localhost:3000/multipart \
  --header 'Content-Type: multipart/form-data' \
  --header 'content-type: multipart/form-data; boundary=---011000010111000001101001' \
  --form my_file=\
  --form my_field=my_value
```

**After**

```
curl --request POST \
  --url http://localhost:3000/multipart \
  --header 'Content-Type: multipart/form-data' \
  --header 'content-type: multipart/form-data; boundary=---011000010111000001101001' \
  --form my_second_file=@/Users/jgiovaresco/Documents/AFTER.gif \
  --form my_first_file=@/Users/jgiovaresco/Documents/Untitled \
  --form 'a_simple_value=my value'
```

### DIFF
--- a/packages/insomnia-app/app/common/har.js
+++ b/packages/insomnia-app/app/common/har.js
@@ -489,6 +489,13 @@ function getRequestPostData(renderedRequest: RenderedRequest): HarPostData | voi
   let params = [];
   if (body.params) {
     params = body.params.map(param => {
+      if (param.type === 'file') {
+        return {
+          name: param.name,
+          fileName: param.fileName,
+        };
+      }
+
       return {
         name: param.name,
         value: param.value,
--- a/packages/insomnia-app/app/common/__tests__/har.test.js
+++ b/packages/insomnia-app/app/common/__tests__/har.test.js
@@ -372,4 +372,58 @@ describe('exportHarWithRequest()', () => {
       settingEncodeUrl: true,
     });
   });
+
+  it('export multipart request with file', async () => {
+    const workspace = await models.workspace.create();
+    const request = Object.assign(models.request.init(), {
+      _id: 'req_123',
+      parentId: workspace._id,
+      headers: [{ name: 'Content-Type', value: 'multipart/form-data' }],
+      parameters: [],
+      method: 'POST',
+      body: {
+        mimeType: 'multipart/form-data',
+        params: [
+          { name: 'a_file', value: '', fileName: '/tmp/my_file', type: 'file' },
+          { name: 'a_simple_field', value: 'a_simple_value' },
+          { name: 'a_second_file', value: '', fileName: '/tmp/my_file_2', type: 'file' },
+        ],
+      },
+      url: 'http://example.com/post',
+      authentication: {},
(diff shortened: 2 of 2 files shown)


nijikokun:
Awesome change, I assume this works for multi-file as well?

jgiovaresco:
> Awesome change, I assume this works for multi-file as well?

You are assuming right ;)
I've updated the unit test to be sure. 😅 

```
curl --request POST \
  --url http://localhost:3000/multipart \
  --header 'Content-Type: multipart/form-data' \
  --header 'content-type: multipart/form-data; boundary=---011000010111000001101001' \
  --form my_file=@/Users/jgiovaresco/Documents/Untitled \
  --form my_field=my_value \
  --form my_second_file=@/Users/jgiovaresco/Documents/AFTER.gif
```

reynolek (approved):
This change makes sense to me.

develohpanda (approved):
LGTM, sorry for the delay in getting to this!

packages/insomnia-app/app/common/har.js
```diff
@@ -489,6 +490,13 @@ function getRequestPostData(renderedRequest: RenderedRequest): HarPostData | voi
   let params = [];
   if (body.params) {
     params = body.params.map(param => {
+      if (param.type === 'file') {
+        return {
+          name: param.name,
+          value: param.fileName,
```
DMarby: While this now correctly includes the path in the generated curl command, it doesn't actually work correctly when you run it, as curl needs a file path to be prepended with `@` in order for it to read it.

If we set `fileName` instead of `value`, httpsnippet will do so:
https://github.com/Kong/httpsnippet/blob/dc08d8338a5e27caa3bdddb0131c3ed8eb92080c/src/targets/shell/curl.js#L48-L52

```suggestion
          fileName: param.fileName,
```

DMarby (changes requested):

packages/insomnia-app/app/common/har.js
jgiovaresco: Arf good catch I should have took a look at httpsnippet 🤦

DMarby (approved):
LGTM, thanks for fixing this @jgiovaresco!

PR #2742, opened 2020-10-19
fix: disable link when plugin has not modules

### LINKED ISSUE #2579: 404 On generated plugin documentation URL
**Describe the bug**

**To Reproduce**
1. Go to plugins
2. Generate a plugin
3. Click in the version number external link arrow indicator
4. See 404

**Expected behavior**
That link is disabled or we add a tool-tip on the arrow that acts as a subtle reminder saying "Published Documentation" so if the user gets a 404, they'll hopefully realize that it needs to be published first.

**Screenshots**
generate-plugin

martinpastore:
Closes #2579 

Captura de Pantalla 2020-10-19 a la(s) 17 41 50

I've validated by empty modules but let me know If this is not correct.

### DIFF
--- a/packages/insomnia-app/app/ui/components/settings/plugins.js
+++ b/packages/insomnia-app/app/ui/components/settings/plugins.js
@@ -197,9 +197,10 @@ class Plugins extends React.PureComponent<Props, State> {
 
     const base = /^insomnia-plugin-/.test(name) ? PLUGIN_HUB_BASE : NPM_PACKAGE_BASE;
     const link = path.join(base, name);
+    const disabled = Object.keys(plugin.module).length === 0;
 
     return (
-      <a className="space-left" href={link} title={link}>
+      <a className={`space-left ${disabled ? 'disabled' : ''}`} href={link} title={link}>
         <i className="fa fa-external-link-square" />
       </a>
     );
--- a/packages/insomnia-app/app/ui/components/settings/plugins.js
+++ b/packages/insomnia-app/app/ui/components/settings/plugins.js
@@ -197,10 +197,12 @@ class Plugins extends React.PureComponent<Props, State> {
 
     const base = /^insomnia-plugin-/.test(name) ? PLUGIN_HUB_BASE : NPM_PACKAGE_BASE;
     const link = path.join(base, name);
-    const disabled = Object.keys(plugin.module).length === 0;
+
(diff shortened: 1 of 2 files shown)


nijikokun:
From a product perspective, we should remove as it feels like there is an error, rather than it's not capable of linking.

martinpastore:
> From

Cool! Thanks for clarificate it. I'll remove it 👌

martinpastore:
@nijikokun now is not showing the link 👌

develohpanda (commented):
I'm not sure if this is the correct way to check whether the documentation link is valid or not, because it doesn't cover many bases.

`plugin.module` _will_ contain keys if the plugin was created locally and had  not been published yet. It would then, again, link to invalid documentation. It will also link to invalid documentation for private plugins.

There are a few options:
- Allow users to override the documentation link from the plugin [`package.json`](https://support.insomnia.rest/article/26-plugins#create-a-plugin).
- Update the website to show a friendlier "plugin not found" page, because the same issue will exist for private plugins that are not on NPM or the Plugin Hub.
- Have the app ping to determine whether documentation exists at the expected link (I don't like this approach)

@nijikokun please advise the desired solution from the above.

martinpastore:
> I'm not sure if this is the correct way to check whether the documentation link is valid or not, because it doesn't cover many bases.
> 
> `plugin.module` _will_ contain keys if the plugin was created locally and had not been published yet. It would then, again, link to invalid documentation. It will also link to invalid documentation for private plugins.
> 
> There are a few options:
> 
> * Allow users to override the documentation link from the plugin [`package.json`](https://support.insomnia.rest/article/26-plugins#create-a-plugin).
> * Update the website to show a friendlier "plugin not found" page, because the same issue will exist for private plugins that are not on NPM or the Plugin Hub.
> * Have the app ping to determine whether documentation exists at the expected link (I don't like this approach)
> 
> @nijikokun please advise the desired solution from the above.

Cool. I was wondering if that case applied for all the possible plugins. Thanks for confirm that

develohpanda:
I will close this for the time being. As it stands there is no reliable way to know for certain that a plugin _doesn't_ have documentation available at the link, for the reasons outlined above. In the future as we work through plugin improvements, we can certainly revisit this. Thank you for the PR 🙏🏽

Issue #1966, opened 2020-02-25
Documentation not updated on schema refresh


dobrynin:
**Describe the bug**
Refreshing the schema does not refresh the schema documentation.

**To Reproduce**
1. Click schema wrench button 
2. Click 'Show Documentation'
3. Observe the state of the schema as shown in the documentation.
4. Make a change in the schema that should change the docs
5. Click 'Refresh Schema'
6. open the documentation again and see that it has not changed appropriately.

**Expected behavior**
Documentation updates when schema changes.
****
**Desktop**
 - OS: macOS 10.14.6
 - Installation Method: downloaded from internet
 - App Version v7.1.1

ranjan-purbey:
May I take up this issue?

develohpanda:
> May I take up this issue?

Hi @ranjan-purbey, feel free to! Sorry about the late response 🤗 it's been a busy few weeks!

PR #2747, opened 2020-10-19
Reloading GraphQL docs on schema reload

antoine29:
#1966 states that the GraphQL docs are not being properly reloaded on a schema reload. Well with this PR the entire docs component would be reloaded on a schema reload.
schemaReload


develohpanda (approved):
LGTM.

With this fix, if someone is browsing the schema, and the schema changes, it will reset the user to the explorer root because a new instance of `GraphqlExporer` will be rendered - exactly what you see in the gif above.

I feel like this behavior is sensible, though, because the schema will have updated and the previous node the user was browsing might not exist anymore.

PR #2749, opened 2020-10-19
Adding deprecated warning icon/message in graphQL docs

### LINKED ISSUE #1831: Show deprecation warning and message in GraphQL schema explorer
- Insomnia Version: 7.0.5
- Operating System: Ubuntu 18.04

## Details

I noticed that the GraphQL editor has an awesome deprecation-warning built-in, but the deprecation warning is not shown in the schema explorer. I think that would be a nice edition - preferably with a big fat warning icon :)

antoine29:
This PR is adding a warning icon to every deprecated field shown in the GraphQL schema explorer, also the deprecating reason is being shown in the icon tooltip. It closes #1831
deprecatedWarn


nijikokun:
Nice work 🎉 

Is it possible to use the [exclamation triangle](https://fontawesome.com/v4.7.0/icon/exclamation-triangle) instead of a circle to better signify more of a warning to the end user?

develohpanda:
There's a `warning` icon in `insomnia-components` you might be able to use for it, will be consistent with other warnings in the code (such as the one you see in the GraphQL body in the screenshot above)

https://storybook.insomnia.rest/?path=/story/iconography-core-icons--reference

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-type.js
```diff
@@ -77,6 +78,8 @@ class GraphQLExplorerType extends React.PureComponent<Props> {
 
     // $FlowFixMe
     const fields = type.getFields();
+    const deprecatedFieldDescription = field =>
+      `The field ${field.name} is deprecated. ${field.deprecationReason}`;
```
develohpanda: 1. Please could you move the creation of this function out of the render flow, so that the function is not recreated on every render?
2. Please add speech marks around the field name (like `The field "currentUser" is deprecated...`) to make it a little clearer.

develohpanda (commented):
Nice work, thanks for the PR!

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-type.js
antoine29: Hey guys. Thank you for the feedback :D
- I replaced the circle warning icon with the one from insomnia-components.
- The deprecationDescription function was moved out from the render function.
- Quote marks were added around the deprecated field name.
fixedDeprecatedWarn

develohpanda (approved):
LGTM!

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-type.js
develohpanda: Nice one. Do you know why the spacing looks a little off with the icon? I wonder if there is a rogue style being applied, adding padding/margin or something 🤔

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-type.js
```diff
@@ -27,6 +29,9 @@ class GraphQLExplorerType extends React.PureComponent<Props> {
     onNavigateField(field);
   }
 
+  deprecatedFieldDescription = field =>
```
DMarby: Not sure what our style guide is, but I don't think we have any other functions defined in a class using this syntax elsewhere in the codebase?

Not sure if our linter should've caught this, what's your thoughts @develohpanda? Imo it's pretty difficult to grok.

develohpanda: Oh good catch, I thought it was moved to _outside_ the class entirely.

I made a similar note [here](https://github.com/Kong/insomnia/pull/2712#discussion_r509881183), to use regular function syntax for classes methods, rather than arrow functions.

I'm _pretty sure_ there is an ES Lint rule for this, which we should enable as long as it covers the right scope

antoine29: The function is pretty simple since is only concatenating the deprecation description string. Don't you think to get rid of this function and just set the concatenated string in the Tooltip message could be a simpler idea?

develohpanda: That works as well 😊

antoine29: got it :D

nijikokun (approved):
Nice work 🎉

DMarby (approved):
Looks great, just tried it out against one of our APIs 👍 

Thank you for taking the time to implement this!

PR #2753, opened 2020-10-20
Beta sync fixes

DMarby:
* Require a user to have sync setup, and to have pushed it to the server to share, instead of having the share dialog create projects just because it was opened.
* Fixes pulling a workspace with no default branch, rather then erroring out we create a workspace with the correct ID and an empty branch.
* Make the sync merge dialog work with more then a single merge conflict.
* Fix workspaces getting deleted when switching to a default branch that's empty, when previously on another branch that has commits.
* Adds logging to failed graphql sync queries.

Fixes INS-235

### DIFF
--- a/packages/insomnia-app/app/sync/vcs/index.js
+++ b/packages/insomnia-app/app/sync/vcs/index.js
@@ -1128,32 +1128,25 @@ export default class VCS {
   }
 
   async _queryProjectTeams(): Promise<Array<Team>> {
-    const run = async () => {
-      const { project } = await this._runGraphQL(
-        `
-        query ($id: ID!) {
-          project(id: $id) {
-            teams {
-              id
-              name
-            }
+    const { project } = await this._runGraphQL(
+      `
+      query ($id: ID!) {
+        project(id: $id) {
+          teams {
+            id
+            name
           }
         }
-      `,
-        {
-          id: this._projectId(),
-        },
-        'project.teams',
-      );
-      return project;
-    };
+      }
+    `,
+      {
+        id: this._projectId(),
+      },
+      'project.teams',
+    );
 
-    let project = await run();
-
-    // Retry once if project doesn't exist yet
-    if (project === null) {
-      await this._getOrCreateRemoteProject();
-      project = await run();
+    if (!project) {
+      throw new Error('Please push the workspace to be able to share it');
     }
 
     return project.teams;
--- a/packages/insomnia-app/app/sync/vcs/index.js
+++ b/packages/insomnia-app/app/sync/vcs/index.js
@@ -716,6 +716,7 @@ export default class VCS {
     const { data, errors } = await fetch.post('/graphql?' + name, { query, variables }, sessionId);
 
     if (errors && errors.length) {
+      console.log(`[sync] Failed to query ${name}`, errors);
       throw new Error(`Failed to query ${name}`);
     }
 
--- a/packages/insomnia-app/app/sync/vcs/index.js
+++ b/packages/insomnia-app/app/sync/vcs/index.js
@@ -727,7 +727,7 @@ export default class VCS {
       `
           query ($projectId: ID!, $ids: [ID!]!) {
             blobsMissing(project: $projectId, ids: $ids) {
-              missing 
+              missing
             }
           }
         `,
@@ -766,7 +766,7 @@ export default class VCS {
     await this._runGraphQL(
       `
       mutation ($projectId: ID!, $branch: String!) {
-        branchRemove(project: $projectId, name: $branch) 
+        branchRemove(project: $projectId, name: $branch)
       }`,
       {
         projectId: this._projectId(),
--- a/packages/insomnia-app/app/sync/vcs/index.js
+++ b/packages/insomnia-app/app/sync/vcs/index.js
@@ -367,8 +367,8 @@ export default class VCS {
     };
   }
 
-  async getHistoryCount(): Promise<number> {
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/sync/vcs/index.js
```diff
@@ -1127,32 +1128,25 @@ export default class VCS {
   }
 
   async _queryProjectTeams(): Promise<Array<Team>> {
-    const run = async () => {
```
develohpanda, line 1130: Why was this a function before 🤔 strange

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
```diff
@@ -256,23 +257,53 @@ class SyncDropdown extends React.PureComponent<Props, State> {
...
 
     await this.refreshMainAttributes({ loadingProjectPull: false });
   }
 
   async _handleSwitchBranch(branch: string) {
     const { vcs, syncItems } = this.props;
+
     try {
       const delta = await vcs.checkout(syncItems, branch);
+      const defaultBranchName = 'master';
```
develohpanda: Could you please set this once at the top of the file (or even in `constants.js`)? Want to avoid multiple instances of what is, effectively, a constant. It's duplicated a few times in this PR/previously.

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
```diff
@@ -256,23 +257,53 @@ class SyncDropdown extends React.PureComponent<Props, State> {
...
+      if (branch === defaultBranchName) {
+        const { historyCount } = this.state;
+        const defaultBranchHistoryCount = await vcs.getHistoryCount(defaultBranchName);
+
+        // If the default branch has no snapshots, but the current branch does
+        // It will result in the workspace getting deleted
+        // So we filter out the workspace from the delta to prevent this
+        if (!defaultBranchHistoryCount && historyCount) {
+          delta.remove = delta.remove.filter(e => e.type !== models.workspace.type);
+        }
```
develohpanda, line 304: 👍

packages/insomnia-app/app/ui/components/dropdowns/workspace-dropdown.js
```diff
@@ -152,15 +152,32 @@ class WorkspaceDropdown extends React.PureComponent<Props, State> {
...
+        await db.upsert(workspace);
+      } else {
+        await newVCS.pull([]); // There won't be any existing docs since it's a new pull
+
+        const flushId = await db.bufferChanges();
+        for (const doc of await newVCS.allDocuments()) {
+          await db.upsert(doc);
+        }
+        await db.flushChanges(flushId);
       }
```
develohpanda, line 179: This entire code-snippet is duplicated from `sync-dropdown.js`. Is it worth extracting and reusing? It looks fairly important.

packages/insomnia-app/app/sync/vcs/index.js
```diff
@@ -1127,32 +1128,25 @@ export default class VCS {
...
+      },
+      'project.teams',
+    );
 
-    // Retry once if project doesn't exist yet
-    if (project === null) {
-      await this._getOrCreateRemoteProject();
-      project = await run();
+    if (!project) {
+      throw new Error('Please push the workspace to be able to share it');
```
develohpanda, line 1149: Is the error message suitable for what the method/user is doing? Which action has the user just taken that could prompt this error?

Not too familiar with sync logic so let me know if it _is_ suitable, too. 😄

develohpanda (changes requested):
Generally looks good, I have a couple of comments

packages/insomnia-app/app/sync/vcs/index.js
DMarby: It was a function since it was ran twice within this function

develohpanda: Ah you're right, the `!project` condition body has now changed. Sweet.

packages/insomnia-app/app/ui/components/dropdowns/workspace-dropdown.js
DMarby: Since the entire pull logic was already duplicated between the sync and the workspace dropdown, I opted to keep it that way for now.

Since the end goal is to replace this entire sync system, I'm trying to make as minimal changes as possible to just get things to a working state, rather then spending time cleaning up the code.

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
DMarby: Yep yep, moved it to the top.
Realized after pushing this PR that there was now two instances of it in the same file from the different fixes 😅 

We're getting rid of this shortly when changing the default branch name (since it shouldn't be a constant in the client), and it's already littered around the codebase, so I don't think it belongs in `constants.js`.

develohpanda (approved):

packages/insomnia-app/app/sync/vcs/index.js
DMarby: Yep, this method is only called by the share modal, and the error is propagated up to said modal:
image
If a user enables sync for a workspace (presses setup sync in the bottom sidebar), but doesn't push (so it hasn't been created on the server side), and try to share it with a team, they get this error.

sonicyeti (approved):
Long overdue, great work!

PR #2756, opened 2020-10-21
Modal prompt to select protofile

develohpanda:
This PR adds the ability to show protofiles uploaded for a particular workspace (but uses dummy data, because the functionality to _upload_ a protofile comes in a subsequent PR).

Logic to handle an empty state:
image

Save button is disabled when no item is selected:
image

And enabled once an item _is_ selected:
image

The requestName and selected protofileId is available in the same place after clicking save
image

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/proto-files-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/proto-files-modal.js
@@ -21,7 +21,7 @@ type State = {|
 
 type ProtoFilesModalOptions = {|
   preselectProtoFileId?: string,
-  onSave?: string => Promise<void>,
+  onSave: string => Promise<void>,
 |};
 
 const INITIAL_STATE: State = {
@@ -46,7 +46,7 @@ class ProtoFilesModal extends React.PureComponent<Props, State> {
   }
 
   async show(options: ProtoFilesModalOptions) {
-    this.onSave = options.onSave || null;
+    this.onSave = options.onSave;
     this.setState({ ...INITIAL_STATE });
 
     this.modal && this.modal.show();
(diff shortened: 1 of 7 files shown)


packages/insomnia-app/app/ui/components/modals/proto-files-modal.js
```diff
@@ -0,0 +1,173 @@
...
+
+const SelectableListItem: React.PureComponent<{ selected?: boolean }> = styled(ListGroupItem)`
+  &:hover {
+    background-color: var(--hl-sm) !important;
+  }
+
+  background-color: ${props => props.selected && 'var(--hl-sm) !important'};
+`;
+
+const ProtoFileListItem = (props: {
```
develohpanda: Extract this component (and a new component - `ProtoFileList`) into `ui`.

packages/insomnia-app/app/ui/components/modals/proto-files-modal.js
```diff
@@ -0,0 +1,121 @@
...
+|};
+
+type State = {|
+  protoFiles: Array<ProtoFile>,
+  selectedProtoFileId: string,
+|};
+
+type ProtoFilesModalOptions = {|
+  preselectProtoFileId?: string,
+  onSave?: string => Promise<void>,
```
DMarby: I don't think that `onSave` should be an optional parameter here.
There's currently no case where we wouldn't want to persist a change in this modal, and not passing this would lead to the Save button just not doing anything, which is confusing.

DMarby (changes requested):

DMarby (approved):
Awesome 🎉

PR #2758, opened 2020-10-21
Release branch for 2020.4.2

DMarby:


DMarby:
Just created this to trigger test runs, this shouldn't be merged directly as it's cherry-picking certain commits from `develop`. Instead, we should cherrypick only the version bump back to `develop` once this has been released.

PR #2763, opened 2020-10-22
Allow for renaming of a protofile

develohpanda:
I introduced a prop to `Editable` to prevent a blank value from being saved. This is useful in many other places (such as unit test names and test suite names as well).

### DIFF
--- a/packages/insomnia-app/app/ui/components/base/editable.js
+++ b/packages/insomnia-app/app/ui/components/base/editable.js
@@ -5,7 +5,7 @@ import KeydownBinder from '../keydown-binder';
 
 export const shouldSave = (oldValue, newValue, preventBlank) => {
   // Should not save if length = 0 and we want to prevent blank
-  if (preventBlank && newValue.length === 0) {
+  if (preventBlank && !newValue.length) {
     return false;
   }
 
@@ -51,10 +51,10 @@ class Editable extends PureComponent {
   }
 
   _handleEditEnd() {
-    const { originalValue, preventBlank } = this.props;
+    const originalValue = this.props.value;
     const newValue = this._input.value.trim();
 
-    if (shouldSave(originalValue, newValue, preventBlank)) {
+    if (shouldSave(originalValue, newValue, this.props.preventBlank)) {
       // Don't run onSubmit for values that haven't been changed
       this.props.onSubmit(newValue);
     }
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/ui/components/base/editable.js
```diff
@@ -3,6 +3,21 @@ import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import KeydownBinder from '../keydown-binder';
 
+const shouldSave = (oldValue, newValue, preventBlank) => {
```
develohpanda: Add unit test

packages/insomnia-app/app/ui/components/base/editable.js
```diff
@@ -3,6 +3,21 @@ import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import KeydownBinder from '../keydown-binder';
 
+export const shouldSave = (oldValue, newValue, preventBlank) => {
+  // Should not save if length = 0 and we want to prevent blank
+  if (preventBlank && newValue.length === 0) {
```
DMarby: Not a blocker, just curious, I know we have a mix of both in the codebase, but generally I'd consider `!newValue.length` to be more idiomatic in JS, is there a particular reason for why you're opting for an explicit `=== 0` instead?

packages/insomnia-app/app/ui/components/base/editable.js
```diff
@@ -36,11 +51,12 @@ class Editable extends PureComponent {
   }
 
   _handleEditEnd() {
-    const value = this._input.value.trim();
+    const { originalValue, preventBlank } = this.props;
```
DMarby: `originalValue` is always going to be null here since there's no property named `originalValue` in props

DMarby (changes requested):
Looks good overall, one minor bug in how we're comparing the previous value to the current one in `Editable`

packages/insomnia-app/app/ui/components/base/editable.js
develohpanda: Oops 🙈 you're right, I meant to alias it to make it clearer from the previous implementation.

packages/insomnia-app/app/ui/components/base/editable.js
develohpanda: Hmmm no particular reason, I can update it. I wonder if there is an ES Lint rule for this code style.

I use both, and both have been used in the codebase.

DMarby: Fine by me to leave it as-is as well, mostly just curious :)

DMarby (approved):

PR #2765, opened 2020-10-22
Fix fetching notifications

DMarby:
We had a slight regression in `develop` from #2676 which broke fetching notifications, as it broke the imports of some of the constants, causing the following error:
```
Failed to fetch user notifications TypeError: Cannot read property 'getAppPlatform' of undefined
```


sonicyeti (approved):
Nice, good catch!

develohpanda (approved):
🔥

PR #2770, opened 2020-10-23
Prevent GraphQL non breaking space

thewheat:
From https://github.com/Kong/insomnia/pull/2704/files#r505491859 it was found that the code to prevent GraphQL non breaking spaces (that cause a syntax error) is not working

This PR fixes the behaviour

**Demo**
- Bottom Insomnia is latest version with the bug
- Top Insomnia is the version with my PR code
- Demo copies invalid text and shows it fails in the latest version but passes with the code change

graphql-nbsp-fixed


packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -804,16 +804,16 @@ class CodeEditor extends React.Component {
       change.update(change.from, change.to, [text]);
     }
 
+    // Don't allow non-breaking spaces because they break the GraphQL syntax
+    if (doc.options.mode === 'graphql' && change.text && change.text.length > 0) {
```
thewheat: `change.text` is an array of the text that is inputted with each element being a line of text that was entered 
- most cases would just be a single line so `change.text.length` == 1
- if `Enter` is pressed or text with newline is pasted then `change.text.length` will be greater than 1

gatzjames:
Merged with latest develop and added more irregular characters that might break the interface.

graphql-whitespace-gif

dimitropoulos (approved):
looked good.  I enjoyed learning of this rick and morty api along the way, haha.

image

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -990,7 +991,7 @@ class CodeEditor extends Component<Props, State> {
     }
   }
 
-  _codemirrorValueBeforeChange(doc, change) {
+  _codemirrorValueBeforeChange(doc: CodeMirror.Editor, change: CodeMirror.EditorChangeCancellable) {
```
develohpanda, line 994: @gatzjames I believe we should be using `CodeMirror.EditorFromTextArea` instead of the generic `CodeMirror.Editor`; that's the only type of editor used AFAIK

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -999,19 +1000,21 @@ class CodeEditor extends Component<Props, State> {
...
       // Convert all whitespace to spaces
-      change.update(change.from, change.to, [text]);
+      change.update?.(change.from, change.to, [text]);
+    }
+
+    // Don't allow non-breaking spaces because they break the GraphQL syntax
+    if (doc.getOption('mode') === 'graphql') {
+      const text = change.text.map(normalizeIrregularWhitespace);
+
+      change.update?.(change.from, change.to, text);
```
develohpanda, line 1010: I think we shouldn't invoke `update` if `text.length === 0` (ie. no lines)?

packages/insomnia-app/app/ui/components/codemirror/normalizeIrregularWhitespace.ts
```diff
@@ -0,0 +1,33 @@
+// List of irregular whitespace characters adopted from https://eslint.org/docs/rules/no-irregular-whitespace#rule-details
```
develohpanda, line 1: Nice fix here 🙂

PR #2777, opened 2020-10-24
Setting same sidebar item sorting for the Edit tag request dropdown

### LINKED ISSUE #2745: Sort request list in dropdown of "Response -> Body attribute" tag editor
When I want to insert a tag using "Response -> Body attribute", in the list of all requests in the dropdown, the requests are not sorted like in the sidebar.

I'd like the requests in the dropdown list to be sorted like in the sidebar.

I often sort the requests as to group them by functionality, and it's hard to navigate through 30+ requests to find the one I need.  

Here's how the dropdown list looks like:
image

Here's the list from the sidebar. Notice that that the "/config" request appears in the first place, which is different from the list displayed in the dropdown.
image

antoine29:
This PR is adding a method to get the same item sorting as is in the request left sideBar, for the tag requests dropdown. For the item sorting I'm using the same sorting criteria found in this [selector code](https://github.com/Kong/insomnia/blob/af6a9b6a78a46402bc254e709b31d8dfa9e4b314/packages/insomnia-app/app/ui/redux/selectors.js#L165-L175).
And as you can see if the sidebar item sorting is rearranged, the change will be reflected in the tag requests dropdown too.
tagReqsSorted
This would close #2745

### DIFF
--- a/packages/insomnia-app/app/ui/components/templating/tag-editor.js
+++ b/packages/insomnia-app/app/ui/components/templating/tag-editor.js
@@ -472,16 +472,11 @@ class TagEditor extends React.PureComponent<Props, State> {
     return prefix;
   }
 
-  sortDocs(allDocs, parentId: string) {
+  sortDocs(requests: Array<any>, requestGroups: Array<any>, parentId: string) {
     let sortedDocs = [];
-    let requests = allDocs[models.request.type];
-    let requestGroups = allDocs[models.requestGroup.type];
-    if (!requests || !requestGroups) return [];
-
-    requests = requests.filter(request => request.parentId === parentId);
-    requestGroups = requestGroups.filter(requestGroup => requestGroup.parentId === parentId);
     requests
-      .concat(requestGroups)
+      .filter(request => request.parentId === parentId)
+      .concat(requestGroups.filter(requestGroup => requestGroup.parentId === parentId))
       .sort((a, b) => {
         if (a.metaSortKey === b.metaSortKey) {
           return a._id > b._id ? -1 : 1;
@@ -492,7 +487,7 @@ class TagEditor extends React.PureComponent<Props, State> {
       .map(doc => {
         if (doc.type === models.request.type) sortedDocs.push(doc);
         if (doc.type === models.requestGroup.type)
-          sortedDocs = sortedDocs.concat(this.sortDocs(allDocs, doc._id));
+          sortedDocs = sortedDocs.concat(this.sortDocs(requests, requestGroups, doc._id));
       });
 
     return sortedDocs;
@@ -500,8 +495,10 @@ class TagEditor extends React.PureComponent<Props, State> {
 
   renderArgModel(value: string, modelType: string) {
     const { allDocs, loadingDocs } = this.state;
-    const sortedDocs = this.sortDocs(allDocs, this.props.workspace._id);
-    const docs = sortedDocs !== [] ? sortedDocs : allDocs[modelType] || [];
+    const requests = allDocs[models.request.type] || [];
+    const requestGroups = allDocs[models.requestGroup.type] || [];
+    const sortedReqs = this.sortDocs(requests, requestGroups, this.props.workspace._id);
+    const docs = modelType === models.request.type ? sortedReqs : allDocs[modelType] || [];
     const id = value || 'n/a';
 
     if (loadingDocs) {
--- a/packages/insomnia-app/app/ui/components/templating/tag-editor.js
+++ b/packages/insomnia-app/app/ui/components/templating/tag-editor.js
@@ -109,6 +109,23 @@ class TagEditor extends React.PureComponent<Props, State> {
     );
   }
 
+  _sortRequests(requests: Array<any>, parentId: string) {
+    let sortedReqs = [];
+    requests
+      .filter(request => request.parentId === parentId)
+      .sort((a, b) => {
+        if (a.metaSortKey === b.metaSortKey) return a._id > b._id ? -1 : 1;
+        else return a.metaSortKey < b.metaSortKey ? -1 : 1;
+      })
+      .map(request => {
+        if (request.type === models.request.type) sortedReqs.push(request);
+        if (request.type === models.requestGroup.type)
+          sortedReqs = sortedReqs.concat(this._sortRequests(requests, request._id));
+      });
+
+    return sortedReqs;
+  }
+
   async _refreshModels(workspace: Workspace) {
     this.setState({ loadingDocs: true });
 
@@ -121,6 +138,10 @@ class TagEditor extends React.PureComponent<Props, State> {
       allDocs[doc.type].push(doc);
     }
 
+    const requests = allDocs[models.request.type] || [];
+    const requestGroups = allDocs[models.requestGroup.type] || [];
+    const sortedReqs = this._sortRequests(requests.concat(requestGroups), this.props.workspace._id);
+    allDocs[models.request.type] = sortedReqs;
     this.setState({
       allDocs,
       loadingDocs: false,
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -452,9 +452,33 @@ class TagEditor extends React.PureComponent<Props, State> {
     return prefix;
   }
 
+  sortDocs(requests: Array<any>, requestGroups: Array<any>, parentId: string) {
+    let sortedDocs = [];
+    requests
+      .filter(request => request.parentId === parentId)
+      .concat(requestGroups.filter(requestGroup => requestGroup.parentId === parentId))
```
cobwebsonsale: *nit* since concat happens every time and both the arrays are filtered in the same way, we could send a combined array of `docs` instead.

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -452,9 +452,33 @@ class TagEditor extends React.PureComponent<Props, State> {
...
+
+    return sortedDocs;
+  }
+
   renderArgModel(value: string, modelType: string) {
     const { allDocs, loadingDocs } = this.state;
-    const docs = allDocs[modelType] || [];
+    const requests = allDocs[models.request.type] || [];
+    const requestGroups = allDocs[models.requestGroup.type] || [];
+    const sortedReqs = this.sortDocs(requests, requestGroups, this.props.workspace._id);
```
cobwebsonsale: Sorting the requests in render method causes this to get triggered multiple times. `_refreshModels` updates the `allDocs` state variable. That may be a better place to update the order. Something like:
```
allDocs[models.request.type] = this.sortDocs(allDocs[models.request.type] || [], allDocs[models.requestGroup.type] || [], this.props.workspace._id);
```

`_refreshModels` is called just once per times popup is shown.
Screenshot 2020-10-27 at 7 46 45 PM

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -452,9 +452,33 @@ class TagEditor extends React.PureComponent<Props, State> {
...
+        if (a.metaSortKey === b.metaSortKey) {
+          return a._id > b._id ? -1 : 1;
+        } else {
+          return a.metaSortKey < b.metaSortKey ? -1 : 1;
+        }
+      })
+      .map(doc => {
+        if (doc.type === models.request.type) sortedDocs.push(doc);
+        if (doc.type === models.requestGroup.type)
+          sortedDocs = sortedDocs.concat(this.sortDocs(requests, requestGroups, doc._id));
```
cobwebsonsale: Yep, this does it. Nice! :)

packages/insomnia-app/app/ui/components/templating/tag-editor.js
antoine29: You're right. So I moved the sorting logic into the refresh method

packages/insomnia-app/app/ui/components/templating/tag-editor.js
antoine29: thank you :D

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -103,6 +103,23 @@ class TagEditor extends React.PureComponent<Props, State> {
     );
   }
 
+  _sortRequests(requests: Array<any>, parentId: string) {
```
cobwebsonsale: You can use types to specify that this is an array of Requests and RequestGroups. Like suggested [here](https://github.com/Kong/insomnia/pull/2738#discussion_r513076716).

antoine29: types added :D

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -103,6 +103,23 @@ class TagEditor extends React.PureComponent<Props, State> {
     );
   }
 
+  _sortRequests(requests: Array<models.request.type | models.requestGroup.type>, parentId: string) {
```
develohpanda: You can import the type directly and use it. `models.request.type` is actually just a string 😆 

```js
import type { Request } from '../../../models/request';
import type { RequestGroup } from '../../../models/requestGroup';
...
_sortRequests(models: Array<Request | RequestGroup>, parentId: string) { 
```

Also, it seems that using `models` instead of `requests` seems more appropriate here for the parameter, since an item in the array can actually be a `Request` or a `RequestGroup`. By extension, when iterating through the array it would also change to `models.filter(model => ...)`

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -103,6 +103,23 @@ class TagEditor extends React.PureComponent<Props, State> {
...
+    let sortedReqs = [];
+    requests
+      .filter(request => request.parentId === parentId)
+      .sort((a, b) => {
+        if (a.metaSortKey === b.metaSortKey) return a._id > b._id ? -1 : 1;
+        else return a.metaSortKey < b.metaSortKey ? -1 : 1;
+      })
+      .map(request => {
```
develohpanda: Nit picking, but this should likely be a `.forEach` rather than `.map` since we don't use the result for anything.

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -103,6 +103,23 @@ class TagEditor extends React.PureComponent<Props, State> {
...
+        if (request.type === models.request.type) sortedReqs.push(request);
+        if (request.type === models.requestGroup.type)
```
develohpanda: I've created some helpers in `packages/insomnia-app/app/models/helpers/is-model.js` to make these a little less verbose. It would then change to:

```js
if (isRequest(model)) { }
if (isRequestGroup(model)) { }
```

packages/insomnia-app/app/ui/components/templating/tag-editor.js
```diff
@@ -103,6 +103,23 @@ class TagEditor extends React.PureComponent<Props, State> {
...
   }
 
+  _sortRequests(requests: Array<models.request.type | models.requestGroup.type>, parentId: string) {
+    let sortedReqs = [];
+    requests
+      .filter(request => request.parentId === parentId)
+      .sort((a, b) => {
+        if (a.metaSortKey === b.metaSortKey) return a._id > b._id ? -1 : 1;
+        else return a.metaSortKey < b.metaSortKey ? -1 : 1;
+      })
```
develohpanda: #2738 is almost complete and it creates a `metaSortKeySort` helper, linked below. Ideally that is used here, so once that PR is merged in we can update this.

https://github.com/cobwebsonsale/insomnia/blob/bca044f74d8bf437b6775ac5a9cf4c9f5f52ba3d/packages/insomnia-app/app/common/sorting.js#L92-L98

Then, this code would just become:
```suggestion
      .sort(metaSortKeySort)
```

develohpanda (changes requested):
Overall looks good, I've got some minor suggestions. 🙂

Sorry about the delay in getting to this, I've been really bogged down for a while!

packages/insomnia-app/app/ui/components/templating/tag-editor.js
develohpanda: 😆 we picked up on the same thing

https://github.com/Kong/insomnia/pull/2777#discussion_r531306885

packages/insomnia-app/app/ui/components/templating/tag-editor.js
antoine29: Sorry for the delay, l lost sight of this PR hehe. Anyways I added the required changes :D

develohpanda (approved):
Nice! Looks awesome now 👏🏽 Thanks for working through it

DMarby (approved):

PR #2784, opened 2020-10-27
Add GrpcRequest and GrpcRequestMeta database models

develohpanda:
What is says on the tin.

The `GrpcRequestMeta` model is a reduced version of the standard `RequestMeta`.

### DIFF
--- a/packages/insomnia-app/app/models/__tests__/grpc-request.test.js
+++ b/packages/insomnia-app/app/models/__tests__/grpc-request.test.js
@@ -51,7 +51,7 @@ describe('create()', () => {
   it('fails when missing parentId', async () => {
     Date.now = jest.fn().mockReturnValue(1478795580200);
     expect(() => models.grpcRequest.create({ name: 'no parentId' })).toThrow(
-      'New gRPC Request missing `parentId`',
+      'New GrpcRequest missing `parentId`',
     );
   });
 });
--- a/packages/insomnia-app/app/models/__tests__/grpc-request.test.js
+++ b/packages/insomnia-app/app/models/__tests__/grpc-request.test.js
@@ -1,0 +1,57 @@
+import * as models from '../index';
+import { globalBeforeEach } from '../../__jest__/before-each';
+
+describe('init()', () => {
+  beforeEach(globalBeforeEach);
+  it('contains all required fields', async () => {
+    Date.now = jest.fn().mockReturnValue(1478795580200);
+
+    expect(models.grpcRequest.init()).toEqual({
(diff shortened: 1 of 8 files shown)


packages/insomnia-app/app/models/__tests__/grpc-request.test.js
```diff
@@ -0,0 +1,57 @@
+import * as models from '../index';
+import { globalBeforeEach } from '../../__jest__/before-each';
+
+describe('init()', () => {
+  beforeEach(globalBeforeEach);
+  it('contains all required fields', async () => {
+    Date.now = jest.fn().mockReturnValue(1478795580200);
```
jgiovaresco, line 7: Wouldn't we need to keep the original implementation of `Date.now` to reset it after tests?

develohpanda: Jest is configured to `resetModules` between each test, so I don't believe it is necessary.

jgiovaresco: The [doc](https://jestjs.io/docs/en/jest-object#jestresetmodules) is saying 

> Resets the module registry - the cache of all required modules. This is useful to isolate modules where local state might conflict between tests.

If I understand correctly, it is related to the mocking module feature (`jest.mock(moduleName, factory, options)`).

I did a simple test case:

```javascript
it('fake date', () => {
  Date.now = jest.fn().mockReturnValue(1478000000000);
        
  console.log('1', Date.now())
        
  jest.resetModules()
        
  console.log('2', Date.now())
});
```

It produces the following:

```
  console.log
    1 1478000000000

  console.log
    2 1478000000000
```

If I didn't make a mistake, the `resetModules` does not reset the original `Date.now` implementation. 
I agree it is not a really big deal because if we want to test something with a date, we will start by mocking the `Date.now` function. 

However, as I learned something, I wanted to share it with you :)

develohpanda: Noted, I'll investigate further when I find some time.

The tests in this PR are copy-paste-edit from other model tests, so it would be good to fix all of them in one go.

DMarby (approved):

packages/insomnia-app/app/models/__tests__/grpc-request.test.js
develohpanda: Add a ticket to come back and address this: INS-262. (Internal link so you won't be able to see it, but just for reference!)

PR #2785, opened 2020-10-27
Save a GrpcRequest after protofile selection

develohpanda:
I would like to add some tests for this logic but it will need a decent chunk of refactoring or setting up a UI testing library. For the sake of speed and because this doesn't contain too much complex business logic, I have not added any.

Fixes INS-198
Fixes INS-261

### DIFF
--- a/packages/insomnia-app/app/models/grpc-request-meta.js
+++ b/packages/insomnia-app/app/models/grpc-request-meta.js
@@ -1,6 +1,7 @@
 // @flow
 import * as db from '../common/database';
 import type { BaseModel } from './index';
+import { prefix as grpcRequestPrefix } from './grpc-request';
 
 export const name = 'gRPC Request Meta';
 export const type = 'GrpcRequestMeta';
@@ -29,6 +30,10 @@ export function migrate(doc: GrpcRequestMeta): GrpcRequestMeta {
 export function create(patch: $Shape<GrpcRequestMeta> = {}): Promise<GrpcRequestMeta> {
   if (!patch.parentId) {
     throw new Error('New GrpcRequestMeta missing `parentId`');
+  }
+
+  if (!patch.parentId.startsWith(`${grpcRequestPrefix}_`)) {
+    throw new Error('Expected the parent of GrpcRequestMeta to be a GrpcRequest');
   }
 
   return db.docCreate(type, patch);
--- a/packages/insomnia-app/app/models/grpc-request-meta.js
+++ b/packages/insomnia-app/app/models/grpc-request-meta.js
@@ -10,6 +10,7 @@ export const canSync = false;
 
 type BaseGrpcRequestMeta = {
   pinned: boolean,
+  lastActive: number,
 };
 
 export type GrpcRequestMeta = BaseModel & BaseGrpcRequestMeta;
@@ -17,6 +18,7 @@ export type GrpcRequestMeta = BaseModel & BaseGrpcRequestMeta;
 export function init() {
   return {
     pinned: false,
+    lastActive: 0,
   };
 }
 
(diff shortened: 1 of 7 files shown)


packages/insomnia-app/app/ui/components/proto-file/proto-file-list.js
```diff
@@ -27,7 +27,7 @@ const ProtoFileList = ({
     {!protoFiles.length && <ListGroupItem>No proto files exist for this workspace</ListGroupItem>}
     {protoFiles.map(p => (
       <ProtoFileListItem
-        key={p.id}
+        key={p._id}
```
develohpanda, line 30: React was showing warnings in console because the keys were not unique. Flow/TS would have been able to help here, but I digress. 😢

sonicyeti (approved):

DMarby (approved):
Nice, seems to be getting saved correctly to disk:
image

DMarby (changes requested):
After some further testing when reviewing #2786, it seems like when a request is saved, instead of GrpcRequestMeta being created and attached to it, a RequestMeta is currently being saved and attached to it:

```
{"_id":"greq_8c07749763d84a46bb6eb64d70f838a0","type":"GrpcRequest","parentId":"wrk_d5966bce9e354f7cbb4873232a858549","modified":1603819339726,"created":1603819339726,"url":"","name":"gRPC 1","protoFileId":"pf_f8628adcae6948319fdaf097c83e6b8f","protoServiceName":"","protoMethodName":"","body":{},"metaSortKey":-1603819339726,"idPrivate":false}

{"_id":"reqm_2102efe286434e4f870c0f30943a0e57","type":"RequestMeta","parentId":"greq_8c07749763d84a46bb6eb64d70f838a0","modified":1603819339772,"created":1603819339772,"previewMode":"friendly","responseFilter":"","responseFilterHistory":[],"activeResponseId":null,"savedRequestBody":{},"pinned":false,"lastActive":1603819339752,"downloadPath":null}
```

I think it makes sense for that to be remedied in this PR, before it is merged.

develohpanda:
That doesn't seem right, I'll investigate what is creating the `requestMeta` object.

`requestMeta` is never created explicitly, it's just created when needed by the UI, and `GrpcRequest` doesn't need it for anything in this PR (because there is no UI to front the database model), so unfortunately I don't think it will be fixed here, even though logically it seems like it should be.

I will investigate adding in a safety net, though, to ensure the `request <- requestMeta` and `grpcRequest <- grpcRequestMeta` relationship is always correct.

develohpanda:
> That doesn't seem right, I'll investigate what is creating the requestMeta object.

On successful creation of a request, that request is activated and a `lastActive` time is stored in the meta object. I've updated the how that method behaves, and added `lastActive` as a field to `GrpcRequestMeta` as well. `lastActive` and `pinned` are the only fields needed.

Something tells me we can probably just use `RequestMeta` for both, but there are fields on `RequestMeta` that don't/won't apply to gRPC.

@DMarby do you have any strong opinions on creating `GrpcRequestMeta`, or just using `RequestMeta`? My preference is to keep it separate even though they are similar.

develohpanda:
`GrpcRequestMeta` is now correctly linked, and `RequestMeta` is not created.

image

packages/insomnia-app/app/models/grpc-request-meta.js
```diff
@@ -10,13 +11,15 @@ export const canSync = false;
 
 type BaseGrpcRequestMeta = {
   pinned: boolean,
+  lastActive: number,
```
develohpanda, line 14: This field is used by the `ctrl + tab` request switcher so I decided to include it here instead of changing too much business logic. INS-263

packages/insomnia-app/app/ui/components/modals/request-create-modal.js
```diff
@@ -56,12 +56,14 @@ class RequestCreateModal extends PureComponent {
...
-          const createdRequestId = 'gr_123';
-          this._onComplete(createdRequestId);
+        onSave: async (protoFileId: string) => {
+          const createdRequest = await models.grpcRequest.create({
+            parentId,
+            name: requestName,
+            protoFileId,
+          });
+
+          this._onComplete(createdRequest._id);
```
develohpanda, line 66: This callback boils up to `App._handleSetActiveRequest`, which calls `App._updateRequestMetaByParentId` (updated below)

DMarby:
> @DMarby do you have any strong opinions on creating `GrpcRequestMeta`, or just using `RequestMeta`? My preference is to keep it separate even though they are similar.

I think keeping it separate makes sense

DMarby (approved):

PR #2786, opened 2020-10-27
Allow sidebar to show grpc requests

develohpanda:
What it says on the tin. 

I reused some higher level components, but instead of modifying `MethodTag` (which is technically for HTTP methods), I created a separate, simple `GrpcTag` component as it is easier to understand.


packages/insomnia-app/app/ui/components/tags/grpc-tag.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const GrpcTag = () => (
+  <div style={{ position: 'relative' }}>
+    <div className={`tag tag--no-bg tag--small method-grpc`}>
```
sonicyeti: Any chance we could go with the #5ac5c5 value found on the visual accent items on the gRPC site itself? If not, all good. I'm thinking longer term when we have actual gRPC badges in play.

sonicyeti (approved):
Awesome, only feedback would be on the color of the tag. The gRPC site leverages 5ac5c5 for accents. This would naturally lend well to a unique badge down the road for other scenarios.

packages/insomnia-app/app/ui/components/tags/grpc-tag.js
DMarby: Agreed that this would be a nice change.
Perhaps we should also change it in the modal dropdown for creating a request to match then as well?

packages/insomnia-app/app/ui/redux/helpers.js
```diff
@@ -0,0 +1,24 @@
...
+import type { RequestGroup } from '../../models/request-group';
+
+export const shouldShowInSidebar = ({ type }: BaseModel): boolean =>
+  type === models.request.type ||
+  type === models.grpcRequest.type ||
+  type === models.requestGroup.type;
+
+type SidebarModels = Request | GrpcRequest | RequestGroup;
+
+export const shouldIgnoreChildrenOf = ({ type }: SidebarModels): boolean =>
```
DMarby: Since this and the one below are sidebar specific due to being tied to the SidebarModels type, I'd like to see them either have less generic names, or for this file to be renamed to sidebar-helpers.js or similar. 

Generally, organizing things under generic names such as "helper", "util", et al. rather then organizing them by what they do/something more specific, is not a great practice.

DMarby (changes requested):
Some notes from testing:
- The generate as code button should be hidden for gRPC requests as httpsnippet/none of our targets currently supports this
- Same with copy as curl
- Settings can be accessed, but are showing a lot of options that doesn't exist for gRPC, so this needs to be remedied by either adjusting the settings modal, or not allowing access to request settings, or such.
- Moving/copying between workspaces seems to be working, but this will need to be re-tested once #2785 is using GrpcRequestMeta instead of RequestMeta
- Pinning also seems to work since it's using RequestMeta, but again, this needs to be re-tested once it's using GrpcRequestMeta correctly.
- Duplicate and Delete also seems to work, but again, will need to be re-tested to ensure that GrpcRequestMeta is correctly used/cleaned up.

Please address the above issues either in this PR, or by creating linear tickets for them so that they don't get lost.

develohpanda:
Nice QA @DMarby! As I was running through it I was happy to see how much of it we (mostly) get for free, but there is still work needed and Linear tickets exist for those.

Again, this PR is **only for showing requests in the sidebar**, not supporting any of those actions.

image

There are some I missed during the initial task creation, which I will now create (namely, removing the items that don't apply to gRPC). They will be fixed in a separate PR, though. This PR makes no changes to the context menu of a request, because there were enough changes required to simply show a gRPC request in the sidebar.

image

packages/insomnia-app/app/ui/redux/helpers.js
develohpanda: I'll see if I can pull the entire selector (the only place where these are used) into a standalone `sidebar-*` file.

I separated these functions out because I wanted to test them, but also didn't want to export them out of the main `selectors.js` file, but I agree in hindsight that I could have named the file better.

packages/insomnia-app/app/ui/components/tags/grpc-tag.js
develohpanda: Yes they would all update, as I made sure to source the color from one location `method-grpc`. I will update, I have only been working through functional work and not exact pixels and colors.

develohpanda: I did take the current yellow from the mocks, but it is not exactly what is in Figma

develohpanda: #5ac5c5
image
image

#ffbf2f (from Figma)
image
image

I have opted to take the color from Figma (#ffbf2f), instead of changing it from the mocks.

DMarby: @nijikokun Do you have any thoughts on this?

DMarby (approved):

packages/insomnia-app/app/ui/components/tags/grpc-tag.js
develohpanda: I added a card in Linear with a link to this conversation so it is not skipped 🙂 INS-266

sonicyeti: This should be in, I'm seeing it in your latest branch. Opted for #2FA3A3 to prevent washout on lighter colors vs. the actual hex on the gRPC site.

PR #2794, opened 2020-10-28
Add support for gRPC request actions

develohpanda:
### This PR:
- adds support for loading a gRPC request as the active request (fixes INS-204)
  - as a result of this, `wrapper-debug` has been refactored to not render a body if a gRPC request is active. The next Epic adds the gRPC body.
- supports deleting a gRPC request (via button and hotkey) (Fixes INS-196)
- supports renaming a gRPC request (via double click and request settings) (Fixes INS-194)
- removes the description input from the settings modal for gRPC requests
- filter by "grpc" text (fixes INS-193)
  
- adds a patch to pinning with hotkeys (we got pin support for free)
- adds a patch to drag and drop requests (fixes INS-197)
  - dragging into a folder that contains a gRPC request was failing
- adds a patch to duplicating

### Affected areas:
- Core, and the Debug tab in Designer (minor layout change)
- request settings modal
- request actions drop-down
- hotkey behavior for delete, pin, open settings

### Self-QA:
(This QA was done incrementally rather than at the end of the PR)
- [x] delete hotkey calls the expected model remove method
- [x] delete action button on drop-down calls the expected model remove method
- [x] rename on double-click calls the expected model update method
- [x] rename from request-settings calls the expected model update method
- [x] description shown and editable for regular requests in request-setttings
- [x] description not shown gRPC requests in request-setttings
- [x] pin/unpin action button on drop-down calls the expect model update method
- [x] pin/unpin hotkey calls the expect model update method
- [x] drag between folders
- [x] drag into a folder that already contains a grpc/http request
- [x] drag drop in above/below a request of the other type (grpc and http, vice versa)
- [x] drag into closed folder (hover to open)
- [x] can duplicate
- [x] can move/copy to a different workspace
  - this currently does not copy the protofile INS-191
- [x] can move folder containing gRPC request to a different workspace
  - this currently does not copy the protofile INS-191

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/request-settings-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/request-settings-modal.js
@@ -12,7 +12,7 @@ import * as db from '../../../common/database';
 import type { Workspace } from '../../../models/workspace';
 import type { Request } from '../../../models/request';
 import type { GrpcRequest } from '../../../models/grpc-request';
-import { isGrpcRequest } from '../../../models/is-model';
+import { isGrpcRequest } from '../../../models/helpers/is-model';
 
 type Props = {
   editorFontSize: number,
--- a/packages/insomnia-app/app/ui/components/modals/request-settings-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/request-settings-modal.js
@@ -13,6 +13,7 @@ import type { Workspace } from '../../../models/workspace';
 import type { Request } from '../../../models/request';
 import type { GrpcRequest } from '../../../models/grpc-request';
 import { isGrpcRequest } from '../../../models/helpers/is-model';
+import * as requestOperations from '../../../models/helpers/request-operations';
 
 type Props = {
   editorFontSize: number,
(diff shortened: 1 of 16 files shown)


packages/insomnia-app/app/ui/components/modals/request-settings-modal.js
```diff
@@ -133,10 +142,16 @@ class RequestSettingsModal extends React.PureComponent<Props, State> {
...
       metaSortKey: -1e9, // Move to top of sort order
       parentId: activeWorkspaceIdToCopyTo,
-    });
+    };
+
+    // TODO: if gRPC, we should also copy the protofile to the destination workspace - INS-267
+
+    isGrpcRequest(request)
+      ? await models.grpcRequest.update(request, patch)
+      : await models.request.update(request, patch);
```
develohpanda, line 154: It's possible to use the following syntax (here and elsewhere) but I chose to keep it explicit. We can do a sweep after this PR and update if need be, but it is safer to keep explicit.

```suggestion
    await models.mustGetModel(request.type).update(request, patch);
```

packages/insomnia-app/app/ui/components/modals/request-settings-modal.js
```diff
@@ -155,12 +170,17 @@ class RequestSettingsModal extends React.PureComponent<Props, State> {
       return;
     }
 
-    const newRequest = await models.request.duplicate(request);
-    await models.request.update(newRequest, {
```
develohpanda, line 159: `duplicate` takes a patch parameter as well, so I have combined the two calls into one.

DMarby (approved):

PR #2795, opened 2020-10-29
Allow users to delete synced workspaces from the API

DMarby:
This allows users to remove synced projects from the API, by adding a new item to the sync dropdown menu.

@nijikokun I know you had some thoughts around the wording/messaging, wasn't sure of a good way to communicate that the workspace will be deleted from the API, but not locally.
Please look it over and let me know what wording you'd prefer.

Fixes INS-236

**Testing**
1. Enable beta sync in the app
2. Create a workspace, set up sync, commit and push
3. Choose "Delete Workspace" from the sync dropdown, and run through the flow.
4. Delete the workspace locally as well, and verify that it no longer shows up in the workspace dropdown under "Remote projects".

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/sync-delete-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/sync-delete-modal.js
@@ -1,0 +1,115 @@
+// @flow
+
+import * as React from 'react';
+import autobind from 'autobind-decorator';
+import Modal from '../base/modal';
+import ModalBody from '../base/modal-body';
+import ModalHeader from '../base/modal-header';
+import type { Workspace } from '../../../models/workspace';
+import VCS from '../../../sync/vcs';
+import { Button } from 'insomnia-components';
+
+type Props = {
+  workspace: Workspace,
+  vcs: VCS,
+};
+
+type State = {
+  error: string,
+  workspaceName: string,
+};
+
+const INITIAL_STATE: State = {
+  error: '',
+  workspaceName: '',
+};
+
+@autobind
+class SyncDeleteModal extends React.PureComponent<Props, State> {
+  modal: ?Modal;
+  input: ?HTMLInputElement;
+
+  constructor(props: Props) {
+    super(props);
+    this.state = INITIAL_STATE;
+  }
+
+  _setModalRef(n: ?Modal) {
+    this.modal = n;
+  }
+
+  _setInputRef(m: ?HTMLInputElement) {
+    this.input = m;
+  }
+
+  _updateWorkspaceName(e: SyntheticEvent<HTMLInputElement>) {
+    this.setState({ workspaceName: e.currentTarget.value });
+  }
+
+  async _handleDelete(e: SyntheticEvent<HTMLFormElement>) {
+    e.preventDefault();
+
+    const { vcs } = this.props;
+
+    try {
+      await vcs.archiveProject();
+      this.hide();
+    } catch (err) {
+      this.setState({
+        error: err.message,
+      });
+    }
+  }
+
+  async show(options: { onHide: Function }) {
+    this.modal && this.modal.show({ onHide: options.onHide });
+
+    // Reset state
+    this.setState(INITIAL_STATE);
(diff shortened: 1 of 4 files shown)


nijikokun (approved):

develohpanda (changes requested):
Code review was good. The testing steps outlined were also good, just found an edge case while doing some exploration.

If you create a workspace with a name that ends or begins with any amount of white-space, the delete modal message trims the white-space even though it is required for confirmation. How would you suggest to handle this case considering the state of data in local and server databases?

In the following screenshot, the workspace name starts and ends with three spaces, but typing the name as prompted does not match, and the delete button is disabled.
image

I know that the intention is to have users type out the workspace name and the pattern is fairly common, however GitHub/Azure (as examples that I have seen) allow you to select and copy the text. Our modal prevents text selection.

image

The only time where I feel allowing text selection might be useful for Insomnia in particular, is where the workspace contains special characters. In theory you could go down the rename and delete route, so I'm open to other options if you want to keep text selection disabled. cc: @nijikokun / @erickrico

DMarby:
Good catch!

I've corrected the CSS to not collapse whitespaces, so that it properly displays when you have more then one whitespace somewhere in the name: 
image

As for being able to select, I don't think there's a good reason to disallow it, so I've made the text in the delete modal selectable as well. The main intention of this style of dialogue is to ensure that the user doesn't just click straight through, but if they want to copy-paste the name or such, that's fine.
image

We could trim the whitespace if it's trailing/prefixing the name, but I don't think there's much to gain by special-casing that.
What's your thoughts @nijikokun?

sonicyeti (approved):

packages/insomnia-app/app/ui/components/modals/sync-delete-modal.js
```diff
@@ -0,0 +1,117 @@
...
+      <Modal ref={this._setModalRef} skinny>
+        <ModalHeader>Delete Workspace</ModalHeader>
+        <ModalBody className="wide pad-left pad-right text-center" noScroll>
+          {error && <p className="notice error margin-bottom-sm no-margin-top">{error}</p>}
+          <p className="selectable">
+            This will permanently delete the{' '}
+            <strong style={{ whiteSpace: 'pre-wrap' }}>{workspace.name}</strong> workspace remotely.
+          </p>
+          <p className="selectable">
+            Please type <strong style={{ whiteSpace: 'pre-wrap' }}>{workspace.name}</strong> to
```
develohpanda: Nitpicking here, but I'd suggest to put the workspaceName node into a component/variable and reuse it so styles are defined in one place for any future changes

```js
const workspaceName = <strong>...</strong>;

// ...permanently delete the{' '}{workspaceName}{' '}workspace remotely..
```

develohpanda (approved):
Looks great!

packages/insomnia-app/app/ui/components/modals/sync-delete-modal.js
DMarby: Updated 👍

develohpanda:
Actually, thinking about this one - are you able to achieve a similar result using the `PromptModal` that already exists? It contains input validation.

I ask because the UI pattern here is very different to all other modals, and preference would be to keep consistent, unless this is a conscious decision cc: @erickrico

DMarby:
Using a different UI pattern here was a conscious design decision made by @nijikokun, as this is a high-impact operation, and none of the existing patterns fitted that well.

I looked at using the PromptModal initially, but it doesn't support text above the input field other than a small label, doesn't provide proper feedback as to when or why a validation is failing, and uses our standard model footer to right-align the submit button (and provides no way to change the state/disable it), which creates a disconnect between the input field and the button that is unwanted for this pattern.

While it would've been possible to modify it to also support this pattern, adding further complexity and special-casing to the PromptModel for this use-case did not feel ideal, which is why I opted to make this a separate component instead.

nijikokun (approved):
@DMarby looks good to me :+1:

PR #2796, opened 2020-10-29
Unit Testing sidebar design integration

sonicyeti:
This PR covers integration of the latest unit testing design sidebar. 

**Features include:**
- Unit tests grouped by their associated testing suite
- Filtering on test suites
- Executing and deleting of test suites within the sidebar
- Collapsing & expanding of suites
- Adding test suites
- Adding individual tests to a suite contextually
- Overall visual design integration

UI Demonstration:

Fixes INS-121
Fixes INS-258


packages/insomnia-components/components/sidebar/sidebar-unit-testing.js
```diff
@@ -0,0 +1,177 @@
...
+  const searchTests = event => {
+    setFilter(event.target.value);
+  };
+
+  const filteredItems = unitTestSuites
+    .filter(suite => {
+      if (filter == null) return suite;
+      else if (suite.name.toLowerCase().includes(filter.toLowerCase())) {
+        return suite;
+      }
```
DMarby: I think this could be simplified a bit, with something along the lines of this:
```suggestion
      return !filter || suite.name.toLowerCase().includes(filter.toLowerCase())
```

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -445,48 +463,35 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
...
+              activeTestSuite={activeId}
+            />
+          )}
         </div>
       </ErrorBoundary>
     );
   }
 
+  componentDidMount() {
+    models.unitTest.all().then(res => {
```
DMarby, line 506: If we're sticking with this approach instead of Redux, I think this should be able to just call `getUnitTests()`?

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -445,48 +463,35 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
...
-                    {testsRunning ? 'Running... ' : 'Run Tests'}
-                  </DropdownItem>
-                  <DropdownItem onClick={this._handleDeleteUnitTestSuite.bind(this, s)}>
-                    Delete Suite
-                  </DropdownItem>
-                </Dropdown>
-              </li>
-            ))}
-          </ul>
+          {unitTests.length > 0 && (
```
DMarby: I don't think this conditional is correct, as it currently makes it impossible to create a test-suite in an empty project, as the sidebar is just blank.

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -214,7 +225,14 @@ class WrapperUnitTest extends React.PureComponent<Props, State> {
     });
   }
 
-  async _handleSetActiveUnitTestSuite(unitTestSuite: UnitTestSuite): Promise<void> {
+  async _handleSetActiveUnitTestSuite(e, unitTestSuite: UnitTestSuite): Promise<void> {
```
DMarby: Not sure if I understand this change, as far as I can tell this function is identical to `_handleSetActiveSuite` below, and the below function does not seem to get called anywhere?

Not sure if this breaks `_handleCreateTestSuite` as well, since that does not have the event included when calling this function?

packages/insomnia-app/app/ui/components/wrapper.js
```diff
@@ -186,6 +186,7 @@ export type WrapperProps = {
   activeUnitTestResult: UnitTestResult | null,
   activeUnitTestSuites: Array<UnitTestSuite>,
   activeUnitTests: Array<UnitTest>,
+  allUnitTests: Array<UnitTest>,
```
DMarby: As far as I can tell, this doesn't seem to be used by anything?

packages/insomnia-components/components/sidebar/sidebar-unit-test-suite-item.js
```diff
@@ -0,0 +1,156 @@
...
+import { useToggle } from 'react-use';
+import { motion } from 'framer-motion';
+import SidebarItem from './sidebar-item';
+import SvgIcon from '../svg-icon';
+
+type Props = {|
+  name: string,
+  parentId: string,
+  unitTests: Array<{
+    _id: string,
```
DMarby, line 14: Would it be possible to use an existing type for this, or define one somewhere, so that this definition doesn't have to be redefined here and for every handler below as it is currently?

Same thing for the SidebarUnitTesting component.

packages/insomnia-components/components/sidebar/sidebar-unit-testing.js
```diff
@@ -0,0 +1,177 @@
...
+  onAddSuiteClick,
+  onTestSuiteClick,
+  onDeleteSuiteClick,
+  onExecuteSuiteClick,
+  onCreateTestClick,
+  activeTestSuite,
+  className,
+}: Props) => {
+  const [filter, setFilter] = React.useState(null);
+  if (unitTestSuites === null) {
```
DMarby: I don't think this is correct, as it will make it impossible to create test suites in an empty workspace.

packages/insomnia-components/components/sidebar/sidebar-unit-test-suite-item.js
```diff
@@ -0,0 +1,156 @@
...
+              <SvgIcon icon="plus" />
+            </span>
+          </StyledTestSuiteActions>
+        )}
+      </SidebarItem>
+      <motion.div
+        initial={{ height: isToggled ? '100%' : '0px', display: isToggled ? 'block' : 'none' }}
+        animate={{ height: isToggled ? '100%' : '0px' }}
+        transition={{ duration: 0.2, ease: 'easeInOut', delay: 0 }}>
+        {unitTests.length !== 0 &&
```
DMarby: I don't think the .length check is needed here, as `.filter` should work fine on an empty array?

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
```diff
@@ -41,13 +39,15 @@ type Props = {|
...
   resultsError: string | null,
+  unitTests: Array<UnitTest> | null,
 |};
 
 @autobind
 class WrapperUnitTest extends React.PureComponent<Props, State> {
   state = {
     testsRunning: null,
     resultsError: null,
+    unitTests: models.unitTest.all(),
```
DMarby, line 50: If it's not a lot of work, I think it might be better to get these from Redux instead of directly from the database, as doing so would mean that there's no need for a refresh function like `getUnitTests`.

You should be able to create a selector similar to `selectActiveUnitTestSuites`, but for getting all tests belonging to the current workspace:
https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/redux/selectors.js#L307-L313

And use that in app.js to add it to our wrapper props:
https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/containers/app.js#L1377

Then you'll be able to access them from `this.props.wrapperProps`, same as how we access activeUnitTests/activeUnitTestSuite in this file.

---

If we do prefer sticking with grabbing them from the database, I'd prefer if we were to fetch just the unittests belonging to this workspace, instead of all the ones in the database.

packages/insomnia-components/components/sidebar/sidebar-unit-test-suite-item.js
```diff
@@ -0,0 +1,156 @@
...
+        <div onClick={toggle}>
+          <SvgIcon icon={isToggled ? 'folder-open' : 'folder'} />
+        </div>
+        <StyledTestSuiteLabel onClick={onTestSuiteClick}>{name}</StyledTestSuiteLabel>
+        {className && (
+          <StyledTestSuiteActions>
+            <span onClick={onDeleteSuiteClick}>
+              <SvgIcon icon="trashcan" />
+            </span>
+            <span onClick={onExecuteSuiteClick}>
```
DMarby: Do we need to handle the case where tests are already running here, as we were doing previously using the `testsRunning` state property?

DMarby (changes requested):
This is awesome, looks really good visually!

Wasn't really able to run through testing of this as I wasn't able to create any new test suites, so will QA it again in the next round of review.

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
sonicyeti: I think perhaps that was via the old UI. Here, even if we have an empty spec, we can still create a suite (and tests) on it.

packages/insomnia-components/components/sidebar/sidebar-unit-test-suite-item.js
sonicyeti: Thanks for the catch! Initially we hadn't planned for actions on the left rail suites for this first pass. Converted them to buttons and now checking to disable accordingly!

packages/insomnia-components/components/sidebar/sidebar-unit-testing.js
sonicyeti: Add suite is part of the UT header component so we can still add them but it did make me realize that we shouldn't actually show the filter input if suites haven't been created, fixed that.

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
DMarby: Sorry, but no, this logic is wrong and needs to be removed.
It works for you since you're pulling *all unit-tests* rather then the ones for just the workspace from the database, but when you have an instance of Designer with no workspaces with unit-tests, this makes the sidebar not render at all:

image

packages/insomnia-components/components/sidebar/sidebar-unit-testing.js
DMarby: Returning here will make the UT header component not render, as it's included further down, so this would block it.

However, since this is checking for null and not empty, in reality, it doesn't have any effect, but I'm not sure of the purpose of it, because I don't think this array can ever be null?

packages/insomnia-app/app/ui/components/wrapper-unit-test.js
sonicyeti: Ah gotcha, hadn't tested this without tests across all workspaces. Yep, will fix now!

packages/insomnia-components/components/sidebar/sidebar-unit-testing.js
sonicyeti: Yanked it entirely.

PR #2798, opened 2020-10-30
Unsafe deletion of a protofile

develohpanda:
Behavior changes:
- Instead of the prompt button, we are using a modal because we would like to seek confirmation from the user after providing some context on repercussions
- After deleting, if that protofile was previously selected, we reset the selection

Screenshot of delete button change:
image

Screenshot of confirmation dialog:
image

@nijikokun please advise on wording for the title, body and buttons of the confirmation modal. It will  also be possible to show a count of affected requests (eg. "10 requests that use proto file..."). Showing a list of items will need more work/design.

QA: 
- After deleting a selected protofile, the save button is disabled (because no protofile is selected)
- After deleting a protofile, loading protofiles for the workspace updates correctly

Fixes INS-209

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/proto-files-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/proto-files-modal.js
@@ -10,7 +10,7 @@ import type { Workspace } from '../../../models/workspace';
 import Modal from '../base/modal';
 import ProtoFileList from '../proto-file/proto-file-list';
 import FileInputButton from '../base/file-input-button';
-import { showError } from './index';
+import { showAlert, showError } from './index';
 import fs from 'fs';
 import path from 'path';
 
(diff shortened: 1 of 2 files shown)


DMarby (approved):

sonicyeti (approved):
Very cool, we have a few cases of destructive interactions in modal, might be worth leveraging the message/confirm DOM element to push verbiage & actions that would typically spawn another modal down the road.

develohpanda:
> Very cool, we have a few cases of destructive interactions in modal, might be worth leveraging the message/confirm DOM element to push verbiage & actions that would typically spawn another modal down the road.

Ya nice observation. At first thought potentially something like a tool-tip to be able to convey more information. I don't really like the nested modals either, but the next iteration of the delete flow requires a modal too so I've kept it this way.

PR #2799, opened 2020-10-30
Extract foundational components from request and response panes

develohpanda:
This PR does some refactoring of components, laying the groundwork to create gRPC panes.

### Nothing selected:
image

### HTTP selected:
image

### gRPC selected:
image

## QA:
There should be no material change to the debug tab in it's various states. I suggest running through QA with the live application and this branch running in parallel.
- no request selected
- request selected with no response history
- request sending (and timer is shown) with no previous response history
- request sending (and timer is shown) _with_ previous response shown
- request cancellation
- changing themes
- any other exploration

### DIFF
--- a/packages/insomnia-app/app/ui/components/panes/pane.js
+++ b/packages/insomnia-app/app/ui/components/panes/pane.js
@@ -1,0 +1,45 @@
+// @flow
+import classnames from 'classnames';
+import React from 'react';
+
+type PaneProps = {
+  className?: string,
+  type: 'request' | 'response',
+};
+
+type PaneHeaderProps = {
+  className?: string,
+};
+
+type PaneBodyProps = {
+  className?: string,
+  placeholder?: boolean,
+};
+
+export const Pane = ({ className, type, children }: PaneProps) => (
+  <section className={classnames(`${type}-pane`, 'theme--pane', 'pane', className)}>
+    {children}
+  </section>
+);
+
+export const PaneHeader = ({ className, children }: PaneHeaderProps) => (
+  <header className={classnames('pane__header', 'theme--pane__header', className)}>
+    {children}
+  </header>
+);
+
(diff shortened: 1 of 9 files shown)


packages/insomnia-app/app/ui/components/wrapper-debug.js
```diff
@@ -252,10 +256,42 @@ class WrapperDebug extends React.PureComponent<Props> {
...
+            </ResizablePaneWrapper>
+          </ErrorBoundary>
+          {dragPanes}
+          <ErrorBoundary showAlert>
+            <ResizablePaneWrapper ref={handleSetResponsePaneRef}>
+              <GrpcResponsePane />
+            </ResizablePaneWrapper>
+          </ErrorBoundary>
+        </React.Fragment>
+      );
```
develohpanda, line 294: There is some refactoring in #2712 which eliminates the need to add repetitive logic for resizable drag panes. That would eliminate the need for `dragPanes` and the `ResizablePaneWrapper` class component, so this file would simplify a lot.

I decided to create a temporary wrapper so that we can move towards functional components for the new pane components that will be needed.

packages/insomnia-app/app/ui/components/panes/pane.js
```diff
@@ -0,0 +1,45 @@
...
+  </section>
+);
+
+export const PaneHeader = ({ className, children }: PaneHeaderProps) => (
+  <header className={classnames('pane__header', 'theme--pane__header', className)}>
+    {children}
+  </header>
+);
+
+export const paneBodyClasses = 'pane__body theme--pane__body';
```
develohpanda, line 31: Exported separately because some views apply these body classes directly to `ReactTabs`, and some use the generic `PaneBody`. In either case, the classes should be identical.

sonicyeti (approved):
Great provisioning and clean-up!

DMarby (approved):

PR #2802, opened 2020-11-01
Add metadata fields to openapi-2-kong package.json

mikaello:
The metadata fields on https://www.npmjs.com/package/openapi-2-kong are a bit scarce, and it would be nice if these were improved:

image


packages/openapi-2-kong/package.json
```diff
@@ -29,5 +29,8 @@
     "webpack": "^4.42.1",
     "webpack-cli": "^3.3.11"
   },
+  "repository": "https://github.com/Kong/insomnia/tree/develop/packages/openapi-2-kong",
+  "bugs": "https://github.com/Kong/insomnia/issues",
```
develohpanda: According to [the docs](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#bugs), this needs to be the following:

```suggestion
  "bugs": {
    "url": "https://github.com/Kong/insomnia/issues"
  },
```

packages/openapi-2-kong/package.json
```diff
@@ -29,5 +29,8 @@
...
+  "license": "MIT",
```
develohpanda: @nijikokun should the license be MIT or Apache2? Inso is set to Apache2.

develohpanda (commented):
Thanks for doing this! One comment, and a question for @nijikokun

packages/openapi-2-kong/package.json
mikaello: Thanks for looking up! I based my changes on an old convention, obviously outdated. Fixed now.

packages/openapi-2-kong/package.json
mikaello: Just a comment to this question: if you set this to Apache2, I think you should change the licence that is distributed with openapi-2-kong, which is currently MIT: 

```shell
$ npm pack openapi-2-kong
$ tar xzf openapi-2-kong-2.2.24.tgz
$ cat package/LICENSE

MIT License

Copyright (c) Insomnia REST Client

Permission is hereby granted....
```

For reference why this came to my attention in the first place, see [discussion in Kong Nation](https://discuss.konghq.com/t/import-swagger-open-api-document-to-create-api-in-kong/510/18?u=michael_johansen).

develohpanda (approved):
🎉

packages/openapi-2-kong/package.json
develohpanda: A [recent PR](https://github.com/Kong/insomnia/pull/3233) went in to set up license fields using feedback from the Legal team, so we can remove this field from this PR. I have created a separate internal ticket INS-658 in order to address your suggestion (with this package and all other packages).

```suggestion
```

packages/openapi-2-kong/package.json
```diff
@@ -31,5 +31,14 @@
     "webpack": "^4.42.1",
     "webpack-cli": "^3.3.11"
   },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Kong/insomnia.git",
+    "directory": "packages/openapi-2-kong"
+  },
```
develohpanda, line 38: Nice improvement here! We should certainly update the other packages to follow this pattern, TIL.

reynolek (approved):
Approved from product perspective

PR #2803, opened 2020-11-01
Fix change log URL

babajidemm:


develohpanda:
Thanks for the PR! This is needed because we generate the full changelog url by combining the base and version. `https://insomnia.rest/changelog/core` itself is not a valid url but `https://insomnia.rest/changelog/core/2020.5.0` is, as is `https://insomnia.rest/changelog/designer/2020.5.0`.

In the future we will look to remove `/core` from the changelog path.

https://github.com/Kong/insomnia/blob/e8d1e9039f090aa34baf181c13c36c368f6413a1/packages/insomnia-app/app/common/constants.js#L91-L94

I will close this PR as such :)

PR #2809, opened 2020-11-04
Support for gRPC Unary and Client streaming

develohpanda:
The following gif shows that you can make unary and client streaming requests to `grpcb.in:9000` using [hello.proto](https://github.com/moul/pb/blob/master/hello/hello.proto).

There is no state management in this PR, as that is a significant task on it's own. Therefore, while sending a request, or starting a stream, there is no spinner or feedback. The `stream`, `commit`, and `cancel` buttons should only respond when a stream is running. When a stream is not running, it should show a message in console saying there is no running call. 

Errors (for example, response errors) are _handled_ and written to console. A future PR will improve the error handling and show them in the UI.

I have completed a self-review of the code, but may have missed some things. Please raise any questions/concerns.

## QA:
Try the following workflows with one request in the sidebar:
- Unary ("SayHello"):
  - `send`
    - you should see a response in console
- Client Streaming ("LotsOfGreetings"):
  - `start` -> `stream` -> `stream` -> `commit`
    - you should see a response in console with each message you sent
  - `start` -> `commit`
    - you should see a response in console
  - `start` -> `stream` -> `cancel` 
  - `start` -> `cancel` 
    - you should see an error in console 
  - `stream`
  - `commit`
  - `cancel`
    - you should see a warning in console saying the client call could not be found

An example message body for each RPC is the following. Try this, as well as some invalid message bodies.
```json
{ "greeting": "insomnia" }
```

There is also functionality for concurrently running multiple RPCs, but the state of a request is not visualized in this PR so can be difficult to validate (no spinner shown, etc).

Fixes INS-223, Fixes INS-224, adds functionality for INS-225 and INS-226


packages/insomnia-app/app/ui/css/constants/dimensions.less
```diff
@@ -1,13 +1,13 @@
...
   /* Font Size */
-  --font-size-xxs: 0.5rem;
-  --font-size-xs: 0.625rem;
-  --font-size-sm: 0.75rem;
-  --font-size-md: 0.8125rem;
-  --font-size-lg: 0.9375rem;
-  --font-size-xl: 1.1875rem;
-  --font-size-xxl: 1.3125rem;
-  --font-size-xxxl: 1.5rem;
+  --font-size-xxs: 8px;
```
develohpanda: Undo this

packages/insomnia-app/app/main/ipc-grpc.js
```diff
@@ -0,0 +1,33 @@
...
+
+export function init() {
+  ipcMain.on(GrpcRequestEventEnum.sendUnary, (e, requestId) =>
+    grpc.sendUnary(requestId, createResponseCallbacks(e)),
+  );
+  ipcMain.on(GrpcRequestEventEnum.startStream, (e, requestId) =>
+    grpc.startClientStreaming(requestId, createResponseCallbacks(e)),
+  );
+  ipcMain.on(GrpcRequestEventEnum.sendMessage, (e, requestId) =>
+    grpc.sendMessage(requestId, createSendError(e)),
```
develohpanda: Specifically, when working with streams, work needs to happen in the main process. If this happens in the render process, a message is never sent on a user action (which is what we need).

Using `ipcMain` is not an efficiency decision, it simply didn't work with the `@grpc/grpc-js` library in the render process.

Ref: https://github.com/grpc/grpc-node/issues/1455#issuecomment-649072697

packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -0,0 +1,126 @@
...
+  requestId: string,
+  respond: ResponseCallbacks,
+): Promise<void> => {
+  // Load method
+  const req = await models.grpcRequest.getById(requestId);
+  const selectedMethod = await protoLoader.getSelectedMethod(req);
+
+  if (!selectedMethod) {
+    respond.sendError(new Error(`The gRPC method ${req.protoMethodName} could not be found`));
+    // TODO: sendEnd
```
develohpanda, line 75: The "sendEnd" event is introduced in the next PR, but I have added a TODO so that I remember.

packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -0,0 +1,144 @@
...
+const createClient = (req: GrpcRequest, respond: ResponseCallbacks): Object | undefined => {
+  if (!req.url) {
+    respond.sendError(req._id, new Error('gRPC url not specified')); // TODO: update wording
+    return undefined;
+  }
+  const Client = grpc.makeGenericClientConstructor({});
+  return new Client(req.url, grpc.credentials.createInsecure());
+};
+
+export const sendUnary = async (requestId: string, respond: ResponseCallbacks): Promise<void> => {
```
develohpanda, line 19: The `sendUnary` and `sendClientStreaming` functions are structurally quite similar, but I have kept them duplicated/separated for the initial pass.

packages/insomnia-app/app/models/__tests__/grpc-request.test.js
```diff
@@ -11,7 +11,6 @@ describe('init()', () => {
       name: 'New gRPC Request',
       description: '',
       protoFileId: '',
-      protoServiceName: '',
       protoMethodName: '',
```
develohpanda, line 14: Service and method no longer need to be stored separately, this was a revelation after working with the `@grpc/grpc-js` library. This library is not well documented, so the initial properties were a best guess (through research into other apps)

packages/insomnia-app/app/network/grpc/proto-loader.js
```diff
@@ -0,0 +1,44 @@
...
+  enums: String,
+  defaults: true,
+  oneofs: true,
+};
+
+const isTypeOrEnumDefinition = (obj: Object) => 'format' in obj; // same check exists internally in the grpc library
+const isServiceDefinition = (obj: Object) => !isTypeOrEnumDefinition(obj);
+
+// TODO: instead of writing to a temp file and loading the protoFile every time methods are required,
+//  add an in-memory caching strategy, indexed by the protoFile._id - INS-272
```
develohpanda, line 20: `@grpc/proto-loader` only accepts a file path, not the contents, so we need to write and read from a temp file in order to load it. INS-272

packages/insomnia-app/app/network/grpc/proto-loader.js
```diff
@@ -0,0 +1,44 @@
...
+
+// TODO: The file path for protoLoader.load can also be a URL, so we can avoid
+//  writing to a file in those cases, but it becomes more important to cache
+//  We also need to think about how to store a reference to a proto file and it's
+//  implications on import/export/sync - INS-271
```
develohpanda, line 25: INS-271 - hyperlinked for easy navigation :)

sonicyeti (approved):
Code looks solid and I was able to successfully complete all of the steps in the Q/A outline. Might be nice to have some sort of proto file hinting via "based on x.proto" tooltip on the left rail request or something. Seems too "set and forget" ATM. Maybe this can be handled as an editable asset in designer or we add a collapsable proto file viewing pane.

packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -0,0 +1,144 @@
...
+import callCache from './call-cache';
+import { ResponseCallbacks } from './response-callbacks';
+
+const createClient = (req: GrpcRequest, respond: ResponseCallbacks): Object | undefined => {
+  if (!req.url) {
+    respond.sendError(req._id, new Error('gRPC url not specified')); // TODO: update wording
+    return undefined;
+  }
+  const Client = grpc.makeGenericClientConstructor({});
+  return new Client(req.url, grpc.credentials.createInsecure());
```
DMarby, line 16: Not a blocker for this PR, but hardcoding `createInsecure()` here will make it impossible to connect to the vast majority of gRPC servers, as TLS is generally used.

Would be good to create a ticket in Linear to follow up on this, as before we ship the MVP we might want to support servers that's using TLS, and at some point we'll also need to configure client certificates, as mutual TLS is the most common authentication method for gRPC, but I'll defer to @nijikokun on whether it's a blocker for the MVP or not/priority.

packages/insomnia-app/app/network/grpc/method.js
```diff
@@ -0,0 +1,35 @@
...
+  client: 2, // 10
+  bidi: 3, // 11
+};
+
+export type GrpcMethodType = $Values<typeof GrpcMethodTypeEnum>;
+
+export const getMethodType = ({
+  requestStream,
+  responseStream,
+}: GrpcMethodDefinition): GrpcMethodType => parseInt(`${+requestStream}${+responseStream}`, 2);
```
DMarby: I think this is overcomplicating things a bit, since there's so few possible states to map in-between.

I'd prefer just a few if statements or similar, which is also how the upstream grpc library handles it: https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/make-client.ts#L134-L146

DMarby (approved):
Looks good overall.

Making the gRPC requests in the main process should be fine for the MVP, but doing so might bring potential issues down the line with blocking the main process, which could cause the entire app to stall, so it would be good to file a Linear ticket to look at re-architecting this in the future to use a separate non-render process instead and whether the gRPC library supports that or not.

packages/insomnia-app/app/network/grpc/index.js
develohpanda: TLS support was descoped for MVP (see tech design)

develohpanda:
> a separate non-render process instead and whether the gRPC library supports that or not.

Thanks, that's a great point and needs some more exploration. For now I've made sure to use asynchronous code where it's possible (eg. file I/O) but it would be worth exploring the use of a separate non-render process altogether.

The actual gRPC logic (`network/grpc/index.js`) is isolated enough so that it doesn't link directly to React/IPC in any way so moving the invocations to a separate location (away from main) should be somewhat simple. (It was all in render initially and I just wrote an interface to move it to main). There is also scope for converting this logic into a class but that is just a refactor from which I don't anticipate any material impact.

For some context about what was failing in particular: unary works fine in the render process, but I came across a hurdle when working with the stream object from streaming RPCs. In the render process, when the invocation into gRPC returns a `call` object (the stream), and you `.write()` a message onto the stream, the write callback is never invoked (indicating the message was not written to the stream, as far as I understand).

I just did a rudimentary experiment now to limit how much is happening in the main process by only writing to the stream in main but doing everything else in render, but that didn't work. In the following screenshot, the callback is never invoked. 😞 
image

Is there anything you can think of, off the top of your head, that might cause this behavior when working with streams in the render process? In particular, if you write a message immediately after starting the stream, the message does get written. But if you start the stream and write a message during individual user interactions, it doesn't get written. Perhaps there is some kind of resource collection, or it's running in a sandbox that cuts a connection to the stream during some part of the life-cycle?

(Also logged as an issue to be addressed later: INS-273)

develohpanda:
@DMarby / @sonicyeti unfortunately this PR requires a re-review, although these are mostly UI changes after merging #2817. I wasn't anticipating as many changes needed to link up. Please re-review since your last review (in case you weren't aware of this option on GH)

image

DMarby:
> Is there anything you can think of, off the top of your head, that might cause this behavior when working with streams in the render process? In particular, if you write a message immediately after starting the stream, the message does get written. But if you start the stream and write a message during individual user interactions, it doesn't get written. Perhaps there is some kind of resource collection, or it's running in a sandbox that cuts a connection to the stream during some part of the life-cycle?

Stream behaviour itself and such in general *should* be identical across processes, nothing comes to mind on top of my head that would cause this behaviour at least, I haven't seen it when using streaming in the render process before.
The theory in the [related issue](https://github.com/grpc/grpc-node/issues/1455#issuecomment-649072697) that you found in the grpc-repo that it's related to the HTTP2 internals in Electron seems plausible, which is definitely unfortunate, since it's nothing we can fix on our end 😅

packages/insomnia-app/app/ui/components/dropdowns/grpc-method-dropdown.js
```diff
@@ -1,43 +1,28 @@
...
-
-type Method = {
-  path: string,
-};
+import type { GrpcMethodDefinition } from '../../../network/grpc/method';
 
 type Props = {
-  methods?: Array<Method>,
-  selectedMethod?: Method,
+  methods: Array<GrpcMethodDefinition>,
```
DMarby, line 7: Nice tightening of the type here 👍

DMarby, line 7: Nice tightening/re-use of the types here 👍

DMarby (approved):
Looks great 👍

sonicyeti (approved):
Amazing!!!

PR #2811, opened 2020-11-04
Stubbing request pane ui

sonicyeti:
This PR covers the scaffolding and styling of the core UI elements for gRPC feature implementation. Items include:

- Client streaming | headers tab system
- Body and supporting streams child tab system
- Editors for child system, read-only for stream instances
- "URL Bar" for gRPC, protocol | one-line editor | method DD | execute

UI provisioning visual:

Fixes INS-202

### DIFF
--- a/packages/insomnia-app/app/ui/components/panes/grpc-request-pane.js
+++ b/packages/insomnia-app/app/ui/components/panes/grpc-request-pane.js
@@ -1,7 +1,219 @@
 // @flow
 import React from 'react';
-import BlankPane from './blank-pane';
+import { Pane, paneBodyClasses, PaneBody, PaneHeader } from './pane';
+import { Tab, TabList, TabPanel, Tabs } from 'react-tabs';
+import classnames from 'classnames';
+import Button from '../base/button';
+import GRPCEditor from '../editors/grpc-editor';
+import OneLineEditor from '../codemirror/one-line-editor';
+import type { Settings } from '../../../models/settings';
 
-const GrpcRequestPane = () => <BlankPane type="request" />;
+type Props = {
+  handleRender: Function,
+  handleGetRenderContext: Function,
+  nunjucksPowerUserMode: boolean,
+  isVariableUncovered: boolean,
+  workspace: Workspace,
+  settings: Settings,
+};
+
+const GrpcRequestPane = (props: Props) => {
+  const {
+    handleRender,
+    handleGetRenderContext,
+    nunjucksPowerUserMode,
+    isVariableUncovered,
+    workspace,
+    settings,
+  } = props;
+
+  return (
+    <Pane type="request">
+      <PaneHeader>
+        <div className="method-grpc pad">gRPC</div>
+        <div className="urlbar">
+          <form className={'form-control'}>
+            <OneLineEditor
+              render={handleRender}
+              forceEditor
+              type="text"
+              getRenderContext={handleGetRenderContext}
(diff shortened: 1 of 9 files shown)


packages/insomnia-app/app/ui/components/panes/grpc-request-pane.js
```diff
@@ -1,7 +1,242 @@
...
+            }}>
+            /hello.HellowService/someOtherStuff/hello.HellowService
+          </DropdownItem>
+          <DropdownItem
+            onClick={() => {
+              console.log('DD Item clicked...');
+            }}>
+            /hello.HellowService/evenMoreThings
+          </DropdownItem>
+        </Dropdown>
```
develohpanda: I suggest converting this into a component that takes in the following props:

```js
type Method = {
	path: string; // this will always be unique
}

type Props = {
	methods: Array<Method>,
	selectedMethod: Method,
}
```

packages/insomnia-app/app/ui/components/panes/grpc-request-pane.js
```diff
@@ -1,7 +1,242 @@
...
+            </Tab>
+            <Tab>
+              <button>Headers</button>
+            </Tab>
+          </TabList>
+          <TabPanel className="react-tabs__tab-panel">
+            <Tabs className={classnames(paneBodyClasses, 'react-tabs', 'react-tabs--nested')}>
+              <div className="tab-action-wrapper">
+                <div className="tab-action-tabs">
+                  <TabList>
```
develohpanda: This nested `TabList` would be good as a component - `GrpcMessages` for example. That can take the following props (in addition to `settings`/anything else needed):

```js
type Message = {
	id: string,
	created: number,
	text: string,	
}
type Props = {
	bodyText: string,
	messages: Array<Message>,
}
```

The tabs would need to be ordered by the `created` time, and the tabs numbered incrementally from that.

This should actually make it trivial for the response pane, I think? If you also added props to show/hide the body tab, and stream+commit buttons, it essentially becomes the response pane!

develohpanda (approved):
LGTM, minor suggestions to create two standalone components in order to assist with linking data together later on. 🎉 

Visually looks 💯 I'm so exicted

develohpanda (approved):
Perfect 😊

packages/insomnia-app/app/ui/css/components/grpc-url-bar.less
```diff
@@ -0,0 +1,67 @@
...
+    padding-right: var(--padding-md);
+    padding-left: var(--padding-md);
+    min-width: 5em;
+    text-align: center;
+    line-height: var(--height-nav);
+    height: 100%;
+    margin-left: auto;
+  }
+
+  /*
```
DMarby, line 61: Is this commented out code needed for anything, or can it be removed?

DMarby (approved):
Looks good 👍 Just one question about some commented out css.

packages/insomnia-app/app/ui/css/components/grpc-url-bar.less
sonicyeti: Thanks for the catch, this stems from a previous implementation of an inline single-line-editor DD, didn't want to drop it in case we shift how this is rolled in but it can go now for sure. I'll pull it!

develohpanda:
Closing this PR because it was merged with #2817

PR #2813, opened 2020-11-05
gRPC state management

develohpanda:
This PR adds state management for gRPC. The React Context implementation stores, per request:
- the request and response messages
  - power the tabs you see in the request and response panes
- whether a stream is open/request is running
  - show a spinner in the sidebar and response pane
  - disable method selection in the dropdown
  - change send button to cancel
  - show the stream and commit buttons
- errors and status responses from the gRPC networking layer
  - bubbled up to the response pane header

This PR also adds the altering of states, depending on various actions. For example, when starting a request or changing the method selection, we want to clear the request state.

I employed a pattern inspired by [this](https://kentcdodds.com/blog/how-to-use-react-context-effectively/) in order to make it safer to use React context. It is a little more verbose, but once the foundations are in place they infrequently need changes. I have added unit tests for the context provider (ensuring IPC listeners are initialized on mount and destroyed on unmount), and for all actions and the reducer (ensuring state changes are as expected).

The general data flow is the following, feel free to ask questions for clarification or further detail about how streaming, error, status events flow through the application.

image

Fixes INS-221, Fixes INS-225, Fixes INS-226, Fixes INS-227


packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -93,13 +102,25 @@ export const commit = (requestId: string) => callCache.get(requestId)?.end();
...
-    respond.sendError(requestId, err);
+    // Don't do anything if cancelled
+    // TODO: test with other errors
+    if (err.code !== GrpcStatusEnum.CANCELLED) {
+      respond.sendError(requestId, err);
+    }
   } else {
     respond.sendData(requestId, value);
   }
+  respond.sendEnd(requestId);
```
develohpanda, line 141: Need to `sendEnd` in the catch block in `_parseMessage` as well

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1290,7 +1290,7 @@ class App extends PureComponent {
 
     return (
       <KeydownBinder onKeydown={this._handleKeyDown}>
-        <GrpcEventWrapper>
+        <GrpcProvider>
```
DMarby, line 1293: I think it might make sense to move this into WrapperDebug so that it only surrounds the PageLayout there, or into Wrapper, because I don't think we'll need to access these contexts outside of that.

packages/insomnia-app/app/ui/context/grpc/create-ipc-listeners.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import { ipcRenderer } from 'electron';
+import { GrpcResponseEventEnum } from '../../../common/grpc-events';
+import grpcActions from './grpc-actions';
+import type { GrpcDispatch } from './grpc-actions';
+
+// TODO: Do we need to clear listeners or will they overwrite?
```
DMarby: ipcRenderer is a standard EventEmitter, so when you register a listener, existing ones will be kept intact, meaning you'll get multiple listeners for the same event. 
Listeners are also not automatically deregistered.

packages/insomnia-app/app/ui/containers/app.js
develohpanda: Note that the immediate child of this context, a div, accepts a uniquenessKey. This causes all its children to re-mount when the logic says it needs to be force refreshed. I don't recall this happening often, however. 

This context, as designed, never needs to be force refreshed, which is why its placed at the top level.

Wrapper-debug unfortunately unmounts and remounts when navigating between tabs in designer, meaning state will be lost each time, so context cannot go there.

Between App and Wrapper, I would most likely favour App. I can investigate moving it one level down, but that's as far as it can go. 

When we have more contexts, a standalone context wrapper component will be used to place all global contexts onto the dom, but since we only have one right now I did it directly.

packages/insomnia-app/app/ui/context/grpc/create-ipc-listeners.js
develohpanda: Thanks. It shouldn't impact us directly because adding main and renderer listeners are only done once per lifetime of the app, regardless of how many gRPC requests or workspaces exist, but I will clear listeners when the context unmounts for completeness.

packages/insomnia-app/app/ui/containers/app.js
DMarby: I think that makes sense, fine to leave it as-is for now then.

packages/insomnia-app/app/ui/context/grpc/create-ipc-listeners.js
develohpanda: I have added logic to clear listeners 👍

packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -136,9 +155,8 @@ const _parseMessage = (request: Request, respond: ResponseCallbacks): Object | u
     return JSON.parse(request.body.text || '');
   } catch (e) {
     // TODO: How do we want to handle this case, where the message cannot be parsed?
-    //  Currently an error will be shown and the RPC stopped, but we can be less destructive
+    //  Currently an error will be shown, but the stream will not be cancelled.
     respond.sendError(request._id, e);
-    // TODO: sendEnd
     return undefined;
   }
```
develohpanda, line 161: This catch block is reached when we try to send a message with invalid JSON that cannot be parsed to an object. If this catch block is reached, the message is technically not accepted, however an invalid message shows up as a new stream in the UI, even though it was not sent to the server. An error does show, but the invalid message appears to be accepted.

How should we handle this? Do we:
- only show a new tab, once a stream has been _sent_ (this could result in a visual delay)
- delete the tab if the stream is not accepted (this could look strange where a tab shows and disappears)
- show a warning on the stream tab
- stop the stream
- something else

@nijikokun

sonicyeti (approved):
Very nice looking state orchestration, glad to be leveraging the context API in such an appropriate way.

packages/insomnia-app/app/ui/components/panes/grpc-request-pane.js
```diff
@@ -110,71 +101,63 @@ const GrpcRequestPane = ({ activeRequest, forceRefreshKey, settings }: Props) =>
             type="text"
             forceEditor
             defaultValue={activeRequest.url}
+            placeholder="grpcb.in:9000"
```
DMarby, line 104: It feels a bit weird to to use an actual service as a placeholder here, in particular since nothing will work without the correct protofile either. 

Not a blocker by any means, but might be worth changing to just an example.com URL or such later, to be more in-line with what we do for other types of requests as well.

cc: @nijikokun

DMarby (approved):
Worked great when I was testing it! :shipit:

develohpanda:
Netlify seems to be failing to deploy storybook, but that's not a blocker to merge.

PR #2814, opened 2020-11-05
Support for server and bidi streaming

develohpanda:
Nothing particular to note. There is a bit of duplication in the grpc network file for each streaming type. It's a little easier to understand, but it would be good to refactor and reuse when this work is revisited.

Fixes INS-282

### DIFF
--- a/packages/insomnia-app/app/network/grpc/index.js
+++ b/packages/insomnia-app/app/network/grpc/index.js
@@ -43,7 +43,6 @@ export const sendUnary = async (requestId: string, respond: ResponseCallbacks): 
 
   // Create client
   const client = _createClient(req, respond);
-
   if (!client) {
     return;
   }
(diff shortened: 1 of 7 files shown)


packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -126,6 +209,22 @@ const _setupStatusListener = (call: Call, requestId: string, respond: ResponseCa
...
 };
 
+const _setupServerStreamListeners = (call: Call, requestId: string, respond: ResponseCallbacks) => {
+  call.on('data', data => respond.sendData(requestId, data));
+
+  call.on('error', (error: ServiceError) => {
+    if (error?.code !== GrpcStatusEnum.CANCELLED) {
+      respond.sendError(requestId, error);
+
+      // Taken through inspiration from other implementation, needs validation
```
DMarby, line 217: Looks reasonable, might be worth filing an issue so we can look into this further in the future?

DMarby (approved):
Awesome 🎉

packages/insomnia-app/app/network/grpc/index.js
develohpanda: Added INS-289 👍

PR #2819, opened 2020-11-07
Update insomnia.less

imtiyazMohammed:


dimitropoulos:
this doesn't seem to apply anymore.  thank you for the PR.  let us know if it does still apply (and, more than that, what the goal in creating it was) and we can reopen.

PR #2820, opened 2020-11-09
Add protofile modal design integration

sonicyeti:
This PR covers the design integration of the gRPC add proto file modal. 

**Notes:**
- Selected state differs from default hover
- Delete icon border removed > set to "danger" color
- All other changes per design spec

**Example:**

**Additional Story for "bordered" instance:**
Screen Shot 2020-11-09 at 10 22 39 AM

Fixes INS-246

PR #2828, opened 2020-11-11
Allow changing the protofile for a request

develohpanda:
The selected method, and the body contents are reset when changing the proto file linked to a request. This PR also adds some edge case handling.

<details>
<summary>Proto file deletion edge case</summary>

@nijikokun there are edge cases that arise on deletion of a proto file, evident especially through this PR. When a request is activated or a message sent, the proto file is loaded again from the contents in the database. But if a proto file has been deleted, those contents are lost, and we can't fetch all of the information required in order to render the UI correctly (the tab title and method drop-downs, for example). In this case, when the protofile for the active request is deleted, the UI is not reset immediately, but on navigating away and returning to the request, the UI will be reset.

</details>

See collapsed comment above for some context about the following changes. The behavior was discussed with @nijikokun before being implemented.

Previously, the methods for a protofile linked to a request were saved in local state, meaning each time the user switched between requests in the sidebar, the methods would be reloaded from the protofile. Because of this, if the protofile for a given request has been deleted and a user tries to activate it, no UI can be rendered because the data no longer exists. Storing the method information itself against a request allows the UX to not be immediately broken (only on the next restart), and has a side-benefit of allowing the business logic and React logic to be more performant.

With this change, an additional action to invalidate the methods was introduced. Invalidating tells React that the next time you need methods, reload them instead of loading from state. Currently, methods are invalidated when a protofile is changed, which is the primary _feature_ introduced by this PR.

I suggest starting your review from the `grpc-reducer.js` file.

QA:
- smoke testing of requests in general: canceling, multiple streams, can make requests etc
- when switching methods, the UI should clear
- when changing protofile, the UI should clear
- if the protofile for a request is deleted, the UI (previous request/response information) should NOT clear
  - it will clear on restart, because it is only stored in memory

Note, if you were to upload the _same_ proto file twice, the application will treat it as two unique files as there is no de-duplication in place.

Fixes INS-211, Fixes INS-272

### DIFF
--- a/packages/insomnia-app/app/network/grpc/proto-loader.js
+++ b/packages/insomnia-app/app/network/grpc/proto-loader.js
@@ -23,7 +23,13 @@ const isServiceDefinition = (obj: Object) => !isTypeOrEnumDefinition(obj);
 //  writing to a file in those cases, but it becomes more important to cache
 //  We also need to think about how to store a reference to a proto file and it's
 //  implications on import/export/sync - INS-271
-export const loadMethods = async (protoFile: ProtoFile): Promise<Array<GrpcMethodDefinition>> => {
+export const loadMethods = async (
+  protoFile: ProtoFile | undefined,
+): Promise<Array<GrpcMethodDefinition>> => {
+  if (!protoFile?.protoText) {
+    return [];
+  }
+
   const tempProtoFile = await writeProtoFile(protoFile.protoText);
   const definition = await protoLoader.load(tempProtoFile, GRPC_LOADER_OPTIONS);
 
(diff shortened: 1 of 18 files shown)


sonicyeti (dismissed):
Very nice! Digging the updated layout as well.

sonicyeti:
@develohpanda Would it be possible to roll in #2820 as part of the response store merge?

DMarby (dismissed):

packages/insomnia-app/app/network/grpc/proto-loader.js
```diff
@@ -16,14 +16,17 @@ const GRPC_LOADER_OPTIONS = {
 const isTypeOrEnumDefinition = (obj: Object) => 'format' in obj; // same check exists internally in the grpc library
 const isServiceDefinition = (obj: Object) => !isTypeOrEnumDefinition(obj);
 
-// TODO: instead of writing to a temp file and loading the protoFile every time methods are required,
-//  add an in-memory caching strategy, indexed by the protoFile._id - INS-272
```
develohpanda, line 20: The file is now loaded and the methods are stored in context, indexed by the request id (not protofile id). These are only reloaded when the protofile for a request is changed, or when the application starts and the request is opened for the first time.

This was done so that React effects ran correctly and no more frequent than necessary. Reloading methods each time the request was activated caused too many edge cases with react effects becoming stale.

sonicyeti (approved):

DMarby (approved):
:shipit:

develohpanda:
Re-reviewed with Niji ✅

PR #2831, opened 2020-11-13
Backwards compitibility in beta sync

develohpanda:
When a new model is introduced and pushed from a particular version of the app, it becomes impossible to pull a snapshot or workspace in an older version of the application, because there are no checks to ensure a model exists before trying to write to it.

I pushed a snapshot from develop, and pulled it in 2020.4.2, and saw the following error.

image

I traced this error to the following piece of code, but only through navigating the code as the stack trace is swallowed https://github.com/Kong/insomnia/blob/8f8acf83f2058abf1e18efccd7964e529bac90a4/packages/insomnia-app/app/common/database.js#L284-L285

Note, this was probably an issue with unit testing models as well. 😅 

I'm not quite sure how to validate this fix... 🤔 re-base it on the commit for 2020.4.2 and try to pull a workspace or snapshot that contains gRPC models, and see how it behaves?

**Do we need to address this backwards compatibility case at all? (I think we do)**

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
@@ -4,7 +4,7 @@ import autobind from 'autobind-decorator';
 import classnames from 'classnames';
 import { Dropdown, DropdownButton, DropdownDivider, DropdownItem } from '../base/dropdown';
 import type { Workspace } from '../../../models/workspace';
-import { showAlert, showModal } from '../modals';
+import { showError, showModal } from '../modals';
 import SyncStagingModal from '../modals/sync-staging-modal';
 import HelpTooltip from '../help-tooltip';
 import Link from '../base/link';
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
```diff
@@ -234,9 +233,9 @@ class SyncDropdown extends React.PureComponent<Props, State> {
       const delta = await vcs.rollbackToLatest(syncItems);
       await db.batchModifyDocs(delta);
     } catch (err) {
-      showModal(ErrorModal, {
+      showError({
         title: 'Revert Error',
-        message: err.message,
+        error: err,
```
develohpanda, line 238: This is more tidy-up, given we have a specific helper and modal with options to show errors and stack traces. By only sending the error (and no message), the message and stack trace will automatically be extracted and shown independently.

`showError` is used in a number of other places - git sync, plugins, etc.

I can remove this change if need be.

sonicyeti (approved):

develohpanda:
Closing this PR, because we likely want to do a server-side fix going forward cc: @DMarby

PR #2832, opened 2020-11-13
Support reuploading a proto file

develohpanda:
When re-uploading, all requests that use that proto file need to be invalidated, so that they refresh their methods the next time they are loaded. Requests that are running, but have had their protofile changed, are cancelled.


sonicyeti (approved):
Nice, digging the upload iconography usage as well.

DMarby:
> What if a request is running? After the protoFile or protoFile contents are changed, the request state is cleared (thus, if something is running, the cancel button will disappear). **Do we cancel the request? **

I think it makes sense to cancel a request if the protofile for it changes.

packages/insomnia-app/app/ui/context/grpc/grpc-context.js
```diff
@@ -55,3 +55,17 @@ export const useGrpc = (requestId: string): [GrpcRequestState, GrpcDispatch] =>
...
+type GrpcContextModalWrapperProps = {|
+  children: (dispatch: GrpcDispatch) => React.Node,
+|};
+
+export const GrpcDispatchModalWrapper = ({ children }: GrpcContextModalWrapperProps) => {
+  const dispatch = useGrpcDispatch();
+
+  React.useEffect(() => {
+    console.log('mounted');
+  }, []);
```
DMarby: Is this just leftover debug code, or does something need to happen here?

packages/insomnia-app/app/common/select-file-or-folder.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+  if (types.includes('directory')) {
+    title += ' Directory';
+  }
+  const options = {
+    title: title,
+    buttonLabel: 'Select',
+    properties: types.map(type => {
+      console.log('TYPE', type);
```
DMarby: I know this was already in the FileInputButton component, but would be nice to get rid of this debug log

DMarby (approved):
Looks good, just two small nitpicks 👍

packages/insomnia-app/app/ui/context/grpc/grpc-context.js
develohpanda: Oh damn, yes leftover. I was checking to make sure it's not remounting continuously because of the wrapper, and it's not. :)

packages/insomnia-app/app/common/select-file-or-folder.js
develohpanda: Ya good idea, I'll get rid of it

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.js
```diff
@@ -62,7 +54,7 @@ const GrpcRequestPane = ({ activeRequest, forceRefreshKey, settings }: Props) =>
             type="text"
             forceEditor
             defaultValue={activeRequest.url}
-            placeholder="example.com:9000"
+            placeholder="brpcb.in:9000"
```
develohpanda: As per Niji's request

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.js
```diff
@@ -62,7 +54,7 @@ const GrpcRequestPane = ({ activeRequest, forceRefreshKey, settings }: Props) =>
             type="text"
             forceEditor
             defaultValue={activeRequest.url}
-            placeholder="example.com:9000"
+            placeholder="grpcb.in:9000"
```
develohpanda, line 57: As per Niji's request

develohpanda:
In order to properly handle cancellation and the various request states, and to remove some of the jumpy behavior (see gif above when changing requests), I needed to do a small amount of refactoring. I've added some additional unit tests for the high-touch custom react hooks as well. cc: @DMarby

DMarby (approved):

PR #2837, opened 2020-11-15
Fix for yarn warning messages wrongly treated as errors

### LINKED ISSUE #2836: Yarn fails plugin installation over a warning (e.g deprecated dependency)
**Describe the bug**
Hi, i have been developing a plugin on Insomnia Designer but i'm experiencing a bug while installing it from the npm registry after have published it. It stops when there is a warning for a deprecated dependency. The build pass on GitHub Action so i don't understand what i'm doing wrong and why it should throw an error over a warning message. I have seen that if i install with yarn directly inside the /plugin folder it works but if it has to install the dependency (node_modules) alone it goes in error.

NPM: https://www.npmjs.com/package/insomnia-plugin-pdf-report
Repository: https://github.com/MrSnix/insomnia-plugin-pdf-report

**To Reproduce**
1. Go to `Preferences -> Plugins`
2. Insert on the input field `insomnia-plugin-pdf-report`
3. Click on `Install Plugin`
4. Watch it fails and see error

**Expected behavior**
It probably shouldn't fail to install on a warning.

**Screenshots**
Error

**Desktop:**
Version: Insomnia Designer 2020.4.2
Release date: 21/10/2020
OS: Darwin x64 20.1.0
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64

**Additional context**
None

MrSnix:
The only thing i added was a check on the stderr in order to prevent warning messages to be wrongly treated as errors by Insomnia.


packages/insomnia-app/app/plugins/install.js
```diff
@@ -159,7 +159,8 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
           return;
         }
 
-        if (stderr) {
+        // If we do not check on 'warning', the plugin setup may fail for any generic message
+        if (stderr && !stderr.toString().includes('warning')) {
```
develohpanda: Two things about this: 

1. I'd prefer this check to be much more explicit, something like what you suggested in Slack. Some regex to match a string that starts with "warning", contains some text in between, then continues with "no longer maintained".

2. This condition will ignore a valid error, if _any_ warnings exist in `stderr`, so it likely needs to be a little smarter to only catch exceptional cases. Potentially splitting on new lines, and ensuring all lines match the above check?

develohpanda (changes requested):
Couple of things to think about here 🙂 Thanks for the PR (and your awesome plugin!)

packages/insomnia-app/app/plugins/install.js
MrSnix: I'll work on it and i'll try to elaborate something better as you suggested. Thanks for feedback

PR #2840, opened 2020-11-16
Fix for random UI crash

### LINKED ISSUE #2805: UI crash
**Describe the bug**
Sometime I have crash in UI. See attachment screenshot.
**To Reproduce**
Sometime
**Expected behavior**
Release date: 10.09.2020
OS: Linux x64 5.8.0-9.3-liquorix-amd64
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.69.1 OpenSSL/1.1.1g zlib/1.2.11 brotli/1.0.7 libssh2/1.9.0 nghttp2/1.41.0
**Screenshots**
Снимок экрана от 2020-11-02 11-22-17

 - OS: ubuntu 18.04
 - Installation Method: snap
 - App Version 2020.4.0

MrSnix:
After some analysis i found out the main problem it's related to the `from.line` which has an `undefined` value in some situations (which i was not able to reproduce). `foldOptions.widget(from, to)` is called internally by CodeMirror when you want to fold code inside the editor itself. This makes React crash when the "Send" button is pressed because the CodeMirror state is not consistent due to invalid folding triggering an infinite setState.

I was thinking about simply return and do not execute any folding on the editor but it will create another inconsistency due to CodeMirror attaching an eventListener on an invalid point. So maybe throwing an error could be the best option.
At the least CodeMirror won't make the app crash when it's not able to retrieve from which line the folding happens and won't attach eventListener on invalid point.

This my proposal for a fix, it closes #2805

Any feedback?


develohpanda (changes requested):
Thanks for investigating a submitting a PR! 🎉 

I think that instead of throwing an error, we simply return `'\u2194'` (the current fallback, see line 330) if either `from` or `to` are falsey, and allow CodeMirror to handle it internally. The event listener would likely be created regardless, because this code doesn't actually do any folding, it only provides the text content. There will be an arrow in the gutter (see below) to fold/unfold regardless.

image

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -312,7 +312,11 @@ class CodeEditor extends React.Component {
         let startToken = '{';
         let endToken = '}';
 
+        // Prevent retrieving an invalid content if undefined
+        if (!from.line || !to.line) return '\u2194';
```
develohpanda: Unfortunately, if `from` or `to` are undefined, this will throw an error. This will need to be changed to:

```suggestion
        // Prevent retrieving an invalid content if undefined
        if (!from?.line || !to?.line) return '\u2194';
```

packages/insomnia-app/app/ui/components/codemirror/code-editor.js
```diff
@@ -312,7 +312,11 @@ class CodeEditor extends React.Component {
...
+
         const prevLine = this.codeMirror.getLine(from.line);
+
         if (prevLine.lastIndexOf('[') > prevLine.lastIndexOf('{')) {
```
develohpanda: The initial reported error was that `lastIndexOf` could not be found on `prevLine`, so there needs to be another check here to make sure to exit if `prevLine` is undefined.

develohpanda (changes requested):
Nice 😄 a few more changes please!

develohpanda (approved):
🎉 LGTM, thanks for the PR

DMarby (approved):
Looks good, thanks for investigating and fixing this!

PR #2846, opened 2020-11-18
Add GrpcRequest and ProtoFile as accepted models for git sync

develohpanda:
image

### DIFF
--- a/packages/insomnia-app/app/sync/git/__tests__/ne-db-plugin.test.js
+++ b/packages/insomnia-app/app/sync/git/__tests__/ne-db-plugin.test.js
@@ -40,6 +40,8 @@ describe('NeDBPlugin', () => {
         models.unitTest.type,
         models.unitTestSuite.type,
         models.workspace.type,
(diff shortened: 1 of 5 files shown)


DMarby (approved):
Looks good overall, worked well when I tried it!

One small thing found while testing is that it's displaying the wrong type in the UI, in that it's showing Request instead of gRPC Request.

I believe this is due to the code below, since we're matching on the path, `.insomnia/GrpcRequest/<foobar>.yaml` will match on the Request type model before getting to GrpcRequest:

https://github.com/Kong/insomnia/blob/ec8f72c09b3637c09f9584ee182bb379f6397351/packages/insomnia-app/app/ui/components/modals/git-staging-modal.js#L213-L218

develohpanda:
Oh I completely missed that in the screenshot... Nice catch and notes. I'll fix this before merging.

DMarby (approved):
:shipit:

develohpanda:
Rebased onto the release branch for 2020.5

PR #2847, opened 2020-11-18
E2E test to install and consume a plugin

develohpanda:
image

We currently don't run packaged tests during a release. Plugin installation requires yarn and it is non-trivial to access it using the build artifacts, so it's easier to test this with the packaged app, hence the conditional test.

Fixes INS-305

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,8 @@
     "preinstall": "node ./scripts/check-version.js",
     "app-build:smoke:core": "cross-env SMOKE_TEST=core npm run app-build",
     "app-build:smoke:designer": "cross-env SMOKE_TEST=designer npm run app-build",
+    "app-package:smoke:core": "cross-env SMOKE_TEST=core npm run app-package",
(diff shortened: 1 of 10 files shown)


DMarby (approved):
Nice, ran perfectly on my machine as well 👍 

Not introduced in this PR (or a blocker for it), but something I noticed when packaging the apps for the smoke tests is that we also produce the .dmg's and the .zip's, which adds quite a bit of extra time to the packaging.
Do you think it might make sense to have electron-builder only package the `.app` and nothing else for smoke tests in the future?

image

develohpanda:
Ooooooh, actually yes! 

In the current state packaging takes 3x longer than building. If we reduce the artifacts and it speeds up, we can package during CI (instead of build during CI). That way tests like this one can run with PRs. It would also reduce the scripts (I really don't like how many scripts there are for smoke tests).

So, worth testing if packaging speeds up significantly by restricting the artifacts. If it doesn't, package tests will run during a release, in which case we need to package everything anyway.

develohpanda:
I added INS-330 to track this.

PR #2853, opened 2020-11-20
Hide the active workspace when moving a folder

### LINKED ISSUE #2849: Hide the active workspace when moving a folder to a new workspace
**Describe the bug**
Moving a folder shows a modal allowing you to select the new workspace, but it also shows the active workspace in that dropdown. This can be quite misleading and confusing, so let's hide it from the dropdown.

**To Reproduce**
1. Create a new workspace (so you have two workspaces with **unique names**)
1. Create a new folder
2. Right click on the folder
3. Click 'Move'
4. Click on the dropdown to select the destination workspace
5. Note that it also shows your current workspace in the dropdown

**Expected behavior**
Don't show the active workspace in the move folder modal

**Screenshots**

In the following screenshot, "Insomnia" should not show in the dropdown.

image

```
Version: Insomnia Core 2020.5.0-beta.1
Release date: 11/18/2020
OS: Windows_NT x64 10.0.19041
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.69.1-DEV OpenSSL/1.1.1d zlib/1.2.11 WinIDN libssh2/1.9.0_DEV nghttp2/1.40.0
```

gravyboat:
Closes #2849

Before:
before

After:
after

This uses a `flatMap` which should have minimal impact on performance. I didn't see any tests for this specific component. If I missed them let me know and I can add some.

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
@@ -12,7 +12,7 @@ import HelpTooltip from '../help-tooltip';
 
 type Props = {
   workspaces: Array<Workspace>,
-  workspace: Workspace,
+  activeWorkspace: Workspace,
 };
 
 type State = {
@@ -79,7 +79,7 @@ class MoveRequestGroupModal extends React.PureComponent<Props, State> {
   }
 
   render() {
-    const { workspaces, workspace } = this.props;
+    const { workspaces, activeWorkspace } = this.props;
     const { selectedWorkspaceId } = this.state;
     return (
       <form onSubmit={this._handleSubmit}>
@@ -93,7 +93,7 @@ class MoveRequestGroupModal extends React.PureComponent<Props, State> {
                 <select onChange={this._handleChangeSelectedWorkspace} value={selectedWorkspaceId}>
                   <option value="n/a">-- Select Workspace --</option>
                   {workspaces.flatMap(w =>
-                    w._id === workspace._id ? (
+                    w._id === activeWorkspace._id ? (
                       []
                     ) : (
                       <option key={w._id} value={w._id}>
--- a/packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
+++ b/packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
@@ -12,6 +12,7 @@ import HelpTooltip from '../help-tooltip';
 
 type Props = {
   workspaces: Array<Workspace>,
+  workspace: Workspace,
 };
 
 type State = {
@@ -78,7 +79,7 @@ class MoveRequestGroupModal extends React.PureComponent<Props, State> {
   }
 
   render() {
-    const { workspaces } = this.props;
+    const { workspaces, workspace } = this.props;
     const { selectedWorkspaceId } = this.state;
     return (
       <form onSubmit={this._handleSubmit}>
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
```diff
@@ -91,11 +92,15 @@ class MoveRequestGroupModal extends React.PureComponent<Props, State> {
...
-                  ))}
+                  {workspaces.flatMap(w =>
+                    w._id === workspace._id ? (
+                      []
+                    ) : (
+                      <option key={w._id} value={w._id}>
+                        {w.name}
+                      </option>
+                    ),
+                  )}
```
develohpanda: I'm guessing you didn't go down the `workspaces.filter(...).map(...)` route to avoid iterating through the array twice: 👏.

Is there a particular reason to use `flatMap` over `map`, though? If you use a `map`, you can simply return `null` in place of an empty array in the conditional.

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
```diff
@@ -12,6 +12,7 @@ import HelpTooltip from '../help-tooltip';
 
 type Props = {
   workspaces: Array<Workspace>,
+  workspace: Workspace,
```
develohpanda: Please change this prop to be `activeWorkspace` to be clearer about what it represents. 🙂

develohpanda (changes requested):
Thanks for the PR! A couple of comments. 🎉

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
gravyboat: I considered this when making the PR but wasn't sure since it was labeled `workspace` in other instances, thanks for the clarification. This is now fixed for all references.

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
gravyboat: > Is there a particular reason to use `flatMap` over `map`, though?

Not really other than personal preference. I don't like using null when I can avoid it and since it's already an array of workspaces here it seems more clear to provide an array. Performance wise it's nearly identical:
flatmap_v_map

Edit: If you prefer a `map` I can update it.

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
develohpanda: Oh, you're right, it is labeled simply `workspace` in other places. I don't know why... since it loses context. I'm not too concerned, but I think keeping context about what the workspace represents is useful for this particular case! 👍

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
develohpanda: That's good to know! I don't really have a preference, I'm on board with avoiding `null` where possible, and I can see cases where `flatMap` will certainly be safer.

I feel that for the sake of consistency in this code-base (only one instance of `flatMap`), and the fact that React knows exactly what to do with a `null`, in this particular instance it might be better to stick with `map`. 🙂 

Btw, what did you use to get that performance snapshot? Looks useful!

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
gravyboat: Agreed!

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
gravyboat: Good point about React knowing what to do with `null` for the conditional rendering. I didn't check to see if `flatMap` had been used elsewhere in the project and this being the only instance would be annoying for maintenance. I've updated the PR with this change. If things look good after that I'll squash the 3 commits in to one.

> Btw, what did you use to get that performance snapshot? Looks useful!

It's https://jsbench.me, I wouldn't use it for anything large, but it's great for testing in these sorts of scenarios.

develohpanda (approved):
Nicely done! 💯

packages/insomnia-app/app/ui/components/modals/move-request-group-modal.js
develohpanda: We squash and merge PRs through GH, so no need to squash into a single commit 😊

DMarby (approved):

gravyboat:
Did a quick merge in of the develop branch to get back up to date since I wasn't sure if you prefer that or a rebase since it will be squashed any way.

PR #2861, opened 2020-11-23
Yarn warnings wrongly treated as errors

### LINKED ISSUE #2836: Yarn fails plugin installation over a warning (e.g deprecated dependency)

MrSnix:
Changes:

0. Modified if condition from `if (stderr)` to `if (stderr && !isFalsePositive(stderr))`
1. Added function `isFalsePositive(std_err)` which takes std as input, split it using `\r?\n` and run `isDeprecatedDependencies(str)` on each line.
2. Added function `isDeprecatedDependencies(str)` which checks if a given message is a plain warning, 
it's commented to better explains what it does.

Feedbacks ?

It resolves #2836

### DIFF
--- a/packages/insomnia-app/app/__tests__/install.test.js
+++ b/packages/insomnia-app/app/__tests__/install.test.js
@@ -1,42 +1,71 @@
-import { isDeprecatedDependencies } from '../plugins/install';
+import { containsOnlyDeprecationWarnings, isDeprecatedDependencies } from '../plugins/install';
 
 describe('install.js', () => {
+  describe('containsOnlyDeprecationWarning', () => {
+    it('should return true when all lines in stderr are deprecation warnings', () => {
+      const stderr =
+        // Warning #1
+        'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.\r\n' +
+        // Warning #2
+        'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.\n' +
+        // Warning #3
+        'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.';
+      expect(containsOnlyDeprecationWarnings(stderr)).toBe(true);
+    });
+    it('should return false when stderr contains a deprecation warning and an error', () => {
+      const stderr =
+        // Warning #1
+        'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.\r\n' +
+        // Error #1
+        'error https://npm.example.net/@types%example/-/nello-1.3.5.tgz:' +
+        'Integrity check failed for "@types/example"' +
+        '(computed integrity doesn\'t match our records, got "sha512-z4kkSfaPg==")\n' +
+        // Warning #2
+        'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.';
+      expect(containsOnlyDeprecationWarnings(stderr)).toBe(false);
+    });
+  });
   describe('isDeprecatedDependencies', () => {
-    it('invalid-dataset-0', () => {
+    it('should not match when the message is falsy', () => {
       expect(isDeprecatedDependencies('')).toBe(false);
-    });
-    it('invalid-dataset-1', () => {
       expect(isDeprecatedDependencies(null)).toBe(false);
-    });
-    it('invalid-dataset-2', () => {
       expect(isDeprecatedDependencies(undefined)).toBe(false);
     });
-    it('valid-warning-0', () => {
+    it('should match with multiple nested dependencies', () => {
       const msg =
         'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
         'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
         'Please, upgrade your dependencies to the actual version of xyz.';
       expect(isDeprecatedDependencies(msg)).toBe(true);
     });
-    it('valid-warning-1', () => {
+    it('should match with one nested dependency', () => {
       const msg =
         'warning insomnia-plugin-xxx-yyy > xyz: ' +
         'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
         'Please, upgrade your dependencies to the actual version of xyz.';
       expect(isDeprecatedDependencies(msg)).toBe(true);
     });
-    it('invalid-warning', () => {
+    it('should not match with an unrelated warning', () => {
       const msg =
-        'warning You are using Node "6.0.0" which is not supported and may encounter bugs or unexpected behaviour.' +
+        'warning You are using Node "6.0.0" which is not supported and may encounter bugs or unexpected behaviour. ' +
         'Yarn supports the following server range: "^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0"';
       expect(isDeprecatedDependencies(msg)).toBe(false);
     });
-    it('invalid-error', () => {
+    it('should not match with an unrelated error', () => {
       const msg =
-        'error https://npm.example.net/@types%example/-/hello-1.3.5.tgz: ' +
+        'error https://npm.example.net/@types%example/-/nello-1.3.5.tgz: ' +
         'Integrity check failed for "@types/example" ' +
-        "(computed integrity doesn't match our records, " +
-        'got "sha512-z4kkSfaPg==")';
+        '(computed integrity doesn\'t match our records, got "sha512-z4kkSfaPg==")';
       expect(isDeprecatedDependencies(msg)).toBe(false);
     });
   });
--- a/packages/insomnia-app/app/__tests__/install.test.js
+++ b/packages/insomnia-app/app/__tests__/install.test.js
@@ -3,16 +3,13 @@ import { isDeprecatedDependencies } from '../plugins/install';
 describe('install.js', () => {
   describe('isDeprecatedDependencies', () => {
     it('invalid-dataset-0', () => {
-      const msg = '';
-      expect(isDeprecatedDependencies(msg)).toBe(false);
+      expect(isDeprecatedDependencies('')).toBe(false);
     });
     it('invalid-dataset-1', () => {
-      const msg = null;
-      expect(isDeprecatedDependencies(msg)).toBe(false);
+      expect(isDeprecatedDependencies(null)).toBe(false);
     });
     it('invalid-dataset-2', () => {
-      const msg = undefined;
-      expect(isDeprecatedDependencies(msg)).toBe(false);
+      expect(isDeprecatedDependencies(undefined)).toBe(false);
     });
     it('valid-warning-0', () => {
       const msg =
--- a/packages/insomnia-app/app/__tests__/install.test.js
+++ b/packages/insomnia-app/app/__tests__/install.test.js
@@ -2,6 +2,18 @@ import { isDeprecatedDependencies } from '../plugins/install';
 
 describe('install.js', () => {
   describe('isDeprecatedDependencies', () => {
+    it('invalid-dataset-0', () => {
+      const msg = '';
+      expect(isDeprecatedDependencies(msg)).toBe(false);
+    });
+    it('invalid-dataset-1', () => {
+      const msg = null;
+      expect(isDeprecatedDependencies(msg)).toBe(false);
+    });
+    it('invalid-dataset-2', () => {
+      const msg = undefined;
+      expect(isDeprecatedDependencies(msg)).toBe(false);
+    });
     it('valid-warning-0', () => {
       const msg =
         'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
--- a/packages/insomnia-app/app/__tests__/install.test.js
+++ b/packages/insomnia-app/app/__tests__/install.test.js
@@ -1,0 +1,34 @@
+import { isDeprecatedDependencies } from '../plugins/install';
+
+describe('install.js', () => {
+  describe('isDeprecatedDependencies', () => {
+    it('valid-warning-0', () => {
+      const msg =
+        'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.';
+      expect(isDeprecatedDependencies(msg)).toBe(true);
+    });
+    it('valid-warning-1', () => {
+      const msg =
+        'warning insomnia-plugin-xxx-yyy > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.';
+      expect(isDeprecatedDependencies(msg)).toBe(true);
+    });
+    it('invalid-warning', () => {
+      const msg =
+        'warning You are using Node "6.0.0" which is not supported and may encounter bugs or unexpected behaviour.' +
+        'Yarn supports the following server range: "^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0"';
+      expect(isDeprecatedDependencies(msg)).toBe(false);
+    });
+    it('invalid-error', () => {
+      const msg =
+        'error https://npm.example.net/@types%example/-/hello-1.3.5.tgz: ' +
+        'Integrity check failed for "@types/example" ' +
+        "(computed integrity doesn't match our records, " +
+        'got "sha512-z4kkSfaPg==")';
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
   });
 }
 
+function isFalsePositive(stderr) {
+  // Let's assume is a false positive
+  let isFalsePositive: boolean = true;
+  // Split into array and remove falsy values (null, undefined, 0, -0, NaN, "", false)
```
develohpanda:
```suggestion
  // Split on line breaks and remove falsy values (null, undefined, 0, -0, NaN, "", false)
```

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
...
+  const arr = stderr.split(/\r?\n/).filter(e => e);
+  // arr.find() stops and return the first match which is not a deprecated dependency warning
```
develohpanda:
```suggestion
  // arr.find() stops and return the first match which is not a deprecated dependency warning. If there are any other warnings/errors, we should exit as normal.
```

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
...
+/**
+ * Provided a string, it checks for the following message:<br>
+ * <<[warning] xxx > yyy > zzz: yyy<n is [no longer maintained] and [not recommended for usage] <br>
+ * due to the number of issues. Please, [upgrade your dependencies] to xxx>> <br>
+ * @param str The error message
+ * @returns {boolean} Returns true if it's a deprecated warning
+ */
+function isDeprecatedDependencies(str: string): boolean {
+  // Search for: warning + dep-list + issue
+  const regex = /(?<keyword>warning)(?<dependencies>[^>:].+[>:])(?<issue>.+)/gm;
```
develohpanda: 1. Because `isDeprecatedDependencies` can be called multiple times as part of `arr.find()`, we shouldn't re-create this regex every time. _I think_, if you declare it outside the function and remove the `global, g` flag from the end of it, you should have the same result.

2. Given you're using the `multi-line, m` flag, do you need to split `stderr` on new lines as well? And conversely, if you are splitting on new lines, the regex doesn't need the multi-line flag?

I _think_ there's some scope for improvement here with splitting into individual lines, and the flags used on the regex.

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
   });
 }
 
+function isFalsePositive(stderr) {
```
develohpanda: Could you please export this function and add some unit tests for it? Place the tests in `packages/insomnia-app/app/plugins/__tests__/install.test.js`.

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
...
+ * Provided a string, it checks for the following message:<br>
+ * <<[warning] xxx > yyy > zzz: yyy<n is [no longer maintained] and [not recommended for usage] <br>
+ * due to the number of issues. Please, [upgrade your dependencies] to xxx>> <br>
+ * @param str The error message
+ * @returns {boolean} Returns true if it's a deprecated warning
+ */
+function isDeprecatedDependencies(str: string): boolean {
+  // Search for: warning + dep-list + issue
+  const regex = /(?<keyword>warning)(?<dependencies>[^>:].+[>:])(?<issue>.+)/gm;
+  let result = false;
```
develohpanda: I suggest avoiding `let` where possible.

In this case, you can just `return true` or `return false` where needed, instead of setting this variable.

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
   });
 }
 
+function isFalsePositive(stderr) {
+  // Let's assume is a false positive
+  let isFalsePositive: boolean = true;
+  // Split into array and remove falsy values (null, undefined, 0, -0, NaN, "", false)
+  const arr = stderr.split(/\r?\n/).filter(e => e);
+  // arr.find() stops and return the first match which is not a deprecated dependency warning
+  if (arr.find(e => !isDeprecatedDependencies(e))) isFalsePositive = false;
```
develohpanda: I think it might be useful to atleast print _all_ deprecation warnings to the console, so we don't swallow them altogether.

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
...
+  const regex = /(?<keyword>warning)(?<dependencies>[^>:].+[>:])(?<issue>.+)/gm;
+  let result = false;
+  // The array[3](issue) contains the message as it is without the dependency list
+  const message = regex.exec(str)[3];
+  // Strict check, everything must be matched to be a false positive
+  if (
+    message &&
+    message.includes('no longer maintained') &&
+    message.includes('not recommended for usage') &&
+    message.includes('upgrade your dependencies')
```
develohpanda, line 202: Could this be included as part of the regex?

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +170,43 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
...
+ * due to the number of issues. Please, [upgrade your dependencies] to xxx>> <br>
+ * @param str The error message
+ * @returns {boolean} Returns true if it's a deprecated warning
+ */
+function isDeprecatedDependencies(str: string): boolean {
+  // Search for: warning + dep-list + issue
+  const regex = /(?<keyword>warning)(?<dependencies>[^>:].+[>:])(?<issue>.+)/gm;
+  let result = false;
+  // The array[3](issue) contains the message as it is without the dependency list
+  const message = regex.exec(str)[3];
```
develohpanda: Since you're using named capture groups (very nice, they make the regex somewhat understandable), you can access the result using names as well.

image

develohpanda (changes requested):
Nice update from the previous PR, and useful comments 🙂 

I've outlined a bunch of questions/comments below, let me know how you get on or if you have any questions!

packages/insomnia-app/app/plugins/install.js
MrSnix: I keep suggesting to leave this out from the regex.
I think regex should be used to match only the general layout of the message rather than the content.
I feel like it could be messy, but i leave the choice to you. Let me know.

MrSnix:
**Changes**:

0. The regex now is compiled outside the function as global variable
1. I removed the flags `/gm` from regex
2. Modified comments as requested
3. Warning are now displayed with console.log() (we won't hide them)
4. Using the dot notation to refers to regex groups
5. Created `install.test.js` with some tests which validate the exported `isDeprecatedDependencies(str)`

Feedbacks ?

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +172,37 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
   });
 }
 
+function isFalsePositive(stderr) {
```
develohpanda: Instead of naming this function `isFalsePositive` I suggest `containsOnlyDeprecationWarnings` for clarity.

This also needs some tests:
1. `should return true when all lines in stderr are deprecation warnings`
2. `should return false when stderr contains a deprecation warning and an error`

packages/insomnia-app/app/__tests__/install.test.js
```diff
@@ -0,0 +1,43 @@
...
+  describe('isDeprecatedDependencies', () => {
+    it('invalid-dataset-0', () => {
+      expect(isDeprecatedDependencies('')).toBe(false);
+    });
+    it('invalid-dataset-1', () => {
+      expect(isDeprecatedDependencies(null)).toBe(false);
+    });
+    it('invalid-dataset-2', () => {
+      expect(isDeprecatedDependencies(undefined)).toBe(false);
+    });
```
develohpanda: These can be combined into one test named `should not match when the message is falsey`

packages/insomnia-app/app/__tests__/install.test.js
```diff
@@ -0,0 +1,43 @@
...
+    it('valid-warning-0', () => {
```
develohpanda:
```suggestion
    it('should match with multiple nested dependencies', () => {
```

packages/insomnia-app/app/__tests__/install.test.js
```diff
@@ -0,0 +1,43 @@
...
+      const msg =
+        'warning insomnia-plugin-xxx-yyy > xyz > xyz > xyz > xyz > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.';
+      expect(isDeprecatedDependencies(msg)).toBe(true);
+    });
+    it('valid-warning-1', () => {
```
develohpanda:
```suggestion
    it('should match with one nested dependency', () => {
```

packages/insomnia-app/app/__tests__/install.test.js
```diff
@@ -0,0 +1,43 @@
...
+      const msg =
+        'warning insomnia-plugin-xxx-yyy > xyz: ' +
+        'xyz is no longer maintained and not recommended for usage due to the number of issues. ' +
+        'Please, upgrade your dependencies to the actual version of xyz.';
+      expect(isDeprecatedDependencies(msg)).toBe(true);
+    });
+    it('invalid-warning', () => {
```
develohpanda:
```suggestion
    it('should not match with an unrelated warning', () => {
```

packages/insomnia-app/app/__tests__/install.test.js
```diff
@@ -0,0 +1,43 @@
...
+      const msg =
+        'warning You are using Node "6.0.0" which is not supported and may encounter bugs or unexpected behaviour.' +
+        'Yarn supports the following server range: "^4.8.0 || ^5.7.0 || ^6.2.2 || >=8.0.0"';
+      expect(isDeprecatedDependencies(msg)).toBe(false);
+    });
+    it('invalid-error', () => {
```
develohpanda:
```suggestion
    it('should not match with an unrelated error', () => {
```

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +172,37 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
...
+  // Split on line breaks and remove falsy values (null, undefined, 0, -0, NaN, "", false)
+  const arr = stderr.split(/\r?\n/).filter(e => e);
+  // Retrieve all matching deprecated dependency warning
+  const warnings = arr.filter(e => isDeprecatedDependencies(e));
+  // Print each deprecation warnings to the console, so we don't hide them.
+  warnings.forEach(e => console.log(e));
```
develohpanda: We tend to scope console messages, so I propose the following change. Could you please add a screenshot for messages in the console after you try to install your plugin?

```suggestion
  warnings.forEach(e => console.log('[plugins] deprecation warning during installation: ', e));
```

packages/insomnia-app/app/plugins/install.js
```diff
@@ -170,6 +172,37 @@ async function _installPluginToTmpDir(lookupName: string): Promise<{ tmpDir: str
   });
 }
 
+function isFalsePositive(stderr) {
+  // Split on line breaks and remove falsy values (null, undefined, 0, -0, NaN, "", false)
+  const arr = stderr.split(/\r?\n/).filter(e => e);
+  // Retrieve all matching deprecated dependency warning
+  const warnings = arr.filter(e => isDeprecatedDependencies(e));
```
develohpanda, line 179: These filters can be combined to reduce the number of times we iterate through the array

```suggestion
  const arr = stderr.split(/\r?\n/);
  // Retrieve all matching deprecated dependency warning
  const warnings = arr.filter(e => e && isDeprecatedDependencies(e));
```

develohpanda (changes requested):
Nice tests; just need to think about naming tests the same way one would think about function names. 🙂

MrSnix:
**Changes**:

0. `isFalsePositive` has been renamed to `containsOnlyDeprecationWarnings`
1. As requested tests have been provided for `containsOnlyDeprecationWarnings`
2. As requested some tests have been unified
3. Tests now have meaningful names
4. As requested output on the console has been modified
5. There is no sixth item in this list

Feedbacks?

packages/insomnia-app/app/plugins/install.js
MrSnix: I cannot do that otherwise the algorithm mismatch the array length and the function returns `false` when it shouldn't.
- **If i do not apply** the last filter, `arr` will contain [msg, null] and `warnings` [msg] and it's mismatched.
- **If i do apply** the last filter, arr will contain [msg] and warnings [msg] and it's a match.

packages/insomnia-app/app/plugins/install.js
MrSnix: image
**Here is it**

packages/insomnia-app/app/plugins/install.js
develohpanda: Ah, that makes sense! 👍

packages/insomnia-app/app/plugins/install.js
develohpanda: Thanks! We should probably change that to `console.warn` instead of `console.log` so it applies colors accordingly. But looks good!

develohpanda (approved):
LGTM 🎉 tested locally with `insomnia-plugin-pdf-report` (has a warning) and `insomnia-plugin-faker` (has no warning).

One note about changing `console.log` to `console.warn` but rest is good!

DMarby (approved):

t0mpl:
Hi everyone, I'm using Insomnia v2020.5.2 and i still have this issue:

Error: Yarn error warning insomnia-plugin-aws-cognito-iam > @aws-amplify/auth > amazon-cognito-identity-js > buffer@4.9.1: This version of 'buffer' is out-of-date. You must update to v4.9.2 or newer
warning "insomnia-plugin-aws-cognito-iam > @aws-amplify/auth > @aws-amplify/core > @aws-sdk/client-cognito-identity > @aws-sdk/middleware-retry > react-native-get-random-values@1.5.0" has unmet peer dependency "react-native@>=0.56".

    at file:///Applications/Insomnia.app/Contents/Resources/app.asar/bundle.js:103518:16
    at ChildProcess.exithandler (child_process.js:295:7)
    at ChildProcess.emit (events.js:223:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Socket.<anonymous> (internal/child_process.js:430:11)
    at Socket.emit (events.js:223:5)
    at Pipe.<anonymous> (net.js:664:12)

Wasn't it suppose to be fix in the last release?

develohpanda:
Hi @ThomasP1988, this was merged but not released with 2020.5.2. It will be included with the next release, likely early 2021.

t0mpl:
Ok thank you, I will wait until there then

nickroberts:
Still having an issue with this. I'm on the latest, 2021.1.1.

Trying to install `insomnia-plugin-reponse`.

```
Error: Yarn error warning insomnia-plugin-response > insomnia-xpath > xmldom@0.1.31: Deprecated due to CVE-2021-21366 resolved in 0.5.0

    at file:///Applications/Insomnia.app/Contents/Resources/app.asar/bundle.js:102243:16
    at ChildProcess.exithandler (child_process.js:295:7)
    at ChildProcess.emit (events.js:223:5)
    at maybeClose (internal/child_process.js:1021:16)
    at Socket.<anonymous> (internal/child_process.js:430:11)
    at Socket.emit (events.js:223:5)
    at Pipe.<anonymous> (net.js:664:12)
```

MrSnix:
This happens because **the check was made really strict on purpose**.

It passes only if the message is shaped as follows: 

[Message format]
_Yarn warnings [dependencies-list]: 
"The following x is **no longer maintained** and **not recommended for usage**, 
please **upgrade your dependencies**."_

From the fix:
```js
    message.includes('no longer maintained') &&
    message.includes('not recommended for usage') &&
    message.includes('upgrade your dependencies')
```

**I could make it less strict** but there were some concerns that could lead to false positives.

**EDIT**: 
To be more clear, we don't really know what Yarn may emit as a message, it may be a real error or something else like a deprecation warning. Currently, it should fail for any given warning message, so we may implement an array of keywords to match inside the warning message, like "out-of-date" or "deprecated" and so on... but there is always some sort of risks with false positives.

develohpanda:
@nickroberts looking closer at the error you've got, this is actually an issue with a downstream package that Insomnia maintains, being consumed by an external plugin. This should be addressed by an open dependabot ticket and I'll look to get that through soon, you can track it in #3245.

I'm also curious to know why you're installing `insomnia-plugin-responses` manually? It's bundled within the release by default already and doesn't get impacted by the yarn warning. 😄 

In any case, a longer term solution for handling yarn responses is still useful.

PR #2862, opened 2020-11-23
Fix dollar not correctly handled on URI

### LINKED ISSUE #2850: URLs with a `$` are not properly handled
**Describe the bug**
The RFC3986 (https://tools.ietf.org/html/rfc3986) about URI list the `$` sign as an allowed subdelimiter, which allows you to expose services on URLs like : `https://www.myservice.com/$graphql`

With the current version of Insomnia, this URI is not well handed, as the `$` is converted into `%24` before the request is sent

**To Reproduce**
I've reproduced the error, in particular with a GraphQL request.

**Expected behavior**
Right now, with Insomnia, the server answer me :
```
{
  "resourceType": "OperationOutcome",
  "text": {
    "div": "POST /%24graphql not found"
  },
  "issue": [
    {
      "severity": "error",
      "code": "not-found",
      "details": {
        "text": "POST /%24graphql not found"
      }
    }
  ]
}
```
Using Postman, everything works well
 - OS: Fedora 32
 - Installation Method: Appimage
 - App Version: 7.1.1

MrSnix:
Added dollar sign on `URL_PATH_CHARACTER_WHITELIST`.
I don't know if there is a better place but seems working as expected.

On query parameters dollars it's still escaped as it should be.

**Before**
Schermata 2020-11-23 alle 22 52 58

**After**
Schermata 2020-11-23 alle 22 52 00

It closes #2850

### DIFF
--- a/packages/insomnia-url/src/querystring.test.ts
+++ b/packages/insomnia-url/src/querystring.test.ts
@@ -178,13 +178,13 @@ describe('querystring', () => {
     });
 
     it('leaves already encoded pathname', () => {
-      const url = smartEncodeUrl('https://google.com/foo%20bar%20baz/100%25/foo');
-      expect(url).toBe('https://google.com/foo%20bar%20baz/100%25/foo');
+      const url = smartEncodeUrl('https://google.com/foo%20bar%20baz/100%25/foo/%24');
+      expect(url).toBe('https://google.com/foo%20bar%20baz/100%25/foo/%24');
     });
 
     it('encodes querystring', () => {
-      const url = smartEncodeUrl('https://google.com?s=foo bar 100%&hi');
-      expect(url).toBe('https://google.com/?s=foo%20bar%20100%25&hi');
+      const url = smartEncodeUrl('https://google.com?s=foo bar 100%&hi$');
+      expect(url).toBe('https://google.com/?s=foo%20bar%20100%25&hi%24');
     });
 
     it('encodes querystring with mixed spaces', () => {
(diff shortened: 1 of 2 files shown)


dimitropoulos:
@reynolek do you know if this will be a breaking change?  For example, perhaps there are current use-cases that depend on the $ being encoded.  I feel like it's probably fine to do what this PR implements, but I'm not sure.


Screenshot_20210405_160039

reynolek:
I am honestly not sure. I feel like we probably *should* be encoding all special characters and in the event that they are expected to be non-encoded feels like a bug that it still worked.

MrSnix:
Small news:
I resolved the conflicts and updated the Typescript support.
Just in case it gets merged.

dimitropoulos (approved):
Maybe I don't have the right context, but this seems like a legit bug when I read through the original issue (https://github.com/Kong/insomnia/issues/2850).

@develohpanda please also take a look

@reynolek can you say more about what you mean by "special characters" in the context we're talking about.  It seems like these characters _should_ be allowed through from my reading of things.

reynolek:
Yeah looks like `$` should be allowed through. I am good with the change as it stems from an RFC, and contains test cases for query params as well as paths, decreasing the chances of breaking it in the future 👍

reynolek (approved):

MrSnix:
Hello everyone, I want only to say that I did some research some time ago and it turned out that usually:

- Dollar is expected to be encoded if it is a query parameter
- Dollar is not expected to be encoded if it is a path sub delimiter

image
Source: [RFC3986 - Reserved Characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2)

Anyway, I'm not an expert so everyone is welcome to tell me I'm wrong.

develohpanda (approved):

night-fury-web:
When will it be released?

reynolek:
@sandguardian It will come out with our next release. It is currently unscheduled due to a larger body of work that is happening atm, which requires us to do quite a bit of cherry picking and history re-writing to accomplish a release without the new feature (as the new feature isnt ready for release).

I wish I could be more specific, but hopefully not too long, we have a lot of great OSS contributions to release, including this one.

PR #2863, opened 2020-11-24
Remove form-control styles from gRPC url bar

develohpanda:
What it says on the tin!

Designer:
image

Core:
image


sonicyeti (approved):
Sweet!

PR #2864, opened 2020-11-24
Validate protofile during upload

develohpanda:
This PR introduces an `AsyncButton`, which is derived from a regular `Button` but if the onClick callback returns a promise, it will disable the button and optionally show a custom loading element until the promise resolves.

I have added two stories for an `AsyncButton` in Storybook as well. The `AsyncButton` is then consumed in the "Add ProtoFile" button (see gif below), and the re-upload button as well.

For validation, both adding and re-uploading a protofile follows the same code path. If there is an error while parsing the proto file, further processing stops and the error is shown.


DMarby (approved):
Looks good!

PR #2868, opened 2020-11-24
Add OS dark mode support

### LINKED ISSUE #3154: System theme synchronization
I really enjoying using Insomnia, there is just one thing that can be improved - theme synchronization with the system theme (light/dark). I'm used to switching whole my computer (mac) to light (beach) and dark (office) mode just by "one-click", just insomnia needs to be switched manually in settings.

flostellbrink:
Just discovered insomnia and I love it, especially the themes and plugins!

One thing I missed was the theme updating when I switch my OS to dark mode.
So I went ahead and added that. Maybe you'll find it useful too.

Not sure how well this works outside of Windows 10, but I am guessing electron should make it work.

image

Figuring out what is a light and what is a dark theme is just based on the background color.
I know that's not ideal, but it seems to work pretty well.
I think there should be some way for themes to override that if they want.

Please feel free to change the default themes, I just picked them arbitrarily.

Closes #3154


flostellbrink:
Hey @develohpanda, did you get a chance to look at this already?

develohpanda:
> Hey @develohpanda, did you get a chance to look at this already?

Hi @flostellbrink! Unfortunately not yet 😞 I've been a little pressed for time given the awesome number of PRs and pushing for a big release last week. I'll be sure to run through it in the coming or the following week! 🤗

packages/insomnia-app/app/plugins/misc.js
```diff
@@ -230,6 +230,39 @@ function wrapStyles(theme: string, selector: string, styles: string) {
   ].join('\n');
 }
 
+export function isDarkTheme(theme: Theme) {
+  // TODO just a heuristic, should be replaced with a property on themes
+  const background = theme.theme.theme.background?.default;
+  return background && Color(background).isDark();
+}
```
develohpanda: I think we can avoid this heuristic entirely and allow _any_ theme to be the default for light or dark schemes.

For example, in VS Code, when specifying the preferred light theme, the user can still choose a dark theme.

image

packages/insomnia-app/app/plugins/misc.js
```diff
@@ -230,6 +230,39 @@ function wrapStyles(theme: string, selector: string, styles: string) {
...
+  }
+
+  if (window.matchMedia(`(prefers-color-scheme: dark)`).matches) {
+    return 'dark';
+  }
+
+  return 'default';
+}
+
+export async function applyColorScheme(settings: Settings): Promise<IColorTheme | null> {
```
develohpanda: Looks like this function doesn't return anything

```suggestion
export async function applyColorScheme(settings: Settings): Promise<void> {
```

packages/insomnia-app/app/ui/components/settings/theme.js
```diff
@@ -45,7 +55,7 @@ class Theme extends React.PureComponent<Props, State> {
         style={{ maxWidth: `${100 / THEMES_PER_ROW}%` }}>
         <h2 className="txt-lg">{theme.theme.displayName}</h2>
         <Button
-          onClick={handleChangeTheme}
+          onClick={() => handleChangeTheme(theme, colorScheme)}
           value={theme.theme.name}
```
develohpanda: The `value` prop is usually send to the `onClick` handler, but since the handler parameter is no longer used the `value` prop itself can be removed.

develohpanda (commented):
This is an awesome PR and I'm pretty excited to get it in! 🎉

Generally it looks good to me. I'd like to propose a few small changes to the Themes tab interface, in order to allow a user to use any theme for any color scheme, instead of us separating light and dark themes.

I'll get some feedback from our designer @erickrico around a nicer interface in the Themes tab, for selecting the light/dark theme when auto-detection is enabled. That will require some refactoring so I haven't reviewed the files I anticipate to change.

develohpanda:
Hi @flostellbrink! Firstly, I've just tested this on macOS and it seems to detect the color scheme as expected 👏🏽 

I had a chat with our PM and designer, and we think it might be good to follow the VS Code approach for now. This would involve not splitting the theme listing into light and dark, but instead allowing for light/dark selection through two drop-downs, which show all themes and default to those in app config.

Most of the work you have done is still applicable, there will just be a few changes in the themes tab itself.

Here's a quick mock-up just with the HTML, pulling elements from the General settings tab.

image

flostellbrink:
@develohpanda Thanks for the feedback and for finding a nice UI solution!

Gonna get back to you on the weekend :)

flostellbrink:
Heey, sorry for the delay.
This should address the issues.

Also got the new look in:
image

Hope this does the trick. Happy Holidays!

packages/insomnia-app/app/ui/components/settings/theme.js
```diff
@@ -114,13 +119,68 @@ class Theme extends React.PureComponent<Props, State> {
...
   }
 
+  renderThemeSelect(scheme: ColorScheme) {
+    const { activeLightTheme, activeDarkTheme, handleChangeTheme } = this.props;
+    const { themes } = this.state;
+
+    const activeColorTheme = {
+      light: activeLightTheme,
+      dark: activeDarkTheme,
+    }[scheme];
```
develohpanda: Interesting approach! I suggest using `if...else` or a `switch` statement instead of a map, especially considering this code could run frequently (render function), object maps have memory implications and a map doesn't seem necessary here. Is there a particular reason you chose to use a map?

packages/insomnia-app/app/ui/components/settings/theme.js
```diff
@@ -114,13 +119,68 @@ class Theme extends React.PureComponent<Props, State> {
 
     return rows.map((row, i) => (
       <div key={i} className="themes__row">
-        {row.map(this.renderTheme)}
+        {row.map(theme => this.renderTheme(theme))}
```
develohpanda: Can be simplified to the previous implementation 🙂 

```suggestion
        {row.map(this.renderTheme)}
```

packages/insomnia-app/app/ui/components/settings/theme.js
```diff
@@ -114,13 +119,68 @@ class Theme extends React.PureComponent<Props, State> {
...
+              type="checkbox"
+              name="autoDetectColorScheme"
+              checked={autoDetectColorScheme}
+              onChange={e => handleAutoDetectColorSchemeChange(e.target.checked)}
+            />
+          </label>
+        </div>
+        <div className="form-row">
+          {this.renderThemeSelect('light')}
+          {this.renderThemeSelect('dark')}
```
develohpanda, line 186: Nice reuse 💯

packages/insomnia-app/app/ui/components/modals/settings-modal.js
```diff
@@ -64,11 +64,29 @@ class SettingsModal extends PureComponent {
...
-  async _handleChangeTheme(theme, persist = true) {
-    await setTheme(theme);
+  async _handleChangeTheme(themeName, colorScheme, persist = true) {
+    const { settings } = this.props;
+
+    const patch = {
+      default: { theme: themeName },
+      light: { lightTheme: themeName },
+      dark: { darkTheme: themeName },
+    }[colorScheme];
```
develohpanda: Same as a previous comment, using a `switch` should be a little more efficient and easier to skim read. 😄

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1217,6 +1217,10 @@ class App extends PureComponent {
 
     // Give it a bit before letting the backend know it's ready
     setTimeout(() => ipcRenderer.send('window-ready'), 500);
+
+    window
+      .matchMedia('(prefers-color-scheme: dark)')
+      .addListener(async () => themes.applyColorScheme(this.props.settings));
```
develohpanda, line 1224: I'm unfamiliar here, do you need to add a listener for `prefers-color-scheme: light` as well? Seems to work without it, but just checking!

packages/insomnia-app/app/ui/components/settings/theme.js
```diff
@@ -114,13 +119,68 @@ class Theme extends React.PureComponent<Props, State> {
...
+  }
+
   render() {
-    return <div className="themes pad-top">{this.renderThemeRows()}</div>;
+    const { autoDetectColorScheme, handleAutoDetectColorSchemeChange } = this.props;
+
+    return (
+      <>
+        <h2>Theme</h2>
+        <div className="themes">{this.renderThemeRows('default')}</div>
```
develohpanda:
```suggestion
        <div className="themes">{this.renderThemeRows()}</div>
```

Argument is now redundant 😄

develohpanda (changes requested):
Awesome! I've just got some minor comments, namely to convert the object maps to switch/if-else statements.

I was doing some testing locally and noticed that when the "use OS color scheme" box is checked, clicking on the themes above doesn't do anything but still changes the selection. We could disable the click handler entirely, and not show the selected 'default' theme, essentially making it read-only. What would you expect to happen, or do you have any other ideas for that case?




Happy belated holidays to you too! Sorry for the delay, I was on vacation. 😄

packages/insomnia-app/app/ui/containers/app.js
flostellbrink: Listening to just one of them seems to do the trick and is what other people have been doing too, I think^^

packages/insomnia-app/app/ui/components/settings/theme.js
flostellbrink: Well I personally like things to be const whenever possible.
The switch alternative to that is to put it into a function that can return a value.
But I am happy to make that mutable and assign it in a switch if you think that makes it more readable.

I am not gonna benchmark this, but I think that it really doesn't make much of a performance difference at all.

packages/insomnia-app/app/ui/components/modals/settings-modal.js
flostellbrink: Sure

flostellbrink:
Heey @develohpanda! Sorry for the delay.
I went ahead and implemented the suggested changes.

You're right, its confusing to change the default theme when it is not enabled.
Disabling those buttons seems like a good enough solution 👍 

Gonna fix those conflicts real quick.

flostellbrink:
Okay, had to rebase, it was getting a bit messy. Quite a lot has happened since last time.

Looks like things are still working though.

packages/insomnia-app/app/ui/components/settings/theme.js
develohpanda: > The switch alternative to that is to put it into a function that can return a value.

I think this is a suitable alternative! It is a bit more readable than the map, for this use case 🤗

flostellbrink: Oh yeah you're right.
I was thinking about the self invoke lambda style.
But it would have been perfectly reasonable to just have this as a normal named function. 🙈

Worked around it with a ternary, hope that's okay too.

develohpanda: A hundred ways to solve the same problem 😂 ternary is fine too, there will only ever be light-dark mode lol

packages/insomnia-app/app/ui/components/modals/settings-modal.js
```diff
@@ -65,11 +65,36 @@ class SettingsModal extends PureComponent {
...
 
-  async _handleChangeTheme(theme, persist = true) {
-    await setTheme(theme);
+  async _handleChangeTheme(themeName, colorScheme, persist = true) {
+    const { settings } = this.props;
+
+    let patch;
+    switch (colorScheme) {
+      case 'light':
+        patch = { theme: themeName };
```
develohpanda: I think this should be setting `lightTheme`?

```suggestion
        patch = { lightTheme: themeName };
```

packages/insomnia-app/app/ui/components/modals/settings-modal.js
```diff
@@ -65,11 +65,36 @@ class SettingsModal extends PureComponent {
...
+        break;
+      case 'dark':
+        patch = { darkTheme: themeName };
+        break;
+      default:
+        patch = { theme: themeName };
+        break;
```
develohpanda, line 82: It might be good to add the case for `'default'` as well, because `ColorScheme` is typed as `'default' | 'light' | 'dark'`

```suggestion
      case 'default':
      default:
        patch = { theme: themeName };
        break;
```

packages/insomnia-app/app/ui/components/settings/theme.js
```diff
@@ -120,8 +127,60 @@ class Theme extends React.PureComponent<Props, State> {
...
+    );
+  }
+
   render() {
-    return <div className="themes pad-top">{this.renderThemeRows()}</div>;
+    const { autoDetectColorScheme, handleAutoDetectColorSchemeChange } = this.props;
+
+    return (
+      <>
+        <h2>Theme</h2>
```
develohpanda: I think we can get away with not having this header because it is on the tab itself 😄 Let's visually keep this part how it was previously

packages/insomnia-app/app/ui/components/settings/theme.js
```diff
@@ -46,8 +53,8 @@ class Theme extends React.PureComponent<Props, State> {
         style={{ maxWidth: `${100 / THEMES_PER_ROW}%` }}>
         <h2 className="txt-lg">{theme.theme.displayName}</h2>
         <Button
-          onClick={handleChangeTheme}
-          value={theme.theme.name}
+          disabled={disabled}
```
develohpanda, line 56: I'm not sure how I feel about the (default) opacity change when the button become disabled. Since the context of these custom buttons is to do with theming, it feels strange to show a skewed color scheme.




Functionally the disabling is good but perhaps we need to indicate "disabled" by:
- ~removing the opacity change~ keep it, see comment below
- disabling the hover animation
- not showing a border around the selected theme (when OS color scheme is enabled)


You'll just need to write a custom CSS class `themes__theme-disabled`.

@erickrico thoughts on the above, specifically for this use case?

develohpanda (commented):
No worries about the delay, thanks for making those updates! It's looking great, almost there 💯

packages/insomnia-app/app/ui/components/settings/theme.js
erickrico: yea i agree, i think we should remove the selected border, but i think we should keep the opacity change, and do remove the hover animation

flostellbrink: Done and done

develohpanda (approved):
Awesome!! Very excited for this 🎉

dimitropoulos (changes requested):
Screenshot_20210301_122735

Can confirm this worked on Linux (Kubuntu).

My only qualm is that the UI isn't crystal clear to me at first as to what it's selecting (and also what is currently selected).  For example, it'd be nice if we could get a preview next to each of these (like exists above) for each dropdown.  Then, that would give the user some indication of what theme is currently active (which we have above), e.g.

Screenshot_20210301_123213

dimitropoulos (commented):
also, is it an option to instead use https://www.electronjs.org/docs/api/native-theme.  It seems better to rely on an electron javascript API than a css media query.  At least with the javascript API we can be notified by the type system if anything ever changes - while with CSS we just have to hope it never breaks.

dimitropoulos (approved):
On second thought, I see that a lot was done in this PR and what was implemented does work quite well from my testing.  Although my suggestion about the preview would be nice we can always iterate - I don't want to hold this up at this juncture.  I came and just immediately looked at the code and the code alone without looking at any of the discussion, my b. haha

develohpanda:
I think it might be worth investigating use of the Electron API mentioned above - if not trivial, we can stick with the CSS query for now.

flostellbrink:
Thanks for bringing up the API option @dimitropoulos!

I just read the [electron guide on dark mode](https://www.electronjs.org/docs/tutorial/dark-mode). That makes it sound like the nativeTheme API is more useful for other use cases like syncing an applications dark mode preferences back into the OS components.

For just updating our theme they recommend this:
> If your app has its own dark mode, you should toggle it on and off in sync with the system's dark mode setting. You can do this by using the prefer-color-scheme CSS media query.

And while I absolutely agree that having a strongly typed API for this would be nice, I think there is also something to be said for just using standard web apis when possible.

Either way, I don't think I can justify spending time on implementing that change at the moment.
Hope you understand.

dimitropoulos:
I most definitely understand.  Thanks for looking into it!!

khyveasna11111908:
JJ

khyveasna

PR #2869, opened 2020-11-24
Fix re-uploading a protofile when it is not linked to a request

develohpanda:
The logic in `proto-files-modal` is difficult to test, but I have added further control and tests with surrounding elements.

Fixes INS-329

### DIFF
--- a/packages/insomnia-app/app/ui/context/grpc/grpc-actions.js
+++ b/packages/insomnia-app/app/ui/context/grpc/grpc-actions.js
@@ -120,12 +120,12 @@ const invalidate = (requestId: string): InvalidateAction => ({
   requestId,
 });
 
-const invalidateMany = async (protoFileId: string): Promise<InvalidateManyAction> => {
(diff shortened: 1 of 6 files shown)


packages/insomnia-app/app/ui/context/grpc/grpc-actions.js
```diff
@@ -120,12 +120,12 @@ const invalidate = (requestId: string): InvalidateAction => ({
   requestId,
 });
 
-const invalidateMany = async (protoFileId: string): Promise<InvalidateManyAction> => {
+const invalidateMany = async (protoFileId: string): Promise<InvalidateManyAction | undefined> => {
```
develohpanda, line 123: This is partly the root cause, TS would have helped detect this sooner and would have required us to handle the undefined case correctly.

DMarby (approved):
:shipit:

sonicyeti (approved):

PR #2883, opened 2020-12-01
gRPC URL bar improvements - streaming type & package groups

develohpanda:
We now show an acronym with a tool-tip for the RPC type. The following is shown when a proto file does not specify a package.

image

In the case where a proto file contains a package, the package name is split and shown in the divider.

image

In the future, where the user could load multiple proto files with multiple packages, the services and methods will be separated with "package" dividers exactly as shown above. Is this a reasonable thing to do? It definitely helps to simplify the UI, removing a _lot_ of repetition when you have nested packages (`/a.nested.package.Service/Method`).

A full path including package would be `/grpcbin.GRPCBin/Index`, while a path without a package is `/GRPCBin/Index`.

The full path is still visible as a tooltip with each item and the dropdown button when collapsed.

image

image

Scaling adjustments:


packages/insomnia-app/app/common/grpc-paths.js
```diff
@@ -0,0 +1,52 @@
+// @flow
+
+import type { GrpcMethodDefinition, GrpcMethodType } from '../network/grpc/method';
+import { groupBy } from 'lodash';
+import { getMethodType } from '../network/grpc/method';
+
+const PROTO_PATH_REGEX = /^\/(?:(?<package>[\w.]+)\.)?(?<service>\w+)\/(?<method>\w+)$/;
```
develohpanda, line 7: image

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.js
```diff
@@ -13,13 +13,32 @@ import { grpcActions, useGrpc, useGrpcIpc } from '../../../context/grpc';
...
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: stretch;
+`;
+
+const UrlEditor = styled.div`
+  flex: 3 0 4em;
+  max-width: 11em;
+`;
```
develohpanda, line 36: This can be `styled(OneLineEditor)` 🤯

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.js
```diff
@@ -13,13 +13,32 @@ import { grpcActions, useGrpc, useGrpcIpc } from '../../../context/grpc';
...
+
+const StyledDropdown = styled.div`
+  flex: 1 0 auto;
+`;
```
develohpanda, line 40: This can by `styled(GrpcMethodDropdown)` by adding a `className` prop to `GrpcMethodDropdown`

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.js
```diff
@@ -13,13 +13,32 @@ import { grpcActions, useGrpc, useGrpcIpc } from '../../../context/grpc';
...
+
+const StyledUrlEditor = styled.div`
+  flex: 3 0 4em;
+  max-width: 11em;
+`;
+
+const StyledDropdown = styled.div`
+  flex: 1 0 auto;
+`;
```
develohpanda, line 40: Url editor will scale 3x faster than the dropdown, but will maintain a min width of `4em` and max width of `11em`. The dropdown will fill the remaining space. (Gif in description above, it's not really noticeable, but is noticeable when both scale at the same rate)

packages/insomnia-app/app/common/grpc-paths.js
develohpanda: This regex may need an update but I was unable to any detailed documentation on what characters are allowed in a package name, method name and service name. Currently we use the `\w` regex matcher (which `matches any letter, digit or underscore. Equivalent to [a-zA-Z0-9_]`).

This logic is tested in `grpc-paths.test.js` so for any failing names a unit test can be added first.

packages/insomnia-app/app/ui/components/tags/grpc-method-tag.js
```diff
@@ -0,0 +1,26 @@
...
+
+const StyledTag = styled.div`
+  width: 1.5em;
+  text-align: right;
+`;
+
+const GrpcMethodTag = ({ methodType }: Props) => (
+  <Tooltip message={GrpcMethodTypeName[methodType]} position="left" delay={500}>
+    <StyledTag>
+      <em>{GrpcMethodTypeAcronym[methodType]}</em>
```
sonicyeti, line 21: Any chance we could wrap these with a generic gRPC badge where the BG is gRPC blue while the text is white, or vice versa?

sonicyeti (approved):
Not a blocker but would love some sort of styling treatment on the acronyms. Rough but something like this might break it up and make it easier to visually digest.

Screen Shot 2020-12-07 at 9 33 33 AM

packages/insomnia-app/app/ui/components/tags/grpc-method-tag.js
develohpanda: I've updated it to be as your screenshot below (with the same styles applied) 😄

develohpanda: Reverted as per discussion in standup + feedback from Niji

packages/insomnia-app/app/common/grpc-paths.js
DMarby: I think it looks reasonable choice, we can always make it less strict in the future if need be.

DMarby (approved):
:shipit:

PR #2885, opened 2020-12-01
fix: remove request id validation 

develohpanda:
In the 2020.5 release we use heuristics to ensure a model (request and grpc-request) is a particular type (because it should genuinely fail if there is a mismatch) using its id. For example, a requestId is always prefixed by `req_*` and folder by `fld_*`. It has been like this since the start of time.

The problem is that not all of the importers follow that heuristic... evident through the swagger2 importer when importing and running `https://petstore.swagger.io/v2/swagger.json` in 2020.5.

Note that the above specification automatically enables OAuth on a request so it will still fail, but there should be no errors in the console. Additionally, if you remove the authentication on a request (for example on `/pet/findByStatus` the request should run correctly, whereas it exhibits adverse behavior in `2020.5.0-beta.2`.

The OpenAPI3 importer applies the id correctly, though.

Related to INS-341

### DIFF
--- a/packages/insomnia-smoke-test/prism/swagger2.yaml
+++ b/packages/insomnia-smoke-test/prism/swagger2.yaml
@@ -1,0 +1,77 @@
+openapi: "3.0.0"
+info:
+  version: 1.0.0
+  title: E2E testing specification
+  description: This is a specification used for E2E testing of Insomnia, and is served via prism
+servers:
+  - url: http://127.0.0.1:4010
+paths:
+  /pets/{id}:
+    get:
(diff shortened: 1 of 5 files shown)


sonicyeti (approved):

packages/insomnia-smoke-test/prism/swagger2.yaml
```diff
@@ -0,0 +1,77 @@
+openapi: "3.0.0"
```
develohpanda, line 1: Damn, this really did sneak in here during the last commit. No matter, it doesn't impact anything.

PR #2888, opened 2020-12-02
Strips newlines when parsing GraphQL curl command

### LINKED ISSUE #1825: Insomnia is not converting some CURL requests to GraphQL properly
- Insomnia Version: 7.0.5
- Operating System: Ubuntu 18.04.3 LTS

## Details
Insomnia is not converting some CURL requests to GraphQL properly. The endlines are kept in the GraphQL output as strings.

Steps to reproduce:
Past this CURL string into insomnia
```
curl 'http://localhost:4000/api/graphql' -H 'Connection: keep-alive' -H 'accept: */*' -H 'Origin: http://localhost:4000' -H 'authorization: Bearer d1aea2f4d2714202bd66f80feb8635d0' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.87 Safari/537.36' -H 'DNT: 1' -H 'content-type: application/json' -H 'Sec-Fetch-Site: same-origin' -H 'Sec-Fetch-Mode: cors' -H 'Referer: http://localhost:4000/test' -H 'Accept-Encoding: gzip, deflate, br' -H 'Accept-Language: sk-SK,sk;q=0.9,cs;q=0.8,en-US;q=0.7,en;q=0.6' --data-binary $'{"operationName":"myQuery","variables":{"f":{}},"query":"query myQuery($testVariable: Boolean\u0021) {\\n  myQuery(filter: $f) {\\n    data {\\n      id\\n      __typename\\n    }\\n    __typename\\n  }\\n}\\n"}' --compressed
```

Following body will be generated:
```
query myQuery($testVariable: Boolean!) {\n  myQuery(filter: $f) {\n    data {\n      id\n      __typename\n    }\n    __typename\n  }\n}\n
```

Note that the **\n** is present in the output which should not be present.

The curl was generated by **Google Chrome Version 78.0.3904.87 (Official Build) (64-bit)**

The issue is that chrome is escaping endlines with extra slash  `\\n`
I tried to generate same CURL request from Firefox which worked fine in insomnia. Difference was that the curl request in FF was containing just `\n`

image

gerrytan:
This PR addresses a bug where GraphQL curl command copied from Chrome is not parsed correctly. 

**To reproduce the bug:**

1. Copy a graphql request as curl from Chrome network inspector
2. Paste it into an Insomnia request address bar
3. Switch the mime type to GraphQL

**Expected**: user is presented with a valid GraphQL query
**Actual**: user is presented with an invalid GraphQL (bunch of \n characters)

**Video of the problem**: [here](https://drive.google.com/file/d/1_xdD3kcAPEO03zBARPn8WFlrT2uYrNIq/view?usp=sharing)

**Video of the new behaviour after this fix**: [here](https://drive.google.com/file/d/1pAqcQIq7eieEmncDN9EHH1sVvD6N5FeK/view?usp=sharing)

Closes #1825


packages/insomnia-app/app/models/request.js
```diff
@@ -212,6 +212,20 @@ export function newBodyRaw(rawBody: string, contentType?: string): RequestBody {
   return { mimeType, text: rawBody };
 }
 
+export function newBodyGraphQL(rawBody: string): RequestBody {
+  try {
+    // Only strip the newlines if rawBody is a parsable JSON
+    JSON.parse(rawBody);
```
dimitropoulos, line 218: There's no other way to do this, and `JSON.parse` is the fastest way available, but it still is one of those things I want to mention - if anyone can think of a way to do this without `JSON.parse`ing arbitrary inputs please say so.  Sometimes in special cases there are ways around it, but from what I can tell this isn't one of those situations.

dimitropoulos (approved):
functionally tested and reviewed code: thanks for the contribution!

reynolek:
Updated branch with `develop`

develohpanda (approved):
Looks and works great! Thanks for the PR, and apologies for the delays 🙏🏽

packages/insomnia-app/app/models/request.js
develohpanda: I don't think this is one of the special cases either so `JSON.parse` seems like a reasonable approach to it.

gerrytan:
Thanks a lot for your help @dimitropoulos and @develohpanda

Issue #2856, opened 2020-11-22
The query value add selection mode，OK?


anyforever:
![Image of example](https://wx3.sinaimg.cn/mw1024/63900ec6ly1gkxw23nbeoj20a905cgli.jpg)

Issue #1802, opened 2019-11-22
[Feature Request] Allow a way to show all the accepted options for a query parameter


sebastien-bartoli:
I use Insomnia to have a collection of all endpoints of my API, as well as documentation. What would help a lot in our swagger export of this, would be to be able to show all the accepted options for one parameter.

For instance, in the example below the `kind` parameter only accept `internal` or `billable` ; the `status` parameter only accept `lead`, `mission_started`, etc. 

<img width="694" alt="Screenshot 2019-11-22 at 17 53 17" src="https://user-images.githubusercontent.com/590363/69444619-007e2300-0d51-11ea-83b6-bef6cf8f380c.png">

It would be really interesting to be able to specify what is accepted or not, maybe as an enum or as a list, to avoid confusion. 

I think stoplight.io did that quite well, and could use inspiration from there.

Thanks

PR #2909, opened 2020-12-04
Introducing enum support

### LINKED ISSUE #1806: [Feature Request] Import Enumerations from Swagger file
As we are able to provide enumerations for parameters in body/query in Swagger, it would be neat if you could implement an option to import enumerations explicitly from Swagger file which would be, again, optional along with custom input provided by user. By making this feature possible, larger groups of developers would spend less time opening swagger file and searching the whole file for enums for certain parameter. For a single developer working on a project, such feature won't make significant difference, but as mentioned, larger teams may often come in confusion where many http requests are used along with their unsemantic params as well.

MrSnix:
**Introduction**
One of the frequently requested features is to support enumeration on a given field.
It is a nice feature even because there is built-in support on OpenAPI and it'd be cool to have it:
```
paths:
  /items:
    get:
      parameters:
        - in: query
          name: sort
          description: Sort order
          schema:
            type: string
            enum: [asc, desc]
```
**Overview**
I didn't start from scratch, I have looked around the project looking for stuff I could modify a bit to work with without introducing new code. I mainly reworked some part of `Dropdown` to build this new little component but before going further I'd like to know if this is the right direction and **I'd be glad to hear some feedback about this.**
Right now, the selection mode allows you to arbitrarily insert new values but you can even pass them inside the component's props and set the component as read-only so you can pick what is already there (as you would do if it was extracted from OpenAPI specs) This is a little gif that shows how it currently works.
test0

This is a proposal, **it's not yet finished** and it should be considered as a draft before a finished pull request.

It closes #1806 #1802 #2856


packages/insomnia-app/app/plugins/context/app.js
```diff
@@ -29,7 +29,7 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): { a
       dialog(
         title,
         body: HTMLElement,
-        options?: {
```
MrSnix, line 32: My IDE kept complaining about it, I found these little mistakes and fixed them.

packages/insomnia-app/app/ui/components/base/dropdown/dropdown.js
```diff
@@ -395,7 +395,13 @@ class Dropdown extends PureComponent {
                 />
               </div>
               {noResults && <div className="text-center pad warning">No match :(</div>}
-              <ul className={classnames({ hide: noResults })}>{dropdownItems}</ul>
+              <ul className={classnames({ hide: noResults })}>
```
MrSnix, line 398: Before if dropdown had no items it'd show anyway, now it displays a proper message

packages/insomnia-app/app/ui/components/dropdowns/hint-dropdown.js
```diff
@@ -0,0 +1,99 @@
...
+  hint?: string,
+  placeholder?: string,
+  items?: Array<string>,
+  defaultValue?: string,
+  submitName?: string,
+  onChange?: (value: string, items: Array<string>) => void,
+};
+
+@autobind
+class HintDropdown extends React.PureComponent<Props> {
```
MrSnix, line 21: I didn't have a better idea... naming stuff is difficult.

packages/insomnia-app/app/ui/components/modals/prompt-modal.js
```diff
@@ -248,21 +275,41 @@ class PromptModal extends React.PureComponent<Props, State> {
...
-          </form>
-        </ModalBody>
-        <ModalFooter>
-          <div className="margin-left faint italic txt-sm tall">{hint ? `* ${hint}` : ''}</div>
-          <button className="btn" onClick={this._handleSubmit}>
-            {submitName || 'Submit'}
-          </button>
-        </ModalFooter>
-      </Modal>
+      <form onSubmit={this._handleSubmit}>
```
MrSnix, line 278: I moved the form on the outside part because it'd generate the following warning:
`Form submission canceled because the form is not connected`

nijikokun:
This is awesome 👏

MrSnix:
Hi all, I got some news:
- I was able to introduce enumeration support on Swagger and OpenAPI 3.0
- I rewrote some tests which were failing due to changes introduced previously

This gif shows the feature on Swagger 2.0 (but as I said, it works on OpenAPI 3.0 too):
test1


**Question**:
I'd like to support enumeration on WSDL as provided by the format:
```
<xs:simpleType name="customerType">
    <xs:restriction base="xs:string">
        <xs:enumeration value="PRIVATE"/>
        <xs:enumeration value="BUSINESS"/>
    </xs:restriction>
</xs:simpleType>
```
but seems like the current WSDL importer relies on Postman.js to make the conversion happen which does not support the enumeration type... if I'm not wrong (?). So right now, I cannot make this happen, is it correct?

Thanks

packages/insomnia-app/app/ui/components/key-value-editor/row.js
```diff
@@ -256,6 +261,14 @@ class KeyValueEditorRow extends PureComponent {
           onChange={this._handleFileNameChange}
         />
       );
+    } else if (pair.type === 'selection') {
```
MrSnix, line 264: When the requested pair (key-value) has `type === 'selection'` displays the new component

packages/insomnia-app/app/ui/components/modals/prompt-modal.js
```diff
@@ -185,10 +214,11 @@ class PromptModal extends React.PureComponent<Props, State> {
 
     return (
       <div key={hint} className={classes}>
-        <Button className="tall" onClick={this._handleSelectHint} value={hint}>
+        <Button type="button" className="tall" onClick={this._handleSelectHint} value={hint}>
```
MrSnix, line 217: I had to explicit the button type otherwise it would "submit" the form

packages/insomnia-app/app/ui/components/modals/prompt-modal.js
```diff
@@ -224,14 +255,10 @@ class PromptModal extends React.PureComponent<Props, State> {
         min={inputType === 'decimal' ? '0.5' : null}
         style={{ textTransform: upperCase ? 'uppercase' : 'none' }}
         placeholder={placeholder || ''}
+        value={currentValue}
       />
     );
 
-    let sanitizedHints = [];
```
MrSnix, line 230: This function used to sanitize and mapping the elements as` <button>`, I split the function logic. Now the `sanitize` functions work on the raw data and in the `render()` they get transformed as buttons

packages/insomnia-importers/src/__tests__/fixtures/openapi3/dereferenced-output.json
```diff
@@ -73,7 +73,13 @@
       "parameters": [
```
MrSnix, line 73: I changed the test output otherwise it would fail on the new data internal representation

develohpanda:
Thanks for putting in a lot of work into this PR, I haven't had a chance to go through it until now. 😞 It's an awesome start! 

I have some high level thoughts about this feature without digging deeply into the implementation details.

Ignoring the OpenAPI importer changes and assuming the role of a user simply using _Insomnia Core_, my initial thoughts are as follows:
- enums are currently stored in-memory, and are lost on application restart or export/import
- a user needs to redefine an enum/selection for a new parameter
- enums can only be used in the HTTP query parameter and header inputs (as these are the only key-value inputs)

There are two main questions to ask regarding enum support:
- how should enums be **stored**?
- how should enums be **consumed**?

Store: as a user, I would like to specify and persist multiple enums _at least_ at a _Workspace_ level, like variables or certificates. I also want to sync and export the enum definitions for my team to consume. (It might be useful to scope an Enum to a folder or request as well, but I digress.)

Consume: as a user, I would like to use my enums in _any_ input field, using a custom template tag (`{% enum 'Status', 'passed' %}` for example). An enum can also be made available through a dropdown in certain inputs, such as header and query parameter key-value-pairs.

Once this core feature for Insomnia is designed, then the OpenAPI importer can also _consume_ it by creating enum models.

I have some additional thoughts around the changes made to the generic `prompt-modal` specifically for enum support, and that it might be good to create a standalone enum editor. This is more an implementation detail that is likely to change after some of the points above are addressed, so I won't dig deeper with it for now.

I see enum support in Insomnia being a _very_ similar paradigm to variables, proto files, and certificates - a "resource" at the workspace level consumable by almost anything.

What do you think, @MrSnix? 😄 @nijikokun feel free to chime in.

develohpanda:
Thanks for putting in a lot of work into this PR, I haven't had a chance to go through it until now. 😞 **It's an awesome start!** 🎉 👏🏽  

I have some high level thoughts about this feature without digging deeply into the implementation details.

Ignoring the OpenAPI importer changes and assuming the role of a user simply using _Insomnia Core_, my initial thoughts are as follows:
- a user needs to re-define an enum/selection, to use the same enum in another location
- enums can only be used in the HTTP query parameter input

There are two main questions to ask regarding enum support:
- how should enums be **stored**?
- how should enums be **consumed**?

**Store:** as a user, I would like to specify and persist multiple enums _at least_ at a _Workspace_ level, like variables or certificates. (It might be useful to scope an Enum to a folder or request as well, but I digress.)

**Consume:** as a user, I would like to use my enums in _any_ input field, using a custom template tag (`{% enum 'Status', 'passed' %}` for example). An enum can also be made available through a dropdown in certain inputs, such as header and query parameter key-value-pairs. In each case, I would like to be able to use the _same_ enum definition against multiple inputs.

Once this core feature for Insomnia is designed, then the OpenAPI importer can also _consume_ it by creating enum models and transforming the data as appropriate for Insomnia.

I have some additional thoughts around the changes made to the generic `prompt-modal` specifically for enum support, and that it might be good to create a standalone enum editor. This is more an implementation detail that is likely to change after some of the points above are addressed, so I won't dig deeper with it for now.

I see enum support in Insomnia being a _very_ similar paradigm to variables, proto files, and certificates - a "resource" at the workspace level consumable by almost anything. This is just one approach, there might be other valid approaches too.

What do you think, @MrSnix? 😄 @nijikokun feel free to chime in.

MrSnix:
Hi @develohpanda, I **agree** with all your initial thoughts! :partying_face: :partying_face: :partying_face:

>  I would like to specify and persist multiple enums at least at a Workspace level, like variables or certificates.
> I would like to use my enums in any input field, using a custom template tag ({% enum 'Status', 'passed' %} for example)

I agree 💯 and I'll work on that 👍🏻 

> I have some additional thoughts around the changes made to the generic prompt-modal specifically for enum support, and that it might be good to create a standalone enum editor.

Yes, I approve.

I will try to do everything I can to develop these features.

develohpanda:
@MrSnix great to hear! If it helps, feel free to share a plan of attack here first. We can review and refine the approach before you writing any code. 😊 I am also happy to give some (delayed) preliminary reviews.

MrSnix:
At the moment I have several thoughts in my head @develohpanda :
- I am wondering if we should provide a specific enum editor... and where should we place it?
I was thinking we may put it near "Cookies", replacing the current text with a button group made by two icons one for the cookie and another one for enums, something like this:
ex0
(Please, don't judge the graphic I did it just to showcase the idea)
Or another idea would be to move the "Cookies" button and "Enums" button inside the "Add button", like this:
ex1
(I really like the idea, even because I don't like how Cookies is currently placed but it's just my personal taste 😆 )
- About the custom template tag I don't know where should I look for implementation 🧐 
(nvm I'll see how it has been done somewhere in the codebase 🤓 )
- Another thing is how to develop the Enum editor? Should it be as big as the Cookie Modal? Do we need a filter? Do we need to display it as a list? Which options should be available on each enum? (aside from "Remove" and "Edit").
- Do you see the Selection component (as seen in the gif) still viable considering the new scope? I may modify it so when selecting it may show the available enumerations and you can pick the one you like on a given field because at this point if I have multiple enums with the current state of the component I can only create one limited to the single request if I'm not using the template tag.

Let me know, I'm always glad to hear feedback.
Thanks :D

MrSnix:
Hi, any feedback or news on that?
I need just some tips about which direction should I follow.

Thanks :D

develohpanda:
Hi @MrSnix, I'll be back from vacation next week, and can take a more in depth look. Apologies for the delays!

Edit: after some conversation with the team we decided to halt this feature for now, see following comment. 🙂

nijikokun:
Hey @MrSnix, thanks for the work that you've done so far. After reading through I think it's best if we put the broad scale enum support on hold as scope is growing quite quickly and it needs more research before we start implementation and I wouldn't want you to add more code only to be told it's not the direction we would want to go.

Right now, things that can be agree'd to:

- As a user I would like for enums to be represented inside requests when converting from OpenAPI
  - This is a requested feature as-is and leveraging a template tag such as below would suffice.
- As a user, I would like to use my enums in any input field, using a custom template tag.
  - This feature while it might introduce some small changes to the template tag modal, should be minimal and be able to support environment variables which would be able to be treated as the store without requiring other interface changes.

Things that still need research, design, and more requests:

- As a user, I would like to specify and persist multiple enums at least at a Workspace level, like variables or certificates.

PR #2921, opened 2020-12-08
Add additional metrics to the notification call

DMarby:


packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1137,6 +1150,18 @@ class App extends PureComponent {
         if (vcs && doc.type === models.workspace.type && type === db.CHANGE_REMOVE) {
           await vcs.removeProjectsForRoot(doc._id);
         }
+
+        // If a request has been created or deleted, increment the stats
+        if (isRequest(doc) || isGrpcRequest(doc)) {
+          switch (type) {
+            case db.CHANGE_INSERT:
+              this._requestStats.createdRequests++;
```
DMarby: We can't do database operations within here safely (such as updating the stats db object), without causing weird side-effects, so the persistence of this needed to be moved out to it's own method that runs on an interval instead.

Not ideal, but should be cheap enough since we immediately exit if there aren't any stats to persist.
This code will hopefully not be around for too long, as once we get Segment implemented and can send events directly, we can get rid of all these counters.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1217,12 +1242,36 @@ class App extends PureComponent {
...
+      this._requestStats = {
+        ...this._requestStats,
+        createdRequests: 0,
+        deletedRequests: 0,
+      };
+
+      // Persist the stats to the database
+      const stats = await models.stats.get();
+      await models.stats.update({
+        ...(createdRequests && { createdRequests: stats.createdRequests + createdRequests }),
```
DMarby: Not strictly needed, but as an extra guard against potential race conditions we only update the fields that needs updating here.

packages/insomnia-app/app/models/stats.js
```diff
@@ -14,6 +14,9 @@ type BaseStats = {
   currentVersion: string | null,
   lastVersion: string | null,
   launches: number,
+  createdRequests: number,
+  deletedRequests: number,
+  executedRequests: number,
```
develohpanda, line 22: Would you think it might be useful to break this down by request type? We've got REST, GQL, gRPC already, and potentially websockets in the future. Can reuse the model:

```js
type RequestStats = {
	created: number,
	deleted: number,
	executed: number,
}

Stats = {
	requests: {
		rest: RequestStats,
		grpc: RequestStats,
		graphql: RequestStats,
	}
}
```

packages/insomnia-app/app/common/analytics.js
```diff
@@ -90,6 +90,28 @@ export function trackPageView(path: string) {
...
+  });
+}
+
+export async function getDeviceId(): Promise<string> {
+  const settings = await models.settings.getOrCreate();
+
+  let { deviceId } = settings;
+  if (!deviceId) {
+    // Migrate old GA ID into settings model if needed
+    const oldId = (window && window.localStorage.gaClientId) || null;
```
develohpanda: Safe access of `window` 👍 Should it be using `window.localStorage.getItem()` though?

packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -170,6 +180,9 @@ export const startBidiStreaming = async (
     return;
   }
 
+  // Update request stats
+  trackExecutedRequest();
```
develohpanda: Is there a desire to track the number of messages for gRPC as well? It would/could be a better representation for gRPC in comparison to REST/GQL.

Can be instrumented in a future PR but just thought I'd mention it 🙂

packages/insomnia-app/app/ui/components/toast.js
```diff
@@ -127,6 +128,10 @@ class Toast extends React.PureComponent<Props, State> {
         autoUpdatesDisabled: !settings.updateAutomatically,
         disableUpdateNotification: settings.disableUpdateNotification,
         updateChannel: settings.updateChannel,
+        deviceId: await getDeviceId(),
+        createdRequests: stats.createdRequests,
+        deletedRequests: stats.deletedRequests,
+        executedRequests: stats.executedRequests,
```
develohpanda, line 134: Out of curiosity, is the `/notification` endpoint used to send stats just because that's the fastest way to go about it in the current state? Doesn't seem that intuitive 🤔 not asking for a change here but for my understanding

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -650,6 +657,9 @@ class App extends PureComponent {
       return;
     }
 
+    // Update request stats
+    trackExecutedRequest();
+
     // NOTE: Since request is by far the most popular event, we will throttle
     // it so that we only track it if the request has changed since the last one
```
develohpanda: Should this note and throttling logic now be removed, or is that unrelated? I couldn't seem to find any legitimate use of `this._sendRequestTrackingKey`, and it seems like this PR ignores the throttling logic entirely.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -738,6 +748,9 @@ class App extends PureComponent {
       return;
     }
 
+    // Update request stats
+    trackExecutedRequest();
+
     // NOTE: Since request is by far the most popular event, we will throttle
     // it so that we only track it if the request has changed since the last noe
```
develohpanda: Same as above regarding throttling note and logic

packages/insomnia-app/app/ui/containers/app.js
develohpanda: Adding the stats logic here will likely mean if someone imports a workspace with 5 requests, it will track 5 created requests. Same with deleting a workspace for example, where all requests will be removed.

This could skew stats significantly, because moving from design to debug in Designer always re-imports (ie. deletes and re-creates IIRC) requests.

Also, it might be worth double checking if there are any unexpected side-effects when db changes are being buffered.

For the most part, the `models.(grpcRequest|request).(create|duplicate|remove)` functions are used correctly, and you can directly update the stats model from there as well, and I think it won't skew stats in Designer this way.

Just wanted to get my thoughts down here to digest, we can talk about this on a call. 🙂 Needs some testing/validation.

packages/insomnia-app/app/ui/containers/app.js
develohpanda: I think it might be worth creating a special function in the stats model to encapsulate this logic: `models.stats.incrementRequestStats(createdRequests, deletedRequests)`.

That way the persistence logic can be solidified with unit tests to an extent, to make sure we don't suffer from data loss. This seems like an important area to be tested.

develohpanda (commented):
Generally looks good, but I'll hold off on approving it as I have some notes about the potential for skewed stats, and one note about a small refactor for the purpose of writing a unit test for the persistence of stats to the db.

> Once we get Segment implemented and can send events directly, we can get rid of all these counters.

By counters, I'm assuming you're also talking about the counters in the `Stats` model? If so, you can most probably disregard some of my comments about collecting stats for rest/gql/grpc/websockets separately, although I will leave them up anyway. 😄

packages/insomnia-app/app/ui/components/toast.js
DMarby: Most of the things that's sent is used to decide what notification to return, eg to trigger a notification after the user has used the app for a while, but it's also used for some metrics as it's the fastest way, yeah.

packages/insomnia-app/app/ui/containers/app.js
DMarby: It can be, yeah, doesn't seem to have been used for several years.
This was used to not spam GA with events, which we don't care about anymore (as there isn't an event for this anymore).
Removed it.

packages/insomnia-app/app/common/analytics.js
DMarby: Both ways are valid ways of accessing it, this is just a leftover from the old code.
Changed for consistency with the rest of the codebase.

packages/insomnia-app/app/ui/containers/app.js
DMarby: Sounds reasonable

packages/insomnia-app/app/common/send-request.js
```diff
@@ -24,6 +24,7 @@ export async function getSendRequestCallbackMemDb(environmentId, memDB) {
 
 export function getSendRequestCallback(environmentId) {
   return async function sendRequest(requestId) {
+    stats.incrementExecutedRequests();
```
DMarby, line 27: This gets called whenever a unit test executes a request within the app, but not when inso runs tests

packages/insomnia-app/app/models/stats.js
```diff
@@ -49,3 +56,40 @@ export async function get(): Promise<Stats> {
...
+
+export async function incrementDeletedRequests() {
+  await incrementRequestStats({ deletedRequests: 1 });
+}
+
+export async function incrementExecutedRequests() {
+  await incrementRequestStats({ executedRequests: 1 });
+}
+
+export async function incrementCreatedRequestsForDescendents(doc: BaseModel) {
```
develohpanda: I would favor restricting the types here because we know for sure this should only be a request group or a workspace (same with the deletion function below)

```suggestion
export async function incrementCreatedRequestsForDescendents(doc: Workspace | RequestGroup) {
```

packages/insomnia-app/app/ui/components/dropdowns/document-card-dropdown.js
```diff
@@ -51,6 +51,8 @@ class DocumentCardDropdown extends React.PureComponent<Props, State> {
         const newWorkspace = await db.duplicate(workspace, { name: newName });
         await models.apiSpec.updateOrCreateForParentId(newWorkspace._id, { fileName: newName });
 
+        models.stats.incrementCreatedRequestsForDescendents(newWorkspace);
```
develohpanda, line 54: Great catch to add it here as well, I had forgotten about this initially

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -351,6 +354,8 @@ class App extends PureComponent {
       selectText: true,
       onComplete: async name => {
         await models.requestGroup.duplicate(requestGroup, { name });
+
+        models.stats.incrementCreatedRequestsForDescendents(requestGroup);
```
develohpanda: Similar to the workspace duplicate logic above in `document-card-dropdown`, the `duplicate()` should return the newly created request group which can be sent to the stats, instead of the original.

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -313,6 +313,10 @@ function handleImportResult(result: ImportResult, errorMessage: string): Array<W
     return [];
   }
 
+  const createdRequests =
+    summary[models.request.type].length + summary[models.grpcRequest.type].length;
+  models.stats.incrementRequestStats({ createdRequests: createdRequests });
```
develohpanda, line 318: I think you mentioned this but just confirming, this code path is not reached by moving from design -> debug?>

develohpanda (approved):
Looks good! I need to do some QA on it but code is good

packages/insomnia-app/app/models/stats.js
DMarby: Sounds reasonable 👍

packages/insomnia-app/app/ui/containers/app.js
DMarby: Good catch! Done

packages/insomnia-app/app/ui/redux/modules/global.js
DMarby: Yep! `handleImportResult` is only called from importFile/importClipBoard/importUri, while moving from design -> debug uses importRaw, which does not call it.

PR #2926, opened 2020-12-10
Support template tags with gRPC

develohpanda:
This PR adds template tag support for the gRPC host input and body editor. Any tags can be used, such as a prompt or data generated using the [Faker](https://insomnia.rest/plugins/insomnia-plugin-faker/) plugin. See comments for additional notes, and feel free to ask if you are confused about data preparation or flow.

The URL auto-complete will also load existing URLs used by other requests in this workspace. This should remain scoped _to the workspace_. This reflects the behavior of HTTP requests.
image


packages/insomnia-components/components/tooltip.js
```diff
@@ -168,10 +168,10 @@ class Tooltip extends React.PureComponent<Props, State> {
   }
 
   _getContainer(): HTMLElement {
-    let container = document.querySelector('#tooltips-container');
+    let container = document.querySelector('#tooltips-container-insomnia-components');
     if (!container) {
       container = document.createElement('div');
-      container.id = 'tooltips-container';
+      container.id = 'tooltips-container-insomnia-components';
```
develohpanda: Undo this

packages/insomnia-app/app/ui/context/grpc/grpc-reducer.js
```diff
@@ -83,6 +83,7 @@ const singleRequestReducer = (state: GrpcState, action: GrpcAction): GrpcState =
     case GrpcActionTypeEnum.start: {
       return _patch(state, requestId, {
         ...oldState,
+        ...CLEAR_GRPC_REQUEST_STATE,
```
develohpanda: Undo this, the clear event should be sent separately

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/use-action-handlers.js
```diff
@@ -0,0 +1,59 @@
...
+import { ipcRenderer } from 'electron';
+import { grpcActions } from '../../../context/grpc';
+import { prepareGrpcMessage, prepareGrpcRequest } from '../../../../network/grpc/prepare';
+
+const _sendStart = async (reqId: string, envId: string, type: GrpcMethodType) => {
+  if (!reqId) {
+    return;
+  }
+  const preparedRequest = await prepareGrpcRequest(reqId, envId, type);
+  ipcRenderer.send(GrpcRequestEventEnum.start, preparedRequest);
```
develohpanda, line 19: dispatch CLEAR to grpc reducer

packages/insomnia-app/app/network/grpc/__tests__/index.test.js
```diff
@@ -0,0 +1,534 @@
+// @flow
```
develohpanda, line 1: ~95% test coverage of `/network/grpc/index.js`

image

packages/insomnia-app/app/network/grpc/__tests__/prepare.test.js
```diff
@@ -0,0 +1,82 @@
+import { prepareGrpcMessage, prepareGrpcRequest } from '../prepare';
```
develohpanda, line 1: image

packages/insomnia-app/app/ui/components/buttons/__tests__/grpc-send-button.test.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
```
develohpanda, line 1: image

packages/insomnia-app/app/network/grpc/index.js
```diff
@@ -25,188 +30,169 @@ const _createClient = (req: GrpcRequest, respond: ResponseCallbacks): Object | u
...
-export const sendUnary = async (requestId: string, respond: ResponseCallbacks): Promise<void> => {
-  // Load method
-  const req = await models.grpcRequest.getById(requestId);
-  const selectedMethod = await protoLoader.getSelectedMethod(req);
-
-  if (!selectedMethod) {
-    respond.sendError(requestId, new Error(`Method definition could not be found`));
-    return;
-  }
+const _makeUnaryRequest = (
```
develohpanda, line 33: The changes in this file will be a little hard to follow in a diff format. The primary changes are the standalone and exported `send{RPCtype}` functions have been consolidated with a single `start` export.

The `start` function takes care of all _common_ setup and validation steps for a request - loading a method definition, creating a client, preparing a message as needed, attaching status listeners, saving to the `callCache`, etc.

The `start` function will determine what kind of RPC is being called, and automatically call the non-exported `_make{RPC}Request` functions, which are responsible for doing any RPC _specific_ setup steps, such as adding server response listeners or preparing a message body during the first call.

This refactor adds and removes a similar amount of code but it removes a lot of duplication. It also consolidates the _untyped_ communications over IPC to reduce the possibility of bugs in that communication layer, which needed to be modified to support template tags.

Much of this logic is now easier to test as well and tests have been added as such with ~95% coverage (up from 0% as there were no tests with the first implementation).

sonicyeti (approved):
Very nice, great coverage too.

packages/insomnia-app/app/common/render.js
```diff
@@ -445,6 +516,7 @@ export async function getRenderedRequestAndContext(
   };
 }
 
+// TODO: I don't think this is used anywhere
```
DMarby: I'd prefer a linear ticket or similar to be filed rather then a TODO in the codebase if we want to do something actionable here (such as remove the function). This comment isn't very constructive as it references self, and isn't actionable.

It also seems like this function is currently used in `har.test.js` and `network.test.js`.

packages/insomnia-app/app/models/helpers/__tests__/query-all-workspace-urls.test.js
```diff
@@ -0,0 +1,96 @@
+// @flow
+
+import { globalBeforeEach } from '../../../__jest__/before-each';
+import * as models from '../../index';
+import { queryAllWorkspaceUrls } from '../query-all-workspace-urls';
+
+describe('queryAllWorkspaceUrls', () => {
+  beforeEach(globalBeforeEach);
+
+  it('should return empty array when no requests exist', async () => {
```
DMarby, line 10: I don't think this reflects what the test checks for

packages/insomnia-app/app/common/render.js
develohpanda: OH yes, my bad. I added it quite early on and didn't want to investigate further at the time. I think I'll remove it and update the tests, since it's not actually used in business logic.

Normally I do try to create liner tickets for TODOs 🙈

DMarby: Sounds good, is there any possibility this could be in-use by a plugin or such that might break?

packages/insomnia-app/app/common/render.js
```diff
@@ -344,14 +349,80 @@ export async function getRenderContext(
...
+
 export async function getRenderedRequestAndContext(
   request: Request,
   environmentId: string | null,
   purpose?: RenderPurpose,
   extraInfo?: ExtraRenderInfo,
 ): Promise<{ request: RenderedRequest, context: Object }> {
   const ancestors = await db.withAncestors(request, [
-    models.request.type,
+    models.grpcRequest.type,
```
DMarby: Could you explain this change?
I see that in `getRenderedGrpcRequest` and `getRenderedGrpcRequestMessage ` we only grab `models.request.type` instead, so opposite of this, but I don't quite understand the reasoning.

packages/insomnia-app/app/common/render.js
```diff
@@ -344,14 +349,80 @@ export async function getRenderContext(
...
+
+  // Render all request properties
+  const renderedRequest = await render(request, renderContext, ignorePathRegex);
+
+  renderedRequest.description = await render(description, renderContext, null, KEEP_ON_ERROR);
+
+  // Remove disabled headers
+  if (renderedRequest.headers) {
+    renderedRequest.headers = renderedRequest.headers.filter(p => !p.disabled);
+  }
```
DMarby: Can be removed since the grpc model doesn't have headers?

DMarby (approved):

packages/insomnia-app/app/common/render.js
develohpanda: I doubt it, but I will have a look through the plugin context logic to see if it's consumed

packages/insomnia-app/app/models/helpers/__tests__/query-all-workspace-urls.test.js
develohpanda: 🤦🏽

packages/insomnia-app/app/common/render.js
develohpanda: Good catch, this shouldn't have changed but I believe the previous implementation is redundant which is why this didn't cause an immediate issue. 

I wrote a `db.withAncestors` test to understand how that logic works and it confirmed my suspicion that the request/grpcRequest type doesn't need to be in the array, only requestGroup and workspace.

I also looked through the code paths and couldn't see immediately a case where anything other than requestGroup or workspace needs to be in the array. The _only_ ancestor a request or grpcRequest can have is a requestGroup and a workspace.

For starters, I consolidated the ancestor loading logic for this file in one function that is used as necessary, to make sure the type arrays don't vary (they shouldn't).

Secondly, in theory, we can remove the request and grpcRequest types from the type array, but I _am_ a little nervous about making that change. 😅

packages/insomnia-app/app/common/render.js
develohpanda: I didn't detect a use for it in the plugin logic. I checked for imports of this file itself, the import contents, and how those imports are used in any of the plugin related logic.

DMarby (approved):

PR #2927, opened 2020-12-10
Tooltip with an initial empty message does not unmount correctly

develohpanda:
This is an interesting bug and occurs through the following process:
- mount Tooltip with no message
- update Tooltip to contain a message
- cause Tooltip to unmount <- this throws an error

An example of where this occurs is when switching from a valid gRPC request to a REST request.
- create a grpc request and select a method
- create a rest request and switch to it

image

The root cause is that the element is moved to the `tooltips-container` _only_ on mount, but if a message does not exist in the initial props then there is [no tooltip that is created](https://github.com/Kong/insomnia/blob/cad2ebc93204176d4fb472d6cb6a948903b288cb/packages/insomnia-components/components/tooltip.js#L204-L206) initially. When a component update occurs and the message does exist, the tooltip is created on the dom as expect, _but it is not moved to the `tooltips-container`_. Later, when the component attempts to unmount and remove the element, the element is found successfully but it is not under the expected parent, hence the error above is shown.

The tooltip in question is on the [`grpc-method-dropdown-button`](https://github.com/Kong/insomnia/blob/cad2ebc93204176d4fb472d6cb6a948903b288cb/packages/insomnia-app/app/ui/components/dropdowns/grpc-method-dropdown/grpc-method-dropdown-button.js#L29-L34), where the message (`fullPath`) is optional. This was introduced in #2883 but the Tooltip bug has existed since before.
 
Fixes INS-359

### DIFF
--- a/packages/insomnia-components/components/__tests__/tooltip.test.js
+++ b/packages/insomnia-components/components/__tests__/tooltip.test.js
@@ -47,7 +47,7 @@ describe('<Tooltip />', () => {
     expectNoTooltip(queryByRole);
   });
 
-  it('should add tooltip is message is empty first then updates', async () => {
+  it('should unmount successfully if message is empty in the first render then and non-empty after update', async () => {
     const childText = 'some child';
     const initialMessage = '';
     const newMessage = 'message';
--- a/packages/insomnia-components/components/__tests__/tooltip.test.js
+++ b/packages/insomnia-components/components/__tests__/tooltip.test.js
@@ -4,13 +4,18 @@ import Tooltip from '../tooltip';
 
 const pause = time => new Promise(resolve => setTimeout(resolve, time));
 
+const expectNoTooltip = queryByRole => {
+  expect(queryByRole('tooltip', { hidden: true })).toBeNull();
+  expect(queryByRole('tooltip')).toBeNull();
+};
+
 describe('<Tooltip />', () => {
   it('should show and hide the tooltip correctly', async () => {
     const childText = 'some child';
     const delay = 200;
     const message = 'message';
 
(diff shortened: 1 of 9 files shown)


packages/insomnia-components/components/__tests__/tooltip.test.js
```diff
@@ -0,0 +1,77 @@
...
+    const childText = 'some child';
+    const message = '';
+
+    const { queryByRole, getByText } = render(<Tooltip message={message}>{childText}</Tooltip>);
+
+    expectNoTooltip(queryByRole);
+    expect(getByText(childText)).toBeTruthy();
+  });
+
+  it('should unmount successfully if message is empty in the first render then and non-empty after update', async () => {
```
develohpanda, line 55: This is the breaking test for the bug identified. If you comment the `componentDidUpdate` logic in `Tooltip` from this PR, this test will break.

packages/insomnia-components/package.json
```diff
@@ -8,24 +8,47 @@
...
+    "resetMocks": true,
+    "resetModules": true,
+    "coveragePathIgnorePatterns": [
+      "/node_modules/"
+    ]
+  },
   "devDependencies": {
     "@babel/core": "^7.7.7",
     "@babel/plugin-proposal-decorators": "^7.8.3",
+    "@babel/plugin-proposal-optional-chaining": "^7.12.7",
```
develohpanda, line 38: Not _strictly_ needed for this PR in the final state, but I've been putting it off for ages and an initial iteration of the fix used it 🙂

sonicyeti (approved):
Great sleuthing!

DMarby (approved):

Issue #203, opened 2017-05-12
[Feature] Per-request Variables


gschier:
## Overview

A common use-case for Insomnia is to define a set of requests up front, then only edit a small subset of values afterward (resource IDs, URL path segments, tokens, etc). This can be quite cumbersome if you need to frequently change a value inside a deep JSON structure, for example.

## Solution 1 – Using placeholders

Placeholders could be defined in a request and Insomnia could detect these and provide an easy key-value editor to change/update them.

This could be done with a custom Nunjucks tag like `{% placeholder "my name" %}`, or maybe show an input for all variables that aren't defined in the currently active environment.

## Solution N

_your idea here_

bkonia:
I agree, this would be very useful. I'd like to be able to enter placeholders in the query parameters. When you submit a request, if the parameters have placeholders, it should automatically display a dialog for the user to input the values for the specific request. This would be much more elegant than manually editing the values in the query fields each time you submit a request.

bkonia:
Another example of how this would be useful...

I'm working with a URL format in which the just the parameter value is appended to the URL, without the parameter name. For instance:

https://www.xzy.com/path/to/api/123

Presently, there's no way to automatically build this format in the query tab. I can manually insert a placeholder in the URL, but then I'd have to manually edit the environment each time I want to use a different value. If we had the ability to create per-request variables, then I could insert the placeholder in the URL and it would prompt me for the value each time I submit the request.

bkonia:
So after reflecting on this a bit more, I'm thinking these placeholders should insert raw values, exactly as entered by the user, not a key/value input like in the Query tab. That would give us the flexibility to enter formats like the one I described, without Insomnia messing it up by trying to turn it into a standard query string key/value pair.

Another approach would be to allow the user to select the format on a per-line basis in the Query tab and also in this new per-request tab I'm proposing. You could have a little toggle switch next to each item, to allow the user to select whether the variable will be inserted in raw format or in key/value format.

mgraham:
Definitely it should not be limited to query string key/value pairs.

@bkonia, if I'm understanding you correctly, you would want something like this:

```
https://www.xzy.com/path/to/api/{% placeholder "my name" %}
```

And then in the "placeholders" tab there would be a list of key/value pairs:

```
my name:  123
```

This would work for me.

However since I would frequently want to edit both query parameters and placeholders for the same request, it would be nice to have them both on the same tab, maybe in a split screen.

bkonia:
[...]

PR #2936, opened 2020-12-13
add path parameters

Vap0r1ze:
I'm coming from issue #203 and wanted to open a PR to discuss my changes. Currently, how I have it working is you add entries to a table in a new "Params" tab, and it replaces the instances of `:${paramName}` (colon preceding parameter name), with the value of said param. I think I should add a visual indicator in the RequestUrlBar for params, similar to how environment variables appear. This is very WIP, so please let me know what must be changed or what is a bad idea. Thanks for being part of this wonderful app!

Image of Functionality:

image


Nisgrak:
This should be a priorized feature.

eltonsandre:
feature that will help a lot, I look forward to this version.

apisaint:
Reading through the code, looks nice so far - great work, ~~haven't determined whether it's backwards compatible yet~~ it is!

> I think I should add a visual indicator in the RequestUrlBar for params, similar to how environment variables appear.

💯% It would be great if the indicator was similar to the env variables, and when you click on it, it focuses the parameter in the params table.

plugins/insomnia-plugin-request/index.js
```diff
@@ -177,15 +178,25 @@ module.exports.templateTags = [
...
+  if (request.pathParameters) {
+    for (const pp of request.pathParameters) {
+      pathParameters.push({
+        name: await context.util.render(pp.name),
+        value: await context.util.render(pp.value),
+      });
+    }
+  }
 
+  const templateUrl = insertPathParameters(url, pathParameters);
```
apisaint, line 197: Nitpick, `templateUrl` makes me think of a regular expression or a template, rather than a compiled output.

packages/insomnia-url/src/path-parameters.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Insert path parameters into the templating within the url
+ * @param url {string} - URL to insert path parameters into
+ * @param pathParameters {{name: string, value: string}[]}
+ * @returns {string}
+ */
+module.exports.insertPathParameters = function(url, pathParameters) {
+  for (const pathParam of pathParameters) {
+    url = url.replace(`:${pathParam.name}`, pathParam.value);
```
apisaint, line 9: Future enhancement could be to add a checkbox for regex replacement support (aka global). That way you can set one parameter for `:user` and it would replace all instances vs the current implementation you'd have to define a `:user` parameter for each usage.

edbrannin:
Maybe I should be suggesting this in the parent issue, and I'm not a maintainer, but:

Would Placeholders be able to supplement Environment variables and use the same syntax?

That way I could, for example, have a request like `{{ my_service.url }}/api/v1/user/{{ userId }}/attributes` with `my_service: { url: "..." }` in the environment and `userId` in the parameters tab.

I thought for a minute that would preclude @apisaint's suggestion (https://github.com/Kong/insomnia/pull/2936#issuecomment-756576846):
> It would be great if the indicator was similar to the env variables, and when you click on it, it focuses the parameter in the params table

...but maybe in this case it would focus on the params table if it's defined there, and open the Environment Variable prompt if it's not?

EDIT: Conversation in the parent issue is also asking about `{{ my_service.url }}/api/v1/user/:userId/attributes` as an option.

scollovati:
+1 on this feature, since it tries to incorporate by default the `Path Variables` as in Postman.
There is this plugin as a temporary workaround https://github.com/gschier/insomnia-plugin-path-parameters, but the url preview does not work.

danshapir:
Any updates?

eltonsandre:
any update on this? Is work still ongoing?

develohpanda:
The parent issue was converted to a discussion and has a response on it to address this [#3551 (comment)](https://github.com/Kong/insomnia/discussions/3551?sort=new#discussioncomment-952413)

ppsirius:
@apisaint what's going on with this pull?
It's a crucial feature why you don't merge it?

packages/insomnia-url/src/path-parameters.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Insert path parameters into the templating within the url
+ * @param url {string} - URL to insert path parameters into
+ * @param pathParameters {{name: string, value: string}[]}
+ * @returns {string}
+ */
+module.exports.insertPathParameters = function(url, pathParameters) {
+  for (const pathParam of pathParameters) {
```
ts14ic, line 8: I'd like to note, that given a URL like `https://example.com/:foo/:fooBar`, a replacement with `{foo: 10, fooBar: 20}` will result in `https://example.com/10/10Bar`. 
To fix this, pathParameters must be sorted by name length in descending order.
This way, `:fooBar` gets replaced with 20 first, and only after that `:foo` gets replaced with 10, without breaking `:fooBar`

ts14ic (changes requested):
Found Insomnia some days ago. Blown away by how flexible and cool it is. 

Given `insertPathParameters` is fixed, and all else (code generation?, generation from swagger/open api?) works OK, can't wait for this to become part of main branch :). 
I don't see any substantial difference between query and path params, so don't understand why so much discussion was needed about this feature.

mugi-luffy:
shouldn't the path parameters be like {paramName} instead of :paramName ?
This will conform with open api 3.0 [https://swagger.io/docs/specification/describing-parameters/](https://swagger.io/docs/specification/describing-parameters/)
The request generation in insomnia should also be fixed, because now {paramName} in design gets converted to {{paramName}} in debug 
image
image

eltonsandre (approved):

edbrannin:
> shouldn't the path parameters be like `{paramName}` instead of `:paramName`?

I've seen both in use, so it may be useful to support both somehow (with or without individual toggles)

jackkav:
Closed as stale.

Nisgrak:
This won't be merged? What a foot shot....

robross0606:
What is the deal here?  This was closed without merge.  Meanwhile, feature requests pointing to this MR are closed as "Completed".

PR #2946, opened 2020-12-15
Serve csv and pdf files locally

develohpanda:
What it says on the tin. This PR initially started as a way to test a response viewer with `text/markdown` but seems like we don't show a rich markdown preview.

Fixes INS-371


packages/insomnia-smoke-test/server/index.js
```diff
@@ -15,11 +15,10 @@ app.get('/pets/:id', (req, res) => {
...
-app.get('/csv', (_, res) => {
-  res
-    .status(200)
-    .header('content-type', 'text/csv')
-    .send(`a,b,c\n1,2,3`);
+const options = { root: 'fixtures' };
+
+app.get('/file/:fileName', (req, res) => {
+  res.sendFile(req.params.fileName, options);
 });
```
DMarby: Express has a helper for serving static files which we can use here. It also handles subdirectories, and prevents potential directory traversal and similar.

```suggestion
app.use('/file', express.static('fixtures'));
```

DMarby (approved):
Looks good, just a minor nitpick about how we're serving the static files.

packages/insomnia-smoke-test/server/index.js
develohpanda: Awesome! This is perfect

PR #2947, opened 2020-12-16
Smoke and Unit testing documentation

develohpanda:
What it says on the tin


DEVELOPMENT.md
```diff
@@ -55,10 +57,22 @@ Insomnia stores data in a few places:
...
 *Eventually, Redux could/should be removed, which would both reduce memory overhead and simplify
 the codebase. NeDB should essentially replace it*
 
+## Automated testing
+
+Unit tests are written using [Jest](https://jestjs.io/), [react-testing-library](https://testing-library.com/docs/react-testing-library),
+while integration tests are written using [Spectron](https://www.electronjs.org/spectron).
+
+Unit tests exist alongside the file under test. For example:
+- `/app/commom/database.js` contains the database business logic
```
DMarby:
```suggestion
- `/app/common/database.js` contains the database business logic
```

packages/insomnia-smoke-test/README.md
```diff
@@ -1,10 +1,107 @@
...
+
+You may also need to run a test multiple times. You can focus a particular test or test suite using [Jest globals](https://jestjs.io/docs/en/api#testonlyname-fn-timeout) such as `it.only` and `describe.skip`, or their aliases `fit` and `xdescribe`, etc.
+
+## General guidelines
+
+### Data
+Individual tests will automatically run against a clean Insomnia data directory to keep data isolated.
+
+### Dependencies
+A test should not depend on any external services unless absolutely necessary. If a particular endpoint is required (eg. for authentication or a specific content type), simply implement a new endpoint in `/server` .
```
DMarby: Nitpick, but it's good to avoid using `simply` in documentation as it can be seen as non-inclusive/trivialising.

packages/insomnia-smoke-test/README.md
```diff
@@ -1,10 +1,107 @@
...
+
+npm run test:smoke:core:package     # Run Core tests 
+npm run test:smoke:core:designer    # Run Designer tests
+```
+
+Each of the above commands will automatically run the Express server, so you do not need to take any extra steps.
+
+## How to write
+
+When writing tests, I suggest to use the scripts in this project directly (instead of from the root, as per the section above). After building and/or packaging your application under test, it will be available under `packages/insomnia-app/{build|dist}` and you can begin writing your test.
```
DMarby: Nitpick, but generally documentation doesn't reference self, so this should perhaps be reworded to `It is recommended to use the scripts in this project directly` or similar.

packages/insomnia-smoke-test/README.md
```diff
@@ -1,10 +1,107 @@
...
+
+You can then run the smoke tests, again from the root:
+
+``` sh
+npm run test:smoke:cli              # Run CLI smoke tests
+npm run test:smoke:core:build       # Run Core smoke tests
+npm run test:smoke:designer:build   # Run Cesigner smoke tests
+```
+
+Sometimes, you need to run tests against a _packaged_ application. Packaging takes longer to do and is only required for edge cases (such as a [plugin installation](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-smoke-test/designer/app.test.js#L36)). To run packaged tests, from the root:
```
DMarby: Might be good to make this link towards a specific commit, as things are likely to be moved around in this file in the future.

I'd also consider rewording this to something like `Sometimes, you might need to run tests against a packaged application`, and adding an explanation of what a `packaged application` means.

DMarby (approved):
Looks good overall, very nice guidelines for the smoke tests 👍

packages/insomnia-smoke-test/README.md
```diff
@@ -1,10 +1,107 @@
...
+It is important for a smoke test to be _readable_ so the flow can be understood, and the (often complicated) implementation details hidden, like in the example below.
+
+![example-test.png](./assets/example-test.png)
+
+In most cases, it will be beneficial to create helper functions under `/modules`, regardless of how reusable they are. Some modules (such as `dropdown`, `tabs` and `settings`) are reusable, while some are specific to certain pages (eg `debug`, `home`, `onboarding`). These can be broken down into more granular modules as the test suite grows.
+
+
+## Contributing a smoke test?
+
+Smoke tests are often flaky, and one attempt to avoid that is to run the final implementation of a test atleast 20 times locally to prove its stability. If a test is unable to achieve this, it is very unlikely to be accepted into the test suite.
```
DMarby: I think this should be reworded to something along the lines of `Smoke tests can potentially be flaky`, as flaky tests are an exception or an indication of an underlying problem, not the norm.

DEVELOPMENT.md
```diff
@@ -55,10 +57,22 @@ Insomnia stores data in a few places:
...
 - A local in-memory NeDB database stores data for data models (requests, folder, workspaces, etc).
 - A local Redux store contains an in-memory copy of all database entities.
+- Multiple React Context stores, defined in `/app/ui/context`.
 
 *Eventually, Redux could/should be removed, which would both reduce memory overhead and simplify
 the codebase. NeDB should essentially replace it*
 
+## Automated testing
+
+Unit tests are written using [Jest](https://jestjs.io/), [react-testing-library](https://testing-library.com/docs/react-testing-library),
```
DMarby: Nitpick, but could perhaps come off as a bit abrasive, I'd consider rewording it to something along the lines of
```
We use Jest and react-testing-library to write our unit tests, and Spectron for integration tests.
```

sonicyeti (approved):

packages/insomnia-smoke-test/README.md
develohpanda: Switched to a permalink 👍🏽

DMarby (approved):

PR #2953, opened 2020-12-17
Add a button to clear oauth2 from the auth options

### LINKED ISSUE #2902: Move or replicate "Clear OAuth 2 session" button location
Hello!

When using OAuth2 with authorization flow we usually need to test some APIs with different users. There is no quick way to clear the authorization session besides the "Clear OAuth 2 session" in the preferences dialog.

I would suggest to bring or replicate the `Clear OAuth 2 session` button to the `Advanced Options` in the auth tab. This would allow changing authenticated users faster than the current way.

This is the button I'm talking about:

image

Best regards,
Rafael Pacheco.

gravyboat:
Closes #2902

This uses the same `initNewOAuthSession` as the preferences menu to ensure the functionality is identical.

@rafaelrenanpacheco is this along the lines of what you were thinking?

oauth2-clear-session


develohpanda (approved):
Very cool! 😎

rafaelrenanpacheco:
Exactly like that @gravyboat, very nice!

Thanks for this PR :+1:

nijikokun (approved):
Like the placement, and the functionality.

Makes me think we should look into something like profiles/session-store. So users can save sessions, and swap between them. Seems like issues surrounding clearing and changing consistently crop up and would be solved by a more robust profile/session-store style solution.

Secondly a minor nitpick, it looks like this button is _always available_, seems like it should be disabled unless a session exists.

gravyboat:
> Secondly a minor nitpick, it looks like this button is always available, seems like it should be disabled unless a session exists.

This is correct, I left it that way since that's how it operates in the `preferences` menu currently as well.

Edit: If `LOCALSTORAGE_KEY_SESSION_ID` was exported the check could be changed to look like:
```
showAdvanced && window.localStorage.getItem(LOCALSTORAGE_KEY_SESSION_ID)
```

It could also be updated to check that const in the preferences menu but that seems like it should probably be a second PR.

develohpanda:
I agree with you @gravyboat, my preference will be a separate PR as it is a separate request.

I have noticed that the action does not have any kind of feedback either, so that would be something to also consider. ☺

PR #2954, opened 2020-12-17
GraphQL Explorer - Make text selectable

### LINKED ISSUE #2899: Feature Request: Copy Text In Schema Documentation Panel
I'm using `2020.5.0` of Insomnia on Mac. 
It would be really useful if I could select and copy the text in the schema documentation panel (the panel on the right from the screenshot). Right now the text is not selectable.

Screen Shot 2020-12-03 at 11 36 16 AM

lykims:
Closes #2899

By inspecting elements, the `user-select` with `none` seems to come from the [body](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/css/lib/chrome/platform_app.css#L3).

However, I don't think it cannot be removed or modified because all texts would be selected.
insomnia_body_select

The next level for CSS change would be for the [GraphQL Explorer](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/renderer.html#L16).
insomnia_graphql_explorer_select

### DIFF
--- a/packages/insomnia-app/app/ui/css/components/graph-ql-explorer.less
+++ b/packages/insomnia-app/app/ui/css/components/graph-ql-explorer.less
@@ -12,6 +12,9 @@
   background: var(--color-bg);
   box-shadow: 0 0 1rem 0 rgba(0, 0, 0, 0.4);
   color: var(--color-font);
+  -webkit-user-select: text;
+  -moz-user-select: text;
+  user-select: text;
 
   .markdown-preview__content *:first-child {
     margin-top: var(--padding-sm);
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/css/components/graph-ql-explorer.less
```diff
@@ -12,6 +12,9 @@
   background: var(--color-bg);
   box-shadow: 0 0 1rem 0 rgba(0, 0, 0, 0.4);
   color: var(--color-font);
+  -webkit-user-select: text;
+  -moz-user-select: text;
+  user-select: text;
```
develohpanda: Because this is an electron only app and we don't need to think about other browsers yet, so it should be safe to omit the `-webkit` and `-moz` styles, just keeping the regular `user-select`. 🙂 

```suggestion
  user-select: text;
```

develohpanda (commented):
Seems reasonable, thanks for raising a PR and explaining your process!

packages/insomnia-app/app/ui/css/components/graph-ql-explorer.less
lykims: Thank you for the suggestion @develohpanda ! I just pushed the change.

develohpanda (approved):
🎉

PR #2957, opened 2020-12-17
Close hint dropdown when a code editor field unmounts

### LINKED ISSUE #2591: The autocomplete dropdown won't close
This problem has happened to me many times. When I click on a field and the autocomplete shows up, if I rapidly click on another request, the dropdown won't leave, no matter where I click. I either have to restart the application or click on one of the options, which will autocomplete the field and then I have to correct it.
This gif shows the problem: (I have added a white target to hide some sensitive URI from a public API.

Untitled Project

 - Windows 10 Pro 120.2212.31.0
 - Installation Method: exe file
 - App Version 2020.3.3

develohpanda:
The bug described in the linked issue is a little tricky to reproduce but before reviewing try to get the hang of it by pressing `ctrl + space` and at the same time switch to a new request in the live application.

If you get the timing right, the request actions dropdown and autocomplete dropdown will open together, and then the application is in the state described in the issue, where the autocomplete dropdown doesn't close on switching requests.

Following the same process, the dropdown will correctly close with this PR (as shown in the gif below)

Closes #2591

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
@@ -37,6 +37,10 @@ CodeMirror.defineExtension('isHintDropdownActive', function() {
   );
 });
 
+CodeMirror.defineExtension('closeHintDropdown', function() {
+  this.state.completionActive?.close();
+});
+
 CodeMirror.defineOption('environmentAutocomplete', null, (cm, options) => {
   if (!options) {
     return;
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.js
```diff
@@ -37,6 +37,10 @@ CodeMirror.defineExtension('isHintDropdownActive', function() {
   );
 });
 
+CodeMirror.defineExtension('closeHintDropdown', function() {
+  this.state.completionActive?.close();
```
apisaint, line 41: Nice, was going to comment that this might not be set but looks like you took care of it!

apisaint (approved):
Was able to reproduce, and tested locally, works. Nice work. LGTM.

ValorMorgan:
I have a similar issue: https://github.com/Kong/insomnia/issues/2991
Though mine isn't based on request switching but URL textbox input delay timing.

sonicyeti (approved):

Issue #2468, opened 2020-07-31
Remove usages of @autobind to allow for react-hot-reloading to work


develohpanda:

MrSnix:
Is this still requested?

If so, what should we do in order to deliver it?

Considering the removal of the `@autobind` decorator and its import, transforming all the functions to arrow functions should solve the .bind() problem. Isn't it? 

**EDIT:**
I was able to perform the transformation from `common function to arrow function` on all the classes using auto bind inside the project, then I successfully removed the decorator and its declaration from the insomnia-app & insomnia-component and it's running like before. Here is a branch on my fork with the transformation: [tech-debt/remove-autobind](https://github.com/Kong/insomnia/commit/4506984e9c1160e45e6976f4efa0f933256f21cd)

Let me know if it can be helpful or I am missing something.

develohpanda:
This _is_ still requested but using arrow functions to replace the `.bind()` is not as clear-cut a solution as it may seem at first sight. Depending on the implementation, arrow functions will be recreated on each render and sent via props to child components. Even if the function implementation does not change, React will detect it as a prop change and cause a re-render. To my understanding, this is generally discouraged because it is inefficient.

I need to put some more thought towards the approach but haven't been able to yet.

PR #2963, opened 2020-12-19
autobind removal and react-hot-reload fix

MrSnix:
As requested by @develohpanda, I'm opening this PR as a proposal about issue #2468.

No precise specifications had been given on how to perform the removal of `@autobind` so with a bit of research I found out we may use the `@babel/plugin-proposal-class-properties` which **is currently declared by Insomnia** and transform all the functions to arrow functions (so `this` is lexically bound and we do not need to use `.bind(this)` anymore) and drop the decorator usage. Why do we need to define arrow functions as class properties? So React.js won't recreate them each time as an inline-arrow function.

Source: [StackOverflow](https://stackoverflow.com/questions/36677733/why-shouldnt-jsx-props-use-arrow-functions-or-bind)
Please **check the best** answer (more **specifically** `Binding methods to this without inlining arrow functions` on point 2.)

**Changes**:
The first commit on this PR :
- Removed the `@autobind` declaration mainly on insomnia-app and insomnia-components.
- Changed all functions to arrow functions.

The second commit on this PR:
- Updated the react-hot-reloading configuration in order to work again and it's dependencies.

Here is a gif that shows react-hot-reloading working again
 
Fix

I'd be glad to hear feedback.
Thanks


apisaint:
Wow, this is a *massive* change. *sweating intensifies*

packages/insomnia-app/app/ui/components/base/button.js
```diff
@@ -1,10 +1,8 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import autobind from 'autobind-decorator';
 
-@autobind
 class Button extends PureComponent {
-  _handleClick(e) {
+  _handleClick = e => {
```
apisaint, line 5: I would suggest to keep parenthesis even when there is a single value.

MrSnix: Parenthesis are automatically removed by Prettier on a single argument function.
I need to **update the Prettier configuration** in order to make it work.

More specifically, I'll **add** this rule:
```
"arrowParens": "always"
```
Then I'll have to run `eslint: fix` across the codebase

Can I do that?

develohpanda: I suggest not making this change at this stage/in this PR.

apisaint: Ah, makes sense that this is something from the formatter. It's a really bad practice to remove parens around single properties, an edge-case, inconsistency in code - making it way more difficult to read through. Unfortunate. Hope the maintainers really change their stance, and a maintainer with more experience chimes in given the lack of conversation from above - especially given the formatter would fix it quickly.

Good timing especially with a change this large. 

That aside, going to read through the changes.

develohpanda: Hi @apisaint, this topic is wholly unrelated to the contents and purpose of this PR. In order to keep conversation on topic and avoid scope creep on an, as you identified, already large PR, I advise away from altering it in this PR and chose to not engage in further reasoning.

Please raise a separate PR, or an issue with your notes and reasoning, and it can be addressed by somebody. 

> Hope the maintainers really change their stance, and a maintainer with more experience chimes in given the lack of conversation from above. 

I don't have an opinion either way as it is an automated change which has little impact during development, but this is rude and misinformed.

apisaint: If you don't have an opinion then I suggest that we make the change. It's directly related to this PR as it introduces thousands of cases. Apologies if you took it personally, but it's from observation and I don't have a way to privately request from those who seem more experienced based on their contributions!

develohpanda: The changes introduced in this changeset are not the only arrow functions in the codebase. When the new rule is applied, it will modify more files in addition to those currently changed. 

Adding this linting rule and fixing all instances, in this PR, is scope creep and a significant overhead for reviewing the contents of _this_ PR. I am not opposed to this rule being added, but it must be a separate pull request.

You are more than welcome to tag in a comment and request a review, but the GitHub reviewers panel is not available (to edit) to contributors.

apisaint: > The changes introduced in this changeset are not the only arrow functions in the codebase. 

Of course, nobody is arguing about that. However, this PR changes consistent functions into inconsistent same-level ones. That's the difference and is what was brought up.

> Adding this linting rule and fixing all instances, in this PR, is scope creep

Nobody said it must be done in this PR. It was suggested to change them. Then, it was found that it was the formatter enforcing it. It can be done in a separate PR. Nobody has an issue with that.

> You are more than welcome to tag in a comment and request a review, but the GitHub reviewers panel is not available (to edit) to contributors.

What I do have an issue with is how you are communicating, from the first comment to now has been deeply personal rather than collaborative and the point you raised at the end does not go against any rules set forth by the organization or the repository. They are your own personal opinions - which personally I disagree with, and would reduce the ability of the community to grow. I will abide by the rules written by the organization - which I have.

apisaint: For the sake of not pinging Snix further I will move this to a new location as I do not want to muddle his PR further. Hopefully, we can have a collaborative conversation there.

develohpanda: In that case it would be worthwhile to introduce the rule (and fix instances that break) introduced to develop first, in preparation for this PR. This is to ensure the modifications in this PR relate only to the linked issue. 

I respect your review and the changes you have requested, but want to avoid tacking on a large changeset onto an already large PR. 👍

develohpanda: > They are your own personal opinions - which personally I disagree with, and would reduce the ability of the community to grow. I will abide by the rules written by the organization - which I have.

Lastly, I will say that the configuration of this repository is ever changing, and the documentation or rules are complete, to our best efforts.

My suggestion to tag whomever you feel appropriate is used throughout this codebase albeit not formally documented. This approach, and which PR panels are editable by contributors is _not_ my personal opinion. If it is something you would like to see changed, we are open to explore those configuration changes. 🙂

> Nobody said it must be done in this PR.

The author of this PR asked after his initial investigation whether this rule can be added, which to my understanding was directed at this PR. 🙂

apisaint: I waited a bit before responding, I really didn't want to. That message makes me not want to contribute. I'm going to assume you are coming from a good place, but when you use smilies and language in such a way it comes off as very passive aggressive and provoking and doesn't bode well for future interactions. I hope that going forward we can have a better dialog and remain professional.

I also apologize for some confusion here you meant the panel in the sidebar, English is not my first language, I thought you meant the review feature itself, the panel - of course - I can only ping 😅

develohpanda: Hi @apisaint! Please know that I do not intend for any comments to be aggressive, provoking or unprofessional at any time. Especially, the brevity of my initial answer was not to ignore or set aside the point that was raised, but perhaps I should have elaborated my reasoning sooner. I apologize for the part I played in our miscommunication and misunderstanding; this is an exception, I'm sure we will have fruitful dialog in future. I respect your feedback here and will take it on board, to the best of my ability.

I'm glad we have an agreement on the resolution (and I see you have already raised a PR - thanks). 😄

Lastly, I've noticed you have been responding to and addressing several PRs and issues, recently. We are a small team split across various aspects of Insomnia as a complete product and I, personally, appreciate the help. 🙏🏽

packages/insomnia-app/app/ui/components/key-value-editor/row.js
```diff
@@ -549,34 +548,30 @@ const dragTarget = {
...
 
 target.prototype.focusValueEnd = function() {
-  this.decoratedRef.current.decoratedRef.current.focusValueEnd();
+  this.decoratedRef.current?.decoratedRef?.current?.focusValueEnd();
 };
 
 target.prototype.focusDescriptionEnd = function() {
-  this.decoratedRef.current.decoratedRef.current.focusDescriptionEnd();
+  this.decoratedRef.current?.decoratedRef?.current?.focusDescriptionEnd();
 };
```
develohpanda, line 575: Nice catch! 👍🏽

packages/insomnia-app/app/ui/components/viewers/response-headers-viewer.js
```diff
@@ -10,15 +10,15 @@ type Props = {
...
+const validateURL = urlString => {
   try {
     const parsedUrl = new URL(urlString);
     if (!parsedUrl.hostname) return false;
     return true;
   } catch (error) {
     return false;
   }
-}
+};
```
develohpanda, line 21: I don't think this one needs to be updated, but it's ok! 🙂

packages/insomnia-app/app/ui/components/viewers/response-viewer.js
```diff
@@ -110,19 +109,19 @@ class ResponseViewer extends React.Component<Props, State> {
...
 
   // eslint-disable-next-line camelcase
-  UNSAFE_componentWillReceiveProps(nextProps: Props) {
+  UNSAFE_componentWillReceiveProps = (nextProps: Props) => {
     this._maybeLoadResponseBody(nextProps);
-  }
+  };
 
-  shouldComponentUpdate(nextProps: Props, nextState: State) {
+  shouldComponentUpdate = (nextProps: Props, nextState: State) => {
```
develohpanda: I don't think it actually matters, but because React's lifecycle functions are never directly used, it might be better to not change these to arrow functions 🤔 thoughts?

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -334,27 +333,26 @@ class App extends PureComponent {
...
-    function __updateDoc(doc, metaSortKey) {
-      return models.getModel(doc.type).update(doc, { metaSortKey });
-    }
+  _recalculateMetaSortKey = async docs => {
+    const __updateDoc = (doc, metaSortKey) =>
+      models.getModel(doc.type).update(doc, { metaSortKey });
 
     return Promise.all(docs.map((doc, i) => __updateDoc(doc, i * 100)));
-  }
+  };
```
develohpanda, line 353: Strange that this is an `async` function, but contains no `await`. Looks like there _is_ an await where it is used, on line 367, so it's okay.

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -423,14 +421,14 @@ class App extends PureComponent {
       submitName: 'Rename',
       selectText: true,
       label: 'Name',
-      onComplete: async name => {
+      async onComplete(name) {
```
develohpanda, line 424: In case I'm missing something, was there a particular reason for this change? Preference would be to keep it as before 🙂

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1417,7 +1415,7 @@ App.propTypes = {
   }),
 };
 
-function mapStateToProps(state, props) {
+const mapStateToProps = (state, props) => {
```
develohpanda: Top level functions don't need to be changed, so I'd favor reverting the changes from here to the end of this file

packages/insomnia-app/app/ui/index.js
```diff
@@ -42,25 +42,17 @@ document.title = getAppLongName();
...
-          <DnDComponent />
-        </Provider>
-      </AppContainer>,
-      document.getElementById('root'),
+    const app = () => (
+      <Provider store={store}>
+        <DnDComponent />
+      </Provider>
     );
+    const TheHottestApp = hot(module)(app);
```
develohpanda, line 50: 😆 🔥

packages/insomnia-app/webpack/webpack.config.development.babel.js
```diff
@@ -39,12 +39,13 @@ module.exports = {
...
+  optimization:{
+    noEmitOnErrors: true,
+    moduleIds: "named"
+  },
   plugins: [
     ...baseConfig.plugins,
     new webpack.LoaderOptionsPlugin({ debug: true }), // Legacy global loader option
-    new webpack.HotModuleReplacementPlugin(),
-    new webpack.NoEmitOnErrorsPlugin(),
-    new webpack.NamedModulesPlugin(),
```
develohpanda: I'm assuming this change is because the plugins are now deprecated; is it necessary or avoidable for hot-reloading? If avoidable can it please be moved to a separate PR? 🙏🏽

packages/insomnia-importers/src/importers/openapi3.js
```diff
@@ -33,103 +33,13 @@ const SUPPORTED_HTTP_AUTH_SCHEMES = [HTTP_AUTH_SCHEME.BASIC, HTTP_AUTH_SCHEME.BE
 const VARIABLE_SEARCH_VALUE = /{([^}]+)}/g;
```
develohpanda: Interesting changes in this file; I didn't dig deep but please move to a separate PR with some notes about what has changed.

Again, let me know if I'm missing something, but I don't think the changes in this file are necessary for hot-reloading 😄

packages/insomnia-app/app/ui/components/sidebar/sidebar-children.js
```diff
@@ -130,7 +129,7 @@ class SidebarChildren extends React.PureComponent<Props> {
 
       const requestGroup = child.doc;
 
-      function hasActiveChild(children) {
+      const hasActiveChild = children => {
```
develohpanda: This seems concerning - children is a parameter of this function and also its parent function `_renderChildren`, so there is a disambiguation issue here. I wouldn't make any changes here, but it seems like we should rename this parameter in another PR, or even move `hasActiveChild` out of `_renderChildren`.

develohpanda (changes requested):
First of all, great work! I'm really excited to have this in. I pulled your branch locally and did some QA, and the quality of life improvement from this is huge.

I have a couple of comments, most of them are observations but one or two changes that I think should be removed from this PR and moved to separate PRs.

Also, as a note - the first step was to get `react-hot-reloader` working again. Now we're in position to explore using [React Fast Refresh](https://github.com/facebook/react/issues/16604) (which is expected to replace react-hot-reloader). There exists an [experimental](https://github.com/pmmmwh/react-refresh-webpack-plugin/) plugin for it at the moment so it's not quite ready, but I'm glad we have hot-reloading in the meantime 😎 

I also found another useful post explaining the use of arrow functions using the babel plugin - https://frontarm.com/james-k-nelson/when-to-use-arrow-functions/.

packages/insomnia-app/app/ui/components/viewers/response-viewer.js
MrSnix: I found this [facebook/react - 10810](https://github.com/facebook/react/issues/10810) and seems like you technically can do it, but it is unnecessary and may also hurt performance a little bit. So, would you like to revert this changes?
Let me know about it

packages/insomnia-app/app/ui/containers/app.js
MrSnix: Nice, so I mark this one as resolved

packages/insomnia-app/app/ui/components/viewers/response-headers-viewer.js
MrSnix: Yes, you are right, it's not necessary because it's not binding anything, anyway while i was doing the conversion to arrow function I switched everything because i thought you would want to keep the same function's declaration/expression syntax for every components. 
I mark this as resolved but if you have something to say, please, let me know.

packages/insomnia-app/app/ui/index.js
MrSnix: image

packages/insomnia-app/package.json
```diff
@@ -96,7 +95,6 @@
   "dependencies": {
     "@hot-loader/react-dom": "^16.8.6",
     "@stoplight/spectral": "^4.1.1",
-    "autobind-decorator": "^2.4.0",
```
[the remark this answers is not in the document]
MrSnix: image

packages/insomnia-importers/src/importers/openapi3.js
MrSnix: Yes, it's out of scope from this PR. I apologize.
I think my IDE did it alone because semantically shouldn't be any difference.
I'm reverting this

packages/insomnia-app/app/ui/containers/app.js
MrSnix: As requested, I'm reverting the changes 👍🏻

packages/insomnia-app/app/ui/containers/app.js
MrSnix: **EDIT**: It was not supposed to be changed. I'm reverting this.

packages/insomnia-app/app/ui/components/sidebar/sidebar-children.js
MrSnix: I'm reverting the changes as requested

packages/insomnia-app/webpack/webpack.config.development.babel.js
MrSnix: > I'm assuming this change is because the plugins are now deprecated;

That's correct 👍🏻 

>  Is it necessary or avoidable for hot-reloading?

**I think it is avoidable and should works just fine anyway**, meanwhile I was fixing this I made the changes according to the Webpack official documentation. :P

> If avoidable can it please be moved to a separate PR?

Yes, I can move it to a separate PR, I felt like considering this one should "improve", "enhance" the wb configuration related to react-hot-reload it would be just fine for this PR. 😃 

Anyway, i'm reverting this as requested. :octocat:

packages/insomnia-app/app/ui/components/viewers/response-viewer.js
MrSnix: Seems like you technically can do it, but it is unnecessary and may also hurt performance a little bit. So, would you like to revert all lifecycles functions? Let me know about it.
I want to clarify I switched everything to arrow functions because i thought you would want to keep the same function's declaration/expression syntax for every components.

EDIT: Reverted as requested

plugins/insomnia-plugin-kong-portal/src/deploy-to-portal.js
```diff
@@ -211,11 +211,11 @@ class DeployToPortal extends React.Component<Props, State> {
     this.setState(newState);
   };
 
-  componentDidUpdate = async () => {
+  async componentDidUpdate() {
```
MrSnix, line 214: This one shouldn't have been modified in the first place, it has been reverted as originally it was.

MrSnix:
**Important**:
This PR was initially a proposal to replace the use of autobind which prevented react-hot-reload from working properly. Recently, after subsequent investigations, we realized that using `@babel/plugin-proposal-class-properties` as a substitute is not efficient enough. This is what has been found out:
- The initialization of arrow functions in class properties are **transpiled into the constructor**
- Arrow functions in class properties **won’t be** in the `prototype` and we can’t call them using super.
- Arrow functions in class properties are **much slower than bound functions**, and both are much slower than the usual functions.
- Class methods **are created once** on the prototype meanwhile **class fields are created on each object**: 
each component will have its own copy of each function.

Therefore, I am closing this PR in favor of a new one that will be opened shortly containing the new proposal.

**References**:
- Medium / [Arrow Functions in Class Properties Might Not Be As Great As We Think](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1)
- Dev.to / [Class fields are coming. Here's what that means for React](https://dev.to/ascorbic/class-fields-are-coming-heres-what-that-means-for-react--3a87)
- Medium / [Arrow functions are disrupting React.Components](https://blog.usejournal.com/arrow-functions-are-disrupting-react-components-63662d35f97b)

packages/insomnia-app/app/ui/components/viewers/response-viewer.js
develohpanda: See https://github.com/Kong/insomnia/pull/2963#issuecomment-765565861

develohpanda:
Adding to the above comment: investigating the behavior of arrow functions of the lifecycle methods helped us to realize that the existing `@autobind` class decorator from [`autobind-decorator`](https://www.npmjs.com/package/autobind-decorator) was not discerning between a React component class and a regular class.

React lifecycle methods (which are used by RHL) [should not be bound](https://github.com/cassiozen/React-autobind#similar-projects), and the `@autoBindMethodsForReact` decorator from [`class-autobind-decorator`](https://www.npmjs.com/package/class-autobind-decorator) specifically ignores React lifecycle methods.

PR #2969, opened 2020-12-26
Added fix for parameters not updating immediately on importing from URL

### LINKED ISSUE #2937: Query params imported from URL only show after refresh
**Describe the bug**
"Import from URL" for query params does not update the UI

**To Reproduce**
1. paste `http://localhost:4000/path?param=test` in the url
2. go to query tab
3. click "import from url"
4. Note that the tab shows the count of query params (1) but the keyvalue table does not update
5. Activate a different request in the sidebar
6. Go back to the previous request
7. Note the query params show correctly

**Expected behavior**
Should update the key value pair table automatically after import

**Screenshots**

**OS**
```
Version: Insomnia Core 2020.5.2
Release date: 10/12/2020
OS: Darwin x64 19.6.0
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
```

ARKEOLOGIST:
Should close #2937 when implemented. Clicking "Import from URL" now imports the URL parameters into the editor immediately.


ARKEOLOGIST:
@develohpanda Do have a look if possible.

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
```diff
@@ -378,10 +378,10 @@ class Editor extends PureComponent {
     return (
       <Lazy delay={pairs.length > 20 ? 50 : -1}>
         <ul className={classes}>
-          {pairs.map((pair, i) => (
+          {this.props.pairs.map((pair, i) => (
```
apisaint, line 381: Now pairs which are added to `this.state.pairs` through `_addPair` won't show. A more ideal solution would be on mount take pairs passed as props and add them one-by-one by using `_addPair`. 

This would allow a reversion of lines 381, 384, and 415 - and it aligns better with react / vue principals.

ARKEOLOGIST: I agree, I'll work on a better solution for this.

PR #2986, opened 2021-01-08
Enable always using parens for arrow functions

apisaint:
From discussions on #2963 (apologies for the pings snix 😓)

This adds `"arrowParens": "always",` to the prettier config. 

I've also ran `npm run lint:fix` here. So all changes to files are from formatters.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -5,5 +5,6 @@
     "singleQuote": true,
     "tabWidth": 2,
     "trailingComma": "all",
-    "printWidth": 100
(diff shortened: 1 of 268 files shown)


apisaint:
Looks like it's an npm issue. I can either force push an amend or push an empty commit to re-trigger, but I can't re-trigger them otherwise than those two methods. Please advise.

apisaint:
Closing since the origin PR has changed directions! :)

Issue #2987, opened 2021-01-08
Add the support to import some npm module on the test section


giosueDelgado:
**Is your feature request related to a problem? Please describe.**

I want to use and import some npm module to improve my test suite with some libraries.

**Describe the solution you'd like**

This feature would greatly extend the potential of the test suite.
In this way the test tool could read data from the database, integrate external services, use some npm util library, etc..

Example with mongodb:

```js

const response1 = await insomnia.send();

const MongoClient = require('mongodb').MongoClient;

// get some data to check also the persistence layer is working correctly

expect(response1.status).to.equal(200);

```

**Describe alternatives you've considered**

There are no particular alternatives, this request would greatly increase the value of the test suite by fully verifying the operations.

**Additional context**

This feature allows you to use any library (properly downloaded before) in the desired project.
Each project can have its own "node_modules" folder

giosueDelgado:
Good evening @apisaint,

great for rapid development that was done in response to an observation, albeit different, that I made.

In my opinion, however, one of the directions to strengthen that section is this.
I think it is essential to be "free" for developers to be able to verify the correct execution of the API that is not just a simple check of the response/status.

I would like to be able to integrate other systems, the most basic example is a database (mysql / mongo) in which they allow me to verify the execution or even prepare a context initialization (clear a table for example).

I think this testing functionality is crucial.
If I can help with the development I would love to do it. But I would need an indication where to look to be able to do this.

I saw that you are very active. Do you think it could be useful to added value to this project (insomnia)?

PR #2994, opened 2021-01-09
Add support for setting custom headers and bodies inside unit tests

apisaint:
This PR introduces the ability to add custom headers, override existing headers, and set custom body values from inside a unit test to the default request or passed request. Tests included.

This came about after looking into the feasibility of #2987. Even if I introduce NPM modules for unit tests, they are not very useful without being able to augment the request directly. This also enables chaining requests in unit tests, by taking the value of a previous request and augmenting the next. Authentication is one such capability.

Example:

```js
// Send initial request unmodified to authenticate the user
const authResponse = await insomnia.send();

// Send secondary request with authentication token passed from previous response
const authenticatedResponse = await insomnia.send('req_123', { 
  authentication: response.data.token 
});

// Expect successful response from an authenticated resource
expect(authenticatedResponse.status).to.equal(200);
```

### DIFF
--- a/packages/insomnia-app/app/common/send-request.js
+++ b/packages/insomnia-app/app/common/send-request.js
@@ -1,6 +1,6 @@
 import * as db from './database';
 import { types as modelTypes, stats } from '../models';
-import { send } from '../network/network';
+import { sendRequest } from '../network/network';
 import { getBodyBuffer } from '../models/response';
 
 export async function getSendRequestCallbackMemDb(environmentId, memDB) {
@@ -17,20 +17,21 @@ export async function getSendRequestCallbackMemDb(environmentId, memDB) {
   await db.batchModifyDocs({ upsert: docs, remove: [] });
 
   // Return callback helper to send requests
-  return async function sendRequest(requestId) {
-    return sendAndTransform(requestId, environmentId);
+  return async function sendRequestCallback(requestId, requestSettings) {
+    return sendAndTransform(requestId, environmentId, requestSettings);
   };
 }
 
 export function getSendRequestCallback(environmentId) {
-  return async function sendRequest(requestId) {
+  return async function sendRequestCallback(requestId, requestSettings) {
+    console.log('settings', requestSettings);
     stats.incrementExecutedRequests();
-    return sendAndTransform(requestId, environmentId);
+    return sendAndTransform(requestId, environmentId, requestSettings);
   };
 }
 
-async function sendAndTransform(requestId, environmentId) {
-  const res = await send(requestId, environmentId);
+async function sendAndTransform(requestId, environmentId, requestSettings) {
+  const res = await sendRequest({ requestId, environmentId, requestSettings });
   const headersObj = {};
   for (const h of res.headers || []) {
     const name = h.name || '';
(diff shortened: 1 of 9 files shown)


apisaint:
Missed a case, looks like there is a [`sendRequest`](https://github.com/Kong/insomnia/blob/fa64e6b5e896c2d043499b6b937e1a8f616d5346/packages/insomnia-testing/src/run/insomnia.js#L76) that can be passed in, and from the looks of it, that's what [the application and CLI does](https://github.com/Kong/insomnia/blob/cad2ebc93204176d4fb472d6cb6a948903b288cb/packages/insomnia-app/app/common/send-request.js#L25). 

To do the most minimal implementation without introducing breaking changes and impacting existing code, the most ideal solution is to create another send exclusively for unit tests. There is already one for authentication that allows sending custom network object, but it does not allow specifying the environment identifier [`sendWithSettings`](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/network/network.js#L853).

My proposed solution is a new function `sendRequest` for two reasons:

1. `sendWithSettings` [does not apply plugins and must be built itself (not based on reqId)](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/network/network.js#L888), while `send` [does](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/network/network.js#L956)
2. Avoids touching existing APIs and introducing breaking changes

In the future it would be nice to use this function as a way to unify the different methods of sending as it has a more flexible argument structure allow for easier API changes (such as unifying into one that can be passed to both application and plugins):

```js
sendRequest(context: {
    requestId: string,
    environmentId?: string,
    requestSettings?: Object,
    extraInfo?: ExtraRenderInfo,
  }
): Promise<ResponsePatch>
```

vs 

```js
send(
  requestId: string,
  environmentId?: string,
  extraInfo?: ExtraRenderInfo,
): Promise<ResponsePatch>
```

This will change the unit test send structure (as previously displayed above) as well:

```js
insomnia.send('req_123', {
  body: { mimeType: 'application/json', text: JSON.stringify({ ... }) },
  headers: [{ name: authentication, value: response.data.token }] 
})
```

This also exposes some features from the application that are not available to the axios provider, but could slowly be ported or the network layer for insomnia could be exported out and replace axios. Either way it is a step forward and the simplest one.

apisaint:
Working example below, note that you can do far more than expected with the Insomnia application, that doesn't translate into the `axios` implementation at the moment. Here I am changing the method to `POST` to test passing along a body. The original request is a `GET` with no body to ensure clean pass.

Screen Shot 2021-01-09 at 3 19 04 AM

MrStanDu33:
Any update on this thread ? 🤔 

Could be such a nice feature !

jackkav:
Closed as stale.

mugi-luffy:
will this ever be merged? This feature is long-overdue!

Baklap4:
@apisaint any chance you can get this up and running with latest develop?

@jackkav personally closing this as stale whilst no other comment from any contributor/member was given seems kinds of hasty. Any chance we can get this opened and brought up to date?

PR #3001, opened 2021-01-14
Add YAML linting and extend accepted yaml content types

### LINKED ISSUE #2999: Add application/yaml body content support
Right now I can choose json/yaml in body content and get nice syntax highlight.
When I choose json I get Content-type: application/json header
But when I choose yaml I get Content-type: text/yaml header

For working with kubernetes api I need application/yaml content type, but if I manually edit content-type header I loose yaml syntax highlight even though it still says YAML as body type.

It would be nice if I could have body syntax highlight and proper header. 
As a possible solution I would keep body syntax highlight after manual edit of content-type header and keep the current default initial value of text/yaml.

**Additional context**
Proper body with wrong header
highlight

Wrong body with proper header
no highlight

develohpanda:
Closes #2999, closes INS-396

The language mode required by CodeMirror in order to configure it as yaml is `yaml`. We need to normalize `text/yaml`, `application/yaml` to `yaml`, the same way some of the other content types are normalized/translated to something CodeMirror can understand.

Additionally, CodeMirror includes JSON lint and YAML lint addons, however Insomnia was only configured with JSON lint. This PR configures the YAML lint plugin to show syntax errors etc.

## Screenshots
With the default `content-type: text/yaml`
image

With `content-type: application/yaml`
image

image

With a lint error:
image

image

Nunjucks tag - variable:
image

Nunjucks tag - complex variable or action:
Shows an error if using the default tag syntax because of unexpected characters. **This is only a UI error**, it does not prevent the app from continuing or tag from resolving

image

image

image

Works if you encapsulate with speech marks
image

image

## QA:
- [x] Tested Design tab in Designer
- [x] Tested both content types
- [x] Tested template tags

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/base-imports.js
+++ b/packages/insomnia-app/app/ui/components/codemirror/base-imports.js
@@ -39,6 +39,7 @@ import 'codemirror/addon/selection/selection-pointer';
 import 'codemirror/addon/display/placeholder';
 import 'codemirror/addon/lint/lint';
 import 'codemirror/addon/lint/json-lint';
+import 'codemirror/addon/lint/yaml-lint';
 import 'codemirror/addon/lint/lint.css';
 import 'codemirror/keymap/vim';
 import 'codemirror/keymap/emacs';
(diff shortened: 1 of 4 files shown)


sonicyeti (approved):
Sweet!

PR #3007, opened 2021-01-18
Fix designer linting notice table visibility

develohpanda:
As part of #2712, redundant CSS was removed, however that resulted in the notice table no longer showing because the grid definition was deleted. This PR fixes it (using flexbox instead of CSS grid) so that the notice table shows again.

Develop:

This PR:


apisaint:
Just came to comment about this on the original PR. Nice work.

apisaint (approved):
Screen Shot 2021-01-21 at 2 36 45 AM

LGTM!

sonicyeti (approved):
Nice!

DMarby (approved):

PR #3018, opened 2021-01-22
feat: remember the last used import Uri

### LINKED ISSUE #3009: [Feature]: remember last used URL in import
I'm always frustrated when Insomnia does not remember my last used URL importing routes from a URL

Insomnia should remember the last used URL when importing routes. It's unlikely that it'll change often.

None

**Additional context**
image

sfabriece:
Closes #3009


packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -360,6 +360,8 @@ export function importUri(workspaceId: string, uri: string, forceToWorkspace?: F
   return async dispatch => {
     dispatch(loadStart());
 
+    window.localStorage.setItem('insomnia.lastUsedImportUri', uri);
```
develohpanda: Is there a particular reason you chose to set this here? I think it might be better placed closer to where it is consumed, within the `onComplete` callback of the prompt options, in `import-export.js`.

The `lastExportFormat` value is stored in this file, but that's because the modal where it is consumed is also shown from this file.

develohpanda (changes requested):
Thanks for the PR! 🎉

packages/insomnia-app/app/ui/redux/modules/global.js
sfabriece: Made changes as recommended @develohpanda 😄

sfabriece:
Moved changes to import-export.js @develohpanda

develohpanda (approved):
Looks good now. Looks to be working after QA as well. Nice one! 🎉 

Note, this only persists in the import/export modal from _Preferences_. If using Designer, and on the homepage you choose to "Create" a document "From URL", it does not persist (which IMO is correct).

DMarby (approved):

Issue #186, opened 2017-05-11
[Suggestion] F5 and Ctrl + R with same action from button Send 


joubertredrat:
Hi,

I have a small sugestion for Insomnia, today you need to click on "Send" button to make request on API, but I think that is good idea to add F5 and/or Ctrl + R for make requests like "Send" button.

I myself often press F5 on insomnia sometimes, then I remember that insomnia isn't a browser. :joy: 

What you think guys?

gschier:
`ctrl+Enter` will trigger the current request to send. I would like to try to stay away from duplicate hotkeys as long as possible. However, this "muscle memory" may be common enough to warrant it.

gschier:
Closing this for now. Can resurrect later if needed.

PR #3022, opened 2021-01-22
Autobind replacement and react-hot-reload update

### LINKED ISSUE #2468: Remove usages of @autobind to allow for react-hot-reloading to work

MrSnix:
As requested by @develohpanda, I'm opening this PR as a proposal about issue #2468.
This PR is an improved version from the previous one #2963 which has been closed in favor of this one.

**Context**
The auto-bind decorator is preventing react-hot-reload from working properly due to binding not only the class functions but even the React lifecycles functions (which is unnecessary and should be avoided) causing the following error:
image

**The current idea**
We aim to replace [autobind-decorator](https://www.npmjs.com/package/autobind-decorator) with [class-autobind-decorator](https://www.npmjs.com/package/class-autobind-decorator) that introduces `@autoBindMethodsForReact` a convenience decorator that skips auto binding methods from the React Component Spec that do not require binding.

Here is a gif that shows react-hot-reloading working again

Test

As always I'm glad to hear feedback.
Thanks

Closes #2468

### DIFF
--- a/packages/insomnia-app/app/ui/index.js
+++ b/packages/insomnia-app/app/ui/index.js
@@ -1,3 +1,4 @@
+import { hot } from 'react-hot-loader';
 import * as React from 'react';
 import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
@@ -9,9 +10,6 @@ import { init as initPlugins } from '../plugins';
 import './css/index.less';
 import { getAppLongName, isDevelopment } from '../common/constants';
 import { setFont, setTheme } from '../plugins/misc';
-import { AppContainer } from 'react-hot-loader';
-import { DragDropContext } from 'react-dnd';
-import DNDBackend from './dnd-backend';
 import { trackEvent } from '../common/analytics';
 import * as styledComponents from 'styled-components';
 import { initNewOAuthSession } from '../network/o-auth-2/misc';
@@ -37,27 +35,17 @@ document.title = getAppLongName();
   // Create Redux store
   const store = await initStore();
 
-  const context = DragDropContext(DNDBackend);
-  const render = Component => {
-    const DnDComponent = context(Component);
+  const render = App => {
+    const TheHottestApp = hot(module)(App);
     ReactDOM.render(
-      <AppContainer>
-        <Provider store={store}>
-          <DnDComponent />
-        </Provider>
-      </AppContainer>,
+      <Provider store={store}>
+        <TheHottestApp />
+      </Provider>,
       document.getElementById('root'),
     );
   };
 
   render(App);
-
-  // Hot Module Replacement API
-  if (module.hot) {
-    // module.hot.accept('./containers/app', () => {
-    //   render(App);
-    // });
-  }
 })();
 
 // Export some useful things for dev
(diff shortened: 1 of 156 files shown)


develohpanda (approved):
Much cleaner PR, and it seems to be working well too. Nice! 🎉 

[react-fast-refresh](https://github.com/facebook/react/issues/16604) isn't stable for [Webpack](https://github.com/pmmmwh/react-refresh-webpack-plugin/) yet, but that would be the next transition when the time comes.

packages/insomnia-app/app/ui/index.js
```diff
@@ -42,25 +42,17 @@ document.title = getAppLongName();
...
   const render = Component => {
     const DnDComponent = context(Component);
-    ReactDOM.render(
-      <AppContainer>
-        <Provider store={store}>
-          <DnDComponent />
-        </Provider>
-      </AppContainer>,
-      document.getElementById('root'),
+    const app = () => (
```
DMarby: It seems like hot reloading causes the sidebar drag functionality to break, after a hot reload occurs, if you try to drag and drop an item in the sidebar (eg to re-order requests), the following error occurs:

image
image

Interacting with the sidebar at all will also keep throwing errors in the sidebar, but it still seems to render correctly.

DMarby (approved):
Thank you for doing this, seems to work great overall!

I noticed that it seems to break some functionality in the sidebar upon hot reloading, but I'll defer to @develohpanda as to whether that's best to fix before merging this or if it can be addressed later.

I'm not sure if we need to also manually ignore some of the lifecycle methods in the files where we use them @develohpanda?
The library automatically ignores these ones: https://github.com/jmrog/class-autobind-decorator/blob/master/src/lib/ReactComponentSpecMethods.js
However, some of them that we use have changed names, and might need to be manually ignored:
```
UNSAFE_componentWillMount
UNSAFE_componentWillUpdate
UNSAFE_componentWillReceiveProps
```
Not sure if there are also any new lifecycle methods we use that should be ignored.
Regardless, that could probably be addressed at a future date.

MrSnix:
In the last commit, I manually added unsafe react lifecycles as to be ignored by the decorator.
About the sidebar error, as @DMarby correctly stated, it is related to `react-dnd` and `react-hot-reload`. (We are so lucky...)
I found this: [react-dnd #186](https://github.com/react-dnd/react-dnd/issues/186)
I'll try to ship a fix for this if I manage to.

**EDIT**: 
I fixed it. Issue no more... I'll play it a bit to see if it works as expected and I'll push the commit.
@develohpanda If you will check it... because I'm not really familiar with `react-dnd`.

develohpanda (changes requested):
Yes the `UNSAFE_*` lifecycle methods should be ignored too (and aren't currently as a result of the library being slightly outdated).

Nice work adding them @MrSnix 🎉, although (if it works) I think it might be safer to apply a common `autoBindConfig` to each instance of the decorator, rather than selectively depending on usage in a component. If specified per component, as it changes over time it is likely someone will forget to update the autoBind config for it.

```js
// common.js or somewhere
const autoBindConfig = {
  methodsToIgnore: ['UNSAFE_componentWillReceiveProps', ...],
};

// component file
@autoBindMethodsForReact(autoBindConfig)
class Component { ... }
```

develohpanda (approved):
🔥

PR #3024, opened 2021-01-25
Prevent deletes and validate proto files when uploading a directory

develohpanda:
This PR resolves two issues with one buffering fix: fixes INS-431 and fixes INS-433.

**Prevent deletes while uploading:** previously, buffering did not work as expected and would show files in the proto files list as they were loaded, instead of _after_ they were loaded. This meant one could delete the root proto directory, even though the loading process had not completed. This has been resolved in this PR by further controlling how the database chooses to flush it's changes to the UI.

**Validate proto files when uploading:** after uploading a directory, this PR will now try to parse each file that was uploaded to make sure all files are valid. If a single invalid file is found, the root proto directory and all of it's children are deleted.


packages/insomnia-app/app/network/grpc/proto-manager/ingest-proto-directory.js
```diff
@@ -40,31 +40,34 @@ const ingestProtoDirectory = async (
...
+
+  const newDirId = models.protoDirectory.createId();
 
   // Read contents
   const entries = await fs.promises.readdir(dirPath, { withFileTypes: true });
 
   // Loop and read all entries
-  const parsePromises: Array<Promise<boolean>> = entries.map(entry => {
+  let filesFound = false;
+  for (const entry of entries.sort((a, b) => a.name.localeCompare(b.name))) {
```
DMarby: Just curious, but is there any particular reason why you're using localeCompare here?

packages/insomnia-app/app/network/grpc/proto-manager/ingest-proto-directory.js
```diff
@@ -40,31 +40,34 @@ const ingestProtoDirectory = async (
...
     const fullEntryPath = path.resolve(dirPath, entry.name);
-    return entry.isDirectory()
-      ? _parseDir(fullEntryPath, createdProtoDir._id)
-      : _parseFile(fullEntryPath, createdProtoDir._id);
-  });
+    const result = await (entry.isDirectory()
```
DMarby: I think we might need some logic handling errors here, as currently they would propagate straight up, meaning any previously inserted files on the current level will get orphaned and won't be cleaned up by the rollback logic, as they don't belong to any parent that exists until it's created further down?

DMarby (changes requested):
Seems to work great 👍
Just one question regarding potential orphaning of db objects

packages/insomnia-app/app/network/grpc/proto-manager/ingest-proto-directory.js
develohpanda: No particular reason, although the name is user defined and I want it to process each file alphabetically (so that errors are raised in the same order each time the user attempts to import).

packages/insomnia-app/app/network/grpc/proto-manager/ingest-proto-directory.js
```diff
@@ -5,66 +5,93 @@ import type { ProtoDirectory } from '../../../models/proto-directory';
...
-  return Boolean(result);
+type IngestResult = {
+  createdDir?: ProtoDirectory | null,
+  createdIds: Array<string>,
+  error?: Error,
 };
 
-const _parseFile = async (entryPath: string, dirParentId: string): Promise<boolean> => {
-  const extension = path.extname(entryPath);
+class ProtoDirectoryLoader {
```
develohpanda, line 14: This diff isn't very nice to look at. 😞 

I decided to move this logic into a class, to avoid passing a "createdIds" array between every function, or saving it globally in the file, because it should be unique per each "ingest process". Let me know if you have any thoughts around this @DMarby.

packages/insomnia-app/app/network/grpc/proto-manager/ingest-proto-directory.js
develohpanda: Added a breaking test, and updated this logic.

packages/insomnia-app/app/network/grpc/proto-manager/ingest-proto-directory.js
DMarby: Sounds like a reasonable approach

DMarby (approved):

PR #3028, opened 2021-01-26
Remove legacy sync

DMarby:
Removes legacy sync, making version control (formerly known as beta) sync enabled by default.

INS-244


develohpanda (approved):
Code looks good! +8 -2,472 is a dream change-set 😆 

Question: what happens to the workspaces that were previously available to my account through legacy sync after this deprecation?

image

PR #3030, opened 2021-01-27
Generate the bundled ssl trust store at compile time

### LINKED ISSUE #2941: Outdated 2017-09-20.pem CAfile?
Hello!

Today we got a new certificate for our endpoints, and sending request from Chrome works just fine. When we tried to send a request from Insomnia the TLS handshake failed with the following trace:

```
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /tmp/insomnia_2020.5.2/2017-09-20.pem
*   CApath: none
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (OUT), TLS alert, unknown CA (560):
* SSL certificate problem: unable to get local issuer certificate
* Closing connection 7
```

This is the certificate shown in the Chrome certificate dialog:

image

One thing that seems odd is the use of the 2017-09-20.pem CAfile, which looks like very outdated.

If this is indeed the issue, any plans to update this CAfile?

DMarby:
This PR replaces the previously hard-coded trust store with one generated from the NodeJS trust store during compile time.
This ensures that we ship a static trust store without needing to manually keep it up to date. 

We mainly do this during compile time instead of during runtime to avoid tying our trust store to the major version of Electron we use, as that is when the NodeJS version is bumped for Electron, as it is generally a smaller endeavour to bump the minor version of our NodeJS toolchain needed, and that is more likely to happen regularly regardless, then to upgrade Electron.
This also allows us to use a remote source for our bundled trust store such as https://curl.haxx.se/docs/caextract.html in the future if needed.

I've also removed the special-casing for macOS for consistency (and as the macOS store previously used may be outdated in some cases), and instead opt to always use the bundled trust store.

Fixes #2941
Fixes INS-319


packages/insomnia-app/app/network/network.js
```diff
@@ -417,24 +417,21 @@ export async function _actuallySend(
...
-          fs.statSync(fullCAPath);
-        } catch (err) {
-          // Doesn't exist yet, so write it
-          mkdirp.sync(baseCAPath);
-          fs.writeFileSync(fullCAPath, CACerts.blob);
-          console.log('[net] Set CA to', fullCAPath);
-        }
+      // Setup CA Root Certificates
+      const baseCAPath = getTempDir();
+      const fullCAPath = pathJoin(baseCAPath, 'ca-certs.pem');
```
DMarby, line 422: There's no need for a unique filename here, as the `getTempDir()` path contains the app version, and the trust store will never change for an existing version (other then if we bump the toolchain version during development, but that's fine)

packages/insomnia-app/send-request/webpack.config.babel.js
```diff
@@ -30,6 +30,14 @@ module.exports = {
           loader: 'babel-loader',
         },
       },
+      {
+        test: require.resolve('../app/network/ca-certs.js'),
```
DMarby, line 34: This is needed so that `insomnia-send-request` correctly gets the bundle, for `inso`.

develohpanda (approved):
🎉

PR #3031, opened 2021-01-28
Add migration for workspace scope and apply defaults

develohpanda:
What it says on the tin, further details in the linked issue.

Closes INS-406

### DIFF
--- a/packages/insomnia-app/app/models/workspace.js
+++ b/packages/insomnia-app/app/models/workspace.js
@@ -2,8 +2,7 @@
 import type { BaseModel } from './index';
 import * as models from './index';
 import * as db from '../common/database';
-import { getAppId, getAppName } from '../common/constants';
-import { APP_ID_DESIGNER } from '../../config';
+import { getAppName } from '../common/constants';
 
 export const name = 'Workspace';
 export const type = 'Workspace';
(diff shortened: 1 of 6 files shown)


DMarby (changes requested):
Changes look good, I think there's a few more places where we create workspaces that we should specify the scope for as well however?

For when removing a workspace from the workspace dropdown:
https://github.com/Kong/insomnia/blob/68949619fecbee12e128eeaf1f77bc12a23ac2d1/packages/insomnia-app/app/ui/components/wrapper.js#L433

And possibly when importing here, but I guess this is kind of moot if we're letting the user decide what kind of workspace to import soon anyway:
https://github.com/Kong/insomnia/blob/1fe337dd2dc353857219f258bb9a739f6e110a3e/packages/insomnia-app/app/common/import.js#L154

packages/insomnia-app/app/models/workspace.js
```diff
@@ -105,3 +106,23 @@ async function _migrateEnsureName(workspace: Workspace): Promise<Workspace> {
...
+async function _migrateScope(workspace: Workspace): Workspace {
+  // Translate the old value
+  type OldScopeTypes = 'spec' | 'debug' | null;
+  switch ((workspace.scope: OldScopeTypes)) {
+    case 'spec': {
+      workspace.scope = 'designer';
+      break;
+    }
+    case 'debug':
+    case null:
```
DMarby, line 126: *Should* never happen, but it might be worth adding a `default:` case here too just as a defensive measure?

develohpanda:
> I think there's a few more places where we create workspaces that we should specify the scope for as well however?

These are all catered for by the default scope applied here: https://github.com/Kong/insomnia/blob/590286d628366b5f62a87c9c43ecfe1b8c999993/packages/insomnia-app/app/models/workspace.js#L21-L27

DMarby (approved):

PR #3032, opened 2021-01-28
Request Pane now displays URL parameters when "Import URL" is clicked

### LINKED ISSUE #2937: Query params imported from URL only show after refresh

ARKEOLOGIST:
Closes #2937 
Request pane should display URL parameters immediately when "Import URL" is clicked.


ARKEOLOGIST:
@develohpanda Can you take a look at this?

develohpanda:
Thanks for the PR @ARKEOLOGIST and sorry about the delay in reviews (on this PR and your previous PR) 🤗 a lot on my plate atm but I'll get to it in the coming week for sure!

ARKEOLOGIST:
@develohpanda Any progress on the review?

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
```diff
@@ -351,6 +351,28 @@ class Editor extends PureComponent {
...
     this._updateFocus();
+    const pairs = [...this.props.pairs];
+    try {
+      const index = [];
+      for (const pair of pairs) {
+        if (this.props.maxPairs !== 1 && !pair.id) {
+          pair.id = generateId('pair');
+          index.push(pair);
+        }
+      }
```
develohpanda: This seems like it's inspired from the constructor here: https://github.com/Kong/insomnia/blob/0779b0b8c435dd861aad70b32a0f0793d213b4f1/packages/insomnia-app/app/ui/components/key-value-editor/editor.js#L32-L38 I'd favor extracting this logic into another method on this class and reusing it, for example `_ensureIds`.

You also shouldn't need `pairs` and `index` as separate arrays. Just update the source array (`pairs`, not `this.props.pairs`), like is done in the constructor snippet above.

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
```diff
@@ -351,6 +351,28 @@ class Editor extends PureComponent {
...
+          prevPairs.push(pair);
+        }
+        console.log(prevPairs);
+        this.setState({
+          pairs: prevPairs,
+        });
+      }
+    } catch (e) {
+      console.log(e);
+    }
```
develohpanda: Is there a particular reason you chose to wrap this logic in a `try-catch`? Seems unnecessary.

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
```diff
@@ -351,6 +351,28 @@ class Editor extends PureComponent {
...
+          pair.id = generateId('pair');
+          index.push(pair);
+        }
+      }
+      if (index.length !== 0) {
+        const prevPairs = [...this.state.pairs];
+        for (const pair of index) {
+          prevPairs.push(pair);
+        }
+        console.log(prevPairs);
```
develohpanda: I think we can avoid this `console.log` statement 😄

develohpanda (changes requested):
Thanks for the PR @ARKEOLOGIST and apologies for the delays! This looks to be working well 🎉 just a couple of comments worth addressing, please.

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
ARKEOLOGIST: New commits address this.

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
ARKEOLOGIST: Please have a look, I have removed it.

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
ARKEOLOGIST: Yep, you're right, I didn't need it.

ARKEOLOGIST:
@develohpanda Please have a look at the changes if possible!

packages/insomnia-app/app/ui/components/key-value-editor/editor.js
develohpanda: I think there's a slight misunderstanding here, let me clarify. 😄 You'll notice that the snippet in the comment above is nearly identical to a part of the logic you introduce. It appears to be logic that should be common across all usages, instead of duplicating it I suggest extracting that particular snippet into a separate method and consuming it in both the constructor and your bug fix.

develohpanda:
Looks like there's an issue when running through the reproduction steps in the original issue. Note: this was working before you made the changes in [`fa7ba9c` (#3032)](https://github.com/Kong/insomnia/pull/3032/commits/fa7ba9cc6912a09d2ff4e0c9d94370dcf5b0b055).

ARKEOLOGIST:
@develohpanda Reverted the changes. They should work now! Changes have been made to the constructor too so that code duplication isn't a thing.

ARKEOLOGIST:
Any updates on this @develohpanda? I've done the changes as asked.

develohpanda:
Hi @ARKEOLOGIST, my apologies for the delays, again. I am fairly bogged down right now with a release coming up and a lot of activity, so I really appreciate your patience with reviews 🙏🏽 rest assured your PR will not be ignored and we will get to it in due time. Thank you for the time you have put into it! 🤗

ARKEOLOGIST:
@develohpanda Sorry for asking again, but did you have the time to look at this? I have added the changes that you wanted.

develohpanda (commented):
Hi @ARKEOLOGIST, it seems this still doesn't work as intended, see gif below where the key-value pairs are missing from the UI at the end.

ARKEOLOGIST:
@develohpanda The issue seems to arise from the fact that that the parameter extraction seems to run again after you come back to the request. Since the parameters already have been removed from the URL - you see no parameters. I'll try and see if I can shift my changes higher up, so that this problem can be averted.

develohpanda:
It happens when you make an edit to the URL without navigating away as well.



This is related to the logic changed within `key-value-editor.js` rather than an issue upstream. When looking at how the `key-value-editor.js` component is behaving, an important question to ask which would guide a solution is the following.

What _should_ happen to `this.state.pairs` if `this.props.pairs` _changes_, given the state is derived from props in the constructor? Are there more suitable lifecycle events to use than `componentDidUpdate`?

I do have a solution but am happy to not give a direct answer if you'd like to do some sleuthing as this is a nice exercise/problem to debug. 🙂

ARKEOLOGIST:
@develohpanda I was stupid. Thanks for giving me the opportunity to figure this out on my own - I'll push my changes shortly and probably record a video as proof of my changes.

develohpanda:
No worries 🤗 take your time

ARKEOLOGIST:
There, it should work now.

jackkav:
Closed as stale.

PR #3042, opened 2021-02-03
Set a minimum value for fontSize setting

### LINKED ISSUE #3041: Insomnia Preference Interface Font Size selector can go below the min value
When using Insomnia on a high resolution display (1440p in this case) the font size defaults may be insufficient. When changing these settings within the options menu both the `Interface Font Size (px)` and `Editor Font Size (px)` values can be set below the minimum threshold of 8 px. While this isn't an issue for the `Editor Font Size (px)` and can be easily fixed, setting this value incorrectly for the `Interface Font Size (px)` such as backspacing from 13 to 1 to 18 can result in the following behavior:

https://user-images.githubusercontent.com/1396878/106685202-c7f4ee80-657c-11eb-9c6a-252b0544ff2c.mp4

**To Reproduce**
1. Go to `Application`
2. Click on `Preferences`
3. Change `Interface Font Size (px)` to a value lower than 8 using keyboard input
4. See the rendering problem that occurs

**Expected behavior**
I expect the interface to be no smaller than the 8 px minimum specified.

**Screenshots**
Video provided above.

 - OS: Windows 10
 - Installation Method: Git clone
 - App Version: Latest development commit

gravyboat:
Closes #3041 

Small fix for #3041. This sets the minimum possible `fontSize` value that can be set via the preferences menu to 8 as specified in https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/settings/general.js#L271

This is what the fix looks like:

https://user-images.githubusercontent.com/1396878/106687643-62efc780-6581-11eb-9e19-711903005b84.mp4

When coming to this solution I considered a couple of options that included things such as creating a new `fontMinimum` settings value that was passed through, or creating a new `BaseSettings` value but decided against it due to the complexity that would add without a lot of gain. There was already precedent set for this sort ternary check, though I'm not super happy about it being a static value though a new const could be declared in `setFont`. I didn't pass through any sort of `min` variable because I was concerned about how `_handleFontSizeChange` would be impacted if I changed the type of `el` from `SyntheticEvent<HTMLInputElement>`.

This also doesn't fix `editorFontSize` (noted it was also a small problem in the original issue) as the way that is set is different and there is no `editor-font-size` property to set so it doesn't work any way.


develohpanda (approved):
Thanks for the PR, and nice catch. Just pulled this PR down to test and it seems to be working, although I have an observation! 😄 

In the current state, I can have 1px saved in settings (and visible under preferences), but the UI renders at 8px. While this looks odd, and IMO is not a deal-breaker for this use case, I think it might be worth a short investigation.

I wonder if we can use a slightly different approach where the font size is not persisted in settings unless it is an acceptable value? This might be a little involved as it seems the min/max attributes only apply when using arrows on the input. Perhaps we could use `onBlur` instead of `onChange`? 🤔 In any case I don't think there's a clear-cut solution for it.

> I'm not super happy about it being a static value though a new const could be declared in setFont

This can be placed in `constants.js`, which already contains constants of this nature 👍🏽 
https://github.com/Kong/insomnia/blob/26fb78ab99f3d16858fd4afe76d9ca3ba8fe7d92/packages/insomnia-app/app/common/constants.js#L140-L151

develohpanda (changes requested):
Oops my bad, I hit approve although there is an item worth investigating! Once you have a chance to look at it, we can re-review 🤗

gravyboat:
> I wonder if we can use a slightly different approach where the font size is not persisted in settings unless it is an acceptable value?

This would definitely solve part of the issue with the visual aspect. There would still be the problem that you can close the settings with an invalid value which is then silently reset and that seems like bad UX. What if we disallow closing the preference or swapping to a different tab while an invalid minimum value is set? Though that would be a bit of an undertaking to ensure it's handled properly everywhere...

> Perhaps we could use `onBlur` instead of `onChange`?

I think that's a good idea and it's a big improvement. We can't use that for the editor text size but since it's in the background it doesn't matter as much and isn't nearly as jarring. Good idea!

> This can be placed in `constants.js`, which already contains constants of this nature

Okay great I'll move it there. I was also thinking we could bump the font max to `24` instead of `20` for accessibility issues in case a user has a visual impairment. If there is some sort of "setting reset" implemented I don't think it should apply to the upper limit. It's a small edge case but there may be users that need a font size larger than `24`, but it still makes sense to keep it as the soft limit to avoid breaking the UI.

Edit: I'm thinking of displaying an error/alert when a value is set incorrectly for the font size. I'll work on it a bit.

develohpanda:
> What if we disallow closing the preference or swapping to a different tab while an invalid minimum value is set? Though that would be a bit of an undertaking to ensure it's handled properly everywhere...

I agree this will be pretty tricky and probably not worth the effort.

See how far you get with a short investigation; @nijikokun could you please provide some product input on the above comments and whether the following statement is a blocker? I encourage you to pull down this PR and run it to see how it behaves.

> In the current state, I can have 1px saved in settings (and visible under preferences), but the UI renders at 8px. While this looks odd, and IMO is not a deal-breaker for this use case, I think it might be worth a short investigation.

gravyboat:
Added the constants as suggested and a potential option where a `showAlert` dialog is shown to notify the user the value cannot be set below 8. It does reset the value on the back end but the text box in the UI doesn't update currently.

Demo:

https://user-images.githubusercontent.com/1396878/107327688-5f1ce300-6a62-11eb-8dd3-bbfe988d3056.mp4

gravyboat:
Moved things to consolidate and I noticed rendering was a bit slow where it was. I'm not quite sure of the best way to force the update in the `input` box yet. I thought about maybe tracking the state like `fonts` and `fontsMono`, but the problem is `renderTextSettings` since it's used everywhere. Maybe tracking the state then doing something like this inside `renderTextSetting` would be an option (this doesn't update the text box still):

```
    return (
        <label>
          {label}
          {help && <HelpTooltip className="space-left">{help}</HelpTooltip>}
          {name === 'fontSize' ? (
            <input
              type={props.type || 'text'}
              name={name}
              defaultValue={settings[name]}
              value={this.state.interfaceFontSize}
              {...props}
              onChange={props.onChange || this._handleUpdateSetting}
            />
          ) : (
            <input
              type={props.type || 'text'}
              name={name}
              defaultValue={settings[name]}
              {...props}
              onChange={props.onChange || this._handleUpdateSetting}
            />
          )}
        </label>
      </div>
    );
```

This is pretty messy though...

nijikokun:
> There would still be the problem that you can close the settings with an invalid value which is then silently reset and that seems like bad UX

I agree with this, resetting is not great, a common pattern for these issues is when the user sets a value below the threshold it snaps to the threshold value along with a tooltip that describes the allowed minimum / maximum value. Additionally, any value lower or higher than the threshold also snaps to the threshold value as well.

develohpanda:
Nice investigations! I agree the ternary inside `renderTextSettings` is not ideal because it's specific to the `fontSize` number input. The alert also is functional but it seems a little intrusive.

Regardless of what UX approach is used (as per Niji's response above), I think this is achievable as a general pattern for number entries using something like the following, inside `_handleUpdateSetting`. There's no need to modify the method signature because you already have access to the min/max attributes on the `currentTarget`

```js
const el = e.currentTarget;
if (el.type === 'number') {
  value = parseInt(el.value, 10) || 0;
  const min = parseInt(el.min, 10);
  const max = parseInt(el.max, 10);

  const lessThanMax = Number.isNaN(max) || value <= max;
  const moreThanMin = Number.isNaN(min) || value >= min;
  const between = lessThanMax && moreThanMin;

  // If not between, exit early
  // or snap to min
  // or snap to max
  if (!lessThanMax) {
    value = max;
  }
}

this.props.updateSetting(el.name, value);
```

gravyboat:
> along with a tooltip that describes the allowed minimum / maximum value.

@nijikokun What would the tooltip look like to you? Right now it gets the orangish outline and then on hover over says what the actual minimum would be. I looked through some of the existing tooltips, I guess maybe show a modified tooltip that actually has the text in red or something? I'm open to ideas since it seems like the alert is understandably seen as being a bit aggressive.

@develohpanda Something like this? If there's a `!between` check that returns early there will need to be some value setting in there I think.
```
  async _handleUpdateSetting(e: SyntheticEvent<HTMLInputElement>): Promise<Settings> {
    const el = e.currentTarget;
    let value = el.type === 'checkbox' ? el.checked : el.value;

    if (el.type === 'number') {
      value = parseInt(value, 10) || 0;
      const min = parseInt(el.min, 10);
      const max = parseInt(el.max, 10);

      const lessThanMax = Number.isNaN(max) || value <= max;
      const moreThanMin = Number.isNaN(min) || value >= min;
      const between = lessThanMax && moreThanMin;

      if (!between) {
        // snap code here
        return;
      }

      if (!lessThanMax) {
        value = max;
      }
    }

    if (el.value === '__NULL__') {
      value = null;
    }

    return this.props.updateSetting(el.name, value);
  }
  ```
I like the idea of moving it in to `_handleUpdateSetting` 👏 The thing I'm not sure about with this is that it still doesn't handle the rendering of the actual component and "snapping" it back to the appropriate value visually. Maybe this is just my lack of experience with class based instead of functional components, but doesn't the state now need to be tracked to properly rerender the component? I guess the only option is to refresh the entire `General` component?

**Comment by gravyboat** [comment]
Quick update to implement the `_handleUpdateSetting` logic, as well as ensuring that the invalid setting is never allowed to be saved and reverts back. The video doesn't capture the hover over when there's an invalid value:

https://user-images.githubusercontent.com/1396878/107584839-0a887d80-6bb2-11eb-89c1-97a934c296ab.mp4

Screenshot of the existing hover over tooltip:

error_tooltip

File: packages/insomnia-app/app/ui/components/settings/general.js [hunk 1]
```diff
@@ -74,11 +78,21 @@ class General extends React.PureComponent<Props, State> {
...
+      const min = parseInt(el.min, 10);
+      const max = parseInt(el.max, 10);
+
+      const lessThanMax = Number.isNaN(max) || value <= max;
+      const moreThanMin = Number.isNaN(min) || value >= min;
+      const between = lessThanMax && moreThanMin;
+
+      if (!between) {
+        value = !lessThanMax ? max : min;
+      }
```
develohpanda: This condition might be more understandable with a couple of minor changes to the variables. 😄 

Something like:
```js
if (moreThanMax) {
	value = max;
} else if (lessThanMin) {
	value = min;
}
```

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -268,9 +283,9 @@ class General extends React.PureComponent<Props, State> {
             </label>
           </div>
           {this.renderNumberSetting('Interface Font Size (px)', 'fontSize', '', {
-            min: 8,
-            max: 20,
-            onChange: this._handleFontSizeChange,
+            min: MIN_INTERFACE_FONT_SIZE,
+            max: MAX_INTERFACE_FONT_SIZE,
+            onBlur: this._handleFontSizeChange,
```
develohpanda, line 301: Looks like this can now work properly with `onChange` but I'd say it's better to go with `onBlur`. 👌🏽 

With `onChange`, even though it doesn't go below the 8px minimum, the UI jumping around between changing from `12` to `1` to `18` is still not a nice experience.

develohpanda (approved):
🔥 

> Right now it gets the orangish outline and then on hover over says what the actual minimum would be.

This seems fine to me.

> and "snapping" it back to the appropriate value visually

What you've got in the latest commit seems OK to me to be honest. It shows the above error if out of bounds; it doesn't set the UI font to absurd values like 1px, and it resets to the min/max when reopening the modal.

If however, you wanted the UI to snap back visually, immediately, you could try apply a unique `key` attribute to the element which would tell React to re-mount that element, or yes, force an update on the entire `General` component. If you wanted to explore this route, here's an example of `forceRefreshKey` used in `wrapper.js`

https://github.com/Kong/insomnia/blob/26fb78ab99f3d16858fd4afe76d9ca3ba8fe7d92/packages/insomnia-app/app/ui/components/wrapper.js#L205

gravyboat:
> If however, you wanted the UI to snap back visually, immediately, you could try apply a unique key attribute to the element which would tell React to re-mount that element, or yes, force an update on the entire General component. If you wanted to explore this route, here's an example of forceRefreshKey used in wrapper.js

After further consideration on this I don't think having it automatically snap back is a good idea. That means users never see the out of bounds notification while also being prevented from entering in values which are out of bounds which is confusing UX and I could see it leading to further issues.

DMarby (approved):

PR #3054, opened 2021-02-05
Stop automatically appending .git to git repository URLs

### LINKED ISSUE #2095: Cannot use Azure DevOps git repo in Insomnia Designer
**Describe the bug**
In the designer when try to import a git repo hosted in Azure DevOps, nothing happens. 

**To Reproduce**
1. Open Designer
2. Click on 'Create' -> 'Git Clone'
3. Configure using Azure DevOps git repo
4. Click 'Done' and observe nothing happens.
5. Open Developer Tools
6. Observe error message "HTTPError: HTTP Error: 404 Not Found"

**Expected behavior**
New Workspace to be added based on the repo

 - App: Insomnia Designer
 - OS: macOS 10.14.5
 - Installation Method: Package downloaded from website
 - App Version: 2020.1.3

**Additional context**
I did some troubleshooting and found out that, insomnia appends string ".git" to the provided git url. This works with github. But some services live Azure DevOps doesn't support .git URL and returns 404 when queried.

**Workaround**
As a workaround I renamed Azure Git Repo to end with .git (Ex: backend.git). Then when configuring Insomnia I used the url without the .git portion.

DMarby:
Fixes #2095
Fixes INS-367

Tested by creating a repo on Azure DevOps and successfully interacting with it, before this PR all operations against it fails.

I've added a migration so that existing repositories that didn't have `.git` at the end of the URL gets it appended, so that the new behaviour only applies to new repositories.
I tested this by setting up a new workspace with a GitHub repo and leaving out the `.git` part of the URL, then applying this PR and verifying that it could still sync.


packages/insomnia-app/app/models/git-repository.js
```diff
@@ -31,10 +32,27 @@ export function init(): BaseGitRepository {
...
+  return doc;
+}
+
+// Append .git to old git URIs to mimic previous isomorphic-git behaviour
+function _migrateURI(doc: GitRepository): GitRepository {
+  if (doc.uriHasBeenMigrated) {
+    return doc;
+  }
+
+  if (!doc.uri.endsWith('.git')) {
```
DMarby: Reason for doing it this way rather then parsing the URL is that it's how `isomorphic-git` did it, so to ensure that we mimic the behaviour we do it the same way: https://github.com/isomorphic-git/isomorphic-git/blob/9176a3a7f0a28b059b0aae135ce5f325f5d737ec/src/managers/GitRemoteHTTP.js#L53

packages/insomnia-app/app/models/git-repository.js
```diff
@@ -12,6 +12,7 @@ type BaseGitRepository = {
     name: string,
     email: string,
   },
+  uriHasBeenMigrated: boolean,
```
DMarby: Not super nice to need to add a new property for this, but not sure of a better approach since we only want to run this migration for old entries.

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
```diff
@@ -185,7 +185,10 @@ class GitSyncDropdown extends React.PureComponent<Props, State> {
         if (gitRepository) {
           await models.gitRepository.update(gitRepository, patch);
         } else {
-          const repo = await models.gitRepository.create(patch);
+          const repo = await models.gitRepository.create({
+            ...patch,
+            uriHasBeenMigrated: true,
```
develohpanda: I think patching this flag should be done within the `create` function rather than out here. If I understand correctly, _any_ time the user will create a git repo in future, this should _always_ be set to true.

In this case it seems possible to contain a flag related to model migration to the model itself, which feels safer.

packages/insomnia-app/app/models/__tests__/git-repository.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+    await models.gitRepository.migrate(oldRepoWithSuffix);
+    await models.gitRepository.migrate(oldRepoWithoutSuffix);
+    await models.gitRepository.migrate(newRepoWithSuffix);
+    await models.gitRepository.migrate(newRepoWithoutSuffix);
+
+    expect(oldRepoWithSuffix.uri).toBe('https://github.com/foo/bar.git');
+    expect(oldRepoWithoutSuffix.uri).toBe('https://github.com/foo/bar.git');
+    expect(newRepoWithSuffix.uri).toBe('https://github.com/foo/bar.git');
+    expect(newRepoWithoutSuffix.uri).toBe('https://github.com/foo/bar');
```
develohpanda: Also want to test that for all instances, the flag `uriHasBeenMigrated` is set (or retained) as `true`.

packages/insomnia-app/app/models/git-repository.js
develohpanda: I agree, not nice to add a one-time entry, but it does seems necessary. There exist database repair scripts which run once on launch (as opposed to model migrations which can run more frequently).

You'd still need a flag somewhere but perhaps `_repairDatabase` is a better place to migrate? Not really fussed, but since this is a one-time migration it might be better in the repair 🤔 

https://github.com/Kong/insomnia/blob/b135a2c637768b5a21ae68a7e39e4ca8e5785ff0/packages/insomnia-app/app/common/database.js#L689-L699

develohpanda (changes requested):
Nice! Seems to be working, I tested with a git repo without the suffix, applied this PR and saw the suffix had been added as expected. 🎉 

A couple of comments to do with how and where `uriHasBeenMigrated` is set and extending the unit test which I think are worth addressing. 

Additionally, I think we need some UI prompts to make sure users don't see 404s if they don't suffix with `.git` manually.

For example, if you try to create a fresh clone of `https://github.com/develohpanda/designer` (note, no suffix), you get 404s while previously, it worked. Perhaps a tooltip or active warning which explains the requirement to suffix with `.git` _except_ in the currently known case with Azure DevOps.

nijikokun:
> Additionally, I think we need some UI prompts to make sure users don't see 404s if they don't suffix with `.git` manually.

Let's not display assumptions about git systems, we know this is the case for this system, but not others and we might ultimately give the user bad advice while attempting to act in good faith. 

If it's a requirement, the input should be sufficient enough, anything further and perhaps in later iterations after we see instances of issue we can add a link to some documentation about specific systems external to the application where we can maintain without updating the application itself.

develohpanda:
> Let's not display assumptions about git systems, we know this is the case for this system, but not others and we might ultimately give the user bad advice while attempting to act in good faith.
> 
> If it's a requirement, the input should be sufficient enough, anything further and perhaps in later iterations after we see instances of issue we can add a link to some documentation about specific systems external to the application where we can maintain without updating the application itself.

The input doesn't seem to check whether `.git` is provided, and we don't mention anywhere that `.git` _needs_ to be provided as the suffix. From my testing and the isomorphic-git documentation, the `.git` suffix is a requirement for GitHub (and likely for GitLab). Omitting the `.git` suffix only seems to work for Azure DevOps. Previously it would automatically add the suffix if it was missing; now it needs to be manually added.

This means, that with GitHub for example, what worked before will seemingly stop working for new clones/repositories (a migration was added for existing repositories).

Given this behavior I think we need to say _something_, even if it is a tool-tip with a link to docs next to the label (like the auth token, for example, below)

image

nijikokun:
> don't mention anywhere that `.git` _needs_ to be provided as the suffix

It's here on the input:
image

develohpanda:
Really doesn't feel like a placeholder is sufficient at explaining the requirement that exists for almost all use cases, but I'll leave that up to you - not a blocker for me, just want to make it quite clear to users.

When a user sees a 404 error with no additional error information provided, they tend to look for the tool-tip, not at the placeholder (which will no longer be visible).

packages/insomnia-app/app/models/__tests__/git-repository.test.js
DMarby: Good catch 👍

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.js
DMarby: Agreed, harder to miss setting it with a create function, done 👍

packages/insomnia-app/app/models/git-repository.js
DMarby: I think that sounds like a better place to put it, done.

packages/insomnia-app/app/models/git-repository.js
```diff
@@ -31,15 +32,19 @@ export function init(): BaseGitRepository {
       name: '',
       email: '',
     },
+    uriNeedsMigration: true,
```
develohpanda, line 35: 👍🏽

packages/insomnia-app/app/models/git-repository.js
```diff
@@ -31,15 +32,19 @@ export function init(): BaseGitRepository {
...
 
-export function migrate<T>(doc: T): T {
+export function migrate(doc: GitRepository): GitRepository {
   return doc;
 }
 
 export function create(patch: $Shape<GitRepository> = {}): Promise<GitRepository> {
-  return db.docCreate(type, patch);
+  return db.docCreate(type, {
+    uriNeedsMigration: false,
```
develohpanda, line 45: 🔥

develohpanda (approved):
Looking good! The discussion above about a tool-tip vs placeholder is not a blocker for this PR 😄

swone1990:
Hey not sure but i tought the issue with the .git suffix is solved.
I still can not checkout AzureDevops Repositories, because the suffix got apppended.

Did I miss something in this Issue?

PR #3055, opened 2021-02-06
Insomnia v4 support for CLI

### LINKED ISSUE #2666: Add insomnia V4 import support to Insomnia CLI
**Unable to use insomnia designer export with CLI**
I am planning to use the inso cli to integrate with CI. I don't plan on create a separate git repo and would like to use it along with my existing repo.

**Importing insomnia export in the CLI**
If the cli can import the insomnia_export file and create the .insomnia folder or directly run it , that can simplify using inso for CI

MrSnix:
TL;DR Contributor loses his sanity due to a crash generated by a typo (I presume)

Hi folks, as requested by PR #2666 I'm working on integrating Insomnia v4 support on CLI using a global option to generate an in-memory database that will be available for the lifecycle of the shell itself.

This is an early preview and I'm still working on it :warning: , I opened it because ***I feel like I need help*** 🌊 ⛵ 🌊 with an issue that is raised while delivering the Insomnia DB object to `insomnia-send-request`.

I'm using the following command to run the cli while developing from the dist folder:
`node inso run test --fromFile ./../src/db/__fixtures__/insomnia-v4/Insomnia.json --verbose`

As declared by the spec, I'm using the following Db model:
```javascript
  export type Database = {|
    ApiSpec: Array<ApiSpec>,
    Environment: Array<Environment>,
    Request: Array<BaseModel>,
    RequestGroup: Array<BaseModel>,
    Workspace: Array<Workspace>,
    UnitTestSuite: Array<UnitTestSuite>,
    UnitTest: Array<UnitTest>,
  |};
```
Sadly, when reaching the following routine, even if I'm delivering the requested model, it dies:
```javascript
  // Load lazily when needed, otherwise this require slows down the entire CLI.
  const { getSendRequestCallbackMemDb } = require('insomnia-send-request');
  const sendRequest = await getSendRequestCallbackMemDb(environment._id, db);
```
Console Output
```text
› Data store configured from file at /Users/**/**/insomnia/packages/insomnia-inso/src/db/__fixtures__/insomnia-v4/Insomnia.json                                                     
› Prompt for document or test suite                                                                                                                                                                     
✔ Select a document or unit test suite · Swagger Petstore 1.0.0 - spc_c3cd1d
› Load api specification with identifier spc_c3cd1d from data store                                                                                                                                     
› Found 1.                                                                                                                                                                                              
› Load workspace with identifier wrk_91f4be0b00f24fe3b7b8fa54a4332ffc from data store                                                                                                                   
› Found 1.                                                                                                                                                                                              
› Load base environment for the workspace wrk_91f4be0b00f24fe3b7b8fa54a4332ffc from data store                                                                                                          
› Found 1.                                                                                                                                                                                              
› Prompt for environment                                                                                                                                                                                
✔ Select an environment · OpenAPI env - env_env_ee775e
› Load base environment for the workspace wrk_91f4be0b00f24fe3b7b8fa54a4332ffc from data store                                                                                                          
› Found 1.                                                                                                                                                                                              
› Load sub environment with identifier env_env_ee775e from data store                                                                                                                                   
› Found 1                   

 ERROR  Cannot read property 'find' of undefined                                                                                                                                                        

  at /Users/**/**/insomnia/packages/insomnia-send-request/dist/index.js:376:14
  at new Promise (<anonymous>)
  at Object.module.exports.database.find (/Users/**/**/insomnia/packages/insomnia-send-request/dist/index.js:375:10)
  at Object.module.exports.database.getWhere (/Users/**/**/insomnia/packages/insomnia-send-request/dist/index.js:397:31)
  at Object.module.exports.database.get (/Users/**/**/insomnia/packages/insomnia-send-request/dist/index.js:406:21)
  at module.exports.database.upsert (/Users/**/**/insomnia/packages/insomnia-send-request/dist/index.js:425:38)
  at Object.module.exports.database.batchModifyDocs (/Users/**/**/insomnia/packages/insomnia-send-request/dist/index.js:537:27)
  at processTicksAndRejections (internal/process/task_queues.js:97:5)
  at async getSendRequestCallbackMemDb (/Users/**/**/insomnia/packages/insomnia-send-request/dist/index.js:28951:3)
  at async runInsomniaTests (/Users/**/**/insomnia/packages/insomnia-inso/dist/index.js:597:23)
```
After some tests, I was able to figure out that the following parameter `type` is **undefined** when he reaches this function:
```javascript
const find = database.find = async function (type, query = {}, sort = {
  created: 1
}) {
  if (db._empty) return _send('find', ...arguments);
  return new Promise((resolve, reject) => {
   // Basically, he access db['undefined'].find() and explodes
    db[type].find(query).sort(sort).exec(async (err, rawDocs) => {
      if (err) {
        return reject(err);
      }
```
After further investigations I found out the problem MAY comes from here:
```javascript
const upsert = database.upsert = async function (doc, fromSync = false) {
  if (db._empty) return _send('upsert', ...arguments);
  const existingDoc = await database.get(doc.type, doc._id);

  if (existingDoc) {
    return database.update(doc, fromSync);
  } else {
    return database.insert(doc, fromSync);
  }
};
```
*The function uses `doc.type` but the current declaration uses `doc._type` (with the underscore)...*
I added the following lines of code on dist.js to check if I was right:
```javascript
const upsert = database.upsert = async function (doc, fromSync = false) {
  if (db._empty) return _send('upsert', ...arguments);
  //TODO It was fixed here
  console.log('type: ' + doc.type)
  console.log('_type: ' + doc._type)
  const existingDoc = await database.get(doc.type, doc._id);
  [...]
};
```
Console Output:
```text
type: undefined
_type: api_spec
type: undefined
_type: environment
type: undefined
_type: environment
type: undefined
_type: request
type: undefined
_type: request
type: undefined
_type: request
type: undefined
_type: workspace
type: undefined
_type: unit_test
type: undefined
_type: unit_test_suite
```

So, I wrote this little function just to check if the real fix was renaming to `type`
```javascript
const rename: ApiSpec = (o, newKey, oldKey) => {
  delete Object.assign(o, { [newKey]: o[oldKey] })[oldKey];
  return o;
};
```

Sadly it crashes anyway, at the same spot for the same reasons apparently.

MrSnix, can you summarise? Things got complicated.
Well, yes.

**Summary**:

1. I deliver the DB model (I assume correctly)
2. `database.upsert` uses the `BaseModel` interface which declares `type` but on my object, I have `_type`
3. `database.upsert` call `database.find(type, db)` with `type = undefined`
4. `database.find(type, db)` dies because he tries to access `db['undefined'].find()`
5. I wrote a little function to rename my object instance fields from `_type` to `type`
6. I retry to see if now works...
7. Crash anyway, same reason.

I have no clue what's happening there.

What should I do?
Any idea?

It closes #2666


packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,37 @@
...
+  // Sanity check - do db files exist?
+  if (!fs.existsSync(path)) return null;
+  // Building an empty database
+  const db: Database = emptyDb();
+
+  // Now, reading and parsing
+  const contents = await fs.promises.readFile(path);
+  const object = JSON.parse(contents);
+  // We are supporting only v4 files
+  if (object.__export_format !== 4) return null;
```
develohpanda: Brilliant!

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,37 @@
...
+    db.RequestGroup.push(...extract(raw, 'request_group'));
+    db.Workspace.push(...extract(raw, 'workspace'));
+    db.UnitTestSuite.push(...extract(raw, 'unit_test_suite'));
+    db.UnitTest.push(...extract(raw, 'unit_test'));
+  }
+
+  return db;
+};
+
+const extract: Array<ApiSpec> = (raw, _type) => {
```
develohpanda: The return type should just be `Array<BaseModel>` because it could be any of them.

packages/insomnia-inso/src/cli.js
```diff
@@ -174,6 +174,7 @@ export function go(args?: Array<string>, exitOverride?: boolean): void {
     .option('-a, --appDataDir <dir>', 'set the app data directory')
     .option('--config <path>', 'path to configuration file')
     .option('--verbose', 'show additional logs while running the command')
+    .option('--fromFile <path>', 'generate an in-memory database from an insomnia-v4.json file')
```
develohpanda:
```suggestion
    .option('--fromFile <path>', 'generate an in-memory database from an insomnia V4 export file')
```

develohpanda (commented):
> The function uses `doc.type` but the current declaration uses `doc._type` (with the underscore)...

You're definitely on the right track with this.

Lets work with the `UnitTest` model for now. When **exporting**, [`models.unitTest.type`](https://github.com/Kong/insomnia/blob/3883dc6feb7d2cf225065ebe9fd635b15e2171c2/packages/insomnia-app/app/models/unit-test.js#L6) is translated to [`EXPORT_TYPE_UNIT_TEST`](https://github.com/Kong/insomnia/blob/1fe337dd2dc353857219f258bb9a739f6e110a3e/packages/insomnia-app/app/common/import.js#L33). It is then set to the property [`_type`](https://github.com/Kong/insomnia/blob/1fe337dd2dc353857219f258bb9a739f6e110a3e/packages/insomnia-app/app/common/import.js#L442-L443), and the [`type`](https://github.com/Kong/insomnia/blob/1fe337dd2dc353857219f258bb9a739f6e110a3e/packages/insomnia-app/app/common/import.js#L458-L459) property is removed. (I'm not sure why this happens, but it does).

Therefore, when **importing**, we just need to do the reverse of this. Translate `model._type = EXPORT_TYPE_UNIT_TEST` to `model.type = 'UnitTest'` (this string value is defined in [`models.unitTest.type`](https://github.com/Kong/insomnia/blob/3883dc6feb7d2cf225065ebe9fd635b15e2171c2/packages/insomnia-app/app/models/unit-test.js#L6)).

The import logic within the electron app (`packages/insomnia-app`) does something like that [here](https://github.com/Kong/insomnia/blob/1fe337dd2dc353857219f258bb9a739f6e110a3e/packages/insomnia-app/app/common/import.js#L205) then [here](https://github.com/Kong/insomnia/blob/1fe337dd2dc353857219f258bb9a739f6e110a3e/packages/insomnia-app/app/common/import.js#L240-L245).

The CLI (`packages/insomnia-inso`) cannot import the same logic, but it can also be a lot simpler.

If you do the following two things, you should be up and running again:
1. Create a map to translate between the "export" type (`'unit_test'`) and the "model" type (`'UnitTest'`)
2. Use that map to translate between `_type` and `type` in your `extract` method for each model, and delete the `_type` property

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,37 @@
...
+
+const insomniaAdapter: DbAdapter = async path => {
+  // Sanity check - do db files exist?
+  if (!fs.existsSync(path)) return null;
+  // Building an empty database
+  const db: Database = emptyDb();
+
+  // Now, reading and parsing
+  const contents = await fs.promises.readFile(path);
+  const object = JSON.parse(contents);
```
develohpanda: Ah, you might want to use `YAML.parse` here, instead of `JSON.parse`, because the file can be YAML or JSON.

There is the option of using `insomnia-importers` to do this, but I don't see any benefit in pulling that package _at this stage_.

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,37 @@
...
+
+  if (raw?.length !== 0) {
+    db.ApiSpec.push(...extract(raw, 'api_spec'));
+    db.Environment.push(...extract(raw, 'environment'));
+    db.Request.push(...extract(raw, 'request'));
+    db.RequestGroup.push(...extract(raw, 'request_group'));
+    db.Workspace.push(...extract(raw, 'workspace'));
+    db.UnitTestSuite.push(...extract(raw, 'unit_test_suite'));
+    db.UnitTest.push(...extract(raw, 'unit_test'));
+  }
```
develohpanda: Nice reuse 👏🏽

MrSnix:
I think I'm ready for another review. I don't know if I'm missing something else, please point it out as always. 
I'll provide the new requested changes as quick as possible.

Thanks :D

nijikokun:
Great work, will simplify a lot of workflows 🎉

develohpanda:
@nijikokun we probably need to look at consolidating the `--app-data-dir` and `--from-file` flags into one; potentially keep the `--app-data-dir` flag but deprecate it in favor of `--app-data-src` which can be a directory (install directory) _or_ a file (like the export file).

Do you have any thoughts about it at this stage, or we roll with introducing a new `--from-file` option?

nijikokun:
> @nijikokun we probably need to look at consolidating the `--app-data-dir` and `--from-file` flags into one; potentially keep the `--app-data-dir` flag but deprecate it in favor of `--app-data-src` which can be a directory (install directory) _or_ a file (like the export file).
> 
> Do you have any thoughts about it at this stage, or we roll with introducing a new `--from-file` option?

There are a few future concerns I have that I'd like to walk through before making a breaking change to existing flags. Given that we want to support running multiple collections tests at once (entire space) for example what would be the best option.

Given the following options:

- Individual `v4.json` file.
- Directory of multiple `v4.json` files
- Insomnia application data directory
- Git Sync `.insomnia` directory

I will note, I do like the succinctness of `--from-file`. Looking at other tools they use very short and very direct terms such as `--file` and `--dir`, as close to these as possible is ideal. Looking at mocha, since I'm on it's page, they use `--file <file|directory|glob>` so perhaps a single option isn't as confusing. 

So something like `--src <file|directory|glob>`?

Thoughts?

develohpanda:
Yep that's what I'm thinking too. We can definitely work with a single option and try multiple db adapters (in priority) until one works, and I'm liking the succinctness of `--src`.

@MrSnix I'd say we replace `--from-file <file>` with `--src <file|directory>`. You might need to use [`fs.existsSync(src) && fs.lstatSync(src).isDirectory()`](https://stackoverflow.com/questions/15630770/node-js-check-if-path-is-file-or-directory) to determine whether we're working with a file or directory, and load accordingly.

In addition we can still keep `--app-data-dir` around for backwards compatibility for now, but mark as deprecated in documentation. And anywhere we use `--app-data-dir` currently in the codebase, we also handle the `--src` flag.

We can introduce glob support in a future PR since that'll be a little more involved.

How does that sound?

MrSnix:
Ok, ready for another review!

**Changes**:
- Rewrote a lot of tests
- Updated the way loadDb() works.
It should be more **clear and clean** (at least, I hope so).
I spent a lot of time figuring out how to make it short and simple, 
- Introduced some more checks on loadDb()
- Changed `--from-file` to `--src`
- Updated everything that was necessary
- I'm taking the habit to make smaller commits which should be easier to review.

Please, let me know what you think and if I forgot something. (as always :P)

packages/insomnia-inso/src/cli.js
```diff
@@ -174,6 +174,7 @@ export function go(args?: Array<string>, exitOverride?: boolean): void {
     .option('-a, --appDataDir <dir>', 'set the app data directory')
```
develohpanda:
```suggestion
    .option('-a, --appDataDir <dir>', 'set the app data directory (deprecated; use --src instead')
```

I don't think it can be removed altogether otherwise the alias will not be recognized correctly.

packages/insomnia-inso/src/commands/export-specification.js
```diff
@@ -11,9 +11,9 @@ export type ExportSpecificationOptions = GlobalOptions & {
 
 export async function exportSpecification(
   identifier: ?string,
-  { output, workingDir, appDataDir, ci }: ExportSpecificationOptions,
+  { output, workingDir, appDataDir, ci, src }: ExportSpecificationOptions,
 ): Promise<boolean> {
-  const db = await loadDb({ workingDir, appDataDir, filterTypes: ['ApiSpec'] });
+  const db = await loadDb({ workingDir, appDataDir, filterTypes: ['ApiSpec'], src: src });
```
develohpanda:
```suggestion
  const db = await loadDb({ workingDir, appDataDir, filterTypes: ['ApiSpec'], src });
```

Could you please make this change against the other instances as well? I won't comment on each one 🙂 I'm sure there's an ES Lint rule for this

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,78 @@
...
+  UnitTest: 'unit_test',
+};
+
+const insomniaAdapter: DbAdapter = async (path, filterTypes) => {
+  // Sanity check - do db files exist and is it a file?
+  if (!fs.existsSync(path) || fs.lstatSync(path).isDirectory()) return null;
+  // Building an empty database
+  const db: Database = emptyDb();
+  // Building promise
+  return new Promise(async function(resolve, reject) {
```
develohpanda: Is there a particular reason you chose to use this syntax of explicitly creating a Promise? It's not a common pattern in this code-base so preference would be to stick with regular `async-await`.

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+import fs from 'fs';
+import { emptyDb } from '../index';
+import type { Database, DbAdapter } from '../index';
+import type { BaseModel } from '../models/types';
+import YAML from 'yaml';
+
+const keyToType = {
```
develohpanda: I suggest renaming this to be very clear - `modelTypeToExportTypeMap`

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,78 @@
...
+
+    if (raw?.length !== 0) {
+      types.forEach(type => db[type].push(...read(raw, keyToType[type])));
+    }
+
+    resolve(db);
+  });
+};
+
+const read: Array<BaseModel> = (raw, _type) => {
```
develohpanda: Could you please add types for `raw` and `_type`?

Two points to think about to encourage improved variable names:
- `raw` in itself is a fairly vague variable name. Someone looking at this code will need to understand how it's being used to figure out what `raw` represents. I suggest renaming it to `importedModels` or something similar, to be clearer.
- is the string value of `_type` expected to be the base model type (`'ApiSpec'`) or export type (`api_spec`), given we know there is a discrepancy there?

Variable names are very important for self-documenting code, and especially helpful for yourself and those who come after you to look at this code. It was a headache for us to figure out the type differences, lets make it clear for those looking at the code after us 😄

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,78 @@
...
+  // Extract fields of the same types
+  const extracted = extract(raw, _type);
+  // Renaming from ._type to .type
+  const renamed = rename(extracted, _type);
+  // Now replacing with the correct type
+  return replace(renamed, _type);
```
develohpanda: It seems like these three steps can happen in the same iteration. In this implementation, the code is traversing the `raw` array 3x, but it can be done in one. Note, you don't necessarily need to use `.map` and `.filter`; where appropriate you can always revert to a simple for loop and do everything inside the loop body 🤗

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,78 @@
...
+        ),
+      );
+    }
+    const raw = object?.resources;
+
+    // Filter data on type
+    const types = filterTypes?.length ? filterTypes : Object.keys(db);
+
+    if (raw?.length !== 0) {
+      types.forEach(type => db[type].push(...read(raw, keyToType[type])));
```
develohpanda: Is the `keyToType[type]` mapping necessary here, given it _also_ exists in the `replace` method?

packages/insomnia-inso/src/db/index.js
```diff
@@ -42,37 +44,81 @@ export type DbAdapter = (
...
+  const currentSrc = src || appDataDir || workingDir || '.';
+  const exists = currentSrc && fs.existsSync(currentSrc);
+  const isDirectory = exists && fs.lstatSync(currentSrc).isDirectory();
+  const pathname = `\`${path.resolve(currentSrc)}\``;
+
+  if (!exists) logger.debug(`Data source does not exists at ${pathname}`);
+
+  // Directory-based adaptors
+  if (isDirectory && workingDir) db = await asGitAdapter(workingDir, filterTypes);
+  if (isDirectory && appDataDir) db = await asNeDbAdapter(appDataDir, filterTypes);
```
develohpanda: `appDataDir` is being deprecated in favor of `src`; so, shouldn't the `NeDbAdapter` take in `src || appDataDir`?

packages/insomnia-inso/src/db/index.js
```diff
@@ -42,37 +44,81 @@ export type DbAdapter = (
...
+
+  const currentSrc = src || appDataDir || workingDir || '.';
+  const exists = currentSrc && fs.existsSync(currentSrc);
+  const isDirectory = exists && fs.lstatSync(currentSrc).isDirectory();
+  const pathname = `\`${path.resolve(currentSrc)}\``;
+
+  if (!exists) logger.debug(`Data source does not exists at ${pathname}`);
+
+  // Directory-based adaptors
+  if (isDirectory && workingDir) db = await asGitAdapter(workingDir, filterTypes);
```
develohpanda: Previously the git adapter only looked at `workingDir || '.'`.

But we're now providing a universal option to provide the data source via `src`, so it would make sense for the git adapter to support it via `src || workingDir || '.'`.

It will exit early if data is not found in the expected format, so that's okay.

https://github.com/Kong/insomnia/blob/b99228d08030a7b6b90ffc4afb4fab1e0564b4f8/packages/insomnia-inso/src/db/adapters/git-adapter.js#L12

packages/insomnia-inso/src/db/index.js
```diff
@@ -42,37 +44,81 @@ export type DbAdapter = (
...
+  if (isDirectory && appDataDir) db = await asNeDbAdapter(appDataDir, filterTypes);
+  // File-based adaptors
+  if (exists && !isDirectory) db = await asInsomniaAdapter(src, filterTypes);
```
develohpanda: Note, the data loading priority should be
- git adapter
- insomnia export adapter
- nedb adapter
- blank

If one of the adapters matches (doesn't return null), **we should not attempt to load another adapter.**

packages/insomnia-inso/src/db/index.js
```diff
@@ -42,37 +44,81 @@ export type DbAdapter = (
...
-  }
-
-  // return empty db
-  if (!db) {
+  // In-memory database instance
+  let db = {
+    adapter: null,
+    descriptor: null,
+    data: null,
+  };
```
develohpanda: Please add flow types for this.

packages/insomnia-inso/src/db/index.js
```diff
@@ -42,37 +44,81 @@ export type DbAdapter = (
...
+
+  /* In order to support both options,
+   * we use this constant (currentSrc) that
+   * gives precedence on the newest option (src) */
+
+  const currentSrc = src || appDataDir || workingDir || '.';
+  const exists = currentSrc && fs.existsSync(currentSrc);
+  const isDirectory = exists && fs.lstatSync(currentSrc).isDirectory();
+  const pathname = `\`${path.resolve(currentSrc)}\``;
```
develohpanda: Could you please explain why this is wrapped in backticks?

packages/insomnia-inso/src/db/index.js
```diff
@@ -42,37 +44,81 @@ export type DbAdapter = (
...
+
+  if (!exists) logger.debug(`Data source does not exists at ${pathname}`);
```
develohpanda: This needs some refinement as it seems somewhat misleading or generic...

For example, in the instance where inso is called as simply `inso run test` _and if no git repository is found_ the expectation is to look at the app data directory through `envPaths` - `currentSrc` doesn't check for that, and could even provide a misleading result.

develohpanda (changes requested):
Large review for a large changeset, bear with me! :hugs: Good work with the insomnia adapter 👏🏽 

The changes in `packages/insomnia-inso/src/db/index.js` in particular are quite tricky to review because of all the rules present and all of the changed unit tests. I don't think it quite matches the current behavior so I have reservations about accepting a large refactor to that file.

If it is possible, my preference would be to stick with the existing logic closely and try fit an extra `if` block in the logic to support insomnia export files. There may be some degree of code duplication and that's okay.

But before you embark on another refactor, were you facing issues with the existing logic that pushed you to look at making changes there? Just keen to understand the decision, in case I am missing something. 😄

MrSnix:
> Large review for a large changeset, bear with me! 🤗 Good work with the insomnia adapter 👏🏽

Smaller commits, longer changeset 😍 
(I bear with you, you shouldn't bear with me xD, I write a lot of stuff, I increase your workload)

Thanks, appreciated.

> The changes in `packages/insomnia-inso/src/db/index.js` in particular are quite tricky to review because of all the rules present and all of the changed unit tests. I don't think it quite matches the current behavior so I have reservations about accepting a large refactor to that file.

That's okay. I can revert to the original logic and work on it. 👍🏻 (no prob, really)

> But before you embark on another refactor, were you facing issues with the existing logic that pushed you to look at making changes there? Just keen to understand the decision, in case I am missing something. 😄

Yes, sure.
1. I usually refactor before updating or modifying something on the current code "hoping" to leave the code in a better state.
_(Even if, I honestly think, this refactor it's not so huge, was it?)_
2. I **tried** to improve the readability and reduce the complexity of the `loadDb` function.
3. I felt like some stuff of the current implementation is quite tricky and a little bit (a tiny bit) weird:
- Why is the CLI allowed to look for stuff inside a non-existing path? I know this is handled in some way but... still it's not that linear. For example, we know that a given path does not exist when we have tried loading all the adapters on it...Wat?
It's not even that simple, we know that there is no data in that directory because all loaders failed so "officially" the directory may be valid but we don't know for sure if loaders failed due to data absence or path not existent... do you see my point?
- Some adaptors have some IO checks before start working on data. Why these steps can't be done before and leave the adaptor alone with its code about being an adaptor rather than checking for data consistency.
- Is it really necessary to check for `db` nullity in order to see if the next adapter is needed?
- Why do we have 4/5 loggers which say the same stuff instead of looping on it?
- Why can't we reduce some `if` with variables? (I know, the logic stays the same but it is cleaner)

I hope I clarified my point but I see, in some ways, this is a bit "legacy" so maybe it's better to not alter the current flow.

**I want to thank you for the time you took to review my code, you so awesome (y), every change was made with the best intention but I'm okay if it is not the right way**. Thanks again 😃 and again...and again...and again...

~Let me know if you want definitely revert to the previous state because I saw some review comments left on it.~

packages/insomnia-inso/src/db/index.js
MrSnix: My mistake.

packages/insomnia-inso/src/commands/export-specification.js
MrSnix: Fixed it

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
MrSnix: Done

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
MrSnix: Adaptors were using this syntax, I thought it had to be used here too.

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
MrSnix: Yes, I'll try to make more attention to this.

packages/insomnia-inso/src/db/index.js
MrSnix: Done

MrSnix:
As requested:

**Changes**:
- Tests have been reverted to their original state
- `loadDb()` has been reverted to its original state (plus my code)
- I addressed the changes requested by the code review
- I simplified the multiple iterations in one single loop

**Basic question:**
I don't know exactly what I have to do when I have a directory of multiple v4.json files, should I merge everything in one big database? Is it correct?

packages/insomnia-inso/src/cli.js
```diff
@@ -171,9 +171,10 @@ export function go(args?: Array<string>, exitOverride?: boolean): void {
   // Global options
   cmd
     .option('-w, --workingDir <dir>', 'set working directory')
-    .option('-a, --appDataDir <dir>', 'set the app data directory')
+    .option('-a, --appDataDir <dir>', 'set the app data directory (deprecated; use --src instead')
     .option('--config <path>', 'path to configuration file')
     .option('--verbose', 'show additional logs while running the command')
+    .option('--src <file|directory>', 'Set the app data source')
```
develohpanda: Matching the other options 🤓 

```suggestion
    .option('--src <file|dir>', 'set the app data source')
```

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,47 @@
...
+  if (!fs.existsSync(path) || fs.lstatSync(path).isDirectory()) return null;
+  // Init Db var
+  const db = emptyDb();
+  // Now, reading and parsing
+  const content = await fs.promises.readFile(path, { encoding: 'utf-8' });
+  const object = YAML.parse(content);
+  // We are supporting only v4 files
+  if (object.__export_format !== 4) return null;
+  // Read resources field which should contains all available objects
+  const importedModels = object?.resources;
```
develohpanda: Optional chaining seems unnecessary here, especially given the line before this already accesses a property on `object`. The following `forEach` will also fail if this exits early as `null`.

Also a suggestion to rename `object` to `parsed` or something similar. "object" is a very generic name and generally discouraged.

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+import fs from 'fs';
+import type { DbAdapter } from '../index';
+import YAML from 'yaml';
+import { emptyDb } from '../index';
+
+const modelTypeToExportTypeMap = {
```
develohpanda: A short comment here to explain why this exists might be worth-while for future explorers 🙂

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,47 @@
...
+  const db = emptyDb();
+  // Now, reading and parsing
+  const content = await fs.promises.readFile(path, { encoding: 'utf-8' });
+  const object = YAML.parse(content);
+  // We are supporting only v4 files
+  if (object.__export_format !== 4) return null;
+  // Read resources field which should contains all available objects
+  const importedModels = object?.resources;
+  // Check if we have to filter
+  const restricted = filterTypes && filterTypes?.length !== 0;
```
develohpanda: No need for optional chaining as the first part of this already ensures `filterTypes` exists

```suggestion
  const restricted = filterTypes && filterTypes.length !== 0;
```

packages/insomnia-inso/src/db/adapters/insomnia-adapter.js
```diff
@@ -0,0 +1,47 @@
...
+    // Rename field, transform value and delete obsolete one
+    item.type = modelTypeToExportTypeMap[item._type];
+    delete item._type;
+    // Filtering
+    if (restricted) {
+      db[item.type] && filterTypes.includes(item.type) && db[item.type].push(item);
+    } else {
+      db[item.type] && db[item.type].push(item);
+    }
+  });
```
develohpanda: Great start but this is doing more than necessary and there are some improvements I can see.

Let's say we are loading only `ApiSpec` from a large V4 export file (by setting `filterTypes`). In this implementation, regardless of if an item is an `ApiSpec` or not, we are mutating the item by adding a `.type` property and deleting the `._type`.
 
We should entirely skip doing anything to a model if we intend to ignore it.

Additionally, `array.includes()` doesn't happen in linear time so as the lists of `importedModels` and `filterTypes` grow (`importedModels` could be _huge_), this could become a real bottleneck. It's better to convert the `filterTypes` array to a set for this use case ([ref](https://stackoverflow.com/a/53423936)).

packages/insomnia-inso/src/db/index.js
```diff
@@ -42,30 +43,45 @@ export type DbAdapter = (
...
 
+  appDataDir &&
+    logger.warn(
+      'The option --app-data-dir has been deprecated and will be removed in future releases.\n' +
+        'Please, consider using --src as alternative',
+    );
+
   // return empty db
   if (!db) {
     logger.warn(
```
develohpanda: This warning needs to update now to include the new data source.

> No git, app data store, or Insomnia V4 export file was found, re-run ... information

develohpanda (changes requested):
Almost there! 🎉 A couple of observations

> I don't know exactly what I have to do when I have a directory of multiple v4.json files, should I merge everything in one big database? Is it correct?

No need to handle that case with this PR (nor do I know the desirable answer to that at this stage, because there are nuances with each approach).

abbasc52:
Any update on this? this PR seems to be pending for a long time

develohpanda:
Hey @abbasc52, sorry about the slow progress with this, it _will_ be merged I just don't have a timeline right now 😞 will see if I can get to having a look at it again in the coming week.

MrSnix:
Hi @develohpanda 😄, I resolved all the conflicts and rewrote everything in Typescript.
I did the best I could, I hope I didn't introduce new bugs 🐜 with this one.

Let me know as always!

EDIT: I ran an empty commit due to a weird Windows failure unrelated to my code.

develohpanda:
Thanks @MrSnix, awesome! On my radar 👀

develohpanda:
These options are getting a little confusing now, so to write the rules out in plain English:

### Current state (after this PR)

We have three data options:
* `--workingDir`: used for specifying the current working directory, both for data input (only git) and output (eg [exporting a spec](https://github.com/Kong/insomnia/blob/5f4c19da35e90ffa2f3ae27e9b309ee48f9d4e16/packages/insomnia-inso/src/commands/export-specification.ts#L28-L28))
* `--src`: used to specify the V4 export file or the Insomnia app data directory
* `--appDataDir`: deprecated in this PR, superseeded by `--src`

We have three data sources:
* the git data _directory_
	* search for `.insomnia` filter in `--workingDir` or `.`
	* only if no `--appDataDir` or `--src` set
* a V4 insomnia export _file_
	* search for file using `--src`
* the Insomnia app data _directory_
	* search for app data using `--src`, then `--appDataDir`, then the [default location](https://support.insomnia.rest/article/202-application-data)

For a code example, see [here](https://github.com/MrSnix/insomnia/blob/839dd61ecaf901182cc8784be6b32bd0dafc369e/packages/insomnia-inso/src/db/index.ts#L49-L49):

<details>
<summary>Screenshot</summary>

image

</details>

### End state

I think (in the future) if we were to update the _git_ data source to also look at `--src` instead of only `--workingDir`, we'll be able to simplify this and say something like the following:

> Inso will search for a `.insomnia` folder in your working directory, otherwise fall back to your default [Insomnia application directory](https://support.insomnia.rest/article/202-application-data). To override the data source, or to use a V4 export file, use `--src`.

I'll revisit the options with @wdawson some time, no changes requested in this PR right now!

develohpanda:
@dimitropoulos see original feature request [here](https://github.com/Kong/insomnia/issues/2666), this comment https://github.com/Kong/insomnia/pull/3055#issuecomment-871148003 and the final PR implementation, you should be able to skip the remaining conversations

develohpanda (approved):
I made a couple of updates [here](https://github.com/Kong/insomnia/pull/3055/commits/839dd61ecaf901182cc8784be6b32bd0dafc369e), but PR looks good to me, pending the result of discussing this and getting product approval on it https://github.com/Kong/insomnia/pull/3055#issuecomment-871148003! 🙌🏽 

Thanks a **tonne** @MrSnix. 🥳

develohpanda:
> I think (in the future) if we were to update the _git_ data source to also look at `--src` instead of only `--workingDir`, we'll be able to simplify this and say something like the following:
> 
> > Inso will search for a `.insomnia` folder in your working directory, otherwise fall back to your default [Insomnia application directory](https://support.insomnia.rest/article/202-application-data). To override the data source, or to use a V4 export file, use `--src`.

We discussed and agreed on this end state, but it would be as a separate PR, so no more blockers on this PR. Just waiting for another review from @dimitropoulos 🙌🏽

StevenReitsma:
Any news here? This is a great feature that I would gladly make use of!

packages/insomnia-inso/src/db/adapters/insomnia-adapter.ts
```diff
@@ -0,0 +1,94 @@
...
+  _id: string;
+  _type: string;
+  parentId: string;
+}
+
+const transformRawToImported = (item: RawBaseModel): BaseModel => ({
+  _id: item._id,
+  parentId: item.parentId,
+  type: modelTypeToExportTypeMap[item._type],
+});
```
develohpanda: Looks like this mapping is missing the `name` property, which is the likely cause for why the test suite name is being read as `undefined` and the spec name is the `_id` 🤔 

image

Not sure if this happened during the rebase or prior to that

develohpanda (changes requested):

packages/insomnia-inso/src/db/adapters/insomnia-adapter.ts
dimitropoulos: https://user-images.githubusercontent.com/15232461/129989337-94fe17af-76a6-4e63-870c-c83c9fd4725f.mp4

here's a video of it (you can use the insomnia v4 fixture in this project and get the same result).  and btw, since this video was taken it was determined that the suite name (like shown in the comment above) is where this error is stemming from.

develohpanda: Fixed now, shows the expected content

image

develohpanda:
QA with four invalid files

image

QA with two valid files

image

develohpanda (approved):
LGTM. Thanks so much for the time you put into this and for your patience @MrSnix 🙌🏽 

### QA
<details>
<summary>🟢 Four invalid files</summary>

image

</details>

<details>
<summary>🟢 Two valid files</summary>

image

</details>

<details>
<summary>🟢 Deprecation warning</summary>

image

</details>

dimitropoulos (approved):
re-tested everything and all looks good now.  I re-ran linting as it has changed since this was last rebased and also made a very minor adjustment to an error message.  thanks again @MrSnix!  This is a big one!

MrSnix:
I'd like to thank @develohpanda and @dimitropoulos for the support and the time spent with me on reviewing this one:
you are always available, kind, and helpful. 

I'm super glad to help this beautiful community.

StevenReitsma:
Is there going to be a new inso release soon that includes this PR? Building the project myself is easy but an npm release is even easier!

wdawson:
Hey, @StevenReitsma (and @MrSnix ). We're going to try to get a release out by the end of October. No promises but that's our target right now. Hopefully, we'll have a beta build before that too!

PR #3066, opened 2021-02-09
route path end marker regex

dimitropoulos:
Generating Declarative Configuration for the input route path `/pet` using o2k produces `/pet` instead of `/pet$` which inside of Kong becomes `/pet/*`, which is incorrect.  This PR makes all route paths regexes and adds the ending marker in all cases.


packages/openapi-2-kong/src/common.js
```diff
@@ -64,15 +64,12 @@ export function generateSlug(str: string, options: SlugifyOptions = {}): string
...
 
 export function pathVariablesToRegex(p: string): string {
   const result = p.replace(pathVariableSearchValue, '(?<$1>\\S+)');
-  if (result === p) {
-    return result;
-  }
-
-  // If anything was replaced, it's a regex, so add a line-ending match
+  // add a line ending because it is a regex
   return result + '$';
```
dimitropoulos, line 73: my only concern is that if someone passes a regex that already ends with `$` it will get another, e.g. `/foo$` will go to `/foo$$`.  I was about to write a test for this but when I noticed that it meant that this never happens in any test case (because without checking for this all tests passed) so I then concluded that my ignorance is leading me astray in assuming an input to `pathVariablesToRegex` may already be a string ending with `$`.

I wanted to call that assumption out, though, to be sure.

nijikokun: Since regular expressions aren't allowed in OpenAPI, we can assume that the user meant it as a special character, and we can treat the cases of this on a per-user basis.

dimitropoulos: sweet - thanks!

packages/openapi-2-kong/src/common.js
```diff
@@ -64,15 +64,12 @@ export function generateSlug(str: string, options: SlugifyOptions = {}): string
   return slugify(str, options);
 }
 
+/** characters in curly brances not immediately followed by `://`, e.g. `{foo}` will match but `{foo}://` will not. */
```
develohpanda, line 67: 👍🏽

develohpanda (approved):
Looks to be working; nice 💯

packages/openapi-2-kong/src/__fixtures__/petstore-expanded.yaml
```diff
@@ -141,7 +141,7 @@ components:
         name:
           type: string
         tag:
-          type: string    
+          type: string
```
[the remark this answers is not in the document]
dimitropoulos: hahah :)

PR #3068, opened 2021-02-09
fix: should only open as design document if scope is designer

develohpanda:
Previously, `workspace.scope` could equal `null`, in which case a check existed to ensure that if a "spec" exists, it opens as a design document.

Now that Core and Designer are merged, a workspace from Core should open as a collection and show the requests.

A case in which this does _not_ happen is if the workspace in Core was imported from an API spec (for example, [petstore](https://gist.githubusercontent.com/gschier/4e2278d5a50b4bbf1110755d9b48a9f9/raw/801c05266ae102bcb9288ab92c60f52d45557425/petstore-spec).

Import the following workspace into the latest develop and try to open it - it will navigate to the "Debug" tab because an `apiSpec` entry exists, but it will not show the toggle switches because the workspace is a _collection from Core_.

[bug.txt](https://github.com/Kong/insomnia/files/5954125/bug.txt)

Now that we know for _certain_ (as per the migration in `workspace`) that `workspace.scope` will either be `collection` or `designer`, we can simply use that to navigate between the different types of workspaces.

This PR:

Develop:

(Note that the contents of the card and the click handler vary depending on the condition, hence the differences in the gifs)

cc: @codeasashu

Fixes #INS-463


DMarby (approved):
Nice catch 👍

codeasashu:
Thank you :)

PR #3072, opened 2021-02-10
use operationId where possible in o2k generation

dimitropoulos:
Currently o2k generates custom incremental route names, Swagger already has a mechanism which allows the user to give routes a unique identifier through operationId which o2k should use before falling back to generated route names.


packages/insomnia-app/package.json
```diff
@@ -100,7 +100,7 @@
     "@hot-loader/react-dom": "^16.8.6",
     "@stoplight/spectral": "^4.1.1",
     "aws4": "^1.9.0",
-    "axios": "^0.19.0",
+    "axios": "^0.21.1",
```
DMarby: I think this change is fine, but in general we try to avoid unrelated changes in a PR, so in the future it would be good if something like this went into a separate PR

nijikokun (approved):
💯

packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -95,20 +95,31 @@ export function generateRouteName(
...
-    const pathSlug = generateSlug((pathItem: Object)['x-kong-name']);
+  // as a product requirement, `x-kong-name` values takes precedence over `operationId` or other OpenAPI values
+
+  // method-level `x-kong-name` has the highest priority
+  if (typeof pathItem[method]['x-kong-name'] === 'string') {
+    return pathItem[method]['x-kong-name'];
+  }
+
+  // path-level `x-kong-name` has the second highest priority
+  if (typeof pathItem['x-kong-name'] === 'string') {
```
nijikokun, line 108: Actually, missed this in my initial review, the order here - while directive based - should also be hierarchical. Since `x-kong-name` is on the path, and not the method it should be as follows:

Current:

- `path.method.x-kong-name`
- `path.x-kong-name`
- `path.method.operationId`
- `generated`

Expected:

- `path.method.x-kong-name`
- `path.method.operationId`
- `path.x-kong-name`
- `generated`

nijikokun: Actually, missed this in my initial review, the order here - while directive based - should also be hierarchical. Since `x-kong-name` is on the path, and not the method it should be as follows:

Current:

- `$.path.method.x-kong-name`
- `$.path.x-kong-name`
- `$.path.method.operationId`
- `generated`

Expected:

- `$.path.method.x-kong-name`
- `$.path.method.operationId`
- `$.path.x-kong-name`
- `generated`

nijikokun (changes requested):

DMarby (approved):
Awesome 🎉

packages/openapi-2-kong/src/__fixtures__/api-with-examples.expected.json
```diff
@@ -9,14 +9,14 @@
...
-          "name": "Simple_API_overview-path-get",
+          "name": "x-kong-name-override-at-method",
           "paths": ["/$"]
         },
         {
           "methods": ["GET"],
           "strip_path": false,
           "tags": ["OAS3_import", "OAS3file_api-with-examples.yaml"],
-          "name": "Simple_API_overview-path_1-get",
+          "name": "Simple_API_overview-x_kong_name_override_at_path_item-get",
```
develohpanda: Are we aware of any name length limits in Kong?

packages/openapi-2-kong/src/__fixtures__/api-with-examples.yaml
```diff
@@ -6,7 +6,9 @@ servers:
   - url: http://backend.com/path
 paths:
   /:
+    x-kong-name: x-kong-name-override-at-path-item
     get:
+      x-kong-name: x-kong-name-override-at-method
```
develohpanda, line 11: 💯

packages/openapi-2-kong/src/declarative-config/services.js
develohpanda: I recall there is similar logic in the kong-for-kubernetes generation, they should probably match when we get around to updating that one as well

packages/openapi-2-kong/types/openapi3.flow.js
```diff
@@ -1,5 +1,9 @@
 // @flow
 
+declare type XKongName = {
+  'x-kong-name'?: string,
+};
```
develohpanda, line 5: Here comes the venture into flow types 😆 

These flow declaration files are _mostly_ pretty good. TS didn't complain too much after converting, yay!

develohpanda (approved):
LGTM, awesome!

packages/openapi-2-kong/types/openapi3.flow.js
dimitropoulos: :)

packages/openapi-2-kong/src/declarative-config/__tests__/names.test.js
```diff
@@ -0,0 +1,77 @@
...
+        },
+      },
+    },
+    '/planet-smasher',
+    'post',
+  );
+  expect(name).toEqual(expected);
+};
+
+describe('names', () => {
```
DMarby, line 28: 🔥

DMarby (approved):
:shipit:

packages/openapi-2-kong/src/__fixtures__/api-with-examples.expected.json
dimitropoulos: I hope not, haha, because one of the prior ones was (before this PR) `httpbin-Returns_the_specified_HTTP_status_code_or_a_random_status_code_if_more_than_one_are_given-delete`

develohpanda: 🤣

dimitropoulos: (resolving since no name is now longer than before)

nijikokun (approved):

PR #3073, opened 2021-02-10
Migration verbiage changes

develohpanda:
image

image

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper-migration.js
+++ b/packages/insomnia-app/app/ui/components/wrapper-migration.js
@@ -77,11 +77,15 @@ const Options = ({ start, cancel }: OptionsProps) => {
   return (
     <>
       <p>
-        <strong>Would you like to migrate data from Insomnia Designer?</strong>
+        <strong>Migrate from Insomnia Designer</strong>
+      </p>
+      <p>
+        Insomnia Designer and Core are now Insomnia! Select the items below you'd like to migrate
+        from Designer.
       </p>
       <div className="text-left margin-top">
         <BooleanSetting
-          label="Use Designer Settings"
+          label="Copy Designer Application Settings"
           name="useDesignerSettings"
           options={options}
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/ui/components/wrapper-migration.js
```diff
@@ -77,11 +77,15 @@ const Options = ({ start, cancel }: OptionsProps) => {
   return (
     <>
       <p>
-        <strong>Would you like to migrate data from Insomnia Designer?</strong>
+        <strong>Migrate from Insomnia Designer</strong>
+      </p>
+      <p>
+        Insomnia Designer and Core are now Insomnia! Select the items below you'd like to migrate
+        from Designer.
```
develohpanda, line 84: This one seems interesting; a lot of things will auto migrate but this wording implies they can only choose the three things (settings, plugins, responses).

reynolek: Are the things that would auto migrate directly associated with the 3 listed things?

develohpanda: Nope; the additional items are all the database models (which users will recognize as workspaces, requests, environments), so naturally they're not optional. These settings are just some additional options they have.

reynolek: Perhaps we can do something like "Along with the required items, you may optionally migrate the below items from Designer" or something? @nijikokun

nijikokun: Actually, thinking, we should let users have the choice of whether or not to migrate workspaces (disabling should disable responses as well)

develohpanda: INS-469

nijikokun (approved):

PR #3074, opened 2021-02-11
Add pull dropdown to homepage

develohpanda:

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.js
@@ -144,15 +144,15 @@ class RemoteWorkspacesDropdown extends React.Component<Props, State> {
       return !(workspaceExists && localProjectExists);
     });
 
+    const button = (
+      <Button variant="contained" bg="surprise" className={className}>
+        Pull
+        <i className="fa fa-caret-down pad-left-sm" />
+      </Button>
+    );
+
     return (
-      <Dropdown
-        onOpen={this._refreshRemoteWorkspaces}
-        renderButton={() => (
-          <Button variant="contained" bg="surprise" className={className}>
-            Pull
-            <i className="fa fa-caret-down pad-left-sm" />
-          </Button>
-        )}>
+      <Dropdown onOpen={this._refreshRemoteWorkspaces} renderButton={button}>
         <DropdownDivider>
           Remote Workspaces{' '}
           <HelpTooltip>
--- a/packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.js
+++ b/packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.js
@@ -1,0 +1,189 @@
+// @flow
+
+import * as React from 'react';
+import { autoBindMethodsForReact } from 'class-autobind-decorator';
+import { AUTOBIND_CFG } from '../../../common/constants';
+import * as session from '../../../account/session';
+import VCS from '../../../sync/vcs';
+import type { Project } from '../../../sync/types';
+import { Dropdown, DropdownDivider, DropdownItem, Button } from 'insomnia-components';
+import type { Workspace } from '../../../models/workspace';
+import HelpTooltip from '../help-tooltip';
+import * as models from '../../../models';
+import * as db from '../../../common/database';
+import { showAlert } from '../modals';
+
+type Props = {
+  className?: string,
+  vcs?: VCS,
+  workspaces: Array<Workspace>,
+};
+
+type State = {
+  loading: boolean,
+  localProjects: Array<Project>,
+  pullingProjects: { [string]: boolean },
+  remoteProjects: Array<Project>,
+};
+
+@autoBindMethodsForReact(AUTOBIND_CFG)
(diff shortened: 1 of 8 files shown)


packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.js
```diff
@@ -0,0 +1,189 @@
...
+              pullingProjects[p.id] ? (
+                <i className="fa fa-refresh fa-spin" />
+              ) : (
+                <i className="fa fa-cloud-download" />
+              )
+            }>
+            <span>
+              Pull <strong>{p.name}</strong>
+            </span>
+          </DropdownItem>
```
develohpanda, line 182: Remove unused code from workspace-dropdown

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.js
```diff
@@ -0,0 +1,189 @@
+// @flow
```
develohpanda, line 1: Most of this (especially the pull logic) is a straight copy-paste from `workspace-dropdown`

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.js
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  render() {
+    const { className, workspaces } = this.props;
+
+    const { loading, remoteProjects, localProjects, pullingProjects } = this.state;
+
+    if (!session.isLoggedIn()) {
+      return null;
+    }
```
develohpanda, line 135: Should this show a "log in" button or dropdown item if not signed in?

packages/insomnia-app/app/ui/components/dropdowns/workspace-dropdown.js
```diff
@@ -303,50 +193,10 @@ class WorkspaceDropdown extends React.PureComponent<Props, State> {
             <i className="fa fa-globe" /> Share <strong>{activeWorkspace.name}</strong>
           </DropdownItem>
 
-          <DropdownDivider>Switch Workspace</DropdownDivider>
-
-          {nonActiveWorkspaces.map(w => {
-            const isUnseen = !!unseenWorkspaces.find(v => v._id === w._id);
```
DMarby, line 309: Not a blocker, maybe out of scope for this PR, but I guess the unseenWorkspaces prop/redux selectors/etc can be removed completely now, because it doesn't seem to be used anywhere anymore with this change?
There are some other props in here that are unused now as well in regards to workspaces since the workspaces list has moved.

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
```diff
@@ -117,6 +118,12 @@ class SyncDropdown extends React.PureComponent<Props, State> {
       }
     }
 
+    // Don't really need to wait for this to finish
+    models.workspaceMeta.updateByParentId(workspace._id, {
```
DMarby: Since `refreshMainAttributes` only gets called after `syncItems` has changed when switching branches, this leads to stale data showing when switching to a new branch and then going back to the dashboard, as no documents gets modified, so syncItems doesn't change.

I don't think there's an easy way to resolve that without potentially breaking the already brittle state machine, so I'd suggest just setting these to null in `_handleSwitchBranch` for now, since imo not showing anything is better than showing stale data.

DMarby (approved):
Looks and works great!
One small comment about some stale data when switching branches.

packages/insomnia-app/app/ui/components/dropdowns/workspace-dropdown.js
develohpanda: Yeah great catch! I'll clean this up today. The initial plan was to remove the workspace-dropdown entirely but since we've decided to keeping it around it makes sense to clean this up. :)

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
develohpanda: I thought there might be some edge cases here :( I'll remove this change and yeah, favor not showing anything. Quite often the git-sync one is also stale but I won't touch that.

Currently we also don't have a way to show "this collection is linked to beta sync" on the homepage but it can be treated as a separate feature request

packages/insomnia-app/app/ui/components/dropdowns/workspace-dropdown.js
develohpanda: Actually, for the sake of time and avoiding conflicts I won't do this in the current PR. This file has more changes in #3076, so I'll tidy it up with the next (and final) PR in this area for INS-412. 🤗

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.js
develohpanda: Have removed it all. Any opinions on reverting the change I made to `vcs.getHistory()`? It's small and has unit tests, and I think we'll likely need it when we eventually come to fix this.

DMarby: Sounds fine to keep, definitely a useful thing to have around

PR #3076, opened 2021-02-11
Move share button and put settings on all pages

develohpanda:
## Settings button + tooltip + hotkey
image

## Account
image

image

## Share button for collections
image

## Account & Preferences in documents
image

Fixes INS-408, Fixes INS-410, Fixes INS-411


packages/insomnia-app/app/ui/components/wrapper-debug.js
```diff
@@ -59,19 +61,29 @@ class WrapperDebug extends React.PureComponent<Props> {
     const collection = activeWorkspace.scope === 'collection';
     const designer = !collection;
 
+    const share = collection && (
```
DMarby: I think we want to hide this when not logged in as well for now, for consistency with the sync button, and as currently clicking it while logged out just gives you an error message in the share modal? cc: @nijikokun

DMarby (approved):

DMarby:
Looks great! Just one question regarding whether the share button should be hidden while not logged in or not.

PR #3087, opened 2021-02-14
Insomnia CLI - Reporter integration

### LINKED ISSUE #3059: No HTML Reporter integration (INSO CLI)
It was communicated to us that 

**inso run test myTestinso spc_ba58e1 --reporter html --bail --keepFile** 

This command should generate results in html file seems to be it doesn't work that way.

HTML report generation for results are required  and it's important feature for anything.

Also tried installing https://www.npmjs.com/package/mocha-simple-html-reporter but it doesn't work. 

Mocha HTML reporter integration doesn't work with insomnia

image

Please contact nijji from insomnia product team for more information

MrSnix:
The following PR contains these changes:
- Removed blocking-code for external reporters
- ~When using an external reporter (installed with npm) an info message is printed to say "be careful, we try our best but it may fail...if you can use our internal reporters".~
- If the reporter is external, now it is able to print on console (because of some Insomnia code preventing it), so the result can be eventually piped
- If the external reporter is not found, it is now caught from Mocha and displayed to the user in a nice way

To-Do:
- ~Handle arguments given to the reporter~

Closes #3059

### DIFF
--- a/packages/insomnia-inso/package.json
+++ b/packages/insomnia-inso/package.json
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "@stoplight/spectral": "^5.4.0",
-    "commander": "^5.1.0",
+    "commander": "^6.1.0",
     "consola": "^2.15.0",
     "cosmiconfig": "^6.0.0",
     "enquirer": "^2.3.6",
--- a/packages/insomnia-inso/README.md
+++ b/packages/insomnia-inso/README.md
@@ -190,7 +190,7 @@ Scope by document name or id using an external reporter
 
 ``` sh
 inso run test spc_ff27a1 --reporter my-custom-reporter
-inso run test spc_ff27a1 --reporter my-custom-reporter --reporter-options key1=value1 key2=value2
+inso run test spc_ff27a1 --reporter my-custom-reporter --reporterOptions key1=value1 key2=value2
 ```
 **Important**:
 Mocha and the external reporter must be installed as global in order to work
--- a/packages/insomnia-inso/README.md
+++ b/packages/insomnia-inso/README.md
@@ -147,7 +147,8 @@ The test runner is built on top of Mocha, thus many of the options behave as the
 |Option|Alias|Description|
 |- |- |- |
 | `--env <identifier>` | `-e` |the environment to use - an environment name or id |
-| `--reporter <value>` | `-r` |reporter to use, options are `dot, list, min, progress, spec` (default: `spec` )|
+| `--reporter <value>` | `-r` |specify reporter to use (default: `spec` )|
+| `--reporter-options <option> [options...]` | `-ro` |reporter-specific options|
 | `--testNamePattern <regex>` | `-t` | run tests that match the regex|
 | `--bail` | `-b` | abort ("bail") after the first test failure|
 | `--keepFile` | | do not delete the generated test file (useful for debugging)|
@@ -169,7 +170,7 @@ inso run test "Sample Specification" --env "OpenAPI env"
 inso run test spc_46c5a4 --env env_env_ca046a
 ```
 
-Scope by the a test suite name or id
+Scope by test suite name or id
 
 ``` sh
 inso run test "Math Suite" --env "OpenAPI env"
@@ -184,6 +185,15 @@ inso run test spc_46c5a4 --reporter progress --bail --keepFile
 ```
 
 More examples: [#2338](https://github.com/Kong/insomnia/pull/2338).
+
+Scope by document name or id using an external reporter
+
+``` sh
+inso run test spc_ff27a1 --reporter my-custom-reporter
+inso run test spc_ff27a1 --reporter my-custom-reporter --reporter-options key1=value1 key2=value2
+```
+**Important**:
+Mocha and the external reporter must be installed as global in order to work
 
 ## `$ inso export spec [identifier]`
 
--- a/packages/insomnia-inso/package-lock.json
+++ b/packages/insomnia-inso/package-lock.json
@@ -2444,9 +2444,9 @@
 			"integrity": "sha1-p9BVi9icQveV3UIyj3QIMcpTvCU="
 		},
 		"commander": {
-			"version": "5.1.0",
-			"resolved": "https://registry.npmjs.org/commander/-/commander-5.1.0.tgz",
-			"integrity": "sha512-P0CysNDQ7rtVw4QIQtm+MRxV66vKFSvlsQvGYXZWR3qFU0jlMKHZZZgw8e+8DSah4UDKMqnknRDQz+xuQXQ/Zg=="
+			"version": "6.2.1",
+			"resolved": "https://registry.npmjs.org/commander/-/commander-6.2.1.tgz",
+			"integrity": "sha512-U7VdrJFnJgo4xjrHpTzu0yrHPGImdsmD95ZlgYSEajAn2JKzDhDTPG9kBTefmObL2w/ngeZnilk+OV9CG3d7UA=="
 		},
 		"commondir": {
 			"version": "1.0.1",
--- a/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
+++ b/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
@@ -164,8 +164,8 @@ Run Insomnia unit test suites
 Options:
   -e, --env <identifier>         environment to use
   -t, --testNamePattern <regex>  run tests that match the regex
-  -r, --reporter <reporter>      reporter to use, options are [dot, list, min,
-                                 progress, spec] (default: spec)
+  -r, --reporter <reporter>      standard reporters options are [dot, list,
+                                  min, progress, spec] (default: spec)
   -b, --bail                     abort (\\"bail\\") after first test failure
   --keepFile                     do not delete the generated test file
   -h, --help                     display help for command"
--- a/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
+++ b/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
@@ -162,11 +162,11 @@ exports[`Snapshot for "inso run test -h" 1`] = `
 Run Insomnia unit test suites
 
 Options:
-  -e, --env <identifier>         environment to use
-  -t, --testNamePattern <regex>  run tests that match the regex
-  -r, --reporter <reporter>      standard reporters options are [dot, list,
-                                  min, progress, spec] (default: spec)
-  -b, --bail                     abort (\\"bail\\") after first test failure
-  --keepFile                     do not delete the generated test file
-  -h, --help                     display help for command"
+  -e, --env <identifier>                         environment to use
+  -t, --testNamePattern <regex>                  run tests that match the regex
+  -r, --reporter <reporter>                      standard reporters options are [dot, list,  min, progress, spec] (default: spec)
+  -ro, --reporter-options <option> [options...]  reporter-specific options
+  -b, --bail                                     abort (\\"bail\\") after first test failure
+  --keepFile                                     do not delete the generated test file
+  -h, --help                                     display help for command"
 `;
--- a/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
+++ b/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
@@ -162,11 +162,11 @@ exports[`Snapshot for "inso run test -h" 1`] = `
 Run Insomnia unit test suites
 
 Options:
-  -e, --env <identifier>                         environment to use
-  -t, --testNamePattern <regex>                  run tests that match the regex
-  -r, --reporter <reporter>                      specify report to use (default: spec)
-  -ro, --reporter-options <option> [options...]  reporter-specific options
-  -b, --bail                                     abort (\\"bail\\") after first test failure
-  --keepFile                                     do not delete the generated test file
-  -h, --help                                     display help for command"
+  -e, --env <identifier>                        environment to use
+  -t, --testNamePattern <regex>                 run tests that match the regex
+  -r, --reporter <reporter>                     specify report to use (default: spec)
+  -ro, --reporterOptions <option> [options...]  reporter-specific options
+  -b, --bail                                    abort (\\"bail\\") after first test failure
+  --keepFile                                    do not delete the generated test file
+  -h, --help                                    display help for command"
 `;
--- a/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
+++ b/packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
@@ -164,7 +164,7 @@ Run Insomnia unit test suites
 Options:
   -e, --env <identifier>                         environment to use
   -t, --testNamePattern <regex>                  run tests that match the regex
-  -r, --reporter <reporter>                      standard reporters options are [dot, list,  min, progress, spec] (default: spec)
+  -r, --reporter <reporter>                      specify report to use (default: spec)
   -ro, --reporter-options <option> [options...]  reporter-specific options
   -b, --bail                                     abort (\\"bail\\") after first test failure
   --keepFile                                     do not delete the generated test file
--- a/packages/insomnia-inso/src/cli.test.ts
+++ b/packages/insomnia-inso/src/cli.test.ts
@@ -179,6 +179,10 @@ describe('cli', () => {
       expect(() => inso('run test -r')).toThrowError();
     });
 
+    it('should throw error if reporterOptions argument is missing', () => {
+      expect(() => inso('run test -ro')).toThrowError();
+    });
+
     it('should call runInsomniaTests with expected options', () => {
       inso('run test uts_123 -e env_123 -t name -r min -b --keepFile');
       expect(runInsomniaTests).toHaveBeenCalledWith('uts_123', {
--- a/packages/insomnia-inso/src/cli.ts
+++ b/packages/insomnia-inso/src/cli.ts
@@ -59,7 +59,7 @@ const makeTestCommand = (commandCreator: CreateCommand) => {
     .option('-t, --testNamePattern <regex>', 'run tests that match the regex')
     .option(
       '-r, --reporter <reporter>',
-      `reporter to use, options are [${reporterTypes.join(', ')}] (default: ${defaultReporter})`,
+      `standard reporters options are [${reporterTypes.join(', ')}] (default: ${defaultReporter})`,
     )
     .option('-b, --bail', 'abort ("bail") after first test failure')
     .option('--keepFile', 'do not delete the generated test file')
--- a/packages/insomnia-inso/src/cli.ts
+++ b/packages/insomnia-inso/src/cli.ts
@@ -57,10 +57,7 @@ const makeTestCommand = (commandCreator: CreateCommand) => {
     .description('Run Insomnia unit test suites')
     .option('-e, --env <identifier>', 'environment to use')
     .option('-t, --testNamePattern <regex>', 'run tests that match the regex')
-    .option(
-      '-r, --reporter <reporter>',
-      `standard reporters options are [${reporterTypes.join(', ')}] (default: ${defaultReporter})`,
-    )
+    .option('-r, --reporter <reporter>', `specify report to use (default: ${defaultReporter})`)
     .option('-ro, --reporter-options <option> [options...]', 'reporter-specific options')
     .option('-b, --bail', 'abort ("bail") after first test failure')
     .option('--keepFile', 'do not delete the generated test file')
--- a/packages/insomnia-inso/src/cli.ts
+++ b/packages/insomnia-inso/src/cli.ts
@@ -1,6 +1,6 @@
 import { ConversionOption, conversionOptions, generateConfig } from './commands/generate-config';
 import { getVersion, logErrorExit1, exit } from './util';
-import { reporterTypes, runInsomniaTests, TestReporter } from './commands/run-tests';
+import { runInsomniaTests, TestReporter } from './commands/run-tests';
 import { lintSpecification } from './commands/lint-specification';
 import { exportSpecification } from './commands/export-specification';
 import { parseArgsStringToArgv } from 'string-argv';
--- a/packages/insomnia-inso/src/cli.ts
+++ b/packages/insomnia-inso/src/cli.ts
@@ -58,7 +58,7 @@ const makeTestCommand = (commandCreator: CreateCommand) => {
     .option('-e, --env <identifier>', 'environment to use')
     .option('-t, --testNamePattern <regex>', 'run tests that match the regex')
     .option('-r, --reporter <reporter>', `specify report to use (default: ${defaultReporter})`)
-    .option('-ro, --reporter-options <option> [options...]', 'reporter-specific options')
+    .option('-ro, --reporterOptions <option> [options...]', 'reporter-specific options')
     .option('-b, --bail', 'abort ("bail") after first test failure')
     .option('--keepFile', 'do not delete the generated test file')
     .action((identifier, cmd) => {
--- a/packages/insomnia-inso/src/cli.ts
+++ b/packages/insomnia-inso/src/cli.ts
@@ -61,6 +61,7 @@ const makeTestCommand = (commandCreator: CreateCommand) => {
       '-r, --reporter <reporter>',
       `standard reporters options are [${reporterTypes.join(', ')}] (default: ${defaultReporter})`,
     )
+    .option('-ro, --reporter-options <option> [options...]', 'reporter-specific options')
     .option('-b, --bail', 'abort ("bail") after first test failure')
     .option('--keepFile', 'do not delete the generated test file')
     .action((identifier, cmd) => {
--- a/packages/insomnia-inso/src/commands/run-tests.test.ts
+++ b/packages/insomnia-inso/src/commands/run-tests.test.ts
@@ -103,7 +103,16 @@ describe('runInsomniaTests()', () => {
       bail: false,
       keepFile: false,
       reporter: 'custom-reporter',
-      reporterOptions: ['key1=value1', 'key2=value2'],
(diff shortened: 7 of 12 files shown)


packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -24,74 +23,104 @@ export type RunTestsOptions = GlobalOptions & {
...
+  // Loading database instance
+  const db = await loadDb(options);
 
-  const { reporter, bail, keepFile, appDataDir, workingDir, env, ci, testNamePattern } = options;
+  // Check if any provider has been provided (Yeah, comedy king)
+  const isExternal = isExternalReporter(options);
 
-  const db = await loadDb({ workingDir, appDataDir });
+  if (isExternal) {
+    logger.info(`Reporter "${reporter}" is not recognized and is treated as external`);
```
nijikokun: While informational, we should avoid printing messages so they are not a part of any output piped to a file for other actions

nijikokun (changes requested):

packages/insomnia-inso/src/commands/run-tests.js
MrSnix: Not even if  - - verbose option is provided? 🤔

nijikokun: Yes, as it’s more likely an intentional choice and the output desired is to go into other commands expecting specific formatting

MrSnix:
@nijikokun 
I wanted to put my hands on this before but my week has been busy 😭 

Anyway, it was challenging but in the end, I **won**! Let's begin.

**Context**
In order to handle the `--reporter-options` I had to make some changes inside the `insomnia-testing` package,
unluckily it was shipped with `insomnia-inso` package using the npm repository and I wasn't able to
make changes on it. So, the first thing I did was to use the `npm link` as a substitute for the dependency,
in that way the changes I was doing on the original code were reflected inside the package.
After my playground 🎮  was ready, I started doing researches. 👨🏻‍🔬 

I spent some time on the [Mocha.js API](https://mochajs.org/api/mocha) documentation and I found out
I could use the following interface to pass the argument to the reporter [Mocha Reporter API](https://mochajs.org/api/mocha#reporter). 
So, I modified `runInternal()` to support the new argument (`--reporter-options`) then, 
I had to add a new argument to insomnia-cli in order to parse the string containing the options for the reporter. 
The problems didn't end there. Mocha API wants an object meanwhile I was able to parse only a string. 
I built a function to extract an object from the string which was given as argument and BOOM! 💥 . 
It happened. It worked.

**Let's see it**

`inso run test --reporter mocha-simple-html-reporter --reporter-options output=/.../.../output.html`

test

Now, here is our little friend.
image

Now, **I have some questions**:
1. Should I suppress the network messages?
2. I did changes on `insomnia-testing` too, should I push them separately or can I push everything here?

I'll wait for an answer before pushing, so I can commit accordingly. 🥇 

Thanks

MrSnix:
**I want to add one more thing because I'm not really sure about it** and  
if someone more expert than me can confirm it that would be great.  

In order to make `mocha-simple-html-reporter` be recognized by `insomnia-cli`
installing it as `global package` wasn't enough.
_(but I may assume it may be because `mocha-simple-html-reporter` is not invokable from the CLI)_

I had to install it as a dependency in order to make it works but **here comes the strange part.**

**Before** linking `insomnia-testing` with `insomnia-inso` I had to add the dependency to the `insomnia-inso` package
to make it recognizable. But **AFTER** linking it was not recognized anymore.  
Then, I **removed** the dependency from `insomnia-inso` package and I added it to the `insomnia-testing` package and... 
it is recognized again... **What?!**

So, my first assumption was that the first `node_modules` folder he finds is the one it searches into before throwing an error.

The final answer comes from here, the actual implementation:
```js
Mocha.prototype.reporter = function(reporterName, reporterOptions) {
[...]
    var reporter;
    // Try to load a built-in reporter.
    if (builtinReporters[reporterName]) {
      reporter = builtinReporters[reporterName];
    }
    // Try to load reporters from process.cwd() and node_modules
    if (!reporter) {
      try {
        reporter = require(reporterName);
      } catch (err) {
        if (err.code === 'MODULE_NOT_FOUND') {
          // Try to load reporters from a path (absolute or relative)
          try {
            reporter = require(path.resolve(utils.cwd(), reporterName));
          } catch (_err) {
            _err.code === 'MODULE_NOT_FOUND'
 [...]
```
[Source Code - Line 273](https://mochajs.org/api/mocha.js.html)

They are basically doing `require()`.

The require function will look for files in the following order:
1. Built-in core Node.js modules (like fs, path)
2. NPM Modules. It will look in the node_modules folder.
3. Local Modules. If the module name has a ./, / or ../, it will look for the directory/file in the given path.

**So here comes my final question:**
If I install `insomnia-inso` and `mocha-simple-html-reporter` as global packages, launching them from the OS command line, will make them work?

This answer should be helpful for @develohpanda (I think?) which was asking about 3rd party modules integrations.

develohpanda:
> **So here comes my final question:**
> If I install `insomnia-inso` and `mocha-simple-html-reporter` as global packages, launching them from the OS command line, will make them work?
> 
> This answer should be helpful for @develohpanda (I think?) which was asking about 3rd party modules integrations.

This was the expectation but it definitely needs some research; as you've found in your experiments it doesn't seem to work?

Ideally we should avoid installing/bundling external reporters within Inso, so if installing globally is not trivial, we'll need to look at a different approach.

The following suggestions are untested, just some findings:
- Looks like you can use [`NODE_PATH`](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders) to have Node load from a different location, eg. global modules.
- Another approach might be to create a custom wrapper reporter in Inso, which uses [`import-global`](https://www.npmjs.com/package/import-global) to extract the globally installed reporter.

> So, the first thing I did was to use the npm link as a substitute for the dependency,
in that way the changes I was doing on the original code were reflected inside the package.

You should be able to get away without using `npm link` because we're using `lerna` so dependencies should be linked. Are you running `inso` in [watch mode](https://github.com/Kong/insomnia/tree/develop/packages/insomnia-inso#development)?

> I did changes on insomnia-testing too, should I push them separately or can I push everything here?

If they're related to this feature request I suggest keeping them in this PR

nijikokun:
> as you've found in your experiments it doesn't seem to work?

From reading it sounds like the experiment was local to the development see @MrSnix earlier comments about how he was linking through npm, not global installation. It would be good to test global installation, as it sounds - from what I read - the require *should* look in the same `node_modules` directory as the globally installed `inso` and should find the relative `mocha-simple-html-reporter` package.

`npm link` *should* do this however, it doesn't always map node module lookups correctly. Another way is to use [`npm pack`](https://docs.npmjs.com/cli/v7/commands/npm-pack) and then `npm install -g <packfile>`

> Ideally we should avoid installing/bundling external reporters within Inso, so if installing globally is not trivial, we'll need to look at a different approach.

Agree'd

develohpanda:
I think `npm link` is being used to make sure changes to `insomnia-testing` can be seen by `insomnia-inso`? Regardless, I misinterpreted the last question, thanks for the clarification. 😊

Another approach for installing the dev version of inso globally would be to build inso, then run `npm install -g packages/insomnia-inso` directly. Worth expanding the readme with this (or the pack file) .

If that finds the reporter, epic!

MrSnix:
**Before reading:**
_Warning_: I don't **really** know why it works like this (I can just assume) but this is what I found out.

This time it was a real adventure. 🤦🏻 

**Let's begin**
I used `npm pack` on `insomnia-inso` and built our beloved ❤️ `insomnnia-inso.tar`.
I executed `npm install -g ./insomnia-inso.tar` then I proceeded to install `mocha-simple-html-reporter` globally.
```shell
~ % npm -g list --depth=0 
/Users/xyz/.nvm/versions/node/v12.18.3/lib
├── insomnia-inso@2.2.28
├── mocha-simple-html-reporter@2.0.0
```

I wrote the command `inso run test --reporter mocha-simple-html-reporter`, I selected everything I had to... 
then I pressed Enter..........

Failure. 💥 

I was astonished. I was expecting it to work. The reporter was not recognized. 
So, I started thinking ... why it is not working? Both are installed globally... they should see each other.
I stared at the screen for a while and I was about to set the cli on fire ... **then something clicked-in.**

I took `mochawesome`, I installed it globally then I run again the same command...

**Working**, the reporter was recognized.

So, I run again `mocha-simple-html-reporter` with insomnia

**Working**, the reporter was recognized.

But this time I knew what I had to look for, **I felt enlightened.** 💡 

```shell
% npm -g list --depth=0
/Users/xyz/.nvm/versions/node/v12.18.3/lib
├── insomnia-inso@2.2.28
├── mocha@8.3.0
├── mocha-simple-html-reporter@2.0.0
├── mochawesome@6.2.2
```

Here the solution:
**If Mocha.js is NOT installed globally, other reporters won't be recognized.**

I don't know about your opinion but I supposed that:
if `insomnia-cli` (specifically `insomnia-testing`) already declares `mocha` as a direct dependency, it should have worked.
Well, it turned out: "No" it doesn't.

So I went on `mochawesome` repository and started doing some considerations which met my expectations 📖 .

Even if they declare and use `mocha` the same way as `insomnia-testing`, the `key difference` is that **they DO NOT list `mocha` as dependency BUT as **peer-dependency**** which makes sense now!
Proof: [mochawesome - package.json](https://github.com/adamgruber/mochawesome/blob/master/package.json) 

Maybe you knew already about it, but I was confused because nor `mocha-simple-html-reporter` nor `insomnia-inso` nor `insomnia-testing` declares `mocha` as a peer dep.

So, I strongly advise applying the same logic to `insomnia-inso` and / or `insomnia-testing` package. 📦 

After that, everything seemed working.
It took me some hours but definitely really funny 🤯  and worth it. 😂 

**So, now I ask, what's next?**
I'd like to keep testing a bit the new code and change the `package.json` as stated before.
Am I missing anything else?

**Q&A**

> You should be able to get away without using `npm link` because we're using `lerna` so dependencies should be linked. 

Lerna builds the packages in each directory but seems like it does not correlate (link) them. 
(I don't know a lot about Lerna, so I may be wrong 😢 , I apologize if the case)

> Are you running `inso` in [watch mode](https://github.com/Kong/insomnia/tree/develop/packages/insomnia-inso#development)?

The link reference `npm run start`, so I dug a bit and it is using webpack watch mode. 🤔 
So I went on the configuration and turned out it is building `insomnia-inso` (as expected) 😄  but is not linking or referencing `insomnia-testing` at all 😭 . I keep highlighting it references the npm general public copy (on the npm registry) rather than the local copy. (Correct me if wrong)

Thanks for reading!

develohpanda:
It looks like `mochawesome` is a reporter instead of a runner, so there isn't direct comparison between that and inso/insomnia-testing.

Also, peer dependencies are not automatically installed, and expecting users to install inso and mocha separately, globally, is not a desirable solution at the moment (last option), so this will need some more exploration unfortunately.

Lerna _does_ link them, but it doesn't alter the `package.json` so you won't see it there. As long as you run `npm run bootstrap` from the root of the repository, it will link dependencies correctly and not use the npm general copy for dependencies it can find in the monorepo.

develohpanda:
After playing around with this approach locally, it seems non-trivial and might need some webpack magic to handle it... so, careful how deep into the abyss you explore for now @MrSnix 😅

packages/insomnia-inso/package.json
```diff
@@ -63,7 +63,7 @@
   },
   "dependencies": {
     "@stoplight/spectral": "^5.4.0",
-    "commander": "^5.1.0",
+    "commander": "^6.1.0",
```
develohpanda, line 62: Is a major upgrade required for the contents of this PR? Ideally we should decouple this and do it separately, but if it is strictly needed we can keep it here.

packages/insomnia-inso/src/cli.js
```diff
@@ -61,8 +61,9 @@ function makeTestCommand(createCommand: CreateCommandType) {
     .option('-t, --testNamePattern <regex>', 'run tests that match the regex')
     .option(
       '-r, --reporter <reporter>',
-      `reporter to use, options are [${reporterTypes}] (default: ${defaultReporter})`,
+      `standard reporters options are [${reporterTypes}] (default: ${defaultReporter})`,
```
develohpanda: Let's change this and work with the Mocha example:

```suggestion
      `specify reporter to use (default: ${defaultReporter})`,
```

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -60,38 +86,31 @@ export async function runInsomniaTests(
...
+    reporter,
+    reporterOptions,
+    bail,
+    keepFile,
+    sendRequest,
+    testFilter: testNamePattern,
+  };
+
+  return isExternal
+    ? await runTestsCli(testFiles, config)?.catch(e => isReporterFailure(reporter, e.toString()))
```
develohpanda: This line has a mixture of `async-await`, `promise.catch` and null-coallescing inside a ternary.

There is a lot going on and in cases like these it's generally good to favor being verbose for the sake of clarity, so lets expand this to an `if-else` statement, and use a `try-catch` so the data flow is clearer at a glance. 🙂

packages/insomnia-testing/src/run/index.js
```diff
@@ -119,18 +123,38 @@ function writeTempFile(src: string): string {
...
+ * @param data An array of 'equal'(=) separated key value pairs
+ */
+function extractObject(data: Array<string>): Object {
+  const obj = {};
+  data.forEach(arg => {
+    const [key, value] = arg.split('=', 2);
+    obj[key] = value;
+  });
+  return obj;
+}
```
develohpanda: This seems like it's extracting CLI arguments for use by Mocha. But `insomnia-testing` doesn't inherently know that it is being used via the CLI, so from my understanding this logic actually belongs in `insomnia-inso`.

`insomnia-testing` shouldn't need to manipulate the reporter options before sending them to mocha.

Am I missing some context here? If so, please let me know 🙂

In any case, whether this belongs in `insomnia-testing` or `insomnia-inso` it's a perfect candidate for input/output unit tests so I'd love to see some added please!

packages/insomnia-testing/src/run/index.js
```diff
@@ -119,18 +123,38 @@ function writeTempFile(src: string): string {
...
+
 type CliOptions = InsomniaOptions & {
   reporter?: Reporter,
```
develohpanda: I believe this should now be typed as `string | Function`?

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -60,38 +86,31 @@ export async function runInsomniaTests(
...
   // Generate test file
-  const testFileContents = await generate(
-    suites.map(suite =>
-      createTestSuite(
-        suite,
-        db.UnitTest.filter(t => t.parentId === suite._id),
-      ),
-    ),
-  );
+  const testFiles = await getTestFileContent(db, suites);
```
develohpanda: This actually generates a single test file

```suggestion
  const testFile = await getTestFile(db, suites);
```

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -60,38 +86,31 @@ export async function runInsomniaTests(
     return false;
   }
 
-  // Find environment
-  const workspaceId = suites[0].parentId;
-  const environment = env
-    ? loadEnvironment(db, workspaceId, env)
-    : await promptEnvironment(db, !!ci, workspaceId);
+  // Find env
+  const environment = await getEnvironments(db, ci, suites, env);
```
develohpanda:
```suggestion
  const environment = await findEnvironment(db, ci, suites, env);
```

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -19,38 +18,65 @@ export const TestReporterEnum = {
...
+  if (err.includes('invalid reporter')) {
+    logger.fatal(`The following reporter \`${reporter}\` was not found!`);
+  } else {
+    logger.fatal(`An unknown error occurred: ${err}`);
   }
+  return false;
+}
+
+function isExternalReporter({ reporter }: RunTestsOptions): boolean {
+  return reporter && !TestReporterEnum[reporter];
```
develohpanda: Because mocha supports [these reporters](https://github.com/mochajs/mocha/blob/master/lib/reporters/index.js) internally, `TestReporterEnum` needs to be updated so that this check (`isExternalReporter`) works as expected.

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -19,38 +18,65 @@ export const TestReporterEnum = {
...
 };
 
-function validateOptions({ reporter }: RunTestsOptions): boolean {
-  if (reporter && !TestReporterEnum[reporter]) {
-    const reporterTypes = Object.keys(TestReporterEnum).join(', ');
-    logger.fatal(`Reporter "${reporter}" not unrecognized. Options are [${reporterTypes}].`);
-    return false;
+export function isReporterFailure(reporter: string, err: string): void {
+  if (err.includes('invalid reporter')) {
+    logger.fatal(`The following reporter \`${reporter}\` was not found!`);
```
develohpanda:
```suggestion
    logger.fatal(`Reporter "${reporter}" not found: ${err}`);
```

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -19,38 +18,65 @@ export const TestReporterEnum = {
...
+  } else {
+    logger.fatal(`An unknown error occurred: ${err}`);
```
develohpanda: Because this is a complete failure we don't necessarily need to prefix it with anything.

https://github.com/Kong/insomnia/blob/eff9cda1b2862863de0ce832dee48b00e454bcea/packages/insomnia-inso/src/commands/lint-specification.js#L43

```suggestion
    logger.fatal(err);
```

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -19,38 +18,65 @@ export const TestReporterEnum = {
 export type RunTestsOptions = GlobalOptions & {
   env?: string,
   reporter: $Keys<typeof TestReporterEnum>,
```
develohpanda: I believe this should be typed as string now

```suggestion
  reporter: string,
```

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -1,5 +1,4 @@
 // @flow
```
develohpanda: Could you please add tests for this file to ensure the options behave correctly? They belong here: https://github.com/Kong/insomnia/blob/eff9cda1b2862863de0ce832dee48b00e454bcea/packages/insomnia-inso/src/__tests__/cli.test.js#L152-L192

develohpanda (changes requested):
Nice work 🎉 Some comments to get this PR ready. The README also needs updating to reflect the option changes 😄

packages/insomnia-inso/package.json
MrSnix: Yes, because it supports variable arguments (varargs) natively (I use them to retrieve `reporter-options`).
I **checked their changelog** while upgrading from 5.1 to 6.1 , there were **only two breaking changes** but I didn't find any code related to them in our codebase:

> Breaking - the options property has the same case as a flag. e.g. flag -n accessed as opts().n (previously uppercase)

> Breaking - throw an error if there might be a clash between an option name and a Command property, with advice on how to resolve 

[Changelog](https://github.com/tj/commander.js/blob/master/CHANGELOG.md)

packages/insomnia-inso/src/cli.js
MrSnix: Changed

packages/insomnia-inso/src/commands/run-tests.js
MrSnix: Are you sure you want to print the stack-trace (err) even if it is a simple not found exception?

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -80,7 +80,17 @@ async function getTestFileContent(db, suites) {
...
+    const key = arg.substring(0, arg.indexOf('='));
+    let value = arg.substring(arg.indexOf('=') + 1);
+
+    // Everything is treated as a string, except:
+    if (value === 'true' || value === 'false') value = value === 'true';
+    else if (value === 'null') value = null;
+    else if (value === 'undefined') value = undefined;
+    else if (!isNaN(value)) value = Number(value);
+
+    // Assign
```
MrSnix: While I was building this function, I realized that `commander.js` treats everything as a string and doesn't do any kind of parsing on the values. 
So for example:`inso run test -r some-reporter -ro booleanKey=true` is translated as `{ booleanKey: "true" }` rather then `{ booleanKey: true }`. **I don't know** what should be the desired behaviour. I quickly looked up on some runners and I saw the options are expressed as a string and boolean mainly, so I supposed that a minimal parsing on common values like 'booleans' should be done... What do you think?

packages/insomnia-inso/src/commands/__tests__/run-tests.test.js
```diff
@@ -93,6 +85,49 @@ describe('runInsomniaTests()', () => {
     );
   });
 
+  it('should return true if reporter options are decoded properly', async () => {
```
develohpanda: Nice test!

packages/insomnia-inso/src/commands/__tests__/run-tests.test.js
```diff
@@ -93,6 +85,49 @@ describe('runInsomniaTests()', () => {
...
+        key2: 'value2',
+        path: '/tmp/9047ue.txt',
+        delimiter: '"="',
+        delimiter2: '=',
+        someBoolStuff: true,
+        someNumber: 5,
+        someNumber2: 10.52,
+      },
+      sendRequest: expect.any(Function),
+      testFilter: undefined,
```
develohpanda: Seems redundant

```suggestion
```

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -10,47 +9,106 @@ import { loadEnvironment, promptEnvironment } from '../db/models/environment';
...
-  identifier: ?string,
+  identifier?: string,
   options: RunTestsOptions,
 ): Promise<boolean> {
-  if (!validateOptions(options)) {
-    return false;
-  }
+  const { reporter, reporterOptions, ci, bail, keepFile, testNamePattern, env } = options;
+  // Loading database instance
+  const db = await loadDb(options);
```
develohpanda: Unfortunately this line change results in Flow type errors, and it should return to the previous implementation where the options object is constructed specifically for `loadDb` as `{ workingDir, appDataDir }`.

packages/insomnia-inso/src/commands/run-tests.js
develohpanda: Given Mocha [doesn't parse them](https://github.com/mochajs/mocha/blob/34643e4c0821aeb8d6977c1942bc106c9363789a/lib/cli/run.js#L195-L215) I would lean towards not parsing them as well, but I would be able to give a better answer after testing the behavior, which I wasn't able to get working

packages/insomnia-inso/src/commands/run-tests.js
```diff
@@ -66,6 +73,29 @@ async function getTestFileContent(db, suites) {
   );
 }
 
+/**
+ * Extract an object from string given a key=value format
+ * @param data An array of 'equal'(=) separated key value pairs
+ */
+function getReporterOptions(data: Array<string> = []): Object {
```
develohpanda: This looks fairly generic (and useful), might be good to move to https://github.com/Kong/insomnia/blob/develop/packages/insomnia-inso/src/util.js and add tests accordingly

develohpanda (changes requested):
Haven't done a thorough code review after your changes, but while testing this I can't seem to get my reporter options to flow through to the external reporter. Here are my steps, after checking out this branch:

```
npm run bootstrap
cd packages/insomnia-inso
npm pack
npm install -g <package>.tgz mocha@8 mochawesome
cd src/db/__fixtures__/git-repo
inso run test --reporter mochawesome --reporterOptions reportFilename=abcd json=false --env env_env_ca046a uts_fe901c
```

image

As you can see in the command, I specify that the `reportFilename` should be `abcd` and to not output a JSON report, but it appears neither of those options were followed.

Does this work for you? If so, could you show me what you've done to make this work, using the `reporter` and `reporterOptions` I'm using above?

Here are my globally installed npm packages. `insomnia-inso` has a `2.2.32-alpha.0` version number because I manually changed the version to something unique before running `npm pack`.
image

Also I think this change is complex enough, especially to be able to dev and test locally, that we might need a special section in the README for it.

MrSnix:
Hi @develohpanda, I introduced the typescript support on this PR too. 
I even addressed your requests in the last commit. As always, I'm open to any feedback.

This time I was able to make it work simply using Lerna itself... I don't know if you upgraded the dependencies and eventually the bug got fixed but this is what I did and it worked:

```
1. npm run clean
2. npm run bootstrap
3. node insomnia-inso/bin/inso run test 
    --appDataDir "/.../Application Support/Insomnia" 
    -r .../node/v12.18.3/lib/node_modules/mocha-simple-html-reporter 
    -ro output=/Users/xyz/Desktop/out.html
 
✔ Select a document or unit test suite · Swagger Petstore 1.0.2 - spc_f93e89
✔ Select an environment · OpenAPI env - env_env_59a715
The report is written in "/Users/xyz/Desktop/out.html" file.

Process finished with exit code 0
```

Proof: 
image

*Note*: 
I'm using absolute paths because I have multiple node versions and I'm running the dev version.

**About your issue:**
You cannot use `npm pack` because when you will install the tar globally it will download from the registry the insomnia-testing package that doesn't have the support for custom reporter options. So basically `insomnia-inso` gets installed as dev copy and is able to pass another reporter name than the usual one with it's options but due to using the registry copy of `insomnia-testing` it won't be able to retrieve and use correctly the reporter option parameter due to being the prod copy.

I hope it's clear.

develohpanda:
Some of the difficulties in this PR related to both inso and the custom reporter being installed from npm into the same directory, and inso not being able to see/access it.

We might be able to explore a different route or solution, now that we ship a single executable for inso and inso can be decoupled from npm.

PR #3092, opened 2021-02-17
Add plugin bundle to default config

develohpanda:
`insomnia-plugin-kong-bundle` installed by default, and shouldn't show document action or config generator plugins for collections.

Note: I have not marked `insomnia-plugin-kong-bundle` as unlisted; meaning, it will still appear on the website. Should it be unlisted?

Additionally, should the previous, manually installed `insomnia-plugin-kong-bundle`, be deleted when migrating from Designer to Core?

image


DMarby (dismissed):

nijikokun:
> Note: I have not marked `insomnia-plugin-kong-bundle` as unlisted; meaning, it will still appear on the website. Should it be unlisted?

Right now, the plugin hub doesn't respect these flags yet, so no need.

However, what I would like to do instead of unlisting them is to mark them as something like `bundled` so we can still show them on the website, perhaps in their own section, so we can still get SEO and visibility for them. Perhaps inspiration for others as well, demonstrating how core functionality is also plugin based.

> Additionally, should the previous, manually installed insomnia-plugin-kong-bundle, be deleted when migrating from Designer to Core?

Yes

develohpanda:
> However, what I would like to do instead of unlisting them is to mark them as something like `bundled` so we can still show them on the website, perhaps in their own section, so we can still get SEO and visibility for them. Perhaps inspiration for others as well, demonstrating how core functionality is also plugin based.

They already are :) https://github.com/Kong/insomnia/blob/eb4ae99f8428e99b8322e5c1b4125ddc14d316fb/plugins/insomnia-plugin-kong-bundle/package.json#L9

EDIT: oh I think you're referring to them being bundled within the app, rather than a "bundle" of plugins

develohpanda:
❗ blocked because I can't figure out why **unrelated** tests are failing

develohpanda:
❗ This is a blocker.

Inso fails because `insomnia-send-request` attempts to load all plugins that are pre-installed in Insomnia. We've pre-installed `insomnia-plugin-kong-bundle`, and inso can't find the plugin.

image

Easy, we can fix that by installing `insomnia-plugin-kong-bundle` in the `insomnia-inso` package. But then it fails because `window` cannot be found.

I think we may need to extend the networking calls used by `insomnia-send-request` to override loading plugins, because inso doesn't work with plugins right now anyway.

packages/insomnia-app/app/common/migrate-from-designer.js
```diff
@@ -103,6 +103,9 @@ async function migratePlugins(designerDataDir: string, coreDataDir: string) {
 
   await removeDirs(designerPlugins, corePluginDir);
   await copyDirs(designerPlugins, designerPluginDir, corePluginDir);
+
+  // Remove plugin bundle from installed plugins because it's included with the app now
+  await removeDirs(['insomnia-plugin-kong-bundle'], corePluginDir);
```
DMarby: Do we want to remove the individual plugins as well, in-case someone installed them directly instead of the bundle?

DMarby (approved):

packages/insomnia-app/app/common/migrate-from-designer.js
develohpanda: Nice catch, I think we do

PR #3101, opened 2021-02-22
Update onboarding flow

develohpanda:
The desired workflow is described in the parent issue: INS-407.

You can test different workflows by changing the flags in general preferences and restarting the app, or opening the migrate modal and progressing through the positive (migrate) and negative (skip) flows.

image

There aren't really any additional screenshots to add here.

### DIFF
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -111,10 +111,12 @@ export function init(): BaseSettings {
     useBulkParametersEditor: false,
     validateSSL: true,
 
-    // Feature flags
     hasPromptedToMigrateFromDesigner: false,
-    // TODO: This shouldn't default to false otherwise even users who update will be sent through onboarding
-    hasPromptedOnboarding: false,
+
+    // Users should only see onboarding during first launch, and anybody updating from an
+    // older version should not see it, so by default this flag is set to true, and is toggled
+    // to false during initialization
+    hasPromptedOnboarding: true,
   };
 }
 
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -56,6 +56,7 @@ type BaseSettings = {
   useBulkParametersEditor: boolean,
   validateSSL: boolean,
   hasPromptedToMigrateFromDesigner: boolean,
+  hasPromptedOnboarding: boolean,
 };
 
 export type Settings = BaseModel & BaseSettings;
@@ -112,6 +113,8 @@ export function init(): BaseSettings {
 
     // Feature flags
     hasPromptedToMigrateFromDesigner: false,
+    // TODO: This shouldn't default to false otherwise even users who update will be sent through onboarding
+    hasPromptedOnboarding: false,
   };
 }
 
@@ -134,6 +137,11 @@ export async function create(patch: $Shape<Settings> = {}): Promise<Settings> {
 }
 
 export async function update(settings: Settings, patch: $Shape<Settings>): Promise<Settings> {
+  return db.docUpdate(settings, patch);
+}
+
+export async function patch(patch: $Shape<Settings>): Promise<Settings> {
+  const settings = await getOrCreate();
   return db.docUpdate(settings, patch);
 }
 
--- a/packages/insomnia-app/app/models/settings.js
+++ b/packages/insomnia-app/app/models/settings.js
@@ -147,10 +147,10 @@ export async function patch(patch: $Shape<Settings>): Promise<Settings> {
   return db.docUpdate(settings, patch);
 }
 
-export async function getOrCreate(patch: $Shape<Settings> = {}): Promise<Settings> {
+export async function getOrCreate(): Promise<Settings> {
   const results = await db.all(type);
   if (results.length === 0) {
-    return create(patch);
+    return create();
   } else {
     return results[0];
   }
--- a/packages/insomnia-app/app/ui/components/settings/general.js
+++ b/packages/insomnia-app/app/ui/components/settings/general.js
@@ -551,10 +551,12 @@ class General extends React.PureComponent<Props, State> {
             <h2>Development</h2>
             <div className="form-row pad-top-sm">
               {this.renderBooleanSetting(
-                'Has prompted to migrate',
+                'Has been prompted to migrate from Insomnia Designer',
                 'hasPromptedToMigrateFromDesigner',
               )}
-              {this.renderBooleanSetting('Has prompted Onboarding', 'hasPromptedOnboarding')}
+            </div>
+            <div className="form-row pad-top-sm">
+              {this.renderBooleanSetting('Has seen onboarding experience', 'hasPromptedOnboarding')}
             </div>
           </>
         )}
--- a/packages/insomnia-app/app/ui/components/settings/general.js
+++ b/packages/insomnia-app/app/ui/components/settings/general.js
@@ -4,6 +4,7 @@ import * as fontScanner from 'font-scanner';
 import * as electron from 'electron';
 import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import {
+  ACTIVITY_MIGRATION,
   AUTOBIND_CFG,
   EDITOR_KEY_MAP_DEFAULT,
   EDITOR_KEY_MAP_EMACS,
@@ -39,7 +40,7 @@ type Props = {
   updateSetting: Function,
   handleToggleMenuBar: Function,
   handleRootCssChange: Function,
-  handleSetActivityMigration: () => void,
+  handleSetActiveActivity: () => void,
 };
 
 type State = {
@@ -109,7 +110,7 @@ class General extends React.PureComponent<Props, State> {
   }
 
   _handleStartMigration() {
-    this.props.handleSetActivityMigration();
+    this.props.handleSetActiveActivity(ACTIVITY_MIGRATION);
     this.props.hideModal();
   }
 
@@ -542,6 +543,7 @@ class General extends React.PureComponent<Props, State> {
                 'Has prompted to migrate',
                 'hasPromptedToMigrateFromDesigner',
               )}
+              {this.renderBooleanSetting('Has prompted Onboarding', 'hasPromptedOnboarding')}
             </div>
           </>
         )}
@@ -561,7 +563,7 @@ class General extends React.PureComponent<Props, State> {
 function mapDispatchToProps(dispatch) {
   const global = bindActionCreators(globalActions, dispatch);
   return {
-    handleSetActivityMigration: global.setActivityMigration,
+    handleSetActiveActivity: global.setActiveActivity,
   };
 }
 
--- a/packages/insomnia-app/app/ui/components/settings/general.js
+++ b/packages/insomnia-app/app/ui/components/settings/general.js
@@ -4,7 +4,6 @@ import * as fontScanner from 'font-scanner';
 import * as electron from 'electron';
 import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import {
-  ACTIVITY_MIGRATION,
   AUTOBIND_CFG,
   EDITOR_KEY_MAP_DEFAULT,
   EDITOR_KEY_MAP_EMACS,
@@ -40,7 +39,7 @@ type Props = {
   updateSetting: Function,
   handleToggleMenuBar: Function,
   handleRootCssChange: Function,
-  handleSetActiveActivity: (activity?: GlobalActivity) => void,
+  handleSetActivityMigration: () => void,
 };
 
 type State = {
@@ -110,7 +109,7 @@ class General extends React.PureComponent<Props, State> {
   }
 
   _handleStartMigration() {
-    this.props.handleSetActiveActivity(ACTIVITY_MIGRATION);
+    this.props.handleSetActivityMigration();
     this.props.hideModal();
   }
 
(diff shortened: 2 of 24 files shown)


packages/insomnia-app/app/models/settings.js
```diff
@@ -112,6 +113,8 @@ export function init(): BaseSettings {
 
     // Feature flags
     hasPromptedToMigrateFromDesigner: false,
+    // TODO: This shouldn't default to false otherwise even users who update will be sent through onboarding
+    hasPromptedOnboarding: false,
```
develohpanda: ❗ This is a blocker. How can we better initialize this value? It should be false if this is a fresh install of Insomnia v2021.1, or it should be true if this is an update.

packages/insomnia-app/app/ui/components/wrapper-onboarding.js
```diff
@@ -77,6 +78,7 @@ class WrapperOnboarding extends React.PureComponent<Props, State> {
 
   _handleImportFile() {
     const { handleImportFile } = this.props;
+    // TODO: this is going to blow away migrated data, needs to be smarter about forcing to current workspace
```
develohpanda: ❗ This is a blocker

packages/insomnia-app/app/ui/components/wrapper-onboarding.js
```diff
@@ -89,6 +91,7 @@ class WrapperOnboarding extends React.PureComponent<Props, State> {
       placeholder: 'https://example.com/openapi-spec.yaml',
       label: 'URI to Import',
       onComplete: value => {
+        // TODO: this is going to blow away migrated data, needs to be smarter about forcing to current workspace
```
develohpanda: ❗ This is a blocker

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -228,12 +231,50 @@ export function loadRequestStop(requestId) {
   return { type: LOAD_REQUEST_STOP, requestId };
 }
 
-export function setActiveActivity(activity?: GlobalActivity) {
-  activity = activity === DEPRECATED_ACTIVITY_INSOMNIA ? ACTIVITY_DEBUG : activity;
+function _getNextActivity(settings: Settings, currentActivity: GlobalActivity): GlobalActivity {
```
develohpanda, line 234: Many changes in this file; the main entry points are:
- `goToNextActivity()` used by sequential flows (like migration, onboarding, home) to automatically go to the "next" activity, depending on different conditions
- `setActiveActivity(activity)` used by any explicit navigation, for example the tab switcher
- `initActiveActivity()` as the name suggests, used for initialization of this redux module to ensure we launch with the expected activity

packages/insomnia-smoke-test/core/migration.test.js
```diff
@@ -0,0 +1,64 @@
...
+
+  beforeEach(async () => {
+    app = await launchApp(path.join(__dirname, '..', 'fixtures', 'basic-designer'));
+  });
+
+  afterEach(async () => {
+    await stop(app);
+  });
+
+  it('can skip migration and proceed onboarding', async () => {
```
develohpanda, line 20: Local macOS runs
Screen Shot 2021-02-22 at 8 23 36 PM

packages/insomnia-smoke-test/core/migration.test.js
```diff
@@ -0,0 +1,64 @@
...
+    await home.expectTotalDocuments(app, 1);
+    await home.expectDocumentWithTitle(app, 'Insomnia');
+
+    await app.restart();
+    await client.focusAfterRestart(app);
+
+    await home.documentListingShown(app);
+  });
+
+  it('can migrate and proceed onboarding', async () => {
```
develohpanda, line 40: Local macOS runs
Screen Shot 2021-02-22 at 9 56 52 PM

packages/insomnia-app/app/models/settings.js
develohpanda: Are we able to use notification local storage keys to initialize this value? Doesn't feel very reliable. Might need to build something in to indicate install vs update 🤔

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1127,6 +1127,41 @@ class App extends PureComponent {
     }
   }
 
+  async _handleDbChange(changes) {
```
develohpanda, line 1130: Extracted from inline-arrow function so that the listener can be removed on unmount

packages/insomnia-app/app/ui/redux/modules/__tests__/global.test.js
```diff
@@ -0,0 +1,279 @@
...
+const mockStore = configureMockStore(middlewares);
+
+const createSettings = (hasPromptedMigration: boolean, hasPromptedOnboarding: boolean) => {
+  const settings = models.settings.init();
+  settings.hasPromptedOnboarding = hasPromptedOnboarding;
+  settings.hasPromptedToMigrateFromDesigner = hasPromptedMigration;
+  return settings;
+};
+
+describe('global', () => {
```
develohpanda, line 42: These cover almost all flows of a user navigating between views. Note that the `initActiveActivity` is distinctly different from `setActiveActivity` because the `init` happens on launch and only once.

image

packages/insomnia-app/app/ui/css/components/onboarding.less
```diff
@@ -64,6 +64,8 @@
       flex-direction: column;
       padding: var(--padding-md) var(--padding-lg) var(--padding-md);
       color: var(--color-font);
+      max-height: 400px;
+      overflow: auto;
```
develohpanda, line 68: It works but this is a fairly basic solution to prevent page overflow when the window height is small and the user clicks "advanced options" on the migration page.

packages/insomnia-app/app/models/settings.js
DMarby: We have some logic in main.js for detecting whether an update happened or not, to decide whether to show the update notification, perhaps we could set this value there?

https://github.com/Kong/insomnia/blob/958ba792560dd7ef68f26206e9bef5663cf4b7ae/packages/insomnia-app/app/main.development.js#L151

develohpanda: Epic!! I should be able to use an approach inspired by this. Thank you!

packages/insomnia-app/app/ui/components/wrapper-onboarding.js
develohpanda: After discussion, during onboarding, we don't need to force import into the current workspace. If anything, we should force import into a new workspace, so as to not clutter anything existing.

This cluttering issue is more prevalent now that we have a potential migration step _prior_ to onboarding which could import x number of workspaces.

packages/insomnia-app/app/ui/components/wrapper-onboarding.js
develohpanda: Same as https://github.com/Kong/insomnia/pull/3101/files#r581441418

packages/insomnia-app/app/models/settings.js
```diff
@@ -137,10 +142,15 @@ export async function update(settings: Settings, patch: $Shape<Settings>): Promi
   return db.docUpdate(settings, patch);
 }
 
-export async function getOrCreate(patch: $Shape<Settings> = {}): Promise<Settings> {
+export async function patch(patch: $Shape<Settings>): Promise<Settings> {
+  const settings = await getOrCreate();
+  return db.docUpdate(settings, patch);
+}
+
+export async function getOrCreate(): Promise<Settings> {
```
develohpanda, line 150: `getOrCreate()` shouldn't take a `patch` parameter (like the previous implementation) because it is ignored if a model has already been created. This is quite misleading and can cause issues. Luckily, there were no usages (outside of unit tests) of `getOrCreate(patch)` so we can safely remove the parameter.

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -628,18 +673,57 @@ export function init() {
...
+
+    let overrideActivity = null;
+    switch (activeActivity) {
+      // If relaunched after a migration, go to the next activity
+      // Don't need to do this for onboarding because that doesn't require a restart
+      case ACTIVITY_MIGRATION:
+        overrideActivity = _getNextActivity(settings, activeActivity);
+        break;
+      // Always check if user has been prompted to migrate or onboard
+      default:
```
DMarby, line 713: It is possible to get stuck in the onboarding flow if `hasPromptedOnboarding` is true, but the localStorage activity key is not set (thus rendering the onboarding flow due to the == null condition in `wrapper.js`, can be reproduced by deleting the localstorage key and restarting the app), since that we the current activity is null, causing _getNextActivity to always return null, not sure if that needs to be addressed?

DMarby (approved):
Tested a bunch of the different scenarios, seems to work great!
Found one potential issue, but not sure if it needs to be addressed or not.

packages/insomnia-app/app/ui/redux/modules/global.js
develohpanda: Nice catch! I have raised a separate PR to handle invalid values at #3116

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -536,24 +536,28 @@ class General extends React.PureComponent<Props, State> {
...
+          </button>
+        </div>
+        <hr className="pad-top" />
+
         {isDevelopment() && (
           <>
+            <h2>Development</h2>
             <div className="form-row pad-top-sm">
               {this.renderBooleanSetting(
                 'Has prompted to migrate',
```
nijikokun:
```suggestion
                'Has been prompted to migrate from Insomnia Designer',
```

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -536,24 +536,28 @@ class General extends React.PureComponent<Props, State> {
...
                 'hasPromptedToMigrateFromDesigner',
               )}
+              {this.renderBooleanSetting('Has prompted Onboarding', 'hasPromptedOnboarding')}
```
nijikokun:
```suggestion
              {this.renderBooleanSetting('Has seen onboarding experience', 'hasPromptedOnboarding')}
```

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -536,24 +536,28 @@ class General extends React.PureComponent<Props, State> {
...
+            Migrate
+          </button>
+        </div>
+        <hr className="pad-top" />
+
         {isDevelopment() && (
           <>
+            <h2>Development</h2>
             <div className="form-row pad-top-sm">
               {this.renderBooleanSetting(
```
nijikokun, line 550: Let's make each flag an individual row.

nijikokun (changes requested):

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -536,24 +536,28 @@ class General extends React.PureComponent<Props, State> {
...
 
         <h2>Migrate from Designer</h2>
+        <div className="form-row pad-top-sm">
+          <button
+            className="btn btn--clicky pointer"
+            style={{ padding: 0 }}
+            onClick={this._handleStartMigration}>
+            Migrate
+          </button>
+        </div>
```
nijikokun, line 543: Can we make the button `inline-block` rather than `full-width`?

packages/insomnia-app/app/ui/components/wrapper-onboarding.js
```diff
@@ -89,7 +89,7 @@ class WrapperOnboarding extends React.PureComponent<Props, State> {
       placeholder: 'https://example.com/openapi-spec.yaml',
       label: 'URI to Import',
       onComplete: value => {
-        handleImportUri(value, ForceToWorkspaceKeys.current);
+        handleImportUri(value, ForceToWorkspaceKeys.new);
```
nijikokun, line 92: 🙌

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -228,8 +231,52 @@ export function loadRequestStop(requestId) {
   return { type: LOAD_REQUEST_STOP, requestId };
 }
 
-export function setActiveActivity(activity?: GlobalActivity) {
-  activity = activity === DEPRECATED_ACTIVITY_INSOMNIA ? ACTIVITY_DEBUG : activity;
+function _getNextActivity(settings: Settings, currentActivity: GlobalActivity): GlobalActivity {
+  switch (currentActivity) {
+    case ACTIVITY_MIGRATION:
+      // Have we the onboarding step? Go to home, otherwise go to onboarding
```
nijikokun:
```suggestion
      // Have we shown the onboarding step? Go to home, otherwise go to onboarding
```

packages/insomnia-app/app/ui/components/settings/general.js
develohpanda: image

packages/insomnia-app/app/ui/components/settings/general.js
develohpanda: Looks a little strange that way as well.

image

We could also move it under the data tab as a single button that says "Migrate from Designer".

image

@nijikokun / @erickrico let me know where you'd like this to move to :)

develohpanda:
Updated screenshot 
image

DMarby (approved):

nijikokun (approved):

PR #3106, opened 2021-02-22
Onboarding flow ent to end tests and automatic screenshots on failure

develohpanda:
This PR introduces smoke tests for the following two workflows. Each has run reliably 20 times in a row. Because the migration and onboarding flows are first time, it also restarts the application to ensure the user proceeds to the expected page.

In addition, this PR extends the smoke testing capability by taking and uploading a screenshot for each test failure automatically. It also provides the capability to taking screenshots explicitly, if necessary for debugging, for instance.

## Can skip migration and onboarding
Screen Shot 2021-02-22 at 8 23 36 PM

## Can proceed through migration and onboarding
Screen Shot 2021-02-22 at 9 56 52 PM

Closes INS-493


.github/workflows/test.yml
```diff
@@ -30,11 +30,16 @@ jobs:
...
-#        run: npm run test:smoke:build
+      - name: Build for smoke tests
+        run: npm run app-build:smoke
+      - name: Run smoke tests
+        timeout-minutes: 10 # sometimes jest fails to exit - https://github.com/facebook/jest/issues/6423#issuecomment-620407580
+        run: npm run test:smoke:build
+      - name: Upload smoke test screenshots
+        uses: actions/upload-artifact@v2
+        if: always()
+        with:
```
DMarby, line 41: Probably makes sense to mute the warnings for no files found, since that's currently expected
```suggestion
        with:
          if-no-files-found: ignore
```

packages/insomnia-smoke-test/core/app.test.js
```diff
@@ -6,7 +6,7 @@ import * as settings from '../modules/settings';
 import fs from 'fs';
 import { basicAuthCreds } from '../fixtures/constants';
 
-describe('Application launch', function() {
+xdescribe('Application launch', function() {
```
DMarby, line 9: Probably good to not leave this in, in-case more tests are added to this file in the future

packages/insomnia-smoke-test/designer/app.test.js
```diff
@@ -10,7 +9,7 @@ import { isPackage, launchApp, stop } from '../modules/application';
 const itIf = condition => (condition ? it : it.skip);
 it.if = itIf;
 
-describe('Application launch', function() {
+xdescribe('Application launch', function() {
```
DMarby, line 12: Probably good to not leave this/the `xit` below in, in-case more tests are added to this file in the future

packages/insomnia-smoke-test/modules/application.js
```diff
@@ -52,10 +59,9 @@ const launch = async config => {
     // Don't remove chromeDriverArgs
     // https://github.com/electron-userland/spectron/issues/353#issuecomment-522846725
     chromeDriverArgs: ['remote-debugging-port=9222'],
+    // waitTimeout: 40000,
```
DMarby: Should this be uncommented or removed?

packages/insomnia-smoke-test/designer/app.test.js
```diff
@@ -22,18 +21,7 @@ describe('Application launch', function() {
...
-
-    await onboarding.welcomeMessageShown(app);
-    await onboarding.clickDontShare(app);
-    await onboarding.clickSkipImport(app);
-
-    await home.documentListingShown(app);
-  });
-
-  it.if(isPackage())('can install and consume a plugin', async () => {
+  xit.if(isPackage())('can install and consume a plugin', async () => {
```
DMarby, line 24: Unrelated to this PR/change, no need to address it here, but I guess this test doesn't really verify that the plugin is installed correctly anymore, now that the plugin bundle is installed by default?

DMarby (approved):
Awesome 🎉 Few minor nitpicks/questions

packages/insomnia-smoke-test/core/app.test.js
develohpanda: There's a linear ticket to update these INS-416, the intention is to fix the tests and un-ignore them

packages/insomnia-smoke-test/designer/app.test.js
develohpanda: Same as above

packages/insomnia-smoke-test/modules/application.js
develohpanda: Needs to be removed, sneaky bugger

packages/insomnia-smoke-test/designer/app.test.js
develohpanda: Hmmm yes, I suppose this can be changed to just check it's installed. But I'd still like to keep a plugin installation test around so I'll update this to work with a different plugin

packages/insomnia-smoke-test/modules/application.js
develohpanda:
```suggestion

```

packages/insomnia-smoke-test/core/app.test.js
DMarby: Righto, makes sense 👍

PR #3111, opened 2021-02-23
SSH/SCP/HTTPS git URLs with/without .git suffix; update isomorphic-git

dimitropoulos:
to reproduce the changes in this PR, make a git repo and clone it the 6 following ways and observe that all git functions work as expected.

1. `git@github.com:<your username>/<your repo>`
1. `git@github.com:<your username>/<your repo>.git`
1. `http://github.com/<your username>/<your repo>`
1. `http://github.com/<your username>/<your repo>.git`
1. `https://github.com/<your username>/<your repo>`
1. `https://github.com/<your username>/<your repo>.git`

closes: INS-464

I would love a pointer or two on how best to write some unit testing for this in the context of remote repositories being involved (I guess, then, it'd really be an end-to-end test at that point).


develohpanda:
> I would love a pointer or two on how best to write some unit testing for this in the context of remote repositories being involved (I guess, then, it'd really be an end-to-end test at that point).

We explored this during the last iteration of changes to git sync, but decided against it because ultimately it is the responsibility of `isomorphic-git` to ensure it can clone from various providers. On exploring their source code I found they _do_ have integration tests for this, so we didn't pursue it further.

We can certainly explore writing smoke tests to ensure Insomnia behaves as expected on success and failure when cloning various types of repositories (ideally integration tests but to my knowledge they will require more refactoring than we have time for), but any deeper and we're testing 3rd-party code.

packages/insomnia-app/app/sync/git/fs-plugin.js
```diff
@@ -3,71 +3,40 @@ import fs from 'fs';
...
+  | fs.FSPromise.unlink
+  | fs.FSPromise.readdir
+  | fs.FSPromise.mkdir
+  | fs.FSPromise.rmdir
+  | fs.FSPromise.stat
+  | fs.FSPromise.lstat
+  | fs.FSPromise.readlink
+  | fs.FSPromise.symlink;
+
+export const fsPlugin = (basePath: string) => {
```
develohpanda: Are there any usages of `fsPlugin` which rely on the `basePath` parameter default of `/`?

packages/insomnia-app/app/sync/git/fs-plugin.js
```diff
@@ -3,71 +3,40 @@ import fs from 'fs';
...
-  }
-
-  async symlink(targetPath: string, filePath: string, ...x: Array<any>) {
-    return this._callbackAsPromise(fs.symlink, filePath, ...x);
-  }
-
-  _callbackAsPromise<T>(fn: Function, filePath: string, ...args: Array<any>): Promise<T> {
-    return new Promise((resolve, reject) => {
-      filePath = path.join(this._basePath, path.normalize(filePath));
-      const callback = args.find(arg => typeof arg === 'function');
```
develohpanda, line 64: Eeeek, `callback` had no usage? Why do you think that was the case? I genuinely don't know, but it seems a little strange...

packages/insomnia-app/app/sync/git/http-plugin.js
```diff
@@ -0,0 +1,34 @@
+import { axiosRequest } from '../../network/axios-request';
+
+export const http = {
```
develohpanda: Comment is lost when translating `packages/insomnia-app/app/sync/git/http.js` to this.

For others reviewing, the http plugin API has changed and requires the export to be an object containing a function property named `request`, hence this update.

packages/insomnia-app/app/sync/git/routable-fs-plugin.js
```diff
@@ -31,7 +31,9 @@ export function routableFSPlugin(defaultFS: Object, otherFS: { [string]: Object
     return result;
   };
 
-  const methods = {};
+  const methods = {
+    enumerable: true,
```
develohpanda: To my understanding this is added specifically because of a nuance in `isomorphic-git`? If so, please add a comment to explain why. With no context it seems like a magical property/value.

packages/insomnia-app/app/sync/git/utils.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+export const translateSSHtoHTTP = (url: string) => {
+  // handle "shorter scp-like syntax"
+  url = url.replace(/^git@([^:]+):/, 'https://$1/');
+  // handle proper SSH URLs
+  url = url.replace(/^ssh:\/\//, 'https://');
```
develohpanda, line 6: Would it be worth providing a fully-formed example of the URL syntax each of these regex's is handling? As a bonus, it would be great to see those examples as unit tests rather than code comments 🎉

packages/insomnia-app/app/sync/git/utils.js
```diff
@@ -0,0 +1,16 @@
...
+  return url;
+};
+
+export const addDotGit = ({ url }: { url: string }) => {
```
develohpanda, line 10: `{ url }: { url: string }` JS 🤕

packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal.js
```diff
@@ -158,7 +158,6 @@ class GitRepositorySettingsModal extends React.PureComponent<Props, State> {
                 <input
                   required
                   ref={this._setInputRef}
-                  type="url"
```
develohpanda, line 161: Does the absence of this mean we need to replace it with custom validation? Feels like we should, although not a blocker.

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,59 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
+        const fs = MemPlugin.createPlugin();
+
+        repoSettingsPatch.uri = translateSSHtoHTTP(repoSettingsPatch.uri);
 
         // Pull settings returned from dialog and shallow-clone the repo
-        const { credentials, uri: url } = repoSettingsPatch;
+        const {
+          credentials: { username, token: password },
+          uri: url,
+        } = repoSettingsPatch;
```
develohpanda: A matter of personal preference but I'm _generally_ not a fan of aliases in this manner because it can be difficult to follow when you know the shape of models you are working with but the names don't match up in your code.

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,59 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
-            core,
-            dir: GIT_CLONE_DIR,
-            gitdir: GIT_INTERNAL_DIR,
-            singleBranch: true,
-            url,
-            ...credentials,
-            depth: 1,
-            noGitSuffix: true,
-          });
+          // TODO: see if there's a way to retry
```
develohpanda: If we don't intend to do this in this PR I suggest at least creating a Linear ticket, and then either linking that ticket in the comment or removing the comment altogether.

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,59 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
         } catch (err) {
+          if (!cloneParams.url.endsWith('.git')) {
+            try {
+              await git.clone({
+                ...cloneParams,
+                url: addDotGit(cloneParams),
+              });
+
+              // by this point the clone was successful, so update with this syntax
+              repoSettingsPatch.uri = url;
```
develohpanda: Shouldn't we update _with_ the dot git?

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,59 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
+              // TODO: verify there's not a way to combine these
+              showError({
+                title: 'Error Cloning Repository with .git',
+                message: error.message,
+                error,
+              });
+            }
+          }
+
           showError({ title: 'Error Cloning Repository', message: err.message, error: err });
```
develohpanda: This flow is going to show two errors, right 🤔

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,59 @@ class WrapperHome extends React.PureComponent<Props, State> {
 
         const core = Math.random() + '';
 
-        // Create in-memory filesystem to perform clone
-        const plugins = git.cores.create(core);
-        const fsPlugin = MemPlugin.createPlugin();
-        plugins.set('fs', fsPlugin);
+        const fs = MemPlugin.createPlugin();
```
develohpanda: I feel strongly that it's not a good idea to create a variable which is the same as a fairly common base node module. _Even if_ it has the same API, it is a proxy to the base `fs` module and can be a source of future (and current - when I was reviewing this PR) confusion.

develohpanda (changes requested):
A few observations and comments from a preliminary review. I haven't had a chance to thoroughly go through the main `git-vcs.js` file yet!

packages/insomnia-app/app/sync/git/__tests__/git-vcs.test.js
```diff
@@ -58,12 +58,11 @@ describe('Git-VCS', () => {
       expect(await vcs.status(fooTxt)).toBe('*added');
     });
 
-    it('Returns empty log without first commit', async () => {
+    it.only('Returns empty log without first commit', async () => {
```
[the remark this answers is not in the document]
dimitropoulos: eek..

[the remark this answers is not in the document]
dimitropoulos: I will admit, I'm fairly used to having https://github.com/jest-community/eslint-plugin-jest/blob/HEAD/docs/rules/no-disabled-tests.md turned on and not having to spend brainpower to worry/think about this.  Maybe we can add it?

packages/insomnia-app/app/sync/git/fs-plugin.js
dimitropoulos: no.  not according to the only two calls I am aware of (both in `app.js`) which `path.join` literal strings.

packages/insomnia-app/app/sync/git/fs-plugin.js
dimitropoulos: it's the kind of thing I think you think you'll need when making a callback->promise converter like this, but in this case happened to never be used.  one way or the other, this function isn't called anywhere else in any context so far as I can see so it's fine to remove dead code.

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -187,21 +239,22 @@ export default class GitVCS {
...
     return {
       name: name || '',
       email: email || '',
     };
   }
 
   async setAuthor(name: string, email: string): Promise<void> {
-    await git.config({ ...this._baseOpts, path: 'user.name', value: name });
-    await git.config({ ...this._baseOpts, path: 'user.email', value: email });
+    console.log({ ...this.baseOpts, name, email });
```
DMarby: Is this just leftover debug logging?

packages/insomnia-app/app/sync/git/http-plugin.js
dimitropoulos: this is good feedback, anyway.  I've added descriptions for both clients (they're called clients now, I guess, not plugins).

packages/insomnia-app/app/sync/git/routable-fs-plugin.js
dimitropoulos: yes, it's in the documentation, but it's absolutely required because it's the thing that tells isomorphic-git that we're using the promise API:

> There are actually TWO possible interfaces for an fs object: the classic "callback" API and the newer "promise" API. If your fs object provides an enumerable promises property, isomorphic-git will use the "promise" API exclusively.

I added a comment on this.

packages/insomnia-app/app/sync/git/utils.js
dimitropoulos: it's done this way to be more ergonomic for how we're actually using it (which is, in all cases, with objects that have a `url` property, e.g. `cloneParams`).

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: I agree with you, actually, but the authors of isomorphic-git do not.  that's why it was done this way.  they expose a property called `fs`.  I'd alias it to `fsClient` but [above](https://github.com/Kong/insomnia/pull/3111#discussion_r598453398) you said you don't want alias.  How would you like me to proceed?

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: removed in 786b35d9fac8eaa89175e858064549ce2ac15b25 as it turns out that the way they suggest we do this is via onAuthFailure, which I think is a level of indirection we really don't need here.  it's much clearer to do it the imperative way.

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: ok, I tried to update to match your personal preference in 78bb66b160b2af501b14386f9ccc7a266667c863 lemme know if that wasn't what you were asking for.

packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal.js
dimitropoulos: I think the answer is no, actually, we don't want validation here.  There are just so many ways to specify git urls and we want to allow users to at least be able to try using their method.  With this validation in place, the scp-style that this PR implements cannot be used, for example.  If we tried to write our own regex, I think we would miss use-cases pretty quickly.  After all, if the url is invalid they're going to get a clear error message telling them so when isomorphic-git tries it.

packages/insomnia-app/app/sync/git/utils.js
dimitropoulos: added unit tests here f093413030037612ee5f72d60f26ed1e6725229c

packages/insomnia-app/app/sync/git/routable-fs-plugin.js
dimitropoulos: added in 855d4b103337801faa7f62726a20094a7e7bb991

packages/insomnia-app/app/sync/git/http-plugin.js
dimitropoulos: implemented in db508782549cf4627cc32ceef4b9913fe6b9f35d

packages/insomnia-app/app/sync/git/__tests__/git-vcs.test.js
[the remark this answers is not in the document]
dimitropoulos: removed in ea35b04287335cae17cc6ecd7a999fab01a500a1

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: why? I feel like I would be frustrated if I put something that didn't have the .git and it got changed (for what I see) to have the .git.  Ultimately, this is an implementation detail on our part, isn't it?

Don't get me wrong, more than happy to make that change, but I'd like to understand the rationale.

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -48,83 +74,109 @@ export type PushResponse = {
...
-
-    this._baseOpts = {
+  }) {
+    this.baseOpts = {
       dir: directory,
       gitdir: gitDirectory,
-      noGitSuffix: true,
+      fs,
+      http,
+      onAuth: onAuth(gitCredentials),
```
DMarby: Do we want to pass `onMessage` here as well, to match `init`?

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -48,83 +74,109 @@ export type PushResponse = {
...
+      ...this.baseOpts,
+      onAuthFailure,
+      onAuthSuccess,
+      url,
+      singleBranch: true,
+    };
+    try {
+      await git.clone(cloneParams);
+    } catch (error) {
+      if (!url.endsWith('.git')) {
```
DMarby: I'm not sure I think this belongs here, since the call-site of `initFromClone` should pass the URL from the `GitRepository` model, which should have gotten the correct URL set during setup, but that's related to 
the discussion in https://github.com/Kong/insomnia/pull/3111#discussion_r598454386 😅

packages/insomnia-app/app/ui/components/wrapper-home.js
DMarby: I also think it makes more sense to update it to the URL with dot git here, so that for one, it's also consistent with how we handle the URLs for the ssh syntax translation.
It also means that we don't need to duplicate logic for it, and can rely on the URL in the database being something `isomorphic-git` can handle, rather than needing to always check and account for it.

Currently, we have duplicate logic between here and `initFromClone`, and also a potential edge-case if this codepath get triggered here by an error from the server, but for some reason doesn't trigger in `initFromClone`, with unknown outcomes.

packages/insomnia-app/app/ui/components/wrapper-home.js
DMarby: From my testing it only shows the `Error Cloning Repository` error, never the other one, so that gets swallowed, which doesn't seem right, but I might be missing something.

I think the flow should be as follows:
1. If we fail to clone a repo, try with .git appended to the URL
2. If the appended .git URL clone also fails, show a single modal with the errors from *both* attempts, as otherwise we might cover up eg an initial 401 from the first one, and a 404 for the second one, in the case of a repo URL that should not have `.git`.

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -379,9 +430,9 @@ export default class GitVCS {
     }
   }
 
-  async _repoExists() {
+  async repoExists() {
     try {
-      await git.config({ ...this._baseOpts, path: '' });
+      await git.setConfig({ ...this.baseOpts, path: '' });
```
DMarby: I think we could potentially use `getConfig` instead, to be guaranteed to be non-destructive? But not sure

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,53 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
-        const { credentials, uri: url } = repoSettingsPatch;
+        const cloneParams = {
+          fs,
+          http,
+          core,
+          dir: GIT_CLONE_DIR,
+          gitdir: GIT_INTERNAL_DIR,
+          singleBranch: true,
+          url,
+          onAuth: () => ({
```
DMarby: Might make sense to pass the onMessage/onAuthFailure etc hooks here as well, so we get better logs if something goes wrong? Or would that mean we don't get an error thrown?

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,53 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
-            gitdir: GIT_INTERNAL_DIR,
-            singleBranch: true,
-            url,
-            ...credentials,
-            depth: 1,
-            noGitSuffix: true,
-          });
+          await git.clone(cloneParams);
         } catch (err) {
+          if (!cloneParams.url.endsWith('.git')) {
```
DMarby: I'm a bit uncertain about how to test this logic, GitHub in particular seems to correctly listen to a URL ending both with `.git` and one without for my test repo now, I was able to input a URL without `.git` even on the release version of our app. 
Not sure if I'm missing anything, but did we ever test that this failed previously, or have GitHub perhaps changed their behaviour?

Not sure if we still need this logic for other providers, what's your thoughts @nijikokun?

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,53 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
+            try {
+              await git.clone({
+                ...cloneParams,
```
DMarby: I think we should use a new instance of `MemPlugin` when retrying here, I saw errors about "the remote already existing" from `isomorphic-git` when giving the URL for a repo that didn't exist, so I'm not sure this logic actually works at all currently?

DMarby (changes requested):

packages/insomnia-app/app/ui/components/wrapper-home.js
nijikokun: > Not sure if we still need this logic for other providers, what's your thoughts @nijikokun?

Is there any harm in leaving it in? Seems like it could provide potential saving grace for providers that *require* `.git` that we aren't aware of.

DMarby: > Is there any harm in leaving it in? Seems like it could provide potential saving grace for providers that _require_ `.git` that we aren't aware of.

I don't think there's any particular harm, so sounds reasonable to leave it in to cover those cases to me.

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: ugg.. yes.. I'm sorry.  will remove.

packages/insomnia-app/app/sync/git/__tests__/git-vcs.test.js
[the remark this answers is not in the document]
develohpanda: Let's do it! Looks like a good rule to have

packages/insomnia-app/app/sync/git/fs-plugin.js
develohpanda: 👍🏽 will double-check as well.

packages/insomnia-app/app/ui/components/wrapper-home.js
develohpanda: Don't think you're missing anything. It was either two errors or one gets swallowed because the `showError` utility always shows the same instance of the `ErrorModal`.

I think we should actually only show the first error. The second attempt (with the .git suffix) is something we do internally, transparent to the user.

If the .git suffix succeeds, we update the URL (as per a previous convo), and if it fails, the user doesn't need to be prompted about it because it's not something they did. (But maybe we should still log into console, as well as a `console.log('[git] attempting clone with .git suffix')` message).

DMarby: Even tho it is something we do transparently, I don't think showing just the first (or the second) error is ideal, as that would cover up valid errors in certain cases. 
Showing just the second error would cause problems in the scenario mentioned in my previous comment, and showing just the first error would be a problem if for example you get a 404 for a non .git URL, and then a 401/403 on the correct URL when .git was automatically appended, because your token is invalid, as getting an error saying 404 is not helpful for debugging that.

Since there are so many potential edge cases, I'd prefer exposing both errors to the user rather than having them need to go look in the logs/console.

nijikokun: > Showing just the second error would cause problems in the scenario mentioned in my previous comment, and showing just the first error would be a problem if for example you get a 404 for a non .git URL, and then a 401/403 on the correct URL when .git was automatically appended, because your token is invalid, as getting an error saying 404 is not helpful for debugging that.

Showing both sounds good, we do capture both, I'm sure we can come up with a way to word both into a single sentence.

Something like:

> Error cloning repository, attempts with and without `.git` suffix ended in failure:

develohpanda: > if for example you get a 404 for a non .git URL, and then a 401/403 on the correct URL when .git was automatically appended, because your token is invalid

Ah okay, yes, this is a totally valid situation and I agree.

In that case we can probably skip sending a stack trace to `showError` (because we have two stacks and they don't provide much value) but combining the messages would be the play. 👏🏽

packages/insomnia-app/app/sync/git/utils.js
develohpanda: Nice tests!

packages/insomnia-app/app/sync/git/utils.js
develohpanda: Ya fair enough. Just feels unnecessary as we're only using one property (and are unlikely to ever use anything else for this utility). I've not done it this way in this codebase because flow never detected errors.

IMO the simpler option would be to just make it really dumb and accept only a string, leave it up to the consumers to send the right value, decoupling this utility from requiring an object with a particular shape. Anyway not a strong opinion because TS will cater for it better in future.

packages/insomnia-app/app/ui/components/wrapper-home.js
develohpanda: Because the URL can never be changed once git sync is configured. If we _know_ that omitting the `.git` suffix fails and adding it works, we should patch the model and add the suffix to prevent a try-retry the next time it needs to be used. Agree with the points raised above.

packages/insomnia-app/app/ui/components/wrapper-home.js
develohpanda: I prefaced with saying it's personal preference because I'm open to hearing other views on it, not because I expect others to code as I would. 🙂 

To explain further, naturally it's situational but from my experiences (and why it's become a preference),
```js
const { foo, bar } = item;
...
someFunc({ bar, baz: foo}); 
```
is easier to follow and debug than
```js
const { foo: baz, bar } = item;
...
someFunc({ bar, baz}); 
```
_especially_ when the rename while destructuring (`foo` to `baz`) can happen far away from where it's used and several levels deep (as is the case with the original change in this PR). In most situations it produces harder to follow code and if it's avoidable IMO avoid it.

To address https://github.com/Kong/insomnia/commit/78bb66b160b2af501b14386f9ccc7a266667c863, my feedback was around the alias/destructured rename rather than destructuring itself, and this has rebounded a bit further than I anticipated so hopefully my explanation above offers some clarification. 😅 Regardless, open to hearing other opinions.

packages/insomnia-app/app/ui/components/wrapper-home.js
develohpanda: I'd argue that isomorphic-git's use of `fs` is different to ours and their use of the `fs` variable is justified. They allow consumers to override the concept of a "file-system" for the entire package, and regardless of the client implementation it _is_ the `fs` for isomorphic-git.

In our case, we prepare and provide a proxy for isomorphic-git (actually two: `MemPlugin` and `NeDbPlugin`) and also use the real thing in other areas of the code, and should not name the proxy the same as the real thing.

A good example of this is the `NeDbPlugin`. This is also a file-system plugin for isomorphic-git and follows the same interface, but it would be completely unreasonable to name it `fs` in our code when we also use the real `fs` module, and especially given it doesn't behave the same was as `fs` does.

It would be better to avoid the confusion that comes with having `fs` and a proxy of `fs`, both be called `fs`, so `fsClient` sounds good.

(I also clarified the use of destructured renaming/aliasing further above and I hope that clarifies why it's not the case here)

packages/insomnia-app/app/sync/git/fs-plugin.js
```diff
@@ -3,71 +3,42 @@ import fs from 'fs';
...
+
+  const wrap = (fn: FSWraps) => async (filePath: string, ...args: Array<any>): Promise<T> => {
+    const modifiedPath = path.join(basePath, path.normalize(filePath));
+    return fn(modifiedPath, ...args);
+  };
+
+  return {
+    promises: {
+      /** this `enumerable` property is crucial, because it is the mechanism for telling isomorphic-git to exclusively use the promise API. {@link https://isomorphic-git.org/docs/en/fs#implementing-your-own-fs} */
+      enumerable: true,
```
develohpanda: Actually, backtracking here, the `enumerable` property should not be needed in this client/plugin. The [source code](https://github.com/isomorphic-git/isomorphic-git/blob/f2e38053733f4c8ca4b7784d5c2a56addcf36bf4/src/models/FileSystem.js#L13-L14) in `isomorphic-git` uses `Object.getOwnPropertyDescriptor` to get the property, and that descriptor contains the [four data attributes of a property](https://www.javascripttutorial.net/javascript-object-properties/): `value`, `configurable`, `writable` and `enumerable`.

I tested this and voila!

image

Today I learned!

It probably didn't work before because this was a class with methods, rather than an object with function properties. This raises the question about how `routableFsPlugin` behaves and whether we should be setting the `enumerable` property manually on `routableFsPlugin` at all or actually adjusting the implementation.

packages/insomnia-app/app/ui/components/wrapper-home.js
develohpanda: Worth testing what it does but I think this is a good idea as well.

packages/insomnia-app/app/ui/components/wrapper-home.js
develohpanda: > Not sure if I'm missing anything, but did we ever test that this failed previously, or have GitHub perhaps changed their behaviour?

As far as I can recall it did used to fail 🤔 I agree as well though, probably worth keeping it around

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -26,15 +26,41 @@ type GitCredentialsToken = {
   token: string,
 };
 
+const onAuth = ({ username, token }: GitCredentialsToken = {}) => () => ({
```
develohpanda: Should this be typed as `GitCredentials` (defined further down in this file)?

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -26,15 +26,41 @@ type GitCredentialsToken = {
...
-  tree: string,
-  author: GitAuthor & {
-    timestamp: number,
+  commit: {
+    message: string,
+    tree: GitRef,
+    author: GitAuthor & {
+      timezoneOffset: number,
+      timestamp: number,
+    },
```
develohpanda: Seems like these properties should go into a new `GitTimestamp` type definition?

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -48,83 +74,109 @@ export type PushResponse = {
...
+const gitInternalDirName = 'git';
 export const GIT_INSOMNIA_DIR_NAME = '.insomnia';
 
-export const GIT_INTERNAL_DIR = path.join(GIT_CLONE_DIR, _gitInternalDirName);
+export const GIT_INTERNAL_DIR = path.join(GIT_CLONE_DIR, gitInternalDirName);
 export const GIT_INSOMNIA_DIR = path.join(GIT_CLONE_DIR, GIT_INSOMNIA_DIR_NAME);
 
-export default class GitVCS {
-  _git: Object;
-  _baseOpts: { dir: string, gitdir?: string, noGitSuffix: boolean };
```
develohpanda, line 59: The general pattern in this code-base is to prefix class properties and methods that are intended to be private with `_`, not sure why this was changed

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -48,83 +74,109 @@ export type PushResponse = {
...
     directory: string,
-    fsPlugin: Object,
+    fs: Object,
     gitDirectory: string,
-  ) {
-    this._git = git;
-    git.plugins.set('fs', fsPlugin);
-
-    this._baseOpts = {
+  }) {
```
develohpanda: An inline type definition doesn't really help with readability. Can we make these explicit types (even if single-use)?

packages/insomnia-app/app/sync/git/git-vcs.js
develohpanda: Nice catch

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -271,13 +328,15 @@ export default class GitVCS {
...
     console.log('[git] Pull remote=origin', await this.getBranch());
     trackEvent('Git', 'Pull');
 
     return git.pull({
-      ...this._baseOpts,
-      ...creds,
+      ...this.baseOpts,
+      onAuth: onAuth(gitCredentials),
+      onAuthFailure,
+      onAuthSuccess,
```
develohpanda: I think creating a `getAuthCallbacks(credentails)` utility function might be useful to serve the same purpose as `this._baseOpts` and avoid the manual repetition.

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -310,34 +371,24 @@ export default class GitVCS {
...
-    }
-
-    if (err && err.code === 'ResolveRefError') {
-      return [];
-    }
-
-    if (err) {
-      throw err;
-    } else {
-      return log;
```
develohpanda, line 329: Wowee this was something special 🤣 Nice refactor

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: cool cool, thanks for the explanation.  updated in 40a3d3af90aff3628583e76b7ad094e5facf6088

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: lemme know if c944f893f80abd93899c425365f4546561a1922e is what you all have in mind.

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: fixed in a more systematic way in 7cb61c53b2329f88d861c1d505fcb94c5ec60387

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: 2abb9cc4362a5a5a2d692349b273aa7e01de8ca3 changed this and one other that was nearly identical in usage (`getAuthor`)

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: 6f9d009b75fcc796271fce5c28698e9351f4ad7d

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: sure, I'll note that I don't agree here, but that's fine - no need to.

The reason is that for these one-off use-cases it can be tough to know what the types are without going to track down the backing type definition object.  Even for things that may seem obvious.  For example, in one of these `directory` is optional and in the other it isnt:

```ts
  async init({ directory, fs, gitDirectory }: InitOptions) {
```

```ts
  async initFromClone({ url, gitCredentials, directory, fs, gitDirectory }: InitFromCloneOptions) {
```

When the types are right there, you can see it without having to hover your mouse or go find a type.

```ts
  async init({
    directory,
    fs,
    gitDirectory,
  }: {
    directory?: string,
    fs?: Object,
    gitDirectory?: string,
  }) {
```

```ts
  async initFromClone({
    url,
    gitCredentials,
    directory,
    fs,
    gitDirectory,
  }: {
    url: string,
    gitCredentials: GitCredentials,
    directory: string,
    fs: Object,
    gitDirectory: string,
  }) {
```

dimitropoulos: fixed in cd1f28070c1dd4b1f1e363cf99307339bbe9f48a

develohpanda: Fair enough, I see your point of view 🙂

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: totally agree.  before I saw your comment I did https://github.com/Kong/insomnia/pull/3111#discussion_r600061885 which had a simiar effect of removing these.  now that I think about it, though, a function that returns an object we could spread in that takes an optional `credentials` is better.  Lemme try to do that real quick.

dimitropoulos: I implemented this (it contains `onMessage` as well, which behaves very similarly so I named it `gitCallbacks`) in 6152dd5efabe0c2cf61a6b9100d88d457aa83215

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: because they're not actually private, that's the only reason.  I'd be happy to change it back if you'd like.  I almost always make this change because (and it's not hard to find examples) inevitably these things leak out anyway since they're not actually private.

once we're in typescript we can mark this as private and have the type system enforce it, or use the https://www.typescriptlang.org/docs/handbook/classes.html#ecmascript-private-fields which will absolutely guarantee that it's private.

dimitropoulos: in order to move this PR along faster, I reverted this change in 3aa41ec0228ce0803c4dcad458e90392b0696dec

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: implemented in 5219f9aa21d021845cb505bd67438bed508a21a0

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: it was removed as part of b667a4faa85780a3ea88d8325eeffa7d38148605

packages/insomnia-app/app/sync/git/git-vcs.js
develohpanda: Thanks for reverting! I'm on the same page with you but given we don't have a good alternative (until we're in TS) I'm inclined to keep as-is. It also reduces the change-set so that's a win!

packages/insomnia-app/app/sync/git/fs-plugin.js
dimitropoulos: I'm skeptical, since the docs say it must be there and also because disabling it here doesn't change that it was actually also added in `routableFsPlugin` (not `fsPlugin` where your screenshot is from).  I implemented what you're asking for in cd34d1591174f50fe7aee140233b23121cd703ac but I'd like you to take another look, please, to verify that the test you ran above still includes `enumerable: true` after this change (or, tell me the steps you took so I can run the same test).

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: check out b667a4faa85780a3ea88d8325eeffa7d38148605.  should be all set now.

packages/insomnia-app/app/sync/git/fs-plugin.js
develohpanda: Will do! I couldn't find it in the docs, could you maybe screenshot where it asks for an enumerable property (not attribute) to be set?

dimitropoulos: the link is in the code comment line you selected for this comment: https://isomorphic-git.org/docs/en/fs#implementing-your-own-fs
Screenshot_20210324_151237

develohpanda: > I'm skeptical, since the docs say it must be there

Yeah I looked at that link and couldn't find anything about the property. In any case I'll double check this while testing and we can talk through it on a call!

packages/insomnia-app/app/ui/components/wrapper-home.js
DMarby: @dimitropoulos Almost, but not quite, I meant something like this:
```
try {
  // Clone the repo
} catch (err) {
  // If the repo URL does not end with .git, try cloning it with .git
  if (!endsWith('.git')) {
    try {
      // Clone the repo again but with the .git URL

      // The clone was successful, update the URL so that it's the new one that ends in .git
      repoSettingsPatch.uri = url;
      // Don't show an error, and don't return, as we need the flow to continue
    } catch (error) {
      // Show an error that combines the error messages from both the initial clone attempt, as well as the second one with the .git URL, and return.
      return false;
    }
  } else {
    // Show an error with the error message from the initial clone attempt, and return.
    return false; 
  }
}
```

packages/insomnia-app/app/sync/git/fs-plugin.js
dimitropoulos: I don't understand.  Are we talking about the same thing?  Here is where it's mentioned (highlighted):

Screenshot_20210325_082922

dimitropoulos: is this a semantics thing that's made more complicated by the fact that they have a literal object key of the name "enumerable" in their object.  Is what they mean in the above:

> If your `fs` object provides a property named `promises` that can be enumerated,

or does it mean

> If your `fs` object provides a property named `enumerable` on the `promises` property,

?

I understand what you mean, now.  I was parsing it as the second one but that was wrong.  It's the first one.  I was influenced heavily by the object itself (when you inspect it) having `enumerable: true` attached on it).

dimitropoulos: resolving for now since it's already been removed, thanks for explaining.

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: ok.  I simply don't share the confusion you're talking about avoiding.  Still, I see your point and I can understand how it may be confusing.

I updated this as well as the language referring to these as plugins since they are not any longer called that by isomorphic git.

memplugin 84df0548abb14dde40748adf5291b965aa9b17d4
httpplugin f571046f5243a265c865b3ee7d63ea817e58df24
nedbplugin 0a95f0074bec5ea5184784fc1e2deb0d92aef3bd
routablefsplugin d08cb2fedab11d5d8ce93bcd7badeb408106dc81
fsplugin 52c1654ecf7732a360b850f9fba31869f6287890

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: ok, I implemented this logic in 6222a403bdb16fd7498cfdf4263c95cd9e74c78c

I realized that the return of this function is never used (later in the same function it returns undefined or returns void) so I updated that while I was here to be accurate (and consistent) which enabled reducing the branching (but the logic should be identical to what you wrote).

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: I believe 4c56dc78f5565a782d8736b495d37e6726b4ad10 implements what you're asking for since it removes the renaming all together, if it didn't, please let me know.

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: 8aa2e42510dd5d58d02a88a8cc3f21288032710b

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: fixed in 8aa2e42510dd5d58d02a88a8cc3f21288032710b

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: db82fd64e92ccc38968aebc495ebc1637bde754a

dimitropoulos:
(as far as I am aware all comments have been addressed: please take another look)

packages/insomnia-app/app/ui/components/wrapper-home.js
develohpanda: Looks good :)

packages/insomnia-app/app/sync/git/__tests__/git-rollback.test.js
```diff
@@ -148,15 +148,15 @@ describe('git rollback', () => {
...
       expect(await vcs.status(bazTxt)).toBe('unmodified');
 
-      // Ensure the two files have been removed from the fs (memplugin)
-      await expect(fs.promises.readFile(fooTxt)).rejects.toThrowError(
+      // Ensure the two files have been removed from the fs (memClient)
+      await expect(fsClient.promises.readFile(fooTxt)).rejects.toThrowError(
         `ENOENT: no such file or directory, scandir '${fooTxt}'`,
       );
-      await expect(fs.promises.readFile(barTxt)).rejects.toThrowError(
+      await expect(fsClient.promises.readFile(barTxt)).rejects.toThrowError(
```
develohpanda, line 155: 🙌🏽

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -43,78 +54,95 @@ export type PushResponse = {
...
 };
 
-// isomorphic-git internally will default an empty ('') clone directory to '.'
-// Ref: https://github.com/isomorphic-git/isomorphic-git/blob/4e66704d05042624bbc78b85ee5110d5ee7ec3e2/src/utils/normalizePath.js#L10
-// We should set this explicitly (even if set to an empty string), because we have other code (such as fs plugins
-// and unit tests) that depend on the clone directory.
+type InitOptions = {
+  directory?: string,
+  fs?: Object,
+  gitDirectory?: string,
```
develohpanda, line 60: I don't think these should be optional, right?

packages/insomnia-app/app/sync/git/git-vcs.js
```diff
@@ -43,78 +54,95 @@ export type PushResponse = {
...
+  _baseOpts: {
+    dir: string,
+    gitdir?: string,
+    fs: Object,
+    http: Object,
+    onMessage: (message: string) => void,
+    onAuthFailure: (message: string) => void,
+    onAuthSuccess: (message: string) => void,
+    onAuth: () => void,
+  } = gitCallbacks();
```
develohpanda, line 95: I think this will complain when we go to TS because `dir`, `fs` and `http` are not optional but `gitCallbacks()` doesn't return an object containing those properties.

Regardless I think it might make sense to split these into `this._baseOpts` and `this._callbacks` which should work better handling this discrepancy.

This isn't a blocker nor does it require any action in this PR, just an observation and something we can adjust in future.

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,55 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
+              url: dotGitUrl,
+            });
+
+            // by this point the clone was successful, so update with this syntax
+            repoSettingsPatch.uri = dotGitUrl;
+          } catch (dotGitError) {
+            showError({
+              title: 'Error Cloning Repository: failed to clone with and without `.git` suffix',
+              message: `failed to clone with original url (${originalUrl}): ${originalUrlError.message};\n\nalso failed to clone with \`.git\` suffix added (${dotGitUrl}): ${dotGitError.message}`,
+            });
```
develohpanda: Because we don't have a stack to show maybe this would be more suitable as `showAlert` instead of `showError` as they have slightly different UI. Also, the message can actually be a React Node if we want to style this in a particular way 😄 

Again just an FYI, no action expected

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,55 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
+              fs: fsClient,
+              url: dotGitUrl,
+            });
+
+            // by this point the clone was successful, so update with this syntax
+            repoSettingsPatch.uri = dotGitUrl;
+          } catch (dotGitError) {
+            showError({
+              title: 'Error Cloning Repository: failed to clone with and without `.git` suffix',
+              message: `failed to clone with original url (${originalUrl}): ${originalUrlError.message};\n\nalso failed to clone with \`.git\` suffix added (${dotGitUrl}): ${dotGitError.message}`,
```
develohpanda: Just to follow the other error messaging patterns in the app, the modal title is usually quite brief and the message contains all the content.

```suggestion
              title: 'Error Cloning Repository',
              message: `Failed to clone with original url (${originalUrl}): ${originalUrlError.message};\n\nAlso failed to clone with \`.git\` suffix added (${dotGitUrl}): ${dotGitError.message}`,
```

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1027,37 +1026,43 @@ class App extends PureComponent {
...
-          // All app data is stored within the a namespaced directory at the root of the
-          // repository and is read/written from the local NeDB database
-          [GIT_INSOMNIA_DIR]: pNeDb,
-
-          // All git metadata is stored in a git/ directory on the filesystem
-          [GIT_INTERNAL_DIR]: pGitData,
-        },
-      );
+
+      /** All app data is stored within the a namespaced directory at the root of the repository and is read/written from the local NeDB database */
```
develohpanda:
```suggestion
      /** All app data is stored within a namespaced GIT_INSOMNIA_DIR directory at the root of the repository and is read/written from the local NeDB database */
```

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1027,37 +1026,43 @@ class App extends PureComponent {
...
+      const neDbClient = NeDBClient.createClient(activeWorkspace._id);
+
+      /** All git metadata is stored in a git/ directory on the filesystem */
```
develohpanda:
```suggestion
      /** All git metadata in the GIT_INTERNAL_DIR directory is stored in a git/ directory on the filesystem */
```

packages/insomnia-app/app/ui/containers/app.js
```diff
@@ -1027,37 +1026,43 @@ class App extends PureComponent {
...
+      const gitDataClient = fsClient(baseDir);
+
+      /** All data outside the directories listed below will be stored in an 'other' directory. This is so we can support files that exist outside the ones the app is specifically in charge of. */
+      const otherDatClient = fsClient(path.join(baseDir, 'other'));
+
+      const routableFS = routableFSClient(otherDatClient, {
```
develohpanda, line 1045: A comment to explain how the above clients (and the purpose they serve) are orchestrated by the `routableFSClient` would be useful. Something like:

```suggestion
      /** The routable FS client directs isomorphic-git to read/write from the database or from the correct directory on the file system while performing git operations. */
      const routableFS = routableFSClient(otherDatClient, {
```

develohpanda (commented):
Code looks good!

### QA with a GitHub repository that contains valid `.insomnia` contents:
🔴 `git@github.com:<your username>/<your repo> `
🔴 `git@github.com:<your username>/<your repo>.git`
Clones successfully, but then fails with a protocol error and no git sync dropdown is shown when opening the document
image


🔴 `http://github.com/<your username>/<your repo>`
🔴 `http://github.com/<your username>/<your repo>.git`
Fails silently


🔴 `https://github.com/<your username>/<your repo>`
🔴 `https://github.com/<your username>/<your repo>.git`
Clones successfully, but then shows an auth error and no git sync dropdown is shown when opening the document
image


### TODO
- [x] Test .git suffix fallback behavior with GitLab (which requires the suffix to exist)
- [x] Test commit/push/pull/rollback operations (blocked as per the above screenshots)
- [x] Test that data is not lost with a repository that contains files outside the `.insomnia` directory as well
- [x] Test setting up git sync on a document with all three syntax options)

packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -161,32 +163,55 @@ class WrapperHome extends React.PureComponent<Props, State> {
...
         // Pull settings returned from dialog and shallow-clone the repo
-        const { credentials, uri: url } = repoSettingsPatch;
+        const cloneParams = {
+          ...gitCallbacks({
+            username: repoSettingsPatch.credentials.username,
+            password: repoSettingsPatch.credentials.token,
+          }),
+          fs: fsClient,
+          http: httpClient,
+          core,
```
develohpanda: Looking at the docs, it doesn't look like this `core` property is used any more with version 1.x

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: I wasn't sure and it doesn't hurt to check them one way or the other so I err'd on the side of caution.

packages/insomnia-app/app/sync/git/git-vcs.js
dimitropoulos: right, yeah, in that case we'll just mark `dir`, `fs`, and `http` as optional since, well.. since they really are optional.  I don't want to mess with the types a whole lot here if we can avoid it since there are many places where they are incorrect anyway and I know the impending changes will address it (not just right here but lots of places).

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: fixed in bd34d164607966401b08194d77b166a4a0b4730b

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: because the title is what @nijikokun asked for, I'm going to leave it https://github.com/Kong/insomnia/pull/3111#discussion_r599160786

I've updated the capitalization in 6e42a57a310116085b3dedc866486940aba93a4f

packages/insomnia-app/app/ui/components/wrapper-home.js
dimitropoulos: thanks, fixed in bf7492c97be6907ff3703abf8bf7f5d784a8c06b

develohpanda (approved):
🟢 `git@github.com:<your username>/<your repo> `
🟢 `git@github.com:<your username>/<your repo>.git`
🔴 `http://github.com/<your username>/<your repo>`
Fails silently (not a regression)
🔴 `http://github.com/<your username>/<your repo>.git`
Fails silently (not a regression)
🟢 `https://github.com/<your username>/<your repo>`
🟢 `https://github.com/<your username>/<your repo>.git`
🟢 `.git` suffix fallback works when cloning and setting up git sync
🟢 SSH URL is supported (and translated to https) when cloning and setting up git sync
🟢 Commit/push/pull/rollback still works

develohpanda:
Merging to unblock #3247

PR #3114, opened 2021-02-24
Use the correct the application data directory for inso

### LINKED ISSUE #3058: inso run test command throws error (INSO CLI)
I have created unit tests currently in Insomnia Designer  its working  fine there with UI.
However, INSO CLI is throwing error with below command

**inso run test**
image

I have tried with command
**inso run test --appDataDir .**
this works but we cannot combine other options as mentioned in this document especially reporting doesn't work
I tried with  > output file as well but this seems to be doesnt work with **inso run test --appDataDir .**

image

This is verify critical feature please fix this asap

Logging after having discussion with Nijji

Neither of this command is working:
inso run test "MyTestSuite"
inso run test spc_ba58e1

I believe these are basic commands for inso not sure why this is also not working.

DMarby:
This PR makes inso use the correct application directory path on Windows and Linux. As `env-paths` is an opinionated library with no relation to Electron, I've opted to make a small utility function for this instead.

It also starts looking for the `Insomnia` data directory instead of `Insomnia Designer` one by default, and adds a fallback to for the `Insomnia Designer` data directory if it does not exist.
This creates a potential breakage if a user have both Core and Designer installed, but have not yet migrated their Designer data to Core. Please let me know your thoughts on whether that should be addressed or not (cc @nijikokun).

I have tested this on macOS and Windows, but would be great to have some more people test on those platforms, and also Linux in particular.

Fixes INS-458
Fixes INS-421
Fixes #3058


nijikokun:
> Please let me know your thoughts on whether that should be addressed or not (cc @nijikokun).

Let's note it within the changelog as a breaking change, and to use the previous version.

packages/insomnia-inso/src/db/index.js
```diff
@@ -61,9 +61,21 @@ export const loadDb = async ({
...
+    // Try to load from the Designer data dir, if the Core data directory does not exist
+    if (!db && !appDataDir) {
+      const designerDir = getAppDataDir('Insomnia Designer');
+      db = await neDbAdapter(designerDir);
+      db &&
+        logger.debug(
+          `Data store configured from Insomnia Designer app data directory at \`${path.resolve(
+            designerDir,
+          )}\``,
+        );
```
develohpanda, line 77: On a related note, it might be useful to make a note in this debug message that Designer has been deprecated and to switch to Core?

packages/insomnia-inso/src/db/index.js
```diff
@@ -61,9 +61,21 @@ export const loadDb = async ({
 
   // try load from nedb
   if (!db) {
-    const dir = appDataDir || envPaths(getDefaultAppDataDir(), { suffix: '' }).data;
+    const dir = appDataDir || getAppDataDir(getDefaultAppName());
     db = await neDbAdapter(dir, filterTypes);
     db && logger.debug(`Data store configured from app data directory at \`${path.resolve(dir)}\``);
+
+    // Try to load from the Designer data dir, if the Core data directory does not exist
+    if (!db && !appDataDir) {
```
develohpanda, line 69: 👍🏽

packages/insomnia-inso/src/data-directory.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import path from 'path';
+import { homedir } from 'os';
+
+// getAppDataDir returns the data directory for an Electron app,
+// it is equivalent to the app.getPath('userData') API in Electron.
+// https://www.electronjs.org/docs/api/app#appgetpathname
```
develohpanda, line 7: 👍🏽

develohpanda (approved):
Awesome! 🎉

PR #3120, opened 2021-02-25
Update the website links

DMarby:
Fixes INS-483


packages/insomnia-app/config/config.json
```diff
@@ -9,7 +9,7 @@
   "longName": "Insomnia",
   "synopsis": "The Collaborative API Client and Design Tool",
   "icon": "https://github.com/kong/insomnia/blob/develop/packages/insomnia-app/app/icons/icon.ico?raw=true",
-  "changelogBaseUrl": "https://insomnia.rest/changelog/core",
+  "changelogUrl": "https://insomnia.rest/changelog",
```
DMarby, line 12: The new website does not have individual changelog pages anymore I believe? Please correct me if I'm wrong and this should be changed to something else @nijikokun

nijikokun: Correct

develohpanda (approved):
With the changes made in this PR we now don't link to the changelogs of a particular version - is that a concern? I'm not sure how the new website handles changelogs and links

DMarby:
> With the changes made in this PR we now don't link to the changelogs of a particular version - is that a concern? I'm not sure how the new website handles changelogs and links

The new website only has the `/changelog` page containing all changelogs, no subpages for the individual ones currently, hence the change.

nijikokun (approved):

PR #3121, opened 2021-02-25
Add initial segment events

DMarby:
Quick and dirty initial implementation, see INS-484 for details.
This should probably be refactored to add some more structure once we replace GA with Segment in the future.

Fixes INS-484

### DIFF
--- a/packages/insomnia-app/app/common/analytics.js
+++ b/packages/insomnia-app/app/common/analytics.js
@@ -11,10 +11,13 @@ import {
   getAppVersion,
   getGoogleAnalyticsId,
   getGoogleAnalyticsLocation,
+  getSegmentWriteKey,
   isDevelopment,
 } from './constants';
 import type { RequestParameter } from '../models/request';
 import { getScreenResolution, getUserLanguage, getViewportSize } from './misc';
+import Analytics from 'analytics-node';
+import { getAccountId } from '../account/session';
 
 const DIMENSION_PLATFORM = 1;
 const DIMENSION_VERSION = 2;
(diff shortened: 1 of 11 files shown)


packages/insomnia-app/app/common/analytics.js
```diff
@@ -105,10 +108,59 @@ export async function getDeviceId(): Promise<string> {
   return deviceId;
 }
 
+let segmentClient = null;
+
+export async function trackSegmentEvent(event: String, properties?: Object) {
+  try {
+    if (!segmentClient) {
```
DMarby, line 115: This is a bit of a hack. Same as in other places in the codebase, we need to use `global.require` to override the axios adapter so that we use the NodeJS one rather then the Electron one, or we'll run into CORS issues and other things. `global.require` is not available until things have been initialized, hence why we do it when the method is called.

packages/insomnia-app/app/common/analytics.js
```diff
@@ -105,10 +108,58 @@ export async function getDeviceId(): Promise<string> {
...
```
DMarby, line 115: This is a bit of a hack. Same as in other places in the codebase, we need to use global.require to override the axios adapter so that we use the NodeJS one rather then the Electron one, or we'll run into CORS issues and other things. global.require is not available until things have been initialized, hence why we do it when the method is called. 
Could move this to a dedicated `init` method that we call on app init, but not sure there's much of a benefit.

packages/insomnia-app/app/common/analytics.js
```diff
@@ -105,10 +108,58 @@ export async function getDeviceId(): Promise<string> {
...
+  } catch (err) {
+    console.warn('[analytics] Error sending segment event', err);
+  }
+}
+
 // ~~~~~~~~~~~~~~~~~ //
 // Private Functions //
 // ~~~~~~~~~~~~~~~~~ //
 
+function _getOsName(): string {
```
DMarby, line 155: This is so that the OS name matches with what we send with events from the API, for easier cross referencing

packages/insomnia-app/app/common/analytics.js
develohpanda: > not sure there's much of a benefit

If you don't notice any difference while running probably fine to leave as-is and we can re-architect it while replacing GA with Segment.

I faced a similar issue in `inso` with the `run test` command; importing the package noticeably slowed down the CLI so we now lazy load when needed.

https://github.com/Kong/insomnia/blob/6f3d2666f8dcfd8cfcecbbb58d00691967c39a77/packages/insomnia-inso/src/commands/run-tests.js#L84-L85

develohpanda (approved):

nijikokun (approved):

Issue #3038, opened 2021-02-01
[Feature request] Auto-Generate XML Request Bodies from OpenAPI spec and prevent deletion of manually edited ones


waciii-ra:
**Question**

1. How does one get Insomnia to auto-generate XML Request Bodies?
2. How does one get Insomnia to not delete manually created XML Request Bodies if auto-generate doesn't work or isn't desired?

In contrast to `requestBody.content.application/json` (which does auto-generate Request Body), when using `requestBody.content.application/xml` Insomnia Designer is **NOT** auto-generating a Request Body as implied in documentation [here](https://support.insomnia.rest/article/98-debugging).  And it's deleting any manually created Request Body after navigating to the `Design` tab.

It appears that some **code or plugin** to auto-generate XML is missing and that the **refresh / re-generate request body** logic operates without awareness that there is nothing that is going to auto-generate a new Request Body for XML.

**Context**
I'm trying to create test cases for XML APIs.  I find that I can do that only by manually creating a Request Body.  However once doing so, Insomnia will delete the manually created Request Body after the `Design` Tab is clicked.  This makes Insomnia almost usuable for testing.

The sequence to see the XML Request Body deleted is strange (has a couple extra steps) because Insomnia has a wider issue (also occurs after a Git Update) with not automatically refreshing screens immediately when a change happens. 

To recreate:

1. Click `Design`. Create or access OpenAPI Spec with `requestBody.content.application/xml`
2. Click `Debug`.  Select `XML`.  **See nothing auto-generated**
3. Enter XML Request Body.  _Optionally Send.  Optionally Generate Tests and Execute.  Optionally click back and forth between_ `Debug` _and_ `Test`.  _Optionally restart Insomnia._  **See XML Request Body saved in all cases.**
4. Click `Design`.  Make no changes.  Click `Debug`. Click `Design` or `Test`.  Click `Debug`.  **See XML Request Body deleted**.

waciii-ra:
**NOTE:**  Recreate **requires navigating back and forth** to the Debug tab more than once due to an "Out-of-synch Refresh" problem.  **(ie: you must do Step 4 above)**

**System Environment:**
[recreate_00_SystemEnvironment-About.txt](https://github.com/Kong/insomnia/files/5906495/recreate_00_SystemEnvironment-About.txt)

**Example OpenAPI Spec:**
[...]

PR #2979, opened 2021-01-07
fix: only change environment on fresh import

apisaint:
This change makes it so that when switching between debug and design tabs, it only selects the default environment *the first time*. Subsequent switches will test whether an active environment id has already been set and ignore.

I've updated the comments as well. Since there were no tests around this area, I did not add or edit any - and will not do so since I am not familiar with the tests infrastructure yet.

Impacts #2228
Impacts #2680

Issue #2882, opened 2020-11-30
Design to Debug tab recreates all the queries every time


Videl:
**Describe the bug**
In the Designer, when going from Design to Debug, with the file type Open API 2, the queries are regenerated every time.
In the end I have a tons of duplicate queries that I don't commit. Also I can't edit them for fear that I will delete them when commiting.

**To Reproduce**
1. Find an OpenAPI 2 file or OpenAPI 3 with errors.
2. Remove all errors/warnings of the api file found by the parser if there are
3. Go to the debug tab
4. Wait a bit to see brand new queries

Notice the following in the console:

``
21:19:43.266 › [App] Forcing update from environment change (3) ["update", {…}, false]
21:19:43.300 › [App] Forcing update from environment change (3) ["update", {…}, false]
``

(nothing has been manually changed in the environment part)

**Expected behavior**
If the folder for a query is there, it should not regenerate inside.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Installation Method: .Exe install

Version: Insomnia Designer 2020.4.2
Release date: 21/10/2020
OS: Windows_NT x64 10.0.19041
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.69.1-DEV OpenSSL/1.1.1d zlib/1.2.11 WinIDN libssh2/1.9.0_DEV nghttp2/1.40.0

Maybe a way to manually block the query creation?

nijikokun:
> If the folder for a query is there, it should not regenerate inside.

Since it wouldn't regenerate, what would you expect or like to happen when changes occur?

RollMasterJW:
We can reproduce this with just a json file import. 

Steps to Reproduce:

1. Open Insomnia
2. Create new Document from URL
3. Use: https://api.rmaster.com/docs/endpoints/swagger.json
4.  Fetch and Import
5. Open Document in Debug view (note, only one 'API Version' endpoint)
6. Switch to Design View
7. Switch to Debug View (Note, once loaded there are now 2 'API Version endpoints)
8. Repeat 6 & 7 until confused as to which one was the one you edited and need to save of the 224 endpoints that get added each time

First Open

![image](https://user-images.githubusercontent.com/68345529/101520738-85cb7a00-394a-11eb-8217-58e95df6d0f6.png)

After switching views

[...]

Issue #2442, opened 2020-07-25
Insomnia Designer - operations in the Debug section are lost when switch to the Design section


projser:
I have added folders to organize operations in the Debug section when I changed to the Design section  and return to the Debug section, the operations are removed from the folders.

Steps to Reproduce

Create a Folder in the Debug section, add operations in it. Change to the Design section and return to the Debug section.

I have even commit the changes and tried to imported later on, but the operations and folders have been lost in the debug section too.

macOS 10.15.5
Insomnia version 2020.3.3

filfreire:
Hi @projser thanks for reporting this!

I'm going to go ahead and close this issue as duplicate of #2882 and #4283.

Internally we are aware that there's a handful of problems when switching between Debug and Design tabs, and we'll be addressing these soon.

Feel free to create new issues and report any other problems or improvements you come across!

PR #3124, opened 2021-02-26
Differential (Additive/Removal) patching when switching between design / debug

### LINKED ISSUE #2971: Updated requests rolled back to default after switching to DESIGN mode and then back to DEBUG mode
I have imported the design from the openAPI3.0 file.

I needed a few more requests and tried adding and reordering the requests in the debug mode. but every time I switch back to design mode, my changes are lost and the view resets to the default imported APIs

Here, I added the Auth folder and added a new request `Some New Request`, and ordered the `Login` API inside the `Auth` folder, and renamed it as `Reordered Login inside Auth Folder`.

image

When I switched to the design mode and returned back to the debug mode again, all the changes except newer requests are lost and the default environment variables are overridden.
image

It even clears the form data 

### before
image

### after
image

### Steps To Reproduce

1. Create a blank project
2. In the design tab, add OpenAPI 3.0 design
3. switch to the `DEBUG` tab
4. Reorder the requests (create a folder and move the request inside it, rename the request, etc.)
5. switch back to the `DESIGN` tab
6. again switch back to the `DEBUG` tab

apisaint:
😩 Finally got tired of dealing with it so I tried fixing it.

This patch adds differential patching for imports that occur during the switching between design and debug tabs inside of Designer. As reported through #2971, and others, this patch favors existing data over imported data, except urls (see below), values that exist already on the document will remain unchanged, only new values (including array based values) will be added / removed.

This also includes a bypass feature for urls, currently these options are not exposed through the interface but could be. This feature is an object to allow for future properties for preference-based patching.

This also allows endpoints to stay in position when moved or grouped in folders.

👇 Included:

- Adds `options.enableDiffBasedPatching` and `options.enableDiffDeep` to `importRaw` for backwards compat.
- Adds `options.bypassDiffProps.url` for url bypassing and use an object for future items.
- Adds `diffPatchObj` based on differential patching for objects (works with arrays as well).

💭 Future ideas:

- `hasBeenModifiedByUser` property map object to allow changing properties that haven't been touched by the user with options.
- Notice icon when clicked displays dialogue with changes that a user can choose to apply. Underlying diff method can be augmented to create such patch objects or use a lib.

⚠️ Notes:

Basically this makes changes additive, which is better than the current scenario especially when paired with #2979, what does that imply? See below.

- If it exists it won’t change unless a bypass option is created.
- When the `url` changes for a Request, when the bypass is enabled (true by default here) it will update the url in the url bar but it does not change the `name`, so the sidebar will not auto-update. This can be resolved in a future or with a modified check for the name using the future idea explained above.
- When items / values inside of the Headers change it does not auto-update, this would require either a bypass option, or rely on the future idea.

Related #3038
Closes #2971, #2882, #2442

### DIFF
--- a/packages/insomnia-app/app/common/import.js
+++ b/packages/insomnia-app/app/common/import.js
@@ -266,19 +266,20 @@ export async function importRaw(
     let newDoc: BaseModel;
     if (existingDoc) {
       let updateDoc = resource;
+
+      // Do differential patching when enabled
+      if (enableDiffBasedPatching) {
+        updateDoc = diffPatchObj(resource, existingDoc, enableDiffDeep);
+      }
+
+      // Bypass differential update for urls when enabled
+      if (bypassDiffProps?.url && updateDoc.url) {
+        updateDoc.url = resource.url;
+      }
+
       // If workspace, don't overwrite the existing scope
       if (isWorkspace(model)) {
         (updateDoc: Workspace).scope = (existingDoc: Workspace).scope;
-      } else {
-        // Do differential patching when enabled
-        if (enableDiffBasedPatching) {
-          updateDoc = diffPatchObj(resource, existingDoc, enableDiffDeep);
-        }
-
-        // Bypass differential update for urls when enabled
-        if (bypassDiffProps?.url && updateDoc.url) {
-          updateDoc.url = resource.url;
-        }
       }
 
       newDoc = await db.docUpdate(existingDoc, updateDoc);
--- a/packages/insomnia-app/app/common/import.js
+++ b/packages/insomnia-app/app/common/import.js
@@ -9,7 +9,7 @@ import { CONTENT_TYPE_GRAPHQL, getAppVersion } from './constants';
 import { showError, showModal } from '../ui/components/modals/index';
 import AlertModal from '../ui/components/modals/alert-modal';
 import fs from 'fs';
-import { fnOrString, generateId } from './misc';
+import { fnOrString, generateId, diffPatchObj } from './misc';
 import YAML from 'yaml';
 import { trackEvent } from './analytics';
 import {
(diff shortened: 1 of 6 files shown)


reynolek:
We have this on our list of issues to tackle, thanks for taking the time to look into it! Going to add @nijikokun to review it from a product perspective along with @develohpanda and @dimitropoulos for code.

packages/insomnia-app/app/common/import.js
```diff
@@ -240,7 +241,19 @@ export async function importRaw(
     const existingDoc = await db.get(model.type, resource._id);
     let newDoc: BaseModel;
     if (existingDoc) {
-      newDoc = await db.docUpdate(existingDoc, resource);
+      let updateDoc = resource;
+
+      // Do differential patching when enabled
+      if (options.enableDiffBasedPatching) {
```
nijikokun: `options` isn't checked for existence so this will fail with other import methods.

nijikokun (changes requested):

nijikokun:
Oh wow. Taking a look now  👀 

> * backwards compat.

Beautiful 🙌 

> This also allows endpoints to stay in position when moved or grouped in folders.

Absolutely amazing 🎉 

> * it will update the url in the url bar but it does not change the `name`, so the sidebar will not auto-update. This can be resolved in a future or with a modified check for the name using the future idea explained above.

While we do have `modified` you're right we don't have something that tells us whether it's been modified by the user or through creation, at least to my knowledge, so I agree this can be a future improvement. 👍 

> * Notice icon when clicked displays dialogue with changes that a user can choose to apply. Underlying diff method can be augmented to create such patch objects or use a lib.

Love this idea 💯, maybe this could also be expanded to sync at some point 🤔

> * `hasBeenModifiedByUser` property map object to allow changing properties that haven't been touched by the user with options.

Makes sense, track properties that change, and update the ones that haven't, seems like a good future improvement.  💯 

> * Adds `options.bypassDiffProps.url` for url bypassing and use an object for future items.

👍 

---

Amazing, love the breakdown. Found an issue in the code, but other than that it's definitely an improvement over the current situation and I'll take it.

packages/insomnia-app/app/common/import.js
apisaint: Good catch, instead of worrying about existence decided to default the argument to an object and let type checking do the worrying about input for us.

nijikokun (approved):

dimitropoulos (approved):

dimitropoulos:
(working on making sure I correctly resolve the merge conflict of the prior commit on master then this is next up to be merged)

PR #3125, opened 2021-02-26
Portable windows builds

DMarby:
Adds a portable windows build target, disables auto-updates for it and enables it to get update notifications from the API.

Fixes INS-496

### DIFF
--- a/packages/insomnia-app/app/common/constants.js
+++ b/packages/insomnia-app/app/common/constants.js
@@ -85,8 +85,17 @@ export function isLinux() {
 }
 
 export function updatesSupported() {
-  // Updates are not supported on Linux, or for Windows portable binaries
-  return !(isLinux() || (isWindows() && process.env.PORTABLE_EXECUTABLE_DIR));
+  // Updates are not supported on Linux
+  if (isLinux()) {
+    return false;
+  }
+
+  // Updates are not supported for Windows portable binaries
+  if (isWindows() && process.env.PORTABLE_EXECUTABLE_DIR) {
+    return false;
+  }
+
+  return true;
 }
 
 export function isWindows() {
--- a/packages/insomnia-app/app/common/constants.js
+++ b/packages/insomnia-app/app/common/constants.js
@@ -84,6 +84,11 @@ export function isLinux() {
   return getAppPlatform() === 'linux';
 }
 
+export function updatesSupported() {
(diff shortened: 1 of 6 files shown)


packages/insomnia-app/config/electronbuilder.json
```diff
@@ -63,13 +63,16 @@
   "win": {
     "target": [
       "squirrel",
-      "zip"
```
DMarby, line 66: We built this target but never used the result anywhere, so I've removed it.

nijikokun: Might be best to keep it in and check GitHub stats for usage to see if we can remove in a separate pr

DMarby: Sorry, to clarify, this target was never uploaded to GitHub releases, as our release script discards the artifacts.

nijikokun: Well then, in that case, 🧹 away

nijikokun (approved):

packages/insomnia-app/app/common/constants.js
```diff
@@ -84,6 +84,11 @@ export function isLinux() {
   return getAppPlatform() === 'linux';
 }
 
+export function updatesSupported() {
+  // Updates are not supported on Linux, or for Windows portable binaries
+  return !(isLinux() || (isWindows() && process.env.PORTABLE_EXECUTABLE_DIR));
```
dimitropoulos:
```suggestion
  // updates are not supported on Linux
  if (isLinux()) {
    return false;
  }
  
  // updates are not supported for Windows portable binaries
  if (isWindows() && process.env.PORTABLE_EXECUTABLE_DIR) {
    return false;
  }
  
  return true;
```

dimitropoulos: Although, of course, this is the same logically, I find that for things like this it can help with maintainability (and prevent regressions) if we avoid the demorgan's law confusion thing and also just to split the business logic up separately so you can discuss separately with coments.  It lowers cyclomatic complexity, as a side benefit.

by all means, though, I'm cool with whatever. :)  just thought I'd mention

DMarby: Sounds reasonable, much easier to grok 👍

dimitropoulos (approved):

PR #3130, opened 2021-03-01
Prompt for import as collection or document

develohpanda:
Example docs to test with:
- [OpenAPI](https://gist.githubusercontent.com/gschier/4e2278d5a50b4bbf1110755d9b48a9f9/raw/801c05266ae102bcb9288ab92c60f52d45557425/petstore-spec.yaml) (should prompt once)
- [Single document](https://gist.githubusercontent.com/develohpanda/6f1709f64343a334eb9cde2f573be685/raw/337f2f2e898bddb605d6f90ff2a4bd89e3fd1113/single-doc-export.json) exported from Designer (should prompt once)
- [Multiple documents](https://gist.githubusercontent.com/develohpanda/3d709b194a395b7fdb66f1003a3cf1d0/raw/24a57a787452939d5b411de0a24e7c55a8c75fb2/multiple-docs.json) exported from Designer (should prompt twice and import 3rd & 4th without prompt)

Should not prompt at all if importing from onboarding flow - it should auto import as a Design Document from there. (@nijikokun pls confirm)

### DIFF
--- a/packages/insomnia-app/app/common/__tests__/import.test.js
+++ b/packages/insomnia-app/app/common/__tests__/import.test.js
@@ -378,12 +378,24 @@ describe('export', () => {
 
 describe('isApiSpecImport()', () => {
   it.each(['swagger2', 'openapi3'])('should return true if spec id is %o', (id: string) => {
-    expect(importUtil.isApiSpecImport(id)).toBe(true);
+    expect(importUtil.isApiSpecImport({ id })).toBe(true);
   });
 
   it('should return false if spec id is not valid', () => {
     const id = 'invalid-id';
 
-    expect(importUtil.isApiSpecImport(id)).toBe(false);
+    expect(importUtil.isApiSpecImport({ id })).toBe(false);
+  });
+});
+
+describe('isApiSpecImport()', () => {
+  it.each(['insomnia-4'])('should return true if spec id is %o', (id: string) => {
+    expect(importUtil.isInsomniaV4Import({ id })).toBe(true);
+  });
+
+  it('should return false if spec id is not valid', () => {
+    const id = 'invalid-id';
+
+    expect(importUtil.isInsomniaV4Import({ id })).toBe(false);
   });
 });
--- a/packages/insomnia-app/app/common/__tests__/import.test.js
+++ b/packages/insomnia-app/app/common/__tests__/import.test.js
@@ -388,7 +388,7 @@ describe('isApiSpecImport()', () => {
   });
 });
 
-describe('isApiSpecImport()', () => {
+describe('isInsomniaV4Import()', () => {
   it.each(['insomnia-4'])('should return true if spec id is %o', (id: string) => {
     expect(importUtil.isInsomniaV4Import({ id })).toBe(true);
   });
--- a/packages/insomnia-app/app/common/__tests__/import.test.js
+++ b/packages/insomnia-app/app/common/__tests__/import.test.js
@@ -376,14 +376,14 @@ describe('export', () => {
   });
 });
 
-describe('isApiSpec()', () => {
+describe('isApiSpecImport()', () => {
   it.each(['swagger2', 'openapi3'])('should return true if spec id is %o', (id: string) => {
-    expect(importUtil.isApiSpec(id)).toBe(true);
+    expect(importUtil.isApiSpecImport(id)).toBe(true);
   });
 
   it('should return false if spec id is not valid', () => {
     const id = 'invalid-id';
 
-    expect(importUtil.isApiSpec(id)).toBe(false);
+    expect(importUtil.isApiSpecImport(id)).toBe(false);
   });
 });
--- a/packages/insomnia-app/app/common/import.js
+++ b/packages/insomnia-app/app/common/import.js
@@ -63,6 +63,19 @@ export type ImportResult = {
   summary: { [string]: Array<BaseModel> },
 };
 
+type ConvertResultType = {
+  id: string,
+  name: string,
+  description: string,
+};
+
+type ConvertResult = {
+  type: ConvertResultType,
+  data: {
+    resources: Array<Object>,
+  },
+};
+
 export type ImportOptions = {
   getWorkspaceId: () => Promise<string | null>,
   getWorkspaceScope?: string => Promise<WorkspaceScope>,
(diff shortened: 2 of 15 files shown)


packages/insomnia-app/app/common/import.js
```diff
@@ -242,6 +242,17 @@ export async function importRaw(
     if (existingDoc) {
       newDoc = await db.docUpdate(existingDoc, resource);
     } else {
+      // Set the workspace scope if creating a new workspace
+      //  IF is creating a new workspace
+      //  AND imported resource has no preset scope property OR scope is null
+      //  AND we have a function to get scope
+      if (
+        isWorkspace(model) &&
+        (!resource.hasOwnProperty('scope') || resource.scope === null) &&
```
develohpanda, line 251: This condition is key.

1. The `!resource.hasOwnProperty('scope')` check is for things coming from `insomnia-importers` which don't always have the scope set (eg. the openapi3 and swagger2 importers, see comments in this PR)

2. The `resource.scope === null` check is if importing a file that was exported **from Insomina Designer**. The unfortunate thing is that currently, Insomnia Designer does not set the `scope` to be anything, and we have no way of determining whether a particular V4 JSON/YAML export was created from Designer or Core, hence we must prompt.
	
	image

**The new behavior as a result of fixing the shortcoming of point 2 may become a blocker, so this solution is TBC.**

(Of course, when going through the migration flow we _do_ know it's coming from Designer and can [set the scope](https://github.com/Kong/insomnia/blob/6f3d2666f8dcfd8cfcecbbb58d00691967c39a77/packages/insomnia-app/app/common/migrate-from-designer.js#L207-L212) as such).

packages/insomnia-app/app/common/import.js
```diff
@@ -147,17 +164,14 @@ export async function importRaw(
...
-    let workspace = await models.workspace.getById(workspaceId || 'n/a');
-
-    // If none provided, create a new workspace
-    if (workspace === null) {
-      workspace = await models.workspace.create({ name: 'Imported Workspace' });
-    }
+    const workspace = await models.workspace.getById(workspaceId || 'n/a');
 
     // Update this fn so it doesn't run again
-    generatedIds[WORKSPACE_ID_KEY] = workspace._id;
```
develohpanda, line 158: We don't create the new workspace immediately, all we need here is an id so we just generate one

packages/insomnia-app/app/common/import.js
```diff
@@ -240,8 +254,15 @@ export async function importRaw(
...
+      // If workspace, don't overwrite the existing scope
+      if (isWorkspace(model)) {
+        (resource: Workspace).scope = (existingDoc: Workspace).scope;
+      }
       newDoc = await db.docUpdate(existingDoc, resource);
     } else {
+      if (isWorkspace(model)) {
+        await updateWorkspaceScope(resource, resultsType, getWorkspaceScope);
+      }
       newDoc = await db.docCreate(model.type, resource);
```
develohpanda, line 266: Allow the workspace to naturally be created here, but first figure out what the scope should be

packages/insomnia-app/app/models/workspace.js
```diff
@@ -10,10 +10,17 @@ export const prefix = 'wrk';
 export const canDuplicate = true;
 export const canSync = true;
 
+export const WorkspaceScopeKeys = {
+  designer: 'designer',
+  collection: 'collection',
+};
+
+export type WorkspaceScope = $Keys<typeof WorkspaceScopeKeys>;
```
develohpanda, line 19: Better late than never

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -328,28 +336,22 @@ export function importFile(workspaceId: string, forceToWorkspace?: ForceToWorksp
...
-          result,
-          'The file does not contain a valid specification.',
-        );
+
+        const options: ImportRawConfig = {
+          getWorkspaceScope: askToSetWorkspaceScope(forceToScope),
+          getWorkspaceId: askToImportIntoWorkspace(workspaceId, forceToWorkspace),
+        };
+        const result = await importUtils.importUri(uri, options);
+        handleImportResult(result, 'The file does not contain a valid specification.');
```
develohpanda, line 349: These error messages are very specific. The user will see this error message when importing an invalid postman collection, for example

packages/insomnia-app/app/common/import.js
```diff
@@ -274,17 +295,46 @@ export async function importRaw(
...
+async function updateWorkspaceScope(
+  resource: Workspace,
+  resultType: ConvertResultType,
+  getWorkspaceScope?: string => Promise<WorkspaceScope>,
+) {
+  // Set the workspace scope if creating a new workspace
+  //  IF is creating a new workspace
+  //  AND imported resource has no preset scope property OR scope is null
+  //  AND we have a function to get scope
+  if ((!resource.hasOwnProperty('scope') || resource.scope === null) && getWorkspaceScope) {
```
develohpanda, line 316: This condition is key.

1. The `!resource.hasOwnProperty('scope')` check is for things coming from `insomnia-importers` which don't always have the scope set (eg. the openapi3 and swagger2 importers, see comments in this PR)

2. The `resource.scope === null` check is if importing a file that was exported **from Insomina Designer**. The unfortunate thing is that currently, Insomnia Designer does not set the `scope` to be anything, and we have no way of determining whether a particular V4 JSON/YAML export was created from Designer or Core, hence we must prompt.
	
	image

(Of course, when going through the migration flow we _do_ know it's coming from Designer and can [set the scope](https://github.com/Kong/insomnia/blob/6f3d2666f8dcfd8cfcecbbb58d00691967c39a77/packages/insomnia-app/app/common/migrate-from-designer.js#L207-L212) as such).

packages/insomnia-app/app/common/__tests__/import.test.js
```diff
@@ -376,14 +376,26 @@ describe('export', () => {
...
 
   it('should return false if spec id is not valid', () => {
     const id = 'invalid-id';
 
-    expect(importUtil.isApiSpec(id)).toBe(false);
+    expect(importUtil.isApiSpecImport({ id })).toBe(false);
+  });
+});
+
+describe('isApiSpecImport()', () => {
```
DMarby:
```suggestion
describe('isInsomniaV4Import()', () => {
```

DMarby (approved):

sonicyeti (approved):

PR #3145, opened 2021-03-04
Rename Workspace verbiage

develohpanda:

### DIFF
--- a/packages/insomnia-app/app/common/strings.js
+++ b/packages/insomnia-app/app/common/strings.js
@@ -1,6 +1,4 @@
 // @flow
-import type { Workspace } from '../models/workspace';
-import { isDesigner } from '../models/helpers/is-model';
 import { pluralize } from './misc';
 
 export const strings = {
@@ -15,7 +13,3 @@ export const stringsPlural = {
   document: pluralize(strings.document),
   collection: pluralize(strings.collection),
 };
-
-// Some helpers
-export const getWorkspaceLabel = (w: Workspace) =>
-  isDesigner(w) ? strings.document : strings.collection;
--- a/packages/insomnia-app/app/common/strings.js
+++ b/packages/insomnia-app/app/common/strings.js
@@ -1,6 +1,7 @@
 // @flow
 import type { Workspace } from '../models/workspace';
 import { isDesigner } from '../models/helpers/is-model';
+import { pluralize } from './misc';
 
 export const strings = {
   workspace: 'Workspace',
@@ -8,6 +9,11 @@ export const strings = {
   document: 'Document',
   collection: 'Collection',
   home: 'Dashboard',
+};
+
+export const stringsPlural = {
+  document: pluralize(strings.document),
+  collection: pluralize(strings.collection),
 };
 
 // Some helpers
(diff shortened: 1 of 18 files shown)


DMarby (approved):

packages/insomnia-app/app/common/strings.js
```diff
@@ -10,6 +9,7 @@ export const strings = {
   home: 'Dashboard',
 };
 
-// Some helpers
-export const getWorkspaceLabel = (w: Workspace) =>
-  isDesigner(w) ? strings.document : strings.collection;
+export const stringsPlural = {
+  document: pluralize(strings.document),
+  collection: pluralize(strings.collection),
```
dimitropoulos, line 14: so should we also then add `workspace` to this list and remove `workspaces` above (as well as refactor the app accordingly)?

DMarby: Workspace/Workspaces will be removed from here post-release as they're no longer visibly used, but it's a bit larger of a refactor, there's a separate follow-up PR for it here: https://github.com/Kong/insomnia/pull/3146

packages/insomnia-app/app/common/strings.js
```diff
@@ -10,6 +9,7 @@ export const strings = {
   home: 'Dashboard',
 };
 
-// Some helpers
-export const getWorkspaceLabel = (w: Workspace) =>
-  isDesigner(w) ? strings.document : strings.collection;
+export const stringsPlural = {
```
dimitropoulos, line 12: what if we did something more like:

```ts
export const strings = {
  workspace: {
    singular: 'Workspace',
    plural: 'Workspaces',
  },
  collection: {
    singular: 'Collection',
    plural: 'Collections',
  },
};
```

The benefits of this are:

1. when the singular is used it's gotta always be crystal clear that the singular was intended
2. no multiple imports
3. we can make helper functions that can choose, e.g. `selectString(someNumber, strings.workspace)` could be a function that picks singular or plural based on the number (hopefully that makes sense).
4. no duplication (and therefore possibility of getting out of sync) of the keys `document`, `collection`, like they are now
5. we can use types to force every singular to actually have a plural, which right now is not the case.
6. when I have done things like this in the past, I have found it's useful to also store the correct indefinite article (`a` vs `an`) since it changes based on the sound being a vowel or not (and not necessarily always the letter), for example https://github.com/dimitropoulos/react-ui-roundup/blob/master/components.tsx#L23

dimitropoulos: update: I made INS-521 to capture this for another time.  we've got a release to cut!

dimitropoulos (approved):

nijikokun (approved):

packages/insomnia-app/app/common/strings.js
develohpanda: > we can make helper functions that can choose, e.g. selectString(someNumber, strings.workspace) could be a function that picks singular or plural based on the number (hopefully that makes sense).

Yep makes sense, there's similar code elsewhere (I think in o2k)

> when I have done things like this in the past, I have found it's useful to also store the correct indefinite article (a vs an) since it changes based on the sound being a vowel or not (and not necessarily always the letter),

That's a great idea!

PR #3162, opened 2021-03-07
font weight selector

### LINKED ISSUE #3161: [Feature Request] Allow font weight selection
I prefer to have _medium/semibold_ font weight for code editors because often i find _normal/book_ rendering to thin for my liking. Unfortunately insomnia only allows to select fonts in _normal_ or _book_ weight and even if i have only heavier versions of fonts installed they will not be seen in insomnia settings.

I added a font weight setting for the editor font that will allow user to make font lighter or heavier. Obviously user must have preferred weight installed on their system. Test on my machine shows that if selected font weight does not exists then closest to it will be rendered.

Here is the font weight selector in settings:
image

And here is font weight change in action:
font weight change

Of course there is possibility to compile font in desired weight as _normal_  if the source code is available. But its more a workaround rather than a solution.

danilkhromov:
This is a proposed implementation for font weight selection that adds following features:

  * font weight setting for editor font
  * font weight selector for editor font

This closes #3161 

This is how font weight selector looks like in settings:
font weight selector

And here is font weight change in action:
font weight change

Test on my machine shows that if selected font weight does not exists then closest to it will be rendered.


packages/insomnia-app/app/common/constants.js
```diff
@@ -182,6 +182,17 @@ export const MIN_INTERFACE_FONT_SIZE = 8;
...
+  { label: 'Thin', value: 100 },
+  { label: 'UltraLight', value: 200 },
+  { label: 'Light', value: 300 },
+  { label: 'Normal', value: 400 },
+  { label: 'Medium', value: 500 },
+  { label: 'SemiBold', value: 600 },
+  { label: 'Bold', value: 700 },
+  { label: 'ExtraBold', value: 800 },
+  { label: 'Heavy', value: 900 },
+];
```
develohpanda: Using an array for this purpose works but it seems better suited as an object/lookup 🤔 

```js
export const EDITOR_FONT_WEIGHTS = {
  Thin: 100,
  UltraLight: 200,
  Light: 300,
  ...
}
```

packages/insomnia-app/app/models/settings.js
```diff
@@ -96,6 +97,7 @@ export function init(): BaseSettings {
     clearOAuth2SessionOnRestart: true,
     fontInterface: null,
     fontMonospace: null,
+    fontMonospaceWeight: 400,
```
develohpanda: After converting `EDITOR_FONT_WEITGHTS` to an object, this should be set using `EDITOR_FONT_WEITGHTS.Normal` instead of hard-coded. That way the number doesn't seem magical to the next person looking at it 😄

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -108,6 +109,11 @@ class General extends React.PureComponent<Props, State> {
     app.exit();
   }
 
+  async _handleFontWeightChange(el: SyntheticEvent<HTMLInputElement>) {
+    const settings = await this._handleUpdateSetting(el);
+    setFont(settings);
+  }
```
develohpanda: This seems to be a duplicate of both `_handleFontSizeChange` and `_handleFontChange`. Maybe all the font related items (font, size, weight) could just use the `_handleFontChange` handler?

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -326,10 +332,34 @@ class General extends React.PureComponent<Props, State> {
...
+              {fontsMono ? (
+                <select
+                  name="fontMonospaceWeight"
+                  value={settings.fontMonospaceWeight}
+                  onChange={this._handleFontWeightChange}>
+                  {EDITOR_FONT_WEITGHTS.map((item, index) => (
+                    <option key={index} value={item.value}>
+                      {item.label}
+                    </option>
+                  ))}
```
develohpanda: It's good to avoid using array index as keys in React. While it's not as concerning in this particular use-case, it is still an [anti-pattern](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318).

After converting `EDITOR_FONT_WEITGHTS` to an object, you can use `Object.entries(EDITOR_FONT_WEITGHTS)` to get the key-value pairs.

```suggestion
                  {Object.entries(EDITOR_FONT_WEITGHTS).map(e => (
                    <option key={e[1]} value={e[1]}>
                      {e[0]}
                    </option>
                  ))}
```

develohpanda (changes requested):
Thank you for the PR, and welcome to the project!

This seems like a useful addition. I haven't had a chance to test this locally yet but I just had a couple of minor observations in the code that would be worth addressing 🙂 

Also, since this contains some UI changes, could you please add a screenshot or gif into the PR description? Thanks!

danilkhromov:
Thanks for taking your time to review my pr, and providing useful suggestions

As suggested, i improved following things:
* Converted `EDITOR_FONT_WEITGHTS` from array to object
* Now all font changes handled in one handler
* Added some screenshots/gif to the description

develohpanda (approved):
Looks good! Seems to work as well on macOS.

I'm not so familiar with this, but is there any way to determine which weights of a font are available, so we only show those weights instead of all?

Pinging @nijikokun for product review

danilkhromov:
@develohpanda i was not yet able to find an elegant solution that will allow display only installed weights for each font. But i can give it another try this weekend maybe.

danilkhromov:
@develohpanda so i think i found a solution that works and also not very hard to understand.

now only installed weights are shown for selected font:
show only installed weights

nijikokun:
Great work @danilkhromov looks good.

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -71,10 +83,22 @@ class General extends React.PureComponent<Props, State> {
...
     //  - https://github.com/Kong/insomnia/issues/1835
-    const fontsMono = fonts.filter(i => i.monospace || i.family.match(FORCED_MONO_FONT_REGEX));
+    const fontsMono = allFonts
+      .filter(i => i.monospace || i.family.match(FORCED_MONO_FONT_REGEX && !i.italic))
+      .map(({ family, monospace }) => {
+        const weights = allFonts
+          .filter(i => i.family === family)
+          .map(item => item.weight)
+          .filter((item, index, self) => self.indexOf(item) === index)
+          .sort((a, b) => (a > b ? 1 : -1));
```
develohpanda: This logic to extract the font weights for a particular font might be better suited to a traditional `forEach` and an ES6 `Set` to achieve the same result in fewer iterations.

```js
import { ascendingNumberSort } from '../../../common/sorting';
...
function getWeightsForFont(font: string): Array<number> {
  const weightSet = new Set();
  
  allFonts.forEach(({family, weight}) => {
    if (family === font)
      weightSet.add(weight);
    }
  });

  return [...weightSet].sort(ascendingNumberSort);
}
```

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -71,10 +83,22 @@ class General extends React.PureComponent<Props, State> {
...
       .sort((a, b) => (a.family > b.family ? 1 : -1));
 
-    // Find monospaced fonts
+    // Find monospaced fonts with installed weights for each font family
     // NOTE: Also include some others:
     //  - https://github.com/Kong/insomnia/issues/1835
-    const fontsMono = fonts.filter(i => i.monospace || i.family.match(FORCED_MONO_FONT_REGEX));
+    const fontsMono = allFonts
+      .filter(i => i.monospace || i.family.match(FORCED_MONO_FONT_REGEX && !i.italic))
+      .map(({ family, monospace }) => {
```
develohpanda: Weights are extracted for all monospaced fonts in the `componentDidMount` event, even if we only use one at a time, so there is some scope for improvement here given a user may have many fonts installed and the current implementation has several nested loops.

It think it might be more suitable to extract and show the font-weights only for the font that is selected. You could do this by storing a `editorFontWeights: {[string]: string}` key-value object in the state, separate to `fontsMono`, and change that `editorFontWeights` object as the selected font changes. This way we are only extracting data that's needed, when it's needed, and nothing more.

develohpanda (changes requested):
Nice work figuring out how to filter the font weights! As a user it certainly looks more polished now 🤗

After the changes for filtering fonts I've noticed some improvements that are worth making so marking this PR as such. Let me know how you get on, happy to help get it across the line! 🙂

packages/insomnia-app/app/ui/components/settings/general.js
danilkhromov: Thanks again for reviewing my pr and sorry for taking so long to implement requested changes. So now only weights for selected fonts are extracted and they are changed when selected font changes. I only declared `editorFontWeights` as an `Array<number>` since weight names can differ from font to font so its safer to use number in this case

develohpanda: No worries! Thank you for the updates. I have it on my list to review again 🙂

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -71,14 +86,21 @@ class General extends React.PureComponent<Props, State> {
       .filter(i => ['regular', 'book'].includes(i.style.toLowerCase()) && !i.italic)
       .sort((a, b) => (a.family > b.family ? 1 : -1));
 
-    // Find monospaced fonts
+    // Find monospaced fonts with installed weights for each font family
```
develohpanda, line 89: Good on you for keeping the comments up to date, very much appreciated 🙌🏽

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -71,14 +86,21 @@ class General extends React.PureComponent<Props, State> {
...
     // NOTE: Also include some others:
     //  - https://github.com/Kong/insomnia/issues/1835
-    const fontsMono = fonts.filter(i => i.monospace || i.family.match(FORCED_MONO_FONT_REGEX));
+    const fontsMono = allFonts
+      .filter(i => i.monospace || i.family.match(FORCED_MONO_FONT_REGEX && !i.italic))
```
develohpanda: Are you sure that the `!i.italic` check is in the right place? It seems like it's accidentally part of the `.match()` call. Let me know if that's not the case, I may be missing something 🙂

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -326,10 +371,36 @@ class General extends React.PureComponent<Props, State> {
...
+          <div className="form-control form-control--outlined">
+            <label>
+              Editor Font Weight
+              {fontsMono ? (
+                <select
+                  name="fontMonospaceWeight"
+                  value={settings.fontMonospaceWeight}
+                  onChange={this._handleFontChange}>
+                  {editorFontWeights.map((item, index) => (
+                    <option key={index} value={item}>
```
develohpanda: Definitely want to [avoid using index as key](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). `item` here should always be unique, so we can key with that.

```suggestion
                  {editorFontWeights.map(item => (
                    <option key={item} value={item}>
```

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -108,14 +130,37 @@ class General extends React.PureComponent<Props, State> {
...
+  _renderFontWeightSetting(fontFamily: string) {
+    const weights = this._getWeightsForFont(fontFamily);
+
+    this.setState({
+      editorFontWeights: weights,
+    });
   }
 
   async _handleFontChange(el: SyntheticEvent<HTMLInputElement>) {
+    const currentTarget = { name: el.currentTarget.name, value: el.currentTarget.value };
```
develohpanda, line 157: Prime use case for destructuring! 😄 

```suggestion
    const { currentTarget } = el;
```

packages/insomnia-app/app/ui/components/settings/general.js
```diff
@@ -108,14 +130,37 @@ class General extends React.PureComponent<Props, State> {
...
-  async _handleFontSizeChange(el: SyntheticEvent<HTMLInputElement>) {
-    const settings = await this._handleUpdateSetting(el);
+  _getWeightsForFont(fontFamily: string): Array<number> {
+    if (fontFamily === '__NULL__') {
+      return Object.values(EDITOR_FONT_WEITGHTS);
+    } else {
+      const weightSet = new Set();
 
-    setFont(settings);
+      const fonts = fontScanner.findFontsSync({ family: fontFamily });
```
develohpanda, line 139: This seems OK to me, but if calling into `fontScanner` too frequently does become an issue we may have to explore a different route in future. No changes required right now, though, I think it'll be fine!

Also, I wonder if the issue I noticed on Windows is related to this... 🤔 `fontScanner` has caused tricky issues in the past

develohpanda (commented):
Code generally looks good to me, but while testing on Windows it doesn't seem to be working as intended. 



I will be away for the next week or so, and can take a swing at Windows when I'm back, in case you don't have access to a Windows dev machine. Also, don't worry about keeping up to date with `develop`, I'll make sure it's up to date before merging. 🤗 Thank you for your patience!

danilkhromov:
@develohpanda i have a windows in the vm, so hopefully i will be able to reproduce this behavior there. Im also assuming that this is caused by the `fontScanner`, seems like it returns slightly different data on windows. 
I will look into that on the weekend and also will take care of other things that you mentioned (sorry if there is some obvious mistakes, im mostly developing backend services and i dont have very deep knowledge of js and react). 

Hopefully i will be able to tackle the windows issue, but if not, then your help will be much appreciated :slightly_smiling_face:

develohpanda:
Open-source is a really great way to learn and practice, so no worries at all! and no rush, take your time 🤗

danilkhromov:
> Code generally looks good to me, but while testing on Windows it doesn't seem to be working as intended.
> 
> 
> 
> I will be away for the next week or so, and can take a swing at Windows when I'm back, in case you don't have access to a Windows dev machine. Also, don't worry about keeping up to date with `develop`, I'll make sure it's up to date before merging. hugs Thank you for your patience!

after some investigation it looks like `fontScanner` on windows returns different data compared to what i have on linux (and i assume on macOS as well). `family` in the font object on windows contains full name of each installed variant e.g. `JetBrains Mono Medium` while on linux/macOS `family` contain only `JetBrains Mono` thus same way of collecting weights with `findFontsSync` does not work. Also for some reason `findFontsSync` always returns empty array on windows, even if i use some hardcoded values like `Consolas` or `JetBrains Mono`.

Not sure what would be the best approach in that case, getting all the weights for each font immediately might also result in different behavior on windows since it shows each installed font in separate `family` :disappointed: 

Any suggestions how i can tackle this would be much appreciated.

packages/insomnia-app/app/ui/components/settings/general.js
danilkhromov: for some reason if i do it like that then actual `value` that im getting to find weights is previously selected font instead of the new one :thinking: 
do you know why it behaves like that?

packages/insomnia-app/app/ui/components/settings/general.js
danilkhromov: my bad
fixed now :smiley:

packages/insomnia-app/app/ui/components/settings/general.js
develohpanda: That's...strange, not sure off the top of my head, but if it doesn't work keep it like you had it. My bad! 😔

develohpanda:
Thanks for that investigation! We all just love the differences between operating systems 😝 I'll spend some time on Windows next week and get back to you!

jackkav:
Closed as stale.

PR #3164, opened 2021-03-08
Better handling of .DS_Store files during migration

develohpanda:
When users manually copy-paste a plugin into the plugins folder in Designer, a `.DS_Store` file gets created. The migration should ignore this `.DS_Store` file. The expectation is that for the parts where the migration manually copies data across, it only does so for directories.

Therefore, the primary copy and remove helper functions in the migration can be restricted to only operate on directories.


DMarby (approved):

sonicyeti (approved):
Awesome!!!

PR #3170, opened 2021-03-08
Card order by date modified

sonicyeti:
Document home's cards are now displayed by date modified. Demo creation and modified below:

https://p191.p3.n0.cdn.getcloudapp.com/items/Jru1YDQP/e7dbad15-ab3b-499e-912d-e055e1b19182.mp4

Fixes INS-519

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper-home.js
+++ b/packages/insomnia-app/app/ui/components/wrapper-home.js
@@ -324,7 +324,7 @@ class WrapperHome extends React.PureComponent<Props, State> {
     handleSetActiveWorkspace(id);
   }
 
-  renderCard(w: Workspace) {
+  renderCard(workspace: Workspace) {
     const {
       apiSpecs,
       handleSetActiveWorkspace,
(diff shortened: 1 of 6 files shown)


packages/insomnia-app/app/ui/components/wrapper-home.js
```diff
@@ -504,7 +504,12 @@ class WrapperHome extends React.PureComponent<Props, State> {
     const { filter } = this.state;
 
     // Render each card, removing all the ones that don't match the filter
-    const cards = workspaces.map(this.renderCard).filter(c => c !== null);
+    const cards = workspaces
+      .map(this.renderCard)
+      .filter(c => c !== null)
+      .sort((c1, c2) => {
+        return c2.props.docLog.props.timestamp - c1.props.docLog.props.timestamp;
+      });
```
develohpanda: Drilling into two levels of props in this way, without unit tests, seems very unreliable.

It might be worth updating this logic to something along the lines of:
```js
    const cards = workspaces
      .map(this.renderCard)
      .filter(c => c !== null)
      .sort((c1, c2) => c2.timestamp - c1.timestamp)
      .map(c => c.card);
```
by updating the `this.renderCard` return type to `{ card: React.Node, timestamp: number } | null`.

We can't enforce that that component & prop tree will always remain the same, but with Flow/TS we can enforce return types and make this safer to consume.

develohpanda (dismissed):
Suggested change for accessing the timestamp for ordering

DMarby (approved):

PR #3200, opened 2021-03-17
Fix password editor props

### LINKED ISSUE #3192: Environmental Variables No Longer Supported in Basic Auth Password Field
**Describe the bug**
After a recent update, my environmental variables for the password field became replaced with asterisks. Instead of being able to hit ctrl+space and populate the password with a variable from my local environment, I instead get the option instead saved as text instead of code for the variable.

**To Reproduce**
1. Go to My environment and create an environment that has variables for username/password. 
2. Create a new request and set the authentication to basic.
3. set the basic username to the environmental variable for username. 
4. Try to set the password to the environmental variable for password and note that it doesn't work. 
5. Add the environmental variable for password to the username field and then copy it to the password field. 
6. Unhide the password and see something like -- `{% base64 'decode', 'normal', password %}`

**Expected behavior**
I expected environmental variables to be available in the password field. Being able to ctrl+space was so helpful. This used to be supported

**Screenshots**
image
image
image

 - OS: Windows 10 Pro 20H2
 - Installation Method: Installer 
 - App Version 2021.1.1

**Additional context**

develohpanda:
In the future, TypeScript should handle this better. Sadly Flow didn't pick up on this error.

Closes #3192

### DIFF
--- a/packages/insomnia-app/app/ui/components/editors/auth/basic-auth.js
+++ b/packages/insomnia-app/app/ui/components/editors/auth/basic-auth.js
@@ -103,8 +103,8 @@ class BasicAuth extends React.PureComponent<Props> {
                   password={authentication.password}
                   onChange={this._handleChangePassword}
                   nunjucksPowerUserMode={nunjucksPowerUserMode}
-                  render={handleRender}
-                  getRenderContext={handleGetRenderContext}
+                  handleRender={handleRender}
+                  handleGetRenderContext={handleGetRenderContext}
                   isVariableUncovered={isVariableUncovered}
                 />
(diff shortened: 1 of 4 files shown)


tharun208:
nice. also @develohpanda can we remove the testId's from the actual component ( password editor).

develohpanda:
> nice. also @develohpanda can we remove the testId's from the actual component ( password editor).

Ah, yes good catch. Will do when I'm back at my computer!

sonicyeti (approved):
Great catch!

DMarby (approved):

PR #3211, opened 2021-03-19
Expose identity token for oauth2

### LINKED ISSUE #1423: [Feature Request] OAuth 2 missing Identity token.
- Insomnia Version: `6.2.3`
- Operating System: Mac OSX

## Details

OAuth specification document provides two kinds of token, `access` token and `identity` token, The **Insomnia** OAuth 2 module works beautifully for most of the things as it exposes and uses the `Access_token` (in addition to `refresh_token`) which is used more often, there are a few oddballs however that need `identity_token` for certain calls, like **AWS Cognito** User pool integration with **API Gateway**:

To authenticate requests to API-Gateway endpoints that are protected by AWS Cognito **Authorizers**, the `access_token` will not work! but the `identity_token` which is included in the same response body that provides the `access_token` works! ( although access_token works to make calls to Cognito endpoints itself!! but anyways! )

I think it would be great if we provide this functionality as AWS is a big player and this will come handy in a lot of use-cases, especially since there aren't any other alternative clients, currently, supporting this ( Postman does not have this feature either ), this will be very useful for developers/engineers working in a serverless stack!

Thanks,
- mim

sayoojs:
Closes #1423 

Exposed identity token in the UI for all grant types in Oauth2.
image

Shows both identity token and refresh token in the Ctrl+Space autocomplete list for ease of use.
image

Made sure to handle populating identity token on clicking the `Refresh Token` button.


packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
```diff
@@ -606,6 +606,16 @@ class OAuth2Auth extends React.PureComponent<Props, State> {
               />
             </label>
           </div>
+          <div className="form-control form-control--outlined">
+            <label>
+              <small>Identity Token {tok ? <em>{expireLabel}</em> : null}</small>
+              <input
+                value={(tok && tok.identityToken) || ''}
+                placeholder="n/a"
+                onChange={this._handleUpdateAccessToken}
```
develohpanda: Seems like this should have it's own `this._handleUpdateIdentityToken` handler.

plugins/insomnia-plugin-request/index.js
```diff
@@ -46,15 +46,25 @@ module.exports.templateTags = [
             description: 'cookie value by name',
           },
           {
-            displayName: 'OAuth 2.0 Token',
-            value: 'oauth2',
-            description: 'access token',
+            displayName: 'OAuth 2.0 Access Token',
+            value: 'oauth2-access',
```
develohpanda: Ideally changing value is the right thing to do but changing this value is a breaking change.

On the current release, the tag `{% request 'oauth2', '', 0 %}` resolves correctly but on this branch it renders as
image and clicking on the tag it shows as 
image

The new values you have added below for identity and refresh are all good, but let's keep the access token `value` as it is (`oauth2`) and only change the display name, and add a comment to explain why. How does that sound?

develohpanda (changes requested):
Generally looks good but a couple of changes to make here 😄 

I am also not set up to be able to QA this, if you could outline some of your QA steps to show this works and fails correctly that would be very much appreciated.

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
sayoojs: Yes, will add an input handling function for the identity field. Missed that on first pass.

plugins/insomnia-plugin-request/index.js
sayoojs: Got it! Reverted access token tag value back to `oauth`. Left a comment to address the inconsistency.

sayoojs:
@develohpanda The QA I've done so far is 

- With 'Grant Type' as `Authorization Code`, tried querying an endpoint protected by AWS Cognito user pools which uses the identityToken as Bearer. Successfully retrieved a 200.
- In cases where there is no `identityToken` returned by the auth provider, the input field will show `n/a`, and 'No token' message is shown on auto complete(Ctrl+Space).(Emulated the null return by manually removing the return for this attribute from the get function)
- - In case of either an empty string value or a null. Checked if the 'No token' message showed up on the autocomplete tag's edit screen. 
- Have **NOT** tested any of the other grant types due to not having any existing environments which uses those authentication grant methods at hand. **This is a concern**. 

image
image

I've also made the requested changes and pushed the commit, but since I forked, Vercel, is asking for authentication.

plugins/insomnia-plugin-request/index.js
```diff
@@ -46,15 +46,29 @@ module.exports.templateTags = [
             description: 'cookie value by name',
           },
           {
-            displayName: 'OAuth 2.0 Token',
+            displayName: 'OAuth 2.0 Access Token',
             value: 'oauth2',
-            description: 'access token',
+            /*
```
develohpanda, line 51: 👍🏽

develohpanda (approved):
Thanks for the quick turnaround and your notes! 🙌🏽

I am happy with the code and usability though I'm not sure about the expectation for token expiry (should it be the same as the access token?). We may also need to explore how to QA this, maybe allowing users from the parent issue to QA a beta build.

@DMarby may have some thoughts on QA as well.

sayoojs:
Hi @develohpanda 

I've pushed another commit which addresses handling the Identity Token expiry separately. 
What I've done is decode the `identityToken` JWT to fetch the `exp` and render the expiry from that value.
So on the off chance the expiry for the identity token is different from that of the access token, it should render differently.

Also, any updates on including this in a beta build?

Thanks in advance! :smiley:

develohpanda:
Awesome! I will review again, and I'll raise it with the team next week to make sure we get some more reviews on this PR and figure out a way forward. 🤗

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
```diff
@@ -483,6 +497,32 @@ class OAuth2Auth extends React.PureComponent<Props, State> {
...
+    let { exp } = JSON.parse(window.atob(base64Url));
+    if (!exp) {
+      return '(never expires)';
+    }
+    /* 
+      Need to find exp's digit count and then remove digits or pad it with 0s to 
+      make it exactly 13 digits. As the 'TimeFromNow' component only seems to 
+      render 13 digit epochs correctly.
+    */
+    const expDigitCount = exp.toString().length;
```
develohpanda: Looks like it's using [`moment`](https://momentjs.com/docs/#/parsing/unix-timestamp-milliseconds/) to create the timestamp.

What does the expiry extracted from the token look like/represent? Feels like the padding or removing of digits could cause some strange behavior 🤔

develohpanda (approved):
Nice 🎉

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
```diff
@@ -483,6 +497,32 @@ class OAuth2Auth extends React.PureComponent<Props, State> {
     return { basic: basicFields, advanced: advancedFields };
   }
 
+  static renderExpiryFromJWT(token: string): React.Element<*> | string | null {
+    if (!token) {
+      return null;
+    }
+
+    const base64Url = token.split('.')[1];
```
develohpanda: I just pulled this down locally to test and the Oauth2 tab fails to render for a brand new request as a result of this implementation. The following suggestion matches how `renderExpireAt` is written

```suggestion
  static renderExpiryFromJWT(token: ?OAuth2Token): React.Element<*> | string | null {
    if (!token || !token.identitiyToken) {
      return null;
    }

    const base64Url = token.identitiyToken.split('.')[1];
```

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
```diff
@@ -541,7 +580,8 @@ class OAuth2Auth extends React.PureComponent<Props, State> {
     const { request, oAuth2Token: tok } = this.props;
     const { loading, error, showAdvanced } = this.state;
 
-    const expireLabel = OAuth2Auth.renderExpireAt(tok);
+    const accessExpireLabel = OAuth2Auth.renderExpireAt(tok);
+    const identityExpireLabel = OAuth2Auth.renderExpiryFromJWT(tok.identityToken);
```
develohpanda:
```suggestion
    const identityExpireLabel = OAuth2Auth.renderExpiryFromJWT(tok);
```

develohpanda (changes requested):
There is a bug here such that `window.atob()` fails fatally if the identity token is not encoded correctly (if you just type anything into the input field). The field has a requirement to be encoded in a particular way from what I can see, which is fine, but there shouldn't be a UI failure if it's _not_ encoded as expected. 🤔 

image

Another bug in comments below. Once these are fixed up I can trigger an alpha build, or you can bundle it locally for your OS by running `npm run app-package:smoke` and fetching artefacts from the `packages/insomnia-app/dist` folder.

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
sayoojs: I'll make this change. 
I was trying to keep the function generic, to be able to be used with any JWT :sweat_smile:
I guess since accessToken already has an expiry function, i'll make this one dedicated to identityToken.

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
sayoojs: The extracted expiry for identityToken was a 10 digit epoch representation in my case (Eg: `1617202697` Precision up to seconds).
For accessToken's `expires_in` key its 13 digit epoch (Eg: `1617202697111`  Precision up to milliseconds).

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
develohpanda: Fair enough 😅 it fails because likely somewhere in the render loop this logic is reached while token is `undefined`, and the previous code would fail with an error like `identityToken cannot be found on undefined`

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
develohpanda: I see! Reading some docs about the JWT `exp` property it seems this is usually in seconds. Is this always the case or situational? If always, I'd be more comfortable with a solution what converts seconds to milliseconds by multiplying by 1000, rather than removing or padding digits 🤔 since we're working with numbers the multiplication seems easier to follow.

sayoojs: Oh no, by padding (as shown in snippet below) that's what I'm doing. 

I'm essentially dividing or multiplying by 10^n, where n is calculated as the difference between the number 13 and the actual digit count of epoch generated from the JWT exp property. 
  
`exp = exp * 10 ** (13 - expDigitCount);`

So in cases of epoch seconds, that code multiplies by 10^3. 
In case of epoch microseconds, that code multiplies by 10^-3. Which essentially remove the digits that representing the microseconds.

Maybe I should force cast the result to big int to prevent floating points on x10^-n

develohpanda: Ahhhh okay, thanks for explaining that!!

> Maybe I should force cast the result to big int to prevent floating points on x10^-n

That sounds reasonable.

I think it would be good to extract these few lines into a standalone utility method somewhere and writing a few tests as a way to document input/output of this logic. 🙌🏽

sayoojs: Which lines specifically? The padding/trimming part?

develohpanda: Yep! The padding and trimming

sayoojs: Is there a file with preexisting util functions? Where I can export this to? Or do I just create a static method within this class itself?

develohpanda: This might be a good place! Looks like there a similar helper here too, for snapping a number to a minimum or maximum.

https://github.com/Kong/insomnia/blob/45c210eb41f52ddc8cb67abb424c4bd786c4e0ea/packages/insomnia-app/app/common/misc.js#L444

The associated tests would go in here: https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/common/__tests__/misc.test.js

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
sayoojs: This fix is implemented.

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.js
sayoojs: Moved this functionality to misc.js and added tests. (Do review the tests, this is the first time I've written JS tests)

sayoojs:
@develohpanda  I've implemented the requested changes.

Any idea why the Ubuntu tests might be failing? I couldn't infer anything from the logs. Could you try re-triggering it?

develohpanda (approved):
Thank you so much 🙏🏽 Sometimes Spectron (the E2E test framework we use for Electron) fails to connect to the chrome driver and gives cryptic errors like that. I have re-triggered, no action needed from your side 🤗

kateefimova-eb:
What an amazing feature! Thank you so much 👏

PR #3212, opened 2021-03-19
Replace Netlify with Vercel for Storybook

DMarby:
Storybook is now live on Vercel at http://storybook.insomnia.rest

All build configuration is done through their web interface as per their documentations recommendation, I have added a vercel config file to stop it from posting comments however, as otherwise it gets really noisy, as it posts comments regardless of whether changes has been made or not (we cancel builds if there aren't any changes to the `packages/insomnia-components` directory: 
image

Since the build status/preview link are still accessible through the deployment status and the checks at the bottom of PR/on a commit, I think this is fine.

Fixes INS-453

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,5 +1,0 @@
-[build]
-  publish = "storybook-static"
-  base = "packages/insomnia-components/"
-  command = "npm run convert-svg && npm run build-storybook"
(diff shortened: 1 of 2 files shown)


sonicyeti (approved):
Nice, this seems smooth.

develohpanda (approved):

PR #3231, opened 2021-03-23
Scroll support for unit testing

### LINKED ISSUE #3216: Cannot scroll unit tests when greater than window size
**Describe the bug**
When adding a large number of unit tests (20+ depending on window size) to a test suite, the unit test list does not scroll. If you run a test suite with a large number of tests then the test results do not scroll either.

**To Reproduce**
1. Go to a design document with requests
2. Click on the Test tab
3. Create a New Test Suite
4. Add Tests until the new tests run off the bottom of the window
5. Try to access the last test
6. Run the tests
7. Try to see all the test results

**Expected behavior**
Once the list of unit tests is longer then can be displayed on screen it should allow scrolling.
Once the list of test results is longer then can be displayed it should allow scrolling

 - OS: Windows 10
 - Installation Method: windows installer
 - App Version: Insomnia 2021.1.1

**Additional context**
Work arounds include using the view menu to zoom out. Or using dev tools to add "overflow: auto;" to .unit-tests style. The css fix only works on the tests, I'm not sure what class needs to be overridden to fix the results.

sonicyeti:
Addressing the scrolling regression recently found within unit testing.

Fixes #INS-589
Closes #3216

### DIFF
--- a/packages/insomnia-app/app/ui/css/components/wrapper.less
+++ b/packages/insomnia-app/app/ui/css/components/wrapper.less
@@ -93,28 +93,26 @@
     }
   }
 
-  .pane-one {
+  .pane-one, .pane-two {
     min-width: 0;
-    grid-column-start: 4;
-    grid-row-start: 2;
-    grid-row-end: span 3;
+    max-height:100%;
+    overflow-y: auto;
     background: inherit;
-
     border-left: 1px solid var(--hl-md);
     background: var(--color-bg);
     color: var(--color-font);
   }
 
+  .pane-one {
+    grid-column-start: 4;
+    grid-row-start: 2;
+    grid-row-end: span 3;
+  }
+
   .pane-two {
-    min-width: 0;
     grid-column-start: 6;
     grid-row-start: 2;
     grid-row-end: span 3;
-    background: inherit;
-
-    border-left: 1px solid var(--hl-md);
-    background: var(--color-bg);
-    color: var(--color-font);
   }
 
   .app-header {
--- a/packages/insomnia-app/app/ui/css/components/wrapper.less
+++ b/packages/insomnia-app/app/ui/css/components/wrapper.less
@@ -96,7 +96,6 @@
   .pane-one, .pane-two {
     min-width: 0;
     max-height:100%;
-    overflow-y: auto;
     background: inherit;
     border-left: 1px solid var(--hl-md);
     background: var(--color-bg);
--- a/packages/insomnia-app/app/ui/css/components/unit-tests.less
+++ b/packages/insomnia-app/app/ui/css/components/unit-tests.less
@@ -1,5 +1,6 @@
 .unit-tests {
   height: 100%;
+  overflow-y: auto;
 
   .editor {
     overflow-y: hidden !important;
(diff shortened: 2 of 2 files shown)


packages/insomnia-app/app/ui/css/components/wrapper.less
```diff
@@ -93,28 +93,26 @@
     }
   }
 
-  .pane-one {
+  .pane-one, .pane-two {
```
develohpanda, line 96: Nice cleanup, de-duplication is awesome ❤️

packages/insomnia-app/app/ui/css/components/wrapper.less
```diff
@@ -93,28 +93,26 @@
...
     min-width: 0;
-    grid-column-start: 4;
-    grid-row-start: 2;
-    grid-row-end: span 3;
+    max-height:100%;
+    overflow-y: auto;
```
develohpanda: I'm inclined to suggest we don't add an overflow to the `.pane-x` class but rather make it `height: 100%` and add the overflow to the actual child component (eg. [unit testing](https://github.com/Kong/insomnia/blob/e1812efdb31ed23de3fbdfffad87cc919049e93a/packages/insomnia-app/app/ui/components/wrapper-unit-test.js#L414)). This would mean that the other views also remain unaffected.

This is because the panes actually define the _layout_ generically whereas overflow is concerned with _contents_.

For example, `pane-one` is rendered like this in `page-layout.js` https://github.com/Kong/insomnia/blob/26fb78ab99f3d16858fd4afe76d9ca3ba8fe7d92/packages/insomnia-app/app/ui/components/page-layout.js#L148-L154 and consumed like this in `wrapper-unit-test.js` https://github.com/Kong/insomnia/blob/e1812efdb31ed23de3fbdfffad87cc919049e93a/packages/insomnia-app/app/ui/components/wrapper-unit-test.js#L499-L505.

IMO the `this._renderTestSuites` and `this._renderResults` render functions should control overflow themselves, using the components they render.

Keen to hear your thoughts 😄 the `page-layout` refactor was done last year to allow for the common pane-resizing/visibility logic to exist in one place and be reused.

develohpanda (commented):
Looks like the sidebar needs some love as well 😆 

image

packages/insomnia-app/app/ui/css/components/wrapper.less
sonicyeti: Will cover the sidebar as well, thanks for the catch! I’ll address it higher up on wrapper if possible, I understand what you mean, it would make sense and prove more scalable down the road.

sonicyeti: Ok set, picked up a smidge of redundancy at the UT panel level but I didn't want to break up the BEM-ish structure for the sake of pulling a single occurrence of the overflow prop.

dimitropoulos (approved):
Screenshot_20210324_085734
functionally tested: looks great

packages/insomnia-app/app/ui/css/components/wrapper.less
develohpanda: Sounds reasonable 🥳

develohpanda (approved):
Lookin' good now! 🥳

PR #3233, opened 2021-03-23
Fix license metadata

### LINKED ISSUE #3219: Unknown license in releases
**What is your question?**
I noticed that both the .rpm and .deb packages available in [Releases](https://github.com/Kong/insomnia/releases/) have unknown license in their specifications. I want to know if that's intended or a mistake.

**What is the context?**
Releases

bfsgr:
Adds license field in packages so that linux builds ship with the correct license. 
Closes  #3219, Closes INS-599.


dimitropoulos:
While we're doing this, may I ask that we also change inso to use a valid license?  [Right now it's `Apache2`](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-inso/package.json#L7), which is close, but not on the list of valid licenses https://spdx.org/licenses/.  We need to change it to `Apache-2.0` (or, for that matter, `MIT`).

@nijikokun is that ok?

In case you think this is a pedantic request, some tooling reports this when you install, e.g.:
```
warning package.json: License should be a valid SPDX license expression
```

And, if nothing else, it'd be nice to have that error go away.

nijikokun:
Thanks for raising the issue @dimitropoulos. Let's go ahead and make sure it's properly labeled as `Apache-2.0`.

bfsgr:
> Thanks for raising the issue @dimitropoulos. Let's go ahead and make sure it's properly labeled as `Apache-2.0`.

Done :+1:

dimitropoulos (approved):

bfsgr:
Any idea why windows tests failed?

DMarby:
I am not sure whether the changes in this PR will actually resolve #3219, as it is unclear whether actually having the license in the `package.json` will make `electron-builder` correctly set the license metadata as per the following open issues, or whether we need to include a `debian/copyright` file ourselves or similar:
https://github.com/electron-userland/electron-builder/issues/4270
https://github.com/electron-userland/electron-builder/issues/3914

That said, I still think we should merge the changes in this PR regardless, but I'm not sure if we can close #3219 until we verify the resulting `.deb` and if further changes are needed.

bfsgr:
I tested it locally and putting the license field in `package.json` did create a correctly licensed .rpm, I didn't test it to .deb but I think it's safe to assume it will be correct as well.

Just to be safe I can test it again

bfsgr:
@DMarby  Just checked it locally and yes, both .rpm and .deb build with the correct license. However I'm not familiar with metadata to .AppImage and .snap but I can look into it too

bfsgr:
Digging deeper I've found that the .snap does not ship with a license field inside `meta/snap.yaml`  and I was unable to find metadata inside the .AppImage. I think the bigger issue may be with the .snap package.

But solving this maybe out of scope of this PR

develohpanda (commented):
I'll need to double check whether we should be applying MIT or Apache-2.0; to my knowledge it is the latter, so let's hold off on merging this until confirmed.

I echo the other points raised, however, that we should definitely get this in. Thank you for raising a PR and running through those investigations! 🤗

develohpanda:
A quick update, this is on the backlog to chase up with the right people. Until then, marking as blocked 🙂

develohpanda:
Okay! I have some more information:
* everything that _doesn't_ have a license should go to `Apache-2.0`
* everything that _does_ have a license should stay as-is
* set the root and `insomnia-app` licenses to `MIT` (we don't want to switch the license for the main app just yet)

@reynolek please advise if this doesn't capture it correctly 🙂

reynolek:
That looks correct from the info that I was given 👍

bfsgr:
> Okay! I have some more information:
> 
>     * everything that _doesn't_ have a license should go to `Apache-2.0`
> 
>     * everything that _does_ have a license should stay as-is
> 
>     * set the root and `insomnia-app` licenses to `MIT` (we don't want to switch the license for the main app just yet)
> 
> 
> @reynolek please advise if this doesn't capture it correctly slightly_smiling_face

Alright, just applied the changes,  please review

reynolek (approved):
Approving from product perspective based on Legal team's input

develohpanda (approved):

develohpanda:
Thank you for working through this!

PR #3256, opened 2021-04-01
o2k parameter style support & default

sonicyeti:
This PR covers the setting of a user defined style value on query parameters while defaulting to 'form' if omitted. Previously we were setting this to 'simple' regardless, which is not a [valid option for query parameters](https://swagger.io/docs/specification/serialization/#query).

Fixes #INS-557


reynolek (approved):
LGTM

dimitropoulos (approved):

PR #3264, opened 2021-04-05
fix(tracking data) - Honor data sharing

### LINKED ISSUE #3255: Telemetry is being sent even whilst opted-out + includes personal identifiers
Edit: The Insomnia team released a fix within only a few short days (#3264). I was even offered the opportunity to go over these changes before they were merged. My faith has been fully restored and I'm properly amazed at how this was handled!

---

Today I noticed requests were being made regularly to `https://api.insomnia.rest/notification`. These turned out to be `POST` requests, and they are sending precisely the telemetry that I have explicitly opted out of sharing.

Contrary to the config screen claim that only anonymous data will be sent if one should opt in, several datapoints are included that both the CCPA and GDPR list as personal identifiable information:  `deviceId`, `firstLaunch`, IP address, OS version. It is impossible for a property that is referred to by the word "identifier" (`deviceId`) to be anonymous to any degree, especially when it is combined with other data.

Additionally, Google Analytics appears to happily be collecting events and who knows what else (all the things). You can never legally process this data anyway as it was collected during a period without consent - which, even if it were to be given at a later date, would not apply retroactively. Please consider disabling data collecting completely when the user has not opted in. Close the holes.

--

optout

The configuration setting that I have disabled says:

> Data Sharing
> Help Kong improve its products by sending anonymous data about features and plugins used, hardware and software configuration, statistics on number of requests, collections, documents, etc.
Please note that this will not include personal data or any sensitive information, such as request data, names, etc.

However the following data is being transmitted:

```json
Common HTTP request data
{
  "firstLaunch":[redacted],
  "launches":[redacted',
  "platform":"darwin",
  "app":"com.insomnia.app",
  "version":"2021.2.2",
  "requests":[redacted],
  "requestGroups":[redacted],
  "environments":[redacted],
  "workspaces":[redacted],
  "updatesNotSupported":false,
  "autoUpdatesDisabled":false,
  "disableUpdateNotification":false,
  "updateChannel":"stable",
  "deviceId":[redacted],
  "createdRequests":[redacted],
  "deletedRequests":[redacted],
  "executedRequests":[redacted[
}
```

The fact that this data is being sent to such an inconspicuous URL, one containing the word "notifications" no less, left a very bad taste in my mouth.

However this came about, it is a serious breach of trust and confidence. Today it's telemetry, tomorrow it might be sensitive work data.

I kindly request for this to be fixed as soon as possible.

reynolek:
* Utilizes the Data Sharing preference to determine what data is sent to the notification endpoint

Closes #3255

### DIFF
--- a/packages/insomnia-app/app/common/analytics.js
+++ b/packages/insomnia-app/app/common/analytics.js
@@ -111,6 +111,11 @@ export async function getDeviceId(): Promise<string> {
 let segmentClient = null;
 
 export async function trackSegmentEvent(event: String, properties?: Object) {
+  const settings = await models.settings.getOrCreate();
+  if (!settings.enableAnalytics) {
+    return;
+  }
+
   try {
     if (!segmentClient) {
       segmentClient = new Analytics(getSegmentWriteKey(), {
--- a/packages/insomnia-app/app/ui/components/toast.js
+++ b/packages/insomnia-app/app/ui/components/toast.js
@@ -112,25 +112,32 @@ class Toast extends React.PureComponent<Props, State> {
 
     // Try fetching user notification
     try {
-      const data = {
+      // Telemetry data
+      const trackingData = {
         firstLaunch: stats.created,
         launches: stats.launches,
-        platform: getAppPlatform(),
-        app: getAppId(),
-        version: getAppVersion(),
         requests: await db.count(models.request.type),
         requestGroups: await db.count(models.requestGroup.type),
         environments: await db.count(models.environment.type),
         workspaces: await db.count(models.workspace.type),
-        updatesNotSupported: !updatesSupported(),
-        autoUpdatesDisabled: !settings.updateAutomatically,
-        disableUpdateNotification: settings.disableUpdateNotification,
-        updateChannel: settings.updateChannel,
         deviceId: await getDeviceId(),
         createdRequests: stats.createdRequests,
         deletedRequests: stats.deletedRequests,
         executedRequests: stats.executedRequests,
       };
+
+      // Non-telemetry data
+      const commonData = {
+        platform: getAppPlatform(),
+        app: getAppId(),
+        version: getAppVersion(),
+        updatesNotSupported: !updatesSupported(),
+        autoUpdatesDisabled: !settings.updateAutomatically,
+        disableUpdateNotification: settings.disableUpdateNotification,
+        updateChannel: settings.updateChannel,
+      };
+
+      const data = settings.enableAnalytics ? { ...trackingData, ...commonData } : commonData;
 
       notification = await fetch.post('/notification', data, session.getCurrentSessionId());
     } catch (err) {
--- a/packages/insomnia-app/app/ui/components/toast.js
+++ b/packages/insomnia-app/app/ui/components/toast.js
@@ -14,11 +14,9 @@ import {
   getAppVersion,
   updatesSupported,
 } from '../../common/constants';
-import * as db from '../../common/database';
 import * as session from '../../account/session';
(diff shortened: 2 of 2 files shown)


dimitropoulos:
@pcjmfranken are you able to test this PR?  I am doing so now, anyway, but thought it would be ideal if you could verify that this fixes the problem you described.

packages/insomnia-app/app/ui/components/toast.js
```diff
@@ -112,26 +112,33 @@ class Toast extends React.PureComponent<Props, State> {
...
+        platform: getAppPlatform(),
+        app: getAppId(),
+        version: getAppVersion(),
+        updatesNotSupported: !updatesSupported(),
+        autoUpdatesDisabled: !settings.updateAutomatically,
+        disableUpdateNotification: settings.disableUpdateNotification,
+        updateChannel: settings.updateChannel,
+      };
+
+      const data = settings.enableAnalytics ? { ...trackingData, ...commonData } : commonData;
```
dimitropoulos: I was _gonna_ say that we should (also) just add this to `_trackEvent` in `app/common/analytics.js`, but when I looked I saw that it's already there (in the one function that `_trackEvent` calls, `_sendToGoogle`).

Still, I mention it just to mention that there should ideally be nowhere else where we need to check `enableAnalytics`.

reynolek: Yeah so what I believe happened is that this piece was originally written 4 years ago, I think with the initial goal of helping logged in users make sure they verified their account and such (at least in my understanding of the API code). It has since had other things bolted on to it.

reynolek:
@DMarby Mind just doing a sanity check for me on the API side? I read through it and I believe I have provided all the variable necessary to do the other parts of the notification endpoint.

pcjmfranken:
> @pcjmfranken are you able to test this PR? I am doing so now, anyway, but thought it would be ideal if you could verify that this fixes the problem you described.

@dimitropoulos Will check it out tomorrow when back in the office. This quick action response is greatly appreciated!

packages/insomnia-app/app/common/analytics.js
```diff
@@ -111,6 +111,11 @@ export async function getDeviceId(): Promise<string> {
 let segmentClient = null;
 
 export async function trackSegmentEvent(event: String, properties?: Object) {
+  const settings = await models.settings.getOrCreate();
```
dimitropoulos, line 114: my only concern here is that this queries the database every time and there may be a performance implication of doing so.  since we do this same check for every time google analytics and it hasn't been an issue so far I am hoping it won't be here, either.

reynolek (commented):
LGTM -- Its my own PR so I cant "approve" it, but I looked over @DMarby's changes and they look good.

dimitropoulos (approved):

dimitropoulos:
hey @pcjmfranken we didn't want to wait on this one, so although we merged it before you got a chance to report back - please feel free to do so anyway.  if there is still more work to be done on this we can gladly open up a new PR.

pcjmfranken:
@dimitropoulos I discovered yesterday that none of my mail-in replies have actually made it onto GH. Apologies for the late response. Good call merging this when you did. 

Just going through all of this again, and I'm still amazed by the quick pickup and solution! Thanks again!

dimitropoulos:
Thanks for that @pcjmfranken.

I love working at Kong for a lot of reasons but one of them is the high degree of integrity we apply to things like this.

We don't always get it right, and this issue is an example of that, but we sure-as-heck try to.

And thank YOU for reporting it. :)

Issue #3266, opened 2021-04-06
We're moving to TypeScript! :tada: 


dimitropoulos:
This umbrella issue will outline the transition of the entire insomnia codebase from Flow (or plain JS) to 1st-class idiomatic TypeScript.

# Phases

This work will be conducted in a few phases.

### Reconnaissance

Many projects before us have undergone this same transition.  Here are some examples:

1. Yarn https://github.com/yarnpkg/yarn/issues/6953
2. Jest https://github.com/facebook/jest/issues/7807
3. Graphql-js https://github.com/graphql/graphql-js/pull/2828, https://github.com/graphql/graphql-js/issues/2104
4. Luxon https://github.com/moment/luxon/pull/548
5. Gatsby https://github.com/gatsbyjs/gatsby/issues/21995
6. Expo https://github.com/expo/expo/issues/2164
7. Styled-Components https://github.com/styled-components/styled-components/pull/3460
8. codemirror-graphql https://github.com/graphql/graphiql/issues/957

### Spike

@develohpanda conducted the initial spike using https://github.com/Kong/insomnia/compare/22935eca82e0284dfe4653d1c6cc5991907957ff...122785f0cd5abf43adfc18db5f8f90b49cadc0f4

> Update: we later learned that this missed one little thing: decorators.  This spike was done on o2k, which doesn't contain any decorators.  More on this later, below.

## :beginner: Phase 1 - initial packages

Similar to almost all of the conversions listed above, we will be taking this conversion in a "bottom up" approach where we will be converting many of the things _without_ dependencies _first_.  This conversion will aim to be fairly final in nature.  Some ambient declarations will be necessary here and there for packages we can't upgrade (which, is a common thing to happen in a conversion like this so that we can have access to types), but in general the hope is do a fairly thorough job such that we won't have a step later to come back and fix a lot of things up.

### Phase 1 Checklist
| Task                 | PR | Status |
| -------------------- | -- | -- |
| `insomnia-xpath`     | https://github.com/Kong/insomnia/pull/3267 | Done
| `insomnia-url`       | https://github.com/Kong/insomnia/pull/3267 | Done
| `insomnia-prettify`  | https://github.com/Kong/insomnia/pull/3267 | Done
| `insomnia-importers` | https://github.com/Kong/insomnia/pull/3267 | Done
| `insomnia-cookies`   | https://github.com/Kong/insomnia/pull/3267 | Done
[...]

PR #3267, opened 2021-04-06
[TypeScript] Phase 1

dimitropoulos:
This ticket implements Phase 1 of the TypeScript conversion: https://github.com/Kong/insomnia/issues/3266

The following PRs were "pre-reviews" of all of the work contained in this PR.

1. insomnia-xpath https://github.com/dimitropoulos/insomnia/pull/1
2. insomnia-url https://github.com/dimitropoulos/insomnia/pull/2
3. insomnia-prettify https://github.com/dimitropoulos/insomnia/pull/3
4. insomnia-importers https://github.com/dimitropoulos/insomnia/pull/4
5. insomnia-cookies https://github.com/dimitropoulos/insomnia/pull/5
6. insomina-testing https://github.com/dimitropoulos/insomnia/pull/6
7. insomnia-inso https://github.com/dimitropoulos/insomnia/pull/7


packages/insomnia-inso/src/logger.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export type ModifiedConsola = ReturnType<typeof consola.create> & { __getLogs: () => LogsByType};
+
+const consolaLogger = consola.create({
+  reporters: [
+    new FancyReporter({
+      formatOptions: {
+        // @ts-expect-error something is wrong here, ultimately these types come from https://nodejs.org/api/util.html#util_util_inspect_object_options and `date` doesn't appear to be one of the options.
+        date: false,
```
develohpanda, line 27: Continuing from: https://github.com/dimitropoulos/insomnia/pull/7#discussion_r611285645

@reynolek @dimitropoulos these options come from https://github.com/unjs/consola/blob/4477e87a5153d31d22a2692565959d90599a6b61/src/reporters/fancy.js#L9-L16. The exported TS types appear to be incorrect here

.github/workflows/test.yml
```diff
@@ -28,6 +28,10 @@ jobs:
           node-version: ${{ steps.nvm.outputs.NVMRC }}
       - name: Bootstrap packages
         run: npm run bootstrap
+      - name: Lint
+        run: npm run lint
+      - name: Typecheck # TODO(TSCONVERSION) remove when flow goes away since bootstrap does the TypeScript build which in turn will error if there are type errors
```
develohpanda, line 33: 👍🏽

package.json
```diff
@@ -9,13 +9,13 @@
   },
   "homepage": "https://github.com/kong/insomnia#readme",
   "scripts": {
-    "lint": "eslint . --ext .js,.json",
+    "lint": "lerna run lint --parallel --stream",
```
develohpanda, line 13: I understand that linting in parallel might be faster but duplicating the `lint` script across all packages seems unnecessary/impractical when it comes to making a change and having to remember to update all lint scripts to keep them aligned.

I'm curious to understand your thought process behind making this change (if it's documented somewhere please point me to it). Is it so that each package can operate as a single/standalone entity?

package.json
```diff
@@ -29,7 +29,8 @@
     "app-package:smoke": "cross-env SMOKE_TEST=true npm run app-package",
     "test:smoke:build": "npm run test:build --prefix packages/insomnia-smoke-test",
     "test:smoke:package": "npm run test:package --prefix packages/insomnia-smoke-test",
-    "test:smoke:cli": "npm run test:cli --prefix packages/insomnia-smoke-test"
+    "test:smoke:cli": "npm run test:cli --prefix packages/insomnia-smoke-test",
+    "workspace": "$(npm bin)/lerna run --scope"
```
develohpanda: Would be nice to document how to using this script somewhere. Dev instructions exist in a few places (README.md and DEVELOPMENT.md) which we should try consolidate in the future. If it has already been documented, 👏🏽!

packages/insomnia-cookies/jest.config.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
```
develohpanda, line 1: So it looks like there's a `jest.config.js` at the root, and this is the first one I've come across in a particular package. Are there ways to extend config similar to ESLint so we have _one_ config? If not, I'm curious to understand the slight differences.

For example, the one at root uses `testMatch: ['**/*.test.ts']` but this file uses `testRegex: ['.+\\.test\\.ts$']`

packages/insomnia-cookies/package.json
```diff
@@ -4,9 +4,20 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist",
+    "package.json",
+    "READEM.md"
+  ],
   "scripts": {
-    "test": "jest --silent"
+    "bootstrap": "npm run build",
```
develohpanda, line 15: As an aside, since I'm expecting we might create more packages in the future, and now that the packages are being standardized (`index.ts`, `dist`, the scripts, root dir structure) it would be super useful now to create a template private package so that future packages follow the same conventions as best as possible 🤗

packages/insomnia-importers/.eslintignore
```diff
@@ -0,0 +1,2 @@
+bin
+dist
```
develohpanda, line 2: Does this file need to exist if these ignores are already in the root `.eslintignore`?

packages/insomnia-importers/src/convert.ts
```diff
@@ -0,0 +1,49 @@
...
+  type: {
+    id: string;
+    name: string;
+    description: string;
+  };
+  data: {
+    _type: 'export';
+    __export_format: 4;
+    __export_date: string;
+    __export_source: `insomnia.importers:v${string}`;
```
develohpanda, line 15: 🤯 TIL

dimitropoulos: :)

packages/insomnia-importers/.eslintignore
dimitropoulos: .. yeah, I think so, because we would need to otherwise point to the root eslintignore if it's run from this directory.  I played around with it for a while and eventually decided this was the least worst option available.

packages/insomnia-cookies/package.json
dimitropoulos: I agree.  I hope we can make this happen soon. :)

package.json
dimitropoulos: I did this because it seemed like an error to be honest.  it's a quite significant amount of time, and, afterall, the thing that ultimately runs this is the CI.  It was changed while splitting the typechecking step out on the CI.  More than happy to consider it out of scope and change it back if you'd like - lemme know.

package.json
dimitropoulos: I really would like to avoid it, because it's basically a replacement for `yarn workspace` in that I run `npm run workspace insomnia-inso test` like I would run `yarn workspace insomnia-inso test`.  It was done as a hold-me-over for when we can bring yarn into the project - which has functionality like this and so much more already documented and standardized out-of-the-box.

packages/insomnia-cookies/jest.config.js
dimitropoulos: yes, and in fact I did that at first, but everything got really confused.  I don't at this second remember what it was, but I definitely did try that.

The way to extend it ([recommended by the jest people themselves](https://kongstrong.slack.com/archives/C01N5M0ESCF/p1618236273005900)) is to import the 

```TS
const rootConfig = require('../../jest.config.js');

module.exports = {
  ...rootConfig,
  { otherThings }
}
```

I plan to take another look at this once TypeScript is fully in, but I felt it was out of scope to take it any further trying to get them to be less boiler-platey.

package.json
develohpanda: Ah fair, maybe even just stating that `npm run workspace` behaves the same as `yarn workspace` and linking to the yarn docs? There will be people who aren't familiar with yarn workspaces at all so might not know where to look, and the script itself looks a little cryptic 😆

packages/insomnia-cookies/jest.config.js
develohpanda: Gotcha, I'm on board with it being out of scope as well. Let's document in the following up tasks or create a linear ticket to make sure we come back and resolve this afterwards

packages/insomnia-importers/.eslintignore
develohpanda: > if it's run from this directory

Aha, this explains it

package.json
develohpanda: Just thinking out loud after our verbal convo - how do other lerna projects handle this? A single eslint call at the root, or individual ones in each package?

If we have eslintignore in each package now, I think we should be able to remove it from the root altogether because IDEs should correctly load from the appropriate package (right?)

dimitropoulos: ~predominantly (from my looking-around during Phase 0): they just use yarn.~ EDIT: I misunderstood your question.  They predominantly do something similar to what we had before, but it doesn't work for us right now from the root because different packages use different parsers and you can only have one parser per running eslint instance.  One way or the other, this is fairly optimal for us because it can do it multi-threaded this way so there's no reason I can see to remove it.

re: .eslintignore at the root: we can drastically reduce it, yes, but it will likely need to stay since, after all, there are still files at the root (including the root eslint config itself) that eslint checks.

package.json
dimitropoulos: I'll do whatever you think is best, but this is a band-aid and I really (ideally) don't want anyone to use it.  It's there for a very utilitarian reason: so that we _can_ run things for a single package.

And for that matter, I don't really know if it behaves the same as `yarn workspace` in all cases - I just know about this one way that I'm using it.  I wouldn't want to make the comparison because I don't know how people use `yarn workspace` outside of how _I_ use it.

I'm totally _for_ the spirit of what you're talking about, but this is a fairly dirty hack (in my mind) that I included because there's really no other way to do it (as long as we're not using yarn)... well... I guess I could just make a bash alias for myself locally and remove this from source code?

packages/insomnia-cookies/jest.config.js
dimitropoulos: Screenshot_20210413_113040

It's been documented in the scratchpad under `Things for Later` all along - so I'm going to resolve this.

After Phase 4, I'll make tickets for this and any remaining follow-ups from that scratchpad.

dimitropoulos: I've added it to "Things for later" (I thought it was there, but it wasn't) in the TypeScript Conversion Scratchpad.

Once Phase 4 is done, we'll loop back and make tickets from those items.

package.json
develohpanda: >  I really (ideally) don't want anyone to use it

Ah, okay. My understanding was it's something to be used for everyone, a helper of sorts! If you don't want people using it and it will be short-lived then I have no pressing reasons to document it. 🤗 Thanks for clarifying!

I vote for a bash alias and removing it altogether (if it works trivially for you), otherwise leave it in and we'll cull it eventually!

packages/insomnia-cookies/jest.config.js
develohpanda: Sounds good!

package.json
develohpanda: > different packages use different parsers

I'm not sure I follow here 🤔 Wouldn't they all use `@typescript-eslint/parser` (at phase 2/3)?

> One way or the other, this is fairly optimal for us because it can do it multi-threaded this way

For performance, yes I agree but that's at the cost of misdirection where `.eslintrc.js` is defined at the root, but `.eslintignore` is defined per package 😅

dimitropoulos: They will use the same parser by phase 2, but this is only a PR for phase 1.

Re: `.eslintignore` I'm not sure I see the downside of having individual `.eslintignore`s wherever we need them.  This keeps the packages more isolated from each other, which is ideal in a monorepo because it means that there's that much less you have to know about the other packages in order to examine a specific one.

package.json
dimitropoulos: I'll remove and make a bash alias, then.

dimitropoulos: `alias nw="$(npm bin)/lerna run --scope"`

it's removed.

packages/insomnia-importers/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "outDir": "./dist"
+  },
+  "include": [
+    "src",
+    "jest.config.js",
+    ".eslintrc.js",
+    "package.json"
+
+  ],
+  "exclude": []
```
develohpanda, line 15: Does this need to explicitly exclude `bin`? When I run build `insomnia-importers`, the `bin/cli` file gets updated

dimitropoulos:
closing this PR, as https://github.com/Kong/insomnia/pull/3370 contains this PR's code as well as (almost all of) Phase 2's work.

PR #3273, opened 2021-04-09
`strip_path` explicit support, and `x-kong-route-defaults` support

dimitropoulos:
closes: INS-122, closes INS-555, closes INS-554

### DIFF
--- a/packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
+++ b/packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
@@ -2,315 +2,418 @@
 import { generateServices } from '../services';
 import { parseSpec } from '../../index';
 
+
+/** This function is written in such a way as to allow mutations in tests but without affecting other tests. */
+const getSpec = () => JSON.parse(JSON.stringify({
+  openapi: '3.0',
+  info: { version: '1.0', title: 'My API' },
+  servers: [{ url: 'https://server1.com/path' }],
+  paths: {
+    '/cats': {
+      'x-kong-name': 'Cat stuff',
+      summary: 'summary is ignored',
+      post: {},
+    },
+    '/dogs': {
+      summary: 'Dog stuff',
+      get: {},
+      post: { summary: 'Ignored summary' },
+    },
+    '/birds/{id}': {
+      get: {},
+    },
+  },
+}));
+
+/** This function is written in such a way as to allow mutations in tests but without affecting other tests. */
+const getSpecResult = () => JSON.parse(JSON.stringify({
+  host: 'My_API',
+  name: 'My_API',
+  plugins: [],
+  path: '/path',
+  port: 443,
+  protocol: 'https',
+  tags: ['Tag'],
+  routes: [
+    {
+      name: 'My_API-Cat_stuff-post',
+      strip_path: false,
+      methods: ['POST'],
+      paths: ['/cats$'],
+      tags: ['Tag'],
+    },
+    {
+      name: 'My_API-dogs-get',
+      strip_path: false,
+      methods: ['GET'],
+      paths: ['/dogs$'],
+      tags: ['Tag'],
+    },
+    {
+      name: 'My_API-dogs-post',
+      strip_path: false,
+      methods: ['POST'],
+      paths: ['/dogs$'],
+      tags: ['Tag'],
+    },
+    {
+      name: 'My_API-birds_id-get',
+      strip_path: false,
+      methods: ['GET'],
+      paths: ['/birds/(?<id>[^\\/\\s]+)$'],
+      tags: ['Tag'],
+    },
+  ],
+}));
+
+const xKongPluginRequestValidator = 'x-kong-plugin-request-validator';
+const xKongRouteDefaults = 'x-kong-route-defaults';
+
 describe('services', () => {
   describe('generateServices()', () => {
     it('generates generic service with paths', async () => {
-      const api: OpenApi3Spec = await parseSpec({
-        openapi: '3.0',
-        info: { version: '1.0', title: 'My API' },
-        servers: [{ url: 'https://server1.com/path' }],
-        paths: {
-          '/cats': {
-            'x-kong-name': 'Cat stuff',
-            summary: 'summary is ignored',
-            post: {},
-          },
-          '/dogs': {
-            summary: 'Dog stuff',
-            get: {},
-            post: { summary: 'Ignored summary' },
-          },
-          '/birds/{id}': {
-            get: {},
-          },
-        },
-      });
+      const spec = getSpec();
 
-      const result = generateServices(api, ['Tag']);
-      expect(result).toEqual([
-        {
-          name: 'My_API',
-          protocol: 'https',
-          host: 'My_API',
-          port: 443,
-          path: '/path',
-          plugins: [],
-          tags: ['Tag'],
-          routes: [
-            {
-              name: 'My_API-Cat_stuff-post',
-              strip_path: false,
-              methods: ['POST'],
-              paths: ['/cats$'],
-              tags: ['Tag'],
-            },
-            {
-              name: 'My_API-dogs-get',
-              strip_path: false,
-              methods: ['GET'],
-              paths: ['/dogs$'],
-              tags: ['Tag'],
-            },
-            {
-              name: 'My_API-dogs-post',
-              strip_path: false,
-              methods: ['POST'],
-              paths: ['/dogs$'],
-              tags: ['Tag'],
-            },
-            {
-              name: 'My_API-birds_id-get',
-              strip_path: false,
-              methods: ['GET'],
-              paths: ['/birds/(?<id>[^\\/\\s]+)$'],
-              tags: ['Tag'],
-            },
-          ],
-        },
-      ]);
+      const specResult = getSpecResult();
+
+      const api = await parseSpec(spec);
+      expect(generateServices(api, ['Tag'])).toEqual([specResult]);
     });
 
     it('generates routes with request validator plugin from operation over path over global', async () => {
-      const api: OpenApi3Spec = await parseSpec({
-        openapi: '3.0',
-        info: { version: '1.0', title: 'My API' },
-        'x-kong-plugin-request-validator': { config: { parameter_schema: 'global' } }, // global req validator plugin
-        servers: [
-          {
-            url: 'https://server1.com/path',
-          },
-        ],
-        paths: {
-          '/dogs': {
-            summary: 'Dog stuff',
-            get: {},
-            post: {
-              summary: 'Ignored summary',
-              'x-kong-plugin-request-validator': {
-                // operation req validator plugin
-                config: { parameter_schema: 'operation' },
-              },
-            },
-          },
-          '/cats': {
-            summary: 'Dog stuff',
-            'x-kong-plugin-request-validator': { config: { parameter_schema: 'path' } }, // path req validator plugin
-            get: {},
-            post: {
-              'x-kong-plugin-request-validator': {
-                config: { parameter_schema: 'operation' }, // operation req validator plugin
-              },
-            },
-          },
-        },
-      });
+      const spec = getSpec();
 
-      const result = generateServices(api, ['Tag']);
-      expect(result).toEqual([
+      // global req validator plugin
+      spec[xKongPluginRequestValidator] = { config: { parameter_schema: 'global' }};
+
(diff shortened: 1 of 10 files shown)


packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -63,14 +64,18 @@ export function generateService(
...
       const fullPathRegex = pathVariablesToRegex(routePath);
       const route: DCRoute = {
+        ...rootRouteDefaults,
+        ...pathRouteDefaults,
         tags,
         name: generateRouteName(api, routePath, method),
         methods: [method.toUpperCase()],
         paths: [fullPathRegex],
-        strip_path: false,
+        strip_path: api.strip_path || false,
```
reynolek: So there are 2 things mentioned in that JIRA ticket -- 1) Provide a way to set `strip_path` to override the default and 2) The [docs](https://www.npmjs.com/package/openapi-2-kong#services) explicitly say that `strip_path` is always true, when clearly in the code its default is false.

I think at this point the best thing to do is to update the [README](https://github.com/Kong/insomnia/blob/develop/packages/openapi-2-kong/README.md#servicesroutes) to say that it defaults to false

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -97,19 +90,105 @@ describe('services', () => {
...
+            ...baseSpecResult,
+            routes: baseSpecResult.routes.map(route => ({
+              ...route,
+              foo: 'bar',
+            })),
+          },
+        ]);
+      });
+
+      it('will not override options on the route with `x-kong-route-defaults` at the root', async () => {
```
reynolek: Should this be `will override`?

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -97,19 +90,105 @@ describe('services', () => {
...
+            ...baseSpecResult,
+            routes: baseSpecResult.routes.map(route => ({
+              ...route,
+              ...(route.paths[0] === '/cats$' ? { foo: 'bar' } : {}),
+            })),
+          },
+        ]);
+      });
+
+      it('will not override options on the route with `x-kong-route-defaults` at the root', async () => {
```
reynolek: This test and the one above it have the exact same string description, which makes it hard to know what is different between the two.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -97,19 +90,105 @@ describe('services', () => {
...
+            ...baseSpecResult,
+            routes: baseSpecResult.routes.map(route => ({
+              ...route,
+              foo: route.paths[0] === '/cats$' ? 'foobar' : 'bar',
+            })),
+          },
+        ]);
+      });
+
+      it('allows overriding strip_path at the root', async () => {
```
reynolek: To my knowledge we dont support `strip_path` at the root. I believe we should only support it at the route definition level (or at the `x-kong-route-default` level)

reynolek (changes requested):

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: the linear ticket and the jira ticket got crossed in my mind, sorry, fixed in 6ad54dfe5c6fb799919e05b1b5de74bfc341587f

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: totally: was a mistake: fixed in b92d702bd1eb8739017b1b3cc4a486c1dc560327

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: fixed in b92d702bd1eb8739017b1b3cc4a486c1dc560327

packages/openapi-2-kong/src/declarative-config/services.js
dimitropoulos: fixed in 64498f939c28530bdc61712c850bd1a396408e24

reynolek (approved):

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -2,92 +2,85 @@
...
+
+const baseSpecResult = {
+  name: 'My_API',
+  url: 'https://server1.com/path',
+  plugins: [],
+  tags: ['Tag'],
+  routes: [
+    {
+      name: 'My_API-Cat_stuff-post',
+      strip_path: false,
```
Tieske: I don't think this is wise. `strip_path` can be specified in `x-kong-route-defaults` and if not specified at all it should simply pick up Kong defaults when synced to Kong by decK.

Adding those named properties here will create a tightly coupled implementation. We should try to keep that to a minimum. So only where actual logic is involved, eg. with the request-validator generation.

dwyt @develohpanda @dimitropoulos ?

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -97,19 +90,119 @@ describe('services', () => {
...
+            ...baseSpecResult,
+            routes: [...baseSpecResult.routes].map(route => ({
+              ...route,
+              foo: 'bar',
+            })),
+          },
+        ]);
+      });
+
+      it('x-kong-route-defaults only works at the root', async () => {
```
Tieske: To make it consistent with the plugins, I think this one should be possible on doc/path/operation level. Where the more specific level override the more generic ones (completely, so no merging of objects).

That would be consistent with the plugins, right @develohpanda ?

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -97,19 +90,119 @@ describe('services', () => {
...
+                foo: 'will be ignored, by design',
+              },
+            },
+          },
+        });
+
+        expect(generateServices(api, ['Tag'])).toEqual([baseSpecResult]);
+      });
+
+      it('allows overriding strip_path at the path level', async () => {
```
Tieske, line 438: only allowing `strip_path` to cascade through the levels doesn't make sense to me, and I think it also complicates the implementation (though that's hard for me to judge)

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
develohpanda: Yis that is how plugins behave now - no merging, most specific is used.

https://github.com/Kong/insomnia/blob/70738103d792ca7340464b585cad52842ca7be85/packages/openapi-2-kong/src/declarative-config/plugins.js#L179-L180

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: I strongly agree with you.  The trouble is that the prior implementation hardcoded in `strip_path` to `false` and there was no way to specify it or not.  To change this behavior now would be a breaking change, we feel.

https://github.com/Kong/insomnia/blob/develop/packages/openapi-2-kong/src/declarative-config/services.js#L85

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: I implemented it this way originally, and it was decided against because it kinda defeats the semantics of a top-level default, which is different from other properties.  @reynolek can say more, but that was the general idea.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: I agree, but [as I said above](https://github.com/Kong/insomnia/pull/3273#discussion_r618484885), the issue ultimately is that we had this hardcoded to false for all paths before which kinda locks us into this.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
develohpanda: FWIW I sort of agree with Thijs here about specificity. Use the top-level default but override when set at a more specific location

reynolek: Lets go with making it consistent with other plugins. Looks like I directed you away from that earlier, not realizing we did it other ways in other places.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
Tieske: actually I was wrong. It should be set to `false`, if not specified in `x-kong-route-defaults`, only if `x-kong-route-defaults` sets it to `true` it should be `true`, in all other cases it should be `false`.

Reason being; the part of the path that is matched by Kong, will be stripped. This is because Kong treats it as a path-prefix. But the nature of OpenAPI specs is that the entire path is specified, not just a prefix. So if stripped, it would mean that the entire path is stripped, and hence every endpoint would be reduced to only `/`, which obviously is useless.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
Tieske: agreed, merging objects can have weird edge cases. Let's put the can of worms back on the shelf 😄

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
Tieske: yep, should be as per other comments, full set of defaults, no merging. And strip_path only goes to `true` if there is a `true` in the defaults, if absent it goes to `false`.

dimitropoulos: should be updated now.  enough has changed anyway that I think another top-level look is necessitated.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: should be updated now.  enough has changed anyway that I think another top-level look is necessitated.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
dimitropoulos: should be updated now.  enough has changed anyway that I think another top-level look is necessitated.

packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -81,16 +82,34 @@ export function generateService(
         continue;
       }
 
+      // $FlowFixMe
+      const operationRouteDefaults = pathItem['x-kong-route-defaults'];
```
Tieske: This is actually the "path" default

packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -81,16 +82,34 @@ export function generateService(
...
+      const pathRouteDefaults = api.paths['x-kong-route-defaults'];
```
Tieske: this is not right, a path default on that array does not make sense. The array has a 1-on-1 relation ship with the root document, hence from a scoping perspective this would be identical to `x-kong-route-defaults` on root level.

Tieske (changes requested):
I think we should close this and replace it with #3319

packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -42,9 +42,18 @@ export function generateService(
...
+  const serviceDefaults = api['x-kong-service-defaults'] || {};
+  if (typeof serviceDefaults !== 'object') {
+    throw new Error(`expected 'x-kong-service-defaults' to be an object`);
+  }
+
   const service: DCService = {
+    ...serviceDefaults,
     name,
-    // remove semicolon ie. convert https: to https
+    // remove semicolon i.e. convert `https` to `https`
```
Tieske:
```suggestion
    // remove semicolon i.e. convert `https:` to `https`
```

Tieske (approved):

PR #3278, opened 2021-04-12
Prompt existing users about analytics

DMarby:
Prompts existing users to review their analytics sharing settings once, without changing the onboarding flow for new users.

I implemented this by extracting the analytics step from the onboarding into it's own component, and added a new activity to show it for this usecase, please let me know if you think there's a better approach.

Wasn't sure what to put as the title/subtitle, or whether any of the text should be changed, so I made it identical to the onboarding flow, please let me know if you think these should be changed to something else @reynolek @erickrico.

image

Fixes INS-609

### DIFF
--- a/packages/insomnia-app/app/main.development.js
+++ b/packages/insomnia-app/app/main.development.js
@@ -150,7 +150,11 @@ async function _createModelInstances() {
 async function _updateFlags({ launches }: Stats) {
   const firstLaunch = launches === 1;
   if (firstLaunch) {
-    await models.settings.patch({ hasPromptedOnboarding: false, hasPromptedAnalytics: true });
+    await models.settings.patch({
+      hasPromptedOnboarding: false,
+      // Don't show the analytics preferences prompt as it is part of the onboarding flow
+      hasPromptedAnalytics: true,
+    });
   }
 }
 
--- a/packages/insomnia-app/app/main.development.js
+++ b/packages/insomnia-app/app/main.development.js
@@ -150,7 +150,7 @@ async function _createModelInstances() {
 async function _updateFlags({ launches }: Stats) {
   const firstLaunch = launches === 1;
   if (firstLaunch) {
-    await models.settings.patch({ hasPromptedOnboarding: false });
+    await models.settings.patch({ hasPromptedOnboarding: false, hasPromptedAnalytics: true });
   }
 }
 
--- a/packages/insomnia-app/app/ui/components/wrapper-analytics.js
+++ b/packages/insomnia-app/app/ui/components/wrapper-analytics.js
@@ -1,0 +1,31 @@
+// @flow
+
+import * as React from 'react';
+import type { WrapperProps } from './wrapper';
+import OnboardingContainer from './onboarding-container';
+import Analytics from './analytics';
+import { useDispatch } from 'react-redux';
+import { setActiveActivity } from '../redux/modules/global';
+import { getAppLongName, getAppSynopsis, ACTIVITY_HOME } from '../../common/constants';
+
+type Props = {|
+  wrapperProps: WrapperProps,
+|};
+
+const WrapperAnalytics = ({ wrapperProps }: Props) => {
+  const reduxDispatch = useDispatch();
+  const navigateHome = React.useCallback(() => {
+    reduxDispatch(setActiveActivity(ACTIVITY_HOME));
+  }, [reduxDispatch]);
+
+  return (
+    <OnboardingContainer
+      wrapperProps={wrapperProps}
+      header={'Welcome to ' + getAppLongName()}
+      subHeader={getAppSynopsis()}>
+      <Analytics wrapperProps={wrapperProps} handleDone={navigateHome} />
+    </OnboardingContainer>
+  );
+};
+
+export default WrapperAnalytics;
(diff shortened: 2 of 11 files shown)


packages/insomnia-app/app/main.development.js
```diff
@@ -150,7 +150,7 @@ async function _createModelInstances() {
 async function _updateFlags({ launches }: Stats) {
   const firstLaunch = launches === 1;
   if (firstLaunch) {
-    await models.settings.patch({ hasPromptedOnboarding: false });
+    await models.settings.patch({ hasPromptedOnboarding: false, hasPromptedAnalytics: true });
```
reynolek: Is this supposed to be `true`?

DMarby: Yep, it gets set to true on the first launch of the app here, so that new users don't see the it, as they select analytics preferences during the onboarding flow.

reynolek: rgr make sense.

reynolek (approved):
Approved from Product Perspective.

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -276,7 +277,10 @@ export function setActiveActivity(activity: GlobalActivity) {
       models.settings.patch({ hasPromptedToMigrateFromDesigner: true });
       break;
     case ACTIVITY_ONBOARDING:
-      models.settings.patch({ hasPromptedOnboarding: true });
+      models.settings.patch({ hasPromptedOnboarding: true, hasPromptedAnalytics: true });
```
develohpanda: Similar to Eric's question, `hasPromptedAnalytics` being set to true here seems odd but makes sense, so a short code comment might be worthwhile

packages/insomnia-app/app/ui/components/wrapper-analytics.js
```diff
@@ -0,0 +1,31 @@
...
+import { getAppLongName, getAppSynopsis, ACTIVITY_HOME } from '../../common/constants';
+
+type Props = {|
+  wrapperProps: WrapperProps,
+|};
+
+const WrapperAnalytics = ({ wrapperProps }: Props) => {
+  const reduxDispatch = useDispatch();
+  const navigateHome = React.useCallback(() => {
+    reduxDispatch(setActiveActivity(ACTIVITY_HOME));
```
develohpanda, line 18: Since we have `wrapperProps` and we have `handleSetActiveActivity` on it I would have been inclined to just use `wrapperProps.handleSetActiveActivity`.

In saying that I think dispatching to Redux (and slowly dismantling `WrapperProps`) is the right approach long term to prevent a waterfall of props so we can call this the start of the dismantling of `WrapperProps` 😆 

No action requested here, it's up to you if you want to leverage `wrapperProps.handleSetActiveActivity` and remove the Redux dispatch for now

packages/insomnia-app/app/ui/redux/modules/__tests__/global.test.js
```diff
@@ -298,5 +306,18 @@ describe('global', () => {
       await store.dispatch(initActiveActivity());
       expect(store.getActions()).toEqual([expectedEvent]);
     });
+
+    it('should prompt to change analytics settings', async () => {
```
develohpanda, line 322: 🥳

develohpanda (commented):
> I implemented this by extracting the analytics step from the onboarding into it's own component, and added a new activity to show it for this usecase, please let me know if you think there's a better approach.

Sounds like a reasonable approach! I've just done a prelim review and all looks good to me (haven't tested locally)

packages/insomnia-app/app/ui/redux/modules/global.js
DMarby: Done

packages/insomnia-app/app/ui/components/wrapper-analytics.js
DMarby: Fair enough! I based this on how `goToNextActivity` was called in `MigrationBody`, but wasn't sure what the correct approach was.

develohpanda: Yeah fair play! That function was a new one I created for migration and I didn't want to pipe it through WrapperProps so used redux for it 😅 slowly but surely!

packages/insomnia-app/app/ui/redux/modules/global.js
```diff
@@ -235,8 +236,18 @@ export function loadRequestStop(requestId) {
...
+        return ACTIVITY_ONBOARDING;
+      }
+
+      // Has not seen the analytics prompt? Go to it
+      if (!settings.hasPromptedAnalytics) {
+        return ACTIVITY_ANALYTICS;
+      }
+
+      // Otherwise, go to home
+      return ACTIVITY_HOME;
```
develohpanda, line 250: Nice 🙌🏽

develohpanda (approved):
Tested a few different workflows locally and all LGTM! 🎉 

Note: I pushed [`a2be24f` (#3278)](https://github.com/Kong/insomnia/pull/3278/commits/a2be24f72e39496dd86f14c514b43d87f40b23f2) adding a couple minor tests that I didn't want to block you with 🙂 Two of them will be implemented better once we go to TS.

sonicyeti (approved):
Nice flow updates!

PR #3283, opened 2021-04-14
Request validator plugin respects existing properties and generates required ones

### LINKED ISSUE #2867: [Bug] openapi-2-kong ignoring verbose_response property of request validation plugin when generating config
**Describe the bug**
It appears that inso / openapi-2-kong is ignoring verbose_response property of request validation plugin when generating config for Kong.

**To Reproduce**

$ inso generate config spec.yaml --output generated.yaml

**Expected behavior**

As verbose_response property of request validation plugin is included in the OAS, it should appear in the generated Kong YAML.

 - OS: Linux Mint 20
 - Installation Method: npm
 - App Version 2.2.25

**Additional context**
Please see the attachment for OAS, expected and actual outputs.

[00019330.zip](https://github.com/Kong/insomnia/files/5591963/00019330.zip)

develohpanda:
The [request-validator plugin](https://docs.konghq.com/hub/kong-inc/request-validator/#parameters) has several options to be send under the `config` object. These parameters are:
- `parameter_schema`
- `body_schema`
- `allowed_content_types`
- `verbose_response`

Out of these four, the first three can be generated or be user-defined, and the fourth must be user defined.

In an OpenAPI spec, the user may specify an object with none or some of the config parameters. If a user defines a config parameter, then that config parameter will be used and not generated. If a parameter that can be generated is not defined, then O2K will attempt to generate.

Here is an example spec with the plugin defined on each operation. This is not an exhaustive example, but should server as a good base to begin testing from.

```yaml
openapi: 3.0.2

info:
  title: Example
  version: 1.0.0

servers:
  - url: http://backend.com/path

paths:
  /params:
    get:
      x-kong-plugin-request-validator:
        enabled: true
        config: 
          verbose_response: true
      parameters:
        - in: path
          name: userId
          schema:
            type: integer
          required: true
  /body:
    post:
      x-kong-plugin-request-validator:
        enabled: true
        config: 
          verbose_response: true
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/jsonSchema'
          application/xml:
            schema:
              $ref: '#/components/schemas/xmlSchema'

components:
  schemas:
    jsonSchema:
      type: object
      properties:
        id:
          type: integer
        name:
          type: string
    xmlSchema:
      type: object
      properties:
        prop:
          type: integer
```

If O2K is unable to generate the `body_schema` (from the operation request JSON body) or `parameter_schema` (from the operation parameters), then a default "pass-all" config should be applied, where the `body_schema` is set to `'{}'`.

See the unit tests added, which checks of these rules. Also see [FTI-2252](https://konghq.atlassian.net/browse/FTI-2252). This doesn't address all of the requirements outlined in the FTI, there is a subsequent PR to support plugins (request validator or otherwise) at other levels of the spec (root, server, path, operation).

Fixes INS-457, closes #2867

### DIFF
--- a/packages/openapi-2-kong/src/declarative-config/__tests__/plugins.test.js
+++ b/packages/openapi-2-kong/src/declarative-config/__tests__/plugins.test.js
@@ -83,7 +83,7 @@ describe('plugins', () => {
           parameter_schema: [parameterSchema],
           body_schema: '[{"name":{"type": "string", "required": true}}]',
           verbose_response: true,
-          allowed_content_types: 'application/json',
+          allowed_content_types: ['application/json'],
         },
       };
 
@@ -106,7 +106,7 @@ describe('plugins', () => {
           body_schema: '[{"name":{"type": "string", "required": true}}]',
           // The following properties are missing
           // verbose_response: true,
-          // allowed_content_types: 'application/json',
+          // allowed_content_types: ['application/json'],
         },
       };
 
@@ -122,20 +122,20 @@ describe('plugins', () => {
     });
 
     describe('parameter_schema', () => {
-      it('should return empty array for parameter schema if no parameters exist in operation', () => {
+      it('should not add parameter_schema if no parameters present', () => {
         const plugin = {
           enabled: true,
           config: {},
         };
 
-        const operation = {};
+        const generated1 = generateRequestValidatorPlugin(plugin, {});
+        const generated2 = generateRequestValidatorPlugin(plugin, { parameters: [] });
 
-        const generated = generateRequestValidatorPlugin(plugin, operation);
-
-        expect(generated).toStrictEqual({
-          name: 'request-validator',
-          enabled: plugin.enabled,
-          config: { version: 'draft4', parameter_schema: [] },
+        expect(generated1.config).toStrictEqual({
+          version: 'draft4',
+        });
+        expect(generated2.config).toStrictEqual({
+          version: 'draft4',
         });
       });
 
@@ -194,8 +194,8 @@ describe('plugins', () => {
       });
     });
 
-    describe('body_schema', () => {
-      it('should not add body_schema if no body present', () => {
+    describe('body_schema and allowed_content_types', () => {
+      it('should not add body_schema or allowed_content_types if no body present', () => {
         const plugin = {
           config: {},
         };
@@ -206,7 +206,6 @@ describe('plugins', () => {
 
         expect(generated.config).toStrictEqual({
           version: 'draft4',
-          parameter_schema: [],
         });
       });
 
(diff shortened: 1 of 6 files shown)


packages/openapi-2-kong/src/declarative-config/__tests__/plugins.test.js
```diff
@@ -63,4 +63,244 @@ describe('plugins', () => {
       });
     });
   });
+
+  describe('generateRequestValidatorPlugin()', () => {
```
develohpanda, line 67: image

packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -42,6 +42,7 @@ export function generateService(
   for (const routePath of Object.keys(api.paths)) {
     const pathItem: OA3PathItem = api.paths[routePath];
 
+    // TODO: Add path plugins to route
```
develohpanda, line 45: Completed and comment removed in a subsequent PR

packages/openapi-2-kong/src/declarative-config/__tests__/plugins.test.js
```diff
@@ -63,4 +63,244 @@ describe('plugins', () => {
...
+
+        const generated = generateRequestValidatorPlugin(plugin, operation);
+
+        expect(generated.config).toStrictEqual({
+          version: 'draft4',
+          body_schema: '{}',
+        });
+      });
+
+      it('should throw error if no operation request body content defined', () => {
```
Tieske: not sure what the effect is here.

Consider 200 operations and a request-validator defined.
- 100 POST ones expecting a body (and having it defined)
- 100 GET ones not expecting a body (and hence not defined in the OAS)

Then this would throw errors on the 100 GETs, right? In that case a user cannot specify the root level plugin config, but needs to add 100 specific ones for the POST ones. That will become painful...

develohpanda: Gotcha, should it silently continue if it can't find the appropriate properties? This is me writing a test for existing behaviour but can certainly change it if it's wrong

Tieske: yes. I think that makes sense.

develohpanda: Fixed in [`e41e047` (#3283)](https://github.com/Kong/insomnia/pull/3283/commits/e41e047ac240d9f8af15718674091ead3e824ea8). I also updated the behavior of `parameters_schema` generation to not throw errors.

sonicyeti (approved):

PR #3286, opened 2021-04-14
Custom & default tagging for plugins

### LINKED ISSUE #2682: Insomnia Designer does not tag generated declarative Kong config with tags in the spec
**Describe the bug**
When writing a spec and leveraging tags, generating Kong configuration from the spec does not tag the configuration with the tags in the spec.  

This leads to a situation where teams working on different specs, generating declarative config and running deck to sync config could overwrite other teams configuration.  

**To Reproduce**
1. Use ID or inso with a spec with tags to generate a Kong declarative config.  
2. Notice the tags in the spec are not available in the Kong declarative config.  

**Expected behavior**
The generated Kong declarative config should show the tags in spec.   This will assure when running deck sync that only the appropriate objects are worked on.

sonicyeti:
Abstracting out our default 'OAS3_import' tag to be re-used, applying custom tags to plugins while including the default OAS3_import tag.

Fixes INS-611, closes https://github.com/Kong/insomnia/issues/2682

### DIFF
--- a/packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json
+++ b/packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json
@@ -103,7 +103,7 @@
           "strip_path": false,
           "tags": ["OAS3_import", "OAS3file_httpbin.yaml"],
           "name": "httpbin-cookies-get",
-          "plugins": [{"name": "correlation-id"}],
+          "plugins": [{"config": {"tags": ["OAS3_import"]}, "name": "correlation-id"}],
           "paths": ["/cookies$"]
         },
         {
--- a/packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json
+++ b/packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json
@@ -101,9 +101,9 @@
         {
           "methods": ["GET"],
           "strip_path": false,
-          "tags": ["OAS3_import", "OAS3file_httpbin.yaml"],
           "name": "httpbin-cookies-get",
-          "plugins": [{"config": {"tags": ["OAS3_import"]}, "name": "correlation-id"}],
+          "tags": ["OAS3_import", "OAS3file_httpbin.yaml"],
+          "plugins": [{"name": "correlation-id", "tags": ["OAS3_import"]}],
           "paths": ["/cookies$"]
         },
         {
--- a/packages/openapi-2-kong/src/declarative-config/plugins.js
+++ b/packages/openapi-2-kong/src/declarative-config/plugins.js
@@ -25,11 +25,15 @@ export function generatePlugins(item: Object, generator: GeneratorFn): Array<DCP
 export function generatePlugin(key: string, value: Object): DCPlugin {
   const plugin: DCPlugin = {
     name: value.name || getPluginNameFromKey(key),
-    config: value.config ? value.config : { tags: [''] },
+    config: value.config ? value.config : {},
   };
-
   // Add tags to plugins while appending defaults tags
-  plugin.config.tags ? plugin.config.tags.push(...defaultTags) : (plugin.config.tags = defaultTags);
+  // make flow happy
+  if (!plugin.config) {
+    plugin.config = {};
+  }
+
+  plugin.config.tags = plugin.config.tags ? defaultTags.concat(plugin.config.tags) : defaultTags;
 
   return plugin;
 }
--- a/packages/openapi-2-kong/src/declarative-config/plugins.js
+++ b/packages/openapi-2-kong/src/declarative-config/plugins.js
@@ -25,7 +25,7 @@ export function generatePlugins(item: Object, generator: GeneratorFn): Array<DCP
 export function generatePlugin(key: string, value: Object): DCPlugin {
   const plugin: DCPlugin = {
     name: value.name || getPluginNameFromKey(key),
-    config: value.config ? value.config : {},
+    config: value.config,
   };
   // Add tags to plugins while appending defaults tags
   // make flow happy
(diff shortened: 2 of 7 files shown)


packages/openapi-2-kong/src/declarative-config/plugins.js
```diff
@@ -25,12 +25,16 @@ export function generatePlugins(item: Object, generator: GeneratorFn): Array<DCP
...
     name: value.name || getPluginNameFromKey(key),
+    config: value.config ? value.config : {},
   };
-
-  if (value.config) {
-    plugin.config = value.config;
+  // Add tags to plugins while appending defaults tags
+  // make flow happy
+  if (!plugin.config) {
+    plugin.config = {};
```
develohpanda, line 33: This is redundant if you already have the check on line 28

packages/openapi-2-kong/src/declarative-config/plugins.js
```diff
@@ -25,12 +25,16 @@ export function generatePlugins(item: Object, generator: GeneratorFn): Array<DCP
 export function generatePlugin(key: string, value: Object): DCPlugin {
   const plugin: DCPlugin = {
     name: value.name || getPluginNameFromKey(key),
+    config: value.config ? value.config : {},
```
develohpanda: This fallback is redundant if you have the check on line 32 below

```suggestion
    config: value.config,
```

develohpanda (changes requested):
There are security plugins which are not being assigned tags (and should be). The following is one example but I think all of the plugins in the following fixture result should contain a tag.

https://github.com/Kong/insomnia/blob/eb4ae99f8428e99b8322e5c1b4125ddc14d316fb/packages/openapi-2-kong/src/__fixtures__/security.expected.json#L23

packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json
```diff
@@ -103,7 +103,7 @@
           "strip_path": false,
           "tags": ["OAS3_import", "OAS3file_httpbin.yaml"],
           "name": "httpbin-cookies-get",
-          "plugins": [{"name": "correlation-id"}],
+          "plugins": [{"config": {"tags": ["OAS3_import"]}, "name": "correlation-id"}],
```
develohpanda: Is the tag applied to the correct location? I don't think so. The tag needs to be on the parent object, not in the config property. 

```
{
  "plugins": [{
    "config": {...},
    "tags": ["OAS3_import"],
    "name": "correlation-id"
  }]
}
````

**Review by develohpanda — CHANGES_REQUESTED** [review]

File: packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json (hunk 3 above)
Comment (sonicyeti) [reply packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json]: We need to support tags on the plugin's config object per [FTI-2444](https://konghq.atlassian.net/browse/FTI-2444). Notice the other issue regarding tags in plugin root. 

Current output being (covering 2444):

Screen Shot 2021-04-16 at 9 32 17 AM

Comment (develohpanda) [reply packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json]: [FTI-2444](https://konghq.atlassian.net/browse/FTI-2444) doesn't seem related to this PR? This PR looks like it's fixing [FTI-2305](https://konghq.atlassian.net/browse/FTI-2305), which is reflected from #2682. The YAML example in FTI-2305 has lost all indenting from what was in GitHub (which is crucial for YAML). Worth confirming with someone but to the best of my knowledge tags should not be part of the _config_ object (which is specific to a plugin), but should be on the parent item (because tags is a common property which exists on all entities).

Comment (sonicyeti) [reply packages/openapi-2-kong/src/__fixtures__/httpbin.expected.json]: Ok I'll circle back with Thjis for clarification. The linear ticket (INS-611) touches on the relationship between the above tickets, it's not capture in this PR though.

File: packages/openapi-2-kong/src/declarative-config/plugins.js [hunk 4]
```diff
@@ -25,11 +25,12 @@ export function generatePlugins(item: Object, generator: GeneratorFn): Array<DCP
 export function generatePlugin(key: string, value: Object): DCPlugin {
   const plugin: DCPlugin = {
     name: value.name || getPluginNameFromKey(key),
+    tags: value.tags,
+    config: value.config,
```
develohpanda: 1. Now that tags are no longer placed on the `config` object this logic to handle `value.config` should return to it's previous implementation. 
2. the tags logic can be simplified because we don't need to work with the config object anymore
	  ```js
	  plugin.tags = [
	    ...defaultTags,
	    ...(value.tags ?? []),
	  ];
	  ```
3. I've added some notes for a follow-up task here: https://linear.app/insomnia/issue/INS-175#comment-81a90971

packages/openapi-2-kong/src/index.js
```diff
@@ -30,7 +31,7 @@ export async function generateFromString(
   tags: Array<string> = [],
 ): Promise<ConversionResult> {
   const api: OpenApi3Spec = await parseSpec(specStr);
-  return generateFromSpec(api, type, ['OAS3_import', ...tags]);
+  return generateFromSpec(api, type, [defaultTags.join(), ...tags]);
```
develohpanda: Is the `.join()` necessary? Doing the following should be enough

```suggestion
  return generateFromSpec(api, type, [...defaultTags, ...tags]);
```

Or even
```suggestion
  return generateFromSpec(api, type, defaultTags.concat(tags));
```

develohpanda (changes requested):
Some cleanup left after changes 🧹

develohpanda:
@sonicyeti happy to pair for resolving conflicts here as they arise from one of my recent PRs

develohpanda (approved):
LGTM - I have created a roll-up ticket (INS-629) regarding tags on plugins and marked which parts of that are covered by this PR

PR #3293, opened 2021-04-15
Respect request validator plugin defined at the document root and hierarchy of plugin definitions

develohpanda:
### Rules

The rules for which plugin to use and where to place it in the generated spec are the following:
- all plugins on a `route` get derived from `path -> operation`
- all plugins on a `service` get derived from `root`

There is one special case to do with the request-validator plugin. This is:
- request validator plugin on a `route` gets derived from `root -> path -> operation`

### Example
This means, if a plugin (eg. `x-kong-plugin-key-auth`) exists on the `root`, `path` _and_ `operation` with different config at each level, in the generated declarative config, the:
- `service` entity will have its plugin derived from the `root`
- `route` entity will have its plugin derived from the `operation` (because `operation` takes precedence over `path`)

If the request validator plugin (`x-kong-plugin-request-validator`) exists on the `root`, in the generated declarative config, the:
- `service` entity will have it's plugin derived from the `root` (or be a default config if nothing is user defined)
- `route` entity will have its plugin derived from the `root`

Similarly, if the request validator plugin (`x-kong-plugin-request-validator`) exists on the `root` _and_ `operation`, in the generated declarative config, the:
- `service` entity will have it's plugin derived from the `root` (or be a default config if nothing is user defined)
- `route` entity will have its plugin derived from the `operation`

Fixes INS-610, fixes INS-619


packages/openapi-2-kong/src/common.js
```diff
@@ -147,3 +147,20 @@ export function joinPath(p1: string, p2: string): string {
 
   return `${p1}/${p2}`;
 }
+
+// Select first unique instance of an array item depending on the property selector
+export function distinctByProperty<T>(arr: Array<T>, propertySelector: (item: T) => any): Array<T> {
```
develohpanda, line 153: Moved from `packages/openapi-2-kong/src/kubernetes/plugins.js`

packages/openapi-2-kong/src/__tests__/common.test.js
```diff
@@ -237,4 +238,40 @@ describe('common', () => {
       expect(result.pathname).toBe('/api/v1');
     });
   });
+
+  describe('distinctByProperty()', () => {
```
develohpanda, line 244: Moved from `packages/openapi-2-kong/src/kubernetes/__tests__/plugins.test.js`

packages/openapi-2-kong/src/declarative-config/__tests__/plugins.test.js
```diff
@@ -15,52 +15,35 @@ describe('plugins', () => {
...
           name: 'key-auth',
           config: {
             key_names: ['x-api-key'],
           },
         },
-      ]);
-    });
-  });
-
-  describe('generatePlugin()', () => {
```
develohpanda, line 30: Is only exported for this test, which is covered by a lot of upstream tests already.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -67,6 +67,332 @@ describe('services', () => {
       ]);
     });
 
+    it('generates routes with request validator plugin from operation over path over global', async () => {
```
develohpanda, line 70: image

sonicyeti (approved):
LGTM!!!

Tieske:
>  in the Declarative Config only the first server is used

I discussed with @sonicyeti that all servers (that differ only by `host` or `port`, differences by `protocol` or `path` should be an error) should be included as `targets` in the `upstream`. Just mentioning it here, since I do not know what the impact is exactly.

> Note: `root` and `server` plugins are treated similarly because in the Declarative Config only the first server is used. However, if the same plugin is defined at the `root` _and_ `server`, the config will be derived from the `server`.

I do not get the distinction between `servers` and `root` because `servers` essentially is a property of `root` so that is always a 1-on-1 relation ship. So imo adding anything to `servers` doesn't make sense, everything should be defined at `root`.

Expanding on that;
- currently only the top-level `servers` is used, and `servers` defined on `path` or `operation` level are ignored.
- we create 1 `service` and 1 `upstream` per spec. But technically a `service` is derived from `servers` (as in the accompanying `upstream`.
- if we ever support `servers` on `operation` and `path` levels, those would become additional `service` and `upstream` entities. And also here; `servers` is a property of the `operation` and/or `path`, so also a 1-on-1 relationship.

develohpanda:
> > in the Declarative Config only the first server is used
> 
> I discussed with @sonicyeti that all servers (that differ only by `host` or `port`, differences by `protocol` or `path` should be an error) should be included as `targets` in the `upstream`. Just mentioning it here, since I do not know what the impact is exactly.

Right, so its a requirement for all servers to follow that rule (which we should enforce in the generator logic), therefore allowing for a _single_ `upstream` with multiple targets and a _single_ `service` to be created in the Declarative Config.

> if we ever support servers on operation and path levels, those would become additional service and upstream entities

My running understanding/assumption was that each server in the OAS root should map to an individual `upstream` and an individual `service` in the Declarative Config. Meaning, multiple servers in the OAS root resulting in multiple `upstreams`/`services` in the generated config (going by how the K8s generation works right now, which isn't a great example either). But according to this bullet point, it says that `servers` being defined at a different _level_ is what controls whether new entities are created.

For instance, if an operation contains a `servers` property with multiple servers, a new `service`/`upstream` would be created in the DC _but_ each server in that property (`operation.servers`) can only differ by `host` or `port` and each server will be created as a _target_ in the new upstream.

> technically a `service` is derived from `servers` (as in the accompanying upstream.

Essentially, you're saying a `service`/`upstream` is derived from the `servers` property on a parent (be it `root`, `path` or `operation`) and the array items contained within, not _per entry_.

If I am understanding your explanation correctly, I can see why plugins on a `server` in the OAS don't actually make sense and shouldn't be there (and that the K8s generator is incorrect).

Tieske:
> If I am understanding your explanation correctly, I can see why plugins on a server in the OAS don't actually make sense and shouldn't be there (and that the K8s generator is incorrect).

@develohpanda that understanding is correct

PR #3295, opened 2021-04-15
Update the segment write key for production

DMarby:


reynolek (approved):

reynolek:
This change relates to #3255 -- Creating a new source in segment so that once the new version of the App is deployed that properly honors users data preferences, only clean data exists.

PR #3296, opened 2021-04-16
chore(o2k) better error message

Tieske:
Parameter validation requires either 'schema' or 'content' to be set.
The validation assumed that if it wasn't 'schema', it was automatically
'content'. This change provides a better error message.

### DIFF
--- a/packages/openapi-2-kong/src/declarative-config/plugins.js
+++ b/packages/openapi-2-kong/src/declarative-config/plugins.js
@@ -44,7 +44,12 @@ export function generateRequestValidatorPlugin(obj: Object, operation: OA3Operat
   if (operation.parameters) {
     for (const p of operation.parameters) {
       if (!(p: Object).schema) {
-        throw new Error("Parameter using 'content' type validation is not supported");
+        if ((p: Object).content) {
+          throw new Error("Parameter validation of type 'content' is not supported");
+        }
+        throw new Error(
+          "Parameter validation requires either 'schema' or 'content' (only 'schema' is supported)",
+        );
       }
       config.parameter_schema.push({
         in: (p: Object).in,


develohpanda:
@Tieske this error message has been removed entirely in #3283. I followed the same approach such that if a scheme can be generated from the existing data then generate it, otherwise move on rather than presenting a failure.

In the case of the parameter_scheme is it better to stop and present a failure?

Tieske:
good point. Spec says either `schema` or `content` must be present.

problem might be if someone has `content` specified, and if we then silently ignore it, whilst the user expects it to be added. Is that what it currently does in #3283?

Tieske:
even if `schema` is not present, the `required` flag might still be there, see https://swagger.io/specification/#parameter-object (I think we can safely ignore ` allowEmptyValue`)

develohpanda:
> problem might be if someone has `content` specified, and if we then silently ignore it, whilst the user expects it to be added. Is that what it currently does in #3283?

Yeah that's what happens in #3283. To me it doesn't warrant a fatal failure, nor does it warrant a silent ignore, but we don't really have a way to present warnings in the current state. 

In 3283, if there are three params - two with schema and one with content - it's just going to ignore the param with content specified, whereas previously it would just throw an error when encountering an unsupported param

develohpanda:
> even if `schema` is not present, the `required` flag might still be there, see https://swagger.io/specification/#parameter-object (I think we can safely ignore ` allowEmptyValue`)

This means the `parameter_schema` generation needs to be a little more sophisticated, then 🤔 we can update that after figuring out how to handle params with `content` specified

Tieske:
How about keeping it simple; if there is a `schema`, use it. If not use `{}` as schema (so always add the validation). That covers 2 cases;

- `content` specified, which isn't supported by the plugin
- neither `content` nor `schema` specified, which is in violation with the spec

In the code you can still make the above options 2 separate code branches, each with its own comment as to why we add a blank schema. In absence of a way to generate warnings, we should properly document the "`content` not supported" case, and that it won't warn nor fail.

That said; doesn't Insomnia have a linter for OAS specs? that would probably be the place to warn about the 2nd case mentioned above, since its a spec violation.

develohpanda:
> if there is a schema, use it. If not use {} as schema (so always add the validation)

That seems appropriate, same pattern as the `body_schema`. Will add comments to explain that behavior.

I'll make this update in #3283 so we can close this PR

> That said; doesn't Insomnia have a linter for OAS specs? that would probably be the place to warn about the 2nd case mentioned above, since its a spec violation.

Yes, we use [spectral](https://www.npmjs.com/package/@stoplight/spectral), although we're not on the latest version (but we should be). Spectral has support for advanced, custom linting rules to be built so that's something we can use to enforce many of these rules but at the very least the default implementation of spectral should lint the OpenAPI spec requirements.

develohpanda:
Closing as this will be resolved in #3283

PR #3298, opened 2021-04-16
fix(o2k) update path param regex to match 1 segment only

Tieske:
existing regex `\S+` is equivalent to `[^\s]+` by adding the `/` character as not allowed, we limit the regex to match only 1 path segment

This path: `/tracks/{hello}/world/{there}`
Would create: `/tracks/(?<hello>\S+)/world/(?<there>\S+)$`

When published, this path matches `/tracks/xxx/world/yyy` as it should
With captures:
  `hello`: `xxx`
  `there`: yyy

This path also matches `/tracks/xxx/zzz/world/yyy/andthensome` but it should not
With captures:
  `hello`: `xxx/zzz`
  `there`: `yyy/andthensome`

The last one should have matched only a single path segment, but actually matches multiple.

### DIFF
--- a/packages/openapi-2-kong/src/__tests__/common.test.js
+++ b/packages/openapi-2-kong/src/__tests__/common.test.js
@@ -187,7 +187,9 @@ describe('common', () => {
 
   describe('pathVariablesToRegex()', () => {
     it('converts variables to regex path', () => {
-      expect(pathVariablesToRegex('/foo/{bar}/{baz}')).toBe('/foo/(?<bar>\\S+)/(?<baz>\\S+)$');
+      expect(pathVariablesToRegex('/foo/{bar}/{baz}')).toBe(
+        '/foo/(?<bar>[^\\/\\s]+)/(?<baz>[^\\/\\s]+)$',
+      );
     });
 
     it('does not convert to regex if no variables present', () => {
--- a/packages/openapi-2-kong/src/common.js
+++ b/packages/openapi-2-kong/src/common.js
@@ -68,7 +68,8 @@ export function generateSlug(str: string, options: SlugifyOptions = {}): string 
 const pathVariableSearchValue = /{([^}]+)}(?!:\/\/)/g;
 
 export function pathVariablesToRegex(p: string): string {
-  const result = p.replace(pathVariableSearchValue, '(?<$1>\\S+)');
+  // match anything except whitespace and '/'
+  const result = p.replace(pathVariableSearchValue, '(?<$1>[^\\/\\s]+)');
   // add a line ending because it is a regex
   return result + '$';
 }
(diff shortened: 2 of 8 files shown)


Tieske:
the CI failure doesn't provide a meaningfull hint (at least to me) as to what failed... any help appreciated

packages/openapi-2-kong/src/__tests__/common.test.js
```diff
@@ -187,7 +187,9 @@ describe('common', () => {
 
   describe('pathVariablesToRegex()', () => {
     it('converts variables to regex path', () => {
-      expect(pathVariablesToRegex('/foo/{bar}/{baz}')).toBe('/foo/(?<bar>\\S+)/(?<baz>\\S+)$');
+      expect(pathVariablesToRegex('/foo/{bar}/{baz}')).toBe(
+        '/foo/(?<bar>[^\\/\\r\\n\\t\\f\\v ]+)/(?<baz>[^\\/\\r\\n\\t\\f\\v ]+)$',
+      );
```
develohpanda: The CI failure is to do with this test. The expected result is incorrect? I'm looking at the updated implementation of `pathVariablesToRegex()` below and it matches what the received result is, not the expected result.

image

Since `\s` is equivalent to `\r\n\t\f\v `, just need to update the expected result.

packages/openapi-2-kong/src/declarative-config/__tests__/services.test.js
```diff
@@ -59,7 +59,7 @@ describe('services', () => {
               name: 'My_API-birds_id-get',
               strip_path: false,
               methods: ['GET'],
-              paths: ['/birds/(?<id>\\S+)$'],
+              paths: ['/birds/(?<id>[^\\/\\r\\n\\t\\f\\v ]+)$'],
```
develohpanda: CI also fails with this test for the same reason as my previous comment

image

packages/openapi-2-kong/src/common.js
```diff
@@ -68,7 +68,7 @@ export function generateSlug(str: string, options: SlugifyOptions = {}): string
 const pathVariableSearchValue = /{([^}]+)}(?!:\/\/)/g;
 
 export function pathVariablesToRegex(p: string): string {
-  const result = p.replace(pathVariableSearchValue, '(?<$1>\\S+)');
+  const result = p.replace(pathVariableSearchValue, '(?<$1>[^\\/\\s]+)');
```
develohpanda, line 72: It would be worth adding a comment here to explain this regex, something like

```suggestion
  // match any non-whitespace character, excluding /
  const result = p.replace(pathVariableSearchValue, '(?<$1>[^\\/\\s]+)');
```

develohpanda (approved):
LGTM after the test fixes

image

PR #3310, opened 2021-04-21
Use exact versions and allow publishing of beta NPM packages

develohpanda:
Currently, when releasing a _beta_ version of the app, we need to publish _stable_ versions of NPM packages. For obvious reasons this is can be problematic so this PR updates our scripts to allow for publishing to a beta channel.

Documentation links:
- [`lerna version`](https://github.com/lerna/lerna/tree/main/commands/version) bump and update versions
	- [`--exact`](https://github.com/lerna/lerna/tree/main/commands/version#--exact) easier downgrades since we don't work with semver
	- [`--no-private`](https://github.com/lerna/lerna/tree/main/commands/version#--no-private) ignore private packages
	- [`--preid beta`](https://github.com/lerna/lerna/tree/main/commands/version#--preid) for pre-releases use `beta` in the version number
	- (optional) [`--no-git-tag-version`](https://github.com/lerna/lerna/tree/main/commands/version#--no-git-tag-version) don't commit or tag, useful for a dry run
	- (optional) [`--force-publish`](https://github.com/lerna/lerna/tree/main/commands/version#--force-publish) force bump all packages
- [`lerna publish`](https://github.com/lerna/lerna/tree/main/commands/publish)
	- [`from-git`](https://github.com/lerna/lerna/tree/main/commands/publish#bump-from-git)
	- [`--pre-dist-tag beta`](https://github.com/lerna/lerna/tree/main/commands/publish#--pre-dist-tag-tag) use the `beta` dist tag for pre-release versions

When cutting a release, from now on we only need to run `npm run version` and choose how we want to bump (patch, minor, major, prepatch, preminor, premajor), which will bump the versions and create a git tag as appropriate.

`npm run publish` will simply publish the packages with their given version numbers, if they don't exist on NPM.

For releasing packages, a user needs to only do `npm run version`, select the bump (patch or prepatch), then `npm run publish`.

(A subsequent PR #3311 will automate the `npm run publish` step to happen as part of the a release in GitHub actions).

In order to test this, you can run `npm run version:dry` locally and you'll see a changeset in your local environment.

Closes INS-358, Closes INS-627, Closes INS-503


reynolek (approved):
Nice!

DMarby (approved):

sonicyeti (approved):
Very cool!

PR #3311, opened 2021-04-21
Publish NPM packages through CI

develohpanda:
What it says on the tin. The updated release workflow will:
- run tests on all OSs
- then, publish NPM packages
- then, release each OS bundle

Closes INS-630

### DIFF
--- a/.github/workflows/release-core.yml
+++ b/.github/workflows/release-core.yml
@@ -46,7 +46,11 @@ jobs:
         with:
           node-version: ${{ steps.nvm.outputs.NVMRC }}
       - run: npm run bootstrap
-      - run: npm run publish
+      - run: |
+          npm config set //registry.npmjs.org/:_authToken ${NPM_TOKEN}
+          npm run publish
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN}}
   Windows:
     needs: [ npm ]
     name: Windows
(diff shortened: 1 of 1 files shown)


reynolek (approved):

.github/workflows/release-core.yml
```diff
@@ -14,7 +14,41 @@ env:
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Read Node version from .nvmrc
+        run: echo "##[set-output name=NVMRC;]$(cat .nvmrc)"
+        id: nvm
+      - uses: actions/setup-node@v1
+        with:
+          node-version: ${{ steps.nvm.outputs.NVMRC }}
+      - run: npm run bootstrap
+      - run: npm run publish
```
DMarby: I think we need a step before this to configure authentication with NPM so that we're able to publish?
Since access tokens are bound to a user, not an organisation, it would probably be good to create a new NPM user and add it to our credentials vault, and then add a token that can be used here, as a repository secret.

DMarby (changes requested):

.github/workflows/release-core.yml
develohpanda: Oh, yes we absolutely do, I forgot to add in the description. That's why I created a separate PR for it 😅 I agree, creating an org account to publish with will be good

sonicyeti (approved):

develohpanda:
Reminder to myself to rebase `git rebase --onto develop 928dec2`

DMarby (approved):

PR #3314, opened 2021-04-22
split `url` into components on a `service`

Tieske:
1. the generated `service` entity would have a `url` field. This field is only a sugar-field for Kong since it internally only stored the individual components of the url. The effect of this is when a generated config is synced to Kong, the provided config will have a `url` field where the Kong config will have `host`, `path`, etc., and hence the comparison will always fail, and it will always be marked in the diff and synced, despite that the end result would be identical. By splitting the `url` already in the generated config, this noise will be gone.

2. fixes a bug; now uses the `upstream` name in a `service` instead of an actual hostname from the `servers` block

3. fixes a bug; if a `server` was specified as a path only, then a target would be generated as `"null:null"`, now it will simply not create any targets

(against Insomnia v2021.3.0-beta.1)

Fixes INS-556


sonicyeti (approved):
Great stuff!

packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -36,13 +37,17 @@ export function generateService(
...
   const name = getName(api);
+  const parsedUrl = parseUrl(serverUrl);
 
   // Service plugins
   const globalPlugins = generateGlobalPlugins(api);
 
   const service: DCService = {
     name,
-    url: serverUrl,
+    protocol: parsedUrl.protocol.substring(0, parsedUrl.protocol.length - 1),
```
develohpanda, line 48:
```suggestion
	// remove semicolon ie. convert https: to https
    protocol: parsedUrl.protocol.substring(0, parsedUrl.protocol.length - 1),
```

develohpanda (approved):

develohpanda:
Nice optimization with option 1. I'm guessing that means faster sync times too huh

packages/openapi-2-kong/src/declarative-config/services.js
```diff
@@ -36,13 +37,18 @@ export function generateService(
...
   const name = getName(api);
+  const parsedUrl = parseUrl(serverUrl);
 
   // Service plugins
   const globalPlugins = generateGlobalPlugins(api);
 
   const service: DCService = {
     name,
-    url: serverUrl,
+	// remove semicolon ie. convert https: to https
```
develohpanda:
```suggestion
    // remove semicolon ie. convert https: to https
```

Tieske:
> Nice optimization with option 1. I'm guessing that means faster sync times too huh

@develohpanda yes! forgot to mention that 😄

PR #3318, opened 2021-04-23
[o2k] fix name generation

Tieske:
1. OperationId was not slugified, which could cause illegal names
2. OperationId and x-kong-name on operation level were not prefixed with the spec name

wrt 2; OperationId is required to be unique within a spec, yet since multiple specs are combined within Kong installation, we prefix them by the spec name to prevent accidental name collisions.

Closes INS-652


develohpanda (approved):
🔥

PR #3320, opened 2021-04-23
[o2k] fix targets not getting tags

Tieske:
fixes [INS-651](https://linear.app/insomnia/issue/INS-651/targets-do-not-get-tags)


develohpanda (approved):
🔥

PR #3327, opened 2021-04-26
[o2k] fix parameter default styles

Tieske:
PR #3256 changed the default style from "simple" to "form". But that is not correct, the default style depends on the type of parameter.

This PR corrects it and adds the proper style for each of the 4 types of parameters, as specified in the OAS docs here: https://swagger.io/docs/specification/serialization/

fixes INS-654

### DIFF
--- a/packages/openapi-2-kong/src/declarative-config/plugins.js
+++ b/packages/openapi-2-kong/src/declarative-config/plugins.js
@@ -38,6 +38,13 @@ function generatePlugin([key, value]: [string, Object]): DCPlugin {
  */
 const ALLOW_ALL_SCHEMA = '{}';
 
+const DEFAULT_PARAM_STYLE = {
+  header: 'simple',
+  cookie: 'form',
+  query: 'form',
+  path: 'simple',
+};
+
 function generateParameterSchema(operation?: OA3Operation): Array<Object> | typeof undefined {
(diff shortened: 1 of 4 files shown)


packages/openapi-2-kong/src/declarative-config/plugins.js
```diff
@@ -62,7 +69,7 @@ function generateParameterSchema(operation?: OA3Operation): Array<Object> | type
         required: !!(p: Object).required,
         name: (p: Object).name,
         schema,
-        style: (p: Object).style ?? 'form',
+        style: (p: Object).style ?? DEFAULT_PARAM_STYLE[(p: Object).in],
```
develohpanda: if `p.in` is not `header, cookie, query or path` then this will likely return undefined

develohpanda (changes requested):

develohpanda (approved):
Looks to be working as expected with the following iterations:

```
      parameters:
        - in:
          style: acbd
          name: userId
          schema:
            type: integer
          required: true
```
```
      parameters:
        - in: asdfadsf
          name: userId
          schema:
            type: integer
          required: true
```
```
      parameters:
        - in: path
          name: userId
          schema:
            type: integer
          required: true
```

PR #3328, opened 2021-04-27
Tags for all of the plugins!

develohpanda:
Closes INS-629, best examples are through the fixture updates, requirements are outlined in the linked ticket


develohpanda:
There are `package.json` changes included in this PR which are only for the beta build; I wanted to avoid creating a nested release branch. Please ignore those changes, you can review the first four commits only as the `package.json` changes are only in [`ab0c314` (#3328)](https://github.com/Kong/insomnia/pull/3328/commits/ab0c31447fc1a2516398d859a2151a8305d87936).

Tieske (approved):
lgtm

packages/openapi-2-kong/src/declarative-config/plugins.js
```diff
@@ -1,23 +1,23 @@
...
 
   // Server plugins should load from the api spec root and from the server
   return Object.entries(item)
     .filter(pluginFilter)
-    .map(generatePlugin);
+    .map(e => generatePlugin(e, tags));
 }
 
-function generatePlugin([key, value]: [string, Object]): DCPlugin {
+function generatePlugin([key, value]: [string, Object], tags: Array<string>): DCPlugin {
```
dimitropoulos, line 20: not-blocking and not required: I suggest we consider doing a data-last curry here so we can use it more ergonomically, e.g. `.map(generatePlugin(tags))`.  The signature would become

```ts
const generatePlugin = (tags: string[]) => ([key, value]: [string, Object]): DCPlugin => { .. }
```

dimitropoulos: nevermind.  I take it back.  Let's just move along.  This isn't the time to make these kinds of improvements - for the one reason that it doesn't match the rest of the project's style (yet).

dimitropoulos (approved):
aside from `x-kong-plugin-request-validator` being in the ticket but being handled in a prior commit, it seems that all of the criteria in the ticket was met here.  I'll admit I don't deeply deeply understand what's going on but the code looks good and the tests pass and the areas in the ticket seem to be covered.

PR #3336, opened 2021-04-28
sanitize all electron.shell.openExternal calls

dimitropoulos:
partially closes INS-590

### DIFF
--- a/packages/insomnia-app/.eslintrc.js
+++ b/packages/insomnia-app/.eslintrc.js
@@ -7,6 +7,9 @@ module.exports = {
     '@typescript-eslint/array-type': ['error', { default: 'generic', readonly: 'generic' }],
     '@typescript-eslint/no-use-before-define': 'off', // TSCONVERSION
     '@typescript-eslint/no-explicit-any': 'off', // TSCONVERSION
-    // 'padding-line-between-statements': ['error', { blankLine: "always", prev: ["*"], next: "export"}],
+    'no-restricted-properties': ['error', {
+      property: 'openExternal',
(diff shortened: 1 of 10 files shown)


packages/insomnia-app/app/common/misc.js
```diff
@@ -467,3 +467,9 @@ export function snapNumberToLimits(value: number, min?: number, max?: number): n
 export function isNotNullOrUndefined(obj: any): boolean {
   return obj !== null && obj !== undefined;
 }
+
+export const sanitizedOpenExternal = (href: string) => {
+  if (href.startsWith('http')) {
```
DMarby: Not sure if it might be safer to parse the URL and check the protocol here (`new URL(url).protocol`)?
But I think this is probably good as-is 👍

DMarby (approved):
Awesome! 🎉

packages/insomnia-app/app/common/misc.js
develohpanda: Or maybe a regex check as well because it wasn't immediately apparent to me that https is supported: `/^https?/`

develohpanda (approved):

DMarby (approved):

packages/insomnia-app/app/common/misc.js
dimitropoulos: love it! I went with https://github.com/Kong/insomnia/pull/3336/commits/23fe70fde2e16c9ecfce293bc85e058aa699fa22 since it seemed to be the most simple approach and doesn't require knowledge of regexes (in the aim to be more beginner friendly where possible, although I love me a good regex).

PR #3338, opened 2021-04-28
makes all package.json author fields consistent

dimitropoulos:
While we were releasing, I noticed that Greg and his kong email (not a valid email now, anyway) was on many of the packages.  This PR fixes that and also adds it to any package I saw that was missing it.

### DIFF
--- a/plugins/insomnia-plugin-cookie-jar/package.json
+++ b/plugins/insomnia-plugin-cookie-jar/package.json
@@ -2,6 +2,12 @@
   "name": "insomnia-plugin-cookie-jar",
   "version": "2.2.35",
   "author": "Kong <office@konghq.com>",
+  "contributors": [
+    {
+      "name": "Preston Alvarado",
+      "email": "pjalva1@hotmail.com"
+    }
+  ],
   "description": "Insomnia cookie jar template tag",
   "license": "MIT",
   "repository": "https://github.com/kong/insomnia/tree/master/plugins/insomnia-plugin-cookie-jar",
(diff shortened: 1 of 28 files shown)


plugins/insomnia-plugin-cookie-jar/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "insomnia-plugin-cookie-jar",
   "version": "2.2.35-beta.3",
-  "author": "Preston Alvarado <pjalva1@hotmail.com>",
```
sonicyeti, line 4: Is this ok?

sonicyeti (approved):
LGTM, just confirm we are good on the Preston replace.

plugins/insomnia-plugin-cookie-jar/package.json
reynolek: Per https://github.com/Kong/insomnia/pull/1039 -- Looks like this was in fact written by Preston

develohpanda: Is it possible to set multiple authors? I think not. But since we basically haven't changed this in force I vote for keeping it as the original author.

reynolek: At the very least, we should move current authors to `contributors` in the `package.json`. You cant have multiple authors, but you can have an array of contributors.

dimitropoulos: added in https://github.com/Kong/insomnia/pull/3338/commits/8d9fcbc044aff545318b721670a9bbcb791416b7

PR #3348, opened 2021-05-01
fix: insomnia open dialog for proto directory can't select directories

### LINKED ISSUE #3347: Cannot navigate to any sub folder where my proto files are
When I try to import my directory of proto files, I am unable to navigate to the folder. This may be specific to my OS. I run fedora 33 and downloaded and isntalled the rpm package from github.com.

1. Go to add new request(type as grpc).
2. Click on import from directory
3. Let's say your proto file is in ~/Code/myproject/proto. Try to navigate to that folder.
4. it is not possible to click on any folder in the center pane.

**Expected behavior**
I should be able to click on the folder and import a folder of my choice.

**Screenshots**
image

Version: Insomnia 2021.2.2
Release date: 3/30/2021
OS: Linux x64 5.11.16-200.fc33.x86_64 (fedora 33)
Electron: 9.3.5
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.73.0 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.9 zstd/1.4.9 libidn2/2.3.0 libssh2/1.9.0 nghttp2/1.41.0
**Additional context**

### LINKED ISSUE #3420: Cannot select proto folder on Manjaro linux KDE
**Describe the bug**
* Cannot select folder where my protos are located

**To Reproduce**
* Select proto folder
* Got folder selection prompt with folders unselectable

**Expected behavior**
I can enter into subfolder and navigate to my protos directory

**Screenshots**
[
Screenshot_20210525_113958
](url)

 - OS: [e.g. macOS 10.15.1] linux, KDE
 - Installation Method: [e.g. homebrew]: AUR
 - App Version [e.g. v7.0.3] insomnia-bin 2021.3.0-1

**Additional context**
* You probably have to enable folder selection in the dialog. For file type filter I could only select files, and you probably have to put folders as well

koppathachan:
refer: [Issue](https://github.com/Kong/insomnia/issues/3347)
This change seems to fix above issue, it might not be right.

Closes #3347, closes #3420, closes INS-741


develohpanda:
Thanks for checking this on several OS's and raising a PR! Pinging @dimitropoulos to explore whether this is the best solution, as it seems to be specific to Linux.

Excellent candidate for a smoke test as well, in future.

dimitropoulos:
That ended up being quite the adventure! :)

It's working now on.  @develohpanda passing the puck back to you to try re-verify on Mac after my Linux fix.  The `name` filed looks to have been misinterpreted before as being some kind of title for the dialog.  When, in fact, it was being used to allow or disallow certain files or folders from being selected.

For example if I:
```diff
 filters: [{
   extensions: (extensions?.length ? extensions : ['*']),
+  name: 'ziltoid',
 }],
```

Then I get a dialog like this:
Screenshot_20210525_105426

Note the bottom right.  It says `ziltoid`.  I don't, honestly, understand how this is working because with `ziltoid` ad the `name`, I can't get a folder named `ziltoid`, a file named `ziltoid` or a file named `test.ziltoid` to be selectable.  It seems that we just simply can't have a `name` in there.  There's something fishy here because the electron docs make it look like you can do something very different: https://www.electronjs.org/docs/api/dialog#dialogshowopendialogbrowserwindow-options

Screenshot_20210525_105635

Now take a look back at the original issue and you'll see, likewise, that `Files` is what's filtered in the bottom right.  It's easy to miss:  https://github.com/Kong/insomnia/issues/3347. 


---


Well.  Maybe this was an electron bug that got fixed long ago.  Not sure.  I wasn't able to find any issues about it, but I can see how it might be hard to search for something like this.

develohpanda (approved):
This PR appears to be working on macOS as well 👍🏽 

It appears the following usage might also be broken on Linux for the same reason?

https://github.com/Kong/insomnia/blob/f2609fcaa9beaeacb73d477517a5d01956461997/packages/insomnia-app/app/ui/redux/modules/global.tsx#L369-L385

I believe you can reach this code path from Preferences > Data  > Import Data > From File.

PR #3372, opened 2021-05-12
keep the previous svgr assets dir in gitignore

develohpanda:
when a user pulls the new TS changes, these won't appear as new files to commit, because the SVGR generation location has changed

image


dimitropoulos (approved):
I'm gonna break from tradition a bit and merge this (one-line non-controversial change to a gitignore) so it doesn't become a problem for any upcoming PRs I'm going to be working on today

PR #3375, opened 2021-05-12
remove duplication of all jest configs

dimitropoulos:
This was an issue we came across during the initial TypeScript conversion.  I posted about it in https://github.com/facebook/jest/issues/11399 and they provided a solution which _appears_ to work in all cases but `app`, which fails for a strange reason.

Using, from what I can tell, is exactly the same syntax using exactly the same npm script in app's package.json, you get

```
Preset ../../jest-preset.js not found.
```

If you try to run it.  See 138c3242dfd42784ab2ac0fa6f9f5047ab82b2f8

If we can get past this, we're golden.  Even if not, we can still merge this PR as-is and hope to figure out what's going on with app's config later, maybe.

### DIFF
--- a/jest-preset.js
+++ b/jest-preset.js
@@ -1,20 +1,16 @@
 /** @type { import('@jest/types').Config.InitialOptions } */
 module.exports = {
+  collectCoverage: false,
   globals: {
     'ts-jest': {
       isolatedModules: true,
     },
   },
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
-  testEnvironment: 'node',
-  transform: {
-    '^.+\\.tsx?$': 'ts-jest',
-  },
-  testMatch: [
-    '**/*.test.ts',
-  ],
-  verbose: false,
   resetMocks: true,
   resetModules: true,
-  collectCoverage: false,
+  testEnvironment: 'node',
+  testRegex: ['.+\\.test\\.tsx?$'],
+  transform: { '^.+\\.tsx?$': 'ts-jest' },
+  verbose: false,
 };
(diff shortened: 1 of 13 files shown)


packages/insomnia-app/package.json
```diff
@@ -9,7 +9,7 @@
   "main": "app/main.min.js",
   "scripts": {
     "clean": "tsc --build tsconfig.webpack.json --clean && tsc --build tsconfig.build.json --clean",
-    "test": "npm run test:jest",
+    "test": "jest",
     "test:jest": "cross-env NODE_ENV=test jest --silent",
```
develohpanda: * is NODE_ENV=test set auto-magically/not used in the app?
* can we now remove `test:jest`?

jest-preset.js
```diff
@@ -1,20 +1,16 @@
...
-    '**/*.test.ts',
-  ],
-  verbose: false,
   resetMocks: true,
   resetModules: true,
-  collectCoverage: false,
+  testEnvironment: 'node',
+  testRegex: ['.+\\.test\\.tsx?$'],
+  transform: { '^.+\\.tsx?$': 'ts-jest' },
+  verbose: false,
```
develohpanda, line 15: Unrelated question but do you happen to have a VS Code extension which sets these properties in alphabetical order?

develohpanda (approved):

jest-preset.js
dimitropoulos: yes I do! It's not automatic, but I use https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines in situations like this where I'm matching up things it makes it MUCH easier to not miss things.  It's also useful for quickly/easily making sure a list (or object!) doesn't have duplicate entries.

packages/insomnia-app/package.json
dimitropoulos: 1: yes, jest sets it automagically https://jestjs.io/docs/environment-variables
2: I was gonna ask to be doubly sure, but yes, we can.  I'll do so now.

PR #3376, opened 2021-05-12
Fix CI Intermittent failures

dimitropoulos:
closes: INS-694

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "build": "lerna run build --parallel",
     "lint": "lerna run lint --parallel --stream --no-bail",
     "lint:fix": "lerna run lint:fix --parallel --stream --no-bail",
-    "bootstrap": "npm install && lerna bootstrap && lerna run --parallel --stream bootstrap",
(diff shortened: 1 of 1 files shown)


dimitropoulos:
@develohpanda I think we may also need to specify peerDependencies correctly according to:

> Specify any peerDependencies in the devDependencies of a given leaf package.json, and the topological order will be respected by lerna run.
> https://github.com/lerna/lerna/issues/2009#issuecomment-491971922

develohpanda:
That seems logical, many of those weren't even in flow before so didn't need any build which might explain why it JustWorked™️ before

develohpanda:
@dimitropoulos that link actually states that any `peerDependencies` should be specified as _devDependencies_. In `insomnia-app`, most of these are actually specified as `dependencies`; I don't know if that will have an issue in future but we can cross that bridge if we get to it

develohpanda:
I removed the `peerDependencies` commit for now, I think removing the `--parallel` had already fixed it, and if it doesn't we can revisit this.

Looking at the lerna docs, the `--parallel` option explicitly ignores toplogical order

develohpanda (approved):

PR #3378, opened 2021-05-12
Fix vercel deployments

develohpanda:
Vercel couldn't find TS because it's root directory is not the lerna root, it's `packages/insomnia-components`, so `insomnia-components` needs to work on it's own (currently)


DMarby (approved):
:shipit:

PR #3380, opened 2021-05-13
Fix regression in empty folder state

develohpanda:
During the TS merge, the children check was updated from `!isCollapsed && children.length > 0` to `!isCollapsed && children`. In this particular use case, children can be defined but be an empty array, and if so we want to show a placeholder. This PR brings the folder empty state back by altering the condition to cater for am empty array.


reynolek (approved):

sonicyeti (commented):
Great catch!

PR #3384, opened 2021-05-13
fix node-libcurl imports

dimitropoulos:
Back in the flow days, we needed to reach into these files to get at the enum, but now we can import directly.  It's a fairly strong anti-pattern to reach into `/dist` like this on npms because there are no guarantees about what will be there in the future, which greatly increases the surface-area for breakage in our own code.

### DIFF
--- a/packages/insomnia-app/app/common/render.ts
+++ b/packages/insomnia-app/app/common/render.ts
@@ -277,6 +277,7 @@ export async function render<T>(
 
   return next<T>(newObj, name, true);
 }
+
 export async function getRenderContext(
   request: Request | GrpcRequest | null,
   environmentId: string | null,
@@ -373,6 +374,7 @@ export async function getRenderContext(
   // Generate the context we need to render
   return buildRenderContext(ancestors, rootEnvironment, subEnvironment, baseContext);
 }
+
 export async function getRenderedGrpcRequest(
   request: GrpcRequest,
   environmentId: string | null,
@@ -419,9 +421,10 @@ export async function getRenderedGrpcRequestMessage(
   const renderedBody: RenderedGrpcRequestBody = await render(request.body, renderContext);
   return renderedBody;
 }
+
 export async function getRenderedRequestAndContext(
   request: Request,
-  environmentId: string | null,
+  environmentId?: string | null,
   purpose?: RenderPurpose,
   extraInfo?: ExtraRenderInfo,
 ) {
@@ -431,7 +434,7 @@ export async function getRenderedRequestAndContext(
   const cookieJar = await models.cookieJar.getOrCreateForParentId(parentId);
   const renderContext = await getRenderContext(
     request,
-    environmentId,
+    environmentId || null,
     ancestors,
     purpose,
     extraInfo || null,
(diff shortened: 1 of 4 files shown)


develohpanda (approved):
Result of having flow and not being able to import those directly (I think?)

DMarby (approved):
Looks reasonable, seems like it's breaking the tests however?

dimitropoulos:
@develohpanda @DMarby turns out there was a reason after all.  tried to be decently good about atomic commits if you wanna see what happened, but I was (at least) able to get rid of the call into /dist from the non-test, but the test needs to keep it otherwise you get something like:

```
  ● Test suite failed to run

    The module '/home/dimitri/src/github.com/Kong/insomnia/packages/insomnia-app/node_modules/node-libcurl/lib/binding/node_libcurl.node'
    was compiled against a different Node.js version using NODE_MODULE_VERSION 80.
    This version of Node.js requires NODE_MODULE_VERSION 72. Please try re-compiling or re-installing
    the module (for instance, using `npm rebuild` or `npm install`).

      at Runtime._loadModule (../node_modules/jest-runtime/build/index.js:893:29)
      at Object.<anonymous> (../node_modules/node-libcurl/lib/Easy.ts:10:44)

```

Since I improved a few types and removed a few ignores along the way towards finding this out, please take another look if you could to verify we're good to merge.

packages/insomnia-app/app/common/render.ts
```diff
@@ -419,9 +421,10 @@ export async function getRenderedGrpcRequestMessage(
   const renderedBody: RenderedGrpcRequestBody = await render(request.body, renderContext);
   return renderedBody;
 }
+
 export async function getRenderedRequestAndContext(
   request: Request,
-  environmentId: string | null,
+  environmentId?: string | null,
```
develohpanda, line 427: Checked this and it looks like an acceptable change. Can now remove this fallback: https://github.com/Kong/insomnia/blob/1271f5a925c956452165fa28384c6dfb84f6fdec/packages/insomnia-app/app/network/network.ts#L965

packages/insomnia-app/app/network/network.ts
```diff
@@ -461,7 +446,7 @@ export async function _actuallySend(
       if (renderedRequest.settingSendCookies) {
         // Tell Curl to store cookies that it receives. This is only important if we receive
         // a cookie on a redirect that needs to be sent on the next request in the chain.
-        curl.setOpt(Curl.option.COOKIEFILE, '');
```
develohpanda, line 464: This is what happens when the helper/wrapper and the parent function carry the same name 😆 good catch

develohpanda (approved):
Changes look good to me

DMarby (approved):

packages/insomnia-app/app/network/network.ts
dimitropoulos: I chuckled, too.  We can try as hard as we want, but in the end, this is a thing made by humans.  This is a hard mistake to avoid.

packages/insomnia-app/app/common/render.ts
dimitropoulos: Screenshot_20210517_091947

unfortunately not, `undefined` is not covariant with the environmentId arg of getRenderContext.


Making this argument optional was added to fix this error:
Screenshot_20210517_092504

Which appears to be genuine (every call in `request.test.ts` only passes the first argument, as well as every call in `har.test.ts`).

PR #3385, opened 2021-05-13
Fix issue chaining multiple requests

### LINKED ISSUE #3382: Unable to chain multiple requests
**Describe the bug**
Hi, I have an issue chaining multiple requests. I have a chain of 3 requests where:
- request n° 3 use a response header from request n°2 (always resend behaviour)
- request n° 2 use a response body attribute from request n° 1 (always resend behaviour)
- request n° 1 doesn't use any template tags

So executing the request n° 3 I always get an error because the request n° 1 is never executed preventing infinite recursion.
Is this a bug or isn't this the correct way to chain multiple requests at once?

**Expected behavior**
Executing the request chain without any error executing the request n° 3

**Screenshots**
image

Version: Insomnia 2021.3.0
Release date: 4/28/2021
OS: Darwin x64 20.2.0
Electron: 9.3.5
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.73.0 OpenSSL/1.1.1j zlib/1.2.11 brotli/1.0.9 zstd/1.4.9 libidn2/2.1.1 libssh2/1.9.0 nghttp2/1.41.0

vincendep:
Closes #3382

### DIFF
--- a/plugins/insomnia-plugin-response/__tests__/index.test.js
+++ b/plugins/insomnia-plugin-response/__tests__/index.test.js
@@ -540,15 +540,31 @@ describe('Response tag', () => {
       expect(await tag.run(context, 'raw', 'req_1', '', 'never')).toBe('Response res_existing');
     });
 
-    it('does not resend recursive', async () => {
+    it('does not resend if request has already sent in recursive chain', async () => {
       const requests = [{ _id: 'req_1', parentId: 'wrk_1' }];
       const responses = [];
-      const context = _genTestContext(requests, responses);
+      const context = _genTestContext(requests, responses, { requestChain: ['req_1']});
 
-      const response = await tag.run(context, 'raw', 'req_1', '', 'always');
-      expect(response).toBe('Response res_1')
-    });
+      try {
+        await tag.run(context, 'raw', 'req_1', '', 'always');
+      } catch (err) {
+        expect(err.message).toBe('No responses for request');
+        return;
+      }
+
+      throw new Error('Running tag should have thrown exception');
   });
+
+  it('does send if request has not been sent in recursive chain', async () => {
+    const requests = [{ _id: 'req_1', parentId: 'wrk_1' }];
+    const responses = [];
+
+    const context = _genTestContext(requests, responses, { requestChain: ['req_2']});
+
+    const response = await tag.run(context, 'raw', 'req_1', '', 'always');
+      expect(response).toBe('Response res_1')
+  });
+});
 
   describe('Max Age', () => {
     const maxAgeArg = tag.args[4];
--- a/plugins/insomnia-plugin-response/__tests__/index.test.js
+++ b/plugins/insomnia-plugin-response/__tests__/index.test.js
@@ -542,19 +542,11 @@ describe('Response tag', () => {
 
     it('does not resend recursive', async () => {
       const requests = [{ _id: 'req_1', parentId: 'wrk_1' }];
-
       const responses = [];
+      const context = _genTestContext(requests, responses);
 
-      const context = _genTestContext(requests, responses, { fromResponseTag: true });
-
-      try {
-        await tag.run(context, 'raw', 'req_1', '', 'always');
-      } catch (err) {
-        expect(err.message).toBe('No responses for request');
-        return;
-      }
-
-      throw new Error('Running tag should have thrown exception');
+      const response = await tag.run(context, 'raw', 'req_1', '', 'always');
+      expect(response).toBe('Response res_1')
     });
   });
 
--- a/plugins/insomnia-plugin-response/index.js
+++ b/plugins/insomnia-plugin-response/index.js
@@ -121,9 +121,7 @@ module.exports.templateTags = [
       let response = await context.util.models.response.getLatestForRequestId(id, environmentId);
 
       let shouldResend = false;
-      if (context.context.getExtraInfo('fromResponseTag')) {
-        shouldResend = false;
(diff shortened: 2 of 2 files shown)


plugins/insomnia-plugin-response/index.js
```diff
@@ -121,34 +121,30 @@ module.exports.templateTags = [
...
       } else if (resendBehavior === 'no-history') {
         shouldResend = !response;
       } else if (resendBehavior === 'when-expired') {
-        if (!response) {
-          shouldResend = true;
-        } else {
-          const ageSeconds = (Date.now() - response.created) / 1000;
-          shouldResend = ageSeconds > maxAgeSeconds;
-        }
+        shouldResend = ! response || (Date.now() - response.created) / 1000 > maxAgeSeconds;
```
dimitropoulos: please revert this change

plugins/insomnia-plugin-response/index.js
```diff
@@ -121,34 +121,30 @@ module.exports.templateTags = [
...
-      if (fromResponseTag) {
+      const requestChain = context.context.getExtraInfo('requestChain') || [];
+      if (requestChain.some(id => id === request._id)) {
         console.log('[response tag] Preventing recursive render');
         shouldResend = false;
       }
 
       if (shouldResend && context.renderPurpose === 'send') {
         console.log('[response tag] Resending dependency');
+        const requestChain = context.context.getExtraInfo('requestChain') || []
```
dimitropoulos: Is there a reason to grab this again?  It's already declared 8 lines above (now).

dimitropoulos (changes requested):
Thanks for the quick PR!

Other than two very minor things, we should be ready to go with this.  I can confirm that this fixes the issue.

Here is how I tested this:
[Run in Insomnia}](https://insomnia.rest/run/?label=Multiple%20Chaining%20Test&uri=https%3A%2F%2Fgist.githubusercontent.com%2Fdimitropoulos%2F64fdb4a75a5bd68998dbe1cae5cf71e4%2Fraw%2F9441f985d80e9f8e99e369689e60cb79d07b7abf%2Fmultiple-chain.json)

For this to work you'd need to run `npm run serve` from `packages/insomnia-smoke-test`.

plugins/insomnia-plugin-response/index.js
vincendep: nope...removed :sweat_smile:

dimitropoulos (approved):

dimitropoulos:
@vincendep sorry, I didn't catch it sooner.  any idea what the changes to the package-lock are all about?  I'm done for the day today but I'll take a look tomorrow and maybe we can get this merged tomorrow one way or the other.

vincendep:
> @vincendep sorry, I didn't catch it sooner. any idea what the changes to the package-lock are all about? I'm done for the day today but I'll take a look tomorrow and maybe we can get this merged tomorrow one way or the other.

mmm no idea what those changes are. I ve only run npm run bootstrap && npm run app-start from the root to test the changes I ve made. (But first time I was using a wrong version of node so I had to run a clean)

dimitropoulos:
haha, welp.  I also nave no idea where those lockfile changes came from.  I removed them and ran a bootstrap and they didn't return.  /shrug.

All the same, I re-tested just to make sure and this seems to be working as expected.  Thanks again for contributing!

vincendep:
thank you too...happy to contribute!

dimitropoulos:
ok, sorry to renege, but I would like @develohpanda to take a look at this.  There was a test that needed to be updated.  I studied it for some time, indeed, and concluded that the change we made to the behavior (to fix the bug) needed to be made to the original, but now I'm worried that there's a potential for a different use-case for recursion that there wasn't before.  The test was written in such a way that looks like it is expecting an infinite loop, but now there isn't one.  Since I'm not sure and out of an abundance of caution, let's @develohpanda a chance to take a look.

develohpanda (approved):
This looks good to me - I tested some of the recursive scenarios I could think of both on Live and in this PR and they seem to be working as intended. I like the `requestChain` approach over the boolean flag.

I think it would be good to add a safety net into the chain, however - so it doesn't grow to an unreasonable amount (maybe hard-coded at 5 or 10?) I am mostly imagining malicious use of it; if someone was to export a workspace online with many (tens, hundreds) of requests chained together, if someone was to import and run a request that would be problematic.

plugins/insomnia-plugin-response/__tests__/index.test.js
```diff
@@ -542,19 +542,11 @@ describe('Response tag', () => {
...
-      try {
-        await tag.run(context, 'raw', 'req_1', '', 'always');
-      } catch (err) {
-        expect(err.message).toBe('No responses for request');
-        return;
-      }
-
-      throw new Error('Running tag should have thrown exception');
+      const response = await tag.run(context, 'raw', 'req_1', '', 'always');
+      expect(response).toBe('Response res_1')
```
develohpanda: @dimitropoulos I think this test needs to still test what it describes, though, this time using the `requestChain` context value instead of `fromResponseTag`

develohpanda (changes requested):
Comment on the test which is worth addressing

develohpanda (approved):

plugins/insomnia-plugin-response/__tests__/index.test.js
develohpanda: Fixed in c4b0f25

PR #3386, opened 2021-05-14
ResponseTimer callstack exceeded

dimitropoulos:
closes INS-698

### DIFF
--- a/packages/insomnia-app/app/ui/components/response-timer.tsx
+++ b/packages/insomnia-app/app/ui/components/response-timer.tsx
@@ -14,7 +14,7 @@ export const ResponseTimer: FunctionComponent<Props> = ({ handleCancel, loadStar
     let interval: NodeJS.Timeout | null = null;
     if (isLoading) {
       interval = setInterval(() => {
-        const milliseconds = Date.now() - loadStartTime - 200;
+        const milliseconds = Date.now() - loadStartTime;
         setSeconds(milliseconds / 1000);
       }, 100);
     }
--- a/packages/insomnia-app/app/ui/components/response-timer.tsx
+++ b/packages/insomnia-app/app/ui/components/response-timer.tsx
@@ -36,16 +36,12 @@ export class ResponseTimer extends PureComponent<Props, State> {
   componentDidUpdate() {
     const { loadStartTime } = this.props;
 
-    if (loadStartTime <= 0) {
-      if (this.interval !== null) {
-        clearInterval(this.interval);
-      }
-      return;
+    if (this.interval !== null) {
+      clearInterval(this.interval);
     }
 
-    if (this.interval !== null) {
-      // Just to be sure
-      clearInterval(this.interval);
+    if (loadStartTime <= 0) {
+      return;
     }
 
     this.interval = setInterval(this.handleUpdateElapsedTime, 100);
--- a/packages/insomnia-app/app/ui/components/response-timer.tsx
+++ b/packages/insomnia-app/app/ui/components/response-timer.tsx
@@ -12,7 +12,7 @@ interface State {
 }
 
 @autoBindMethodsForReact(AUTOBIND_CFG)
-class ResponseTimer extends PureComponent<Props, State> {
+export class ResponseTimer extends PureComponent<Props, State> {
   _interval: NodeJS.Timeout | null = null;
 
   state: State = {
@@ -80,5 +80,3 @@ class ResponseTimer extends PureComponent<Props, State> {
     );
   }
 }
-
-export default ResponseTimer;
--- a/packages/insomnia-app/app/ui/components/response-timer.tsx
+++ b/packages/insomnia-app/app/ui/components/response-timer.tsx
@@ -1,77 +1,51 @@
-import React, { DOMAttributes, PureComponent } from 'react';
-import { autoBindMethodsForReact } from 'class-autobind-decorator';
-import { REQUEST_TIME_TO_SHOW_COUNTER, AUTOBIND_CFG } from '../../common/constants';
+import React, { DOMAttributes, FunctionComponent, useEffect, useState } from 'react';
+import { REQUEST_TIME_TO_SHOW_COUNTER } from '../../common/constants';
 
 interface Props {
   handleCancel: DOMAttributes<HTMLButtonElement>['onClick'],
   loadStartTime: number,
 }
 
-interface State {
-  elapsedTime: number;
-  interval: NodeJS.Timeout | null;
-}
+export const ResponseTimer: FunctionComponent<Props> = ({ handleCancel, loadStartTime }) => {
+  const [seconds, setSeconds] = useState(0);
+  const isLoading = loadStartTime > 0;
 
-@autoBindMethodsForReact(AUTOBIND_CFG)
-export class ResponseTimer extends PureComponent<Props, State> {
-  state: State = {
-    elapsedTime: 0,
-    interval: null,
-  };
-
-  componentWillUnmount() {
-    if (this.state.interval !== null) {
-      clearInterval(this.state.interval);
+  useEffect(() => {
+    let interval: NodeJS.Timeout | null = null;
+    if (isLoading) {
+      interval = setInterval(() => {
+        const milliseconds = Date.now() - loadStartTime - 200;
+        setSeconds(milliseconds / 1000);
+      }, 100);
     }
-  }
-
-  componentDidUpdate() {
-    const { loadStartTime } = this.props;
-
-    if (this.state.interval !== null) {
-      clearInterval(this.state.interval);
-    }
-
-    if (loadStartTime <= 0) {
-      return;
-    }
-
-    const handleUpdateElapsedTime = () => {
-      const { loadStartTime } = this.props;
-      const millis = Date.now() - loadStartTime - 200;
-      const elapsedTime = millis / 1000;
-      this.setState({ elapsedTime });
+    return () => {
+      setSeconds(0);
+      if (interval !== null) {
+        clearInterval(interval);
+      }
     };
+  }, [loadStartTime, setSeconds, isLoading]);
 
-    const interval = setInterval(handleUpdateElapsedTime, 100);
-    this.setState({ interval });
-    handleUpdateElapsedTime();
+  if (!isLoading) {
+    return null;
   }
 
-  render() {
-    const { handleCancel, loadStartTime } = this.props;
-    const { elapsedTime } = this.state;
-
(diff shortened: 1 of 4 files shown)


packages/insomnia-app/app/ui/components/response-timer.tsx
```diff
@@ -1,84 +1,61 @@
...
+            });
+          }
+          return 0;
+        });
+        clearInterval(interval);
       }
-      return;
-    }
+    };
+  }, [loadStartTime, setMilliseconds, isLoading, responseTime]);
```
develohpanda: It certainly feels strange for `responseTime` to be a dependency for _this_ effect, the intent of which is simply to increment `milliseconds`, right? It can be broken up into an individual effect to handle the lag, and a custom hook/individual effect to handle the timer incrementing. The custom hook can then, also, be unit tested separately.

Here is an example implementation: this is largely the same as your implementation but separated into standalone effects.
* `milliseconds` is reset each time the timer starts (not when it clears)
* the timing logic is extracted to a custom hook and standalone effect
* calculation of the lag is handled by a standalone effect

```ts
import React, { DOMAttributes, FunctionComponent, useEffect, useState } from 'react';
import { REQUEST_TIME_TO_SHOW_COUNTER } from '../../common/constants';

interface Props {
  handleCancel: DOMAttributes<HTMLButtonElement>['onClick'];
  loadStartTime: number;
  responseTime?: number;
}

const useTimer = (startTime: number, intervalMs: number) => {
  const [milliseconds, setMilliseconds] = useState(0);
  const [running, setRunning] = useState(false);

  useEffect(() => {
    let interval: NodeJS.Timeout | null = null;
    if (startTime > 0) {
      setMilliseconds(0);
      setRunning(true);
      interval = setInterval(() => {
        const milliseconds = Date.now() - startTime;
        setMilliseconds(milliseconds);
      }, intervalMs);
    }
    return () => {
      if (interval !== null) {
        clearInterval(interval);
        setRunning(false);
      }
    };
  }, [startTime, intervalMs]);

  return { milliseconds, running };
};

export const ResponseTimer: FunctionComponent<Props> = ({ handleCancel, loadStartTime, responseTime = 0 }) => {
  const { milliseconds, running } = useTimer(loadStartTime, 100);

  // When we have a response time, and an elapsed time, and have stopped running, log the lag
  useEffect(() => {
    if (!running && milliseconds > 0 && responseTime > 0) {
      console.info(`[ResponseTimer] request setup and tear-down measured to be ${Math.floor(milliseconds - responseTime)}ms`, {
        totalDuration: milliseconds,
        networkDuration: responseTime,
      });
    }
  }, [running, milliseconds, responseTime]);

  if (!running) {
    return null;
  }

  const seconds = milliseconds / 1000;
  return (
    <div className="overlay theme--transparent-overlay">
      <h2>
        {seconds >= REQUEST_TIME_TO_SHOW_COUNTER ? `${seconds.toFixed(1)} seconds` : 'Loading'}...
      </h2>
      <div className="pad">
        <i className="fa fa-refresh fa-spin" />
      </div>
      <div className="pad">
        <button
          className="btn btn--clicky"
          onClick={handleCancel}
        >
          Cancel Request
        </button>
      </div>
    </div>
  );
};

```

packages/insomnia-app/app/ui/components/response-timer.tsx
```diff
@@ -1,84 +1,61 @@
...
-      if (this._interval !== null) {
-        clearInterval(this._interval);
+    return () => {
+      if (interval !== null) {
+        setMilliseconds(milliseconds => {
+          if (milliseconds > 0) {
+            console.info(`[ResponseTimer] lag measured to be ${Math.floor(milliseconds - responseTime)}ms`, {
+              wallTime: milliseconds,
+              responseTime,
+            });
```
develohpanda:
```suggestion
            console.info(`[ResponseTimer] request setup and tear-down measured to be ${Math.floor(milliseconds - responseTime)}ms`, {
              totalDuration: milliseconds,
              networkDuration: responseTime,
            });
```

packages/insomnia-app/app/ui/components/response-timer.tsx
```diff
@@ -1,84 +1,61 @@
...
-
-  componentDidUpdate() {
-    const { loadStartTime } = this.props;
-
-    if (loadStartTime <= 0) {
-      if (this._interval !== null) {
-        clearInterval(this._interval);
+    return () => {
+      if (interval !== null) {
+        setMilliseconds(milliseconds => {
```
develohpanda: Even though this is in it's own closure having two variables with the exact same name (`milliseconds`) in nested code-blocks can be very misleading when skimming through the code.

develohpanda (changes requested):
The discrepancy in the times seen could also be a result of this: https://github.com/Kong/insomnia/blob/1271f5a925c956452165fa28384c6dfb84f6fdec/packages/insomnia-app/app/network/network.ts#L927-L947


Also something here doesn't quite add up: 
image

Anyway, a couple of suggested improvements.

packages/insomnia-app/app/ui/components/response-timer.tsx
dimitropoulos: I don't see what this accomplishes, to be honest, because the thing you said you were trying to avoid (responseTime being a dependency) was not avoided.  The difference now is just that we introduced new state (i.e. `running`) to share state.

This implementation has a bug, which is that it will log twice when you switch to a new request.  To reproduce:
- make `Request 1`, see the log occur
- switch to `Request 2` and see the log occur again

The trouble is, that _some_ of the state you see logged again is from `Request 1`, e.g. `totalDuration`, while the other is from whatever request you just selected, e.g. `networkDuration`.

-----

All the same, I really see the value here in what you're aiming for and I like the "single responsibility principle" adherence a lot.

I'd like to develop this more because I think there's value in it, but for now because I don't want this PR to stick around for another day, I've just removed the log line altogether.  We really are gonna need a nice `useTimer` bulletproof implementation that fills our need because we have timers like this ALL OVER the app and I'd love to work with you more on developing one that's clean and works well (and has tests), but this bug has reminded me that it's out of scope for the callstack exceeded bug this PR set out to fix.

Thanks, looking forward to making progress on this in the near future.

dimitropoulos: removed in https://github.com/Kong/insomnia/pull/3386/commits/a2e2ddfe46fd8a59358a8d274cefbe0eb9511393

packages/insomnia-app/app/ui/components/response-timer.tsx
dimitropoulos: see https://github.com/Kong/insomnia/pull/3386#discussion_r633498495

packages/insomnia-app/app/ui/components/response-timer.tsx
dimitropoulos: I don't strongly agree, haha, but I guess I wouldn't because I wrote the code. :stuck_out_tongue: 


I see where you're coming from, though, and I removed a similar case in 9506e39f82a41ed5fc31c28e53e48923475e8c16

This line you commented on was removed in https://github.com/Kong/insomnia/pull/3386/commits/a2e2ddfe46fd8a59358a8d274cefbe0eb9511393

dimitropoulos:
@develohpanda great find on the discrepancy between timers.  Looks like the rabbit hole goes a bit further down.  cool!

And I really liked where you were headed with the `useTimer` implementation.  We're really gonna need something like that in the future, but I think it is a little too much for what I want to accomplish with this PR, which is just to fix the callstack-exceeded bug.

We need to discuss more (and, haha, pun intended, take our _time_ in doing so), have tests, etc. for a `useTimer` implementation that works for us because we do similar things to this with timers all over the app and I think we'll need something like it.  And once we have something we have tested and works well we can contribute it to `react-use`.  I removed this stuff for now but I'm excited pick up where we left off in this PR at a later date.

Please take another look if/when you can :)

packages/insomnia-app/app/ui/components/response-timer.tsx
develohpanda: > I don't see what this accomplishes, to be honest, because the thing you said you were trying to avoid (responseTime being a dependency) was not avoided. 

You misunderstood 🙂 there were two effects in the previous implementation bundled into one, and the thing I said was `responseTime` doesn't make sense as a dependency for _this_ effect. The two effects are:
- increment time at a regular interval until unmount (`useTimer()`)
- log the lag once we have a response (standalone effect)

The additional `running` state is to avoid placing the timeout in a shared location.

> This implementation has a bug, which is that it will log twice when you switch to a new request.

Not a rigorously tested code snippet, it was merely an example (as stated) to demonstrate the separation of effects.

However you captured the essence with the `single responsibility principle` - when I was testing the initial implementation I noticed the timer effect was being called multiple times as the `responseTime` prop changed which didn't conceptually make sense - _when this component mounts, start a timer to and show the elapsed time_.

Likely the bug you described above is a conditional I was missing in the "handle lag" effect, but ultimately the two _are_ separate.

I'm glad you've backtracked on handling the lag, though, because that also was also out of scope for the callstack exceeded bug. 👍🏽

packages/insomnia-app/app/ui/components/response-timer.tsx
```diff
@@ -1,84 +1,51 @@
...
-        clearInterval(this._interval);
+    return () => {
+      if (interval !== null) {
+        setMilliseconds(0);
+        clearInterval(interval);
       }
-      return;
-    }
+    };
+  }, [loadStartTime, setMilliseconds, isLoading]);
```
develohpanda: I believe the setState function isn't needed as a dependency because React ensures it will never change (ES lint doesn't complain either, I'll link docs when I find em)

develohpanda (approved):
LGTM, I validated and pushed the endpoint I added in order to test this (which can be used for smoke tests later)

packages/insomnia-app/app/ui/components/response-timer.tsx
dimitropoulos: ahh, right right, yes, this is mistake.  it looked a little different for a bit when I was working with it and at the time it was require.  definitely not needed:
> React guarantees that setState function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the useEffect or useCallback dependency list.
> - https://reactjs.org/docs/hooks-reference.html#usestate

dimitropoulos:
```suggestion
  }, [loadStartTime, isLoading]);
```

PR #3387, opened 2021-05-15
Theme preview tile for OS themes 

### LINKED ISSUE #3358: Make it easier to see which theme is selected for OS themes
When selecting an OS color scheme, it can be hard to see which theme is being selected for each color scheme.

image

Show a preview tile directly below each drop-down, only when `Use OS color scheme` is enabled.

Viewing the themes list, but the tiles are disabled and the color faint.

**Additional context**
None

roguib:
Closes #3358 🎉

## What I did? 
1. I've extracted the ```svg``` that is used to preview the theme into a function named ```renderThemePreview```.
2. Then I invoked ```renderThemePreview``` under the select dropdown used for selecting the theme if the option ```use OS color scheme``` is enabled.

image

From the feature request:
> Show a preview tile directly below each drop-down, **only when Use OS color scheme is enabled**.

I decided not to hide the preview tile and instead rendered it disabled since it was hard to notice that the preview tile appeared on the bottom of the select just by looking at the scrollbar. Here you can find the problem I'm trying to describe:

https://user-images.githubusercontent.com/30463479/118369706-b4f2b100-b5a4-11eb-80f4-6152f5b0f7dd.mp4

The final implementation looks as follows:

https://user-images.githubusercontent.com/30463479/118371129-685ea400-b5ab-11eb-9574-7a1ebd158f9d.mp4

Let me know if you see anything that can be improved, I'll change it ASAP.


roguib:
Not sure about the failing test in macOS since I can't test the application in that environment but it seems it's not reproducible in Windows/Linux. Do you know what could be the cause?

develohpanda:
> Not sure about the failing test in macOS since I can't test the application in that environment but it seems it's not reproducible in Windows/Linux. Do you know what could be the cause?

Seemed to be a smoke test timeout, these can happen sometimes. I have re-run 👍🏽

develohpanda (approved):
Looks to be working, thanks for the PR! 👏🏽

develohpanda:
> From the feature request:
> 
> > Show a preview tile directly below each drop-down, **only when Use OS color scheme is enabled**.
> 
> I decided not to hide the preview tile and instead rendered it disabled since it was hard to notice that the preview tile appeared on the bottom of the select just by looking at the scrollbar. Here you can find the problem I'm trying to describe:

Seems fine to me for the time being. We do have another task to bring the OS color scheme selector above the list of themes (instead of below, as it is now), and in that case hiding the previous would work better.

I pulled this branch locally and did a quick experiment with moving the selector and changing the behavior slightly, but will need some design feedback before going ahead with the following. This is unrelated to your PR though, which can go through regardless (pending another review/QA).

roguib:
Thanks for your review @develohpanda! I'm going to have a look at the issues page and start working on another issue.

reynolek (approved):
LGTM from prod -- I love the feel of instant results.

PR #3388, opened 2021-05-16
add search for graphQL schemas

### LINKED ISSUE #2720: Documentation Browser - Supporting Search
I'm having a frustrating experience with the documentation browser. I'm working in a large project with a large number of GraphQL queries. The documentation browser shows the queries but there are so many and they don't appear to be ordered in a meaningful way (although I understand it's probably the way the introspection query returns the data).

Adding support for search, ctrl-f or at least listing queries and mutations alphabetically would be a tremendous help

Hademumel:
This PR adds search functionality to the graphQL schema explorer.

The search input is visible on the initial schema component and can be jumped to by using a shortcut. The search results are divided into two categories "Types" and "Fields". The number of found elements is thereby limited to 100 per category, and the update function is debounced, in order to allow performant searches in large schemas.

Closes #2720

grafik
grafik


develohpanda:
Wow, this is awesome! Thanks for opening a PR, added to my list for review. 🙌🏽 

In the meantime, since this change involves some UI could you please add screenshots/gifs into the PR description? Thanks!

Hademumel:
Is there nobody who can review this? I implemented this feature cause I want to use it myself 😅

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-type.tsx
```diff
@@ -74,70 +78,88 @@ class GraphQLExplorerType extends PureComponent<Props> {
     );
   }
 
+  static renderFieldsList(
```
develohpanda: This function was extracted and made static to that it can be re-used in `GraphQLExplorerSearchResults`, but it would be better suited as a standalone component.

This code-base is transitioning to Function components and hooks, and while that's not a specific ask for this PR for the larger components, using static helpers from other class components definitely slows that conversion process.

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-search-results.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  componentDidMount() {
+    this.updateSearchResults();
+  }
+
+  updateSearchResults = debounce(() => {
+    if (this.ref.current) {
+      this.setState({
+        foundFields: this.renderFoundFields(),
+        foundTypes: this.renderFoundTypes(),
```
develohpanda: Curious why you chose to store the rendered elements in state, rather than the filtered data (`GraphQLType` and `GraphQLFieldWithParentName`) and leave the rendering up to the render method.

If there's no specific reason I think it might be cleaner to refactor this and only store the filter results in state

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-schema.tsx
```diff
@@ -57,15 +60,27 @@ class GraphQLExplorerSchema extends PureComponent<Props> {
...
-          <li>{this.renderSubscriptionType()}</li>
-        </ul>
+        {filter ? (
+          <GraphQLExplorerSearchResults
+            schema={schema}
+            filter={filter}
+            onNavigateType={onNavigateType}
+            onNavigateField={onNavigateField}
+          />
+        ) : (
```
develohpanda: Should this ternary exist in `GraphQLExplorerSchema` or in `GraphQLExplorer`?

It looks like the component `GraphQLExplorerSearchResults` does not use anything from `GraphQLExplorerSchema` and the props introduced just drill down. For that reason I think this logic of whether to show the explorer or search results, belongs in the parent component here: https://github.com/Hademumel/insomnia/blob/1d79de95c2bbf15a575e624c9a3818f9118adeac/packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer.tsx#L234-L244

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-search-results.tsx
```diff
@@ -0,0 +1,125 @@
...
+    if (prevProps.filter !== this.props.filter) {
+      this.updateSearchResults();
+    }
+  }
+
+  componentDidMount() {
+    this.updateSearchResults();
+  }
+
+  updateSearchResults = debounce(() => {
```
develohpanda: IMO this is the wrong place to debounce. We should debounce storing the input value into state rather than debounce the render. I see (in `GraphQLExplorer`) that you use a regular `input`. There is a `DebouncedInput` component in this codebase, are you able to use that in place of the `input` tag?

develohpanda (changes requested):
Hi @Hademumel! Thank you for this PR and apologies for the delays, we have a lot of work flowing through.

I've just done a preliminary code review. In general this PR looks good, although there are some aspects of the implementation that are a little confusing and worth refactoring - please see my comments below. I have also pushed a commit to resolve conflicts and fix up some type definitions. 😄 

As an aside, when you update this PR, could you please make sure to exclude any extra changes that get included as a result of formatting rules in your code-editor (I presume)? It reduces the noise in a code-review and thus helps us to review faster. I updated these for now, but given there will need to be an iteration and further review, please keep it in mind. Thank you! 🙏🏽

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-type.tsx
Hademumel: I extracted the function to its own functional component

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-search-results.tsx
Hademumel: I also think thats clearer, I refactored according to the suggestion.

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-type.tsx
Hademumel: I extracted this to its own functional component

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-schema.tsx
Hademumel: Good point, I moved the SearchResults to the parent component

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-search-results.tsx
Hademumel: I refactored this, so the Debounced Component is used. I had to make some changes to it though.

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-explorer-search-results.tsx
```diff
@@ -0,0 +1,151 @@
...
+        {numberOfTypes > MAX_RENDERED_ELEMENTS && (
+          <p>And {numberOfTypes - MAX_RENDERED_ELEMENTS} more types found...</p>
+        )}
+      </>
+    );
+  }
+
+  renderFoundFields() {
+    const { onNavigateType, onNavigateField } = this.props;
+    const fields: GraphQLFieldWithParentName[] = [...this.state.foundFields];
```
develohpanda: Can this just use `this.state.Foundfields` or does it need to be a new array?

develohpanda (commented):
Thank you for making those changes! I [updated](https://github.com/Kong/insomnia/pull/3388/commits/c7e5fffa7276c7e4e2d3ed9b43eed878eb5f9d70) how refs are being used with `DebouncedInput` to match the existing pattern to simplify (even though the existing pattern is not great, consistency is king). Overall LGTM 🔥 I will hold off on approval until the following questions regarding UX have been addressed from the team.

While testing I noticed that empty states could be better represented. Perhaps with a note under each header (or change the header itself, or not show at all). @wdawson would you have any suggestions on behavior and/or wording?

image

image

Additionally, @erickrico could you please suggest how best to represent/style this concatenation message? This PR will only show 100 results, but if more are found it will present a message like below. Should it be italicized and/or use a fainter text color etc.

image

wdawson:
> While testing I noticed that empty states could be better represented. Perhaps with a note under each header (or change the header itself, or not show at all). @wdawson would you have any suggestions on behavior and/or wording?

@erickrico may have thoughts here too. But I'd suggest hiding the headers completely if nothing is found for them. If there are no hits across any of the headers, then I'd show some message like "No results found."

erickrico:
> > While testing I noticed that empty states could be better represented. Perhaps with a note under each header (or change the header itself, or not show at all). @wdawson would you have any suggestions on behavior and/or wording?
> 
> @erickrico may have thoughts here too. But I'd suggest hiding the headers completely if nothing is found for them. If there are no hits across any of the headers, then I'd show some message like "No results found."

I agree to remove the headers, we should just display something like "no results found, try agin with a different schema", and for the "show 30 more results" @develohpanda  i think using a lighter color works best, also what happens when a user click on that, will we load the 30+ results beneath the first batch or results?

develohpanda:
> what happens when a user click on that, will we load the 30+ results beneath the first batch or results?

Currently nothing, that's not a button, just a label. I think the intention is for users to make the search/filter more specific, though it would probably be better if it would show another batch of results 🤔 @erickrico

gatzjames:
👍 I tested it with a couple of different graphql schemas and it works really well!

Hademumel:
I worked on the proposed changes. Thanks for the suggestions. I used --color-surprise as the color for the "more elements found" hint. Im not sure if this is a good color to use, but I felt it was a good choice to differentiate between the other colors in the schema explorer.

develohpanda:
@Hademumel thank you for making those updates, much appreciated!

I have pushed a few more commits to improve the TS types (following on from my previous commits). I also pushed an update to replace search with a fuzzy search algorithm Insomnia already uses in other areas, but largely your changes are great! 🙌🏽  I have been testing with various schemas alongside and it seems to be working smoothly. I'm super excited for us to get this through, it's bound to be so useful.

develohpanda (approved):
### QA

<details>
<summary>🟢 Click to navigate to docs</summary>



</details>

<details>
<summary>🟢 Fuzzy search finds all results</summary>

image

</details>

<details>
<summary>🟢 Fuzzy search finds only types</summary>

image


</details>

<details>
<summary>🟢 Fuzzy search finds only fields</summary>

image

</details>

<details>
<summary>🟢 Fuzzy search finds nothing </summary>

image


</details>

<details>
<summary>🟢 Show more results, clear search </summary>

This batch size is set to 100, but for this demo I pushed it down to 5.




</details>

gatzjames (approved):
💯

dimitropoulos (approved):
Worked well with the GitHub API.  Thanks!

Screenshot_20210825_162835

PR #3390, opened 2021-05-17
fixes `redux-devtools`-specific error, `"sync" is not available`

dimitropoulos:
closes: INS-702

until https://github.com/reduxjs/redux-devtools/pull/711 is available to us, our best course of action is to remove it for now.


develohpanda (approved):
Oh wow, between the time I checked the linked issue this morning and now, it's already been merged! I'll leave it to you to re-instate the plugin once is published?

dimitropoulos:
Yep.  I followed up with them a bit ago re: when the next release will be, then from there I'll follow up with the `electron-devtools-installer` people to see if we can get a patch release with this updated dependency.  If not, we _may_ be able to do some dependency magic to force it, but I'm hoping to just do it the normal way and get everyone to release the fix.

develohpanda:
I think `electron-devtools-installer` doesn't actually bundle it, only stores the app store id, but worth double checking that!

PR #3391, opened 2021-05-17
fixes react controlled input error

dimitropoulos:
closes INS-703

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.tsx
+++ b/packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.tsx
@@ -270,10 +270,14 @@ class WorkspaceEnvironmentsEditModal extends PureComponent<Props, State> {
   }
 
   async _updateEnvironment(
-    environment: Environment,
+    environment: Environment | null,
     patch: Partial<Environment>,
     refresh = true,
   ) {
+    if (environment === null) {
+      return;
+    }
+
     const { workspace } = this.state;
     // NOTE: Fetch the environment first because it might not be up to date.
     // For example, editing the body updates silently.
@@ -301,7 +305,6 @@ class WorkspaceEnvironmentsEditModal extends PureComponent<Props, State> {
       clearTimeout(this.colorChangeTimeout);
     }
     this.colorChangeTimeout = setTimeout(async () => {
-      // @ts-expect-error -- TSCONVERSION environment can be null
       await this._updateEnvironment(environment, {
         color,
       });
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.tsx
```diff
@@ -348,24 +341,16 @@ class WorkspaceEnvironmentsEditModal extends PureComponent<Props, State> {
...
 
-  async _handleClickColorChange(environment: Environment) {
-    const color = environment.color || '#7d69cb';
-
+  _handleClickColorChange(environment: Environment) {
     if (!environment.color) {
-      await this._handleChangeEnvironmentColor(environment, color);
+      // TODO: fix magic-number. Currently this is the `suprise` background color for the default theme, but we should be grabbing the actual value from the user's actual theme instead.
+      const color = environment.color || '#7d69cb';
+      this._handleChangeEnvironmentColor(environment, color);
```
dimitropoulos:
```suggestion
      // TODO: fix magic-number. Currently this is the `surprise` background color for the default theme, but we should be grabbing the actual value from the user's actual theme instead.
      this._handleChangeEnvironmentColor('#7d69cb');
```

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.tsx
```diff
@@ -508,16 +498,17 @@ class WorkspaceEnvironmentsEditModal extends PureComponent<Props, State> {
...
                       />
                       {activeEnvironment.color ? 'Change Color' : 'Assign Color'}
                     </DropdownItem>
 
                     <DropdownItem
                       value={activeEnvironment}
-                      onClick={this._handleUnsetColor}
+                      onClick={() => {
+                        this._handleChangeEnvironmentColor(activeEnvironment, null);
+                      }}
```
develohpanda: Could you please provide context into this change (class function -> inline arrow function)?

dimitropoulos: In the commit message I mentioned "inlined one-liners". The reason for doing this was to reduce all the indirection so I could tell what was actually going on.  I put some of them back, but I left the one-line simple ones.  In this case, the main reason for doing this was that we are calling `this._getActiveEnvironment()` twice in a row just for the ceremony of preventing re-renders when (I would wager, but have not measured) the database call and subsequent array search that occurs to do so far outweighs whatever performance gain we got from the avoided re-render, which, by the way, would be solved by memoizing the callback anyway.
 
Still, in the interest of moving forward, I've reverted this in e801afea4f64ffce803e17ad8fb059d28bb3a295

develohpanda (approved):

PR #3395, opened 2021-05-18
use same `array-type` rule, project-wide

dimitropoulos:
This was a holdover from the initial TypeScript conversion that we had planned to fix after the initial large merge.  That time is now.


dimitropoulos:
(if this ever needs to be rebased, the correct course of action is to keep the first (one line) commit, rebase that, and then just run `npm run lint:fix` again, since that's literally all I did in the first place).

@develohpanda please feel free to merge at will

develohpanda (approved):

develohpanda:
Merging now to avoid rebasing

PR #3396, opened 2021-05-18
Feat/o2k improvements

dimitropoulos:
A recent PR (https://github.com/Kong/insomnia/pull/3340) on inso went through a lot of iterations to try and find a suitable solution to the problem, and during that time there were some small things fixed along the way.  I pulled those things out of the work on that PR and collected them here in the interest of cleanliness (but also because it looks like that PR may not merge super soon).

Best to read commit by commit, I tried to keep it very clean and document any changes where necessary.


packages/openapi-2-kong/src/declarative-config/upstreams.test.ts
```diff
@@ -4,6 +4,7 @@ import { getSpec } from './utils';
 
 const xKongUpstreamDefaults = 'x-kong-upstream-defaults';
 
+/** This function is written in such a way as to allow mutations in tests but without affecting other tests. */
```
develohpanda, line 7: 👍🏽

packages/openapi-2-kong/src/types/outputs.ts
```diff
@@ -1,32 +1,26 @@
...
-import {
-  KubernetesConfig,
-  KubernetesMethodConfig,
-  KubernetesPluginConfig,
-} from './kubernetes-config';
+import { KubernetesDocument } from './kubernetes-config';
 
 export type ConversionResultType = 'kong-declarative-config' | 'kong-for-kubernetes';
 
+export interface Warnings {
```
develohpanda: nit pick, this should be a singular `Warning` right?

packages/insomnia-inso/src/commands/generate-config.test.ts
```diff
@@ -14,7 +14,8 @@ const generate = _generate as jest.MockedFunction<typeof _generate>;
 const generateFromString = _generateFromString as jest.MockedFunction<typeof _generateFromString>;
 const writeFileWithCliOptions = _writeFileWithCliOptions as jest.MockedFunction<typeof _writeFileWithCliOptions>;
 
-const mockConversionResult: ConversionResult = {
+const mockConversionResult: KongForKubernetesResult = {
+  // @ts-expect-error -- TSCONVERSION the tests seem to suggest that this is valid, yet it is not allowed by the types.
```
develohpanda, line 18: Yeah I believe this is because the document is just a string rather than a complete K8s document. This was done for simplicity (because flow wasn't very helpful) but we can look to enforce it better now.

packages/openapi-2-kong/src/types/kubernetes-config.ts
```diff
@@ -79,6 +79,6 @@ export interface KubernetesConfig {
   spec: K8sSpec;
 }
 
-export type KubernetesDocument = KubernetesConfig
+export type KubernetesManifest = KubernetesConfig
```
develohpanda: 👍🏽

develohpanda (commented):
Need to QA so not approving as yet but code review checks out

packages/openapi-2-kong/src/types/outputs.ts
dimitropoulos: most definitely.  great catch.  Will update.

dimitropoulos: updated in https://github.com/Kong/insomnia/pull/3396/commits/33643d2d4f81079b1fe5e199a965740f62341e36

packages/insomnia-inso/src/commands/generate-config.test.ts
dimitropoulos: that's good to know, I don't want to get into it quite yet because I'd like to have a little more time to look at/understand how it works for these functions taking strings that represent json or yaml for valid specs, or objects that represent valid specs themselves.

ultimately, a package where we can have common types, basically we can use https://github.com/Kong/khcp/blob/master/apps/konnect-api/src/common/globals.ts#L23 as a jumping off point.  That way, we'll be able to (at least a little better) describe the contents of these strings.  it's a little bigger than what I want to dip into for this PR, but just to share what's on my mind.

develohpanda (approved):

PR #3397, opened 2021-05-19
Enable ESLint & TS for smoke tests

develohpanda:
What it says on the tin, minimal changes


packages/insomnia-smoke-test/tsconfig.build.json
```diff
@@ -0,0 +1,23 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "composite": true,
+    "outDir": "dist",
+    "rootDir": ".",
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true, // this is required because spectron depends on electron but it is not locatable by typescript for the purpose of types
+    "noImplicitAny": false
```
develohpanda, line 10: ‼️ temporary

dimitropoulos: cool

dimitropoulos (approved):
the only thing I found/could think of was: should the package.json have `main` and `types` entries?  non-blocking.

develohpanda:
> the only thing I found/could think of was: should the package.json have `main` and `types` entries? non-blocking.

Not needed strictly because this doesn't actually output anything. The entire package is a package of test suites.

PR #3400, opened 2021-05-19
Fixes clean scripts

dimitropoulos:
closes: INS-708

These are mostly fairly straightforward changes, all notated in the commit descriptions where more context may be helpful.  Please feel free to read commit-by-commit for an easier review experience.

The only one that I can see has any potential to make any logical change is the last commit, see the commit description for more.


packages/insomnia-app/package.json
```diff
@@ -6,7 +6,7 @@
   "description": "The Collaborative API Design Tool",
   "author": "Kong <office@konghq.com>",
   "license": "MIT",
-  "main": "app/main.min.js",
+  "main": "build/main.min.js",
```
DMarby, line 9: I wonder if this is actually used by anything, given that we just managed to package the app with this being wrong and running for development seem to be working as well? 😅

DMarby (approved):

packages/insomnia-app/package.json
dimitropoulos: it's used by tooling a lot, yeah.  The main reason we didn't see a big problem is that we don't consume `insomnia-app`'s NPM anywhere (like, as a library) since it's the top of the iceberg, but we really should have this.  We also need a reference to the root types, but I guess I'll do that at a later date.  This is enough for one PR.

packages/insomnia-components/package.json
```diff
@@ -15,15 +15,17 @@
...
+    "clean": "rimraf dist storybook-static && npm run clean:svg",
+    "clean:svg": "rimraf src/assets/svgr",
+    "clean:svg:index": "rimraf src/assets/svgr/index.tsx",
+    "convert-svg": "npm run clean:svg && svgr --config-file svgr.config.js --out-dir src/assets/svgr src/assets && npm run clean:svg:index",
     "jest:watch": "jest --watch",
     "lint": "eslint . --ext .js,.ts,.tsx",
     "lint:fix": "npm run lint -- --fix",
-    "postclean": "rimraf dist",
-    "prepare": "npm run convert-svg && npm run build",
+    "prepare": "npm run build",
```
develohpanda, line 28: @dimitropoulos some context from the commit message - `lerna bootstrap` executes `npm install` in each package, which in turn calls `prepare` as part of the NPM [lifecycle](https://docs.npmjs.com/cli/v7/using-npm/scripts#npm-install), resulting in SVG generation and build during bootstrap.

Because the recommended setup _was_ `npm run bootstrap && npm run app-start` from lerna root, this wasn't an issue (but granted, some misdirection). 

Now with the addition of `npm run build` at lerna root the misdirection became evident, and using the `prebuild` script looks like a great solution to it!

packages/insomnia-app/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "scripts": {
     "clean": "tsc --build tsconfig.webpack.json --clean && tsc --build tsconfig.build.json --clean",
+    "postclean": "rimraf build",
```
develohpanda: Do we also want to delete `dist` here (the result of packaging, not building)? I would think so

> I'll be honest, I'm not entirely sure what the reason is for having `postclean` scripts in every package with (mostly) `rimraf dist` (i.e. instead of just including that in the main clean script at the end), but that's clearly the convention - so for now I'm just following it

Inherited, we can totally change it up

packages/insomnia-components/package.json
dimitropoulos: yep, I sorta gathered that, although what you said helps a lot to confirm.  Thanks, I'm happy with the consistency and lack-of-pitfalls of the prebuild approach.

packages/insomnia-app/package.json
dimitropoulos: I'm not sure, to be honest about `dist`.  Evidently (by the fact that a portion of this PR needed to be reverted) to me, it's still not clear what _exactly_ is going on with where dependencies are being dropped and why (and for dev or prod).

dimitropoulos: Thanks for mentioning it re: postclean.  I don't at all mind how it is now, just was wondering really.

PR #3401, opened 2021-05-19
fixes build artifact location for local development

dimitropoulos:
In the prior PR (https://github.com/Kong/insomnia/pull/3400) on updating the clean the following sequence of events happened.

1. I ran the `build` script on the `insomnia-app` package
1. The build succeeded
1. I updated the `clean` script to remove all build artifacts (the subject matter of that PR)
1. I ran the `build` script on the `insomnia-app` package
1. The build succeeded
1. I noticed that the recent build did not include `app/main.min.js`.
    - I, therefore, concluded that this file was not output by the build.
    - It would make sense that it was still there because
        - I know it _used to be_ output by the build script
        - the build scripts were recently updated
        - it would still be there in my filetree from days or weeks ago because it's in the `.gitignore`
  
The wrong assumption turned out to be that the build produced artifacts in either case, despite it actually succeeding.  Unlike every other package, to _actually_ build `insomnia-app` locally, you must trick it by running `SMOKE_TEST=true npm run build`.  The exit code if you don't do this is `0`.  This should be changed, but I have done so in a separate PR https://github.com/Kong/insomnia/pull/3402 to allow for discussion there.

For now, I want to make sure `develop` is good, so I am submitting this PR to fix the immediate problem.

The good thing to note is that the change required only affects local development, at least so far as I can tell.

### DIFF
--- a/packages/insomnia-app/.eslintignore
+++ b/packages/insomnia-app/.eslintignore
@@ -1,6 +1,7 @@
 build
 bin
 send-request
+**/main.min.js
 webpack.config.*.js
 webpack.config.*.js.map
 webpack.config.*.d.ts
--- a/packages/insomnia-app/.gitignore
+++ b/packages/insomnia-app/.gitignore
@@ -1,5 +1,5 @@
 dist
 build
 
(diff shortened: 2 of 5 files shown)


reynolek (approved):

khyveasna11111908:
Inbox me to the store and I'll have

PR #3402, opened 2021-05-19
fixes error code used when build fails

dimitropoulos:
closes: INS-709

So far as I can tell, this behavior was introduced more than a year ago in https://github.com/Kong/insomnia/commit/c6a7c4d68272b3c4f3d78d4dc9031ae8e49b3f7a#diff-a3f8249a67ca4e9b7a97494a05d806944b48ea1ff0a37ca4774c3782d5e90607R24.

Since this just caused a bug (https://github.com/Kong/insomnia/pull/3401), it seems like a good thing to fix.

## Before
Before this PR, it would look like this:
Screenshot_20210519_120001

## After
And with this change, it looks like this:
Screenshot_20210519_115848

## Thoughts?

I can't think of a reason or use-case where we'd want the exit code to be success (exit code `0`) when it didn't build.  I sorta can understand that it's a tad murky in this case, but still.  We really need `npm run build` to like... actually work locally without hacking it with a `SMOKE_TEST` environment variable.

### DIFF
--- a/packages/insomnia-app/scripts/build.ts
+++ b/packages/insomnia-app/scripts/build.ts
@@ -198,7 +198,7 @@ export const start = async ({ forceFromGitRef }: { forceFromGitRef: boolean }) =
         console.log(`[build] git ref \`${gitRef}\` found`);
       }
       console.log('[build] Skipping build because no version was found (the version is derived from the git ref)');
-      process.exit(0);
+      process.exit(1);
     }
 
     if (appConfig.version !== version) {


develohpanda:
> So far as I can tell, this behavior was introduced more than a year ago in [c6a7c4d#diff-a3f8249a67ca4e9b7a97494a05d806944b48ea1ff0a37ca4774c3782d5e90607R24](https://github.com/Kong/insomnia/commit/c6a7c4d68272b3c4f3d78d4dc9031ae8e49b3f7a#diff-a3f8249a67ca4e9b7a97494a05d806944b48ea1ff0a37ca4774c3782d5e90607R24).

This commit was the initial designer merge from a [separate fork](https://github.com/Kong/studio/) around a year ago, in case you ever need to dig further into history (I've had to a couple of times to get context)

> I can't think of a reason or use-case where we'd want the exit code to be success (exit code 0) when it didn't build

Agreed, be it a silent failure or actively skipped it still didn't _build_.

> without hacking it with a SMOKE_TEST environment variable

In the build script there is a check to make sure the tag matches the version it's trying to build (which is a good safety net to retain when building for publishing) but that meant we needed some way bypassing that validation, which at the time was the SMOKE_TEST variable.

By default the build script validates the tag & version, and we disable the validation with the SMOKE_TEST variable. Perhaps that could be flipped, where by default it _doesn't_ validate, and instead is enabled if some condition is met (eg. a flag in CI? something else?)

develohpanda (approved):

DMarby (approved):

khyveasna11111908 (approved):
How are things there I think it's just the store

PR #3405, opened 2021-05-19
updates codemirror (in anticipation of upcoming PR bugfix)

dimitropoulos:
https://github.com/codemirror/CodeMirror/pull/6690 is the PR that fixes https://linear.app/insomnia/issue/INS-710/typeerror-cannot-read-property-hasgutter-of-undefined which is an error that occurs quite commonly (`Cannot read property 'hasGutter' of undefined`).

Doing these very-minor-fixes now means that once that PR is released we'll simply just need to update our dependency (and no further changes).

Fixes INS-710


packages/insomnia-app/app/codemirror.d.ts
```diff
@@ -0,0 +1,13 @@
+import 'codemirror';
+
+interface InsomniaExtensions {
+  isHintDropdownActive: () => boolean;
+  closeHintDropdown: () => void;
+}
```
develohpanda: Cool! There are a few more defined in `app/ui/components/codemirror/extensions/` if you wanted to add them, likely used in other areas of the code and not in code-editor itself

develohpanda:
Code checks out, waiting on the bump of code-mirror to QA one time. Shall we mark as draft until then?

packages/insomnia-app/app/codemirror.d.ts
dimitropoulos: oh cool, I should have checked (I assumed they were all defined in one place, but now I can see that how they're _actually_ organized is also very sensible).  Added in f52de783ccaae186a9352375d27b60187034d429.

develohpanda (approved):
Looks to be working! Will merge once CI passes.

PR #3407, opened 2021-05-20
show both name and uri during import through insomnia protocol

develohpanda:
Fixes INS-731

Protocols to test with:
* live: `insomnia://app/import?uri=uri-param&name=name-param`
* this PR: `insomniadev://app/import?uri=uri-param&name=name-param` (after `npm run app-start`, go to this URL in your browser)

Users would previously see this:
image

Users will now see this:
image

To try with valid import data:
* live: `insomnia://app/import?uri=https://gist.githubusercontent.com/gschier/4e2278d5a50b4bbf1110755d9b48a9f9/raw/801c05266ae102bcb9288ab92c60f52d45557425/petstore-spec.yaml&name=should-not-be-hidden`
* this PR: `insomniadev://app/import?uri=https://gist.githubusercontent.com/gschier/4e2278d5a50b4bbf1110755d9b48a9f9/raw/801c05266ae102bcb9288ab92c60f52d45557425/petstore-spec.yaml&name=should-not-be-hidden`

> Note: this PR does not address how the prompt looks, it's already looked bad with a long URL, and looks worse now with the name property.


dimitropoulos:
The code certainly looks just fine, but I tried to test this and learned that I can't, for whatever reason, open these links, despite insomnia running.

The manual way on linux would be to call xdg-open directly, and when I do, this happens:
Screenshot_20210520_082330

dimitropoulos:
according to https://github.com/electron-userland/electron-builder/issues/4035#issuecomment-621131126 followed by https://github.com/electron-userland/electron-builder/pull/4909#issuecomment-696557530 this was fixed in https://github.com/electron-userland/electron-builder/releases/tag/v22.9.1, so I've upgraded `electron-builder` (and `electron-builder-squirrel-windows` to be in sync) to `v22.11.5` (we were on `v22.5.1`) to see if it works, and it did not....

reynolek (approved):
Verified on mac

PR #3408, opened 2021-05-20
adds logging for when setting default protocol client fails

dimitropoulos:
closes: INS-732

Screenshot_20210520_082330

This fails on Linux, but not sure why yet.  Here is an example with logging from this PR:
Screenshot_20210520_085548

see https://www.electronjs.org/docs/api/app#appsetasdefaultprotocolclientprotocol-path-args


dimitropoulos:
according to https://github.com/electron-userland/electron-builder/issues/4035#issuecomment-621131126 followed by https://github.com/electron-userland/electron-builder/pull/4909#issuecomment-696557530 this was fixed in https://github.com/electron-userland/electron-builder/releases/tag/v22.9.1, so I've upgraded `electron-builder` (and `electron-builder-squirrel-windows` to be in sync) to `v22.11.5` (we were on `v22.5.1`) to see if it works, and it did not....

Due to that, I guess this PR is as far as we're gonna get for now.  It's helpful, anyway

develohpanda (approved):

PR #3409, opened 2021-05-20
fixes 'previewHidden' of undefined error

dimitropoulos:
closes: INS-733

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper-design.tsx
+++ b/packages/insomnia-app/app/ui/components/wrapper-design.tsx
@@ -26,7 +26,6 @@ interface Props {
 }
 
 interface State {
-  previewHidden: boolean;
   lintMessages: {
     message: string;
     line: number;
@@ -42,7 +41,6 @@ class WrapperDesign extends PureComponent<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      previewHidden: props.wrapperProps.activeWorkspaceMeta?.previewHidden || false,
       lintMessages: [],
     };
   }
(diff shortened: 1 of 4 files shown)


packages/insomnia-app/app/ui/components/wrapper-design.tsx
```diff
@@ -58,18 +56,9 @@ class WrapperDesign extends PureComponent<Props, State> {
...
-        const previewHidden = this.state.previewHidden;
-        await models.workspaceMeta.updateByParentId(workspaceId, {
-          previewHidden,
-        });
-      },
-    );
+    const workspaceId = this.props.wrapperProps.activeWorkspace._id;
+    const previewHidden = Boolean(this.props.wrapperProps.activeWorkspaceMeta?.previewHidden);
+    await models.workspaceMeta.updateByParentId(workspaceId, { previewHidden: !previewHidden });
   }
```
develohpanda, line 62: I very nearly missed all of the notes you put in the commit message. It _does feel_ a little less snappy for a UI toggle but this could be a combination of hardware differences, and/or mental bias while visually comparing alongside the live application.

I see `workspaceMeta.previewHidden` as just persisting the state in the background to be able to restore later, and that the visual toggle itself should not be dependent on a database round-trip. For this reason, my vote would be to revert this back to how it was, but there are likely plenty of areas in the app which don't follow this guideline as well.

I'll let you consider the pros and cons and choose an approach.

develohpanda (approved):

develohpanda (approved):

packages/insomnia-app/app/ui/components/wrapper-design.tsx
dimitropoulos: Ok, thanks for mentioning that.  I've been mulling it over all morning and I think I want to keep the commit for these reasons:
1. there is no perceptible performance difference, as measured, even on the fastest gaming monitors (and, anyway, a 1ms is not perceptible to humans in sound or sight).
1. this is considered an antipattern, generally speaking: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html (although exceptions exist, I don't think this is one of them)
1. I perceive the place to optimize this, if optimization is necessary, to be at the db level itself.  we could cache the DB in memory, for example, in a different way that wouldn't cost as much time which would have a cast-a-wide-net effect on all calls in the app.
1. and, as you said, this isn't a pattern that's very common in the code in general so I'd like to keep things simple

PR #3414, opened 2021-05-21
INS-674 - First pass at making the send button more visible

### LINKED ISSUE #3152: Make the Send button more visually prominent
The new layout and design (both light and dark mode) in 2021.1.0 makes it harder to quickly locate the 'Send' button on a request. I believe the addition of the new sync button at the top right as well as the extra header space and theme changes has made the 'Send' button merge into the background more. At least for my workflow, the 'Send' button is what I click on the most, so it should more more prominent than everything else.

I'd like to see the Send button stand out a little more from it's surrounding elements, either by making it's appearance more striking, or additionally toning down some of the surrounding elements slightly.

Muscle memory might eventually counter these changes for me, but it's sure to be frustrating to newer users. The design language really needs to lend itself to what's most important.

reynolek:
Closes #3152
Closes INS-674

This is a first pass attempt at a way to make the send button more visible in the current layout. I verified all current core themes to make sure each had the improvement.

Below are some example screenshots of different themes, with hover states as well.

Screen Shot 2021-05-21 at 2 12 20 PM
Screen Shot 2021-05-21 at 2 14 14 PM
Screen Shot 2021-05-21 at 2 12 30 PM
Screen Shot 2021-05-21 at 2 14 33 PM
Screen Shot 2021-05-21 at 2 12 48 PM
Screen Shot 2021-05-21 at 2 14 44 PM

### DIFF
--- a/packages/insomnia-components/src/button/button.tsx
+++ b/packages/insomnia-components/src/button/button.tsx
@@ -8,12 +8,17 @@ export const ButtonSizeEnum = {
   Medium: 'medium',
 } as const;
 
+// These variants determine how the `bg` color variable is handled
+// Outlined sets the `bg` color as an outline
+// Contained sets the `bg` color as the background color
+// Text sets the `bg` color as the text color
 export const ButtonVariantEnum = {
   Outlined: 'outlined',
   Contained: 'contained',
   Text: 'text',
 } as const;
 
+// Sets the `bg` color to a themed color
 export const ButtonThemeEnum = {
   Default: 'default',
   Surprise: 'surprise',
--- a/packages/insomnia-app/app/ui/components/buttons/grpc-send-button.tsx
+++ b/packages/insomnia-app/app/ui/components/buttons/grpc-send-button.tsx
@@ -12,8 +12,9 @@ interface Props {
 
 const buttonProps: ButtonProps = {
   className: 'tall',
+  bg: 'surprise',
   size: 'medium',
-  variant: 'text',
+  variant: 'contained',
   radius: '0',
 };
 
--- a/packages/insomnia-app/app/ui/components/settings/theme.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/theme.tsx
@@ -69,6 +69,10 @@ class Theme extends PureComponent<Props, State> {
               height="10%"
               className="theme--pane__header--sub bg-fill"
             />
+            {/* Send Button */}
+            <g>
+              <rect x="53%" y="10%" width="9%" height="10%" className="success-fill" />
+            </g>
           </g>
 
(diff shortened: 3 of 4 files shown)


dimitropoulos:
I rebased it for you and added 33fbde6e14cc3f646bbfd61df62dcf90564556ec which does:


The color isn't right (obvi) but I had a moment to throw it in there.  Since this change is all about making it standout/so prominent, I thought that it would be appropriate for the preview to have it.  Not 100% sure how to theme it, but I'm sure we can figure it out by observing how the other things (e.g. sidebar) are themed.  I only had a few moments between two calls to do it so I thought it'd be better than nothing to get the ball rolling with at least placing it.

reynolek:
Ah interesting! I didnt even think to update that little SVG. I'll dig in on how to get proper colors in.

reynolek:
Updated the colors

image

dimitropoulos:
no biggie, but I'm curious: why is it that in the core default the color for the dark theme is `#a39af9` while the swatch is `#796dc5`.  For all of the rest of them it seems to match, but for just core-default it's brighter.

image

develohpanda (approved):
LGTM!

develohpanda (changes requested):
If you add a long URL, there's no padding any more (there wasn't before either, but by virtue of the button having the same background color it didn't matter)

image

The button for gRPC also needs the same theme color: https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/buttons/grpc-send-button.tsx
image

reynolek:
Code updated to change the gRPC button (since it uses `insomnia-components` as its more modern) and updated the CSS to add a small left margin to prevent text from pushing up against the rest/graphql send button

packages/insomnia-components/src/button/button.tsx
```diff
@@ -8,12 +8,17 @@ export const ButtonSizeEnum = {
   Medium: 'medium',
 } as const;
 
+// These variants determine how the `bg` color variable is handled
+// Outlined sets the `bg` color as an outline
+// Contained sets the `bg` color as the background color
+// Text sets the `bg` color as the text color
```
develohpanda, line 14: Nice comments - this is kind of messy though, I'm hoping when we come back to this we can clean it up. 🤞🏽

develohpanda (approved):
Looks so fresh now!

image

image

packages/insomnia-components/src/button/button.tsx
reynolek: Yeah all good -- This type of thing just took longer to grok than it should have (although some of that was because live reloading wasnt working for me properly...).

Its at least something, and will be useful in its current form until better addressed.

PR #3419, opened 2021-05-24
[o2k] copy plugin objects as is (and respect enabled flag)

dimitropoulos:
closes: INS-648
note: this is stacked off of https://github.com/Kong/insomnia/pull/3416

@Tieske please take a look at https://github.com/Kong/insomnia/pull/3419/commits/d21787c2112c931e5458585c7027ae56c08e85d1 which respects the enabled flags and https://github.com/Kong/insomnia/pull/3419/commits/f27bc298f0eca06fe6732b709dba7ec0a8dbd8ee which allows passing more than just `name`, `config`, and `tags`.

If those are correct, the only remaining question is whether it's also correct that we should be adding tags into the resulting plugin config.  Presently we do, and the current test (intentionally, for now, until I hear from you) fails on this:


Tieske (commented):
lgtm

- the `enabled` property should not be generated. If it is specified in the input, take that, if not then do not add it. Result is that the user specified value is honored, and if not specified, it will fall back on Kong defaults. So this part is ok.
- the `tags` property should be added to the plugin config (as for each Kong entity). So the code is right, and the test needs updating.

dimitropoulos:
ok, I added the tags like you said in 4ccc12909289f9c24a80782d99ee6f53ff21430f

reynolek (approved):

packages/openapi-2-kong/src/declarative-config/plugins.ts
```diff
@@ -147,10 +147,16 @@ export function generateRequestValidatorPlugin({ plugin, tags, operation }: {
...
+
   const dcPlugin: DCPlugin = {
     config,
-    tags: [...(tags ?? []), ...(plugin.tags ?? [])],
-    enabled: Boolean(plugin.enabled ?? true),
+    tags: [
+      ...(tags ?? []),
+      ...(plugin.tags ?? []),
+    ],
+    ...enabled,
```
develohpanda, line 159: I wonder if we should skip tags entirely if both `tags` and `plugin.tags` are empty, to not give the final plugin an empty tags property.

packages/openapi-2-kong/src/declarative-config/services.test.ts
```diff
@@ -142,25 +149,28 @@ describe('services', () => {
       };
       spec.paths['/cats'].get = {};
       // operation req validator plugin
-      spec.paths['/cats'].post[xKongPluginRequestValidator] = {
-        config: {
-          parameter_schema: 'operation',
-        },
-      };
+      if (spec.paths['/cats'].post) {
```
develohpanda, line 152: Hmmm I generally avoid conditionals in unit tests because they can start causing false positives, but I assume this is needed for types to resolve correctly?

Is there another approach we can use here, or add `else` cases to fail the test if the conditional is not met?

packages/openapi-2-kong/src/declarative-config/services.ts
```diff
@@ -159,6 +159,6 @@ export function generateRouteName(
 
   // replace all `/` with `-` except the ones at the beginning or end of a string
   const replacedRoute = routePath.replace(/(?!^)\/(?!$)/g, '-');
-  const pathSlug = generateSlug(pathItem['x-kong-name'] || replacedRoute);
+  const pathSlug = generateSlug(pathItem[xKongName] || replacedRoute);
```
develohpanda, line 162: I love this approach 💯

packages/openapi-2-kong/src/declarative-config/utils.ts
```diff
@@ -1,7 +1,7 @@
 import { OpenApi3Spec } from '../types/openapi3';
 
 /** This function is written in such a way as to allow mutations in tests but without affecting other tests. */
-export const getSpec = (): OpenApi3Spec =>
+export const getSpec = (overrides: Partial<OpenApi3Spec> = {}): OpenApi3Spec =>
```
develohpanda, line 4: 👍🏽

packages/openapi-2-kong/src/types/openapi3.ts
```diff
@@ -1,39 +1,40 @@
...
-  'x-kong-plugin-request-validator'?: {
-    enabled: boolean;
+  [xKongPluginRequestValidator]?: {
+    enabled?: boolean;
     config: {
-      verbose_response: boolean,
-    }
-  }
+      verbose_response?: boolean;
+      parameter_schema?: 'global' | 'path' | 'operation';
```
develohpanda, line 30: This property doesn't seem right at all, but it also has a `bodySchema` which is defined under the `DcPluginConfig`.

Documentation here: https://docs.konghq.com/hub/kong-inc/request-validator/#parameters

packages/openapi-2-kong/src/declarative-config/services.test.ts
dimitropoulos: Screenshot_20210526_103126

Yep, exactly.  We can cast.  I'll do so in the next PR.

dimitropoulos: 989adc7cf0a71a71ebf7b70187ce21d793b6557a

packages/openapi-2-kong/src/declarative-config/services.ts
dimitropoulos: :)

packages/openapi-2-kong/src/types/openapi3.ts
dimitropoulos: while I agree with you and the docs agree with you, the code (ours) does not: https://github.com/Kong/insomnia/blob/develop/packages/openapi-2-kong/src/declarative-config/services.test.ts#L140-L262

Let's follow up on this separately, like you're saying.  The current type simply satisfies the existing code.  If the existing code is wrong, I'm a full supporter of fixing :)

develohpanda: Ya current code would be wrong (atleast in unit tests). 👍

packages/openapi-2-kong/src/declarative-config/plugins.ts
dimitropoulos: the next PR does a lot of work with tags, take a look there and see if your opinion's changed any. #3428

PR #3423, opened 2021-05-26
Skip apt repo publish if not stable tag

develohpanda:
Closes INS-742

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Ensure stable release tag
       id: stable-tag
       run: |
-        if [[ ${{ github.event.inputs.tag }} =~ ^core@\d{4}\.\d+\.\d+$$ ]]; then
+        if [[ ${{ github.event.inputs.tag }} =~ ^core@\d{4}\.\d+\.\d+$ ]]; then
           echo ::set-output is-stable=match::true
         fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -8,19 +8,36 @@ on:
...
+
   apt_repository:
     name: Publish to apt repository
     runs-on: ubuntu-latest
+    needs: extract_channel
+    if: needs.extract_channel.outputs.channel == 'stable'
 
     steps:
     - name: Download .deb from GitHub release
+      if: steps.stable-tag.outputs.is-stable == 'true'
```
DMarby: I guess this is not needed anymore since the check is moved to a job instead? Same for the one below

DMarby (changes requested):

DMarby (approved):

develohpanda:
@DMarby I tested this and it didn't exit/continue properly 😆 will need to pair briefly

DMarby:
> @DMarby I tested this and it didn't exit/continue properly 😆 will need to pair briefly

Forgot that we were doing POSIX regex when reviewing 😅 `\d` isn't supported, so we'll need to change it to this instead and it should work:
```
^core@[0-9]{4}\.[0-9]+\.[0-9]+$
```

develohpanda:
Looks like it's working now!

PR #3424, opened 2021-05-26
Add space data model

develohpanda:


packages/insomnia-app/app/models/space.ts
```diff
@@ -0,0 +1,45 @@
...
+import { generateId } from '../common/misc';
+
+export const name = 'Space';
+export const type = 'Space';
+export const prefix = 'sp';
+export const canDuplicate = false;
+export const canSync = true;
+
+interface BaseSpace {
+  name: string;
```
DMarby, line 12: Not really relevant for this PR, but how are you imagining that associating spaces with their corresponding remote spaces in the API will work later on?
Are we going to add a `remoteId` or similar that will indicate that it's a remote space, or will the database ID of the space just be set to the remote ID? (would mean that it's not using the `sp` prefix, since space ID's start with `team_` in the API), or do something different entirely?

Given that we sync the entire workspace object we might need to make sure that we override the parentId of workspaces when pulling changes depending on our approach 🤔

packages/insomnia-app/app/models/__tests__/workspace.test.ts
```diff
@@ -26,8 +26,6 @@ describe('migrate()', () => {
     for (const cert of certs) {
       expect(typeof cert.modified).toBe('number');
       expect(typeof cert.created).toBe('number');
-      delete cert.modified;
```
DMarby, line 29: Would be good to update the comment above as well if we're no longer deleting these? Seems like this is currently breaking the tests however

packages/insomnia-app/app/models/space.ts
```diff
@@ -0,0 +1,45 @@
+import { database as db } from '../common/database';
+import type { BaseModel } from './index';
+import { generateId } from '../common/misc';
+
+export const name = 'Space';
+export const type = 'Space';
+export const prefix = 'sp';
+export const canDuplicate = false;
+export const canSync = true;
```
DMarby: Not sure if `canSync` should be true, given that this is mostly a collection to group things, which I'd expect to be created by user action as needed (eg when switching to a remote space for the first time or choosing to create a local space)?
At least with the sync API, we sync objects with the database on a workspace basis, so syncing the space for each workspace would definitely break things/cause conflicts.

DMarby (approved):

packages/insomnia-app/app/models/space.ts
develohpanda: Left it out of this PR because I wasn't convinced about an approach yet 😆 I'm thinking to track the team id on a separate `remoteId` field, instead of it be the database id. Removing that coupling from the remote data will give us local/unlinked spaces for free.

(The initial description in the linear ticket has a `type: 'remote' | 'local'` field but not going ahead with that)

What do you think?

packages/insomnia-app/app/models/__tests__/workspace.test.ts
develohpanda: Hmm I was getting errors locally; I'll look at this again and add comments if needed 👍🏽

packages/insomnia-app/app/models/space.ts
develohpanda: Yes that's valid, I'll flip this to false. Need to validate it doesn't end up in any git-sync workflows too, since the `canSync` flag isn't used everywhere it should be 🤣 🙄

packages/insomnia-app/app/models/space.ts
DMarby: I think that sounds reasonable

PR #3427, opened 2021-05-27
Add Space selector, with the logic to select a space

develohpanda:
This PR introduces the space selector, and links it into Redux. It uses a constant - `BASE_SPACE_ID` - to determine the base space, which doesn't actually exist, but will be used to filter the list of workspaces with no parent, in the future.

This PR does not have any UI for creating or deleting a space, however for testing purposes you can do so via the command line. When running in dev mode, run `await models.space.create({name: 'some name'});` in the developer console to create spaces, which will then show in the space dropdown and be selectable.

Any additional items in the dropdown list (eg settings, logout, create) will be added when implementing their functionality too.

This PR also tidies up the breadcrumbs component and its props.

### DIFF
--- a/packages/insomnia-app/app/ui/components/workspace-page-header.tsx
+++ b/packages/insomnia-app/app/ui/components/workspace-page-header.tsx
@@ -53,7 +53,7 @@ const WorkspacePageHeader: FunctionComponent<Props> = ({
         <Fragment>
           <img src={coreLogo} alt="Insomnia" width="24" height="24" />
           <Breadcrumb crumbs={[{ id: 'home', node: strings.home, onClick: homeCallback },
-{ id: 'workspace', node: <Fragment key="workspace-dd">{workspace}</Fragment> }]}/>
+            { id: 'workspace', node: <Fragment key="workspace-dd">{workspace}</Fragment> }]}/>
         </Fragment>
       }
       gridCenter={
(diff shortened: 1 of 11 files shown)


packages/insomnia-app/app/ui/components/workspace-page-header.tsx
```diff
@@ -44,18 +45,16 @@ const WorkspacePageHeader: FunctionComponent<Props> = ({
...
         <Fragment>
           <img src={coreLogo} alt="Insomnia" width="24" height="24" />
           <Breadcrumb
-            className="breadcrumb"
-            // @ts-expect-error -- TSCONVERSION
-            crumbs={[strings.home, <Fragment key="workspace-dd">{workspace}</Fragment>]}
-            onClick={homeCallback}
-          />
+            crumbs={[{ id: 'home', node: strings.home, onClick: homeCallback },
+              { id: 'workspace', node: <Fragment key="workspace-dd">{workspace}</Fragment> }]}/>
```
develohpanda, line 57: Need to add an eslint rule to force arrays onto new lines/format this in a readable manner

dimitropoulos (approved):
Screenshot_20210527_152558

Works as advertised :)

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -377,7 +377,7 @@ class WrapperHome extends PureComponent<Props, State> {
             gridLeft={
               <Fragment>
                 <img src={coreLogo} alt="Insomnia" width="24" height="24" />
-                <Breadcrumb className="breadcrumb" crumbs={[getAppName()]} />
```
dimitropoulos, line 380: I was gonna ask if there was corresponding CSS for `.breadcrumb` that needs to be removed, but it looks like there isn't.

develohpanda:
CI smoke test failed in the last run on linux because it failed to launch the app, but the review above is from linux. Merging this

PR #3433, opened 2021-05-31
Create, rename and delete a local space

develohpanda:
## Create new space
image
image
image

## Settings
image

### Not visible when base space is selected
image

## Delete
image

> Note: we may want to iterate on/improve the space deletion flow to seek confirmation in a stricter way (type in space name).

> Note: UI tests have not yet been added - I tried to write some but our Redux implementation needs a little work before it's ready for quality testing of components that are linked to Redux.

### DIFF
--- a/packages/insomnia-app/app/test-utils.ts
+++ b/packages/insomnia-app/app/test-utils.ts
@@ -1,16 +1,12 @@
 import * as modals from './ui/components/modals';
 import type { ErrorModalOptions } from './ui/components/modals/error-modal';
+import { PromptModalOptions } from './ui/components/modals/prompt-modal';
 
-export const getAndClearShowPromptMockArgs = () => {
+export const getAndClearShowPromptMockArgs = (): PromptModalOptions => {
   const mockFn = modals.showPrompt as jest.Mock;
-  const { title, submitName, defaultValue, onComplete } = mockFn.mock.calls[0][0];
+  const options = mockFn.mock.calls[0][0] as PromptModalOptions;
   mockFn.mockClear();
-  return {
-    title,
-    submitName,
-    defaultValue,
-    onComplete,
-  };
+  return options;
 };
 
 export const getAndClearShowAlertMockArgs = () => {
--- a/packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
+++ b/packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
@@ -71,7 +71,7 @@ class SpaceSettingsModal extends PureComponent<Props> {
         <ModalBody key={`body::${space._id}`} className="pad">
           <div className="form-control form-control--outlined">
             <label>
-            Name
+              Name
               <DebouncedInput
               // @ts-expect-error -- TSCONVERSION props are spread into an input element
                 type="text"
--- a/packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
+++ b/packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
@@ -56,59 +56,42 @@ class SpaceSettingsModal extends PureComponent<Props> {
     this.modal && this.modal.hide();
   }
 
-  renderModalHeader() {
-    const { space } = this.props;
-    if (!space) {
-      return null;
-    }
-
-    return (
-      <ModalHeader key={`header::${space._id}`}>
-        {strings.space.singular} Settings{' '}
-        <div className="txt-sm selectable faint monospace">{space._id}</div>
-      </ModalHeader>
-    );
-  }
-
-  renderModalBody() {
-    const { space } = this.props;
-    if (!space) {
-      return null;
-    }
-
-    return (
-      <ModalBody key={`body::${space._id}`} className="pad">
-        <div className="form-control form-control--outlined">
-          <label>
-            Name
-            <DebouncedInput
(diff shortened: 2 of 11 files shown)


dimitropoulos:
(I rebased this to get the changes for the string-to-single-object PR that merged before it)

packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+interface ReduxDispatchProps {
+  handleRemoveSpace: (space: Space) => void;
+}
+
+interface Props extends ReduxStateProps, ReduxDispatchProps { }
+
+@autoBindMethodsForReact(AUTOBIND_CFG)
+class SpaceSettingsModal extends PureComponent<Props> {
```
dimitropoulos, line 28: is there any reason not to use a function component here?  my understanding (perhaps uncommunicated?) was that we are trying not to make class components going forward.

develohpanda: The API for modals currently only works with class components, to my understanding, have a look at how all modals are shown. I don't want to spend time changing that right now.

packages/insomnia-app/app/ui/components/request-url-bar.tsx
```diff
@@ -182,13 +182,16 @@ class RequestUrlBar extends PureComponent<Props, State> {
       inputType: 'decimal',
       title: 'Send After Delay',
       label: 'Delay in seconds',
+      // @ts-expect-error TSCONVERSION string vs number issue
```
dimitropoulos, line 185: why can't we change `PromptModalOptions` to use `string | number`, or, make it generic?

packages/insomnia-app/app/test-utils.ts
```diff
@@ -1,16 +1,12 @@
 import * as modals from './ui/components/modals';
 import type { ErrorModalOptions } from './ui/components/modals/error-modal';
+import { PromptModalOptions } from './ui/components/modals/prompt-modal';
 
-export const getAndClearShowPromptMockArgs = () => {
+export const getAndClearShowPromptMockArgs = (): PromptModalOptions => {
```
dimitropoulos, line 5:
```ts
import { showAlert, showError, showModal, showPrompt } from './ui/components/modals';

export const getAndClearShowPromptMockArgs = () => {
  const mockFn = showPrompt as jest.Mock<typeof showPrompt, Parameters<typeof showPrompt>>;
  const options = mockFn.mock.calls[0][0];
  mockFn.mockClear();
  return options;
};

export const getAndClearShowAlertMockArgs = () => {
  const mockFn = showAlert as jest.Mock<typeof showAlert, Parameters<typeof showAlert>>;
  const { title, okLabel, addCancel, message, onConfirm } = mockFn.mock.calls[0][0];
  mockFn.mockClear();
  return {
    title,
    okLabel,
    addCancel,
    message,
    onConfirm,
  };
};

export const getAndClearShowErrorMockArgs = () => {
  const mockFn = showError as jest.Mock<typeof showError, Parameters<typeof showError>>;
  const options = mockFn.mock.calls[0][0];
  mockFn.mockClear();
  return options;
};

export const getAndClearShowModalMockArgs = () => {
  const mockFn = showModal as jest.Mock<typeof showModal, Parameters<typeof showModal>>;
  const args = mockFn.mock.calls[0][1];
  mockFn.mockClear();
  return args;
};
```

I suggest we:
1. drop the redundant function return type
2. give a generic to jest.Mock (a shame it doesn't use `Parameters` internally... hmm... maybe we should submit a PR and find out why or if it can?)
3. then remove the unnecessary cast e.g. `as PromptModalOptions`.

packages/insomnia-app/app/ui/components/request-url-bar.tsx
develohpanda: We can, didn't relate to the scope of this PR

dimitropoulos (approved):
while I'd like to see the changes I suggested, I can't say they're blocking.  please feel free to merge after you implement them (or not - it's cool either way).

I especially like the 2nd "are you sure" modal.

packages/insomnia-app/app/ui/components/request-url-bar.tsx
dimitropoulos: Well.. I guess my perspective is that if it causes us to need add new error like this then it's as "in scope" as it will ever be.  My thought process is that we should be always reducing the number of these errors, not adding.  If you're saying you wanna do in an immediate follow-up PR, I can get down with that.

develohpanda: > reducing the number of these errors, not adding

This PR is not adding new errors, only surfacing ones that already existed because of improved typing. We're not in a worse place than we were before.

Weighing the pros and cons of increasing the dev, review and QA that will need to happen outside of the planned work in this PR by fixing this particular error (which is non-trivial), I opted to mark as an error and resolve at a later date

packages/insomnia-app/app/test-utils.ts
develohpanda: Will do in subsequent PR, much nicer approach using the jest.Mock generic

dimitropoulos: https://github.com/Kong/insomnia/pull/3435/files

PR #3436, opened 2021-06-02
Fetch and store remote spaces

develohpanda:
This follows the same process as the "Pull" drop-down - it will load remote spaces each time it is opened or the VCS object is updated.

Remote spaces and local spaces currently show together, but settings are not available for remote spaces.


develohpanda:
This was merged accidentally - re-opened [here](https://github.com/Kong/insomnia/pull/3441)

PR #3438, opened 2021-06-02
fix DebouncedInput (and delay handling) and change handlers

dimitropoulos:
During the review in a prior PR, a bug in DebouncedInput was discovered, namely that if the delay (which is a prop) changes, the component behavior will not change to debounce (or not) since the behavior is set only in the constructor.

Along the way, there were a few other low-hanging fruit to address, so I did those in a limited capacity.

### DIFF
--- a/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
+++ b/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
@@ -22,24 +22,22 @@ interface State {
   debouncedOnChange: ExtendedAttributes['onChange'];
 }
 
-const defaultDelay = 500;
-
 @autoBindMethodsForReact(AUTOBIND_CFG)
 export class DebouncedInput extends PureComponent<Props, State> {
   _hasFocus = false;
   _input: HTMLTextAreaElement | HTMLInputElement | null = null;
 
   state: State = {
-    delay: defaultDelay,
+    delay: this.props.delay ?? 500,
     onChange: this.props.onChange,
-    debouncedOnChange: debounce(this.props.onChange, this.props.delay || defaultDelay),
+    debouncedOnChange: debounce(this.props.onChange, this.props.delay),
   }
 
   static getDerivedStateFromProps(props: Props, state: State) {
     if (state.delay !== props.delay || state.onChange !== props.onChange) {
       return {
         onChange: props.onChange,
-        debouncedOnChange: debounce(props.onChange, props.delay || defaultDelay),
+        debouncedOnChange: debounce(props.onChange, props.delay),
         delay: props.delay,
       };
     }
--- a/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
+++ b/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
@@ -13,7 +13,7 @@ type InheritedAttributes = Omit<InputHTMLAttributes<TargetElement>, keyof Extend
 
 interface Props extends InheritedAttributes, ExtendedAttributes {
   textarea?: boolean;
-  delay?: number;
+  delay: number;
 }
 
 interface State {
@@ -27,14 +27,14 @@ export class DebouncedInput extends PureComponent<Props, State> {
   _input: HTMLTextAreaElement | HTMLInputElement | null = null;
 
   state: State = {
-    delay: this.props.delay ?? 500,
-    onChange: this.props.delay ? debounce(this.props.onChange, this.props.delay) : this.props.onChange,
+    delay: this.props.delay,
+    onChange: debounce(this.props.onChange, this.props.delay),
   }
 
   static getDerivedStateFromProps(props: Props, state: State) {
     if (state.delay !== props.delay || state.onChange !== props.onChange) {
       return {
-        onChange: props.delay ? debounce(props.onChange, props.delay) : props.onChange,
+        onChange: debounce(props.onChange, props.delay),
         delay: props.delay,
       };
     }
--- a/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
+++ b/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
@@ -16,18 +16,44 @@ interface Props extends InheritedAttributes, ExtendedAttributes {
   delay?: number;
 }
 
+interface State {
+  delay: number;
+  onChange: ExtendedAttributes['onChange'];
+  debouncedOnChange: ExtendedAttributes['onChange'];
+}
+
+const defaultDelay = 500;
+
 @autoBindMethodsForReact(AUTOBIND_CFG)
-export class DebouncedInput extends PureComponent<Props> {
+export class DebouncedInput extends PureComponent<Props, State> {
   _hasFocus = false;
   _input: HTMLTextAreaElement | HTMLInputElement | null = null;
 
+  state: State = {
+    delay: defaultDelay,
+    onChange: this.props.onChange,
+    debouncedOnChange: debounce(this.props.onChange, this.props.delay || defaultDelay),
+  }
+
+  static getDerivedStateFromProps(props: Props, state: State) {
+    if (state.delay !== props.delay || state.onChange !== props.onChange) {
+      return {
+        onChange: props.onChange,
+        debouncedOnChange: debounce(props.onChange, props.delay || defaultDelay),
+        delay: props.delay,
+      };
+    }
+    return null;
+  }
+
   _handleChange(event: ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) {
-    const { delay, onChange } = this.props;
+    const { delay } = this.props;
+    const { onChange, debouncedOnChange } = this.state;
     const { value } = event.target;
     if (!delay) {
       onChange(value);
     } else {
-      debounce(onChange, delay || 500)(value);
+      debouncedOnChange(value);
     }
   }
 
--- a/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
+++ b/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react';
+import React, { ChangeEvent, PureComponent } from 'react';
 import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import { AUTOBIND_CFG } from '../../../common/constants';
 import { debounce } from '../../../common/misc';
@@ -15,20 +15,15 @@ interface Props {
 export class DebouncedInput extends PureComponent<Props> {
   _hasFocus = false;
   _input: HTMLTextAreaElement | HTMLInputElement | null = null;
-  _handleValueChange: Props['onChange'] | null = null;
 
-  constructor(props: Props) {
-    super(props);
-
-    if (!props.delay) {
-      this._handleValueChange = props.onChange;
+  _handleChange(event: ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) {
+    const { delay, onChange } = this.props;
+    const { value } = event.target;
+    if (!delay) {
+      onChange(value);
     } else {
-      this._handleValueChange = debounce(props.onChange, props.delay || 500);
+      debounce(onChange, delay || 500)(value);
     }
-  }
-
-  _handleChange(e: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) {
-    this._handleValueChange?.(e.target.value);
   }
 
   _handleFocus(e: React.FocusEvent<HTMLTextAreaElement | HTMLInputElement>) {
--- a/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
+++ b/packages/insomnia-app/app/ui/components/base/debounced-input.tsx
@@ -19,7 +19,6 @@ interface Props extends InheritedAttributes, ExtendedAttributes {
 interface State {
   delay: number;
   onChange: ExtendedAttributes['onChange'];
-  debouncedOnChange: ExtendedAttributes['onChange'];
 }
 
 @autoBindMethodsForReact(AUTOBIND_CFG)
@@ -29,15 +28,13 @@ export class DebouncedInput extends PureComponent<Props, State> {
 
   state: State = {
     delay: this.props.delay ?? 500,
-    onChange: this.props.onChange,
-    debouncedOnChange: debounce(this.props.onChange, this.props.delay),
+    onChange: this.props.delay ? debounce(this.props.onChange, this.props.delay) : this.props.onChange,
   }
 
   static getDerivedStateFromProps(props: Props, state: State) {
     if (state.delay !== props.delay || state.onChange !== props.onChange) {
       return {
-        onChange: props.onChange,
-        debouncedOnChange: debounce(props.onChange, props.delay),
+        onChange: props.delay ? debounce(props.onChange, props.delay) : props.onChange,
         delay: props.delay,
       };
     }
@@ -45,14 +42,9 @@ export class DebouncedInput extends PureComponent<Props, State> {
   }
 
   _handleChange(event: ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) {
-    const { delay } = this.props;
-    const { onChange, debouncedOnChange } = this.state;
+    const { onChange } = this.state;
     const { value } = event.target;
-    if (!delay) {
-      onChange(value);
-    } else {
-      debouncedOnChange(value);
-    }
+    onChange(value);
   }
 
   _handleFocus(e: React.FocusEvent<HTMLTextAreaElement | HTMLInputElement>) {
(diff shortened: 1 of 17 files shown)


packages/insomnia-app/app/ui/components/base/debounced-input.tsx
```diff
@@ -15,22 +15,17 @@ interface Props {
...
-    if (!props.delay) {
-      this._handleValueChange = props.onChange;
+  _handleChange(event: ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) {
+    const { delay, onChange } = this.props;
+    const { value } = event.target;
+    if (!delay) {
+      onChange(value);
     } else {
-      this._handleValueChange = debounce(props.onChange, props.delay || 500);
+      debounce(onChange, delay || 500)(value);
```
develohpanda: This change doesn't actually work. It now generates a _new_ debounced handler with every keystroke, removing the memoization we get with the previous implementation.

Yes initializing the handler in the constructor isn't right either, in case the `onChange` or `delay` props are changed

Develop:


This PR:

develohpanda (changes requested):

packages/insomnia-app/app/ui/components/base/debounced-input.tsx
dimitropoulos: ok, all fixed now.

<details>
<summary>expand to see a version of `space-settings-modal.tsx` you can test with</summary>

```tsx
import React, { PureComponent } from 'react';
import { autoBindMethodsForReact } from 'class-autobind-decorator';
import { AUTOBIND_CFG } from '../../../common/constants';
import { DebouncedInput } from '../base/debounced-input';
import Modal from '../base/modal';
import ModalBody from '../base/modal-body';
import ModalHeader from '../base/modal-header';
import PromptButton from '../base/prompt-button';
import * as models from '../../../models/index';
import { Space } from '../../../models/space';
import { selectActiveSpace } from '../../redux/selectors';
import { bindActionCreators } from 'redux';
import * as spaceActions from '../../redux/modules/space';
import { connect } from 'react-redux';

interface ReduxStateProps {
  space?: Space;
}

interface ReduxDispatchProps {
  handleRemoveSpace: (space: Space) => void;
}

interface State {
  delay: number;
}

interface Props extends ReduxStateProps, ReduxDispatchProps { }

class SpaceSettingsModal extends PureComponent<Props, State> {
  modal: Modal | null = null;

  state: State = {
    delay: 500,
  }

  constructor(props: Props) {
    super(props);
    setTimeout(() => {
      console.log('changing the delay to 2 seconds');
      this.setState({ delay: 2000 });
    }, 10_000);

    setTimeout(() => {
      console.log('changing the onChange and the delay to 3 seconds');
      this._handleRename = async () => { console.log('new onChange (does nothing)'); };
      this.setState({ delay: 3000 });
    }, 20_000);
  }

  _handleSetModalRef(n: Modal) {
    this.modal = n;
  }

  _handleRemoveSpace() {
    if (this.props.space) {
      this.props.handleRemoveSpace(this.props.space);
    }

    this.hide();
  }

  async _handleRename(name: string) {
    const { space } = this.props;

    if (space) {
      await models.space.update(space, { name });
    }
  }

  show() {
    this.modal && this.modal.show();
  }

  hide() {
    this.modal && this.modal.hide();
  }

  render() {
    const { space } = this.props;
    if (!space) {
      return null;
    }

    console.log('rendering space-settings-modal');
    return (
      <Modal ref={this._handleSetModalRef} freshState>
        <ModalHeader key={`header::${space._id}`}>
          {strings.space.singular} Settings{' '}
          <div className="txt-sm selectable faint monospace">{space._id}</div>
        </ModalHeader>
        <ModalBody key={`body::${space._id}`} className="pad">
            <label>
              Name
              <DebouncedInput
                type="text"
                delay={this.state.delay}
                placeholder="My Space"
                defaultValue={space.name}
                onChange={this._handleRename}
              />
            </label>
          </div>
          <h2>Actions</h2>
          <div className="form-control form-control--padded">
            <PromptButton
              onClick={this._handleRemoveSpace}
              addIcon
              className="width-auto btn btn--clicky inline-block">
              <i className="fa fa-trash-o" /> Delete
            </PromptButton>
          </div>
        </ModalBody>
      </Modal>
    );
  }
}

const mapStateToProps = (state): ReduxStateProps => {
  const space = selectActiveSpace(state);
  return { space };
};

const mapDispatchToProps = (dispatch): ReduxDispatchProps => {
  const boundSpaceActions = bindActionCreators(spaceActions, dispatch);
  return {
    handleRemoveSpace: boundSpaceActions.removeSpace,
  };
};

export default connect(mapStateToProps, mapDispatchToProps, null, { forwardRef: true })(SpaceSettingsModal);
```

</details>

I guess, upon really looking at it, it ALSO didn't respond to changes to the change handler (the `onChange` prop).  That's pretty bad, actually, come to think of it.  That's also fixed now. d11a54ab910bd578a3bbf2ccdedb4b9ce87137f3

While on the one hand I realllly don't like having to use derived state, this really is the (one?) use-case where it's required, where we actually need to know the previous state since we're calling a function (debounce) that spans multiple state transitions.

packages/insomnia-app/app/ui/components/base/debounced-input.tsx
```diff
@@ -1,34 +1,60 @@
...
-    this._handleValueChange?.(e.target.value);
+  _handleChange(event: ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) {
+    const { delay } = this.props;
+    const { onChange, debouncedOnChange } = this.state;
+    const { value } = event.target;
+    if (!delay) {
+      onChange(value);
+    } else {
+      debouncedOnChange(value);
+    }
```
develohpanda: I think we can simplify this by having a single `onChange` property in state, which gets set to `props.onChange` or `debounce(props.onChange, ...)` depending on `props.delay` being defined.

As an aside, the `defaultDelay` fallback seems redundant since the debounced handler will never be reached if `props.delay` is not set 🤔

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -80,19 +80,29 @@ const BASE_CODEMIRROR_OPTIONS: CodeMirror.EditorConfiguration = {
...
+  | 'style'
+>
+
+interface ExtendedAttributes {
+  onKeyDown?: (event: React.KeyboardEvent, value?: string) => void;
+  onChange?: (value?: string) => void;
+}
 
-interface Props {
+interface Props extends InheritedAttributes, ExtendedAttributes {
```
develohpanda, line 101: This looks like a nice approach for extending and overriding but inheriting types correctly types

packages/insomnia-app/app/ui/components/codemirror/one-line-editor.tsx
```diff
@@ -1,46 +1,51 @@
...
   getRenderContext?: Function;
   nunjucksPowerUserMode?: boolean;
   getAutocompleteConstants?: Function | null;
-  placeholder?: string;
-  className?: string;
   forceEditor?: boolean;
   forceInput?: boolean;
   isVariableUncovered?: boolean;
   // TODO(TSCONVERSION) figure out why so many components pass this in yet it isn't used anywhere in this
   disabled?: boolean;
```
develohpanda, line 40: Could it be that this is also an inherited attribute?

packages/insomnia-app/app/ui/components/editors/body/form-editor.tsx
```diff
@@ -3,9 +3,10 @@ import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import { AUTOBIND_CFG } from '../../../../common/constants';
 import KeyValueEditor from '../../key-value-editor/editor';
 import { HandleGetRenderContext, HandleRender } from '../../../../common/render';
+import { RequestBodyParameter } from '../../../../models/request';
 
 interface Props {
-  onChange: Function,
+  onChange: (parameters: RequestBodyParameter[]) => void,
   parameters: any[],
```
develohpanda: Should this also be `RequestBodyParameter[]`?

packages/insomnia-app/app/ui/components/editors/body/url-encoded-editor.tsx
```diff
@@ -3,9 +3,10 @@ import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import { AUTOBIND_CFG } from '../../../../common/constants';
 import KeyValueEditor from '../../key-value-editor/editor';
 import { HandleGetRenderContext, HandleRender } from '../../../../common/render';
+import { RequestBodyParameter } from '../../../../models/request';
 
 interface Props {
-  onChange: Function;
+  onChange: (parameters: RequestBodyParameter[]) => void;
   parameters: any[];
```
develohpanda: Should this also be `RequestBodyParameter[]`?

packages/insomnia-app/app/ui/components/key-value-editor/editor.tsx
```diff
@@ -20,7 +21,7 @@ const LEFT = 37;
 const RIGHT = 39;
 
 interface Props {
-  onChange: Function;
+  onChange: (headers: RequestHeader[]) => void;
   pairs: any[];
```
develohpanda: Should this also be `RequestHeader[]`?

packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
```diff
@@ -73,7 +73,6 @@ class SpaceSettingsModal extends PureComponent<Props> {
             <label>
               Name
               <DebouncedInput
-              // @ts-expect-error -- TSCONVERSION props are spread into an input element
```
develohpanda, line 76: So glad to have this fixed

packages/insomnia-app/app/ui/components/key-value-editor/row.tsx
```diff
@@ -23,20 +33,11 @@ interface Props {
...
-  pair: {
-    id: string,
-    name: string,
-    value: string,
-    description: string,
-    fileName: string,
-    type: string,
-    disabled: boolean,
-  },
-  readOnly?: boolean,
```
develohpanda, line 35: I am assuming there is nothing that sends the `readOnly` prop?

packages/insomnia-app/app/ui/components/codemirror/one-line-editor.tsx
```diff
@@ -1,46 +1,51 @@
...
-  id?: string;
-  type?: string;
   mode?: string;
-  onBlur?: Function;
-  onKeyDown?: Function;
-  onFocus?: Function;
-  onChange?: Function;
-  onPaste?: Function;
   render?: Function;
   getRenderContext?: Function;
```
develohpanda: See types here

https://github.com/Kong/insomnia/blob/3998e8b48969fdf0a668435c8b017384c5cc38a0/packages/insomnia-app/app/ui/components/editors/body/body-editor.tsx#L41-L42

develohpanda (commented):
QA'd locally and looks to be working! A couple of comments so not approving yet 🙌🏽

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
dimitropoulos: :)

packages/insomnia-app/app/ui/components/key-value-editor/editor.tsx
dimitropoulos: good call: 86f51efd059309fdedb4bd8992bc5cf411eec53d

packages/insomnia-app/app/ui/components/editors/body/url-encoded-editor.tsx
dimitropoulos: good call: 86f51efd059309fdedb4bd8992bc5cf411eec53d

packages/insomnia-app/app/ui/components/editors/body/form-editor.tsx
dimitropoulos: good call: 86f51efd059309fdedb4bd8992bc5cf411eec53d

packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
dimitropoulos: it's a great feeling, isn't it :)

packages/insomnia-app/app/ui/components/key-value-editor/row.tsx
dimitropoulos: no, opposite, things were sending it but this component isn't/wasn't using it.

I did the git spelunking so you could double check my work, but the TLDR; is `readOnly` was never actually used but was allowed before when types weren't checked, but is not now. 

Actually, it _was_ once upon a time using it, but [`KeyValueEditorRow` was just passing it to `OneLineEditor`](https://github.com/Kong/insomnia/blob/bdb4b4e6610f69e84a0ae7ea1296078c59bbe88e/packages/insomnia-app/app/ui/components/key-value-editor/row.js#L219), which [wasn't using it](https://github.com/Kong/insomnia/blob/bdb4b4e6610f69e84a0ae7ea1296078c59bbe88e/packages/insomnia-app/app/ui/components/codemirror/one-line-editor.js).

So far as I can tell, this has been wrong all along.  At [the time this was added](https://github.com/Kong/insomnia/commit/a25b62b994b283bffb6ba0a947a9b96332e6d2f1#diff-d2a68d459e2d7ba43df733366da4c7db204b7759746b037e0aebae4ae9082d1fR218) if you look at [this component at that time](https://github.com/Kong/insomnia/blob/a25b62b994b283bffb6ba0a947a9b96332e6d2f1/packages/insomnia-app/app/ui/components/codemirror/one-line-editor.js) you should see that it wasn't used then, either.

dimitropoulos: Perhaps the confusion arose because CodeEditor (but not OneLineEditor) actually uses `readOnly`: https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx#L1140

packages/insomnia-app/app/ui/components/codemirror/one-line-editor.tsx
dimitropoulos: Ahh, cool.  I was wondering how I missed this because I know I did a gloabl search before.  Turns out some are prefixed with `handle` and some are not.  Fun stuff.



It'd be cool if they were consistent although this PR already does plenty - I don't want to expand the scope.  Next time.

dimitropoulos: fixed in 26ec600afe1117bc12a264acd25f70a55a84fe12

packages/insomnia-app/app/ui/components/base/debounced-input.tsx
dimitropoulos: I removed the default delay in c63036d7f1b784fcf7d3de52c043dda1baf0aabc but I'd like to ask you to reconsider re: combining the onChange.

The reason is that it will make the code (and debugging) more complicated since you never know at any given moment which one you're working with.  I implemented what I believe you're asking for in 914474d7213f5dc7ac6f7580a1ccfb8510ab20c1.  I agree that the code is more tight, but I think it comes at the cost of readability.

HOWEVER, I was thinking we can actually make this a lot better by simply requiring that there's always a delay and that it's always debounced.  I'm not sure if that matters, but it seems like if you didn't want debouncing you shouldn't use DebouncedInput.  From that perspective, maybe it makes sense to just remove the code path that uses the callback _without_ a debounce (like the component now supports).  Thoughts?

The only place I see it used without a delay is in OneLineEditor, and still there it's not clear to me if it's just written that way with the knowledge that the delay is implicitly there as a fallback.  I implemented this just as a proof-of-concept here dceb8ab951bb823007df7942d64c0cfb848b9487.

dimitropoulos: OHHH I think I remember a reason they need to be separate.  Doesn't that mean that otherwise there's no way to compare if the prop changed.  If we always set it to something new, it would create an infinite loop since `state.onChange !== props.onChange` will always evaluate to true, right?

Screenshot_20210630_124518

packages/insomnia-app/app/ui/components/codemirror/one-line-editor.tsx
dimitropoulos: that's just the thing, though.  I don't see anywhere where this is actually passed or spread into a child component.

Can you show me a behavior of the app that is different when `disabled` is true or false, as passed as a prop to `OneLineEditor`?  And if so, what code enables the passing of the value of this prop to an element in the DOM?

jackkav:
Closed as stale.

PR #3441, opened 2021-06-03
Fetch and store remote spaces

develohpanda:
This follows the same process as the "Pull" drop-down - it will load remote spaces each time it is opened or the VCS object is updated.

Remote spaces and local spaces currently show together, but settings are not available for remote spaces.

### DIFF
--- a/packages/insomnia-app/app/common/database.ts
+++ b/packages/insomnia-app/app/common/database.ts
@@ -50,14 +50,16 @@ export const database = {
     const promisesUpserted: Promise<BaseModel>[] = [];
     const promisesDeleted: Promise<void>[] = [];
 
-    // @ts-expect-error -- TSCONVERSION upsert operations are optional
-    for (const doc of operation.upsert) {
-      promisesUpserted.push(database.upsert(doc, true));
+    if (operation.upsert !== undefined) {
+      for (const doc of operation.upsert) {
+        promisesUpserted.push(database.upsert(doc, true));
+      }
     }
 
-    // @ts-expect-error -- TSCONVERSION remove operations are optional
-    for (const doc of operation.remove) {
-      promisesDeleted.push(database.unsafeRemove(doc, true));
+    if (operation.remove !== undefined) {
+      for (const doc of operation.remove) {
+        promisesDeleted.push(database.unsafeRemove(doc, true));
+      }
     }
 
     // Perform from least to most dangerous
--- a/packages/insomnia-app/app/common/database.ts
+++ b/packages/insomnia-app/app/common/database.ts
@@ -44,27 +44,13 @@ export const database = {
     return database.find<T>(type);
   },
 
-  batchModifyDocs: async function(operation: Operation) {
+  batchModifyDocs: async function({ upsert = [], remove = [] }: Operation) {
     if (db._empty) return _send<void>('batchModifyDocs', ...arguments);
     const flushId = await database.bufferChanges();
-    const promisesUpserted: Promise<BaseModel>[] = [];
-    const promisesDeleted: Promise<void>[] = [];
-
-    if (operation.upsert !== undefined) {
-      for (const doc of operation.upsert) {
-        promisesUpserted.push(database.upsert(doc, true));
-      }
-    }
-
-    if (operation.remove !== undefined) {
-      for (const doc of operation.remove) {
-        promisesDeleted.push(database.unsafeRemove(doc, true));
-      }
-    }
 
     // Perform from least to most dangerous
-    await Promise.all(promisesUpserted);
-    await Promise.all(promisesDeleted);
+    await Promise.all(upsert.map(doc => database.upsert(doc, true)));
+    await Promise.all(remove.map(doc => database.unsafeRemove(doc, true)));
 
     await database.flushChanges(flushId);
   },
--- a/packages/insomnia-app/app/models/space.ts
+++ b/packages/insomnia-app/app/models/space.ts
@@ -22,6 +22,7 @@ export type Space = BaseModel & BaseSpace;
 export function init(): BaseSpace {
   return {
     name: 'My Space',
+    remoteId: undefined, // this is necessary for the model init logic to work properly
   };
 }
 
--- a/packages/insomnia-app/app/models/space.ts
+++ b/packages/insomnia-app/app/models/space.ts
@@ -14,6 +14,7 @@ export const BASE_SPACE_ID = 'base-space';
 
 interface BaseSpace {
   name: string;
+  remoteId?: string;
 }
 
 export type Space = BaseModel & BaseSpace;
--- a/packages/insomnia-app/app/models/space.ts
+++ b/packages/insomnia-app/app/models/space.ts
@@ -14,7 +14,7 @@ export const BASE_SPACE_ID = 'base-space';
 
 interface BaseSpace {
   name: string;
-  remoteId?: string;
+  remoteId: string | null;
 }
 
 export type Space = BaseModel & BaseSpace;
@@ -22,7 +22,7 @@ export type Space = BaseModel & BaseSpace;
 export function init(): BaseSpace {
   return {
     name: 'My Space',
-    remoteId: undefined, // this is necessary for the model init logic to work properly
+    remoteId: null, // `null` is necessary for the model init logic to work properly
   };
 }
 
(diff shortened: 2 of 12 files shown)


packages/insomnia-app/app/common/database.ts
```diff
@@ -50,14 +50,16 @@ export const database = {
     const promisesUpserted: Promise<BaseModel>[] = [];
     const promisesDeleted: Promise<void>[] = [];
 
-    // @ts-expect-error -- TSCONVERSION upsert operations are optional
-    for (const doc of operation.upsert) {
-      promisesUpserted.push(database.upsert(doc, true));
+    if (operation.upsert !== undefined) {
+      for (const doc of operation.upsert) {
+        promisesUpserted.push(database.upsert(doc, true));
```
dimitropoulos: perhaps more idiomatic if we do:

```ts
operation.upsert?.foreach(doc => {
  promisesUpserted.push(database.upsert(doc, true);
});
```

dimitropoulos: (same for the remove below)

packages/insomnia-app/app/sync/store/drivers/base.ts
```diff
@@ -1,5 +1,4 @@
 export interface BaseDriver {
-  new (config: Record<string, any>): void;
```
dimitropoulos, line 2: good, I was wondering about this.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -4,34 +4,56 @@ import { useDispatch, useSelector } from 'react-redux';
...
+import { useRemoteSpaces } from '../../hooks/space';
 import { setActiveSpace } from '../../redux/modules/global';
 import { createSpace } from '../../redux/modules/space';
 import { selectActiveSpace, selectSpaces } from '../../redux/selectors';
 import { showModal } from '../modals';
 import SpaceSettingsModal from '../modals/space-settings-modal';
 
-const mapSpace = ({ _id, name }: Space) => ({ id: _id, name });
-const defaultSpace = { id: BASE_SPACE_ID, name: getAppName() };
+type SpaceSubset = Pick<Space, '_id'> & Pick<Space, 'name'> & Pick<Space, 'remoteId'>;
```
dimitropoulos:
```suggestion
type SpaceSubset = Pick<Space, '_id' | 'name' | 'remoteId'>;
```

packages/insomnia-app/app/common/database.ts
dimitropoulos: actually, on second blush, we could make this all much simpler by removing the intermediate datastructures:

```ts
  batchModifyDocs: async function({ upsert = [], remove = [] }: Operation) {
    if (db._empty) return _send<void>('batchModifyDocs', ...arguments);
    const flushId = await database.bufferChanges();

    // Perform from least to most dangerous
    await Promise.all(upsert.map(doc => database.upsert(doc, true)));
    await Promise.all(remove.map(doc => database.unsafeRemove(doc, true)));

    await database.flushChanges(flushId);
  }
  ```

Comment (dimitropoulos) [reply packages/insomnia-app/app/common/database.ts]: (random side-note, it's too bad we don't have these with a data-last (auto-curried) API, because then we can use it more easily, e.g.

```diff
- upsert.map(doc => database.upsert(doc, true))
+ upsert.map(database.upsert(true))
```
)

packages/insomnia-app/app/models/space.ts
```diff
@@ -14,13 +14,15 @@ export const BASE_SPACE_ID = 'base-space';
...
   name: string;
+  remoteId?: string;
 }
 
 export type Space = BaseModel & BaseSpace;
 
 export function init(): BaseSpace {
   return {
     name: 'My Space',
+    remoteId: undefined, // this is necessary for the model init logic to work properly
```
dimitropoulos:
```suggestion
  remoteId: string | undefined;
}

export type Space = BaseModel & BaseSpace;

export function init(): BaseSpace {
  return {
    name: 'My Space',
    remoteId: undefined, // this is necessary for the model init logic to work properly
```

The next version of TypeScript (4.4) will have (https://github.com/microsoft/TypeScript/pull/43947) some logic that distinguishes between these two things.  Let's try to stay accurate about it now so it'll be easier to transition when the time comes (soon).

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: I implemented this in https://github.com/dimitropoulos/insomnia/commit/261c28f086b22131e97356f585c19256698f90ae if you wanna just pull it in

packages/insomnia-app/app/common/database.ts
dimitropoulos: I implemented this in https://github.com/dimitropoulos/insomnia/commit/1158baab11b6a746a364f9458afaec57921715f3 if you wanna just pull it in

packages/insomnia-app/app/models/space.ts
dimitropoulos: I think we should consider using `null` for this since it is more specific and will serialize when we import/export:

see: https://github.com/dimitropoulos/insomnia/commit/ede3ca97d82e4254d2599edb7a428099cbe68fd6

packages/insomnia-app/app/ui/hooks/space.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Space } from '../../models/space';
+
+export const useRemoteSpaces = (vcs?: VCS) => {
+  const [loading, setLoading] = useState(false);
+
+  const refresh = useCallback(async () => {
+    if (vcs && isLoggedIn()) {
+      setLoading(true);
+
+      const teams = await vcs.teams();
```
DMarby, line 16: What will happen if an exception is thrown here, eg if the session has expired?

packages/insomnia-app/app/ui/hooks/space.ts
```diff
@@ -0,0 +1,32 @@
...
+      const spaces = await Promise.all(teams.map(team => models.initModel<Space>(
+        models.space.type,
+        { _id: `${models.space.prefix}_${team.id}`, remoteId: team.id, name: team.name },
+      )));
+      await database.batchModifyDocs({ upsert: spaces });
```
DMarby, line 25: Not a blocker for this PR, but I guess we're going to want to figure out how to handle the user logging out fairly soon, eg whether we should still show remote spaces or not in the dropdown in that scenario, as well as what happens when switching between accounts.

DMarby (approved):

packages/insomnia-app/app/ui/hooks/space.ts
develohpanda: And this is why we have code reviews 💯 I'll add some error handling in there which aligns with how the other sync code handles errors

packages/insomnia-app/app/ui/hooks/space.ts
develohpanda: Yah. For this particular one I thought about using the "Pull" drop-down model where you click to download/clone, but it didn't feel worthwhile. Because we don't allow the deletion of remote spaces, currently you can't _delete_ these remote spaces that get created in the database. 

I'll show you how a delete behaves in the app for remote spaces, so we can potentially address it. I'll track it in a Linear ticket.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
develohpanda: I was trying to use `Pick<space, ['id', 'name']>` but couldn't get it sorted. This is great!

packages/insomnia-app/app/models/space.ts
develohpanda: Nice catch 👍🏽

dimitropoulos (approved):

packages/insomnia-app/app/ui/hooks/space.ts
develohpanda: Maybe no error handling for the time being, the pull drop-down behaves the same way (continued loading spinner and error in console) when the session id is invalid

image

develohpanda: INS-757

packages/insomnia-app/app/ui/hooks/space.ts
develohpanda: INS-758

PR #3446, opened 2021-06-04
Create a workspace in the active space and filter dashboard for active space

develohpanda:
Fixes INS-748 and Fixes INS-749

### DIFF
--- a/packages/insomnia-app/app/models/workspace.ts
+++ b/packages/insomnia-app/app/models/workspace.ts
@@ -59,7 +59,7 @@ export async function create(patch: Partial<Workspace> = {}) {
   return db.docCreate<Workspace>(type, patch);
 }
 
-export async function all() {
+export async function all(): Promise<Workspace[]> {
   const workspaces = await db.all<Workspace>(type) || [];
 
   if (workspaces.length === 0) {
--- a/packages/insomnia-app/app/models/workspace.ts
+++ b/packages/insomnia-app/app/models/workspace.ts
@@ -7,13 +7,9 @@ import { ValueOf } from 'type-fest';
 import { isSpaceId } from './helpers/is-model';
 
 export const name = 'Workspace';
-
 export const type = 'Workspace';
-
 export const prefix = 'wrk';
-
 export const canDuplicate = true;
-
 export const canSync = true;
 
 export const WorkspaceScopeKeys = {
@@ -59,19 +55,19 @@ export async function create(patch: Partial<Workspace> = {}) {
   return db.docCreate<Workspace>(type, patch);
 }
 
-export async function all(): Promise<Workspace[]> {
-  const workspaces = await db.all<Workspace>(type) || [];
+export async function all() {
+  const workspaces = await db.all<Workspace>(type);
 
-  if (workspaces.length === 0) {
-    // Create default workspace
-    await create({
-      name: getAppName(),
-      scope: WorkspaceScopeKeys.collection,
-    });
-    return all();
-  } else {
+  if (workspaces.length > 0) {
     return workspaces;
   }
+
+  // Create default workspace
+  await create({
+    name: getAppName(),
+    scope: WorkspaceScopeKeys.collection,
+  });
+  return db.all<Workspace>(type);
 }
 
 export function count() {
--- a/packages/insomnia-app/app/ui/components/wrapper-home.tsx
+++ b/packages/insomnia-app/app/ui/components/wrapper-home.tsx
@@ -48,8 +48,8 @@ import { strings } from '../../common/strings';
 import { descendingNumberSort } from '../../common/sorting';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import * as workspaceActions from '../redux/modules/workspace';
-import * as gitActions from '../redux/modules/git';
+import { createWorkspace } from '../redux/modules/workspace';
+import { cloneGitRepository } from '../redux/modules/git';
 import { MemClient } from '../../sync/git/mem-client';
 import { SpaceDropdown } from './dropdowns/space-dropdown';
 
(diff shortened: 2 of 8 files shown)


packages/insomnia-app/app/models/workspace.ts
```diff
@@ -59,7 +59,7 @@ export async function create(patch: Partial<Workspace> = {}) {
   return db.docCreate<Workspace>(type, patch);
 }
 
-export async function all() {
+export async function all(): Promise<Workspace[]> {
```
dimitropoulos: I see.. hey, what would you think about doing something like this?

```ts

export async function all() {
  const workspaces = await db.all<Workspace>(type);
  if (workspaces.length > 0) {
    return workspaces;
  }

  // Create default workspace
  await create({
    name: getAppName(),
    scope: WorkspaceScopeKeys.collection,
  });
  return db.all<Workspace>(type);
}
```

That way we drop the recursion.

dimitropoulos: (also, from what I could tell the default to `[]` isn't necessary, but double check that if you go with this exact implementation)

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -58,13 +58,16 @@ interface RenderedCard {
   lastModifiedTimestamp?: number | null;
 }
 
-interface Props {
+interface ReduxDispatchProps {
```
dimitropoulos, line 61: _NOICE_

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1596,12 +1597,13 @@ function mapStateToProps(state, props) {
...
     workspaceMetas,
-    // @ts-expect-error -- TSCONVERSION
-    workspaces,
   } = entitiesLists;
   // @ts-expect-error -- TSCONVERSION
   const settings = entitiesLists.settings[0];
   // Workspace stuff
+
+  // @ts-expect-error -- TSCONVERSION https://github.com/reduxjs/reselect#accessing-react-props-in-selectors
+  const workspaces = selectWorkspacesForActiveSpace(state, props);
```
dimitropoulos, line 1606: good stuff, good stuff.

packages/insomnia-app/app/ui/redux/modules/__tests__/workspace.test.ts
```diff
@@ -15,11 +15,17 @@ jest.mock('../../../../common/analytics');
 const middlewares = [thunk];
 const mockStore = configureMockStore(middlewares);
 
+const createSpace = async () => await models.initModel(models.space.type);
```
dimitropoulos:
```suggestion
const createSpace = () => models.initModel(models.space.type);
```

packages/insomnia-app/app/ui/redux/modules/__tests__/workspace.test.ts
```diff
@@ -15,11 +15,17 @@ jest.mock('../../../../common/analytics');
...
+
 describe('workspace', () => {
   beforeEach(globalBeforeEach);
   describe('createWorkspace', () => {
     it('should create document', async () => {
-      const store = mockStore();
+      const space = await createSpace();
+      const entities = { spaces: { [space._id]: space } };
+      const global = { activeSpaceId: space._id };
+      const store = mockStore({ entities, global });
```
dimitropoulos: maybe we move these three lines into `createSpace`? it seems to be parametric with the other usage of `createSpace`

packages/insomnia-app/app/ui/redux/modules/__tests__/workspace.test.ts
```diff
@@ -51,7 +58,47 @@ describe('workspace', () => {
...
+        createWorkspace({
+          scope: WorkspaceScopeKeys.collection,
+        }),
+      );
+      const { title, submitName, defaultValue, onComplete } = getAndClearShowPromptMockArgs();
+      expect(title).toBe('Create New Request Collection');
+      expect(submitName).toBe('Create');
+      expect(defaultValue).toBe('My Collection');
+      const workspaceName = 'name';
+      await onComplete(workspaceName);
```
dimitropoulos: Screenshot_20210608_160524

Are you seeing errors in this file in these new lines?  e.g. this one is awaited but it doesn't return a promise (does it?) and seems to (from the usages of showPrompt I looked at) genuinely possibly be undefined.

packages/insomnia-app/app/ui/redux/modules/__tests__/workspace.test.ts
```diff
@@ -51,7 +58,47 @@ describe('workspace', () => {
...
+      const workspaces = await models.workspace.all();
+      expect(workspaces).toHaveLength(1);
+      const workspace = workspaces[0];
+      expect(workspace.name).toBe(workspaceName);
+      expect(workspace.scope).toBe(WorkspaceScopeKeys.collection);
+      expect(workspace.parentId).toBe(space._id);
```
dimitropoulos:
```suggestion
      expect(workspaces[0]).toMatchObject({
        name: workspaceName,
        scope: WorkspaceScopeKeys.collection,
        parentId: space._id,
      });
```

Perhaps something like this would be good?

packages/insomnia-app/app/ui/redux/modules/workspace.ts
```diff
@@ -22,13 +23,14 @@ const actuallyCreate = (patch: Partial<Workspace>, onCreate?: OnWorkspaceCreateC
   };
 };
 
-export type CreateWorkspaceCallback = (arg0: {
+export const createWorkspace = ({ scope, onCreate }: {
```
dimitropoulos, line 26: thank you re: `arg0`

dimitropoulos (commented):
Screenshot_20210608_162821

is this an expected behavior - that a new space may have no collections?

develohpanda:
> is this an expected behavior - that a new space may have no collections?

Yes. Currently the app only requires an active workspace, _any_ active workspace, it doesn't have to be in a particular space. I'd like to in the future remove that requirement of an active workspace, so this empty state is valid.

dimitropoulos (approved):

PR #3451, opened 2021-06-04
fix CodeMirror crash

dimitropoulos:
I noticed the following crash of CodeMirror occurring when you switch between different states of the response viewer:
Screenshot_20210604_152527

This leaves the app in a pretty nasty state that you have to restart to get out of:
Screenshot_20210604_152557

I bisected and confirmed that this started when we updated CodeMirror in https://github.com/Kong/insomnia/pull/3405

I searched the CodeMirror repo and found https://github.com/codemirror/CodeMirror/issues/4135 which suggested what I had already suspected, that we are calling something wrong.  That means the first course of action is to remove some of the places we ignored TypeScript errors and see if I uncover something.

and sure enough...

Thankfully, TypeScript alerted me to the fact that we were calling this wrong in the first place:
Screenshot_20210604_161744

I looked at the TypeScript types, and the problem became clear.

`setSelection`, which is what we were calling, has an API like this:
```ts
{
  /** Set a single selection range. anchor and head should be {line, ch} objects. head defaults to anchor when not given. */
  setSelection(
      anchor: Position,
      head?: Position,
      options?: { bias?: number; origin?: string; scroll?: boolean },
  ): void;
}
```

but `setSelections`, which is what we wanted to call, has an API like this:
```ts
{
  /**
   * Sets a new set of selections. There must be at least one selection in the given array. When primary is a
   * number, it determines which selection is the primary one. When it is not given, the primary index is taken from
   * the previous selection, or set to the last range if the previous selection had less ranges than the new one.
   * Supports the same options as setSelection.
   */
  setSelections(
      ranges: Array<{ anchor: Position; head: Position }>,
      primary?: number,
      options?: { bias?: number; origin?: string; scroll?: boolean },
  ): void;
}
```

and the fix:
Screenshot_20210604_162317

I feel that I definitely would have missed this subtle but critical distinction without TypeScript.

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
@@ -430,11 +430,10 @@ class CodeEditor extends Component<Props, State> {
   _restoreState() {
     const { uniquenessKey } = this.props;
 
-    // @ts-expect-error -- TSCONVERSION only try access if uniquenessKey is defined
-    if (!editorStates.hasOwnProperty(uniquenessKey)) {
+    if (uniquenessKey === undefined) {
       return;
     }
-    if (uniquenessKey === undefined) {
+    if (!editorStates.hasOwnProperty(uniquenessKey)) {
       return;
     }
     if (!this.codeMirror) {
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
@@ -442,7 +442,7 @@ class CodeEditor extends Component<Props, State> {
     this.codeMirror?.setHistory(history);
     // NOTE: These won't be visible unless the editor is focused
     this.codeMirror?.setCursor(cursor.line, cursor.ch, { scroll: false });
-    this.codeMirror?.setSelection(selections, undefined, { scroll: false });
+    this.codeMirror?.setSelections(selections, undefined, { scroll: false });
 
     // Restore marks one-by-one
     for (const { from, to } of marks || []) {
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
@@ -437,17 +437,20 @@ class CodeEditor extends Component<Props, State> {
     if (uniquenessKey === undefined) {
       return;
     }
+    if (!this.codeMirror) {
+      return;
+    }
 
     const { scroll, selections, cursor, history, marks } = editorStates[uniquenessKey];
-    this.codeMirror?.scrollTo(scroll.left, scroll.top);
-    this.codeMirror?.setHistory(history);
+    this.codeMirror.scrollTo(scroll.left, scroll.top);
+    this.codeMirror.setHistory(history);
     // NOTE: These won't be visible unless the editor is focused
-    this.codeMirror?.setCursor(cursor.line, cursor.ch, { scroll: false });
-    this.codeMirror?.setSelections(selections, undefined, { scroll: false });
+    this.codeMirror.setCursor(cursor.line, cursor.ch, { scroll: false });
+    this.codeMirror.setSelections(selections, undefined, { scroll: false });
 
     // Restore marks one-by-one
     for (const { from, to } of marks || []) {
-      this.codeMirror?.foldCode(from, to);
+      this.codeMirror.foldCode(from, to);
     }
   }
 
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
@@ -291,7 +291,10 @@ class CodeEditor extends Component<Props, State> {
   }
 
   getSelectionStart() {
-    // @ts-expect-error -- TSCONVERSION
+    if (!this.codeMirror) {
+      return;
+    }
+
     const selections = this.codeMirror.listSelections();
 
     if (selections.length) {
@@ -302,7 +305,10 @@ class CodeEditor extends Component<Props, State> {
   }
 
   getSelectionEnd() {
-    // @ts-expect-error -- TSCONVERSION
+    if (!this.codeMirror) {
+      return;
+    }
+
     const selections = this.codeMirror.listSelections();
 
     if (selections.length) {
@@ -313,14 +319,16 @@ class CodeEditor extends Component<Props, State> {
   }
 
   focusEnd() {
-    if (this.codeMirror) {
-      if (!this.hasFocus()) {
-        this.focus();
-      }
-
-      const doc = this.codeMirror.getDoc();
-      doc.setCursor(doc.lineCount(), 0);
+    if (!this.codeMirror) {
+      return;
     }
+
+    if (!this.hasFocus()) {
+      this.focus();
+    }
+
+    const doc = this.codeMirror.getDoc();
+    doc.setCursor(doc.lineCount(), 0);
   }
 
   hasFocus() {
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -1223,8 +1229,8 @@ class CodeEditor extends Component<Props, State> {
               display: 'none',
             }}
             readOnly={readOnly}
-            autoComplete="off" // NOTE: When setting this to empty string, it breaks the _ignoreNextChange
-            //   logic on initial component mount
+            autoComplete="off"
+            // NOTE: When setting this to empty string, it breaks the _ignoreNextChange logic on initial component mount
```
reynolek, line 1236: Is this comment associated with `defaultValue` or with `autoComplete`? The original placement implies `autoComplete`, but the new placement indicates `defaultValue`. Can we change this to `When setting <whatever property> to empty string, it breaks the _ignoreNextChange logic ...`

dimitropoulos: nope.  unfortunately, this was one thing that the flow-to-ts script pretty thoroughly botched.  we knew this and accepted it.

here it is before TS
https://github.com/Kong/insomnia/blob/bdb4b4e6610f69e84a0ae7ea1296078c59bbe88e/packages/insomnia-app/app/ui/components/codemirror/code-editor.js#L1049

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -331,42 +339,55 @@ class CodeEditor extends Component<Props, State> {
...
-        },
-        {
-          line: -1,
-          ch: -1,
-        },
-        {
-          scroll: false,
-        },
-      );
+    if (this.codeMirror?.isHintDropdownActive()) {
```
develohpanda, line 375:
```suggestion
    // Never do this if dropdown is open
    if (this.codeMirror?.isHintDropdownActive()) {
```

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -412,27 +433,20 @@ class CodeEditor extends Component<Props, State> {
...
+      return;
+    }
 
-    // @ts-expect-error -- TSCONVERSION only try access if uniquenessKey is defined
     const { scroll, selections, cursor, history, marks } = editorStates[uniquenessKey];
-    // @ts-expect-error -- TSCONVERSION
-    this.codeMirror.scrollTo(scroll.left, scroll.top);
-    // @ts-expect-error -- TSCONVERSION
-    this.codeMirror.setHistory(history);
+    this.codeMirror?.scrollTo(scroll.left, scroll.top);
```
develohpanda: Seems like this function can also follow the similar pattern used elsewhere, to return if `this.codeMirror` is not defined

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -412,27 +433,20 @@ class CodeEditor extends Component<Props, State> {
     if (!editorStates.hasOwnProperty(uniquenessKey)) {
       return;
     }
+    if (uniquenessKey === undefined) {
+      return;
+    }
```
develohpanda: There's an expect-error on line 432, that can be resolved if this check occurred first, right, or this check can be included in the if statement on line 433?

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -607,16 +619,13 @@ class CodeEditor extends Component<Props, State> {
...
       ? '\t'
-      // @ts-expect-error -- TSCONVERSION
-      : new Array(this.codeMirror.getOption('indentUnit') + 1).join(' ');
+      : new Array((this.codeMirror?.getOption?.('indentUnit') || 0) + 1).join(' ');
   }
 
   _handleBeautify() {
-    // @ts-expect-error -- TSCONVERSION
-    this._prettify(this.codeMirror.getValue());
+    this._prettify(this.codeMirror?.getValue());
```
develohpanda, line 631: Can `this._prettify` accept undefined?

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -1018,13 +1026,12 @@ class CodeEditor extends Component<Props, State> {
...
 
-    // @ts-expect-error -- TSCONVERSION
-    const value = this.codeMirror.getDoc().getValue();
+    const value = this.codeMirror?.getDoc().getValue() || '';
     // Disable linting if the document reaches a maximum size or is empty
-    const shouldLint =
-      value.length > MAX_SIZE_FOR_LINTING || value.length === 0 ? false : !this.props.noLint;
-    // @ts-expect-error -- TSCONVERSION
-    const existingLint = this.codeMirror.options.lint || false;
+    const isOverMaxSize = value.length > MAX_SIZE_FOR_LINTING;
```
develohpanda, line 1034: 👍🏽

develohpanda (commented):
> I noticed the following crash of CodeMirror occurring when you switch between different states of the response viewer

Could you please add a gif or add the exact reproduction steps, for QA purposes

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
dimitropoulos: wow, that's incredible, when you take a step back and look it it, what you're talking about it so obvious, but if you look at just each part it's hard to see.  great suggestion f44333cdcefe51ba5b90b41247f7f74618b2e4d1

dimitropoulos: (I checked and didn't see anywhere else in this file where there was such an obvious case of this that wouldn't be a logical change)

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
dimitropoulos: I didn't do this because I was trying not to make any logical changes, but now that you point it out this can't be a logical change since, haha, `undefined` can't be an object key (only strings and symbols can be).

8e52fecfc82332aebfa48874ded730b6ba400546

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
dimitropoulos: well from a types perspective the answer is definitely `yes` because:

```ts
  _prettify(code?: string) {
    this._codemirrorSetValue(code, true);
  }
```

-->

```ts
  _codemirrorSetValue(code?: string, forcePrettify = false) {
    if (typeof code !== 'string') {
      console.warn('Code editor was passed non-string value', code);
      return;
    }
```

But I was trying not to make any (other) logical changes here so I didn't change this.  The difference is now that instead of throwing a type error when we try to access `this.codeMirror?.getValue()` we'll get a console warning from `_codemirrorSetValue` that is thankfully a pretty helpful message.

I was trying not to go very far in this PR since I was really only making minor adjustments to help me find the bug.  Ultimately, if you're down for it, `_handleBeautify` is only called in one place and `_prettify` is only called by `_handleBeautify`:

```ts
  _handleBeautify() {
    this._prettify(this.codeMirror?.getValue());
  }

  _prettify(code?: string) {
    this._codemirrorSetValue(code, true);
  }
```

We could simplify things quite a bit by just doing:
```ts

  _prettify() {
    const code = this.codeMirror?.getValue();
    if (!code) {
      return;
    }
    this._codemirrorSetValue(code, true);
  }
```

and passing `_prettify` to the `onClick` instead of `_handleBeautify`.

develohpanda: Don't get me wrong, not asking to take this PR any further, I'm only asking whether undefined is accepted 👍🏽 thanks for the explanation, this is fine

develohpanda (approved):
### QA steps

#### On develop to reproduce
- create a request
- make a text selection on the response preview
- switch to the raw response
- switch back to preview mode
- observe error in console, and selection not preserved

#### On this PR
- create a request
- make a text selection on the response preview
- switch to the raw response
- switch back to preview mode
- no error, selections preserved

👍🏽

dimitropoulos:
thanks for the steps: looking back at the original repo steps I gave `when you switch between different states of the response viewer` could have been a heck of a lot clearer, heh. :smile_cat:

PR #225, opened 2017-05-17
Add option to toggle Menu Bar showing for Windows/Linux

Subzidion:
## What
Add option to toggle Menu Bar showing for Windows/Linux

## Why
Menu Bar is ugly to some people, so we should be able to hide it.

## Details
Set AutoHideMenuBar to true by default, and add an option in the settings panel to show the menu bar. Also add hot key for opening settings in the response pane since it is not visible by default.

## Notes
I added an updater function in `components/settings/general.js` to show/hide the Menu Bar when the option was toggled. I only wanted it to be called when the checkbox is updated instead of on every render of a component. If there's a better place to do the update, I can change it.

**Related Issue**: Closes #142

PR #3455, opened 2021-06-07
Provides Space id to "export all"

dimitropoulos:
closes: INS-725


packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -563,32 +563,23 @@ const writeExportedFileToFileSystem = (filename: string, jsonData: string, onDon
...
       let exportPrivateEnvironments = false;
       const privateEnvironments = environments.filter(e => e.isPrivate);
 
       if (privateEnvironments.length) {
-        const names = privateEnvironments.map(e => e.name).join(', ');
+        const names = privateEnvironments.map(environment => environment.name).join(', ');
         exportPrivateEnvironments = await showExportPrivateEnvironmentsModal(names);
       }
 
-      const fileNamePrefix = (workspace ? workspace.name : 'Insomnia All').replace(/ /g, '-');
```
develohpanda, line 590: What is the rationale behind dropping the kebab-casing?

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -563,32 +563,23 @@ const writeExportedFileToFileSystem = (filename: string, jsonData: string, onDon
   fs.writeFile(filename, jsonData, {}, onDone);
 };
 
-export const exportAllToFile = (workspaceId?: string) => async dispatch => {
+export const exportAllToFile = () => async dispatch => {
```
develohpanda: Do we ever call this action with a `workspaceId`? If not, 51485e031c41b76cbd28eda86c00168f709b3882 seems reasonable.

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
```diff
@@ -38,9 +43,9 @@ class SelectModal extends PureComponent<{}, State> {
     this.doneButton = n;
   }
 
-  _handleDone() {
+  _handleDone = async () => {
```
develohpanda: IMO we shouldn't mix-match function properties and should set this back to how it was. Is there some context I am missing, other than the things discussed [here](https://github.com/Kong/insomnia/pull/2963#issuecomment-765565861)?

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
```diff
@@ -1,94 +1,77 @@
...
-  state: State = initialState;
-
-  _setModalRef(m: Modal) {
-    this.modal = m;
-  }
-
-  _setDoneButtonRef(n: HTMLButtonElement) {
-    this.doneButton = n;
-  }
+export class SelectModal extends PureComponent<{}, SelectModalShowOptions> {
```
develohpanda, line 29: As part of this component refactor, it might be worth adding some docs/guidelines about how you see future modals being written

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
```diff
@@ -1,94 +1,77 @@
...
+  modal = createRef<Modal>();
+  doneButton = createRef<HTMLButtonElement>();
```
develohpanda, line 31: Refs 👍🏽

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -176,7 +176,7 @@ export const newCommand = (command: string, args) => async dispatch => {
         ),
         addCancel: true,
       });
-      dispatch(importUri(args.workspaceId, args.uri));
+      await importUri(args.workspaceId, args.uri)(dispatch);
```
develohpanda: A thunk can return an async function but you still call it as `await dispatch(importUri())`, you don't need to call the thunk and pass the dispatch in to the result ([ref](https://github.com/reduxjs/redux-thunk/blob/290acf90fa5afac5b49f286bb3d8fc51aa864ed3/README.md#L282-L286))

It doesn't feel like a good change to make just to satisfy the current (semi-working) types, because it goes against any redux-thunk documentation for future, even if it works

packages/insomnia-app/app/plugins/context/data.ts
```diff
@@ -8,55 +8,49 @@ interface PluginImportOptions {
...
+        format || 'json',
+      ),
+
+      har: async ({
+        workspace,
+        includePrivate,
+      }: HarExport = {}) => exportWorkspacesHAR(
+        workspace || null,
+        Boolean(includePrivate),
+      ),
```
develohpanda, line 53: > I wish I understand why `har` is called `har`, but `json` is called `insomnia`.  oh well.  I can't see where this is used outside of the tests.

Because `insomnia` is the Insomnia proprietary format, either as yaml or json, and `har` is a separate format. `insomnia` is not an alias for `json`.

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -318,6 +318,17 @@ export const selectActiveSpace = createSelector<any, {}, {}, Space | undefined>(
   },
 );
 
+// TODO(TSCONVERSION) type this properly when doing the rest of this file
```
develohpanda: This comment can be removed I think because `selectActiveSpace` is already typed, which means this selector is as well, by extension

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -587,7 +587,7 @@ const writeExportedFileToFileSystem = (filename: string, jsonData: string, onDon
   fs.writeFile(filename, jsonData, {}, onDone);
 };
 
-export const exportAllToFile = () => async (dispatch: Dispatch) => {
+export const exportAllToFile = (dispatch: Dispatch) => async (spaceId?: string) => {
```
develohpanda: This is the wrong way around, isn't it... when an action creator [returns a function](https://github.com/reduxjs/redux-thunk/blob/290acf90fa5afac5b49f286bb3d8fc51aa864ed3/README.md#motivation), that function already receives dispatch. We're just doing extra work here manually not following any of the examples set by `redux-thunk`

ie.

```ts
export const exportAllToFile = (spaceId?: string) => (dispatch: Dispatch) => {...};

// called later
dispatch(exportAllToFile('abc'));
```

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -36,6 +38,15 @@ export const ImportExport: FC<Props> = ({
     });
   }, [handleImportUri]);
 
+  const space = useSelector(selectActiveSpace);
+  const spaceName = useSelector(selectActiveSpaceName);
+  const activeWorkspace = useSelector(selectActiveWorkspace);
+  const activeResourceScope = activeWorkspace.scope === 'collection' ? strings.collection : strings.document;
```
develohpanda: https://github.com/Kong/insomnia/blob/5f4c19da35e90ffa2f3ae27e9b309ee48f9d4e16/packages/insomnia-app/app/common/get-workspace-label.ts#L5-L6

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -53,11 +64,11 @@ export const ImportExport: FC<Props> = ({
           <DropdownDivider>Choose Export Type</DropdownDivider>
           <DropdownItem onClick={handleShowExportRequestsModal}>
             <i className="fa fa-home" />
-              Current {strings.document.singular} / {strings.collection.singular}
+              Export Requests from the "{activeWorkspace.name}" {activeResourceScope.singular}
```
develohpanda: We're not _only_ exporting requests - perhaps `Export the "abcd" collection`?

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -4,19 +4,21 @@ import Link from '../base/link';
...
 import { strings } from '../../../common/strings';
+import { useSelector } from 'react-redux';
+import { selectActiveSpace, selectActiveSpaceName, selectActiveWorkspace } from '../../redux/selectors';
 
 interface Props {
   handleImportFile: () => void;
   handleImportClipBoard: () => void;
   handleImportUri: (uri: string) => void;
-  handleExportAllToFile: () => void;
+  handleExportAllToFile: (spaceId?: string) => void;
```
develohpanda: Because (and only because) `import-export.tsx` is a part of `components/settings`, it's intended use is within the settings modal, what are your thoughts on passing a `hideSettingsModal` prop here, instead of the export handler.

That way, we can use `useDispatch` directly in this component to dispatch the export action, and `settings-modal` does not need to be connected into redux.

develohpanda (changes requested):
Curious to understand why you are writing and consuming thunks as `actionCreator(dispatch)(param)` instead of `dispatch(actionCreator(param))`.

When consumed as the latter, thunks are written as the following, allowing access into state without any changes required for when the action creator is invoked for dispatch.

```ts
const actionCreator = (arg: string) => (dispatch: Dispatch, getState: Function) => { }
```

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: What else are we exporting? You know, I was wondering this and tried to figure it out but eventually gave up.  Since the function is called `handleShowExport*Requests*Modal` and the modal is called `Export*Requests*Modal` I went with that.

The thing is... it doesn't export the collection - it gives you a modal that then allows you to pick things (so far as I could tell, only requests) from the collection or document.

We don't need to get super deep into this I guess in this PR (maybe out of scope, however it's not like it's a big change: it's just a copy adjustment).

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: we talked about this the other day, sorry if I didn't explain it well.  the rational is that since we know we'll need it soon I left it there (and optional).  I was using it while testing but we can remove it.

still, I agree with you.  easy enough to add later.  no use in having dead code: a92f51c847a06398629bd68b9779cc306bdb5379

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: sounds good, changed in a92f51c847a06398629bd68b9779cc306bdb5379

packages/insomnia-app/app/plugins/context/data.ts
dimitropoulos: ahh, ok, that makes sense.  thanks.

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: yep, updated in a92f51c847a06398629bd68b9779cc306bdb5379

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: perfect, thanks.  was reaching for this but couldn't think of it when the time came.  fb0036fa26117da527118d6a1f8e47390c7c2e0f

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: great suggestion.  we were talking about this but I was afraid it would cause a cascade of other things and balloon in scope.  it didn't!  cool.  fixed in 4a5309a8e1f7e9b453982134d45c476fa352fbf7

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: sure.  they're all in need of work, but I agree that a special case for this one is not necessary e89be51625846bb5ad3eee1f521e0393261b3f7f

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
dimitropoulos: sure: 438fa86e311dd2923614b38e54448a700c98b100

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
dimitropoulos: this was a two-step thing, the [next commit](86945f389b999aef1f4d06610d43de900f73acc5) of which provided some context in the commit message:

```
it's too big of a change to make this a hook right here and now, but we can do the following to make this easier to work with:

1. use createRef instead of custom code
1. actually make private thing private
1. remove unnecessary autobind decorator
1. alphabetize props (to make it easier to make sure none were missed)
1. move `onDone` to state.  from what I can tell, there is no reason to do this, but please verify.  from looking at the two usages of this component, they're always the same anyway but if they change I would imagine we would want to not call the previous onDone, but rather the new one.
```

I'll add to that by saying that the intention here is to get as close as we can to hooks since this PR is an inappropriate place to to the main refactor that will allow us to use hooks for modals.  That's the only reason this function exists as a class method at all (bound or not).

All the same, it doesn't mix-match anymore (see the next commit), but if you have any qualms with the current implementation please just let me know and I'll switch it to class methods immediately.

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: not intentional.  I was taking the string as is and didn't catch that it was being wrapped.  good catch: 16808c7e62faa5e017839d7e9122a34ded68445e.  At least it's more obvious now :) so win-win

packages/insomnia-app/app/ui/components/settings/import-export.tsx
develohpanda: > Since the function is called handleShowExport*Requests*Modal and the modal is called Export*Requests*Modal I went with that.

I wouldn't always trust function names haha. Remember, older/legacy codebase that's gone through a lot of fundamental changes.

It _does_ export the collection. You have the option to choose _some_ requests when exporting a particular workspace, but Insomnia still does export all the auxiliary models - the workspace/collection, the cookie jar, environments, folders, unit tests, etc, and some of these are not optional eg environments.

https://github.com/Kong/insomnia/blob/6e073c8a38f4b9c783786e3a2c149a10c7d38dec/packages/insomnia-app/app/common/import.ts#L507-L530

This drop-down item exports a workspace and _in addition_ allows you to choose specific requests to export; it doesn't only export requests.

dimitropoulos: perfect.  very helpful.  makes much more sense, then.  It'd be great to _show_ users what we're actually exporting, then, even with disabled (but checked) checkboxes for things like environments, cookiejar, etc.

packages/insomnia-app/app/models/workspace.ts
```diff
@@ -16,20 +15,29 @@ export const canDuplicate = true;
...
 }
 
+export type BaseWorkspace = DesignWorkspace | CollectionWorkspace;
+
+export type WorkspaceScope = BaseWorkspace['scope'];
+
+export const WorkspaceScopeKeys: Record<WorkspaceScope, WorkspaceScope> = {
+  design: 'design',
+  collection: 'collection',
+} as const;
```
develohpanda: I suggest typing this in the following way, so long as it still works everywhere else. The duplication especially at the workspace model definition level is... 🤔 

```ts
interface GenericWorkspace<TScope extends 'collection' | 'design'> {
  name: string;
  description: string;
  certificates?: any;
  scope: TScope;
}

export type DesignWorkspace = GenericWorkspace<'design'>;
export type CollectionWorkspace = GenericWorkspace<'collection'>;
export type BaseWorkspace = DesignWorkspace | CollectionWorkspace;

export type WorkspaceScope = BaseWorkspace['scope'];

export const WorkspaceScopeKeys: Record<WorkspaceScope, WorkspaceScope> = {
  design: 'design',
  collection: 'collection',
} as const;
```

There is an improvement to make here with dropping `WorkspaceScopeKeys` (a remnant of pseudo-flow days) but that's out of scope for this PR.

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
```diff
@@ -1,83 +1,73 @@
...
 
-  hide() {
-    this.modal && this.modal.hide();
+  #handleSelectChange = (event: React.SyntheticEvent<HTMLSelectElement>) => {
+    this.setState({ value: event.currentTarget.value });
   }
 
-  show(data: Record<string, any> = {}) {
-    const { title, message, options, value, onDone, onCancel } = data;
-    this._doneCallback = onDone;
```
develohpanda, line 58: It's interesting that previously `onDone` was not put into state. Can you think of any reasons why that might be the case?

packages/insomnia-app/app/ui/components/modals/settings-modal.tsx
```diff
@@ -25,12 +25,10 @@ export const TAB_INDEX_THEMES = 2;
...
-  handleShowExportRequestsModal: Function;
-  handleExportAllToFile: Function;
-  handleImportFile: Function;
-  handleImportUri: Function;
-  handleToggleMenuBar: Function;
-  handleImportClipBoard: Function;
+  handleImportFile: () => void;
+  handleImportUri: (uri: string) => void;
+  handleToggleMenuBar: (hide: boolean) => void;
+  handleImportClipBoard: () => void;
```
develohpanda: This can be removed now, right?

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -1,98 +1,111 @@
...
+    showPrompt({
       title: 'Import Data from URL',
       submitName: 'Fetch and Import',
       label: 'URL',
       placeholder: 'https://website.com/insomnia-import.json',
-      onComplete: uri => {
+      onComplete: async (uri: string) => {
         window.localStorage.setItem('insomnia.lastUsedImportUri', uri);
-        this.props.handleImportUri(uri);
+        await importUri(activeWorkspace._id, uri)(dispatch);
```
develohpanda: Same thing here as per previous feedback

```suggestion
        await dispatch(importUri(activeWorkspace._id))
```

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -1,98 +1,111 @@
...
+        hideSettingsModal();
```
develohpanda, line 35: Do we need to wrap in a `try-finally` so the settings modal gets hidden properly in case of error?

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -1,98 +1,111 @@
...
+    });
+  }, [hideSettingsModal, activeWorkspace, dispatch]);
+
+  const showExportRequestsModal = useCallback(() => {
+    showModal(ExportRequestsModal);
+    hideSettingsModal();
+  }, [hideSettingsModal]);
+
+  const handleExportAllToFile = useCallback(async () => {
+    dispatch(exportAllToFile);
```
develohpanda:
```suggestion
    await dispatch(exportAllToFile());
```

`exportAllToFile` needs to update too 🤔 from my understanding async actions themselves don't take dispatch, they return a function which accept dispatch (as well as state and any extra overrides, following redux-thunk docs)

```diff
- export const exportAllToFile = async (dispatch: Dispatch) => {
+ export const exportAllToFile = () => async (dispatch: Dispatch) => {
```

packages/insomnia-app/app/ui/components/settings/import-export.tsx
develohpanda: For some yes, but not all - there's simplicity in not adding more than is necessary. Users don't necessarily need to know or understand the models that exist under the hood which always get exported no matter what. Just adds more noise in that case.

develohpanda:
```suggestion
              Export the "{activeWorkspace.name}" {activeResourceScope.singular}
```

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1573,6 +1576,88 @@ class App extends PureComponent<Props, State> {
   }
 }
 
+async function _moveDoc(docToMove, parentId, targetId, targetOffset) {
```
develohpanda, line 1579: I'm assuming nothing changed here and the function was just moved further up in the file

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1707,107 +1792,28 @@ function mapStateToProps(state, props) {
...
-    handleGoToNextActivity: global.goToNextActivity,
-    handleSetActiveWorkspace: global.setActiveWorkspace,
-    handleImportFileToWorkspace: global.importFile,
-    handleImportClipBoardToWorkspace: global.importClipBoard,
-    handleImportUriToWorkspace: global.importUri,
-    handleCommand: global.newCommand,
-    handleExportFile: global.exportWorkspacesToFile,
-    handleExportRequestsToFile: global.exportRequestsToFile,
-    handleInitializeEntities: entities.initialize,
-    handleMoveDoc: _moveDoc,
```
develohpanda, line 1729: Why have some of these actions been dropped? Are we certain they are not used?

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -835,9 +835,3 @@ export const initActiveActivity = () => (dispatch, getState) => {
   const initializeToActivity = overrideActivity || activeActivity;
   dispatch(setActiveActivity(initializeToActivity));
 };
-
-export const init = () => [
-  initActiveSpace(),
-  initActiveWorkspace(),
-  initActiveActivity(),
-];
```
develohpanda: I think this change should be switched back to how it was. It gives a clear indication about how redux should initialize this global module. Making `redux/modules/index.ts` aware of which _specific_ actions need to be dispatched in order to initialize the global module doesn't seem appropriate.

packages/insomnia-app/app/ui/redux/modules/index.ts
```diff
@@ -1,40 +1,48 @@
...
+  onCommand(bound.newCommand);
 
-  for (const action of global.init()) {
-    // @ts-expect-error -- TSCONVERSION
-    store.dispatch(action);
-  }
+  store.dispatch(initActiveSpace());
+  store.dispatch(initActiveWorkspace());
+  // @ts-expect-error -- TSCONVERSION need to merge in Redux-Thunk types to root
+  store.dispatch(initActiveActivity());
```
develohpanda: See comment above

packages/insomnia-app/app/ui/redux/modules/index.ts
```diff
@@ -1,40 +1,48 @@
 import { bindActionCreators, combineReducers } from 'redux';
-import * as entities from './entities';
+import { allDocs, addChanges, initializeWith, reducer as entitiesReducer } from './entities';
```
develohpanda: It feels like this file is generally a good example of a scope-creep-y change

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -299,25 +300,34 @@ export const selectActiveUnitTests = createSelector(
...
+export const selectSpaces = createSelector(
   selectEntitiesLists,
   (entities) => {
     // @ts-expect-error -- TSCONVERSION
     return entities.spaces;
   },
 );
 
-// TODO(TSCONVERSION) type this properly when doing the rest of this file
-export const selectActiveSpace = createSelector<any, {}, {}, Space | undefined>(
```
develohpanda: The generic being typed here actually means that this selector will return `Space | undefined`. The comment pertains to `{}` because state is not typed currently, but typing the expected return type means all downstream types work, and once we have state typed, then it will all work out of the box.

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -299,25 +300,34 @@ export const selectActiveUnitTests = createSelector(
...
   state => state.global.activeSpaceId,
   (entities, activeSpaceId) => {
-    // @ts-expect-error -- TSCONVERSION
     return entities.spaces[activeSpaceId];
   },
 );
 
+export const selectActiveSpaceName = createSelector(
+  selectActiveSpace,
+  (activeSpace?: Space) => {
```
develohpanda: If `selectActiveSpace` is typed correctly (which it was, see my previous comment), then this callback does not need to be typed, `createSelector` will know that the activeSpace is typed as `Space | undefined`

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -299,25 +300,34 @@ export const selectActiveUnitTests = createSelector(
...
+    if (activeSpace === undefined) {
+      return 'Insomnia';
```
develohpanda: This selector shouldn't be concerned with providing a fallback/default, because it doesn't know how the rest of the app works when `activeSpace` is undefined. It should really just return `string | undefined`

develohpanda (commented):
I didn't comment everywhere the redux dispatch is being done differently to the documentation (if I'm mistaken please provide further context), but the previous comments unfortunately weren't addressed

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
develohpanda: My honest opinion is that we should switch this back to the previous implementation, discuss how we want to approach the improvement of modal architecture (thanks for adding the readme), and only after agreement start to make these changes. So my vote will be to switch back to class methods - the changes you made with refs are still good!

packages/insomnia-app/app/ui/redux/modules/global.tsx
develohpanda: Doesn't look like this change aligns to the review? 🤔

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
dimitropoulos: ok, I'll do that first thing tomorrow.

packages/insomnia-app/app/ui/containers/app.tsx
dimitropoulos: yes

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
dimitropoulos: 6aaa7964000aec63036e8ed38016f79a44294290

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: 1f019e884069a1dd23bb22e74d020f83cbc2d1e0

packages/insomnia-app/app/models/workspace.ts
dimitropoulos: interesting suggestion 2b0bbd26f5fb7eea85f09f00edc46fae0db40eab yeah, it uncovered a few little things once I got working with it.  thanks.

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
dimitropoulos: same.  yeah, the best I can come up with is that react clears the state when the component unmounts (but that doesn't mean the callback will get garbage collected).  I didn't observe this bug happening, but if it exists then we kinda want to know about it since in hooks land we won't get away with "working around" react like this.  As with anything, I'm happy to change this if you want - just lemme know.

packages/insomnia-app/app/ui/components/modals/settings-modal.tsx
dimitropoulos: yes, the consequence of this involved removing an actual function but I checked like 3 times and it legit appears to not be used: 1029f43b13e9000d58cba05fa6dee90b1d84e169

please double-check

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: I'm kinda confused as to why this looked wrong to you on GitHub because this was updated in 8a9b5a8af7fd19741275505499f8d29ecb62ef28 which was pushed 7 hours before your comment here

packages/insomnia-app/app/ui/containers/app.tsx
dimitropoulos: I manually checked each and every one, but I would encourage you to do the same.

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: this isn't relevant to redux-thunks because the invocation is evaluated before dispatch is called:

```ts

const callee = console.log;

const myFunction1 = () => async someArg => { console.log(someArg); };
callee(myFunction1());

const myFunction2 = async someArg => { console.log(someArg); };
callee(myFunction2);
```

these two should be equivalent.  So would be:

```ts
const myFunction3 = () => () => () => async someArg => { console.log(someArg); };
callee(myFunction3()()());
```

All redux-thunk sees is the result of the invocation.  I didn't see any reason to needlessly wrap, but if you have one I'm open minded.

There are others that suffer from the same problem that was fixed with `exportAllToFile` (e.g. `initActiveActivity`, `goToNextActivity` and two, `loadStart` and `loadStop` that are functions for no reason), but I opted to not touch the others because I don't want the changeset to expand here even more.  I did actually start fixing `initActiveActivity` for what it's worth but ended up backing out the changes once I saw that it would mean a lot more changes.  I can fix the ones that weren't immediately affected by this PR if you want in a separate PR?

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: I see where you're coming from re: awareness.  f52922ad45030fdcad47daa5cd9e974cb015aa16

packages/insomnia-app/app/ui/redux/modules/index.ts
dimitropoulos: ditto

packages/insomnia-app/app/ui/redux/modules/index.ts
dimitropoulos: I'm a bit confused: we've discussed this a few times and my understanding is that you want to go with the "fix them as we encounter them" approach for these kinds of things.  It'd be easy enough to do a grep for `import * as ` in the codebase (there are 350 results) and change them all in one PR but my understanding is that you prefer to do things as they are in and affecting code we're already touching.

If you recall, I _did_ start changing all of these in the Phase 2 PR and you asked me to revert it (which, I did).

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: I didn't actually type that for what it's worth, but thanks for clarifying.

abe28ec88d4ece5c570a71930e3acf147014883e goes with the selectors you wrote in the previous PR

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: see abe28ec88d4ece5c570a71930e3acf147014883e but I believe this is as you are asking for, now

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: ok, updated in 659050e6cc79f325fe568b6e08c76374ff9f2c8d

hey and while we're on the topic: what's the situation in which there wouldn't be a space?

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: I agree.  that's why I always try to link the commit - in case I miss something or misunderstand or otherwise just get turned around.  how's this 7de65b552d4cf8d3a131776d91097b0c43034b34

packages/insomnia-app/app/ui/components/modals/select-modal.tsx
develohpanda: Was more wondering whether you had found anything I hadn't, seems like an OK change

packages/insomnia-app/app/ui/components/modals/settings-modal.tsx
develohpanda: Seems alright, just double checked. It looks like the `handleToggleMenuBar` functionality was lost at some point in time.... it was introduced here: https://github.com/Kong/insomnia/pull/225. Sort of inclined to not remove `handleToggleMenuBar` and reinstate the UI toggle switch for it in a separate PR. I didn't dig further as to where/why it was dropped.

packages/insomnia-app/app/ui/components/settings/import-export.tsx
develohpanda: Yeah I'm not sure either 🤔 maybe some sort of caching issue? Ah well

packages/insomnia-app/app/ui/components/settings/import-export.tsx
develohpanda: Ah! I see what you mean. Using the syntax in `myFunction1` is safer.

For example, if the action creator is updated to a function which receives a parameter (which is a valid scenario), TS will complain correctly. If you don't invoke the action creator, TS will not complain.

As an example, lets say the action creator needs to now accept a parameter.

If invoking the action creator:
```diff
- const actionCreator = () => async dispatch => { ... };
+ const actionCreator = (arg: string) => async dispatch => { ... };

dispatch(actionCreator()); // TS will error at compile time
```

If not invoking the action creator like in your change:
```diff
- const actionCreator = async dispatch => { ... };
+ const actionCreator = (arg: string) => async dispatch => { ... };

dispatch(actionCreator); // TS will NOT error at compile time
```

image

> There are others that suffer from the same problem

I don't necessarily agree that the others are _broken_ and so I am certainly opposed to changing them for the above reason

packages/insomnia-app/app/ui/containers/app.tsx
develohpanda: This is a good example for why reviews are unfortunately so time consuming, because most of these are not concerned with the scope for this feature. 😞

packages/insomnia-app/app/ui/redux/selectors.ts
develohpanda: The base space (which is a grouping of all the workspaces without a parent). This is not an actual _space_, it's a placeholder.

https://github.com/Kong/insomnia/blob/f625b34a77909fb9b76b3f5d2ec8d84a98bc6c28/packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx#L17-L21

develohpanda (commented):
(ignore)

packages/insomnia-app/app/models/workspace.ts
```diff
@@ -3,38 +3,40 @@ import * as models from './index';
...
 
-export const WorkspaceScopeKeys = {
-  design: 'design',
-  collection: 'collection',
-} as const;
-
-export type WorkspaceScope = ValueOf<typeof WorkspaceScopeKeys>;
-
-interface BaseWorkspace {
+interface GenericWorkspace<Scope extends string> {
```
develohpanda: The scope shouldn't extend string, though, it should extend the union of `design | collection`, because those are the only valid entries

packages/insomnia-app/app/models/workspace.ts
```diff
@@ -3,38 +3,40 @@ import * as models from './index';
...
 export type Workspace = BaseModel & BaseWorkspace;
 
-export function init() {
-  return {
-    name: `New ${strings.collection.singular}`,
-    description: '',
-    scope: WorkspaceScopeKeys.collection,
-  };
-}
+export const init = (): CollectionWorkspace => ({
```
develohpanda:
```suggestion
export const init = (): BaseWorkspace => ({
```

packages/insomnia-app/app/models/workspace.ts
```diff
@@ -123,39 +125,31 @@ async function _migrateEnsureName(workspace: Workspace) {
...
-  // Translate the old value
-  type OldScopeTypes = 'spec' | 'debug' | 'designer' | null;
-
-  switch (workspace.scope as OldScopeTypes) {
-    case 'spec': {
-      workspace.scope = WorkspaceScopeKeys.design;
+function _migrateScope(workspace: MigrationWorkspace) {
+  switch (workspace.scope) {
+    case 'collection':
+    case 'design':
```
develohpanda:
```suggestion
    case WorkspaceScopeKeys.collection:
    case WorkspaceScopeKeys.design:
```

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -1,98 +1,111 @@
...
     const lastUsedImportUri = window.localStorage.getItem('insomnia.lastUsedImportUri');
-    const promptOptions = {
+    const defaultValue = lastUsedImportUri ? { defaultValue: lastUsedImportUri } : {};
+    showPrompt({
       title: 'Import Data from URL',
       submitName: 'Fetch and Import',
       label: 'URL',
       placeholder: 'https://website.com/insomnia-import.json',
-      onComplete: uri => {
+      onComplete: async (uri: string) => {
```
develohpanda: Does this still need to be async? There's no await in the function.

packages/insomnia-app/app/ui/components/wrapper.tsx
```diff
@@ -685,12 +679,6 @@ class Wrapper extends PureComponent<WrapperProps, State> {
...
             <SettingsModal
               ref={registerModal}
-              handleShowExportRequestsModal={handleShowExportRequestsModal}
-              handleExportAllToFile={handleExportFile}
-              handleImportClipBoard={this._handleImportClipBoard}
-              handleImportFile={this._handleImportFile}
-              handleImportUri={this._handleImportUri}
-              handleToggleMenuBar={handleToggleMenuBar}
               settings={settings}
             />
```
develohpanda, line 683: 😍 funfact: we shouldn't change it in this PR but settings is just this https://github.com/Kong/insomnia/blob/1f0612a41817058387d3517f981235c61ba7fb90/packages/insomnia-app/app/ui/redux/selectors.ts#L43-L46

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -836,8 +836,8 @@ export const initActiveActivity = () => (dispatch, getState) => {
...
 };
 
-export const init = () => [
-  initActiveSpace(),
-  initActiveWorkspace(),
-  initActiveActivity(),
+export const initActions = [
+  initActiveSpace,
+  initActiveWorkspace,
+  initActiveActivity,
```
develohpanda: This doesn't quite switch it back to what it was; we should invoke the action creators and send the actions to `redux/modules/index.ts` to be dispatched, see my reasoning https://github.com/Kong/insomnia/pull/3455#discussion_r648827987

develohpanda (changes requested):
As is the expectation with PRs that involve UI changes, please update the PR description with a gif to show everything working, and also add any extra notes for things to be aware of, decisions/tradeoffs you made.

### QA results

- The smoke test failures seem legitimate, the app takes much longer to launch

- Navigation is broken
	- can't open a workspace

 
	- after creating a workspace to open one, can't go back to the homepage or change the active tab


- If the base space is active, the export drop-down is bugged
image

- Cannot export the active workspace

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: I don't agree with you, but that's fine - I hear where you're coming from and now I understand much better why you wanted this change.  All good. :) updated in d210bd008221306c55a29e8671a51092693d7eda

dimitropoulos: regarding
> I don't necessarily agree that the others are broken and so I am certainly opposed to changing them for the above reason

Take a look at the redux docs https://redux.js.org/tutorials/fundamentals/part-2-concepts-data-flow#actions

Screenshot_20210610_061412

and compare that to one of the actions I mentioned in the list of ones I felt were broken:

Screenshot_20210610_061516

I listed the specific ones instead of talking in generalities to try to make that part clear, but reading back I could have just shown this comparison to _actually_ make it clear.  Sorry bout that.

packages/insomnia-app/app/ui/components/modals/settings-modal.tsx
dimitropoulos: Ok, if you're alright then I'm alright with it.  Just like you were saying re: passing a `spaceId` to the exporAllToFile function: no sense in keeping dead code once we discover it to be dead.

We can always come and grab this toggleMenuBar function later.  It will be safe and sound here in git.

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: As far as I can tell the answer is "no" because the importUri can't throw because it's already wrapped in its own try-catch-finally

Screenshot_20210610_063357

packages/insomnia-app/app/ui/containers/app.tsx
dimitropoulos: I hear you.  My offer stands to add the dead code back if you wish.  After all, it's not lost on me that my assessment (and yours) could be incorrect in some assumption we don't know we're making about the life/death of the ones that were dropped.  While it appears dead to me, I definitely could be wrong, and if you even just don't have the bandwidth to check just say the word and I'll gladly bring them all back.

packages/insomnia-app/app/ui/redux/modules/index.ts
dimitropoulos: sure, reverted in: 9b0d2ad4c0b045d25d3bfd3dc5ffbada92ab03e6

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: ok, thanks for explaining

packages/insomnia-app/app/models/workspace.ts
dimitropoulos: 6c719b0479928f1c12c6179c3b7f91860c887de7

packages/insomnia-app/app/models/workspace.ts
dimitropoulos: for what it's worth, I only didn't keep this because above you [said](https://github.com/Kong/insomnia/pull/3455#discussion_r647868273) you wanted to stop using it.  I agree totally with the part about not doing that refactor in this PR (hence why I didn't), but since I was directly touching these lines I didn't keep this part because of your feedback.

dimitropoulos: e64bfba9f20403c7ee9785e2e6bc725fbc5f8c09

packages/insomnia-app/app/ui/components/wrapper.tsx
dimitropoulos: sweet.  hopefully, if nothing else, this work with ImportExport will show that it's very possible to remove all the `_handle`ers being drilled like they are.

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: huh.  welp.  gotta say I don't see the connection at all.  You were talking about thunks before but two out of the three of these (`initActiveSpace` and `initActiveWorkspace`) return objects.

Actually, this is literally the opposite of what you were asking for [before](https://github.com/Kong/insomnia/pull/3455#discussion_r647872817), isn't it?  Because now it's doing 
- [A] `store.dispatch(action())` where `action` returns a function _or_ an object literal

and here you're asking for it to do
- [B] `store.dispatch(action)` where `action` is a function that returns an object literal _or_ a function that returns a thunk.  

Whereas before you were asking for evaluating the function at the point of dispatch.

I don't remember, but I bet the reason I did this before is because I was in the mode of trying to follow what you were talking about above regarding how you want these to be dispatched.  Of course, haha, _now_ I have the benefit of having https://github.com/Kong/insomnia/pull/3455#discussion_r648827987 which does a great job of explaining what you were thinking (much more than I understood at the time I wrote this code).

Still, you're right that my aim was to make it how it was before so I hope 6d48c9c7498df24b29861c72c583208791e07e9d is what you're looking for. :smile:

packages/insomnia-app/app/models/workspace.ts
dimitropoulos: I don't agree but I've made the change you're asking for here: 2399c85c4029528ee8f84d307653f103ac6eecd1

My rationale is/was:
1. This appears to me to be a clear violation of the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) which states:
    > Abstractions should not depend on details.
 
    In this case, the abstraction is the higher-level generic, and the detail is the type implementations `DesignWorkspace` and `CollectionWorkspace`.  Forcing the higher-level abstraction to have explicit knowledge of the lower-level details
1. It is still true that design and collection are the only valid entries because the generic type is private to this module.  Due to that, no other consumer in the code can import do `GenericWorkspace<'foo'>` where something is asking for a `BaseWorkspace`.
1. This means duplicating the state of which strings are allowed.  Thankfully, unlike most duplication, TypeScript will catch it if they get out of sync.
1. I (wrongly) interpreted it was just a little mistake in something you were sketching out quickly to demonstrate merging the two prior separate declarations and I didn't take it literally or mention it, which I should have done.  My bad.  apologies.

packages/insomnia-app/app/ui/containers/app.tsx
dimitropoulos: update: this caused one (very easy to fix (af675460b3798a36574bc1784b12851dd0603505) but very critically breaking) bug, so I reverted them all in 3ae1f286dff30570e06cb90b313b89dee056ff88.

(the only exception is what used to be `handleExportFile` which is the thing specifically worked on by the ticket for this PR)

dimitropoulos:
re:
> If the base space is active, the export drop-down is bugged

I guess we both missed this in review - the fallback was there originally but after you asked to remove it in https://github.com/Kong/insomnia/pull/3455#discussion_r647880362 I didn't know how to put the app in the state with a undefined space.

updated in fb27f01a6d807b1b399917a11fe0f62ff1290cb8

Screenshot_20210610_081055

dimitropoulos:
re: the other things in https://github.com/Kong/insomnia/pull/3455#pullrequestreview-680330004

They were all fixed by af675460b3798a36574bc1784b12851dd0603505 (essentially a one-liner).

(I'm working on making a gif now but having technical troubles)

packages/insomnia-app/app/models/workspace.ts
develohpanda: I can see where you are coming from but the union of `'design' | 'collection'` is being used as a type the same way `string` is a type. We know for certain that the scope can only be one of those two and not any string so it does feel like a reasonable use 🤔 

Is it reasonable to say that `Scope extends string` is also violation of that principle for the same reason - it is restricting what it can be? 

Perhaps another (untested) option can be:

```ts
interface CommonWorkspace { ... }; // excluding scope
interface DesignWorkspace extends CommonWorkspace {
	scope: 'design';
}
interface CollectionWorkspace extends CommonWorkspace {
	scope: 'collection';
}

```

Tonne of different ways we can go about it, I suppose. Thanks for explaining your rationalle

dimitropoulos: cool cool

packages/insomnia-app/app/ui/redux/modules/global.tsx
develohpanda: > Actually, this is literally the opposite of what you were asking for before, isn't it? ... before you were asking for evaluating the function at the point of dispatch.

It's not different. It relates directly to the comment I explained above, of not passing _action creators_ to dispatch, actually invoking the action creators to extract the action (be they object literals or thunk functions).

It just so happens that in most places we invoke the action creator at the time of dispatch (for example, `useDispatch()(exportAllToFile())`), but in this case we are preparing a _list of actions to initialize the global module_, so we should invoke all of the action creators there, then send them off for dispatch.

Does this explanation help to show how the two are different?

> Note: I understand and agree that in cases like `loadStart()`, it doesn't need to be an action creator function. However, in the case of `initActiveWorkspace`, it does need to be an action creator function because some logic needs to be executed before the action is created.

packages/insomnia-app/app/ui/redux/modules/index.ts
```diff
@@ -1,40 +1,48 @@
 import { bindActionCreators, combineReducers } from 'redux';
```
develohpanda: Please revert this entire file and extract it to a separate PR, because the changes here are unrelated to the ticket being resolved. (You can, of course, keep the circular dependency comment if it is still valid).

My [previous comment](https://github.com/Kong/insomnia/pull/3455#discussion_r647878958) referred to this entire file being out of scope, not the single import (it was edited in the same minute as the initial comment, to mention the file, and not after the fact).

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -836,7 +836,7 @@ export const initActiveActivity = () => (dispatch, getState) => {
   dispatch(setActiveActivity(initializeToActivity));
 };
 
-export const init = () => [
+export const initActions = [
```
develohpanda: Please undo this change. These actions creators previously would be invoked _when required_ (when they are to be dispatched, without `/redux/modules/index.ts` knowing which specific actions they are, back to the comment about awareness).

The action creators will now be invoked when this file is imported, and there is no way of ensuring that the resources required by these action creators (now or in the future) will be available the first time this file is imported.

develohpanda (changes requested):
### QA

<details>
<summary>🟢 Export a single workspace</summary>

1. Create a collection titled "ABC"
2. Create a request
3. Open preferences and export using `Export the "ABC" collection`
	
Result: The export file only contains the children of that workspace

</details>

<details>
<summary>🟢 Export all workspaces in a local space</summary>

1. Create a new space titled "ABC"
2. Create a collection
3. Create a request
8. Open preferences and export using (`All documents and collections in the "ABC" space`)

Result: the exported file contains all workspaces (and their children) in the "ABC" space

</details>

<details>
<summary>🔴 Export all workspaces in the base space</summary>

1. Create a new space
2. Create a workspace (document or collection)
3. Create a request
4. Go to dashboard
5. Activate the base "Insomnia" space
6. Create a workspace
7. Create a request
8. Open preferences and export using (`All documents and collections in the "Insomnia" space`)

Result: the exported file contains workspaces in multiple spaces (but should only contain workspaces in the base space, ie. those workspaces with `parentId = null`)

image

There are several ways to go about this, feel free to propose a solution and request review before starting on code if that will be easier.

</details>

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: please note that the way I originally did it didn't have this problem (nor did the way it was before this PR).  I only invoked the functions because you asked me to specifically here:

> we should invoke the action creators and send the actions to redux/modules/index.ts to be dispatched
> https://github.com/Kong/insomnia/pull/3455#discussion_r648844598

The original change was not a logical change (as far as I'm aware, I'd be curious to know if it wasn't).

I'm trying to follow your feedback exactly to make it exactly how you want it, and I can see that I focused on the specifics you said instead of the general statement (i.e. put it back how it was).

This happened in this situation twice, actually: because here you said:
> I think this change should be switched back to how it was. It gives a clear indication about how redux should initialize this global module. Making redux/modules/index.ts aware of which specific actions need to be dispatched in order to initialize the global module doesn't seem appropriate.
> https://github.com/Kong/insomnia/pull/3455#discussion_r647878207

And so, there too, I sent you back a commit that tried to implement the specific thing you were asking for.  In https://github.com/Kong/insomnia/pull/3455#discussion_r649086152 was trying to provide a side-by-side comparison to explain why it seemed like you were asking for opposite things but I think maybe we are just approaching it differently - not sure to be honest, but no biggie either way :smile:.

Well all the same, apologies for the mixup.  76e95507a69d696a2bad93910e6665a8e5903253 should be logically equal as well as stylistically equal to what was there before.

packages/insomnia-app/app/ui/redux/modules/index.ts
dimitropoulos: > it was edited in the same minute as the initial comment

off topic, but just to mention (cuz it's kinda hidden) you can see timestamps by hovering over the dates.  The initial comment was made at 8:26pm EDT and the edit was 7 minutes later at 8:33.  I can't say for sure whether I saw the original and then assumed it hadn't changed or whether I simply only saw the original and misread.

Either way, thanks for saying it clearly here.  Happy to do so.

d81b5e9fbcf9b2de9b88b725996d9261db5e3edb should make it so there are no changes at all in this PR for this file.

dimitropoulos:
re:
> Export all workspaces in the base space

This seems to be out of scope of the ticket: "Export just needs to know which Space to export"

I absolutely agree that this work needs to be done asap, but it's gotta be in a followup PR based on my understanding.  I can work on it immediately after this PR is merged, as far as I am concerned (I'll make another ticket if so).

cc: @reynolek lemme know if you disagree

develohpanda:
> This seems to be out of scope of the ticket: "Export just needs to know which Space to export" ... it's gotta be in a followup PR based on my understanding

The ticket title (while not great) and description do say that export needs to know which space to export. From a user's and UX perspective, _the base space is a space_ and when exporting the base space, only the items in the base space should be in the export. So handling the base space is not out of scope and this PR is incomplete for the feature. 

In saying that, to get things moving I am okay with merging this and resolving the bug with the base space in a subsequent PR.

develohpanda (approved):
> This seems to be out of scope of the ticket: "Export just needs to know which Space to export" ... it's gotta be in a followup PR based on my understanding

The ticket title (while not great) and description do say that export needs to know which space to export. From a user's and UX perspective, _the base space is a space_ and when exporting the base space, only the items in the base space should be in the export. So handling the base space is not out of scope and this PR is incomplete for the feature. 

In saying that, to get things moving I am okay with merging this and resolving the bug with the base space in a subsequent PR.

I suggest waiting for product review from @reynolek (see QA results here: https://github.com/Kong/insomnia/pull/3455#pullrequestreview-681363315) before merging this 👍🏽

reynolek (approved):
Approved under condition of new ticket created for base space

dimitropoulos:
INS-763

PR #3456, opened 2021-06-08
Filter remote collections by the active space

develohpanda:
Closes INS-751

Notes:
- the "pull" dropdown will:
	- not show at all, if not logged in
	- not show at all, on the base and local spaces
	- will show disabled on a remote space after logging out
	- will show as enabled on a remote space if logged in
- when we pull a remote collection, we now set the parentId to be that of the remote space
- because we set the parentId, we need to update the diff logic to ignore the `parentId` property, the same way the `modified` property is ignored
- the above doesn't work out of the box - `modified` gets deleted, but `parentId` needs to be reset back to null, so that the hash does not change. This was written first as a failing test here [`3681288` (#3456)](https://github.com/Kong/insomnia/pull/3456/commits/3681288e25f0c206cc445b1be3087fc5bf6fac7f) and then fixed in a subsequent commit

Cleanup:
- I added schemas and builders for several sync models, because the tests were failing as they were all testing with incorrect types
- Converted `RemoteWorkspacesDropdown` to a function component, and extracted the refresh & pull logic into a custom hook
- Extracted the pull logic to be reused in `RemoteWorkspacesDropdown` and `SyncDropdown`

TODO:
- [x] How should the collections in the private team be shown?
	- They don't need any special treatment, all remote collections will (eventually) belong to a team/space
- [x] Should the pull drop-down be hidden on local spaces and the base space? (IMO yes)
	- Yes
- [x] Pull and set the parent id on the workspace. ~This will then be stored after sync 🤔 is that okay? How does this work with older clients~ ignore the parentId of the workspace for sync
	- the parent id is deterministic for all users because the the `space._id` for a remote space is created from the remote team id (https://github.com/Kong/insomnia/blob/5774a0ac096c32e90f52c39a751d285a7de90a70/packages/insomnia-app/app/ui/hooks/space.ts#L17-L24)


packages/insomnia-app/app/sync/vcs/ignore-keys.ts
```diff
@@ -0,0 +1,34 @@
...
+  return false;
+};
+
+export const clearIgnoredKeys = (doc: BaseModel) => {
+  for (const key of ALWAYS_IGNORED_KEYS) {
+    delete doc[key];
+  }
+
+  for (const { key, defaultValue } of getModelIgnoredKeys(doc.type)) {
+    // If we delete the key, we then have a different hash because of the deletion
```
DMarby: Is this because `parentId` exists on already synced documents and is set to `null`, as that's the default value?
It's not immediately obvious to me if that's the case based on these comments, hence the question 😅

packages/insomnia-app/app/sync/vcs/index.ts
```diff
@@ -154,10 +156,7 @@ export default class VCS {
     return {
       stage,
       unstaged,
-      key: hashDocument({
-        stage,
-        unstaged,
-      }).hash,
+      key: hash({ stage, unstaged }).hash,
```
DMarby: Just to double check, but was this broken out to a method which doesn't do the modified/parentId removal since we're not actually hashing a document here?

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
```diff
@@ -11,65 +8,64 @@ import * as models from '../../../models';
...
-    });
-  }
-
-  async _handlePullRemoteWorkspace(project: Project) {
-    const { vcs } = this.props;
-
+    setLoading(true);
+    setRemoteProjects(await vcs.remoteProjects(spaceRemoteId));
+    setLocalProjects(await vcs.localProjects());
+    setLoading(false);
```
DMarby: Since we're now setting state in each operation here, rather then collecting it and then setting the state once once all operations succeeded, there *might* be some edge cases where we get partial/invalid state if an operation throws an error.

Not sure if we should opt for something like this as a bit of a defensive measure?
```suggestion
    setLoading(true);
    const remoteProjects = await vcs.remoteProjects(spaceRemoteId); 
    const localProjects = await vcs.localProjects();
    setRemoteProjects(remoteProjects);
    setLocalProjects(localProjects);
    setLoading(false);
```

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
```diff
@@ -94,101 +91,104 @@ class RemoteWorkspacesDropdown extends Component<Props, State> {
...
+  if (!isRemoteSpace) {
+    return null;
   }
 
-  componentDidMount() {
-    this._refreshRemoteWorkspaces();
-  }
-
-  renderButton(disabled = false) {
+  // Show a disaled button if remote space but not logged in
```
DMarby:
```suggestion
  // Show a disabled button if remote space but not logged in
```

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
```diff
@@ -85,6 +81,7 @@ class RemoteWorkspacesDropdown extends Component<Props, State> {
         const workspace: Workspace = await models.initModel(models.workspace.type, {
           _id: project.rootDocumentId,
           name: project.name,
+          parentId: spaceId || null,
```
DMarby: Not a blocker for this PR, but similar logic to this cloning also exists in the SyncDropdown component for when you have a local workspace that has the same ID as a remote one, so I think we're likely going to need it there as well later?

DMarby (commented):
Didn't do much functional testing yet, but looks good so far!

packages/insomnia-app/app/sync/vcs/ignore-keys.ts
develohpanda: Yes that's right.

I'll update this slightly, because I don't think we can truly _delete_ any key (unless its a brand new key which has previously not been synced). I'll update this code slightly to make that clearer, and update the comment 🙏🏽

packages/insomnia-app/app/sync/vcs/index.ts
develohpanda: That's right!

https://github.com/Kong/insomnia/blob/51bcab77d11b5186cc5b85b045798824eb946461/packages/insomnia-app/app/sync/vcs/util.ts#L454-L477

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
develohpanda: Good catch, this seems like a reasonable approach to take. [`3c2e1c8` (#3456)](https://github.com/Kong/insomnia/pull/3456/commits/3c2e1c8f26ed97937844f77412e22ebc215170cc)

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
develohpanda: Oh I vaguely remember when this was duplicated to the sync-dropdown as well, nice catch. I'll see how complex it is to address in this PR, otherwise at least make a record of it.

packages/insomnia-app/app/sync/vcs/ignore-keys.ts
develohpanda: 561ac39165b48c798dc320ab8dfa53870a48a1ec

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
develohpanda: Updated in d473b1a6492ff9ba56f8d95f435af8e2194d185a

reynolek:
LGTM -- Only thing that I think can use improvement (but definitely not blocking) is the initial pull of remote spaces. The blink in for the data is a bit jarring...Theres probably a UX improvement like a spinner or some clear indicator that something is occurring. 

Slower connections may take longer to load which would then cause this to feel worse.

reynolek (approved):

develohpanda:
> The blink in for the data is a bit jarring...Theres probably a UX improvement like a spinner or some clear indicator that something is occurring.

Yah it is, I'll have to double check but I think develop behaves the same way as well.

develohpanda:
Oh wait, you mean for the initial pull of remote _spaces_. There is a spinner already, it just loads  real fast. And also, the initial pull will happen on mount and not on open, so should be less jarring than the gif shows.

packages/insomnia-app/app/common/constants.ts
```diff
@@ -94,9 +94,12 @@ export const UPDATE_CHANNEL_BETA = 'beta';
 export const UPDATE_URL_MAC = 'https://updates.insomnia.rest/builds/check/mac';
 export const UPDATE_URL_WINDOWS = 'https://updates.insomnia.rest/updates/win';
 
-// AP
+// API
 export const API_BASE_URL = 'https://api.insomnia.rest';
 
+// SYNC
+export const DEFAULT_BRANCH_NAME = 'master';
```
dimitropoulos, line 101: Off-topic to the PR, but on-topic to this change: I can see that this was here before this PR and just moved to this file from `sync-dropdown.tsx`, but I'm curious why it isn't in `config.json` so it's configurable?  Don't get me wrong, I'm not trying to say that this is the time to make that change, but rather just wondering if there's any intention of giving the user control over this in the future.

packages/insomnia-app/app/models/space.ts
```diff
@@ -42,6 +42,10 @@ export function getById(_id: string) {
   return db.getWhere<Space>(type, { _id });
 }
 
+export function getByRemoteId(remoteId: string) {
+  return db.getWhere<Space>(type, { remoteId });
+}
```
dimitropoulos: I don't understand where/how this function is used.  any pointers?

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
```diff
@@ -0,0 +1,159 @@
...
+import VCS from '../';
+import { DEFAULT_BRANCH_NAME } from '../../../common/constants';
+import * as models from '../../../models';
+import { Workspace } from '../../../models/workspace';
+import { globalBeforeEach } from '../../../__jest__/before-each';
+import MemoryDriver from '../../store/drivers/memory-driver';
+import { projectSchema } from '../../__schemas__/type-schemas';
+import { pullProject } from '../pull-project';
+
+jest.mock('../');
```
dimitropoulos: this seems like the kind of thing that has a high potential of breakage in the future since it's
- sensitive to file rearrangements
- also doesn't contain anything that you might grep when looking to rename the file it points at
- will not be picked up by language server protocol automatic renames like a regular import would

can you think of a more resilient way to do this?

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  afterEach(() => {
+    expect(vcs.setProject).toHaveBeenCalledWith(project);
+    expect(vcs.checkout).toHaveBeenCalledWith([], DEFAULT_BRANCH_NAME);
+  });
+
+  describe('creating a new project', () => {
+    beforeEach(() => {
+      (vcs.getRemoteBranches as jest.MockedFunction<typeof vcs.getRemoteBranches>).mockResolvedValue([]);
```
dimitropoulos: instead of having to cast to `jest.MockedFunction` at every usage, is there a way we can type `vcs` (once) where it's declared a few lines above?

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    afterEach(() => {
+      expect(vcs.pull).toHaveBeenCalledWith([]);
+    });
+
+    it('should overwrite the parentId only for a workspace with null', async () => {
+      // Arrange
+      const spaceId = undefined;
+      const w = await models.workspace.create({ _id: project.rootDocumentId, name: project.name });
+      const r = await models.request.create({ parentId: w._id });
```
dimitropoulos: to me, readability would strongly benefit from not using single letter variable names here, but I'll leave it up to you to decide

packages/insomnia-app/app/common/constants.ts
DMarby: The value of this is currently tied to the internals of the API, so I'm not sure if it would make much sense for it to be configurable in any capacity right now. Longer term we do want to get rid of this hardcoded value from the app itself to make it possible to change.

packages/insomnia-app/app/models/space.ts
develohpanda: Ah, it was used in an earlier iteration, good catch.

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
develohpanda: The `jest.mock` function has never been sensitive to file rearrangements (AFAIK), and it's less an issue of the path parameter used by it

The correct solution would be to not export VCS as default from an `index.ts`: https://github.com/Kong/insomnia/blob/0c476f8ca22d109ad6c15af571a32ece86012e0a/packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts#L2

Although, this still won't fix the line you commented on, because `jest.mock` does not (AFAIK) update as you rearrange files.

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
develohpanda: Do you know of a way we can use `jest.MockedClass` to achieve this? I tried but couldn't get it working in a timebox. I'll add a type to the declaration above though, good idea.

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
develohpanda: [`d81a73a` (#3456)](https://github.com/Kong/insomnia/pull/3456/commits/d81a73a1243b214f658dc793f73631431121b534)

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
develohpanda: [`abebd9c` (#3456)](https://github.com/Kong/insomnia/pull/3456/commits/abebd9c225b1be9a13c7d852e248283fe6de7bc8) & [`3a78864` (#3456)](https://github.com/Kong/insomnia/pull/3456/commits/3a78864a294c2ce87793440dd182838a710720b5)

packages/insomnia-app/app/models/space.ts
develohpanda: [`ea5eb22` (#3456)](https://github.com/Kong/insomnia/pull/3456/commits/ea5eb220fd9df5aef7c95a9e771919621c7b3d81)

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
```diff
@@ -4,21 +4,23 @@ import { DEFAULT_BRANCH_NAME } from '../../../common/constants';
 import * as models from '../../../models';
 import { Workspace } from '../../../models/workspace';
 import { globalBeforeEach } from '../../../__jest__/before-each';
-import MemoryDriver from '../../store/drivers/memory-driver';
 import { projectSchema } from '../../__schemas__/type-schemas';
 import { pullProject } from '../pull-project';
+import { mocked } from 'ts-jest/utils';
+import { MockedObjectDeep } from 'ts-jest/dist/utils/testing';
```
develohpanda: Hmm this shouldn't be here

develohpanda:
> Slower connections may take longer to load which would then cause this to feel worse.

@reynolek this is what users will see the first (and every) time they open the drop-down, with the spaces having already loaded in the background, and checking for a refresh when opened. I have artificially slowed down the loading, to show the spinner.

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
develohpanda: https://github.com/Kong/insomnia/pull/3471

dimitropoulos: > Although, this still won't fix the line you commented on, because jest.mock does not (AFAIK) update as you rearrange files.

Right.  The (now approved) PR you put up at least solves one of the 3 cases, so I'll take what I can get.  I can't see the 3rd point ever being addressed by LSP (although, I've been surprised before, haha).  And the first point is mitigated by the 2nd being solved.

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
dimitropoulos: cool, looks much better!

DMarby (approved):

packages/insomnia-app/app/sync/vcs/__tests__/pull-project.test.ts
develohpanda: >  I can't see the 3rd point ever being addressed by LSP (although, I've been surprised before, haha)

Yeah, still not the case lol. At least tests properly fail when jest doesn't mock properly, but unfortunate that it's essentially a "runtime" (while running tests) failure.

PR #3462, opened 2021-06-09
Make _getOrCreateRemoteProject able to specify what team the project should belong to

DMarby:
Implements the new functionality from https://github.com/Kong/insomnia-api/pull/151, while leaving current behaviour untouched.

The `teamId` (current active space `remoteId`) needs to be passed to `_createRemoteProject` somehow, the easiest way would probably be to just pass it through the push/pull callsites, but I'll leave that up to what you think is best when implementing the functionality that uses this @develohpanda

### DIFF
--- a/packages/insomnia-app/app/sync/vcs/index.ts
+++ b/packages/insomnia-app/app/sync/vcs/index.ts
@@ -1212,7 +1212,7 @@ export default class VCS {
     const symmetricKey = await crypt.generateAES256Key();
     const symmetricKeyStr = JSON.stringify(symmetricKey);
 
-    const teamKeys: any[] = [];
+    const teamKeys: {accountId: string, encSymmetricKey: string}[] = [];
     let encSymmetricKey: string | undefined;
 
     if (teamId && teamPublicKeys) {
--- a/packages/insomnia-app/app/sync/vcs/index.ts
+++ b/packages/insomnia-app/app/sync/vcs/index.ts
@@ -1215,7 +1215,11 @@ export default class VCS {
     const teamKeys: {accountId: string, encSymmetricKey: string}[] = [];
     let encSymmetricKey: string | undefined;
 
-    if (teamId && teamPublicKeys) {
+    if (teamId) {
+      if (!teamPublicKeys?.length) {
+        throw new Error('teamPublicKeys must not be null or empty!');
+      }
+
       // Encrypt the symmetric key with the public keys of all the team members, ourselves included
       for (const { accountId, publicKey } of teamPublicKeys) {
         teamKeys.push({
--- a/packages/insomnia-app/app/sync/vcs/index.ts
+++ b/packages/insomnia-app/app/sync/vcs/index.ts
@@ -514,11 +514,11 @@ export default class VCS {
 
   async _createRemoteProject(workspaceId: string, workspaceName: string, teamId?: string) {
     if (teamId) {
-      var teamKeys = await this._queryTeamMemberKeys(teamId)
-      return this._queryCreateProject(workspaceId, workspaceName, teamId, teamKeys.memberKeys)
+      const teamKeys = await this._queryTeamMemberKeys(teamId);
+      return this._queryCreateProject(workspaceId, workspaceName, teamId, teamKeys.memberKeys);
     }
 
-    return this._queryCreateProject(workspaceId, workspaceName)
+    return this._queryCreateProject(workspaceId, workspaceName);
   }
 
   async push() {
--- a/packages/insomnia-app/app/sync/vcs/index.ts
+++ b/packages/insomnia-app/app/sync/vcs/index.ts
@@ -505,11 +505,20 @@ export default class VCS {
     let project = await this._queryProject();
 
     if (!project) {
-      project = await this._queryCreateProject(localProject.rootDocumentId, localProject.name);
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/sync/vcs/index.ts
```diff
@@ -1163,32 +1172,98 @@ export default class VCS {
...
     // Generate symmetric key for ResourceGroup
     const symmetricKey = await crypt.generateAES256Key();
     const symmetricKeyStr = JSON.stringify(symmetricKey);
-    // Encrypt the symmetric key with Account public key
-    const encSymmetricKey = crypt.encryptRSAWithJWK(publicKey, symmetricKeyStr);
+
+    const teamKeys: any[] = [];
+    let encSymmetricKey: string | undefined;
+
+    if (teamId && teamPublicKeys) {
```
develohpanda: What should happen if `teamPublicKeys` is an empty array - is it safe to assume it never will be an empty array?

packages/insomnia-app/app/sync/vcs/index.ts
```diff
@@ -1163,32 +1172,98 @@ export default class VCS {
...
+      publicKey: string;
+    }[],
+  ) {
     // Generate symmetric key for ResourceGroup
     const symmetricKey = await crypt.generateAES256Key();
     const symmetricKeyStr = JSON.stringify(symmetricKey);
-    // Encrypt the symmetric key with Account public key
-    const encSymmetricKey = crypt.encryptRSAWithJWK(publicKey, symmetricKeyStr);
+
+    const teamKeys: any[] = [];
```
develohpanda:
```suggestion
    const teamKeys: {accountId: string, encSymmetricKey: string}[] = [];
```

develohpanda (approved):
> The `teamId` (current active space `remoteId`) needs to be passed to `_createRemoteProject` somehow, the easiest way would probably be to just pass it through the push/pull callsites, but I'll leave that up to what you think is best when implementing the functionality that uses this @develohpanda

That sounds good 🙏🏽 will sort this out when I link it up

Code looks good, I haven't tested it, since it's not linked to anything yet, and the smoke tests pass which means the app still runs.

packages/insomnia-app/app/sync/vcs/index.ts
DMarby: Nice!

packages/insomnia-app/app/sync/vcs/index.ts
DMarby: It *should* never be an empty array, and the API will just reject it as `teamPublicKeys` must not be empty if `teamId` is set, but I've rewritten this check to be a bit more defensive now, and so that it's more obvious as well.

DMarby:
@develohpanda Feel free to merge this after approval, it's fully functional now that the API has been deployed, but I've opted not to enable auto-merge since this stuff is a bit brittle 😅 

Easiest way to test it if you want to is probably to hardcode `teamId` in `_createRemoteProject`, and then create a new collection/setup sync/create snapshot + push. You can verify that the collection have been shared correctly with the team after doing so by using the "Share" button, it should show up as already shared with the team.

packages/insomnia-app/app/sync/vcs/index.ts
```diff
@@ -1163,32 +1172,102 @@ export default class VCS {
...
-    const { publicKey } = this._assertSession();
+  async _queryTeamMemberKeys(
+    teamId: string,
+  ): Promise<{
+    memberKeys: {
+      accountId: string;
+      publicKey: string;
+    }[];
+  }> {
+    const { teamMemberKeys } = await this._runGraphQL(
```
dimitropoulos, line 1183: totally up to you, but if you wanted to update
```ts
  async _runGraphQL<T = Record<string, any>>(
    query: string,
    variables: Record<string, any>,
    name: string,
  ): Promise<T> {
  ```
  
Then it would provide some type checking here that we're otherwise not getting flowing from the call site to the return rather than the other way around.

Then we can just do 

```ts
  async _queryTeamMemberKeys(teamId: string) {
    const { teamMemberKeys } = await this._runGraphQL<{
      teamMemberKeys: {
        memberKeys: {
          accountId: string;
          publicKey: string;
        }[];
      }
    }>(   
```

here.

By the way, it's is possible to autogen correct types (e.g. https://www.apollographql.com/blog/tooling/apollo-codegen/typescript-graphql-code-generator-generate-graphql-types does it) but we're likely not going to be at that point until we finish the TypeScript conversion.

DMarby: Really like this, will do so in a later PR as per our Slack discussion :)

packages/insomnia-app/app/sync/vcs/index.ts
develohpanda: Awesome 🙌🏽

develohpanda:
> Easiest way to test it if you want to is probably to hardcode `teamId` in `_createRemoteProject`, and then create a new collection/setup sync/create snapshot + push. You can verify that the collection have been shared correctly with the team after doing so by using the "Share" button, it should show up as already shared with the team.

Just tested this way and it seems to be working! See logs 

image

I then deleted the collection, removed the hard-coded teamId, and followed the same approach again, this time the share modal only shows "Private" and no particular team.

develohpanda (approved):

PR #3465, opened 2021-06-11
Add response time in api response to test latency

### LINKED ISSUE #3297: Implement a response time calculation to test API response latency.
I would like to test my API endpoints latency by adding a test that measures it. Currently, Postman uses responseTime: 

```
pm.test("Response time is less than 1000ms", function () {
    pm.expect(pm.response.responseTime).to.be.below(1000);
});
```

After talking to @reynolek  - the solution would be seem to be as follows: 

"It would need a variable added to the request config (a start time), and then once the request has returned, calculate the time between return and start time in milliseconds, and include that in the data being returned by send."

**Additional context**
Looking to move away from Postman, so hoping to have this added. Thank you!

amaaniqbal:
Closes #3297

### DIFF
--- a/packages/insomnia-app/app/common/send-request.ts
+++ b/packages/insomnia-app/app/common/send-request.ts
@@ -4,6 +4,10 @@ import { send } from '../network/network';
 import { getBodyBuffer } from '../models/response';
 import * as plugins from '../plugins';
 
+function toMs (num) {
+  return Math.round(num * 1000) / 1000;
+}
+
 export async function getSendRequestCallbackMemDb(environmentId, memDB) {
   // Initialize the DB in-memory and fill it with data if we're given one
   await db.init(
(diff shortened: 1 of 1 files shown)


amaaniqbal:
@develohpanda I created a PR for #3297. Please have a look!

reynolek (approved):
Hey @amaaniqbal ! Thanks for the PR. Looking at the code, this is functional and does the requested ask in the open issue.

There is some clean up that I would like to do, mostly around handling the proper types. The function `toMs` injects an implicit `any` type, which is one thing we are looking to eliminate going forward. However to fix this, there are larger changes that need to be made in insomnia's networking code, some of which falls out of scope of the PR, and I wouldn't expect any contributor to make those changes.

develohpanda (approved):
Thanks for the PR! This is awesome

PR #3466, opened 2021-06-11
Update types in networking layer for response times in testing

reynolek:
Type fixes for #3465 along with related type changes. See self comments for context.


packages/insomnia-app/app/network/network.ts
```diff
@@ -212,7 +212,7 @@ export async function _actuallySend(
           url: renderedRequest.url,
           parentId: renderedRequest._id,
           error: err.message,
-          elapsedTime: 0,
+          elapsedTime: (curl.getInfo(Curl.info.TOTAL_TIME) as number || 0) * 1000,
```
reynolek: I added `|| 0` because of how this was originally defined. Technically I wouldnt expect this to ever be 0, but just in case there are edge cases when `TOTAL_TIME` isnt set in this instance, 0 would maintain current functionality.

reynolek: Side note, the type provided for `curl.getInfo` is `string | number | null` since `getInfo` wraps a lot of different functionality.

`TOTAL_TIME` is always a number, making this type cast safe.

develohpanda: This code-path seems to be reached while running plugin response hooks, so in that case I feel like this elapsedTime being 0 is actually valid :thinking:

packages/insomnia-app/app/network/network.ts
```diff
@@ -843,6 +842,7 @@ export async function _actuallySend(
           {
             statusMessage,
             error,
+            elapsedTime: curl.getInfo(Curl.info.TOTAL_TIME) as number * 1000,
```
develohpanda, line 845: I think this should maybe be reverted? Not totally sure, but now when you send a request to a server not running (causing node-libcurl to error), you see a timestamp

image

packages/insomnia-app/app/network/network.ts
reynolek: Ah ok, and I assume plugins do not use curl to communicate. Ill reset this back to zero and add a comment as to why it's 0.

packages/insomnia-app/app/network/network.ts
reynolek: I have no real strong opinion here, but I would actually consider this a bug fix. A request to an NXDOMAIN does still take time to issue a curl and determine that it is an NXDOMAIN.

How useful is the info for how fast an NXDOMAIN returned? Probably not very useful. But that doesn't mean it is 0.

dimitropoulos: @develohpanda could you say more about why you would expect a request (no matter the result) to take `0ms` please?  I would expect it to be non-zero.

packages/insomnia-app/app/network/network.ts
develohpanda: They would but I'm not sure what this would mean for plugins, will need to experiment with plugin code that flows through this code path

packages/insomnia-app/app/network/network.ts
develohpanda: That sounds reasonable TBH, just wanted to double check that was an intentional change. If I think about it, Insomnia _is_ sending off a request, it doesn't matter that the server on the other side doesn't exist/failed to respond, Insomnia's networking layer still went over the network to do something. (This applies to the previous plugin comment as well, but needs some more QA there). As you were!

develohpanda (approved):

packages/insomnia-app/app/network/network.ts
reynolek: Let's come back to it and leave the functionality unchanged. I don't know enough about the plugin process to feel confident unlike in other spots.

Issue #2585, opened 2020-09-03
GraphQL autocomplete query fields losing sorting


brunovigano:
**Describe the bug**
Consuming a graphql endpoint when using the shortcut CTRL + Space to autocomplete the query fields and using the arrow keys to navigate between the indexes, Insomnia switches the index to the top of the list by itself.

**Expected behavior**
Be able to navigate between the indexes of the suggested fields without losing their order.

**Screenshots**
![ezgif com-gif-maker (2)](https://user-images.githubusercontent.com/25367686/92144160-ab496480-edec-11ea-9bbd-63d441aa39bd.gif)

**Desktop (please complete the following information):**
Version: Insomnia Core 2020.4.0-beta.6
Release date: 01/09/2020
OS: Windows_NT x64 10.0.19041
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.69.1-DEV OpenSSL/1.1.1d zlib/1.2.11 WinIDN libssh2/1.9.0_DEV nghttp2/1.40.0

develohpanda:
Closed by #3468

Issue #1015, opened 2018-06-27
[Bug] Selection in GraphQL autocomplete list is reset to the first item when scrolling by arrows


4ekki:
- Insomnia Version: 5.16.6
- Operating System: Win10

## Details
[Bug] Selection in GraphQL autocomplete list is reset to the first item when scrolling by arrows

S2R: 
1. Create new request with GraphQL type and set address to http://try.sangria-graphql.org/graphql 
2. Type in 'query {}'
3. Press [Ctrl]+[Space] to get autocompletion list 
4. Press [Arrow down] button several times

As a result, on #4 selection is reverted to the first item in the list. 

On this gif, I press [Arrow down] twice to select '__schema' item. After that, 'human' item becomes selected automatically. 
![gif](https://user-images.githubusercontent.com/38757986/41994701-fb3b4932-7a57-11e8-973d-31b570823ca2.gif)

gschier:
Hmm, that's odd. Going to take a look now

gschier:
Hmm, I can't seem to reproduce this issue (on Mac). Can you describe more details around when it happens? Is it after a delay, always on the 3rd item, etc? What if the autocomplete list is longer?

4ekki:
Reproduces every time on my machine, doesn't depend on list length or the exact final position. 
Also, I've reproduced it with scrolling the list by mousewheel, you just need a bit longer list for the scrollbar appear.

gschier:
So it always happens on the 3rd item?

4ekki:
No, it doesn't depend on the final position, can happen on 2nd or 5th, doesn't matter. But looks like it happens after the same time interval, ~1sec after opening the list.

4ekki:
This one is still actual in 6.0.2

gschier:
I was never able to reproduce it, unfortunately. Perhaps someone who is experiencing the problem wants to take a shot at fixing (or at least debugging) it?

rafaelrenanpacheco:
I use Insomnia with Arch Linux and Gnome as DE, and for a long time I see this bug in the GraphQL autocomplete dropdown.

nicholaslythall:
Confirming this is still an issue in 7.1.1 on macOS Catalina 10.15.3

nickwinsor:
+1 on this. 
I experience this issue hundreds of times a day. 

One thing to add though is that letting the autocomplete dropdown sit for a second before navigating, bypasses the bug.

[...]

PR #3468, opened 2021-06-13
Fix for GraphQL - now autocomplete does not reset anymore

### LINKED ISSUE #3437: Cursor jumps back to first index during autocomplete
**Describe the bug**
The cursor almost always jumps to the first index after ~ 1 second when I am navigating through the autocomplete dialog / overlay.

**To Reproduce**
1. Create any GraphQL request
2. write your query against a schema that you have previously fetched
3. press `control + space` to show autocomplete dialog and quickly move with cursor keys to an item that is not the first in list
4. see the cursor jump back to the first index

**Expected behavior**
Cursor does not jump to first index.

**Screenshots**

 - OS: macOS 11.4
 - Installation Method: https://insomnia.rest/download

```
Version: Insomnia 2021.3.0
Release date: 4/28/2021
OS: Darwin x64 20.5.0
Electron: 9.3.5
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64
node-libcurl: libcurl/7.73.0 OpenSSL/1.1.1j zlib/1.2.11 brotli/1.0.9 zstd/1.4.9 libidn2/2.1.1 libssh2/1.9.0 nghttp2/1.41.0
```

MrSnix:
**Before you read**:
It is not so simple to understand this bug, I will try to be as clear as possible.

**The problem**
The autocomplete feature in GraphQL, under some conditions, gets reset to the first item when scrolling by arrows.
These conditions occur based on many factors. First, the user and then the hardware.

This happens because two events get fired at the same time (more or less), one causing the first render and the second one causing the second re-render (the reset itself). Specifically, I'm talking about the `_codemirrorTriggerCompletionKeyUp` event and `cm.ShowHints` issued by the keyboard event `Ctrl+Space`.

- `_codemirrorTriggerCompletionKeyUp` is used to display the hint panel when a user has just finished typing something.
- `Ctrl+Space` is used to force the displaying of the hint panel wherever requested.

**So, how they conflict with each other?**

When the user forces the hint panel to show up using `Ctrl+Space`, this little fellow starts running **even if he shouldn't**, every single time.

``` js
this._autocompleteDebounce = setTimeout(() => {
  doc.execCommand('autocomplete');
}, 700);
```

**Some users complained that this bug happened about a second after the hint panel opened. 
Well, this is the reason, the delay is set at 700ms** 

In this gif, I used `console.trace()` to determine who was the caller of the `autocomplete` feature, you can clearly see the second render comes from the `_autocompleteDebounce` event that is calling `cm.showHint` even if it's already open.
**The second render reset the position.**

Jun-13-2021 16-36-34

image

**How did you fix this?**
Well, like this.

``` js
// You don't want to re-trigger the hint dropdown if it's already open
// for other reasons, like forcing its display with Ctrl+Space
if (this.codeMirror?.isHintDropdownActive()) {
  return;
}

this._autocompleteDebounce = setTimeout(() => {
  doc.execCommand('autocomplete');
}, 700);
```

Now the second re-render never happen if already open.

Jun-13-2021 16-47-23

**Why on some systems happens so often meanwhile on others never show up?**
Because `setTimeout` on a faster system usually gets executed way before the hint panel even show up to the user, so basically the user never sees the second render happening even if it happens.

**Have you tested it thoroughly?**
Not really. On my macOS, this bug happened on average once every 15. Now, after the fix, it seems to no longer happen.

It closes #3437, #2585 and #1015


develohpanda:
Need to QA this but code looks reasonable, nice catch on `clearInterval`!

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -913,7 +913,7 @@ class CodeEditor extends Component<Props, State> {
 
   _codemirrorEndCompletion() {
     if (this._autocompleteDebounce !== null) {
-      clearInterval(this._autocompleteDebounce);
+      clearTimeout(this._autocompleteDebounce);
```
dimitropoulos, line 931: definitely.  great catch.  I searched the codebase for all `setTimeout` calls to see if we were doing this elsewhere and found none.  Unfortunately the return type of `setTimeout` and `setInterval` is exactly the same, `NodeJS.Timeout` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/timers.d.ts

zdeneksejcek:
Please merge this, it's do annoying, it happens to me every time - I select first item in list instead of selected one, so I have to delete it and do CMD+SPACE again. Sometimes it does me couple of times before I realise I have to wait 1s before I select something. It makes my workflow slow :(

develohpanda (approved):
I QA'd this and it seems to be working better now. I was able to reproduce the jumpy behavior on the live app (following some of the steps in the linked issues, using `https://api.spacex.land/graphql/`) and couldn't reproduce it in this PR 🙌🏽

gatzjames (approved):
Also QA'd and looks really good to me!

MrSnix:
Yeah, I'm glad I was able to help on this. Yuppie!

PR #3472, opened 2021-06-14
Remove issue management bot for now

develohpanda:
Until we are able to configure it properly, it seems to just be adding too much noise


dimitropoulos (approved):
approving since we agreed in standup that we'd do this

PR #3474, opened 2021-06-15
Automatically enable sync for collections created in a remote space

develohpanda:
### Demo and QA steps

<details>
<summary>Create collection in base space (no sync options)</summary>

</details>

<details>
<summary>Create collection in local space (no sync options)</summary>

</details>

<details>
<summary>Create collection in remote space (auto enable sync)</summary>

</details>

<details>
<summary>Create collection in remote space while signed out (need to manually setup sync after log in, as normal)</summary>

</details>

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
+++ b/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
@@ -36,7 +36,7 @@ const REFRESH_PERIOD = 1000 * 60 * 1;
 interface Props {
   workspace: Workspace;
   workspaceMeta?: WorkspaceMeta;
-  space?: Space;
+  space: Space;
   vcs: VCS;
   syncItems: StatusCandidate[];
   className?: string;
--- a/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
+++ b/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
@@ -26,6 +26,8 @@ import { strings } from '../../../common/strings';
 import { pullProject } from '../../../sync/vcs/pull-project';
 import { Space } from '../../../models/space';
 import { WorkspaceMeta } from '../../../models/workspace-meta';
+import { pushSnapshotOnInitialize } from '../../../sync/vcs/initialize-project';
+
 // Stop refreshing if user hasn't been active in this long
 const REFRESH_USER_ACTIVITY = 1000 * 60 * 10;
 // Refresh dropdown periodically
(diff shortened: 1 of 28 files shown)


packages/insomnia-app/app/ui/redux/modules/git.tsx
```diff
@@ -130,7 +130,7 @@ const noDocumentFound = (gitRepo: GitRepository) => {
       okLabel: 'Yes',
       addCancel: true,
       message: `No ${strings.document.singular.toLowerCase()} found in the repository for import. Would you like to create a new one?`,
-      onConfirm: async () => await dispatch(createWorkspaceWithGitRepo(gitRepo)),
```
develohpanda, line 133: This doesn't return a promise so doesn't need async-await

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
```diff
@@ -382,7 +391,10 @@ class SyncDropdown extends PureComponent<Props, State> {
     }
 
     return (
-      <DropdownButton className="btn--clicky-small btn-sync btn-utility wide text-left overflow-hidden row-spaced">
+      <DropdownButton
+        className="btn--clicky-small btn-sync btn-utility wide text-left overflow-hidden row-spaced"
+        disabled={initializing}
```
develohpanda, line 396: When the drop-down is opened it runs some refresh logic. If the drop-down is opened while pushing during initialize, the push fails. We can either do `vcs.newInstance().push()` but I feel like disabling the drop-down is the cleaner option, because you shouldn't do anything in the drop-down while initializing anyway.

DMarby (approved):

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
dimitropoulos: too bad our dropown doesn't support a loading state e.g. https://ant.design/components/dropdown/
Screenshot_20210617_152717

dimitropoulos (changes requested):
(we debugged it on a call but updating here for posterity) there as a breaking bug whereby if you followed the reproduction steps exactly (i.e. without clicking into a collection in the remote steps) you would have a broken vcs instance and sync will not work.

develohpanda:
> (we debugged it on a call but updating here for posterity) there as a breaking bug whereby if you followed the reproduction steps exactly (i.e. without clicking into a collection in the remote steps) you would have a broken vcs instance and sync will not work.

Looks like this is not related to the vcs instance being broken, but the bug (where sync fails to set up automatically) occurs when the collection being created is the _first workspace (document or collection) in the space_. This is why the gif in the description above works but it didn't work on your computer, following the same steps.

This is a deeply nested issue but it is caused by the following two facts:
1. the database is mirrored in redux, but through events (which cause background updates)
2. the requirement for an active workspace to always exist (even on the dashboard where it is not applicable)

I have created a separate ticket (thus will be a separate PR (#3494) to handle this bug INS-771 (currently in progress). As a solution I will be concretely exploring how difficult it is to remove the requirement for an active workspace to always exist

@dimitropoulos @reynolek

dimitropoulos (approved):

PR #3476, opened 2021-06-15
Introducing tooltip helper for Vim mode on MacOS

### LINKED ISSUE #3448: Press and Hold not working on Vim keymap (on MacOS)
**Describe the bug**
When I enable Vim keymap, long pressing a move key (`j` or `k` for instance) only moves the cursor once.

**To Reproduce**
1. Go to 'Settings', enable Vim keymap in Text Editor Key Map
2. Open a request body and long press `j`
3. The cursor only moves one line down

**Expected behavior**
The cursor should keep moving down until I stop pressing the key

**Screenshots**

 - OS: macOS 11.4
 - Installation Method: Download
 - App Version: 2021.3.0

**Additional context**
I tried to run command `defaults write NSGlobalDomain ApplePressAndHoldEnabled -bool true` and restart Insomnia but it did not work

MrSnix:
**Summary**:
This line has been introduced to make the Vim mode correctly work on macOS.
Contrary to other platforms, if you preventDefault() a keypress event, no repeat events are fired for that key as it is being held down. This snippet, reintroduce the event firing disabling the Apple behaviour. The flag applies only at app level (just for Insomnia), it won't influence other applications.

When set to **false**, the custom apple behaviour is disabled and works as expected by code-mirror.

`systemPreferences.setUserDefault('ApplePressAndHoldEnabled', 'boolean', false);`

Jun-15-2021 19-47-04

I'm holding `j` and `k` in this gif

**Important**:
I have put this one-liner near the `BrowserWindow` object because I feel like it's more appropriate, it is part of the main Electron interface. As always, you can commit to this branch and put it somewhere else if you have a better and more appropriate place. Lastly, you must wrap the statement inside an `if(isMac())` to prevent it from running against Windows or Linux systems.

It closes #3448


packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -407,6 +407,31 @@ export function createWindow() {
...
+    * Note [2]:
+    * The type definition for:
+    * setUserDefault(key: string, type: string, value: string): void
+    * is expecting a string for the 'value' field but not really...
+    * The implementation is expecting a primitive value, not a string.
+    * That's why the type-checking has been disabled in the next line.
+    * */
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
dimitropoulos:
```suggestion
    // @ts-expect-error <description of why there's a type error on this line>
```

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -407,6 +407,31 @@ export function createWindow() {
...
+    * (i.e. defaults write -g ApplePressAndHoldEnabled -bool true),
+    * this will override app-level preferences.
+    *
+    * Note [2]:
+    * The type definition for:
+    * setUserDefault(key: string, type: string, value: string): void
+    * is expecting a string for the 'value' field but not really...
+    * The implementation is expecting a primitive value, not a string.
+    * That's why the type-checking has been disabled in the next line.
+    * */
```
dimitropoulos:
```suggestion
    */
```

doing this mirrors how jsdoc comments terminate

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -407,6 +407,31 @@ export function createWindow() {
         type: 'separator',
       },
     );
+
+    /*
+    * This line has been introduced to make the Vim mode correctly work on MacOS
+    * Contrary to other platforms, if you preventDefault() a keypress event,
+    * no repeat events are fired for that key as it is being held down.
+    * This snippet, reintroduce the event firing disabling the Apple behaviour.
```
dimitropoulos: I'm not sure I understand what this line is saying.  Perhaps try to rephrase?

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -407,6 +407,31 @@ export function createWindow() {
...
+    * if you've set ApplePressAndHoldEnabled at a global level
+    * (i.e. defaults write -g ApplePressAndHoldEnabled -bool true),
+    * this will override app-level preferences.
+    *
+    * Note [2]:
+    * The type definition for:
+    * setUserDefault(key: string, type: string, value: string): void
+    * is expecting a string for the 'value' field but not really...
+    * The implementation is expecting a primitive value, not a string.
+    * That's why the type-checking has been disabled in the next line.
```
dimitropoulos: Regarding line wrapping, I know we aren't very consistent about this (yet!) in this project, but it's ideal to avoid wrapping these kinds of comments for the sake of line length and nothing else.  For example, I frequently split screen with line wrapping and here's how this would look in (even) a fairly wide split:



compare that to a more inlined version (not apples to apples):

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -407,6 +407,31 @@ export function createWindow() {
...
+    * This line has been introduced to make the Vim mode correctly work on MacOS
+    * Contrary to other platforms, if you preventDefault() a keypress event,
+    * no repeat events are fired for that key as it is being held down.
+    * This snippet, reintroduce the event firing disabling the Apple behaviour.
+    * The flag applies only at app-level (just for Insomnia),
+    * it won't influence other applications.
+    *
+    * Note [1]:
+    * if you've set ApplePressAndHoldEnabled at a global level
+    * (i.e. defaults write -g ApplePressAndHoldEnabled -bool true),
```
dimitropoulos: it's helpful if you use backticks to delineate code from text, for example:

```
    /*
    * This line has been introduced to make the Vim mode correctly work on MacOS.
    * Contrary to other platforms, if you `preventDefault()` a keypress event no repeat events are fired for that key as it is being held down.
    * This snippet, reintroduce the event firing disabling the Apple behaviour. The flag applies only at app-level (just for Insomnia), and won't influence other applications.
    *
    * Note: if you've set `ApplePressAndHoldEnabled` at a global level (i.e. `defaults write -g ApplePressAndHoldEnabled -bool true`), this will override app-level preferences.
    */

    // @ts-expect-error the type definition for: `setUserDefault(key: string, type: string, value: string): void` is expecting a string for the 'value' field but not really... The implementation is expecting a primitive value, not a string.
    systemPreferences.setUserDefault('ApplePressAndHoldEnabled', 'boolean', false);
```

packages/insomnia-app/app/main/window-utils.ts
dimitropoulos: see https://github.com/Kong/insomnia/pull/3476/files#r652096873 for a concrete example

MrSnix:
@dimitropoulos 
I have applied the requested changes. 
As always, let me know. 
I'm open to any feedback :D

I apologize if I missed something.

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -407,6 +407,18 @@ export function createWindow() {
...
+
+    /*
+    * This line has been introduced to make the Vim mode correctly work on MacOS.
+    * Contrary to other platforms, when `ApplePressAndHoldEnabled` is enabled, it stops from firing repeatedly the keypress event while the key is held down after calling `preventDefault()`.
+    * The flag applies only at app-level (just for Insomnia), it won't influence other applications.
+    *
+    * Note:
+    * If you've set `ApplePressAndHoldEnabled` at a global level (i.e. `defaults write -g ApplePressAndHoldEnabled -bool true`), this will override app-level preferences.
+    */
+
```
dimitropoulos: oh, only one final thing is that maybe we should remove this empty line because the comment starts with `this line` and then isn't butt-up right against the line it's talking about?

dimitropoulos (commented):
I can't test it since I'm on linux, but the code looks good to me!  thanks!

DMarby:
@MrSnix Thank you for fixing this, this is awesome!

@reynolek @MrSnix @develohpanda @dimitropoulos I don't have a strong opinion, but do you think it might make sense/be possible to only conditionally set the system preference if vim mode is enabled (and perhaps use `removeUserDefault` when vim mode is disabled to fallback to the system global setting)?

By default in maCOS if you hold eg `a`, rather then `a` repeating itself a bunch of times, you get a popup menu for special characters: 
image

With this change as it is currently, I think we'd lose that behaviour within the app even when vim mode is disabled, which I think would be good to avoid if possible.

reynolek:
Ah good call -- It'd be even better if it was limited to non-insert mode with vim bindings enabled (although I dont know if thats possible).

MrSnix:
Hi @DMarby, thanks for your feedback :D
I did some test about switching at runtime the flag but **it is not** working as I'd expect.

I will investigate and try to let you know more tomorrow.

develohpanda:
> Do you think it might make sense/be possible to only conditionally set the system preference if vim mode is enabled (and perhaps use removeUserDefault when vim mode is disabled to fallback to the system global setting)?

1. It feels like this should be an option which is given to the user, because even with Vim mode enabled, some users may want to keep using the special-characters instead of press-and-hold for navigation. We're adding one feature but taking another away.
    1. If we provide an option, what should this default to, depending on the OS level setting?
1. On looking at how other IDEs handle this, the official vim plugin for VS Code does not provide an option for VS Code to internally override it (remember, VS Code is also electron), but rather specifies in it's [README](https://github.com/VSCodeVim/Vim#mac). IntelliJ also does not provide any official settings for it, as several [SO](https://stackoverflow.com/a/15109812) answers suggest.

With the above reasons in mind, leaning especially on VS Code's/IntelliJ's approach, I suggest we don't programmatically disable this option, rather add an info icon with a link to documentation which describes to disable this option using the command:
```
defaults write com.insomnia.app ApplePressAndHoldEnabled -bool false 
```

image

develohpanda:
I have added documentation [here](https://support.insomnia.rest/article/203-key-maps): `https://support.insomnia.rest/article/203-key-maps`

Documentation in the app is defined [here](https://github.com/Kong/insomnia/blob/5f4c19da35e90ffa2f3ae27e9b309ee48f9d4e16/packages/insomnia-app/app/common/documentation.ts) for use throughout the app.

develohpanda (changes requested):
Following on from the reasons outlined in https://github.com/Kong/insomnia/pull/3476#issuecomment-861906734, I suggest we only link to documentation and have users change the setting themselves.

I have added documentation [here](https://support.insomnia.rest/article/203-key-maps): `https://support.insomnia.rest/article/203-key-maps`

Documentation links in the app are defined [here](https://github.com/Kong/insomnia/blob/5f4c19da35e90ffa2f3ae27e9b309ee48f9d4e16/packages/insomnia-app/app/common/documentation.ts) for use throughout the app.

reynolek:
So in my testing, I disabled the value. In Insomnia press and works. If I switch over to iMessages or my browsers, press and hold still brings up other accented letters.

But other apps like Discord it no longer works (is discord electron? I always forget if they transitioned away or not).

develohpanda:
> But other apps like Discord it no longer works (is discord electron? I always forget if they transitioned away or not).

@reynolek which command did you run?

In my testing, I ran `defaults write com.insomnia.app ApplePressAndHoldEnabled -bool false`, restarted both Insomnia and Discord - I see repeating characters in Insomnia and the special characters in Discord.

reynolek:
Ahh you know what, i did it globally 👍 

EDIT: To be more explicit, I ran `defaults write -g ApplePressAndHoldEnabled -bool false`

MrSnix:
As requested, I applied the changes.

Jun-16-2021 21-24-43

Feel free to modify the text content if there is a better alternative.
As always, I'm open to any feedback :D

develohpanda (approved):
Made a slight wording adjustment, but LGTM! 🎉 

image

dimitropoulos (approved):

PR #3480, opened 2021-06-16
have and use guards for all models

dimitropoulos:
this came up during a previous PR that only some models have guards, and even those with guards, the guards are not used consistently.

This PR aims to be laser focused on just organizing (and ensuring usage) of guards for all models.  Nothing else.  I tried very hard to have one commit per guard.  Thankfully, the vast majority of the line changes were simply import updates.


packages/insomnia-app/app/common/sorting.ts
```diff
@@ -9,10 +9,9 @@ import {
   SORT_TYPE_ASC,
   SORT_TYPE_DESC,
 } from './constants';
-import type { Request } from '../models/request';
-import type { GrpcRequest } from '../models/grpc-request';
-import type { RequestGroup } from '../models/request-group';
-import { isGrpcRequest, isRequest, isRequestGroup } from '../models/helpers/is-model';
+import { isRequest, Request } from '../models/request';
+import { GrpcRequest, isGrpcRequest } from '../models/grpc-request';
+import { isRequestGroup, RequestGroup } from '../models/request-group';
```
dimitropoulos, line 14: this particular diff, and many many like it in this PR, is a good example of why once you have types, you don't stand to gain a lot by creating index files (which, in the way I mean it, I would consider `is-modal` to have been a sort of index file for guards).

packages/insomnia-app/app/sync/vcs/ignore-keys.ts
```diff
@@ -1,6 +1,5 @@
 import { BaseModel } from '../../models';
-import { isWorkspace } from '../../models/helpers/is-model';
-import { Workspace } from '../../models/workspace';
+import { isWorkspace, Workspace } from '../../models/workspace';
```
dimitropoulos, line 3: This is another example of what I was talking about above regarding "index" files

dimitropoulos (commented):
Just finished a self-review.  Looks good.  If the build passes (which proves there are no messed up imports) really the only thing that could go wrong is if something like this got messed up:

```diff
-    const requestGroups = workspaceChildren.filter(d => d.type === models.requestGroup.type);
+    const requestGroups = workspaceChildren.filter(isRequestGroup);
```
like this (example of a bad change using `isRequest` instead of `isRequestGroup`):
```diff
-    const requestGroups = workspaceChildren.filter(d => d.type === models.requestGroup.type);
+    const requestGroups = workspaceChildren.filter(isRequest);
```

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -197,7 +199,7 @@ export const selectWorkspaceRequestsAndRequestGroups = createSelector(
   selectActiveWorkspaceEntities,
   entities => {
     return entities.filter(
-      e => e.type === models.request.type || e.type === models.requestGroup.type,
+      e => isRequest(e) || isRequestGroup(e),
```
develohpanda, line 202: This is where that helper from ramda would come in handy right, gets rid of the inline arrow function :D

develohpanda (approved):
Looks good, very excited to have this tidied up, it's been a long time coming!

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: yes, exactly, it'd be `entities.filter(anyPass([isRequest, isRequestGroup]))`

develohpanda: Wouldn't be opposed to pulling it in. Does it have an overhead like lodash? I tend to avoid lodash unless absolutely necessary

dimitropoulos: if you're talking about the bundle size, it's tree-shakable so I don't think the bundle impact is super relevant.  It's something like 50% less unpacked size than lodash anyway, but both are so tiny that I don't think it matters for our app.

PR #3483, opened 2021-06-17
Use safe state and reducer dispatch hooks

develohpanda:
On develop:

This PR:

* couldn't get `react-use` working
* converted the state management in `useRemoteWorkspaces` into a state-reducer - fewer state changes during refresh

Open to any feedback on this, not sure on the best solution (which doesn't involve adding support for `AbortController` to the VCS fetch and to the database implementation)

This whole unmounted state arises from the fact that we force refresh the app (via a key) when the active workspace changes. I'd like to decouple this in the future if we can.
https://github.com/Kong/insomnia/blob/47a8d1fa65a2b761e8ef62f65360fbd88b0073f8/packages/insomnia-app/app/ui/containers/app.tsx#L1502-L1506


reynolek (commented):
Im probably not qualified to review the code and say whether this the best approach, but I *do* love squashing random stacktraces. All about that. 👍

dimitropoulos (approved):

PR #3487, opened 2021-06-18
Don't create a workspace unless it is absolutely the last

develohpanda:
Closes INS-770

Previously, deleting the last workspace in a space would force create a workspace. But now, it will only force create a workspace when there are truly no workspaces present in _any_ space.

This is because the app requires at least one workspace to always exist (hoping to remove this requirement in the future)

https://www.loom.com/share/0f6cb6d9816249eeb00d24d90b2dbf6d


develohpanda:
Converting to draft, this is now actually redundant after https://github.com/Kong/insomnia/pull/3494. Will close PR once I've double checked everything

develohpanda:
Note: I plan to slightly modify this PR to remove the feature and keep it as a cleanup, because it converts a large dropdown component into FC and hooks

develohpanda:
Included in https://github.com/Kong/insomnia/pull/3494

PR #3492, opened 2021-06-22
Disable cert validation option while running tests in inso

develohpanda:
Cert validation is enabled by default.

This is settings initialization:
https://github.com/Kong/insomnia/blob/b5fe4357108f0e347c02918fc37edf381e2b16c7/packages/insomnia-app/app/models/settings.ts#L131-L131

This is settings being consumed by the networking layer:
https://github.com/Kong/insomnia/blob/c0412b1092a81dd569a6f2e0f619cf25f8ff73ed/packages/insomnia-app/app/network/network.ts#L420-L427

This PR provides a `--disableCertValidation` option to `inso run test`, the same way you can disable certificate validation in the UI through preferences.

Closes INS-773


wdawson:
Sweet! I'm not sure if we try to model the options after any other tools or not (e.g. `curl` has the `-k` or `--insecure` flags, which I think are very vague, and `wget` has the verbose `--no-check-certificate` flag). I think the name you've picked is good but it does make me wonder if it will just convert all my https tests to use http. It's not immediately obvious from the name alone that it will disable the certificate validation. That said, as curl proves, I don't think it has to be clear from the name alone. Anyway, just food for thought.

develohpanda:
> Sweet! I'm not sure if we try to model the options after any other tools or not (e.g. `curl` has the `-k` or `--insecure` flags, which I think are very vague, and `wget` has the verbose `--no-check-certificate` flag). I think the name you've picked is good but it does make me wonder if it will just convert all my https tests to use http. It's not immediately obvious from the name alone that it will disable the certificate validation. That said, as curl proves, I don't think it has to be clear from the name alone. Anyway, just food for thought.

Good point, maybe we could use a `--disableCertificateValidation` option. It's supposed to mirror this option from the Insomnia GUI preferences

image

(as an aside, not entirely sure why the inso options are not kebab-cased but there must be a reason that I can't remember off the top of my head)

wdawson:
> Good point, maybe we could use a `--disableCertificateValidation` option. It's supposed to mirror this option from the Insomnia GUI preferences

Yeah, great idea. Let's go with `--disableCertValidation` to keep it slightly shorter (reduced chance for typos).

develohpanda:
QA'd with the following workspace (import using clipboard)

<details>
<summary>Workspace</summary>

```
{"_type":"export","__export_format":4,"__export_date":"2021-07-28T22:44:22.242Z","__export_source":"insomnia.desktop.app:v2021.4.1-alpha.1","resources":[{"_id":"req_46e45d2a04c44a2996071375727221bd","parentId":"wrk_58b31c7c30164c6e9926f90e298f05e4","modified":1627511649497,"created":1627511641208,"url":"https://expired.badssl.com","name":"My Request","description":"","method":"GET","body":{},"parameters":[],"headers":[],"authentication":{},"metaSortKey":-1627511641208,"isPrivate":false,"settingStoreCookies":true,"settingSendCookies":true,"settingDisableRenderRequestBody":false,"settingEncodeUrl":true,"settingRebuildPath":true,"settingFollowRedirects":"global","_type":"request"},{"_id":"wrk_58b31c7c30164c6e9926f90e298f05e4","parentId":null,"modified":1627511636272,"created":1627511636272,"name":"PR#3492","description":"","scope":"design","_type":"workspace"},{"_id":"env_2ce142b81eae533fc9670b8f0448f09c86eca0b8","parentId":"wrk_58b31c7c30164c6e9926f90e298f05e4","modified":1627511636281,"created":1627511636281,"name":"Base Environment","data":{},"dataPropertyOrder":null,"color":null,"isPrivate":false,"metaSortKey":1627511636281,"_type":"environment"},{"_id":"jar_2ce142b81eae533fc9670b8f0448f09c86eca0b8","parentId":"wrk_58b31c7c30164c6e9926f90e298f05e4","modified":1627511636283,"created":1627511636283,"name":"Default Jar","cookies":[],"_type":"cookie_jar"},{"_id":"spc_65a55035795e4838aa03a38821c132b8","parentId":"wrk_58b31c7c30164c6e9926f90e298f05e4","modified":1627511636275,"created":1627511636275,"fileName":"PR#3492","contents":"","contentType":"yaml","_type":"api_spec"},{"_id":"uts_c041cc29e9ca4fea8e98229f01488487","parentId":"wrk_58b31c7c30164c6e9926f90e298f05e4","modified":1627511661965,"created":1627511661965,"name":"New Suite","_type":"unit_test_suite"},{"_id":"ut_e7575cb9b2234cb991c9e44badf730a0","parentId":"uts_c041cc29e9ca4fea8e98229f01488487","modified":1627511665144,"created":1627511663558,"requestId":"req_46e45d2a04c44a2996071375727221bd","name":"Returns 200","code":"const response1 = await insomnia.send();\nexpect(response1.status).to.equal(200);","_type":"unit_test"}]}
```

</details>

Demo showing the validate certs flag failing the request and failing the unit test



The test failing in inso without the certs flag (and succeeding with)

image

packages/insomnia-app/app/common/send-request.ts
```diff
@@ -1,12 +1,18 @@
 import { BaseModel, stats, types as modelTypes } from '../models';
+import * as models from '../models';
 import { getBodyBuffer } from '../models/response';
+import { Settings } from '../models/settings';
 import { send } from '../network/network';
 import * as plugins from '../plugins';
-import { database as db } from './database';
+import { database } from './database';
```
dimitropoulos, line 8: > let the floodgates open!

packages/insomnia-inso/README.md
```diff
@@ -151,6 +151,7 @@ The test runner is built on top of Mocha, thus many of the options behave as the
 | `--testNamePattern <regex>` | `-t` | run tests that match the regex|
 | `--bail` | `-b` | abort ("bail") after the first test failure|
 | `--keepFile` | | do not delete the generated test file (useful for debugging)|
+| `--disableCertValidation` | | disable certificate validation|
```
dimitropoulos, line 154: would it be more friendly to users viewing the help if we said:
```suggestion
| `--disableCertValidation` | | disable SSL certificate validation|
```

packages/insomnia-inso/src/cli.ts
```diff
@@ -64,6 +64,7 @@ const makeTestCommand = (commandCreator: CreateCommand) => {
     )
     .option('-b, --bail', 'abort ("bail") after first test failure')
     .option('--keepFile', 'do not delete the generated test file')
+    .option('--disableCertValidation', 'disable certificate validation')
```
dimitropoulos, line 67: would it be more friendly to users viewing the help if we said:

```suggestion
    .option('--disableCertValidation', 'disable SSL certificate validation')
```

dimitropoulos (commented):
checks out:

Screenshot_20210728_232402

dimitropoulos (commented):
My onnnnllly thing is that in `insomnia-app` we frame it in the positive (`validateSSL`) and in `inso` we frame it in the negative (`disableCertValidation`).

There are only 6 little places where we use `validateSSL`:



I think we should take just a tiny moment and consider that it will make for a more cohesive overall user experience between these two products if we unify these things.  Perhaps we switch gears and do `--validateSSL` and `--validateSSL=false` in `inso`, or perhaps we update the input (and corresponding language) in the app to be in the negative (i.e. so the UI would have a checkbox that had a label `Disable SSL Certificate Validation`.

@wdawson any thoughts on that?

I'm totttallly cool with whatever you all think - but just wanted to mention it.  I know it's a small thing at this moment, but I believe these kinds of things add up over time and I just want to keep the user experience at top quality wherever possible.

packages/insomnia-inso/src/cli.ts
develohpanda: Will defer to @wdawson on that!

develohpanda:
You're right, in the app it is positive but not because the setting itself is called `validateSSL`, it's because the text in preferences is positive (even though it's just a checkbox there).

image

I agree with and prefer positive & consistent conditions as well, but the slight difference with Inso is that we have defaults, and flags to _change_ those defaults, and IMO should avoid a breaking change. That is to say, if we have a default of true (positive), then the flag to switch of that default has to be false (negative). You might ask, why can't cert validation be off by default, and switched on by a flag?

That's because the app has certificate validation on by default as well, and the previous versions of inso have hard-coded to on, so we have to write a flag to switch them off otherwise we risk breaking CI pipelines.

For example, in the current live version `inso run test` has cert validation ON. If we were to make that opt-in/positive, then `inso run test` would have validation OFF, and now requires a user specified flag to preserve previous behavior - `inso run test --validateSSL`.

> or perhaps we update the input (and corresponding language) in the app to be in the negative (i.e. so the UI would have a checkbox that had a label `Disable SSL Certificate Validation`.

Keeping the above in mind, IMO this would be a better solution

DMarby (approved):

wdawson:
> My onnnnllly thing is that in insomnia-app we frame it in the positive (validateSSL) and in inso we frame it in the negative (disableCertValidation).

_Great_ points made by @dimitropoulos and @develohpanda above. I generally agree: consistency is better/preferable. Given the defaults we want in `inso` (validate on by default), it seems like we're stuck with a "negative" flag (`--disableCertValidation`) there. When looking at the options in Insomnia though, I have never seen a tool phrase an option like "Don't follow redirects." I think it's much more important to be consistent with things in the same context than in various tools you may provide. For that reason, I think we should keep things as they are and live with this inconsistency across `inso` and Insomnia.

develohpanda:
> I think it's much more important to be consistent with things in the same context than in various tools you may provide.

This is a fair point and I agree 🙌🏽

packages/insomnia-inso/src/cli.ts
develohpanda: I created a ticket to make sure we address it: INS-830

PR #3494, opened 2021-06-22
Remove the need for an active workspace to always exist

develohpanda:
The underlying approach is if we type the redux state correctly (avoiding `as` casts because those can still force errors), we should be able to flow down through each component and exit early in components which require an active workspace, thereby removing the need for an active workspace to always exist.

Removing the requirement for an active workspace will fix the bug described [here](https://github.com/Kong/insomnia/pull/3474#issuecomment-864891381), which arises as a result of the need for an active workspace to always exist.

No workspaces (thus, no active workspace)!
image

This PR is somewhat split up into decent commits but they start to fragment towards the end as I play whack-a-mole.

In addition:
- I updated the `DocumentCardDropdown` component into a functional component because it was done as part of another (now redundant) PR but the event handler changes from that PR were needed here.
- I fixed a bug related to the original report which existed within the VCS stage/unstage functions where they would not treat the input as immutable.

There is potentially some scope for additional unit tests but this is ready for code-review nonetheless.

QA:
- [x] status/stage/unstage/snapshot in insomnia sync (race conditions with fast stage/unstage?)
- [x] import/export behavior while in a workspace, or on the home page
- [x] git sync (checking for regressions, no changes)
- [x] quick switcher from the homepage or a workspace (cmd + p)

Closes INS-771


packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -78,15 +87,17 @@ export const selectWorkspacesForActiveSpace = createSelector(
...
+  (workspaces, activeWorkspaceId, activeActivity) => {
+    // Only return an active workspace if we're in an activity
+    if (activeActivity && isWorkspaceActivity(activeActivity)) {
+      const workspace = workspaces.find(w => w._id === activeWorkspaceId);
+      return workspace;
+    }
+
+    return undefined;
   },
 );
```
develohpanda, line 102: The update to this selector is what ultimately fixes the bug. The return type of this selector changes from `Workspace` to `Workspace | undefined`, and this waterfalls down through the App and Wrapper components

packages/insomnia-app/app/ui/components/modals/request-switcher-modal.tsx
```diff
@@ -146,8 +153,20 @@ class RequestSwitcherModal extends PureComponent<Props, State> {
...
+
+    let goToActivity: GlobalActivity;
+
+    if (activeActivity && isWorkspaceActivity(activeActivity)) {
+      goToActivity = activeActivity;
+    } else {
+      goToActivity = isDesign(workspace) ? ACTIVITY_SPEC : ACTIVITY_DEBUG;
+    }
+
+    this.props.handleSetActiveActivity(goToActivity);
```
develohpanda, line 167: Almost identical to https://github.com/Kong/insomnia/blob/611d73804f058dc59ef1b6adbbe64677aa974763/packages/insomnia-app/app/ui/components/wrapper-home.tsx#L158-L169

packages/insomnia-app/app/models/helpers/query-all-workspace-urls.ts
```diff
@@ -11,12 +11,11 @@ export const queryAllWorkspaceUrls = async (
   const docs = await db.withDescendants(workspace, reqType);
   const urls = docs
     .filter(
-      (d: any) =>
+      d =>
         d.type === reqType &&
         d._id !== reqId && // Not current request
-        (d.url || ''), // Only ones with non-empty URLs
+        ((d as Request | GrpcRequest).url || ''), // Only ones with non-empty URLs
```
dimitropoulos: instead of the cast, is it all the same if we do 
```ts
    .filter((d: Request | GrpcRequest) => (
      d.type === reqType &&
      d._id !== reqId && // Not current request
      (d.url || '') // Only ones with non-empty URLs
    ))
```
like the map below?

packages/insomnia-app/app/ui/components/wrapper-design.tsx
```diff
@@ -66,6 +72,12 @@ class WrapperDesign extends PureComponent<Props, State> {
       wrapperProps: { activeApiSpec },
       handleUpdateApiSpec,
     } = this.props;
+
+    if (!activeApiSpec) {
+      return;
+    }
+
+    // TODO: this seems strange, should the timeout be set and cleared on every change??
```
dimitropoulos, line 80: that's a good question.. hmm..

packages/insomnia-app/app/ui/components/wrapper.tsx
```diff
@@ -147,55 +124,14 @@ export interface WrapperProps {
...
-  activeUnitTestSuite: UnitTestSuite | null;
-  activeRequestResponses: Response[];
-  activeWorkspace: Workspace;
-  activeCookieJar: CookieJar;
-  activeEnvironment: Environment | null;
-  activeGitRepository: GitRepository | null;
-  activeUnitTestResult: UnitTestResult | null;
-  activeUnitTestSuites: UnitTestSuite[];
-  activeUnitTests: UnitTest[];
-  activeWorkspaceClientCertificates: ClientCertificate[];
```
dimitropoulos, line 188: :heart_eyes:

packages/insomnia-app/app/ui/components/wrapper.tsx
```diff
@@ -633,52 +587,53 @@ class Wrapper extends PureComponent<WrapperProps, State> {
...
-              editorLineWrapping={settings.editorLineWrapping}
-              handleRender={handleRender}
-              handleGetRenderContext={handleGetRenderContext}
-              nunjucksPowerUserMode={settings.nunjucksPowerUserMode}
-              handleRemoveWorkspace={this._handleRemoveActiveWorkspace}
-              handleDuplicateWorkspace={handleDuplicateWorkspace}
-              handleClearAllResponses={this._handleActiveWorkspaceClearAllResponses}
-              isVariableUncovered={isVariableUncovered}
-            />
+              {activeApiSpec ? <WorkspaceSettingsModal
```
dimitropoulos, line 619: formatting seems a little wonky here: but there are eslint rules for this in the react plugin.  we can loop back later.  they have autofixers.  all good.  finishing the typescript conversion is a much bigger priority in my mind (first and foremost)

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1464,15 +1459,13 @@ class App extends PureComponent<Props, State> {
...
-    this._ensureWorkspaceChildren(nextProps);
+  UNSAFE_componentWillReceiveProps() {
+    this._ensureWorkspaceChildren();
   }
 
   // eslint-disable-next-line camelcase
   UNSAFE_componentWillMount() {
-    // @ts-expect-error -- TSCONVERSION this function doesn't even accept props but should it?
-    this._ensureWorkspaceChildren(this.props);
+    this._ensureWorkspaceChildren();
```
dimitropoulos, line 1468: :+1:

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1492,16 +1485,15 @@ class App extends PureComponent<Props, State> {
       forceRefreshCounter,
       forceRefreshHeaderCounter,
     } = this.state;
-    const uniquenessKey = `${forceRefreshCounter}::${activeWorkspace._id}`;
+    const uniquenessKey = `${forceRefreshCounter}::${activeWorkspace?._id || 'n/a'}`;
```
dimitropoulos, line 1488: I'd love to hear an explanation of this from you.  from my reading, `'n/a/'` is a way to exit early and avoid a database lookup.  However, it looks like maybe `null` also serves the same purpose.  Why not use null?

And if there's a reason to keep `'n/a/'` here, it'd be cool maybe to put it in `constants.ts` with it's own TSDoc explaining what it is and what it's for.

packages/insomnia-app/app/ui/redux/modules/__tests__/helpers.test.ts
```diff
@@ -2,21 +2,26 @@ import { askToImportIntoWorkspace, ForceToWorkspaceKeys } from '../helpers';
...
   it('should return id if forcing to a current workspace', () => {
     const currentWorkspaceId = 'currentId';
-    const func = askToImportIntoWorkspace(currentWorkspaceId, ForceToWorkspaceKeys.current);
+    const func = askToImportIntoWorkspace({ workspaceId: currentWorkspaceId, forceToWorkspace: ForceToWorkspaceKeys.current });
     expect(func()).toBe(currentWorkspaceId);
   });
 
   it('should prompt the user if not forcing', () => {
-    modals.showModal = jest.fn();
+    (modals as any).showModal = jest.fn();
```
dimitropoulos: to me it's preferable here if we use a `ts-expect-error` and say that yes, while this is normally readonly, we are overwriting it in the test to mock things.  For that matter, should we just be using `jest.mock` instead of doing it this way?

packages/insomnia-app/app/ui/redux/sidebar-selectors.ts
```diff
@@ -77,7 +94,7 @@ export const selectSidebarChildren = createSelector(
         const hasMatchedChildren = child.children.find(c => c.hidden === false);
         // Try to match request attributes
         const name = child.doc.name;
-        const method = isGrpcRequest(child.doc) ? 'gRPC' : child.doc.method;
+        const method = isGrpcRequest(child.doc) ? 'gRPC' : isRequest(child.doc) ? child.doc.method : '';
```
dimitropoulos, line 97: just double checking, is the final fallback here (`''`) valid?

dimitropoulos (approved):
functionally tested and it seems to work great.  it's a really really good feeling to have a lot of these fixes in.  so powerful to be able to know all the places that needed to exit early.  it also feels good to not need to have an active workspace.

packages/insomnia-app/app/ui/containers/app.tsx
develohpanda: The `uniquenessKey` causes a force refresh of the app with this concatenation. We could do something like `[forceRefreshCounter, activeWorkspace?.id].filter(x => x).join('::')` to cover for the fact that a workspace is now optional.

The value `'n/a'` is non-significant

packages/insomnia-app/app/ui/redux/modules/__tests__/helpers.test.ts
develohpanda: Nice catch a558535727fb47dc8e4e2108c332bc154b2f0eac

packages/insomnia-app/app/ui/redux/sidebar-selectors.ts
develohpanda: Yeah it's valid, it's only used for fuzzy matching and a method is only relevant for a grpc request or a regular request

packages/insomnia-app/app/ui/components/wrapper.tsx
develohpanda: The rules would be nice too honestly, I feel like it's a small effort for a large gain, where the TS conversion is a larger effort so will take more time. In any case, formatting feedback I will always rely on an auto-fixer to control haha

dimitropoulos: yep, fo sho

packages/insomnia-app/app/ui/containers/app.tsx
dimitropoulos: oh, interesting.  ok.

packages/insomnia-app/app/ui/redux/modules/__tests__/helpers.test.ts
dimitropoulos: cool cool. thanks

packages/insomnia-app/app/ui/redux/sidebar-selectors.ts
dimitropoulos: cool

packages/insomnia-app/app/models/helpers/query-all-workspace-urls.ts
develohpanda: Consistency is king! 8555dc888df7f10e3ff80add66e08362d8ff77e1

packages/insomnia-app/app/ui/components/wrapper-design.tsx
develohpanda: I don't think it should be implemented the way it is, defeats the purpose. It's basically the same as my original observation on the `DebouncedTimeout` refactor PR

PR #3503, opened 2021-06-23
adds devtools shortcut

dimitropoulos:
Multiple times per day I hit F12 and nothing happens.  I've been doing this every day for a few months now, and finally had a moment (during a meeting I was listening in on) quickly add do it.

### DIFF
--- a/packages/insomnia-app/app/common/hotkeys.ts
+++ b/packages/insomnia-app/app/common/hotkeys.ts
@@ -86,7 +86,7 @@ export const hotKeyRefs = {
   SHOW_MONITOR: defineHotKey('activity_monitor', 'Show Monitor Activity'),
   SHOW_HOME: defineHotKey('activity_home', 'Show Home Activity'),
   FILTER_DOCUMENTS: defineHotKey('documents_filter', 'Focus Documents Filter'),
-  TOGGLE_DEVTOOLS: defineHotKey('toggle-devtools', 'Toggle Chrome DevTools'),
+  TOGGLE_DEVTOOLS: defineHotKey('toggle_devtools', 'Toggle Chrome DevTools'),
 };
 
 /**
@@ -385,16 +385,14 @@ const defaultRegistry: HotKeyRegistry = {
       { ctrl: true, keyCode: keyboardKeys.f.keyCode },
     ],
   },
-  ...(isDevelopment() ? {
-    [hotKeyRefs.TOGGLE_DEVTOOLS.id]: {
-      macKeys: [
-        { keyCode: keyboardKeys.f12.keyCode },
-      ],
-      winLinuxKeys: [
-        { keyCode: keyboardKeys.f12.keyCode },
-      ],
-    },
-  } : {}),
+  [hotKeyRefs.TOGGLE_DEVTOOLS.id]: {
+    macKeys: [
+      { keyCode: keyboardKeys.f12.keyCode },
+    ],
+    winLinuxKeys: [
+      { keyCode: keyboardKeys.f12.keyCode },
+    ],
+  },
 };
 
 const copyKeyCombs = (sources: KeyCombination[]) => sources.map(source => ({
@@ -433,6 +431,8 @@ export function newDefaultRegistry() {
  * Get the key combinations based on the current platform.
  */
 export function getPlatformKeyCombinations(bindings: KeyBindings) {
+  console.log(JSON.parse(JSON.stringify({ bindings })));
+
   if (isMac()) {
     return bindings.macKeys;
   }
--- a/packages/insomnia-app/app/common/hotkeys.ts
+++ b/packages/insomnia-app/app/common/hotkeys.ts
@@ -1,5 +1,5 @@
 import { keyboardKeys } from './keyboard-keys';
-import { ALT_SYM, CTRL_SYM, isMac, META_SYM, SHIFT_SYM } from './constants';
+import { ALT_SYM, CTRL_SYM, isDevelopment, isMac, META_SYM, SHIFT_SYM } from './constants';
 import { strings } from './strings';
 
 /**
@@ -135,6 +135,7 @@ export const hotKeyRefs: Record<string, HotKeyDefinition> = {
   SHOW_MONITOR: defineHotKey('activity_monitor', 'Show Monitor Activity'),
   SHOW_HOME: defineHotKey('activity_home', 'Show Home Activity'),
   FILTER_DOCUMENTS: defineHotKey('documents_filter', 'Focus Documents Filter'),
+  TOGGLE_DEVTOOLS: defineHotKey('toggle-devtools', 'Toggle Chrome DevTools'),
 };
 
 /**
@@ -293,6 +294,12 @@ const defaultRegistry: HotKeyRegistry = {
     keyComb(false, false, false, true, keyboardKeys.f.keyCode),
     keyComb(true, false, false, false, keyboardKeys.f.keyCode),
   ),
+  ...(isDevelopment() ? {
+    [hotKeyRefs.TOGGLE_DEVTOOLS.id]: keyBinds(
+      keyComb(false, false, false, false, keyboardKeys.f12.keyCode),
+      keyComb(false, false, false, false, keyboardKeys.f12.keyCode),
+    ),
+  } : {}),
 };
 
 function copyKeyCombs(sources: KeyCombination[]): KeyCombination[] {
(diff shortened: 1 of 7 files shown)


packages/insomnia-app/app/common/hotkeys.ts
```diff
@@ -135,181 +86,326 @@ export const hotKeyRefs: Record<string, HotKeyDefinition> = {
...
-  ),
-  [hotKeyRefs.SHOW_HOME.id]: keyBinds(
-    keyComb(false, false, true, true, keyboardKeys.h.keyCode),
-    keyComb(true, false, true, false, keyboardKeys.h.keyCode),
-  ),
-  [hotKeyRefs.FILTER_DOCUMENTS.id]: keyBinds(
-    keyComb(false, false, false, true, keyboardKeys.f.keyCode),
-    keyComb(true, false, false, false, keyboardKeys.f.keyCode),
-  ),
+  [hotKeyRefs.WORKSPACE_SHOW_SETTINGS.id]: {
```
develohpanda, line 96: Assuming/hoping there was nothing lost in translation here

packages/insomnia-app/app/common/hotkeys.ts
```diff
@@ -336,10 +431,8 @@ export function newDefaultRegistry(): HotKeyRegistry {
 
 /**
  * Get the key combinations based on the current platform.
- * @param bindings
- * @returns {Array<KeyCombination>}
  */
-export function getPlatformKeyCombinations(bindings: KeyBindings): KeyCombination[] {
+export function getPlatformKeyCombinations(bindings: KeyBindings) {
```
develohpanda, line 433: IMO it's generally good to _keep_ the return types where they exist or even in general. They can be redundant for documentation because TS can infer return types, but they also catch unexpected return types earlier (which I've been bitten by recently). Anyway not a blocker

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -284,10 +284,9 @@ export function createWindow() {
         },
       },
       {
-        label: `Toggle ${MNEMONIC_SYM}DevTools`,
+        label: `Toggle Chrome ${MNEMONIC_SYM}DevTools`,
```
develohpanda: Don't think this change is necessary 🤔

develohpanda (approved):

packages/insomnia-app/app/main/window-utils.ts
dimitropoulos: the rational is that there are at least two devtools now in the app (there weren't when this was originally made), the "redux devtools" and "chrome devtools".  An arguable third is react: react calls it "react developer tools" which is almost the same thing.

packages/insomnia-app/app/common/hotkeys.ts
dimitropoulos: I hear you.  The rationale is that where we're going (100% strict mode with no holes in internal types) it doesn't matter.  I know the bite of which you speak: but that shouldn't exist where we're going since, after all, if the return type of the function you call changes in such a way that is incompatible with your implementation then your implementation should absolutely error.

The reverse is also true (for example, I've given examples of this in the #typescript channel on slack) where we specify the return type but the internal type is something different and you can sort of short-circuit the type system and end up with runtime type errors (which, this will thankfully also be much better when we're in strict mode).

packages/insomnia-app/app/common/hotkeys.ts
dimitropoulos: I did this in the most systematic way I know: all at once with multicursor.  Therefore there was no manual translation.  Even still (and even with separating these changes into their own commit) I just checked again now by manually lining them up and they all looked 100% accurate to me.

dimitropoulos:
hey @develohpanda I know you approved it, but I want to just double check you're cool with my responses to your questions before I merge.  we're in no rush on these changes and it's not a high priority.

packages/insomnia-app/app/common/hotkeys.ts
develohpanda: > you can sort of short-circuit the type system and end up with runtime type errors

Could you elaborate on what you're thinking here for short circuiting - are you referring to casting? We're not quite there yet w.r.t. strict mode and the ideal "no-holes" scenario, so that bite will exist for the foreseeable future regardless of the approach which inherently makes it unsafe/tricky. 

> if the return type of the function you call changes in such a way that is incompatible with your implementation then your implementation should absolutely error

This depends on which direction you approach the problem from. "As a consumer I expect xyz to be returned", or "as a function, this is my input and output - my signature". The latter has a smaller surface area because you aren't reliant on consumers to tell you about errors in your function implementation (though consumers will error, defining the return type gives you a source of truth).

Coding to a function signature allows you to write the tests for expected return types and link it up in pseudocode, before writing up the actual implementation and then leaving TS to infer the return types. Think about the Dependency Injection approach, where you're _only_ working with interfaces (though I know we are not using DI per-se, this is just an example).

This comment explains my POV (and what has tripped me up a few times in our codebase) quite well: https://www.reddit.com/r/typescript/comments/ihpnme/opinion_on_explicit_return_types/g31pw1e

packages/insomnia-app/app/main/window-utils.ts
develohpanda: They're all embedded within the Chrome Developer Tools though, which still makes the "Chrome" distinction redundant. I don't think Microsoft Edge says "Chrome Developer Tools" even though it's built on Chromium (assumption, I don't have it installed to check)?

packages/insomnia-app/app/common/hotkeys.ts
develohpanda: (To be clear, not requesting any changes to this PR relating to this, only discussing 😄)

packages/insomnia-app/app/main/window-utils.ts
dimitropoulos: Not trying to be pedantic, but there's no tool called "Chrome Developer Tools" so far as I can tell (or am aware).

There's also no product called "DevTools", and that's all I was just trying to clarify since we (now) have 3 tools that are easily confusable via this (Chrome DevTools, Redux DevTools, React Developer Tools).

Maybe it will help elaborate on my mental process to hear that I was considering maybe in the future adding an item that will open the redux tools in a separate window (without opening the Chrome DevTools).  I, and I know other developers as well (since they built this functionality into the redux devtools by request and I have also worked with other developers that do this) like to work with the tools popped out as a separate window.  While I'll admit I wasn't _really_ thinking too deeply about this change, perhaps that's a motivation for why I think it's useful to clarify because to me an entry that says something like "toggle DevTools" is ambiguous.

re: Edge: I don't see what edge and chromium have to do with it since we're talking about the Chrome DevTools and not any other tool.  The Edge DevTools are not the same as the Chrome DevTools.  It's got a lot more tools and features, actually.  If we had the Edge DevTools as an option (and dang... you know what maybe we really could have both if we wanted?? not sure) wouldn't it make sense to clarify that in the menu item?

image

Totally happy to change it back, but it seemed like an uncontroversial change to me to make something clear that wasn't ambiguous when this line was first introduced, but is (to me) ambiguous now that we have 3 tools with very similar names in the app, yah know?

dimitropoulos: oh wow, haha, apparently we really can integrate _both_ the Chrome DevTools _and_ the Edge DevTools! 

https://youtu.be/jtGkiBJNbTQ

https://marketplace.visualstudio.com/items?itemName=ms-edgedevtools.vscode-edge-devtools

develohpanda: "Chrome Developer Tools" was merely me assuming _we_ had used DevTools as a short-form for Developer Tools, it's more the meaning behind it that I was referring to.

Perhaps I didn't explain clearly enough where the Edge comparison came in but it doesn't matter, the point I'm getting at is that Redux and React DevTools aren't available in production nor are they standalone, they exist as extensions tacked onto the Chrome DevTools embedded within Electron.

In Chrome, I have React and Redux DevTools installed but the line item to open simply says "Developer tools", and it's a similar story on Firefox.

image

image


As such I think we should leave Chrome off the title; to a user, these are developer tools available within the _Insomnia_ application.

dimitropoulos: ok, I'll make the change tomorrow

dimitropoulos: c901eb94e89c9eb14f7bd8464a28beec44207fc7

dimitropoulos:
I was doing a final QA pass before merging and found something but I don't have time to look into it more today so I'm pulling back into draft for now.

jackkav:
Closed as stale.

PR #3506, opened 2021-06-28
Improvements to theme selection

dimitropoulos:
This PR implements improvements to theme selection:

https://user-images.githubusercontent.com/15232461/123688471-089d1f00-d820-11eb-9473-930f85a34ff7.mp4

I have a few items I'd like to draw attention to:
1. There's some kind of thing happening with the SVG where a border appears around the badge because of the background svg.  I've tried using a border instead of a box shadow, z index hacks, moving the absolutely positioned badges up and beyond into the border.  For whatever reason it persists.  In this example I've manually changed the top background rect to have a black background to prove that it is indeed from the svg underneath.
Screenshot_20210628_144200
1. I didn't know how literal to take the design spec for the checkbox.  I left it as is for now, but there don't appear to be any other examples of checkboxes with these subtext labels in the app.  Would like to hear opinions on most insomnia-native approaches for this before I hand-code it.
1. @erickrico lemme know where I can grab the svg sun and moon icons you want to use.  I just did font-awesome for now - but it's super super simple to switch them out.
1. from a UX perspective, I think we should consider (and! this can happen in a future iteration!) some way of making it crystal-clear which theme is actively selected when they're in OS-priority mode.  Even a little thing that communicates "right now, we have detected your OS theme mode to be 'dark'" below the checkbox maybe could do the trick.


packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,389 @@
...
+      case 'default':
+      default:
+        patch = {
+          theme: themeName,
+        };
+        break;
+    }
+
+    applyColorScheme({
+      ...settings,
```
dimitropoulos: I was meaning to ask if this part is actually required?

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,389 @@
...
+  },
+}));
+
+interface Props {
+  activeTheme: PluginTheme['name'];
+  handleAutoDetectColorSchemeChange: (arg0: boolean) => void;
+  autoDetectColorScheme: boolean;
+  activeLightTheme: PluginTheme['name'];
+  activeDarkTheme: PluginTheme['name'];
+  settings: any; // TODO
```
dimitropoulos: it would be great to get a pointer on the correct type for this.  the parent component doesn't have it either.  is it as simple as `Settings` from the models/settings file?

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,389 @@
...
+import styled from 'styled-components';
+import { ColorScheme, getThemes } from '../../../plugins';
+import { applyColorScheme, PluginTheme } from '../../../plugins/misc';
+import HelpTooltip from '../help-tooltip';
+import * as models from '../../../models';
+import { RequireExactlyOne } from 'type-fest';
+
+const THEMES_PER_ROW = 5;
+
+type DarkOrLight = RequireExactlyOne<{ $isDark: boolean; $isLight: boolean }>
```
develohpanda: Why not use `isDark` and `!isDark`, or `theme: 'dark' | 'light'`?

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: What makes you feel as though it is not required? Looking at the code inside `applyColorScheme` it doesn't seem like something that can be removed without a refactor...

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: Looking at usages further down, and the waterfall input starting at https://github.com/Kong/insomnia/blob/c24d08ad663bf0ebd46d7e3bf10a2ae3b231553c/packages/insomnia-app/app/ui/components/wrapper.tsx#L645-L648, it appears to be the Settings database model

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,389 @@
...
+  settings,
+  handleAutoDetectColorSchemeChange,
+}) => {
+  const [themes, setThemes] = useState<PluginTheme[]>([]);
+
+  useEffect(() => {
+    getThemes().then(pluginThemes => {
+      setThemes(pluginThemes.map(({ theme }) => theme));
+    });
+  }, [activeDarkTheme, activeLightTheme, activeTheme, autoDetectColorScheme, settings]);
```
develohpanda: Does the entire settings object need to be a dependency for this effect? The other four options seem sufficient

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,389 @@
...
+      <rect x="90%" y="85%" width="5%" height="8%" style={{ fill: 'var(--color-info)' }} />
+    </g>
+  </svg>
+);
+
+const IndividualTheme: FC<{
+  isActive: boolean;
+  isDark: boolean;
+  isLight: boolean;
+  isInOsThemeMode: boolean;
```
develohpanda, line 220: These two (`isDark` and `isLight`) being individual props doesn't seem like the right approach. Could you explain why you chose this path?

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,389 @@
...
+          onChange={event => {
+            handleAutoDetectColorSchemeChange(event.target.checked);
+          }}
+        />
+      </label>
+    </div>
+  );
+
+  const onChangeTheme = useCallback((themeName: string, colorScheme: ColorScheme) => {
+    let patch;
```
develohpanda:
```suggestion
    let patch: Partial<Settings> = {};
```

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,389 @@
...
+    applyColorScheme({
+      ...settings,
+      ...patch,
+    }).then(() => {
+      models.settings.update(settings, patch).then(() => {
+        getThemes().then(pluginThemes => {
+          setThemes(pluginThemes.map(({ theme }) => theme));
+        });
+      });
+    });
```
develohpanda: Generally this code-base uses `async-await`; could you elaborate why you chose to use `then-catch` here and elsewhere?

develohpanda (changes requested):
Overall, I think it is possible to remove the `settings` prop entirely. None of the code-paths that I followed, that use the `settings` object, used anything _other than_ the properties extracted here:

image

Potentially, you could choose to define something like the following, and remove the duplicated props

```ts
type ThemeSettings = Pick<Settings, 'theme' | 'lightTheme' | 'darkTheme' | 'autoDetectColorScheme'>
```

A couple of questions and observations, so will hold off on approval.

-----

Now that this is converted to a FC, should you wish to, it's possible to remove props entirely and use `selectSettings` to fetch settings, and pull the theme change handler (`handleAutoDetectColorSchemeChange`) directly into `ThemePanel` because it is not used by anything else

develohpanda:
> 2. I didn't know how literal to take the design spec for the checkbox.  I left it as is for now, but there don't appear to be any other examples of checkboxes with these subtext labels in the app.  Would like to hear opinions on most insomnia-native approaches for this before I hand-code it.

Sorry I don't quite follow, it's the same as in the "General" tab in preferences.

image

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: because async is not supported with useEffect.  the error is:

> Warning: useEffect function must return a cleanup function or nothing. Promises and useEffect(async () => ...) are not supported, but you can call an async function inside an effect..

There are other options (e.g. https://github.com/facebook/react/issues/14326) but I thought this was the simplest, considering how short it is.  happy to change to the thing Dan suggested.

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: I know what you mean on gut reaction, and I tried to keep it "simple" but since in some situations there are actually three options (dark, light, default), it ends up just being the simplest to be able to say: "here is the style for when it's light" and not "here are the styles for when it's _not_ dark (i.e. is that always clear to the reader that that means light or could it be mistaken for a default?)".

If there were just two states for these UI elements, I'd definitely see the rationale for using `isDark` and `!isDark` or whathaveyou.  Doing it the way I have is "dumber" for sure (in my opinion), but it's also much more obvious - so much so that it seems best.

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: re the former: https://github.com/Kong/insomnia/pull/3506#discussion_r665408838

re the latter: I considered that, but all that does is push the logic into the style function, then I'd have to do `theme === 'dark' ?` in all of them instead of `isDark ?`.  That was the main reason I think.

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: image

Settings has like 47 properties, but these functions only actually look at 4 of them.  That's the reason it seems like it would work fine to just pass the updated part we care about.

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: thanks for confirming: updated in bb148dca312f52669f5ad2e25b75da350715d16a

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: the reason I thought it mattered is: what happens if a new theme is added?  This should re-render.  But it won't unless we put `settings` in there.

All of this squirrely logic with `getThemes` would be great to avoid (with a state selector maybe?) if you have any suggestions.  I was trying very very hard to restrict refactors to this file so I was trying to just work with what I had.

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: ahh, thanks, fixed in c916f6a401dc8da7d879342ee4f034eb4788fa24

dimitropoulos:
@develohpanda regarding https://github.com/Kong/insomnia/pull/3506#issuecomment-875181333 what I was asking about was the subtext part (the part that starts with "Pick your")

image

dimitropoulos:
re: https://github.com/Kong/insomnia/pull/3506#pullrequestreview-700471937, check out badf3c63014593998894dc9b6f4a9a91581a6bca and 575721556a46fe90ef11acaa5f7b50b0a8585d3f.

Good feedback, thanks!

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: Ah, sweet, yeah. Similar to the subset I have in the review description further down 👍

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: Noted, I'll think about this and get back to ya

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: Continuing with the latter, what about using `isLight(theme)` and `isDark(theme)` utility functions in the function? It seems uncommon to enforce logic/rules through props in this way, but this is coming from an ignorant place

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: I separated this comment from the earlier one because this usage (from what I can see) will allow me to specify _both_ `isDark` and `isLight` as true, which is why it didn't seem right.

But now that I look at this again; a single theme can be set as _both_ the light and dark theme selection, and that's what these props specify

https://github.com/dimitropoulos/insomnia/blob/575721556a46fe90ef11acaa5f7b50b0a8585d3f/packages/insomnia-app/app/ui/components/settings/theme-panel.tsx#L385-L386

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: > the reason I thought it mattered is: what happens if a new theme is added? This should re-render. But it won't unless we put settings in there.

Is this actually the case? Plugins and Themes exist on different tabs, which will force an update to occur regardless.



Even if you install via the Insomnia protocol (used by the insomnia.rest/plugins), you are forced into the plugins tab after installation (`insomniadev://plugins/install?name=insomnia-plugin-platypus-dark-theme`)



Although I wasn't able to reproduce this, fundamentally your observation is correct in that if the themes panel was open and at the same time a theme was installed in the background, this wouldn't update. An event of some sort needs to propagate through the app to say "themes have updated, go recheck", and that event (or dependency) isn't necessarily the `settings` object.

This is because the themes (plugins) themselves aren't stored in the `settings` object, [only whether they are enabled or not](https://github.com/Kong/insomnia/blob/b5fe4357108f0e347c02918fc37edf381e2b16c7/packages/insomnia-app/app/models/settings.ts#L50-L50). If you feel strongly about adding settings as a dependency here, I would only use `settings.pluginConfig` instead of the whole thing (or just remove it altogether).

In saying that, if I'm missing a step of how to reproduce this, let me know and I can check again.

Additionally, as I thin about this more, we shouldn't have the other dependencies there at all. This means _every time_ the theme selection is changed, Insomnia will go and load all themes again, which seems a bit excessive. What do you think?

As such I'd propose one of the following two options:

```suggestion
  }, [settings.pluginConfig]);  // only fetch themes when plugin config changes
```

```suggestion
  }, []); // only fetch themes on mount
```

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
develohpanda: But this particular one isn't in `useEffect`, it's in `useCallback`, which can be async, hence the question

For effects, I suggest we set a convention (through consistent implementation) of async effects using `async-await` instead of `try-catch`. Especially as more things move into hooks and async effects are needed, this will mean we have one format of working with promises (`async-await`) instead of two (`try-catch`)

Two existing examples from this codebase:

https://github.com/Kong/insomnia/blob/0c97fd7a88f56ff06f7003c5744f1accd8005d2d/packages/insomnia-app/app/ui/hooks/space.ts#L32-L35 

https://github.com/Kong/insomnia/blob/5f4c19da35e90ffa2f3ae27e9b309ee48f9d4e16/packages/insomnia-app/app/ui/components/panes/grpc-request-pane/use-proto-file-reload.ts#L20-L25

Not too concerned about any particular syntax (eg the two examples above are both different) ,but it's certainly jarring to keep switching between `async-await` and `try-catch`

develohpanda:
I'm confused where the subtext is coming from 🤔 on macOS on this PR I see the following, as a tooltip

image

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,394 @@
...
+
+  const isThemeActive = useCallback((theme: PluginTheme) => {
+    if (autoDetectColorScheme) {
+      return theme.name === activeLightTheme || theme.name === activeDarkTheme;
+    }
+    return theme.name === activeTheme;
+  }, [activeLightTheme, activeDarkTheme, activeTheme, autoDetectColorScheme]);
+
+  const onChangeAutoDetectColorScheme = useCallback((event: ChangeEvent<HTMLInputElement>) => {
+    const autoDetectColorScheme = event.target.checked;
```
develohpanda: It's very misleading to use the same variable in this scope while it is also used in a higher scope, please change it to be unique.

While reviewing this code I was wondering why this variable (which I know to exist in the FC scope) wasn't passed as a dependency to `useCallback`.

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,394 @@
...
+    applyColorScheme({
+      ...settings,
+      autoDetectColorScheme,
+    }).then(() => {
+      models.settings.update(settings, {
+        autoDetectColorScheme,
+      });
```
develohpanda: You could use `models.settings.patch` which removes the need for `settings` (thus a clean dependency array) - there will always only be one settings object, and `.patch` will update that one

develohpanda (changes requested):
A couple of things, to summarize:
- few questions about hook dependencies
- request to use `async-await` instead of `try-catch` (OS contributors have been guided in the past against `try-catch`, as is the convention in this code-base, so let's try stick with it)
- avoid using the same variables names in nested scopes
- is [this](https://github.com/Kong/insomnia/pull/3506#discussion_r664866560) something you are hoping to address in this PR, or leave for later?

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
dimitropoulos: ok, fixed up in a2c4a4d12d5edab339e1d13898f4f3c8e0957fa0 by using the "patch" technique used elsewhere (and before this PR)

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -0,0 +1,322 @@
...
+  isActive,
+  isDark,
+  isLight,
+  isInOsThemeMode,
+  onChangeTheme,
+  theme,
+}) => {
+  const { displayName, name } = theme;
+
+  const createChangeHandler = useCallback((mode: ColorScheme) => () => onChangeTheme(name, mode), [name, onChangeTheme]);
```
develohpanda: I renamed this callback to indicate better what it is actually doing. Previously, it was just labelled `onChangeTheme`, but it is actually a generator that returns a change handler. [`fed0a3c` (#3506)](https://github.com/Kong/insomnia/pull/3506/commits/fed0a3cc9dcd99f9c6d67f1f858ae66eb53478f2)

With that in mind, memoizing a generator doesn't really do a great deal. This is because we are only memoizing the generator function, not it's return value. In this case, we are memoizing the function that creates the change handler, not the change handler itself. Further down, when sending the result of `createChangeHandler` to the `OverlaySide` component as a prop, we are sending a new function on every render, not a memoized change handler.

Is there anything that can be done to improve this?

develohpanda (commented):
I made a couple of updates to move related logic together, it was scattered. There was also unnecessary deconstruction and aliasing (and then reversed aliasing) happening, which really shouldn't exist, so I fixed that.

All of the callbacks that exist in the custom hook (`useThemes`) which existed in the FC itself prior to the refactor, are somewhat unnecesary. Following through all off the code paths, it appears that those callbacks pretty much always change.

Something about the UX also seems a bit clunky compared to the live app, but I can't quite put my finger on it. If you run the two apps side by side, and switch themes (without the OS auto detect mode), do you notice anything? I made an attempt to improve this, looking at the previous implementation, to some degree of success: [`f31d506` (#3506)](https://github.com/Kong/insomnia/pull/3506/commits/f31d506604356b8ff89c5893f17ae7aaaa942aa5)

dimitropoulos:
@develohpanda I believe all feedback has been addressed.  please take another peek.

develohpanda (approved):

PR #3774, opened 2021-07-01
Force to design workspace when coming from git

develohpanda:
When cloning a git repository, any workspaces coming from it should always be design documents (`scope = 'design'`).

However, sometimes when a repository that was created from an early version of Insomnia Designer, the workspace might have `scope = null`, which [automatically migrates](https://github.com/Kong/insomnia/blob/611d73804f058dc59ef1b6adbbe64677aa974763/packages/insomnia-app/app/models/workspace.ts#L144-L165) (correctly) to a `collection`, which unfortunately disables all git functionality.

If we are coming from git then always force the scope to design, this is a safe assumption to make.

To test, I synced a repository from Insomnia, deleted the workspace from Insomnia, and in GitHub manually changed the scope to be null, then cloned again in Insomnia.

image

On the live app, after cloning and opening the cloned workspace, a user is presented with the following:
image

But they should be presented with the following, as per this PR:
image

This PR also will force the workspace to always be committed when viewed in the git staging modal, to make sure migrations are persisted. This is okay, because the workspace model itself has very few properties and it is safe to force any changes to be committed.

image

> Note, git sync only works in the base space for the time being, so do not test in a different space (INS-781 will address this)

### DIFF
--- a/packages/insomnia-app/app/common/migrate-from-designer.ts
+++ b/packages/insomnia-app/app/common/migrate-from-designer.ts
@@ -6,11 +6,10 @@ import * as models from '../models';
 import { database as db } from './database';
 import { getModelName } from '../models';
 import { difference } from 'lodash';
-import type { Workspace } from '../models/workspace';
 import type { Settings } from '../models/settings';
 import fsx from 'fs-extra';
 import { trackEvent } from './analytics';
-import { WorkspaceScopeKeys } from '../models/workspace';
+import { forceWorkspaceScopeToDesign } from '../sync/git/force-workspace-scope-to-design';
 
 async function loadDesignerDb(
   types: string[],
@@ -201,9 +200,7 @@ export default async function migrateFromDesigner({
 
       // For each workspace coming from Designer, mark workspace.scope as 'design'
       if (modelType === models.workspace.type) {
-        for (const workspace of entries) {
-          (workspace as Workspace).scope = WorkspaceScopeKeys.design;
-        }
+        entries.forEach(forceWorkspaceScopeToDesign);
       }
 
       const entryCount = entries.length;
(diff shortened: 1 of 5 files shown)


wdawson (approved):
LGTM. Thanks for the find ⛵

packages/insomnia-app/app/common/migrate-from-designer.ts
```diff
@@ -201,9 +200,7 @@ export default async function migrateFromDesigner({
 
       // For each workspace coming from Designer, mark workspace.scope as 'design'
       if (modelType === models.workspace.type) {
-        for (const workspace of entries) {
-          (workspace as Workspace).scope = WorkspaceScopeKeys.design;
-        }
+        entries.forEach(forceWorkspaceScopeToDesign);
```
dimitropoulos, line 203: thumb_pacha-italian-chef-pacha-edits-when-the-sun-53678014

packages/insomnia-app/app/sync/git/ne-db-client.ts
```diff
@@ -182,13 +173,13 @@ export class NeDBClient {
     }
 
     if (fileBuff) {
-      const doc = YAML.parse(fileBuff.toString());
+      const doc: BaseModel = YAML.parse(fileBuff.toString());
       return new Stat({
         type: 'file',
         mode: 0o777,
         size: fileBuff.length,
+        // @ts-expect-error should be number instead of string https://nodejs.org/api/fs.html#fs_stats_ino I think flow should have detected this
```
dimitropoulos: since we're touching this line: maybe change the part that mentions flow?

dimitropoulos (approved):
Before this PR:

Screenshot_20210702_134357


After this PR:
Screenshot_20210702_134959

PR #3775, opened 2021-07-01
git clone into the active space

develohpanda:
Similar to Insomnia Sync, we want to ignore the workspace parentId when syncing to and from git. This is because a git repository exists at the level of a single workspace, and a space exists one level above, which is not synced into git.

This means we have to:
a) reset the parentId to it's original value when writing into git
b) set the parentId to be the active space when reading from git and writing into nedb

See self-comments and code comments.

Initially, when a user would activate a space, then clone from git, that cloned workspace would end up in the base space always. [Demo](https://www.loom.com/share/ed0cc19d1ef3434d9a062a8a97879215)

Now, when a user activates a space, then clones from git, that cloned workspace will exist and remain in the space it was cloned into. [Demo](https://www.loom.com/share/49d2bf8b8c0f469e92b9861ad204903d)

Partially closes INS-781


packages/insomnia-app/app/sync/git/ne-db-client.ts
```diff
@@ -7,23 +7,27 @@ import { GIT_INSOMNIA_DIR_NAME } from './git-vcs';
 import parseGitPath from './parse-git-path';
 import { BufferEncoding } from './utils';
 import { SystemError } from './system-error';
+import { resetKeys } from '../vcs/ignore-keys';
```
develohpanda: Ideally move this into a common location so that "vcs" and "git" can both use the same `resetKeys` logic without invading into each others' domain

packages/insomnia-app/app/sync/git/ne-db-client.ts
```diff
@@ -90,6 +96,15 @@ export class NeDBClient {
...
     }
 
+    if (isWorkspace(doc)) {
+      console.log('[git] setting workspace parent to be that of the active space', { original: doc.parentId, new: this._spaceId });
+      // Whenever we write a workspace into nedb we should set the parentId to be that of the current space
+      // This is because the parentId is not synced into git, so it will be cleared whenever git writes the workspace into the db, thereby removing it from the space
+      // In order to reproduce this bug, comment out the following line, then clone a repository into a local space, then open the workspace
+      // @ts-expect-error parentId can be string or null for a workspace
+      doc.parentId = this._spaceId;
+    }
```
develohpanda: Force the workspace parentId to be that of the active space when writing from git into NeDB

packages/insomnia-app/app/ui/redux/modules/git.tsx
```diff
@@ -252,15 +256,18 @@ export const cloneGitRepository = ({ createFsClient }: {
                 const docPath = path.join(modelDir, docFileName);
                 const docYaml = await fsClient.promises.readFile(docPath);
                 const doc: BaseModel = YAML.parse(docYaml.toString());
+                if (isWorkspace(doc)) {
+                  // @ts-expect-error parentId can be string or null for a workspace
+                  doc.parentId = activeSpace?._id || null;
+                }
```
develohpanda, line 262: Set the workspace parentId to be that of the active space, on clone

packages/insomnia-app/app/sync/git/__tests__/ne-db-client.test.ts
```diff
@@ -139,6 +144,60 @@ describe('NeDBClient', () => {
...
+      // Assert
+      expect(upsertSpy).toHaveBeenCalledTimes(1);
+      expect(upsertSpy).toHaveBeenCalledWith(workspaceInDb, true);
+
+      // Cleanup
+      upsertSpy.mockRestore();
+    });
+
+    it('should force to a design workspace when writing', async () => {
+      // Assemble
```
dimitropoulos: `Arrange`?  Isn't it `Arrange`?  (doesn't matter, was just wondering if there was a reason you did this).

packages/insomnia-app/app/sync/git/ne-db-client.ts
```diff
@@ -90,6 +98,15 @@ export class NeDBClient {
       throw new Error(`Doc type does not match file path [${doc.type} != ${type || 'null'}]`);
     }
 
+    if (isWorkspace(doc)) {
+      console.log('[git] setting workspace parent to be that of the active space', { original: doc.parentId, new: this._spaceId });
+      // Whenever we write a workspace into nedb we should set the parentId to be that of the current space
+      // This is because the parentId (or a space) is not synced into git, so it will be cleared whenever git writes the workspace into the db, thereby removing it from the space on the client
+      // In order to reproduce this bug, comment out the following line, then clone a repository into a local space, then open the workspace, you'll notice it will have moved into the base space
+      // @ts-expect-error parentId can be string or null for a workspace
```
dimitropoulos, line 106: I think we should have a convention or something for the workarounds we need to do for orphaned workspaces so when we fix it later we can come back and fix them.  There are more than I can keep track of at this point.

packages/insomnia-app/app/sync/git/__tests__/ne-db-client.test.ts
develohpanda: Ah you're right 😆 it is arrange

packages/insomnia-app/app/sync/git/ne-db-client.ts
dimitropoulos: (we chatted about this - same page)

packages/insomnia-app/app/ui/redux/modules/__tests__/git.test.tsx
```diff
@@ -194,7 +203,7 @@ describe('git', () => {
     it('should fail if exception during clone and not try again if uri ends with .git', async () => {
       const memClient = MemClient.createClient();
       const err = new Error('some error');
-      (shallowClone as jest.Mock).mockRejectedValue(err);
+      shallowClone.mockRejectedValue(err);
```
dimitropoulos, line 206: nice to see all these working now without casts.

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -67,7 +68,7 @@ export const selectSpaces = createSelector(
 export const selectActiveSpace = createSelector(
   selectEntities,
   (state: RootState) => state.global.activeSpaceId,
-  (entities, activeSpaceId) => {
+  (entities, activeSpaceId): Space | undefined => {
     return entities.spaces[activeSpaceId || 'n/a'];
   },
```
dimitropoulos: oh.  I see.  Is what's happening here is that we're trying to work around this error?

image

If so, I strongly feel that we should do this in a much less tricky way, for example:

```suggestion
  (entities, activeSpaceId) => {
    return activeSpaceId ? entities.spaces[activeSpaceId] : undefined;
  },
```

which matches the type you specified.

image

The reason I find it most tricky is that there are a good few places in the app which actually use this value, `'n/a'`.  If this isn't one of them (and, I'm just now realizing it probably is not one of them, although until now that's how I've been reading this line when I saw it before), I think we should change this.

dimitropoulos (approved):
Approving on the assumption that if I was correct on https://github.com/Kong/insomnia/pull/3775/files#r671479673 that we'll fix it before merging or fix in a followup PR.  I don't want to block this whole PR on that alone.  Time to merge this puppy.

Otherwise, worked well when I functionally tested it.

packages/insomnia-app/app/ui/redux/selectors.ts
develohpanda: Great feedback and I agree; will improve this

develohpanda: `'n/a'` is used in a bunch of different places but it doesn't mean anything, it may as well be `''`. Ideally we get rid of them but in some cases it's nice to have an actual meaningless value than nothing at all

develohpanda: Fixed in [`3c614e6` (#3775)](https://github.com/Kong/insomnia/pull/3775/commits/3c614e6afb2d5317d90e53e62a6f451b1e565da0)

packages/insomnia-app/app/sync/git/__tests__/ne-db-client.test.ts
```diff
@@ -217,7 +217,7 @@ describe('NeDBClient', () => {
     });
 
     it('should throw error if type does not match', async () => {
-      // Assemble
+      // Arrange
```
dimitropoulos, line 220: This is also happening in `ne-db-clients.test.ts`.  Now seems as good a time as any to fix if you're waiting on this PR.

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -68,8 +67,8 @@ export const selectSpaces = createSelector(
 export const selectActiveSpace = createSelector(
   selectEntities,
   (state: RootState) => state.global.activeSpaceId,
-  (entities, activeSpaceId): Space | undefined => {
-    return entities.spaces[activeSpaceId || 'n/a'];
+  (entities, activeSpaceId) => {
+    return activeSpaceId ? entities.spaces[activeSpaceId] : undefined;
```
dimitropoulos, line 71: thanks.

PR #3776, opened 2021-07-01
Don't auto create a workspace if none exist

develohpanda:
Follow up to https://github.com/Kong/insomnia/pull/3494

This means we can start using correctly initialized redux state in unit tests (where possible, for now)

### DIFF
--- a/packages/insomnia-app/app/models/workspace.ts
+++ b/packages/insomnia-app/app/models/workspace.ts
@@ -1,7 +1,6 @@
 import type { BaseModel } from './index';
 import * as models from './index';
 import { database as db } from '../common/database';
-import { getAppName } from '../common/constants';
 import { strings } from '../common/strings';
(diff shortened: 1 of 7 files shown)


packages/insomnia-app/app/models/workspace.ts
```diff
@@ -70,18 +69,7 @@ export async function create(patch: Partial<Workspace> = {}) {
...
-    return workspaces;
-  }
-
-  // Create default workspace
-  await create({
-    name: getAppName(),
-    scope: WorkspaceScopeKeys.collection,
-  });
-  return db.all<Workspace>(type);
+  return await db.all<Workspace>(type);
```
dimitropoulos, line 72: NOICE

dimitropoulos (approved):
Screenshot_20210706_162809

Works great in a squeaky-clean environment.

PR #3779, opened 2021-07-01
adds eslint rule for requiring default case statements

dimitropoulos:
due to a recent bug that would have been caught by us having this rule.

short and sweet: there were only 7 places in the app where we needed this and they were all very straightforward.

### DIFF
--- a/packages/insomnia-app/app/common/select-file-or-folder.ts
+++ b/packages/insomnia-app/app/common/select-file-or-folder.ts
@@ -40,7 +40,7 @@ export const selectFileOrFolder = async ({ itemTypes, extensions }: Options) => 
           return 'openDirectory';
 
         default:
-          unreachableCase(type, `unrecognized file type: "${type}"`);
+          unreachableCase(type, `unrecognized item type: "${type}"`);
       }
     }),
     // @ts-expect-error https://github.com/electron/electron/pull/29322
--- a/packages/insomnia-app/app/common/select-file-or-folder.ts
+++ b/packages/insomnia-app/app/common/select-file-or-folder.ts
@@ -1,4 +1,5 @@
 import { OpenDialogOptions, remote } from 'electron';
+import { unreachableCase } from 'ts-assert-unreachable';
 
 interface Options {
   itemTypes?: ('file' | 'directory')[];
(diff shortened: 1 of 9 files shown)


packages/insomnia-app/app/common/select-file-or-folder.ts
```diff
@@ -37,6 +38,9 @@ export const selectFileOrFolder = async ({ itemTypes, extensions }: Options) =>
 
         case 'directory':
           return 'openDirectory';
+
+        default:
+          unreachableCase(type, `unrecognized file type: "${type}"`);
```
develohpanda:
```suggestion
          unreachableCase(type, `unrecognized item type: "${type}"`);
```

packages/insomnia-app/app/network/grpc/index.ts
```diff
@@ -170,6 +170,9 @@ export const start = async (
     case GrpcMethodTypeEnum.bidi:
       call = _makeBidiStreamRequest(requestParams);
       break;
+
+    default:
+      return;
   }
 
   if (!call) {
```
dimitropoulos, line 178: I debated on removing this `if-block`, but it looks like some of the above functions can return undefined so, for now at least, it's better to leave it I think.

packages/insomnia-app/app/ui/components/viewers/response-timeline-viewer.tsx
```diff
@@ -76,10 +76,9 @@ class ResponseTimelineViewer extends PureComponent<Props, State> {
       case 'TEXT':
         prefix = '* ';
         break;
-    }
 
-    if (prefix === null) {
-      return null;
+      default:
+        return null;
     }
```
develohpanda, line 82: Lol this is great

develohpanda (approved):
LGTM - should we hold off on merging this one so there are fewer conflicts for merging the hotfix branch into develop? Not sure, there's conflicts either way

packages/insomnia-app/app/network/grpc/index.ts
develohpanda: Sounds reasonable 👍🏽

packages/insomnia-app/app/ui/components/viewers/response-timeline-viewer.tsx
dimitropoulos: isn't it? haha

dimitropoulos:
it's written in such a way as to be line-for-line what's in the hotfix branch so it should't be a conflict.  and even if it is a conflict... I'd be happy to resolve it.

develohpanda:
Okay! Happy to merge then, we can cross that bridge when we get to it

PR #3782, opened 2021-07-02
Feat/eslint import order

dimitropoulos:
Lately, we've been having a lot of merge conflicts that happened only because of imports.  This PR addresses that in an automatic and systematic way by introducing `eslint-plugin-simple-import-sort`.

Despite being a lot of changes, there is actually only [one](d0730a861e69c764d756d3ce2d0cbe340d6279dd) non-config code file touched in this PR (which is very minor).  The rest was automatic by running `npm run lint:fix`.

The sequence of events was as follows:
1. I added `import/order` using the defaults
1. I saw that it wasn't alphabetizing.  We want that.  It defaults to false so I turned it on.
1. I saw that `import/order` doesn't handle side effect imports by design.  Ok.  Welp.  I turned on an option an option from the docs `warnOnUnassignedImports` to address this.
1. The option wasn't available because we were using an out of date version of the plugin.  As an experiment to see how bad it would be: I updated every eslint-related dependency and ran `lint:fix` again.  There were no changes and only one line reporting a new error! ok cool!
1. I addressed that in d0730a861e69c764d756d3ce2d0cbe340d6279dd
1. While doing so I noticed that we have included the `semistandard` config but it doesn't actually change the rules.  See the commit description on c705ff4df011cef88dfa4e295e27dd49fb050618 to see how I found that.  I removed it since it's unnecessary but also since eventually I can see us wanting to have our own `eslint-config-insomnia` anyway and in that case we wouldn't want to depend on any other config.
1. Now I was able to add the `import/order` rule.  I noticed a few problems:
  1. it didn't fully sort some files with side effect imports (again this is by design) so I fixed that in fc8daffc667ace87139a472c5a5b99b93c0a33c8
  1. some files required running the rule multiple times.  This seemed like a bug to me.  I reported in https://github.com/benmosher/eslint-plugin-import/issues/2143 with an example of one file that required being run 6 times (!!).
1. I turned off `import/order` and turned on `simple-import-sort/sort-imports`.
1. I reverted the two commits from `import/order` (since, I thought it might be interesting to the reviewer instead of just pruning it from git) and ran `lint:fix` with `simple-import-sort`. It worked great the first time (from what I can tell).

I know that sounds like a lot: but all of the above happened pretty quickly.

No rush on this PR.  We can always drop the commits that touch code and re-run since there are (aside from exactly 5 characters on one line) no manual changes in this PR.

@develohpanda feel free to merge whenever you feel if you approve.  Didn't want to make it auto-merge in case there's some timing with the release that's more convenient.


.eslintrc.js
```diff
@@ -27,6 +26,8 @@ module.exports = {
     'json',
     'filenames',
     'react-hooks',
+    'import',
```
develohpanda, line 29: Can we remove this `import` plugin now that we're using `simple-import-sort`?

develohpanda (approved):

develohpanda (changes requested):
I even did a hard-reset, but it looks like spacing is a bit off :(

image

dimitropoulos:
cool. It only took like 90 seconds to get to the bottom of that one, it was because of a change in `index.tsx.`  I tried to get to the bottom of it quickly, but I didn't want to take too much time on it.  I did discover that it breaks when the css is imported _before_ App.  I figure we can loop back another time if so.

Ultimately: these are the files that have changes (from the linter) that have potential for logical differences:

Screenshot_20210722_173237

develohpanda (approved):
I briefly QA'd this; spec editor, templating and & sending requests (which all go through code-mirror) and it seems ok to me. We'll just need to keep a lookout for things

PR #3787, opened 2021-07-06
Import into the active space

develohpanda:
Because there is conversation about removing import/export from the Data tab in the context-less/global Preferences, I opted to _not_ prompt the user for which space they want to import into, rather sticking with the approach of always importing into the currently active space (the same way we export the "current" document)

The only exception to this "import into the active space" rule, is that if a workspace already exists and the import is simply updating the existing workspace, we retain the space the workspace exists in.

This PR removes a lot of the prop-drilling and ensures each "import" code-path references the redux action directly so it is easier to trace. The first challenge with this PR is making sure all "import" code-paths work as expected. Every code path that goes through the redux actions will also, automatically, be sent the active space [ref](https://github.com/Kong/insomnia/blob/616b6991236b3cacbfeec8c36c8dbdab35f582bf/packages/insomnia-app/app/ui/redux/modules/import.ts#L45-L46)

* When exporting, we will now reset keys that shouldn't be sync'd exported (same code-path as Insomnia sync)
* When importing one/multiple workspaces:
  * if the workspace already exists, retain the existing parentId/space it is in and run import
  * if the workspace doesn't already exist, create it in the currently active space

The various ways in which you can import are:
* drag and drop an export file into the client window [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/containers/app.tsx#L1369-L1401)
* from the placeholder request pane (new workspace, no requests) [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/panes/placeholder-request-pane.tsx#L64-L66)
* from the create dropdown on the dashboard [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper-home.tsx#L311-L322)
* from the onboarding screen (`window.localStorage.setItem('insomnia::meta::activity', 'onboarding');` then refresh) [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper-onboarding.tsx#L129-L134)
* from the Insomnia protocol https://github.com/Kong/insomnia/pull/3407 [ref](https://github.com/Kong/insomnia/blob/b47e11ef18bd854dee5e879c8bdeb4d29b1524d3/packages/insomnia-app/app/ui/redux/modules/global.tsx#L183-L194)
* from Preferences > Data [ref](https://github.com/Kong/insomnia/blob/acee8978c11793c125567dcdb031576427635df7/packages/insomnia-app/app/ui/components/settings/import-export.tsx#L91-L102)
* when moving from design to debug [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper.tsx#L275-L284)
* from pasting in the URL bar [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper.tsx#L203-L231)

Demos:
- Import an API spec in various ways, notice how it always stays in the space that was created [demo](https://www.loom.com/share/41aea4808a11459487fcf656b9c74314)
- Export, delete and import multiple into a space [demo](https://www.loom.com/share/e247170b628b430290f49d10df2ea82a)
- Export and import without deleting, so the Workspaces already exist in another space (they should not be moved) [demo](https://www.loom.com/share/6a1c7ec72af344b59d8a12b28e5176e1)
- Import using insomnia protocol (you can experiment with `insomniadev://app/import?uri=https://gist.githubusercontent.com/dimitropoulos/64fdb4a75a5bd68998dbe1cae5cf71e4/raw/3eb58fce4363ad0e5a0d273176956932e21a64d1/multiple-chain.json`) [demo](https://www.loom.com/share/976785c1a03042bab1f3ac53e3a19d3b)

Closes INS-724 and partially closes INS-781

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper-onboarding.tsx
+++ b/packages/insomnia-app/app/ui/components/wrapper-onboarding.tsx
@@ -4,17 +4,18 @@ import 'swagger-ui-react/swagger-ui.css';
 import { showPrompt } from './modals';
 import type { BaseModel } from '../../models';
 import { AUTOBIND_CFG, getAppLongName, getAppName, getAppSynopsis } from '../../common/constants';
-import type { HandleImportFileCallback, HandleImportUriCallback, WrapperProps } from './wrapper';
+import type { WrapperProps } from './wrapper';
 import { database as db } from '../../common/database';
-import { ForceToWorkspaceKeys } from '../redux/modules/helpers';
 import OnboardingContainer from './onboarding-container';
 import { isWorkspace, WorkspaceScopeKeys } from '../../models/workspace';
 import Analytics from './analytics';
+import { bindActionCreators } from 'redux';
+import { importFile, importUri } from '../redux/modules/import';
+import { connect } from 'react-redux';
+import { ForceToWorkspace } from '../redux/modules/helpers';
 
-interface Props {
+interface Props extends ReturnType<typeof mapDispatchToProps> {
   wrapperProps: WrapperProps;
-  handleImportFile: HandleImportFileCallback;
-  handleImportUri: HandleImportUriCallback;
 }
 
 interface State {
@@ -64,9 +65,8 @@ class WrapperOnboarding extends PureComponent<Props, State> {
   }
 
   _handleImportFile() {
-    const { handleImportFile } = this.props;
-    handleImportFile({
-      forceToWorkspace: ForceToWorkspaceKeys.new,
+    this.props.handleImportFile({
+      forceToWorkspace: ForceToWorkspace.new,
       forceToScope: WorkspaceScopeKeys.design,
     });
   }
@@ -80,7 +80,7 @@ class WrapperOnboarding extends PureComponent<Props, State> {
       label: 'URI to Import',
       onComplete: value => {
         handleImportUri(value, {
-          forceToWorkspace: ForceToWorkspaceKeys.new,
+          forceToWorkspace: ForceToWorkspace.new,
           forceToScope: WorkspaceScopeKeys.design,
         });
       },
@@ -160,4 +160,16 @@ class WrapperOnboarding extends PureComponent<Props, State> {
   }
 }
 
-export default WrapperOnboarding;
+const mapDispatchToProps = (dispatch) => {
+  const bound = bindActionCreators({
+    importFile,
+    importUri,
+  }, dispatch);
+
+  return ({
+    handleImportFile: bound.importFile,
+    handleImportUri: bound.importUri,
+  });
+};
+
+export default connect(null, mapDispatchToProps)(WrapperOnboarding);
--- a/packages/insomnia-app/app/ui/components/wrapper.tsx
+++ b/packages/insomnia-app/app/ui/components/wrapper.tsx
@@ -79,7 +79,6 @@ import type { GlobalActivity } from '../../common/constants';
 import ProtoFilesModal from './modals/proto-files-modal';
 import { GrpcDispatchModalWrapper } from '../context/grpc';
 import WrapperMigration from './wrapper-migration';
-import type { ImportOptions } from '../redux/modules/global';
 import WrapperAnalytics from './wrapper-analytics';
 import { HandleGetRenderContext, HandleRender } from '../../common/render';
 import { RequestGroup } from '../../models/request-group';
@@ -133,10 +132,6 @@ export type WrapperProps = AppProps & {
   vcs: VCS | null;
   gitVCS: GitVCS | null;
 }
-
-export type HandleImportFileCallback = (options?: ImportOptions) => void;
-export type HandleImportClipboardCallback = (options?: ImportOptions) => void;
-export type HandleImportUriCallback = (uri: string, options?: ImportOptions) => void;
 
 interface State {
   forceRefreshKey: number;
@@ -300,18 +295,6 @@ class Wrapper extends PureComponent<WrapperProps, State> {
 
   _handleUpdateSettingsUseBulkParametersEditor(useBulkParametersEditor: boolean) {
     return models.settings.update(this.props.settings, { useBulkParametersEditor });
-  }
-
-  _handleImportFile(options?: ImportOptions) {
-    this.props.handleImportFileToWorkspace({ workspaceId: this.props.activeWorkspace?._id, ...options });
-  }
-
-  _handleImportUri(uri: string, options?: ImportOptions) {
-    this.props.handleImportUriToWorkspace(uri, { workspaceId: this.props.activeWorkspace?._id, ...options });
-  }
-
-  _handleImportClipBoard(options?: ImportOptions) {
-    this.props.handleImportClipBoardToWorkspace({ workspaceId: this.props.activeWorkspace?._id, ...options });
   }
 
   _handleSetActiveResponse(responseId: string | null) {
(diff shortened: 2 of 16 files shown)


wdawson (commented):
> The only exception to this "import into the active space" rule, is that if a workspace already exists and the import is simply updating the existing workspace, we retain the space the workspace exists in.

This is strange to me. As a user, I'm in the context of "another space" and the app just told me that it successfully imported some workspaces, but it's not showing up on my screen anywhere. Putting myself in a user's shoes, this flow doesn't seem very intuitive. Perhaps it would be better to warn them and say something like, "We found these collections/designs in other spaces (list them). Are you sure you want to import them into the FooBar space?" where FooBar is the current space. It seems to me that if a user wanted to update existing workspaces in a different space, they wouldn't have a problem switching to that space and working there.

> Import using insomnia protocol

I think we're going to need to ask which space you want to import this into if and only if you have multiple spaces. My reasoning is that if you have multiple spaces and you're not working in the app, just browsing the web, you stumble across something that lets you import insomnia collections. You click that button but it imports something into your team's production Space that you were working in yesterday instead of your personal debugging/testing space. Not good.

packages/insomnia-app/app/ui/redux/modules/import.ts
```diff
@@ -0,0 +1,145 @@
+import electron, { OpenDialogOptions } from 'electron';
```
dimitropoulos, line 1: re: moving these: nice.  `global` is so jam-packed.

packages/insomnia-app/app/ui/redux/modules/index.ts
```diff
@@ -44,3 +44,5 @@ export interface RootState {
   entities: entities.EntitiesState
   global: global.GlobalState,
 }
+
+export type GetState = () => RootState;
```
dimitropoulos, line 48: is it too early to do something like the docs suggest: 
> export type RootState = ReturnType<typeof store.getState>

so, in this case, it'd just be
> export type GetState = typeof store.getState

packages/insomnia-app/app/ui/redux/modules/import.ts
```diff
@@ -0,0 +1,145 @@
...
+    });
+  } finally {
+    dispatch(loadStop());
+  }
+};
+
+export const importUri = (
+  uri: string,
+  options: ImportOptions = {},
+) => async (dispatch: Dispatch, getState: GetState) => {
```
dimitropoulos, line 131: is there a way we can use `ThunkAction` from redux-thunk directly?  Apparently `Dispatch` as typed here is wrong since the actual type is `ThunkDispatch` which is

```ts
export interface ThunkDispatch<S, E, A extends Action> {
  <T extends A>(action: T): T;
  <R>(asyncAction: ThunkAction<R, S, E, A>): R;
}
```

while `Dispatch` (from `redux`) is

```ts
export interface Dispatch<A extends Action = AnyAction> {
  <T extends A>(action: T): T
}
```

packages/insomnia-app/app/ui/components/wrapper.tsx
```diff
@@ -302,18 +297,6 @@ class Wrapper extends PureComponent<WrapperProps, State> {
...
-  }
-
-  _handleImportUri(uri: string, options?: ImportOptions) {
-    this.props.handleImportUriToWorkspace(uri, { workspaceId: this.props.activeWorkspace?._id, ...options });
-  }
-
-  _handleImportClipBoard(options?: ImportOptions) {
-    this.props.handleImportClipBoardToWorkspace({ workspaceId: this.props.activeWorkspace?._id, ...options });
-  }
-
```
dimitropoulos, line 316: > what are... what are these tears in my eyes doing here.. they're.... they're tears of joy!

packages/insomnia-app/app/ui/components/wrapper-onboarding.tsx
```diff
@@ -4,17 +4,18 @@ import 'swagger-ui-react/swagger-ui.css';
...
 import OnboardingContainer from './onboarding-container';
 import { isWorkspace, WorkspaceScopeKeys } from '../../models/workspace';
 import Analytics from './analytics';
+import { bindActionCreators } from 'redux';
+import { importFile, importUri } from '../redux/modules/import';
+import { connect } from 'react-redux';
+import { ForceToWorkspace } from '../redux/modules/helpers';
 
-interface Props {
+interface Props extends ReturnType<typeof mapDispatchToProps> {
```
dimitropoulos, line 17: no biggie or anything or whatever, but it'd be cool to have some kind of convention we stick to, e.g. doing

```ts
type ReduxProps = ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps>
```
Or something like that.  I mean... on the other hand, we are going to be in hooks for each and every component soon enough so maybe it doesn't matter since the selectors we'd then be using will handle all of that for us.  Don't bother changing it here or whatever, just sorta thinking out loud while it's in an diff in front of me.

packages/insomnia-app/app/ui/redux/modules/import.ts
develohpanda: Yeah it is, and gets hard to follow. I want to move the export ones too

packages/insomnia-app/app/ui/redux/modules/index.ts
develohpanda: Once I do that, all the selectors see this error, which starts at `configureStore`

image

```ts
export type GetState = ReturnType<typeof configureStore>['getState'];
export type RootState = ReturnType<GetState>;
```

develohpanda: I have a feeling if we set up the store using `redux-toolkit` we won't face this

develohpanda: Going to leave as-is for now because it produces all these errors 😢

packages/insomnia-app/app/ui/redux/modules/import.ts
develohpanda: Looks like if we explicitly type the `ThunkAction` return type (instead of `ThunkDispatch`), it works too (and automatically types `getState` and any extra parameters). [`b61a581` (#3795)](https://github.com/Kong/insomnia/pull/3795/commits/b61a5813a82c0b7069b64eb06e824869df3eaad6)

packages/insomnia-app/app/ui/components/wrapper-onboarding.tsx
develohpanda: I like that convention! [`8ad40fa` (#3795)](https://github.com/Kong/insomnia/pull/3795/commits/8ad40facd906393be6a347d88ecf6b24284a0490)

packages/insomnia-app/app/ui/redux/modules/index.ts
dimitropoulos: cool cool. all good by me. just was mentioning it as I saw it.

this is an absolutely clear and strong step in the right direction - no complaints from me!

PR #3791, opened 2021-07-06
Allow removing header in default header plugin

ThisNoName:
Default header is global, it might be useful allow removing header using folder environment.  For example, most requests require token, but certain section may not.


plugins/insomnia-plugin-default-headers/src/default-headers.js
```diff
@@ -12,7 +12,12 @@ module.exports = function(context) {
       continue;
     }
 
-    context.request.setHeader(name, value);
-    console.log(`[header] Set default header ${name}: ${value}`);
+    if (value==="null") {
+      context.request.setHeader(name, null);
```
develohpanda: Might be better to use [`removeHeader`](https://github.com/Kong/insomnia/blob/0178533e5284cbcaff2c39a4ba2d560b16422e4c/packages/insomnia-app/app/plugins/context/request.ts) here

```suggestion
      context.request.removeHeader(name);
```

develohpanda (commented):
Thank you for the PR and welcome to the project! 🥳 

Looks reasonable to me, just one comment. Could you please update the [README](https://github.com/Kong/insomnia/blob/549ce23ce8cf202d787ca4df5efc5f25e03311ed/plugins/insomnia-plugin-default-headers/README.md) as well?

plugins/insomnia-plugin-default-headers/README.md
```diff
@@ -21,4 +21,11 @@ Headers can be added by setting a `DEFAULT_HEADERS` environment variable.
 	}
 }
 ```
-
+Default header can be removed by setting value to null. For example, use folder environment variables to remove authorization header from anonymous calls
+```
```
dimitropoulos:
```suggestion
```json
```

dimitropoulos (approved):

develohpanda (approved):

PR #3795, opened 2021-07-07
Import into the active space

develohpanda:
Continuation from #3787 because that was somehow merged (reverted now but the PR cannot be reopened)

Because there is conversation about removing import/export from the Data tab in the context-less/global Preferences, I opted to _not_ prompt the user for which space they want to import into, rather sticking with the approach of always importing into the currently active space (the same way we export the "current" document)

The only exception to this "import into the active space" rule, is that if a workspace already exists and the import is simply updating the existing workspace, we retain the space the workspace exists in.

This PR removes a lot of the prop-drilling and ensures each "import" code-path references the redux action directly so it is easier to trace. The first challenge with this PR is making sure all "import" code-paths work as expected. Every code path that goes through the redux actions will also, automatically, be sent the active space [ref](https://github.com/Kong/insomnia/blob/616b6991236b3cacbfeec8c36c8dbdab35f582bf/packages/insomnia-app/app/ui/redux/modules/import.ts#L45-L46)

* When exporting, we will now reset keys that shouldn't be exported (same code-path as Insomnia sync)
* When importing one/multiple workspaces:
  * if the workspace already exists, retain the existing parentId/space it is in and run import
  * if the workspace doesn't already exist, create it in the currently active space

The various ways in which you can import are:
* drag and drop an export file into the client window [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/containers/app.tsx#L1369-L1401)
* from the placeholder request pane (new workspace, no requests) [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/panes/placeholder-request-pane.tsx#L64-L66)
* from the create dropdown on the dashboard [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper-home.tsx#L311-L322)
* from the onboarding screen (`window.localStorage.setItem('insomnia::meta::activity', 'onboarding');` then refresh) [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper-onboarding.tsx#L129-L134)
* from the Insomnia protocol https://github.com/Kong/insomnia/pull/3407 [ref](https://github.com/Kong/insomnia/blob/b47e11ef18bd854dee5e879c8bdeb4d29b1524d3/packages/insomnia-app/app/ui/redux/modules/global.tsx#L183-L194)
* from Preferences > Data [ref](https://github.com/Kong/insomnia/blob/acee8978c11793c125567dcdb031576427635df7/packages/insomnia-app/app/ui/components/settings/import-export.tsx#L91-L102)
* when moving from design to debug [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper.tsx#L275-L284)
* from pasting in the URL bar [ref](https://github.com/Kong/insomnia/blob/e498b0e9458996fae88ea928a43dcd3d215a3f68/packages/insomnia-app/app/ui/components/wrapper.tsx#L203-L231)

Demos:
- Import an API spec in various ways, notice how it always stays in the space that was created [demo](https://www.loom.com/share/41aea4808a11459487fcf656b9c74314)
- Export, delete and import multiple into a space [demo](https://www.loom.com/share/e247170b628b430290f49d10df2ea82a)
- Export and import without deleting, so the Workspaces already exist in another space (they should not be moved) [demo](https://www.loom.com/share/6a1c7ec72af344b59d8a12b28e5176e1)
- Import using insomnia protocol (you can experiment with `insomniadev://app/import?uri=https://gist.githubusercontent.com/dimitropoulos/64fdb4a75a5bd68998dbe1cae5cf71e4/raw/3eb58fce4363ad0e5a0d273176956932e21a64d1/multiple-chain.json`) [demo](https://www.loom.com/share/976785c1a03042bab1f3ac53e3a19d3b)

Closes INS-724 and partially closes INS-781

### DIFF
--- a/packages/insomnia-app/app/common/export.ts
+++ b/packages/insomnia-app/app/common/export.ts
@@ -31,6 +31,7 @@ import { isCookieJar } from '../models/cookie-jar';
 import { isEnvironment } from '../models/environment';
 import { isUnitTestSuite } from '../models/unit-test-suite';
 import { isUnitTest } from '../models/unit-test';
+import { resetKeys } from '../sync/vcs/ignore-keys';
 
 const EXPORT_FORMAT = 4;
 
@@ -210,6 +211,8 @@ export async function exportRequestsData(
       if (isWorkspace(d)) {
         // @ts-expect-error -- TSCONVERSION maybe this needs to be added to the upstream type?
         d._type = EXPORT_TYPE_WORKSPACE;
+        // reset the parentId of a workspace
+        resetKeys(d);
       } else if (isCookieJar(d)) {
         // @ts-expect-error -- TSCONVERSION maybe this needs to be added to the upstream type?
         d._type = EXPORT_TYPE_COOKIE_JAR;
--- a/packages/insomnia-app/app/common/import.ts
+++ b/packages/insomnia-app/app/common/import.ts
@@ -39,7 +39,7 @@ interface ConvertResult {
 
 export interface ImportRawConfig {
   getWorkspaceId: ImportToWorkspacePrompt;
-  getSpaceId: () => string | null;
+  getSpaceId?: () => string | null;
   getWorkspaceScope?: SetWorkspaceScopePrompt;
   enableDiffBasedPatching?: boolean;
   enableDiffDeep?: boolean;
@@ -335,11 +335,13 @@ async function updateWorkspaceScope(
 
 async function createWorkspaceInSpace(
   resource: Workspace,
-  getSpaceId: () => string | null,
+  getSpaceId?: () => string | null,
 ) {
-  // Set the workspace parent if creating a new workspace during import
-  // @ts-expect-error workspace parent can be null or string
-  resource.parentId = getSpaceId();
+  if (getSpaceId) {
+    // Set the workspace parent if creating a new workspace during import
+    // @ts-expect-error workspace parent can be null or string
+    resource.parentId = getSpaceId();
+  }
 }
 
 export const isApiSpecImport = ({ id }: Pick<ConvertResultType, 'id'>) => (
--- a/packages/insomnia-app/app/common/import.ts
+++ b/packages/insomnia-app/app/common/import.ts
@@ -39,7 +39,7 @@ interface ConvertResult {
 
 export interface ImportRawConfig {
   getWorkspaceId: ImportToWorkspacePrompt;
-  getSpaceId?: () => string | null;
+  getSpaceId?: () => Promise<string | null>;
   getWorkspaceScope?: SetWorkspaceScopePrompt;
   enableDiffBasedPatching?: boolean;
   enableDiffDeep?: boolean;
@@ -259,7 +259,7 @@ export async function importRaw(
       // If workspace, check and set the scope and parentId while importing a new workspace
       if (isWorkspace(model)) {
         await updateWorkspaceScope(resource as Workspace, resultsType, getWorkspaceScope);
-        createWorkspaceInSpace(resource as Workspace, getSpaceId);
+        await createWorkspaceInSpace(resource as Workspace, getSpaceId);
       }
 
       newDoc = await db.docCreate(model.type, resource);
@@ -335,12 +335,12 @@ async function updateWorkspaceScope(
 
 async function createWorkspaceInSpace(
   resource: Workspace,
-  getSpaceId?: () => string | null,
+  getSpaceId?: () => Promise<string | null>,
 ) {
   if (getSpaceId) {
     // Set the workspace parent if creating a new workspace during import
     // @ts-expect-error workspace parent can be null or string
-    resource.parentId = getSpaceId();
+    resource.parentId = await getSpaceId();
   }
 }
 
--- a/packages/insomnia-app/app/common/import.ts
+++ b/packages/insomnia-app/app/common/import.ts
@@ -39,6 +39,7 @@ interface ConvertResult {
 
 export interface ImportRawConfig {
   getWorkspaceId: ImportToWorkspacePrompt;
+  getSpaceId: () => string | null;
   getWorkspaceScope?: SetWorkspaceScopePrompt;
   enableDiffBasedPatching?: boolean;
   enableDiffDeep?: boolean;
@@ -112,6 +113,7 @@ export async function importRaw(
   {
     getWorkspaceId,
     getWorkspaceScope,
+    getSpaceId,
     enableDiffBasedPatching,
     enableDiffDeep,
     bypassDiffProps,
@@ -246,15 +248,18 @@ export async function importRaw(
         updateDoc.url = resource.url;
       }
 
-      // If workspace, don't overwrite the existing scope
+      // If workspace preserve the scope and parentId of the existing workspace while importing
       if (isWorkspace(model)) {
         (updateDoc as Workspace).scope = (existingDoc as Workspace).scope;
+        (updateDoc as Workspace).parentId = (existingDoc as Workspace).parentId;
       }
 
       newDoc = await db.docUpdate(existingDoc, updateDoc);
     } else {
+      // If workspace, check and set the scope and parentId while importing a new workspace
       if (isWorkspace(model)) {
         await updateWorkspaceScope(resource as Workspace, resultsType, getWorkspaceScope);
+        createWorkspaceInSpace(resource as Workspace, getSpaceId);
       }
 
       newDoc = await db.docCreate(model.type, resource);
@@ -306,7 +311,7 @@ async function updateWorkspaceScope(
   resultType: ConvertResultType,
   getWorkspaceScope?: SetWorkspaceScopePrompt,
 ) {
-  // Set the workspace scope if creating a new workspace
+  // Set the workspace scope if creating a new workspace during import
   //  IF is creating a new workspace
   //  AND imported resource has no preset scope property OR scope is null
   //  AND we have a function to get scope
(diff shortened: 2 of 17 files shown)


develohpanda:
@wdawson @dimitropoulos I will link back to your comments on #3787 in this PR as I address them

develohpanda:
@wdawson in response to https://github.com/Kong/insomnia/pull/3787#pullrequestreview-701077880

> > The only exception to this "import into the active space" rule, is that if a workspace already exists and the import is simply updating the existing workspace, we retain the space the workspace exists in.
> 
> This is strange to me. As a user, I'm in the context of "another space" and the app just told me that it successfully imported some workspaces, but it's not showing up on my screen anywhere. Putting myself in a user's shoes, this flow doesn't seem very intuitive. Perhaps it would be better to warn them and say something like, "We found these collections/designs in other spaces (list them). Are you sure you want to import them into the FooBar space?" where FooBar is the current space. It seems to me that if a user wanted to update existing workspaces in a different space, they wouldn't have a problem switching to that space and working there.

There are a few technical limitations here; you can't import the same V4 export file twice and end up with two different workspaces, because the workspace in the export file usually has the _same id_. This means if you import the same file twice, it will just update/merge with the existing workspace. [demo](https://www.loom.com/share/a2fbc546c9fd4cd89c13cb4eb4eaa3df) Extend this slightly, let's say Jill exported a workspace which Jack imported. Then Jill made some updates to that workspace, exported, and sent to Jack to import. When Jack imports, it will update the same previously imported workspace (because the id will not have changed). I don't think we should change this import behavior (if we should, then I think that's a separate discussion).

With that behavior in mind, the things we can do are:
1. activate the space in which the workspace currently exists and go to that dashboard after import
1. inform users that the workspace already existed and in space XYZ and data will be imported into there
1. something else that doesn't change their data, but navigates or informs

With both of these cases, what happens if an export file contains multiple existing workspaces that exist in different places? I think we can only make a best attempt here 🤔

develohpanda:
> > Import using insomnia protocol
> 
> I think we're going to need to ask which space you want to import this into if and only if you have multiple spaces. My reasoning is that if you have multiple spaces and you're not working in the app, just browsing the web, you stumble across something that lets you import insomnia collections. You click that button but it imports something into your team's production Space that you were working in yesterday instead of your personal debugging/testing space. Not good.

I agree with this, great point. Have updated this PR.

More demos for the Insomnia protocol import handler:
- Only base space so don't prompt [demo](https://www.loom.com/share/0d040e9ea4d54dbb91cf29088f89cd44)
- Workspace already exists, so don't prompt [demo](https://www.loom.com/share/ae9d9ac9e821404a9a44358dc9ec2740)
- Prompt for space if multiple exist [demo](https://www.loom.com/share/b0bc335f354b452584852491900cff7b)

In addition, I also enabled a space prompt for the drag-and-drop flow, because you could be in any context but want to import into a different space. @wdawson let me know if I should undo this change, it's a single flag.

develohpanda:
[`b50e636` (#3795)](https://github.com/Kong/insomnia/pull/3795/commits/b50e636e022de9a667ed2175394d22bb6c280366) is an initial pass, I intent to tidy this up 😄

wdawson:
> @wdawson in response to [#3787 (review)](https://github.com/Kong/insomnia/pull/3787#pullrequestreview-701077880)
> 
> > > The only exception to this "import into the active space" rule, is that if a workspace already exists and the import is simply updating the existing workspace, we retain the space the workspace exists in.
> > 
> > ...
> 
> There are a few technical limitations here; you can't import the same V4 export file twice and end up with two different workspaces, because the workspace in the export file usually has the _same id_. This means if you import the same file twice, it will just update/merge with the existing workspace. [demo](https://www.loom.com/share/a2fbc546c9fd4cd89c13cb4eb4eaa3df) Extend this slightly, let's say Jill exported a workspace which Jack imported. Then Jill made some updates to that workspace, exported, and sent to Jack to import. When Jack imports, it will update the same previously imported workspace (because the id will not have changed). I don't think we should change this import behavior (if we should, then I think that's a separate discussion).
> 
> With that behavior in mind, the things we can do are:
> 
> 1. activate the space in which the workspace currently exists and go to that dashboard after import
> 2. inform users that the workspace already existed and in space XYZ and data will be imported into there
> 3. something else that doesn't change their data, but navigates or informs
> 
> With both of these cases, what happens if an export file contains multiple existing workspaces that exist in different places? I think we can only make a best attempt here 🤔

Gotcha. Good context on the immutable IDs. I agree that makes sense to keep if that's the way to do a janky sync. Though, we may want to revisit that down the road when we make the sync experience better. For now, let's keep it.

As for what to do, I think there are two cases (@erickrico let us know if you disagree):

- If you're in a **different space**, my gut says to do `#2` since switching context without warning can be jarring. I expect that would be a prompt with something like, "The workspace you're trying to import already exists in another space (`[name of space]`). How do you want to proceed?" Ideally there would be 3 options in this prompt:
-- "Update my existing workspace": Choosing this will update the existing thing and we should switch to that space and go straight to the updated workspace. Reasoning is that they probably want to work with that thing now.
-- "Duplicate into this space": Choosing this would let them create a new copy of this workspace in this collection with a new ID. Should be the same as the new duplicate flow I think. This option is fine as a stretch I think.
-- "Cancel": Stops the import and keeps the user on their current screen.
- If you're in the **same space**, I think we can switch directly to the workspace you imported without needing a prompt. Reasoning is because we want to get you working with that thing right now.

Side note, will importing clobber any outstanding changes? Or try to do conflict resolution? Might be good to put that in the message and/or docs too.

wdawson:
> > > Import using insomnia protocol
> > 
> > ...
> 
> I agree with this, great point. Have updated this PR.
> 
> More demos for the Insomnia protocol import handler:
> 
> * Only base space so don't prompt [demo](https://www.loom.com/share/0d040e9ea4d54dbb91cf29088f89cd44)
> * Workspace already exists, so don't prompt [demo](https://www.loom.com/share/ae9d9ac9e821404a9a44358dc9ec2740)
> * Prompt for space if multiple exist [demo](https://www.loom.com/share/b0bc335f354b452584852491900cff7b)
> 
> In addition, I also enabled a space prompt for the drag-and-drop flow, because you could be in any context but want to import into a different space. @wdawson let me know if I should undo this change, it's a single flag.

Sweet! Thanks for the quick work. That triple prompt flow is tough. I think it's fine for now but definitely an area to revisit with UX at some point (see last demo @erickrico ). Does it also work if the app is closed? Like will the protocol open the app and still prompt for which space to import to? What is the default space in that case? 

I think the drag-and-drop flow doesn't need to prompt though. My reasoning there is because it feels more likely that you'd drag and drop onto the correct context. It's like dragging files into google drive; it doesn't ask you which folder you want to drop them into.

develohpanda:
> Sweet! Thanks for the quick work. That triple prompt flow is tough. I think it's fine for now but definitely an area to revisit with UX at some point (see last demo @erickrico ).

Yeah I don't like the multiple prompts and alerts at all either. It'll need some discussion, though. On the one hand, we could provide a list of options for import upfront, but those prompts are only used if they're needed in the import flows, not always, so it doesn't quite make sense to ask all the questions to the user upfront 🤔 eg. we currently pass around callbacks to the import logic, saying "if you need xyz, use this to fetch it", and some of those callbacks lead to a prompt. This isn't a particularly great solution but we haven't had a chance to look at it with fresh eyes after the initial approach went in this direction.

https://github.com/Kong/insomnia/blob/b50e636e022de9a667ed2175394d22bb6c280366/packages/insomnia-app/app/ui/redux/modules/import.ts#L54-L59

> Does it also work if the app is closed? Like will the protocol open the app and still prompt for which space to import to? What is the default space in that case?

Yep, it works the same was if the app is closed. It will open the app and go through the same flow.

> I think the drag-and-drop flow doesn't need to prompt though. My reasoning there is because it feels more likely that you'd drag and drop onto the correct context. It's like dragging files into google drive; it doesn't ask you which folder you want to drop them into.

Sounds reasonable. It relates to the discussion recently about making it clear where you are in the app (show the space and workspace everywhere). I'll undo this!

develohpanda:
> * If you're in a different space, my gut says to do `#2` since switching context without warning can be jarring. I expect that would be a prompt with something like, "The workspace you're trying to import already exists in another space ([name of space]). How do you want to proceed?" Ideally there would be 3 options in this prompt:
-- "Update my existing workspace": Choosing this will update the existing thing and we should switch to that space and go straight to the updated workspace. Reasoning is that they probably want to work with that thing now.
-- "Duplicate into this space": Choosing this would let them create a new copy of this workspace in this collection with a new ID. Should be the same as the new duplicate flow I think. This option is fine as a stretch I think.
-- "Cancel": Stops the import and keeps the user on their current screen.
> * If you're in the same space, I think we can switch directly to the workspace you imported without needing a prompt. Reasoning is because we want to get you working with that thing right now.

Both of these options assume there is only _one_ workspace in the file being imported, which is not the case when importing an insomnia document. It _is_ the case when importing an OpenAPI spec or some other import format, but is not always the case, so we likely need another option? 🤔

I also missed this fact when describing my options, but it stands that there could be multiple. We kind of need an "import result report" 🤣

dimitropoulos (approved):
It'd be nice to have some of those `onCancel`s in now, but if you can't get to it in this PR please make a ticket so we can follow up before release.

develohpanda:
> It'd be nice to have some of those `onCancel`s in now

Handled in [`867e705` (#3795)](https://github.com/Kong/insomnia/pull/3795/commits/867e705694e0293d1f11088975d3debf2273dd28). See below, there is no close button, and pressing the escape key is also ignored. 

image

@dimitropoulos

PR #3798, opened 2021-07-08
Upgrade Electron to 11.2.3.

johnwchadwick:
Upgrades Electron from 9.3.5 to 11.2.3.

There are still likely plenty of issues to be resolved, this PR so far just gets the app building and running.

We will need to look into context isolation; it is planned to be required by (I believe) Electron 14.

Reverted back from Electron 12 due to breakages that only occur on Windows; these appear to be related bugs: electron/electron#19554 / electron/electron#22119. It seems to be resolved by disabling Windows Defender Real-time Protection, but `allowRendererProcessReuse = false` doesn't seem to help with Node 14 anymore. I think we are going to need to do some architectural refactors to make it to Electron 12, and thankfully getting to Electron 11 should make said refactors easier. (We can't go to >=11.3 because `node-libcurl` doesn't have builds for it.)


packages/insomnia-app/package.json
```diff
@@ -220,7 +220,7 @@
     "concurrently": "^3.5.0",
     "cross-env": "^2.0.0",
     "css-loader": "^2.1.1",
-    "electron": "9.3.5",
+    "electron": "12.0.14",
     "electron-builder": "^22.5.1",
```
dimitropoulos: while we're here, it might be valuable to upgrade `electron-builder` to `^22.11.7` https://github.com/electron-userland/electron-builder/releases

packages/insomnia-app/package.json
```diff
@@ -220,7 +220,7 @@
...
     "electron-builder-squirrel-windows": "^22.5.1",
```
dimitropoulos: this is in lockstep with electron-builder above

packages/insomnia-app/package.json
```diff
@@ -220,7 +220,7 @@
...
     "electron-devtools-installer": "^3.2.0",
```
dimitropoulos, line 226: github won't let me select the line below, but the line below is for `electron-notarize`.  If it isn't a problem, it'd be good to get this on the next release (which, was only breaking because it dropped node 10 support: https://github.com/electron/electron-notarize/releases)

packages/insomnia-app/package.json
```diff
@@ -220,7 +220,7 @@
     "concurrently": "^3.5.0",
     "cross-env": "^2.0.0",
     "css-loader": "^2.1.1",
```
dimitropoulos, line 222: (not related to this line)
do:
- `electron-context-menu`
- `electron-log`
- `electron-updated`
- `webpack-target-electron-renderer`

also need a bump?

.nvmrc
```diff
@@ -1 +1 @@
-12.18.3
+14.16.0
```
dimitropoulos: did you, perchance, run an `npm run bootstrap` after this?  when I tried to upgrade node a week or two ago, it updated all the lockfiles (`"lockfileVersion": 2`)

dimitropoulos: looks like this happened for npm v7 which shipped with node 15 so maybe actually it's fine for now https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion if we're on node 14.  I tried to update to 16.4.0 I think.  Still, good to run an `npm run bootstrap` just to be sure.

dimitropoulos: any reason not to go with 14.16.1? https://nodejs.org/ja/blog/release/v14.16.1/ there are three CVEs in there (didn't look at them though).

dimitropoulos: actually on second though, I didn't realize the latest 14 release is 14.17.3.  Any reason not to be on that?

dimitropoulos (commented):
Screenshot_20210708_150017
just as a data point: works well in linux so far

.nvmrc
johnwchadwick: The reasoning is visible in your screenshot: Electron comes bundled with Node. I synced the Node version to 14.16.0 because it is the version of Node that comes with Electron 12.0.14. I don't think this is strictly necessary; I'm pretty sure as long as NAPI doesn't bump you can use pretty much any Node version. However, there are real actual API changes between Node.js minor version numbers; for example, from 14.11 to 14.12, `fs.writeFileSync` was modified so that its `data` parameter can accept objects which have an explicit `toString()` method. Such a difference would be detrimental to tests that run in Node, which would not match the behavior of code running in Electron.

The table mapping Electron releases and Node versions is enumerated over here. https://www.npmjs.com/package/electron-releases#releases

dimitropoulos: oh wow, nice.  I was looking and looking for exactly this chart but couldn't find it.

johnwchadwick: I did.

packages/insomnia-app/package.json
johnwchadwick: Bumped all of them, except `electron-context-menu`. Funny enough, bumping that one crashes the app. It would certainly be worth investigating, but probably not right now.

packages/insomnia-app/package.json
johnwchadwick: Done.

packages/insomnia-app/package.json
johnwchadwick: Done.

packages/insomnia-app/package.json
johnwchadwick: Done.

packages/insomnia-app/package.json
```diff
@@ -236,7 +236,7 @@
     "react-hot-loader": "^4.13.0",
     "redux-mock-store": "^1.5.4",
     "source-map-loader": "^1.0.0",
-    "spectron": "^11.1.0",
+    "spectron": "^13.0.0",
```
develohpanda: Interesting, this must be residue from a past time. Spectron isn't needed in this package

develohpanda (approved):
Looks to be working! I tested both REST and gRPC, keyboard shortcuts, insomnia sync and git sync

DMarby (approved):
This is awesome @johnwchadwick!
Do we need to change our use of `BrowserWindow.destroy()` since that seems to have been removed in Electron 10.x?

https://github.com/Kong/insomnia/blob/227caae91979e8ff17546c3ddceae4fba48bb3c2/packages/insomnia-app/app/main/window-utils.ts#L514

https://www.electronjs.org/docs/breaking-changes#planned-breaking-api-changes-110

johnwchadwick:
@DMarby 
> This is awesome @johnwchadwick!
> Do we need to change our use of `BrowserWindow.destroy()` since that seems to have been removed in Electron 10.x?
> 
> https://github.com/Kong/insomnia/blob/227caae91979e8ff17546c3ddceae4fba48bb3c2/packages/insomnia-app/app/main/window-utils.ts#L514
> 
> https://www.electronjs.org/docs/breaking-changes#planned-breaking-api-changes-110

I tried to check all of the individual breaking changes but I appreciate getting a second eye on it.

As far as I can tell, the answer is no. `BrowserWindow.destroy` is not the same as `BrowserView.destroy`: https://www.electronjs.org/docs/api/browser-window#windestroy

DMarby:
> As far as I can tell, the answer is no. `BrowserWindow.destroy` is not the same as `BrowserView.destroy`: https://www.electronjs.org/docs/api/browser-window#windestroy

🤦 That makes sense! I somehow read BrowserView as BrowserWindow

packages/insomnia-app/package.json
johnwchadwick: Finally fixed this; took some extra time because I wanted to make sure I tested bootstrapping from scratch without it just to be sure no funny business was going on.

dimitropoulos (approved):
image

I saw this when I was running it locally (not the alpha) yesterday, but I ran it earlier today and I didn't see it again.  Not sure.  All the same, couldn't find anything else.  looks good!

PR #3803, opened 2021-07-11
Load local GraphQL schema

### LINKED ISSUE #2743: Add support for importing GraphQL schemas from local files.
Complementary to #2629 ...

Sometimes server deployments disable GraphQL introspection. It would be nice if in those situations Insomnia could load the schema from a local SDL file.

lodi-g:
I often work with GraphQL, and sometimes the introspection requests are blocked. In these case, the developers give me a GraphQL schema so I can still have the documentation.
However, Insomnia does not support importing local schema files, only fetching from internet. Therefore I must use Postman for this case, but I'd prefer to use Insomnia.

Closes #2743


packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -321,6 +323,56 @@ class GraphQLEditor extends PureComponent<Props, State> {
...
+      schemaIsFetching: false,
+    };
+
+    const options: OpenDialogOptions = {
+      title: 'Import GraphQL introspection schema',
+      buttonLabel: 'Import',
+      properties: ['openFile'],
+      filters: [
+        {
+          name: 'GraphQL file',
```
develohpanda: I suggest using the following helper if it's possible - we recently faced an issue with the `OpenDialogOptions.filters[*].name` field on Linux, and it was fixed generically here, so reusing this helper will avoid the issue showing up again.

 https://github.com/Kong/insomnia/blob/9f3705a8a8d784c621b29a7ef635c561e9d15990/packages/insomnia-app/app/common/select-file-or-folder.ts#L14-L14

See more context here: https://github.com/Kong/insomnia/pull/3348

develohpanda (commented):
Looking good! I haven't tested with a local schema file yet, but code and UX looks good - submitting this review with one small piece of feedback.

image

Error message looks good too

image

I also pushed a commit to fix up how state is being used to avoid the `as any` 😄

develohpanda (approved):
Tested with https://github.com/apollographql/apollo-tooling/blob/master/__fixtures__/starwars/schema.json and looks to be working well!

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
develohpanda: I addressed this in [`7304fb8` (#3803)](https://github.com/Kong/insomnia/pull/3803/commits/7304fb884bc433c7d6bd4f414df8422b3a6b6271); it wouldn't be possible to use `selectFileOrFolder` in the current state (because some of the options are not exposed as params)

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -321,6 +323,54 @@ class GraphQLEditor extends PureComponent<Props, State> {
...
+    } catch (err) {
+      console.log('[graphql] ERROR: Failed to fetch schema', err);
+      newState.schemaFetchError = {
+        message: `Failed to fetch schema: ${err.message}`,
+        response: null,
+      };
+    }
+
+    if (this._isMounted) {
+      this.setState(existingState => ({ ...existingState, ...newState }));
```
dimitropoulos, line 370: is there a reason to do this instead of `this.setState(newState)`?

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -374,6 +424,12 @@ class GraphQLEditor extends PureComponent<Props, State> {
     );
   }
 
+  _handleSetLocalSchema() {
+    this.setState({ hideSchemaFetchErrors: false }, async () => {
+      await this._loadAndSetLocalSchema();
+    });
```
dimitropoulos, line 430: is this the same as `this.setState({ handleSchemaFetchErrors: false }, this._loadAndSetLocalSchema)`?

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -321,6 +323,54 @@ class GraphQLEditor extends PureComponent<Props, State> {
     }
   }
 
+  async _loadAndSetLocalSchema() {
+    const newState: Partial<State> = {
```
dimitropoulos, line 327: personally, I find it much easier to follow what's going on when we don't create intermediate state like this.  We can avoid it by just setting the values that we need to _when_ we need to:

```ts
  async _loadAndSetLocalSchema() {
    const options: OpenDialogOptions = {
      title: 'Import GraphQL introspection schema',
      buttonLabel: 'Import',
      properties: ['openFile'],
      filters: [
        // @ts-expect-error https://github.com/electron/electron/pull/29322
        {
          extensions: ['', 'json'],
        },
      ],
    };

    const { canceled, filePaths: paths } = await electron.remote.dialog.showOpenDialog(options);

    if (canceled) {
      return;
    }

    try {
      const path = paths[0]; // showOpenDialog is single select
      const file = readFileSync(path);

      const content = JSON.parse(file.toString());
      if (!content.data) {
        throw new Error('JSON file should have a data field with the introspection results');
      }

      if (this._isMounted) {
        return;
      }
      this.setState({
        schema: buildClientSchema(content.data),
        schemaLastFetchTime: Date.now(),
        schemaFetchError: null,
        schemaIsFetching: false,
      });
    } catch (err) {
      console.log('[graphql] ERROR: Failed to fetch schema', err);
      if (!this._isMounted) {
        return;
      }
      this.setState({
        schemaFetchError: {
          message: `Failed to fetch schema: ${err.message}`,
          response: null,
        },
        schemaIsFetching: false,
      });
    }
  }
```

If anyone's amenable to it I think it would improve the maintainability of this code to do the above (or something like it).

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -321,6 +323,54 @@ class GraphQLEditor extends PureComponent<Props, State> {
...
+      properties: ['openFile'],
+      filters: [
+        // @ts-expect-error https://github.com/electron/electron/pull/29322
+        {
+          extensions: ['', 'json'],
+        },
+      ],
+    };
+
+    const { canceled, filePaths: paths } = await electron.remote.dialog.showOpenDialog(options);
```
dimitropoulos, line 344:
```suggestion
    const { canceled, filePaths } = await electron.remote.dialog.showOpenDialog(options);
```

I think we should not alias this.  It isn't being done to avoid a shadowing of `paths` somewhere else (so far as I can tell), and just adds cognitive overhead (to me).  Maybe it'd be good, in kind, to also do:

```ts
      const filePath = filePaths[0]; // showOpenDialog is single select
      const file = readFileSync(filePath);
 ```
 
 below as well

File: packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx [hunk 6]
```diff
@@ -374,6 +424,12 @@ class GraphQLEditor extends PureComponent<Props, State> {
     );
```
dimitropoulos, line 424: (github won't let me _quite_ select the line I want, but I'm talking here about `_handleRefreshSchema`).

As a user of this, the very first thing I did was change the schema on disk and hit "Refresh Schema":
image


And it just returns the error at the bottom "schema not yet fetched".  I think if we're introducing a local schema option, we really need to handle this case.

dimitropoulos (changes requested):
image

Works as advertised with https://raw.githubusercontent.com/apollographql/apollo-tooling/master/__fixtures__/misc/schema.json but I have a few questions.

In general, I think this is a bigger feature than we're making it out to be in terms of UX consideration (and this particular item is a good example of that).  https://github.com/Kong/insomnia/pull/3803#discussion_r669734541 is a good example of that, but another one is: should there be any indication anywhere to a user that they're using a remote vs local schema?  Right now there isn't one, right?  I think their should be.

Don't get me wrong: I love this PR and I really like the use-case it solves, but I also want to be considerate to a good user experience when users are uploading local schemas.

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
develohpanda: I did this; if we don't do it this way, newState cannot be a partial. Try it out locally (thought I think this comment https://github.com/Kong/insomnia/pull/3803#discussion_r669722438  removes the need for intermediate state)

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
develohpanda: IMO this works the same way as proto files, as a one time import. There's no persistent link to a file on the file system (and I think this is good) such that refresh schema should work for local files.

dimitropoulos (approved):

develohpanda:
Thanks for the PR and contribution @lodi-g 😄 We're merging this for functionality, and we'll put in a follow-up PR to address some of the feedback above, as well as a slight improvement to the UX to discern between working with remote schemas (refresh and auto-fetch) and local schemas. 🙌🏽

wdawson (approved):

lodi-g:
Thanks for accepting it @develohpanda. I unfortunately did not have the time to address your feedback but I wholeheartedly agree that the UX of my PR can be improved. Cheers! 🍻

PR #3809, opened 2021-07-13
Add option to change test timeouts (in the GUI and via inso)

develohpanda:
This PR adds the ability to override the timeouts per unit test, and allows inso to override the default test timeout (for all those tests that do not have a timeout override). By default, all timeouts are 5s.

image

When run via the CLI, we see the same results:
image

But via the CI, we can override the default timeout. Increasing to 10s, everything passes:
image

Disabling the timeout, everything also passes:
image

Changing the timeout to 2s via the CI fails every test that doesn't manually override the timeout:
image

Closes https://github.com/Kong/insomnia/discussions/3648, Closes INS-806


.github/workflows/release-npm.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          node-version: ${{ steps.nvm.outputs.NVMRC }}
+      - name: Bootstrap packages
+        run: npm run bootstrap
+      - name: Release NPM packages
+        run: |
+          npm config set //registry.npmjs.org/:_authToken ${NPM_TOKEN}
+          npm run release
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN}}
```
johnwchadwick: Missing newline at EOF.

packages/insomnia-testing/src/run/run.ts
```diff
@@ -23,15 +23,15 @@ const runInternal = async <T>(
...
+  const { bail, keepFile, testFilter, timeout } = options;
 
   // Add global `insomnia` helper.
   // This is the only way to add new globals to the Mocha environment as far as I can tell
   global.insomnia = new Insomnia(options);
   global.chai = chai;
 
   const mocha: Mocha = new Mocha({
-    timeout: 5000,
+    timeout: timeout || 5000,
```
johnwchadwick: nit: Consider using `??`, otherwise a `timeout` value of 0 would be coerced to 5000. I'm not sure it's ever particularly useful, but a value of 0 disables timeouts.

develohpanda: Oh this is a great catch, TY. Not tested yet so still draft 😆 

```suggestion
    timeout: timeout ?? 5000,
```

.github/workflows/release-npm.yml
develohpanda:
```suggestion
          NPM_TOKEN: ${{ secrets.NPM_TOKEN}}

```

johnwchadwick:
Closing this since it is no longer being worked on. We can always rebase this later if it turns out to be something we want to do.

PR #3819, opened 2021-07-16
Ensure a base space always exists (+ fix gRPC sub-environments)

### LINKED ISSUE #3221: Sub-environment variables don't work with gRPC
**Describe the bug**
I recently discovered that the latest version of Insomnia supports environment variables; however I am unable to get this to work. When I put in an environment variable of `127.0.0.1:6565` and attempt to use this, nothing happens in the UI (there are errors in the dev console). If I directly paste `127.0.0.1:6565` in, it works as expected (old behavior). I have tried downgrading to 2020.1.0 but that doesn't fix the issue either. 

**To Reproduce**
In the environment tab, add a new variable. Fill that variable in and make a gRPC request.

**Expected behavior**
The request should be successful, but it is not

**Screenshots**
(These are intentionally cropped)
image
image
image

image

Version: Insomnia 2021.1.1
Release date: 3/9/2021
OS: Darwin x64 19.6.0
Electron: 9.1.1
Node: 12.14.1
V8: 8.3.110.13-electron.0
Architecture: x64

**Additional context**
Error message:
```
18:24:20.309 › Unhandled Promise PromiseRejectionEvent {isTrusted: true, promise: Promise, reason: Error: attempted to output null or undefined value
    at file:///Applications/Insomnia.app/Content…, type: "unhandledrejection", target: global, …}bubbles: falsecancelBubble: falsecancelable: truecomposed: falsecurrentTarget: global {parent: global, opener: null, top: global, length: 0, frames: global, …}defaultPrevented: falseeventPhase: 0isTrusted: truepath: [global]promise: Promise__proto__: Promise[[PromiseStatus]]: "rejected"[[PromiseValue]]: Error: attempted to output null or undefined value
    at file:///Applications/Insomnia.app/Contents/Resources/app.asar/bundle.js:37548:26
    at /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/nunjucks/src/environment.js:41:5
    at RawTask.call (/Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/asap/asap.js:40:19)
    at flush (/Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/asap/raw.js:50:29)
    at processTicksAndRejections (internal/process/task_queues.js:76:11)location: {line: 1, column: 1}path: "url"reason: "undefined"type: "render"message: "attempted to output null or undefined value"stack: "Error: attempted to output null or undefined value↵    at file:///Applications/Insomnia.app/Contents/Resources/app.asar/bundle.js:37548:26↵    at /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/nunjucks/src/environment.js:41:5↵    at RawTask.call (/Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/asap/asap.js:40:19)↵    at flush (/Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/asap/raw.js:50:29)↵    at processTicksAndRejections (internal/process/task_queues.js:76:11)"__proto__: Errorreason: Error: attempted to output null or undefined value
    at file:///Applications/Insomnia.app/Contents/Resources/app.asar/bundle.js:37548:26
    at /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/nunjucks/src/environment.js:41:5
    at RawTask.call (/Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/asap/asap.js:40:19)
    at flush (/Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/asap/raw.js:50:29)
    at processTicksAndRejections (internal/process/task_queues.js:76:11)returnValue: truesrcElement: global {parent: global, opener: null, top: global, length: 0, frames: global, …}target: global {parent: global, opener: null, top: global, length: 0, frames: global, …}timeStamp: 595062.414999993type: "unhandledrejection"__proto__: PromiseRejectionEvent
setTimeout (async)
consoleLog @ /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/electron-log/src/transports/console.js:93
transport @ /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/electron-log/src/transports/console.js:46
runTransport @ /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/electron-log/src/log.js:44
runTransports @ /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/electron-log/src/log.js:27
log @ /Applications/Insomnia.app/Contents/Resources/app.asar/node_modules/electron-log/src/log.js:21
```

develohpanda:
### Changes
* The export functions now always require a list of workspaces (the magic where we default an array of `[null]` has been removed
	* This directly impacts the plugin API but a fallback has been put into place, which we might need to add metrics for (see self-review)
* Converted some of the render function arguments into an object to better cater for optionals
	* As a positive side-effect, fixing these types identified an error in the types [ref](https://github.com/Kong/insomnia/pull/3819#discussion_r674502952). Fixing the types fixes #3221
* If a workspace did not have a `parentId` it is set to the `BASE_SPACE_ID`
* Some existing tests need to call `models.space.all()` to make sure the parent space exists; this is done in `beforeEach` blocks

Please see self-review comments for context in different areas

### Demos
* [Don't export if there are no collections in the selected space ](https://www.loom.com/share/0f76f8b7ee1c4964880b8afbd3e075f8)

Closes INS-812


packages/insomnia-app/app/network/grpc/prepare.ts
```diff
@@ -23,15 +23,16 @@ export const prepareGrpcRequest = async (
...
-    // @ts-expect-error -- TSCONVERSION req can be null but should not try to render if it is null
-    req,
-    environment,
-    RENDER_PURPOSE_SEND,
-    {},
-    canClientStream(methodType),
+    {
+      // @ts-expect-error -- TSCONVERSION req can be null but should not try to render if it is null
+      request: req,
+      environmentId,
```
develohpanda, line 30: Previously we send the entire environment, but it only expects an environmentId, which fixes https://github.com/Kong/insomnia/issues/3221. Yay for improved types!

packages/insomnia-app/app/plugins/context/data.ts
```diff
@@ -23,7 +25,10 @@ const buildImportRawConfig = (options: PluginImportOptions): ImportRawConfig =>
   ),
 });
 
-export const init = () => ({
+// TODO: add metrics here to track how frequently this fallback is being used
```
develohpanda, line 29: How can we add a metric here to check this - through segment?

packages/insomnia-app/app/plugins/context/data.ts
```diff
@@ -39,7 +43,7 @@ export const init = () => ({
         includePrivate,
         format,
       }: InsomniaExport = {}) => exportWorkspacesData(
-        workspace ? [workspace] : [],
+        workspace ? [workspace] : await getWorkspacesInActiveSpace(activeSpaceId),
```
develohpanda, line 46: Initially, sending no workspace would cause all workspaces to export

After introducing spaces, sending no workspace would cause all workspaces _in the base space only, always_ to export (this is because the export logic would fetch all of the workspaces with a `null` parentId - thus those in the base space).

After spaces, the behavior wasn't really ideal - what should happen is if the plugin does not specify which workspace to export, we should by default get all of the workspaces in the active space. This PR updates that behavior by sending the `activeSpaceId` whenever this plugin context is initialized, and a large majority of the changes in this PR are related to getting the space to this `init` function

packages/insomnia-app/app/network/network.ts
```diff
@@ -1003,7 +1005,7 @@ async function _applyRequestPluginHooks(
   for (const { plugin, hook } of await plugins.getRequestHooks()) {
     const context = {
       ...(pluginContexts.app.init(RENDER_PURPOSE_NO_RENDER) as Record<string, any>),
-      ...pluginContexts.data.init(),
+      ...pluginContexts.data.init(renderedContext.getSpaceId()),
```
develohpanda, line 1009: This follows the same pattern as fetching the environment Id four lines down when initializing the network context

packages/insomnia-app/app/plugins/context/data.ts
```diff
@@ -39,7 +47,7 @@ export const init = () => ({
         includePrivate,
         format,
       }: InsomniaExport = {}) => exportWorkspacesData(
-        workspace ? [workspace] : [],
+        workspace ? [workspace] : await getWorkspaces(activeSpaceId),
```
develohpanda, line 52: Initially, sending no workspace would cause all workspaces to export

After introducing spaces, sending no workspace would cause all workspaces _in the base space only, always_ to export (this is because the export logic would fetch all of the workspaces with a `null` parentId - thus those in the base space).

The behavior isn't great - what should happen is if the plugin does not specify which workspace to export, we should by default get all of the workspaces in the _active_ space, which matches more closely to existing behavior.

This PR updates that behavior by sending the `activeSpaceId` whenever this plugin context is initialized, and a large majority of the changes in this PR are related to getting the space to this `init` function.

packages/insomnia-app/app/plugins/context/data.ts
```diff
@@ -23,7 +24,14 @@ const buildImportRawConfig = (options: PluginImportOptions): ImportRawConfig =>
...
 
-export const init = () => ({
+// TODO: add metrics here to track how frequently this fallback is being used
+const getWorkspaces = (activeSpaceId?: string) => 
+  activeSpaceId 
+    ? models.workspace.findByParentId(activeSpaceId) 
+    : models.workspace.all();
+
+// Only in the case of running unit tests from Inso via send-request can activeSpaceId be undefined. This is because the concept of a space doesn't exist in git/insomnia sync or an export file
+export const init = (activeSpaceId?: string) => ({
```
develohpanda, line 36: The only time an `activeSpaceId` is not available is when we are coming from inso (which has no context of a space) through `insomnia-send-request`, and in that case, we return all of the workspaces present. This is only for completeness; it is unlikely there is a workflow that routes through here via inso but in case there is we have an implementation.

packages/insomnia-app/app/common/__tests__/har.test.ts
```diff
@@ -6,543 +6,544 @@ import { AUTH_BASIC } from '../constants';
 import * as harUtils from '../har';
 import * as render from '../render';
 
-describe('exportHar()', () => {
-  beforeEach(globalBeforeEach);
+describe('export', () => {
```
develohpanda, line 9: Some of the unit testing diffs are not lining up well in GitHub - if that is the case, it is most likely because of a top-level describe block that has been introduced (with a common `beforeEach`) causing an indent change on each line. Be sure to hide whitespace changes when reviewing: 
image

packages/insomnia-app/app/models/space.ts
```diff
@@ -64,6 +63,13 @@ export function update(space: Space, patch: Partial<Space>) {
...
+export async function all() {
+  const spaces = await db.all<Space>(type);
+
+  if (!spaces.find(c => c._id === BASE_SPACE_ID)) {
+    await create({ _id: BASE_SPACE_ID, name: getAppName(), remoteId: null });
+    return db.all<Space>(type);
+  }
+
+  return spaces;
 }
```
develohpanda, line 75: The same pattern that existed previously for ensuring that a workspace always existed

packages/insomnia-app/app/common/__tests__/render.test.ts
```diff
@@ -1,846 +1,745 @@
...
+import { Environment } from '../../models/environment';
+import { Workspace } from '../../models/workspace';
 import * as renderUtils from '../render';
 
 jest.mock('electron');
 
-describe('render()', () => {
-  beforeEach(globalBeforeEach);
+const envBuilder = createBuilder(environmentModelSchema);
+const reqGroupBuilder = createBuilder(requestGroupModelSchema);
```
develohpanda, line 13: Using these builders to fix many of the type errors in this file. This file hasn't been updated since flow and my current work was in this area.

packages/insomnia-app/app/common/database.ts
```diff
@@ -549,7 +549,7 @@ export const database = {
 
     let docsToReturn: T[] = doc ? [doc] : [];
 
-    async function next(docs: T[]) {
+    async function next(docs: T[]): Promise<T[]> {
```
develohpanda, line 552: TS didn't auto-detect this return type (and thus didn't type the parent type properly)

packages/insomnia-app/app/common/render.ts
```diff
@@ -54,10 +54,17 @@ export type HandleGetRenderContext = () => Promise<RenderContextAndKeys>;
 export type HandleRender = <T>(object: T, contextCacheKey?: string | null) => Promise<T>;
 
 export async function buildRenderContext(
```
develohpanda, line 56: In a future PR, we should really type the return type of this correctly; it will save us many headaches in the future

packages/insomnia-app/app/common/render.ts
```diff
@@ -280,17 +287,31 @@ export async function render<T>(
...
-  purpose: RenderPurpose | null = null,
-  extraInfo: ExtraRenderInfo | null = null,
+  {
+    request,
+    environmentId,
+    ancestors: _ancestors,
+    purpose,
+    extraInfo,
+  }: RenderContextOptions,
 ): Promise<Record<string, any>> {
```
develohpanda, line 311: In a future PR, we should really type the return type of this correctly; it will save us many headaches

packages/insomnia-app/app/plugins/context/__tests__/data.test.ts
```diff
@@ -47,7 +52,7 @@ describe('app.import.*', () => {
         modified: 999,
         description: '',
         name: 'New',
-        parentId: null,
+        parentId: BASE_SPACE_ID,
```
develohpanda: Should import into the current space

packages/insomnia-app/app/plugins/context/__tests__/data.test.ts
```diff
@@ -95,7 +100,7 @@ describe('app.import.*', () => {
...
```
develohpanda: Should import into the current space

packages/insomnia-app/app/sync/git/__tests__/ne-db-client.test.ts
```diff
@@ -66,11 +68,11 @@ describe('NeDBClient', () => {
       const wrk1Yml = path.join(GIT_INSOMNIA_DIR, models.workspace.type, 'wrk_1.yml');
       const req1Yml = path.join(GIT_INSOMNIA_DIR, models.request.type, 'req_1.yml');
       const reqXYml = path.join(GIT_INSOMNIA_DIR, models.request.type, 'req_x.yml');
-      const pNeDB = new NeDBClient('wrk_1');
+      const pNeDB = new NeDBClient('wrk_1', BASE_SPACE_ID);
       expect(YAML.parse((await pNeDB.readFile(wrk1Yml, 'utf8')).toString())).toEqual(
         expect.objectContaining({
           _id: 'wrk_1',
-          parentId: null,
+          parentId: null, // should be reset to default value instead of the active space id
```
develohpanda, line 75: This is because we set/clear the `parentId` when writing/reading through the git integration

packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
```diff
@@ -15,15 +16,9 @@ import ModalBody from '../base/modal-body';
...
-  space?: Space;
-}
-
-interface ReduxDispatchProps {
-  handleRemoveSpace: (space: Space) => void;
-}
+export type ReduxProps = ReturnType<typeof mapStateToProps> & ReturnType<typeof mapDispatchToProps>;
 
-interface Props extends ReduxStateProps, ReduxDispatchProps { }
+type Props = ReduxProps
```
develohpanda, line 21: Seems trivial, but just following our own set convention :)

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
```diff
@@ -28,19 +29,19 @@ describe('selectors', () => {
...
       await models.space.create();
 
-      // set first as selected
-      const state = await reduxStateForTest({ activeSpaceId: null });
+      // set base as selected
+      const state = await reduxStateForTest({ activeSpaceId: undefined });
 
       const space = selectActiveSpace(state);
-      expect(space).toBeUndefined();
+      expect(space).toStrictEqual(expect.objectContaining<Partial<Space>>({ _id: BASE_SPACE_ID }));
```
develohpanda, line 44: Tests to make sure the base space is returned in cases where we can't find the actual expected space

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -461,10 +461,22 @@ const writeExportedFileToFileSystem = (filename: string, jsonData: string, onDon
...
+  const workspaces = selectWorkspacesForActiveSpace(state);
+
+  if (!workspaces.length) {
+    dispatch(loadStop());
+    showAlert({
+      title: 'Cannot export',
+      message: <>There are no workspaces to export in the <strong>{activeSpaceName}</strong> space.</>,
+    });
+    return;
+  }
```
develohpanda, line 475: Demo linked in description

packages/insomnia-app/app/network/grpc/prepare.ts
develohpanda: image

packages/insomnia-app/app/common/database.ts
dimitropoulos: huh.  I fiddled earlier this morning but didn't see why either.

packages/insomnia-app/app/common/render.ts
```diff
@@ -54,10 +54,17 @@ export type HandleGetRenderContext = () => Promise<RenderContextAndKeys>;
...
-  ancestors: BaseModel[] | null,
-  rootEnvironment: Environment | null,
-  subEnvironment: Environment | null,
-  baseContext: Record<string, any> = {},
+  {
+    ancestors,
+    rootEnvironment,
+    subEnvironment,
+    baseContext = {},
+  }: {
```
dimitropoulos, line 62: I'm always happy to see us moving to this option-argument way

packages/insomnia-app/app/common/render.ts
```diff
@@ -373,24 +394,26 @@ export async function getRenderContext(
 
   baseContext.getEnvironmentId = () => environmentId;
 
+  // It is possible for a space to not exist because this
```
dimitropoulos: can you maybe reword this comment or clarify a bit more?

packages/insomnia-app/app/models/space.ts
```diff
@@ -8,15 +9,13 @@ export const prefix = 'sp';
...
-export const isBaseSpace = ({ _id }: Pick<Space, '_id'>) => _id === BASE_SPACE_ID;
-export const isNotBaseSpace = (space: Pick<Space, '_id'>) => !isBaseSpace(space);
-export const isLocalSpace = ({ remoteId }: Pick<Space, 'remoteId'>) => remoteId === null;
-export const isRemoteSpace = (space: Pick<Space, 'remoteId'>) => !isLocalSpace(space);
-export const spaceHasSettings = (space: Pick<Space, 'remoteId' | '_id'>) => isLocalSpace(space) && !isBaseSpace(space);
+export const isBaseSpace = ({ _id }: Space) => _id === BASE_SPACE_ID;
+export const isNotBaseSpace = (space: Space) => !isBaseSpace(space);
+export const isLocalSpace = ({ remoteId }: Space) => remoteId === null;
+export const isRemoteSpace = (space: Space) => !isLocalSpace(space);
+export const spaceHasSettings = (space: Space) => isLocalSpace(space) && !isBaseSpace(space);
```
dimitropoulos, line 18: was it intentional to remove these `Pick`s?  They were there before because we don't always actually have a full space to pass to these functions.  E.g. in space-dropdown we have a partial space that we work with.

packages/insomnia-app/app/network/grpc/__tests__/prepare.test.ts
```diff
@@ -10,6 +12,9 @@ import { prepareGrpcMessage, prepareGrpcRequest } from '../prepare';
 
 jest.mock('../../../common/render');
 
+const getRenderedGrpcRequest = mocked(_getRenderedGrpcRequest);
+const getRenderedGrpcRequestMessage = mocked(_getRenderedGrpcRequestMessage);
```
dimitropoulos, line 16: noice.

packages/insomnia-app/app/network/grpc/__tests__/prepare.test.ts
```diff
@@ -26,11 +31,12 @@ describe('prepareGrpcRequest', () => {
...
-        env,
-        RENDER_PURPOSE_SEND,
-        {},
-        false,
+        {
+          request: gr,
+          environmentId: env._id,
+          purpose: RENDER_PURPOSE_SEND,
+          skipBody: false,
+        },
```
dimitropoulos, line 39: I know I already commented on it, but the readability improvement here by using an object argument is stunning.  great improvement.

packages/insomnia-app/app/network/grpc/prepare.ts
dimitropoulos: wow.  just wow.

packages/insomnia-app/app/plugins/context/data.ts
dimitropoulos: see my comment above regarding migrations, but I think migrations to a new insomnia format may be the best way.  we can put the analytics call right there in that migration function (from v4 -> v5)

packages/insomnia-app/app/plugins/context/data.ts
dimitropoulos: I think I actually follow all of that.  OK.  cool.

packages/insomnia-app/app/sync/git/__tests__/ne-db-client.test.ts
dimitropoulos: So a thought on that... what would it matter if we left the parentId there?  It wasn't used for anything before, right?

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -109,25 +99,20 @@ export const SpaceDropdown: FC<Props> = ({ vcs }) => {
...
+
   return (
     <Dropdown renderButton={button} onOpen={refresh}>
-      <SpaceDropdownItem
-        isActive={baseSpace._id === activeSpace._id}
-        setActive={setActive}
-        space={baseSpace}
-      />
-
+      {spaces.filter(isBaseSpace).map(renderSpace)}
```
dimitropoulos, line 113: I know we gotta get the show on the road, but though I'd mention ramda's partition for this use case.  most definitely do not need to do here and now just thought it was a good example of where it's useful.

```ts
const [[baseSpace], ...otherSpaces] = partition(isBaseSpace, spaces)
```

packages/insomnia-app/app/ui/components/modals/index.ts
```diff
@@ -24,7 +23,7 @@ export function showPrompt(config: PromptModalOptions) {
   return showModal(PromptModal, config);
 }
 
-export function showAlert(config) {
+export function showAlert(config: AlertModalOptions) {
```
dimitropoulos, line 26: nice one.

packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
dimitropoulos: oh, I'm ALLL for it, don't worry.

packages/insomnia-app/app/ui/components/sidebar/sidebar-children.tsx
```diff
@@ -117,6 +120,7 @@ class SidebarChildren extends PureComponent<Props> {
             request={child.doc}
             hotKeyRegistry={hotKeyRegistry} // Necessary for plugin actions on requests
             activeEnvironment={activeEnvironment}
+            activeSpace={activeSpace}
```
dimitropoulos, line 123: I'm assuming there's a deep reason that I'm missing the explains why we can't get this from redux?

From what I can tell: all of these have it:

Screenshot_20210723_150322

If there's just one thing I'd like to happen in this PR it'd be to get this from redux in all of these cases.  I looked at them and none of them appear to be connected already, which is a bummer, but we gotta start somewhere I suppose.

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
dimitropoulos: with the context of where we were, where we are, and where we're going, this makes sense.  on the other hand if I step back and think about this it seems like this should throw an error in some way.  what is the (valid) situation in which there's no active space?

packages/insomnia-app/app/ui/redux/modules/workspace.ts
```diff
@@ -61,8 +60,7 @@ export const createWorkspace = ({ scope, onCreate }: {
             {
               name,
               scope,
-              // @ts-expect-error TSCONVERSION the common parentId isn't typed correctly
```
dimitropoulos, line 64: man.. this PR is a trip.  when you add it all up there were QUITE a few places where we were dancing around this ghost space thingy.  glad we're addressing it sooner rather than later!

dimitropoulos (approved):
This PR is going to be a real pain in the tuchus to rebase, so I don't want to wait on it.

However, I'd like to maybe start thinking about creating a new version of the insomnia format that accounts for spaces.  That way we can ensure there will never be dangling workspaces (i.e. with null parentId like before).  And that way we only need to check once upon booting the app (and then run a small migration if so) and then never again.

If you do only one thing before merging: it'd be great to connect some of those components to redux so we can grab the active space (rather than prop drilling).  Apologies if I missed a reason why it's currently happening.

packages/insomnia-app/app/models/space.ts
develohpanda: Not anymore, the concept of a partial space is gone 😉

packages/insomnia-app/app/ui/components/sidebar/sidebar-children.tsx
develohpanda: You are correct these just come from redux, but I was hesitant to do that here because the PR was already quite far reaching and I wanted to reduce the changeset by following the simplest approach.

I'll attach the leaf components to redux in an immediate follow-up PR

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
develohpanda: That'd be very useful! 🙌🏽 The more I come across use cases for ramda the more I want to use it

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
develohpanda: The only way someone can reach this situation is if they manually fiddle with the space value in local storage, there's no valid workflow through which someone can reach this state.

And we could show an error but working case by case, in this particular case we are able to self-recover (by showing the base space), so IMO we should self-recover.

packages/insomnia-app/app/sync/git/__tests__/ne-db-client.test.ts
develohpanda: It would matter - even though the `parentId` was not used for anything, the original value stored is `null`, and if we leave the `parentId` there depending on which space each individual user cloned into, git will always show changes. We need to resert this to what the previous value was.

See more context here: https://github.com/Kong/insomnia/blob/eb6795884660ef8a024af97af5d6c5857f0804c8/packages/insomnia-app/app/sync/ignore-keys.ts#L11-L17

packages/insomnia-app/app/plugins/context/data.ts
develohpanda: I do think there's a misunderstanding here and we don't need to move to a V5; this has nothing to do with the Insomnia export format but rather checking how frequently a fallback of "the plugin didn't provide a workspace so export all workspaces" is being used via this plugin API.

develohpanda:
> However, I'd like to maybe start thinking about creating a new version of the insomnia format that accounts for spaces. That way we can ensure there will never be dangling workspaces (i.e. with null parentId like before). And that way we only need to check once upon booting the app (and then run a small migration if so) and then never again.

I'm honestly a bit confused about what impact a new Insomnia export format has on migrating workspaces on launch. The parentId is set during the process of importing a V4 file.

The two are unrelated items and I don't want to overlap them at all so curious where this is coming from; what is the issue (is it still a valid issue?), and why is an Insomnia V5 format the solution for it?

packages/insomnia-app/app/common/render.ts
develohpanda: [`0716ac3` (#3819)](https://github.com/Kong/insomnia/pull/3819/commits/0716ac34b125a4206ea6a09c89402ec2f1b85a8e)

packages/insomnia-app/app/plugins/context/data.ts
develohpanda: https://linear.app/insomnia/issue/INS-823/add-metrics-to-track-how-often-plugin-export-fallback-logic-is-being

PR #3823, opened 2021-07-16
Implement color scheme badge artifact workaround.

johnwchadwick:
Color scheme badges have a small gap artifact caused by an interplay between anti-aliasing and alpha blending. Here, we use pseudo-classes to ensure one opaque pixel of intersection between the shadow border and the color badge.

### Explanation
For an explanation on _why_ this is necessary, please [see this codepen](https://codepen.io/john-chadwick/pen/mdmmWKG).

### Preview
This fix is not entirely perfect; a more rigorous fix might involve putting the box shadow itself into a pseudo-element and making it overlap that way. (Since box-shadows are always clipped by the shape of the element, this would be necessary to work around the problem.)
SHAIHRK2z2

### DIFF
--- a/packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
@@ -26,15 +26,34 @@ const ThemeButton = styled.div<{ $isActive: boolean; $isInOsThemeMode: boolean }
   margin: 'var(--padding-md) var(--padding-md)',
   fontSize: 0,
   borderRadius: 'var(--radius-md)',
-  boxShadow: '0 0 0 1px var(--hl-sm)',
   transition: 'all 150ms ease-out',
+  // This is a workaround for some anti-aliasing artifacts that impact the color scheme badges.
+  // The box shadow is placed on a pseudo-element. When it is active, it is configured to overlap
+  // 1px with the underlying geometry to prevent gaps caused by anti-aliasing.
+  '&:before': {
+    display: 'block',
+    position: 'absolute',
+    boxShadow: '0 0 0 1px var(--hl-sm)',
+    transition: 'all 150ms ease-out',
+    borderRadius: 'var(--radius-md)',
+    width: '100%',
+    height: '100%',
+    content: "''",
+    ...($isActive ? {
+      boxShadow: '0 0 0 var(--padding-xs) var(--color-surprise)',
+      width: 'calc(100% - 2px)',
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -45,6 +44,19 @@ const ThemeButton = styled.div<{ $isActive: boolean; $isInOsThemeMode: boolean }
       visibility: 'hidden', // this prevents alpha-blending problems with the underlying svg bleeding through
     },
   } : {}),
+  // This is a workaround for some anti-aliasing artifacts that impact the color scheme badges.
+  ...($isActive && $isInOsThemeMode ? {
+    '&:not(:hover):before': {
```
reynolek: magic

develohpanda (approved):
That codepen blew me away. Nice one, thanks for fixing this! 🙌🏽 @dimitropoulos I can't seem to find whether a Linear ticket was created for this issue?

johnwchadwick:
Apologies, but I've re-requested review. There were some visual glitches during transitions that were bugging me enough to revisit this and implement the proposed better solution in the original comment (putting the actual box shadow on the pseudo element.)

Animation of improved workaround.

This fixes several artifacts:

- The noticeable jump when switching between OS color scheme mode and back.
- The workaround border being visible during theme transitions, as it was not synchronized with the rest of the animation. There is no longer a separate 'workaround' border.
- The desync between the color badge and the outline when switching themes. This one can actually be fixed without this, but it still was bugging me.
- The border radius is no longer impacted when in OS color scheme mode.

It is still not totally perfect, but at least the remaining artifacts are _a lot_ less noticeable.

dimitropoulos (approved):
tested it again, looks and works great!

johnwchadwick:
OK, OK! One last minor tweak.

The animation when hovering looks a little bit janky. To illuminate, this is what it looks like in 25% speed:

Ring hover animation before

This happens because the ring fades out, but it would probably be better if it only transitioned in spread and not color. In my opinion, it's a pretty big gain for a one-line tweak.

Ring hover animation after

The one-line tweak is that the `:hover:before` rule now specifies a zero-width `box-shadow` instead of `none`.

dimitropoulos (approved):
(that last change was my fault, haha, we got talking about it while waiting for the CI before merging and it's so easy to fix that it was worth waiting for the CI one last time).

develohpanda:
This is amazing @johnwchadwick 😍

PR #3824, opened 2021-07-16
UI for Local/Remote spaces

dimitropoulos:
This ended up being significantly more complicated than I had hoped because... we aren't really setting ourselves up for success for custom styling for these components.  I tried not to veer too far off the trail: but there were two things that must be done to complete this task: the ability to style the dropdown items/svgs/whatever, and the selector for determining whether an workspace is remote.


packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
```diff
@@ -29,7 +29,8 @@ export const RemoteWorkspacesDropdown: FC<Props> = ({ className, vcs }) => {
     pull,
   } = useRemoteWorkspaces(vcs || undefined);
 
-  const isRemoteSpace = Boolean(useSelector(selectActiveSpace)?.remoteId);
+  const space = useSelector(selectActiveSpace);
+  const isRemoteSpace = useSelector(selectIsRemoteSpace(space?._id));
```
develohpanda: We should IMO use utilities instead of depending on redux. Going to redux (even though the selector is cached) actually results in an additional lookup which is unnecessary, but also means we can't reuse this check outside of a context where redux is available.

```ts
export const isLocalSpace = ({ remoteId }: Space) => remoteId === null;
export const isRemoteSpace = (space: Space) => !isLocalSpace(space);
```

In #3819 I introduce the following https://github.com/Kong/insomnia/pull/3819/files?file-filters%5B%5D=.ts#diff-d5e2124c0a2d0e0dff2eac856fa011955d696ffbbcc322afcc7a4ee351f8b858R29-R32

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -1,64 +1,106 @@
...
+const StyledSvgIcon = styled(SvgIcon)({
+  // this is a bit of a hack/workaround to avoid some larger changes that we'd need to do with dropdown item icons
+  marginTop: 1,
+  '& svg': {
+    fill: 'var(--hl)',
+  },
+});
+
+const TooltipIcon = ({ message, icon }: { message: string, icon: SvgIconProps['icon'] }) => (
+  <Tooltip message={message} style={{ display: 'flex' }}>
```
develohpanda: The comments from above need to go here for this style object as well, right?

I think it might be worth memoizing this style object with `useMemo` as well. If we don't, `Tooltip` will see it as a unique prop on every render

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -1,64 +1,106 @@
...
   _id: BASE_SPACE_ID,
   name: getAppName(),
   remoteId: null,
 };
 
-const check = <i className="fa fa-check" />;
-const cog = <i className="fa fa-cog" />;
-const plus = <i className="fa fa-plus" />;
+const Checkmark = styled(SvgIcon)({
+  // this is a bit of a hack/workaround to avoid some larger changes that we'd need to do with dropdown item icons
```
develohpanda: Also worth adding what it's a hack _for_. (ie. this is a hack to align the icon correctly with the text)

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -1,64 +1,106 @@
...
+
+const TooltipIcon = ({ message, icon }: { message: string, icon: SvgIconProps['icon'] }) => (
+  <Tooltip message={message} style={{ display: 'flex' }}>
+    <StyledSvgIcon icon={icon} />
+  </Tooltip>
+);
+
 const spinner = <i className="fa fa-spin fa-refresh" />;
-const home = <i className="fa fa-home" />;
+const home = <TooltipIcon message="Base Space (Always Local)" icon="home" />;
```
develohpanda, line 60: TBC on the wording for this mesage I think, but good to have it

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -1,64 +1,106 @@
...
+const remoteSpace = <TooltipIcon message="Remote Space" icon="globe" />;
+const localSpace = <TooltipIcon message="Local Space" icon="laptop" />;
 
 interface Props {
   vcs?: VCS;
 }
 
+const BoldDropdownItem = styled(DropdownItem)({
+  fontWeight: 500,
```
develohpanda: A few things here:

1. Magic numbers make me wary; is there an existing CSS var we can use? 
2. Why not only make the `{name}` have a fontWeight, rather than the entire `DropdownItem`. Does the fontWeight also apply to SVGs?
3. (most important) Is this to match 1:1 with the designs? IMO it's better to be consistent with what is in the app than with the designs. If we want dropdowns to be bold, we should consider all dropdowns to be bold.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -1,64 +1,106 @@
...
+    _id: spaceId,
+    name,
+  },
+}) => {
+  const dispatch = useDispatch();
+  const setActive = useCallback((id: string) => dispatch(setActiveSpace(id)), [dispatch]);
+  const isRemote = useSelector(selectIsRemoteSpace(spaceId));
+
+  const activeSpace = useSelector(selectActiveSpace);
+  const selectedSpace = activeSpace || baseSpace;
```
develohpanda: Not liking this duplication between `SpaceDropdown` and `SpaceDropdownItem`; the reason `SpaceDropdownItem` initially wasn't extracted into a separate component was because it uses a lot of info from the parent component `SpaceDropdown` already.

If we don't want to use the memoized render function IMO we should send data to `SpaceDropdownItem` as props rather than re-fetch it, so there is a single source of truth.

(herein lies the question about what is too little and too much prop-drilling, but this isn't drilling props between unrelated components, it's just making `SpaceDropdownItem` a bit dumber and using the single source of truth as defined by it's parent)

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -1,64 +1,106 @@
...
+      icon={isBaseSpace ? home : isRemote ? remoteSpace : localSpace}
+      right={isActiveSpace && <Checkmark icon="checkmark" />}
+      value={spaceId}
+      onClick={setActive}
+    >
+      {name}
+    </BoldDropdownItem>
+  );
+};
+SpaceDropdownItem.displayName = 'DropdownItem';
```
develohpanda: I'm fairly opposed to a refactor that requires setting this `displayName` because it makes the code brittle.

Now, instead of only consuming the exported `DropdownItem` component from `insomnia-components`, a developer needs to know the inner workings of the dropdown in the component library to know they need to set this (or why they cannot remove this line). If the implementation ever changed in the component library (which is a totally valid scenario), or the exported component name ever changed, the `SpaceDropdown` would break.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
```diff
@@ -69,18 +111,26 @@ export const SpaceDropdown: FC<Props> = ({ vcs }) => {
...
+        <DropdownItem icon={<StyledSvgIcon icon="gear" />} onClick={showSettings}>
           {strings.space.singular} Settings
         </DropdownItem>
       </>}
+
       <DropdownDivider />
-      <DropdownItem icon={plus} onClick={createNew}>
-        Create new {strings.space.singular.toLowerCase()}
+      <DropdownItem icon={<StyledSvgIcon icon="plus" />} onClick={createNew}>
+        Create or join a {strings.space.singular.toLowerCase()}
```
develohpanda: There's no way to "join" a space; this line should be reverted (even though, I am aware, the designs have it like this).

packages/insomnia-components/src/dropdown/dropdown-item.tsx
```diff
@@ -126,7 +126,7 @@ export class DropdownItem extends PureComponent<DropdownItemProps> {
     );
     return (
       <StyledButton
-        className={buttonClass}
+        className={`${className} ${buttonClass}`.trim()}
```
develohpanda: Any reason for not using classnames? `classnames(classname, buttonClass)`

packages/insomnia-components/src/dropdown/dropdown.tsx
```diff
@@ -15,6 +15,7 @@ import { DropdownItem } from './dropdown-item';
 import { DropdownDivider } from './dropdown-divider';
 import styled, { css } from 'styled-components';
 import { SvgIcon } from '../svg-icon';
+import { any, equals } from 'ramda';
```
develohpanda, line 18: :heart_eyes: nice use of it in this file

packages/insomnia-components/src/svg-icon.tsx
```diff
@@ -1,50 +1,53 @@
-import React, { Component, NamedExoticComponent, ReactNode, SVGProps } from 'react';
+import React, { CSSProperties, Component, NamedExoticComponent, ReactNode, SVGProps } from 'react';
```
develohpanda, line 1: Unfortunately it's really hard to review this file.

Until we don't have an automated lint rule and an autofix committed in develop, as a reviewer, when reordering like this (which makes the diff extremely verbose) sneaks in alongside a feature, it's really hard to review anything in a confident manner.

develohpanda (changes requested):
A good start, but there's some implementation patterns here that are a bit brittle and we should look to improve them.

Excited to have ramda in here (albeit only in `insomnia-components` for now) :raised_hands:

packages/insomnia-components/src/svg-icon.tsx
dimitropoulos: gotcha

f5b6a6d7b0ad800219ce43e8c45b47c3bd23682a reverts it to exactly what it was before this PR
257262c200d10b7d8e8c9443479c710bb2993481 does the sorting and nothing else
ca18b2cbd68d98fb70c36c7dff2a928b13e22d35 is just the changes for this PR

I think that resolves your request here, but if I am mistaken please just unresolve and lemme know

packages/insomnia-components/src/dropdown/dropdown-item.tsx
dimitropoulos: interesting.  so 
image

that was the first thing I tried, but it didn't show up in intellisense so I just went ahead assuming that meant it wasn't in this project.

Turns out it's those dang default exports again.  It doesn't come up because of that.  I'm glad default exports are a fad that is fading away because it creates time-wasters like this (well, for me it does, anyway).  All the same.  fixed in a9b7d2dca9107825108101bf12a7958148bf2ebf

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
dimitropoulos: I don't agree but actually I think you're right about your argument here, haha, if that makes any sense.

Totally a reasonable request: I'll change it to use the newly added helpers (that didn't exist when this PR was created).

To put what I'm saying another way: I don't object to these helpers existing, but that if we're going to be grabbing something from the redux state we should be using redux selectors.  Then that starts a whole conversation about how everything's not yet in redux and so on and so on.  So doing it the way you're suggesting is quite reasonable for where this codebase is right now.  I hear yah.

dimitropoulos: image

I tried to copy it exactly from that other PR (such that there wouldn't be a conflict later, but I get the above error.  I modified the selectors to just need the thing they check for and it's all good now.  Should be a totally compatible change with any other usages.

Fixed in 1057087df94033afda667eba8cc500e6073e9b26

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: Yep, I was following the design.  Ok, I'll change it back.  @erickrico could you please update the design (or @wdawson let us know if there's other language you'd like).

4c1d6c34d5ccbf3e166b3c177c51d8b0c938b627

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: Gotcha.  Fixed in 34c12545d1f928eb7754df4cc48f3a85d31bb839

By the way, are the same problem, then? https://github.com/Kong/insomnia/blob/develop/packages/insomnia-smoke-test/modules/dropdown.ts#L6 and https://github.com/Kong/insomnia/blob/develop/packages/insomnia-smoke-test/modules/debug.ts#L153.

----

However, in case I'm misunderstanding you and you are talking about the behavior where `Dropdown` will completely disregard any component contained within that isn't `DropdownItem` or `DropdownDivider`, that behavior already existed here: https://github.com/Kong/insomnia/blob/develop/packages/insomnia-components/src/dropdown/dropdown.tsx#L493.  In fact, it was a quite big problem for me and I found it very non-obvious (like you're saying) and was the reason why I needed to do this in order to style it.  I checked a few of the top UI libraries (AntD, Blueprint, Material-UI) and they don't silently remove children like we are.  I think we need to address this at some point in some way, but I didn't want to get deep into it in this PR so I just did what I minimally needed to do to allow it to work.

One way or the other, right now, you have to be aware of the innerworkings of DropdownItem.  My fix was to mainly address your very-valid point about SpaceDropdown breaking if DropdownItem's name changes.  That's no longer the case.

dimitropoulos: (please unresolve if you'd like to talk more - seems like this is otherwise the solution to what you were asking for)

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: > Not liking this duplication between SpaceDropdown and SpaceDropdownItem;

I don't love it either, read on for the reason.

> the reason SpaceDropdownItem initially wasn't extracted into a separate component was because it uses a lot of info from the parent component SpaceDropdown already.

Right, but the reason it needs to be now is that it needs to be a component ([it wasn't before](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx#L50)) because it's using hooks and you can't use hooks in a callback or a loop.

____

All the same, I did the best I could to cut down on any duplication.  624aadef8e5bd91511713d0016e3de1794229ff5

I'm not super happy with this because it, like I sorta expected it might, opened a bit of a can of worms with DropdownItem.  

Lemme know what you think.  I'd like to contain the scope as much as possible - that was a large portion of the motivation for how it was.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: > (most important) Is this to match 1:1 with the designs?

Yes, absolutely.  That was the only reason.

Let's figure out if we are going to follow the design or not and if not then we can just skip the search for a CSS var (which, I agree with you, we should be using here).

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: great point: 43031a917a891b7e0f24b16ed6d5f1ec4e97af93 I removed the duplication while I was looking at it, too.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
erickrico: designs updated

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: I'm not clear on how that could cause a rerender.  Could you say a bit more?  As it stands this component will only rerender when `message` or `icon` changes, would it not?  If I'm mistaken about that can you clarify how a rerender would occur when `message` and `icon` have not changed?

That said, I agree with your deeper point.  I think it's unacceptably sloppy to do it the way I had it before.  I changed the approach in 63a5b819355e1b326a9f4dfcc18ef19949418725.  It wasn't at all clear before how critical it was that `style` is used (rather than regular styled components or css) to work around css specificity problems without doing more surgery on dropdown item itself.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: same.  I pinged @wdawson about this.  We'll go with this for now and change at any moment he requests (before this merges or after).

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: I touched base with @erickrico and he updated the designs.  it's now gone in 4369270bf239adeee5e66d7f38ebbd4213628e66.

packages/insomnia-app/app/ui/components/dropdowns/remote-workspaces-dropdown.tsx
develohpanda: > but that if we're going to be grabbing something from the redux state we should be using redux selectors. 

I disagree in this instance.

In this instance we are doing two things:
1. fetching the active space
2. determining whether a space is remote

The first task requires us to fetch from redux, but the second does not; the second is just checking a property on that already fetched entity and I don't see why that requires a round-trip though redux, to fetch the space _again_ and then determine the state of a property on it.

> Then that starts a whole conversation about how everything's not yet in redux and so on and so on.

This doesn't start a convo about things being in/not in redux, because the data _is_ entirely in redux; it's the operations (changing the data) that don't occur through redux (but they update redux through the database hooks that exist). That's an different conversation which has been confused here.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
develohpanda: Here's a demonstration as to why the prop causes a re-render: https://www.loom.com/share/17b073ceadf34948952847078f3e4916, link to the sandbox in the video description.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
develohpanda: I feel like refactoring this to extract a component was extended scope to begin with. It's not a question about whether hooks are being used, but rather how the data is being prepared, which, regardless of a memoized HoC or a standalone component, should come from the same place. 

IMO what you've done is fine but I suggest `SpacesDropdownItem` having the following props and not sending the entire `activeSpace` object - it now, essentially, becomes a controlled component.

```diff
{
  space: SpaceSubset;
-  activeSpace: SpaceSubset;
+  isActive?: boolean;
  setActive: (spaceId: string) => void;
}
```

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
develohpanda: > However, in case I'm misunderstanding you and you are talking about the behavior where Dropdown will completely disregard any component contained within that isn't DropdownItem or DropdownDivider, that behavior already existed here: develop/packages/insomnia-components/src/dropdown/dropdown.tsx#L493.

Yes I think there's a misunderstanding here. I know this is the behavior and that's specifically why I think we should be using the components made available through the component library instead of hijacking it _knowing_ how the `Dropdown` works internally; because it's undocumented behavior. 

For example, here's the documentation for [react-tabs](https://github.com/reactjs/react-tabs#custom-components), which talks about the requirements for custom components. I'm fairly confident that the pattern (but not the implementation) for the dropdown in our code is heavily inspired from react-tabs.

To your point, react-tabs also doesn't silently remove children, and I agree that's not the right approach, but the differentiator is all of those custom frameworks (incl. react-tabs) explains the requirements for creating your own custom components/children. 

I think what you've done is fine; at least calling out "here be dragons" and explaining why a deviation from the norm is needed 👍🏽 

Edit: interestingly, it looks like `react-tabs` also follows a similar approach but instead of the component `displayName` it checks for `role`. See [helpers](https://github.com/reactjs/react-tabs/blob/main/src/helpers/elementTypes.js) and how they are [consumed](https://github.com/reactjs/react-tabs/blob/e6d3989709bc9a7e00ddc1bab4d5c30110475658/src/components/UncontrolledTabs.js#L195).

develohpanda: > By the way, are the same problem, then? develop/packages/insomnia-smoke-test/modules/dropdown.ts#L6 and develop/packages/insomnia-smoke-test/modules/debug.ts#L153.

Not the same problem, but good question. The reason it is not a problem here is because the smoke-tests will huff and puff and fail if the drop-down implementation was to change and an element it expected to find, could not be found. And that's to be expected, smoke-tests need to find and access elements, one way or another, and if those elements change then the smoke tests will fail and also need to change, and that's fine - tests fail because the test suite is out of date or the behavior it is testing is broken.

packages/insomnia-components/src/svg-icon.tsx
```diff
@@ -116,11 +122,13 @@ export type IconId = ValueOf<typeof IconEnum>;
...
   icon: IconId;
   label?: ReactNode;
+  className?: string;
+  style?: CSSProperties;
 }
 
 const SvgIconStyled = styled.div<{
-  theme: ThemeKeys;
-  hasLabel: boolean;
+  $theme: ThemeKeys;
```
develohpanda, line 130: Out of curiosity, is the `$` a new pattern/approach described by styled-components or just a convention?

packages/insomnia-components/src/svg-icon.tsx
develohpanda: Thank you for separating 🙏🏽

packages/insomnia-components/src/dropdown/dropdown-item.tsx
```diff
@@ -1,20 +1,36 @@
...
+
+interface UncontrolledDropdown {
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+}
+
+const isControlledInput = <T extends unknown>(dropdown: DropdownItemProps<T>): dropdown is ControlledDropdown<T> => (
+  Object.prototype.hasOwnProperty.call(dropdown, 'value')
+);
+
+const isUncontrolledInput = <T extends unknown>(dropdown: DropdownItemProps<T>): dropdown is ControlledDropdown<T> => (
```
develohpanda: Should this be the following? 
```suggestion
const isUncontrolledInput = <T extends unknown>(dropdown: DropdownItemProps<T>): dropdown is UncontrolledDropdown => (
```

packages/insomnia-components/src/dropdown/dropdown-item.tsx
```diff
@@ -98,10 +114,10 @@ export class DropdownItem extends PureComponent<DropdownItemProps> {
...
     }
 
-    if (this.props.hasOwnProperty('value')) {
-      onClick(this.props.value, event);
-    } else {
-      onClick(event);
+    if (isControlledInput<T>(this.props)) {
+      this.props.onClick?.(this.props.value, event);
+    } else if (isUncontrolledInput(this.props)) {
+      this.props.onClick?.(event);
```
develohpanda, line 120: I tried to change this to the following, interesting how it causes type errors.

(This is not a suggestion, just showing the diff)
```suggestion
    if (isControlledInput<T>(this.props)) {
      onClick(this.props.value, event);
    } else if (isUncontrolledInput(this.props)) {
      onClick(event);
```

develohpanda (approved):
Small note [here](https://github.com/Kong/insomnia/pull/3824#discussion_r674465486) and a question about the types of `isUncontrolledInput` - aside from that good to go!

image

(I pushed [`12402a9` (#3824)](https://github.com/Kong/insomnia/pull/3824/commits/12402a957d79285739a722b584f420da58caf6b2) to remove a redundant ts-expect-error)

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: > interestingly, it looks like react-tabs also follows a similar approach but instead of the component displayName it checks for role. See helpers and how they are consumed.

That is very interesting, indeed.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
dimitropoulos: > I feel like refactoring this to extract a component was extended scope to begin with.

I had to extract because I had to use a selector which at the time was in redux which required hooks (which requires extraction).  That was the reason it was specifically within scope in the context of what had been already done (again, this was before your work on the non-redux selectors - which ended up being useful).

----

That aside, I really like this suggestion and now that you make it it seems crystal-clear that it should be done.  I did so in 8d2914d7b4c05153538136ccceda369b70bb3af2.

packages/insomnia-components/src/dropdown/dropdown-item.tsx
dimitropoulos: YEP. heh.

YOU THINK YOU KNOW A GUY (the guy being this code)..

dimitropoulos: fixed in 5d800a3dd5dd111196cd80708b4e1a789e5c7a91

packages/insomnia-components/src/dropdown/dropdown-item.tsx
dimitropoulos: I, as well, had to wrestle with it... A LOT.

This particular part, though, is because of a problem with TypeScript below 4.4.  In 4.4 they added https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#cfa-aliased-conditions which _I would expect_ solves this problem.

packages/insomnia-app/app/ui/components/dropdowns/space-dropdown.tsx
wdawson: Can you link me to the designs?

PR #3836, opened 2021-07-23
Create VCS instance even if no active workspace

develohpanda:
A regression was introduced when removing the need for an active workspace to always exist.

When updating the VCS instance, we shouldn't prevent an instance from being created if no active workspace exists (eg while on the homepage). we should only prevent switching to a project if no workspace is active. Additionally, if no workspace is active we should explicitly clear the project.

If on the dashboard after a first launch, remote spaces would fail to load, as a result of this regression. It is now fixed by this PR.


packages/insomnia-app/app/sync/vcs/vcs.ts
```diff
@@ -90,6 +90,10 @@ export class VCS {
     this._project = null;
   }
 
+  clearProject() {
+    this._project = null;
```
develohpanda, line 94: @DMarby I _think_ it's safe for us to not clear the project from `this._store` given we don't do that when switching projects (see method below). What are your thoughts?

DMarby: I think this makes sense, yep.

DMarby (approved):

PR #3843, opened 2021-07-23
add: support for plain PKCE challenge method

### LINKED ISSUE #2973: Add Plain code challenge method in OAuth 2.0 with PKCE
The [OAuth 2.0 with PKCE flow specification](https://tools.ietf.org/html/rfc7636) says in 4.2 Section that the Code Challange could be created using two algorithms:

> 4.2.  Client Creates the Code Challenge
   The client then creates a code challenge derived from the code
   verifier by using one of the following transformations on the code
   verifier:
   plain
      code_challenge = code_verifier
   S256
      code_challenge = BASE64URL-ENCODE(SHA256(ASCII(code_verifier)))

By default Insomnia seems to use only the S256 transformation and some sites offer only the Plain transformation and that provokes that we can not retrieve the token(s).

A possible solution could be that you could be to select the Code Challenge Method to be either Plain or SHA-256 if you select the "Use PKCE" checkbox.

Now, the only way to retrieve the token in a server that only implements Plain code challenge method with Insomnia is create manually the requests of OAuth 2.0 Flow.

Thanks.

aeri:
In reference to the Issue **Add Plain code challenge method in OAuth 2.0 with PKCE #2973**

Closes #2973


packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.ts
```diff
@@ -84,6 +93,7 @@ async function _authorize(
   audience = '',
   resource = '',
   codeChallenge = '',
+  pkceMethod = '',
```
develohpanda, line 95: We should (in a separate PR) change this to an object param rather than marking things as optional/with defaults when the default isn't actually valid (pkceMethod cannot be an empty string)

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.tsx
```diff
@@ -491,6 +514,11 @@ class OAuth2Auth extends PureComponent<Props, State> {
         redirectUri,
         enabled,
       ];
+
+      if (authentication.usePkce){
+          basicFields.splice(5, 0, pkceMethod);
+      }
```
develohpanda: While I think we should use a slightly different approach to insert the code challenge dropdown I'd like to get product approval on the UX first.

Should the challenge dropdown be removed or _disabled_ if PKCE is not enabled? It causes the UI to jump around. @wdawson / @erickrico what would you suggest?

develohpanda (commented):
Thanks for the PR! A question for our PM but overall implementation looks good.

packages/insomnia-app/app/ui/components/editors/auth/o-auth-2-auth.tsx
wdawson: It would be nice to get rid of the jump. Assuming if we _disabled_ the dropdown, we wouldn't have the snapping?

develohpanda: Yes that's right, no snapping if the dropdown is disabled.

develohpanda (approved):
Thanks for the PR 🙌🏽 QA'd the UX and pushed a commit to fix conflicts and the jumpy behavior

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.ts
aeri: Perfect, should I make any modifications?

develohpanda: No need in this PR, but if you'd like to give it a go in a separate PR that would be awesome! 😄

aeri: Great, I'll wait for this PR to be resolved then.

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.ts
```diff
@@ -37,8 +38,15 @@ export default async function(
...
     // @ts-expect-error -- TSCONVERSION
     codeVerifier = _base64UrlEncode(crypto.randomBytes(32));
-    // @ts-expect-error -- TSCONVERSION
-    codeChallenge = _base64UrlEncode(crypto.createHash('sha256').update(codeVerifier).digest());
+
+    if (pkceMethod == c.PKCE_CHALLENGE_S256){
+      // @ts-expect-error -- TSCONVERSION
+      codeChallenge = _base64UrlEncode(crypto.createHash('sha256').update(codeVerifier).digest());
+    }
+    else{
```
dimitropoulos:
```suggestion
    } else {
```

just fixing this via the GitHub UI in anticipation of the CI blocking on https://github.com/Kong/insomnia/pull/3859 which looks like it'll be merged right before this.

packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.ts
```diff
@@ -37,8 +38,14 @@ export default async function(
   if (usePkce) {
     // @ts-expect-error -- TSCONVERSION
     codeVerifier = _base64UrlEncode(crypto.randomBytes(32));
-    // @ts-expect-error -- TSCONVERSION
-    codeChallenge = _base64UrlEncode(crypto.createHash('sha256').update(codeVerifier).digest());
+
+    if (pkceMethod == c.PKCE_CHALLENGE_S256){
```
dimitropoulos:
```suggestion
    if (pkceMethod == c.PKCE_CHALLENGE_S256) {
```

dimitropoulos (approved):

PR #3844, opened 2021-07-23
Make Swagger 2 importer generate deterministic IDs

johnwchadwick:
This is a backport of a feature that was added to the OpenAPI 3.0 importer two years ago in 3e2311e4. As it is now, the placeholder ID gets replaced by a UUIDv4 identifier, which is pseudorandom and therefore not deterministic. This is problematic because the importer will not have any obvious way to determine identity between requests, and so it duplicates them every time.

This isn't perfect, because requests whose identities change or are removed will not be updated as expected, but it will bring Swagger/OpenAPI 2.0 specs to parity with how OpenAPI 3.0 specs work.


develohpanda:
Duplicate of https://github.com/Kong/insomnia/pull/3804 (I think?)

johnwchadwick:
Closed as it is indeed a duplicate.

PR #3847, opened 2021-07-26
[followup] Local GraphQL schema

dimitropoulos:
This is a followup from https://github.com/Kong/insomnia/pull/3803. Closes INS-814.

Each commit links to the thing it followed up on in the commit description.

### DIFF
--- a/packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
+++ b/packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
@@ -342,15 +342,15 @@ class GraphQLEditor extends PureComponent<Props, State> {
       ],
     };
 
-    const { canceled, filePaths: paths } = await electron.remote.dialog.showOpenDialog(options);
+    const { canceled, filePaths } = await electron.remote.dialog.showOpenDialog(options);
 
     if (canceled) {
       return;
     }
 
     try {
-      const path = paths[0]; // showOpenDialog is single select
-      const file = readFileSync(path);
+      const filePath = filePaths[0]; // showOpenDialog is single select
+      const file = readFileSync(filePath);
 
       const content = JSON.parse(file.toString());
       if (!content.data) {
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -342,33 +337,42 @@ class GraphQLEditor extends PureComponent<Props, State> {
...
       const content = JSON.parse(file.toString());
       if (!content.data) {
         throw new Error('JSON file should have a data field with the introspection results');
       }
 
-      newState.schema = buildClientSchema(content.data);
-      newState.schemaLastFetchTime = Date.now();
+      if (this._isMounted) {
+        return;
+      }
```
develohpanda: Shouldn't this be

```suggestion
      if (!this._isMounted) {
        return;
      }
```

develohpanda:
Were we not planning on putting in dropdown-dividers between the docs/remote/local options?

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
dimitropoulos: um... YES!!

fixed in 2b8af1cd781f3e8cc377d637cc18c19d10b1ac97

dimitropoulos:
ahh, right, right, dividers added in https://github.com/Kong/insomnia/pull/3847/commits/41acf622471e99798a01ee2550d777a8229dd005, but I'm not 100% remembering if this is what we talked about.

Screenshot_20210727_142910

develohpanda (approved):
LGTM! Both the error and success states look to be working. The dividers are what we discussed as well (noted in the linear ticket) 👍🏽

gatzjames (approved):
From a UX pov I also feel that figuring out which schema is being used is a bit confusing (local/remote) imo.
Other than that I couldn't find anything else except a bug in the variables editor.
LGTM!

PR #3778, opened 2021-07-01
🚀 Hotfix 2021.4.1 RC

develohpanda:
Closes #3777 

https://github.com/Kong/insomnia/compare/core@2021.4.0...core@2021.4.1

PR #3850, opened 2021-07-27
[feat] Sort Dashboard Cards

gatzjames:
Adds the ability to sort the dashboard cards by
- Last Modified
- Name (Ascending/Descending)
- Newest/Oldest

The UI is similar to the one used in the Collection list and the selected option is persisted in the global state.

image

Closes: INS-543


packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+        return cardB.lastModifiedTimestamp - cardA.lastModifiedTimestamp;
+      case 'NameAscending':
+        return cardA.workspace.name.localeCompare(cardB.workspace.name);
+      case 'NameDescending':
+        return cardB.workspace.name.localeCompare(cardA.workspace.name);
+      case 'DateCreatedAscending':
+        return cardA.workspace.created - cardB.workspace.created;
+      case 'DateCreatedDescending':
+        return cardB.workspace.created - cardA.workspace.created;
+      default:
```
dimitropoulos, line 76: seems like we need to add `DateModifiedAscending` to `SpaceItemsSortOrder` and include that case here and use `unreachableCase`.

```ts
...
      case 'DateCreatedDescending':
        return cardB.workspace.created - cardA.workspace.created;
      case 'DateModifiedAscending':
        return cardB.lastModifiedTimestamp - cardA.lastModifiedTimestamp;
      default:
        unreachableCase(orderBy, `space card ordering "${orderBy}" is invalid`);
    }
...
```

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+      </DropdownItem>
+      <DropdownItem value="NameDescending" onClick={onSelect}>
+        Name Descending
+      </DropdownItem>
+      <DropdownItem value="DateCreatedAscending" onClick={onSelect}>
+        Oldest First
+      </DropdownItem>
+      <DropdownItem value="DateCreatedDescending" onClick={onSelect}>
+        Newest First
+      </DropdownItem>
```
dimitropoulos: @develohpanda what are your thoughts on the approach taken here (i.e. not mapping a data structure).  @gatzjames and I had already talked about it but I'm curious what you think.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+  | 'modifiedLocally'
+  | 'lastCommitTime'
+  | 'lastCommitAuthor'
+  | 'lastActiveBranch'
+  | 'spec'
+  | 'specFormat'
+  | 'apiSpec'
+  | 'specFormatVersion'
+  | 'workspace'
+> | null {
```
dimitropoulos: Is there a reason to specify all of this?  seems like typescript infers it just fine?

Screenshot_20210727_153720

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -349,22 +471,44 @@ class WrapperHome extends PureComponent<Props, State> {
...
+          workspaceMetas,
+          apiSpecs,
+          filter,
+        }),
+      )
       .filter(isNotNullOrUndefined)
-      // @ts-expect-error -- TSCONVERSION appears to be a genuine error
-      .sort((a: RenderedCard, b: RenderedCard) => descendingNumberSort(a.lastModifiedTimestamp, b.lastModifiedTimestamp))
-      .map(c => c?.card);
+      .sort(orderSpaceCards(activeSpace?.order || 'DateModifiedDescending'))
```
dimitropoulos: `DateModifiedDescending` if this is truly the default, then should it be handled in the `default` of the case statement inside `orderSpaceCards`?

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -349,22 +471,44 @@ class WrapperHome extends PureComponent<Props, State> {
...
-    const { workspaces, isLoading, vcs } = this.props.wrapperProps;
+    const { workspaces, isLoading, vcs, activeSpace, workspaceMetas, apiSpecs } =
+      this.props.wrapperProps;
     const { filter } = this.state;
     // Render each card, removing all the ones that don't match the filter
     const cards = workspaces
-      .map(this.renderCard)
+      .map((workspace) =>
+        mapWorkspaceToWorkspaceCard({
+          workspace,
```
dimitropoulos: up to you, but if this is the way it's designed to be used (mapping on workspaces) it may be more natural/better to do

```ts
const mapWorkspaceToWorkspaceCard = ({
  apiSpecs,
  workspaceMetas,
}: {
  apiSpecs: ApiSpec[];
  workspaceMetas: WorkspaceMeta[];
}) => (workspace: Workspace) =>  {
```

which will enable us to compose this better e.g.
```ts
      .map(mapWorkspaceToWorkspaceCard({
        workspaceMetas,
        apiSpecs,
      }))
```

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+        Newest First
+      </DropdownItem>
+    </Dropdown>
+  );
+}
+
+interface MapWorkspaceToWorkspaceCardInput {
+  apiSpecs: ApiSpec[];
+  workspaceMetas: WorkspaceMeta[];
+  filter: string;
```
dimitropoulos: it appears that `filter` is unused.

see https://github.com/Kong/insomnia/pull/3850/files#r677748563 for more context, but this kind of thing is one of the reasons I find it better to do these kinds of things inline because it's much harder to make the mistake of leaving something there that isn't actually needed (anymore).

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+  lastCommitAuthor?: string | null;
+  modifiedLocally?: number;
+  spec: Record<string, any> | null;
+  specFormat: 'openapi' | 'swagger' | null;
+  specFormatVersion: string | null;
+  hasUnsavedChanges: boolean;
+  onSelect: (workspaceId: string, activity: GlobalActivity) => void;
+}
+
+function WorkspaceCard(props: WorkspaceCardProps) {
```
dimitropoulos: is there any way we can move this component out of `wrapper-home`?  I'd really love to hear that we can, if we can.

we're trying to reduce all the "single responsibility" breakage in this file.  I know some of this was already here, but this is a fine time to correct course in the direct context of what you're working on.

packages/insomnia-app/app/models/space.ts
```diff
@@ -20,6 +27,7 @@ export const spaceHasSettings = (space: Space) => isLocalSpace(space) && !isBase
 interface BaseSpace {
   name: string;
   remoteId: string | null;
+  order: SpaceItemsSortOrder;
```
dimitropoulos: is `order` really clear enough? are we going to regret this later when some other aspect of spaces needs to be ordered, you know?

I was gonna suggest `dashboardOrder` but then I realized that that's a bad idea because I absolutely expect this to be at least _potentially_ usable for more things that just the dashboard (e.g. in a dropdown of spaces to select, maybe, at some time).

@develohpanda any ideas?

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+  const workspaceModified = workspaceMeta?.modified;
+
+  const modifiedLocally = isDesign(workspace) ? apiSpec.modified : workspaceModified;
+
+  // Span spec, workspace and sync related timestamps for card last modified label and sort order
+  const lastModifiedFrom = [
+    workspace?.modified,
+    workspaceMeta?.modified,
+    apiSpec.modified,
+    workspaceMeta?.cachedGitLastCommitTime,
```
dimitropoulos, line 131: so here we use `workspaceMeta?.cachedGitLastCommitTime` instead of `lastCommitTime` from above (which is the same thing).  as a reader, I find this confusing and arbitrary.  I would rather just use the actual values and then assign them new names if we have to, but directly in the object that this function returns.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+    lastCommitAuthor,
+    spec,
+    specFormat,
+    specFormatVersion,
+    hasUnsavedChanges,
+  } = props;
+
+  let branch = lastActiveBranch;
+
+  // @TODO Figure out how we should handle a missing timestamp
```
dimitropoulos: @develohpanda any suggestions on this?

packages/insomnia-app/app/common/misc.ts
```diff
@@ -445,6 +445,10 @@ export function snapNumberToLimits(value: number, min?: number, max?: number) {
   return value;
 }
 
-export function isNotNullOrUndefined(obj: any | null | undefined) {
-  return obj !== null && obj !== undefined;
-}
+export function isNotNullOrUndefined<ValueType>(
```
develohpanda, line 448: Love to see it

packages/insomnia-app/app/models/space.ts
develohpanda: There is a question here I'd like to pose to @wdawson first.

Should the workspace sort order be on a _per space_ basis, or for _all spaces on the client_? IMO, the latter.

In some future state we do expect to be able to export and share a space, and in that case the sorting will also be shared. This isn't desirable though (right?). Sorting is always a user specific, it's a user preference.

So IMO this doesn't belong on the `Space` model at all, it belongs in the Settings object, potentially as a `dashboardSortOrder` setting.

packages/insomnia-app/app/ui/redux/modules/space.ts
```diff
@@ -44,3 +44,11 @@ export const removeSpace = (space: Space) => dispatch => {
...
   });
 };
+
+export const updateSpaceItemsOrder =
+  async (space: Space, order: SpaceItemsSortOrder) => {
+    models.space
+      .update(space, {
+        order,
+      });
+  };
```
develohpanda: Usually a helper like this would just sit on the space itself, it doesn't need to be in `redux/modules/...` because it doesn't use state or dispatch in any way

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+        return cardA.workspace.name.localeCompare(cardB.workspace.name);
+      case 'NameDescending':
+        return cardB.workspace.name.localeCompare(cardA.workspace.name);
+      case 'DateCreatedAscending':
+        return cardA.workspace.created - cardB.workspace.created;
+      case 'DateCreatedDescending':
+        return cardB.workspace.created - cardA.workspace.created;
+      default:
+        return cardB.lastModifiedTimestamp - cardA.lastModifiedTimestamp;
+    }
```
develohpanda, line 78: Is it possible to use the sort methods available in https://github.com/Kong/insomnia/blob/0178533e5284cbcaff2c39a4ba2d560b16422e4c/packages/insomnia-app/app/common/sorting.ts? These are all unit tested properly, it would be nice to re-use. It does seem possible without any major refactoring

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+  };
+}
+
+interface OrderFilterDropdownProps {
+  onSelect: (value: SpaceItemsSortOrder) => void;
+}
+
+function OrderFilterDropdown(props: OrderFilterDropdownProps) {
```
develohpanda: Just to follow the pattern in this code-base for function components (inline arrow-funcs typed with FC<>), I suggest `const OrderFilterDropdown: FC<OrderFilterDropdownProps> = ({ onSelect }) => <Dropdown />`

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: I feel strongly that mapping a data structure (like https://github.com/Kong/insomnia/blob/0178533e5284cbcaff2c39a4ba2d560b16422e4c/packages/insomnia-app/app/ui/components/sidebar/sidebar-sort-dropdown.tsx) is a more stable approach for something like this.

As one example of a benefit, it couples the sort type/label _and_ the handler, which means you can't have one without the other. In the current implementation, I will be allowed to remove the "NameAscending" switch case in the sort handler, for example, but the dropdown item will still exist. The only time I will notice this mismatch is in runtime.

Why did you choose to do it this way (without a map) and what are your thoughts on it; what am I missing?

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+  const workspaceMeta = workspaceMetas?.find((wm) => wm.parentId === workspace._id);
+
+  const lastActiveBranch = workspaceMeta?.cachedGitRepositoryBranch;
+
+  const lastCommitAuthor = workspaceMeta?.cachedGitLastAuthor;
+
+  const lastCommitTime = workspaceMeta?.cachedGitLastCommitTime;
+
+  // WorkspaceMeta is a good proxy for last modified time
+  const workspaceModified = workspaceMeta?.modified;
```
develohpanda: The previous implmenetation used `workspace.modified` if `workspaceMeta` didn't exist

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+    workspace,
+  };
+}
+
+interface WorkspaceCardProps {
+  apiSpec: ApiSpec;
+  workspace: Workspace;
+  filter: string;
+  activeSpace: Space;
+  lastActiveBranch?: string | null;
```
develohpanda: Now that we're cleaning this up I think we should avoid the optional and nullable properties; rather standardize to optional

`lastActiveBranch?: string`

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: I agree with this, the sort dropdown and workspace card can be standalone components in separate files

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: Can we investigate how we can end up in a state where `lastModifiedTimestamp` is falsey? It doesn't feel like this is a valid situation so there might be a typing issue further up.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -65,6 +62,261 @@ interface State {
...
+  // @TODO Figure out how we should handle a missing timestamp
+  let log = <TimeFromNow timestamp={lastModifiedTimestamp || new Date()} />;
+
+  if (hasUnsavedChanges) {
+    // Show locally unsaved changes for spec
+    // NOTE: this doesn't work for non-spec workspaces
+    branch = lastActiveBranch + '*';
+    log = (
+      <Fragment>
+        <TimeFromNow className="text-danger" timestamp={modifiedLocally || new Date()} /> (unsaved)
```
develohpanda: I don't think `new Date()` is a reasonable value to put here as a fallback :thinking:

develohpanda:
A couple of initial notes, haven't had a chance to fully review yet

develohpanda:
> Does adding a new option to the Space model require a migration? For now this uses the LastModified as a default sorting option in case a space doesn't have one.

Nope; migrations in this app aren't akin to regular database migrations introducing a new column for instance. Migrations can be used to change entities as required, but because the Insomnia database is essentially a document store, we just need to add a property, give it a default, and that property will be added to all existing documents/entities on launch

packages/insomnia-app/app/models/space.ts
wdawson: Yeah, I would think the sort order should be ephemeral (not stored with the data). How you want the APIs sorted on your screen feels like a separate concern to what a Space is. I could see the simplest thing to be to have a client-side cross-Space preference for how you want to sort the dashboards (i.e. alphabetical, recently updated, created, etc.)

gatzjames: @dimitropoulos agree on the naming not being clear enough. Some other thoughts I had: `sortOrder`, `sortWorkspacesBy`

@develohpanda I initially had the same thought on using a global setting.
My thoughts on having a different sort order per space is that updating the current view of a Space shouldn't affect other views outside of it.
As a user I would switch to another space, change the sort order, go back to the previous space and now everything would look different.
It would also make things cleaner if we were getting the space items list by calling an api where with a global orderBy filter we would have to re-fetch the data for a view that the user might have just visited.
That said, I also wouldn't want to bloat the Space model if this isn't true from a UX perspective.

wdawson: > As a user I would switch to another space, change the sort order, go back to the previous space and now everything would look different.

I agree that it would be confusing depending on how the UX behaved. However, I'd expect switching spaces to be relatively rare. And if you switched space A from alphabetical to most recently created (for example), I don' think that it's completely unexpected for space B to also be sorted that way. What makes it slightly more challenging from a UX perspective is that the attributes you can sort by are not all show on the cards, so it's a bit ambiguous as to how they are sorted sometimes.

dimitropoulos: > the attributes you can sort by are not all show on the cards

This completely changed my mind.  After reading and considering this it seems clear that it needs to be global.  It's just too fiddly to grok all the different states.  In fact, maybe we should duplicate this setting in the main settings so people can configure it there as well?

wdawson: I always hesitate to duplicate settings. I would rather it stay out of a settings menu that is away from the context of the thing they are changing. But @erickrico may have other opinions that I would trust more.

develohpanda: It doesn't need any kind of duplication in settings/Preferences modal. The dropdown on the dashboard can be connected to and update the single global settings field.

Speaking of, since we're persisting this sort order, we should also show a tick for which sort order is selected (similar to how we show a tick to indicate which space is selected)

packages/insomnia-app/app/common/constants.ts
```diff
@@ -299,13 +299,15 @@ export type SortOrder =
...
-export const SORT_TYPE_ASC: SortOrder = 'type-asc';
+export const SORT_NAME_ASC = 'name-asc';
+export const SORT_NAME_DESC = 'name-desc';
+export const SORT_CREATED_ASC = 'created-asc';
+export const SORT_CREATED_DESC = 'created-desc';
+export const SORT_MODIFIED_ASC = 'modified-asc';
+export const SORT_MODIFIED_DESC = 'modified-desc';
+export const SORT_HTTP_METHOD = 'http-method';
+export const SORT_TYPE_DESC = 'type-desc';
+export const SORT_TYPE_ASC = 'type-asc';
```
develohpanda, line 310: Curious why you removed the type from these?

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -4,63 +4,156 @@ import { autoBindMethodsForReact } from 'class-autobind-decorator';
...
+    ReturnType<typeof mapStateToProps> {
   wrapperProps: WrapperProps;
 }
 
 interface State {
   filter: string;
 }
 
+function orderSpaceCards(orderBy: SpaceSortOrder) {
+  return (cardA: Pick<WorkspaceCardProps, 'workspace' | 'lastModifiedTimestamp'>, cardB: Pick<WorkspaceCardProps, 'workspace' | 'lastModifiedTimestamp'>) => {
```
develohpanda, line 64: What is the rationale behind making these props `Pick<WorkspaceCardProps, ...>` instead of the entire `WorkspaceCardProps` interface?

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: Unless I am missing something, it looks like this comment still needs to be addressed.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -4,63 +4,156 @@ import { autoBindMethodsForReact } from 'class-autobind-decorator';
...
 import { importClipBoard, importFile, importUri } from '../redux/modules/import';
 import { activateWorkspace, createWorkspace } from '../redux/modules/workspace';
-import Highlight from './base/highlight';
+import { selectSpaceSortOrder } from '../redux/selectors';
 import SettingsButton from './buttons/settings-button';
 import AccountDropdown from './dropdowns/account-dropdown';
 import { RemoteWorkspacesDropdown } from './dropdowns/remote-workspaces-dropdown';
 import { SpaceDropdown } from './dropdowns/space-dropdown';
-import { WorkspaceCardDropdown } from './dropdowns/workspace-card-dropdown';
+import OrderFilterDropdown from './dropdowns/space-sort-dropdown';
```
develohpanda: Aliasing default exports like this has been used (historically) in a few places and it's _so confusing_ when debugging through it. We are moving away from default exports and unnecessary aliasing anywhere, but especially from aliasing component names and their imports in this way. Let's change `SpaceSortDropdown` to be a named export.

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -632,20 +656,40 @@ export const exportRequestsToFile = (requestIds: string[]) => async (dispatch: D
   });
 };
 
-export function initActiveSpace() {
-  let spaceId: string | null = null;
+export const initActiveSpace = () => (dispatch) => {
```
develohpanda: This initialization helper, the same as `initActiveWorkspace` and `initActiveActivity` should only serve one purpose, to initialize the active space from local storage. I don't think initializing any other loosely related properties should happen here.

We should instead create a separate initializer to for the sort order named `initSpaceSortOrder`, which should only serve one purpose and should be nearly identical to the existing ones in terms of logic and functionality.

Once they are individual functions serving one purpose, then this refactor to a thunk won't be necessary either.

packages/insomnia-app/app/ui/redux/modules/__tests__/global.test.ts
```diff
@@ -250,25 +251,58 @@ describe('global', () => {
...
         type: SET_ACTIVE_SPACE,
         spaceId,
-      };
-      expect(initActiveSpace()).toStrictEqual(expectedEvent);
+      },
+      {
+        'payload': {
+          sortOrder,
+        },
+        'type': 'global/space-sort-order',
```
develohpanda: Should be using the exported constant rather than hard-coding a string

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -392,24 +406,33 @@ class WrapperHome extends PureComponent<Props, State> {
...
+    dispatch
+  );
 
   return ({
     handleCreateWorkspace: bound.createWorkspace,
     handleGitCloneWorkspace: bound.cloneGitRepository,
     handleImportFile: bound.importFile,
     handleImportUri: bound.importUri,
     handleImportClipboard: bound.importClipBoard,
+    setSpaceSortOrder: bound.setSpaceSortOrder,
```
develohpanda: Please use `handleSetSpaceSortOrder` even though the `handleSet...` wording seems redundant. Consistency is king; if we want to drop the `handle` terminology we certainly can but mix-matched naming leads to an overhead

packages/insomnia-app/app/ui/components/wrapper-home.tsx
```diff
@@ -328,23 +308,48 @@ class WrapperHome extends PureComponent<Props, State> {
             <span className="fa fa-search filter-icon" />
           </KeydownBinder>
         </div>
+        <OrderFilterDropdown
+          onSelect={(value) => activeSpace && setSpaceSortOrder(value)}
```
develohpanda: Why is the `activeSpace` check necessary here?

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -84,6 +84,11 @@ export const selectActiveSpace = createSelector(
   },
 );
 
+export const selectSpaceSortOrder = createSelector(
+  (state: RootState) => state.global.spaceSortOrder,
+  sortOrder => sortOrder
+);
```
develohpanda: This should be the following to match the existing global selectors (see the bottom of this file)

```suggestion
export const selectSpaceSortOrder = createSelector(
  selectGlobal,
  global => global.spaceSortOrder,
);
```

develohpanda (changes requested):
Nice one; I tested it locally and it seems to be working well! The sort functionality was built in Flow and it's nice to see it receive some TS love. It's not completely there yet, but it is close; with a few minor changes we can remove a lot of repetition which I hope to address after this PR is merged in. 🙌🏽 

I have a couple of observations and notes that ideally need to be addressed. Most importantly, we should avoid deviating from existing patterns, and if we want to change the existing pattern we should apply that change to all uses of that pattern. I understand that we are making incremental improvements to this codebase but in that process we shouldn't end up in a state where we have multiple naming patterns/implementation patterns for the same underlying concept. In particular this feedback applies to how the `spaceSortOrder` is being initialized from local storage into redux, but it applies generally as well.

Could you also please the previous review comments as resolved and/or link to the commit where it was address, but at the very least mark as resolved; I think you'll find that some of the comments have been missed (eg. one of the sort options isn't available in the dropdown, and I don't know if that's intentional or it was missed).

Lastly, as an aside, while you are self-reviewing this PR, could you please make sure to exclude any extra changes that get included as a result of formatting rules in your code-editor (I presume)? It reduces the noise in a code-review and thus helps to review faster. I'm on board (and encourage) if we want to change any lint & formatting rules, but ideally we should avoid allowing smaller formatting user preferences from sneaking in with PRs.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: Just noticed that the Linear ticket specifies the criteria as below

* Alphabetical (asc/desc)
* Creation Date (asc/desc)
* Last Modified (Default)

I have created a follow-up ticket to address this to not block this PR: https://linear.app/insomnia/issue/INS-841/allow-modified-ascendingdescending-sort

packages/insomnia-app/app/models/space.ts
develohpanda: > Speaking of, since we're persisting this sort order, we should also show a tick for which sort order is selected (similar to how we show a tick to indicate which space is selected)

Added a linear ticket to address this: INS-842

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: Nice one. Removed the check

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: Updated! Agree on consistency plus naming is hard 💯

packages/insomnia-app/app/ui/redux/modules/global.tsx
gatzjames: Good one. My reasoning was to put together all things related to initialising a space together but that could be another fn that calls these but that's out of this context.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: The WorspaceCard also accepts the `onSelect, activeSpace, filter` props which are required and not available on the orderSpaceCards fn. We could also make another type e.g.

```
interface SortableWorkspaceCard {
  workspace: Workspace;
  lastModifiedTimestamp: number;
}
``` 

and change it like that:

```suggestion
  return (cardA: SortableWorkspaceCard, cardB: SortableWorkspaceCard) => {
```

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: Agree. Added a check to update the log if the modifiedLocally has a value.
This probably won't be needed once we remove null from the types.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: I tried removing null from the types but the changes required touch a lot of files and logic.
I'd suggest making another PR for removing null from our types.

packages/insomnia-app/app/ui/redux/modules/space.ts
gatzjames: Removed and using redux store for persistence.

packages/insomnia-app/app/models/space.ts
gatzjames: Great feedback 🙏. Updated the sort order is persisted in the localStorage and the global state in redux.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: Nice one I also think currying makes it cleaner!

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: I tried to follow the write-what-you-know approach tbh. Wasn't sure how things were going to evolve e.g. what would happen if we wanted to add a separator between different items etc. I prefer to have the view logic e.g. how something is going to be shown and interacted with, separate from the data layer as they usually deviate. I agree that this is a more stable approach though plus it makes it more consistent with how things are done in the codebase so I updated it accordingly!

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: Added the unreachableCase handler for the default switch case. 
One thing I wanted to share is that TS will check if all possible values of the switch value are handled and if not will require a default case which I think makes it redundant.

@develohpanda yep showing the First Modified(modified-asc) option might not be what users want plus I think we should add it to the other sorting dropdowns if that's the case for consistency.

packages/insomnia-app/app/common/constants.ts
gatzjames: They are now being used as values for the SpaceSortOrder type too, plus declaring a const string as a union type doesn't provide any additional value (except autocomplete choice) since they are treated as literal types by TS.

packages/insomnia-app/app/ui/components/wrapper-home.tsx
gatzjames: Fixed and now uses the curry version of the fn

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: There was a hotfix we had to do a while back because a default case wasn't handled: https://github.com/Kong/insomnia/pull/3778 & https://github.com/Kong/insomnia/pull/3779. Maybe TS requires strict mode enabled before it errors for these, but until then we'll want to keep using the lint rule and `unreachableCase`

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: Thanks for sharing your approach!

> Wasn't sure how things were going to evolve e.g. what would happen if we wanted to add a separator between different items etc.

Would you agree that we also shouldn't be preparing for an unknown future state? If it evolves, so does the implementation at that time.

> I prefer to have the view logic e.g. how something is going to be shown and interacted with, separate from the data layer as they usually deviate.

This is case-by-case though, isn't it; in this case the view and data layers _shouldn't_ deviate ever. That is to say, every dropdown item should have a handler and we should try and make it as difficult as possible for a handler to _not_ exist. In the initial implementation, how might you avoid a bug caused by a typo in the `value` prop? For example, if you defined `DateCreatedDescending` incorrectly as `CreatedDateDescending`?


Of course it's a balance because we could go to extremes in either direction but if we make it harder for us to make mistakes it's definitely a win IMO

packages/insomnia-app/app/common/constants.ts
develohpanda: Gotcha; yep the autocomplete choice was the only reason the type existed in the first place. Enums aren't useful everywhere, and this is also a remnant of Flow (which doesn't support enums) but I think this is a great candidate for TS enums in a future refactor

packages/insomnia-app/app/ui/components/wrapper-home.tsx
develohpanda: Ah I understand. That's fine for the scope of this PR I think though this card logic could use another sweep further down the track. It just feels more complex than it realistically needs to be (not as a result of this PR, just the original implementation to begin with)

packages/insomnia-app/app/ui/redux/modules/global.tsx
develohpanda: Gotcha; there is a common fn that calls all the initializers already so I see the `initSpaceSortOrder` just adding into that. Is that what you meant?

https://github.com/Kong/insomnia/blob/37815afd2e8774fe1da9211367f3cc7cfaba2c60/packages/insomnia-app/app/ui/redux/modules/global.tsx#L729-L729

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -647,6 +671,23 @@ export function initActiveSpace() {
   return setActiveSpace(spaceId || BASE_SPACE_ID);
 }
 
+export function initSpaceSortOrder() {
```
develohpanda, line 674: Thanks for splitting this out ❤️

develohpanda (approved):
Wooohooo!

wdawson (approved):
Going off Opender's gif, looks great! ⛵

PR #3856, opened 2021-07-28
Use activateWorkspace action to activate a workspace

develohpanda:
Needs a bunch of QA to make sure the correct (or cached) activity activates:

- [x] from quick switcher (cmd + p)
- [x] from homepage
- [x] on creating a document/collection
- [x] on git clone (of a blank repo)

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/proto-files-modal.tsx
+++ b/packages/insomnia-app/app/ui/components/modals/proto-files-modal.tsx
@@ -7,22 +7,22 @@ import { AUTOBIND_CFG } from '../../../common/constants';
 import { GrpcRequestEventEnum } from '../../../common/grpc-events';
 import type { ProtoDirectory } from '../../../models/proto-directory';
 import type { ProtoFile } from '../../../models/proto-file';
-import type { Workspace } from '../../../models/workspace';
 import * as protoManager from '../../../network/grpc/proto-manager';
 import type { GrpcDispatch } from '../../context/grpc';
 import { grpcActions, sendGrpcIpcMultiple } from '../../context/grpc';
-import type { ExpandedProtoDirectory } from '../../redux/proto-selectors';
+import { RootState } from '../../redux/modules';
 import { selectExpandedActiveProtoDirectories } from '../../redux/proto-selectors';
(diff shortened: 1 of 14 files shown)


packages/insomnia-app/app/ui/components/modals/proto-files-modal.tsx
```diff
@@ -122,7 +128,14 @@ class ProtoFilesModal extends PureComponent<Props, State> {
...
   _handleAddDirectory() {
-    return protoManager.addDirectory(this.props.workspace._id);
+    const { workspace } = this.props;
+
+    if (!workspace) {
+      return;
+    }
+    
+    return protoManager.addDirectory(workspace._id);
+
```
dimitropoulos:
```suggestion
```

packages/insomnia-components/src/button/async-button.tsx
```diff
@@ -14,7 +14,7 @@ function isPromise(obj: unknown) {
 }
 
 export interface AsyncButtonProps<T> extends ButtonProps {
-  onClick: (e: React.MouseEvent<HTMLButtonElement>) => Promise<T>;
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => Promise<T> | undefined;
```
dimitropoulos, line 17:
```suggestion
  onClick?: (e: React.MouseEvent<HTMLButtonElement>) => Promise<T>;
```

Should this be this?  It doesn't make a difference now, but it will when we get to TS 4.4.

develohpanda: No no these are slightly different (and very hard to see). This is defining the `onClick` handler as required, but returning `Promise<T> | undefined`

dimitropoulos (approved):
~one minor non-blocking type thing but otherwise~ good to go

I tested all 4 of the QA checklist

packages/insomnia-components/src/button/async-button.tsx
develohpanda: (At least that is my intent, but if the syntax is incorrect I'll fix)

dimitropoulos: ahhh I was reading it too fast.  thanks.  cool.

PR #3867, opened 2021-07-29
use optional chaining

dimitropoulos:
I'm tired of fixing these all the time manually when I see them.  They're so formulaic that it feels fine to just do them all at once in a scope-constrained PR like this.

Normally I rely on a lint rule that autofixes this, but that rule requires the types to actually be pretty dang correct (or else you get runtime errors by not having the optional chain where you actually, it turns out in the end, really need it).

I just `grep`ed around for the most obvious use-cases and fixed the ones I found.

### DIFF
--- a/packages/insomnia-app/app/ui/components/page-layout.tsx
+++ b/packages/insomnia-app/app/ui/components/page-layout.tsx
@@ -58,7 +58,7 @@ class PageLayout extends PureComponent<Props> {
       workspaces,
     } = wrapperProps;
     const realSidebarWidth = sidebarHidden ? 0 : sidebarWidth;
-    const paneTwo = renderPaneTwo && renderPaneTwo();
+    const paneTwo = renderPaneTwo?.();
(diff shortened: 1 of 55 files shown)


packages/insomnia-app/app/ui/components/page-layout.tsx
```diff
@@ -58,7 +58,7 @@ class PageLayout extends PureComponent<Props> {
       workspaces,
     } = wrapperProps;
     const realSidebarWidth = sidebarHidden ? 0 : sidebarWidth;
-    const paneTwo = renderPaneTwo && renderPaneTwo();
+    const paneTwo = renderPaneTwo?.();
```
develohpanda, line 61: As an aside, this component (and it's props) needs some love/a slight refactor. Remind me to show you why sometime, @gatzjames noticed it too I think

develohpanda (approved):

PR #3873, opened 2021-08-03
Pull a project into the correct space

develohpanda:
The core change that happens in this PR is that when we fetch remote projects, we also fetch the team that project exists in. Then, when we pull the project locally either from the sync dropdown inside a collection or pull dropdown from the dashboard, we set the `workspace.parentId` to be `project.team.id` instead of the active space.

In addition, if the remote space does not exist on the client, it will be created using a shared `Team -> Space` mapping so that names and ids are consistent.

Not using the `activeSpace` but using the `project.team.id` ensures that the workspace is always pulled into the correct space in the app.

See reproduction steps in INS-813 for how to end up in a state, where a request collection/workspace exists in a space that is _not_ the correct space.

### Import into a different space, open collection, then pull from sync menu

In this demo, you can see the active space in the window title. There is a comment further down to address the blink that occurs in the UI, but I will do that in a separate PR.

In both demos, the collection has automatically moved under the correct space; I forgot to show at the end but the collection no longer exists in the space it was imported into from the file (my personal space).


packages/insomnia-app/app/sync/vcs/normalize-project-team.ts
```diff
@@ -0,0 +1,17 @@
...
+export interface ProjectWithTeam extends Project {
+  team: Team;
+}
+
+export const normalizeProjectTeam = (project: ProjectWithTeams): ProjectWithTeam => ({
+  id: project.id,
+  name: project.name,
+  rootDocumentId: project.rootDocumentId,
+  // A project is guaranteed to exist on exactly one team
+  team: project.teams[0],
```
develohpanda, line 16: @DMarby the approach in this PR largely hinges on this assumption being true, and to the best of my knowledge it _is_ true but I'd like to triple check.

packages/insomnia-app/app/sync/vcs/vcs.ts
```diff
@@ -121,10 +122,14 @@ export class VCS {
...
   }
 
-  async remoteProjects(teamId?: string) {
+  async remoteProjects(teamId: string) {
     return this._queryProjects(teamId);
   }
 
+  async remoteProjectsInAnyTeam() {
+    return this._queryProjects();
+  }
```
develohpanda, line 131: Wanted to make this distinction very clear to avoid confusion

packages/insomnia-app/app/sync/vcs/vcs.ts
```diff
@@ -1131,7 +1138,8 @@ export class VCS {
       },
       'projects',
     );
-    return projects;
+
+    return (projects as ProjectWithTeams[]).map(normalizeProjectTeam);
```
develohpanda, line 1142: Ideally convert `_runGraphQL` into a generic function instead of casting, but this works for now.

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
```diff
@@ -292,13 +311,22 @@ class SyncDropdown extends PureComponent<Props, State> {
...
     this.setState({
       loadingProjectPull: true,
     });
 
-    await pullProject({ vcs, project: p, space });
+    const pulledIntoSpace = await pullProject({ vcs, project: p, remoteSpaces });
+    if (pulledIntoSpace._id !== space._id) {
+      // If pulled into a different space, reactivate the workspace
+      await handleActivateWorkspace({ workspaceId: workspace._id });
+      // TODO: do we need to show an alert to the user? It's pretty seamless
```
develohpanda: What do we think about this; is a console log sufficient? IMO yes. I'll record a video when I get a chance

packages/insomnia-app/app/sync/vcs/pull-project.ts
```diff
@@ -1,34 +1,36 @@
...
   const remoteBranches = await vcs.getRemoteBranches();
   const defaultBranchMissing = !remoteBranches.includes(DEFAULT_BRANCH_NAME);
 
-  const workspaceParentId = space._id;
+  // Find or create the remote space locally
+  let space = remoteSpaces.find(({ remoteId }) => remoteId === project.team.id);
+  if (!space) {
+    space = await initializeSpaceFromTeam(project.team);
+    await database.upsert(space);
+  }
```
develohpanda, line 28: This change to `getOrCreate` a space matching the `project.team`, instead of using the currently active space, is the key change to this logic for pulling a project. It ensures a remote project always sits under the expected remote space.

packages/insomnia-app/app/ui/redux/modules/workspace.ts
```diff
@@ -96,5 +106,7 @@ export const activateWorkspace = (workspace: Workspace) => {
       const nextActivity = cachedActivity as GlobalActivity ||  (isDesign(workspace) ? ACTIVITY_SPEC : ACTIVITY_DEBUG);
       dispatch(setActiveActivity(nextActivity));
     }
+
+    // TODO: dispatch one action to activate the space, workspace and activity in one go to avoid jumps in the UI
```
develohpanda, line 110: Not doing a single state change results in a flash of the dashboard before switching to the correct workspace. It's not something to change in this PR, but it's an improvement to be made.

packages/insomnia-app/app/ui/components/workspace-page-header.tsx
```diff
@@ -54,15 +53,18 @@ const WorkspacePageHeader: FunctionComponent<Props> = ({
     />
   );
 
+  const crumbs = [
+    { id: 'home', node: activeSpace.name, onClick: homeCallback },
```
develohpanda: Need to revert this (and use `node: strings.home.singular`) but it makes this PR easier to QA.

packages/insomnia-app/app/sync/vcs/normalize-project-team.ts
DMarby: This is a correct assumption, yep.

packages/insomnia-app/app/ui/components/workspace-page-header.tsx
develohpanda: Updated with [`93f465d` (#3873)](https://github.com/Kong/insomnia/pull/3873/commits/93f465de5f5621e0109ea889415d93d6c823d507)

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
develohpanda: See demo in description

wdawson:
Looks a lot better than the existing behavior. I'm curious why we don't import the collection into the right space in the first place. I guess we don't include the space id in the export or something?

develohpanda:
> Looks a lot better than the existing behavior. I'm curious why we don't import the collection into the right space in the first place. I guess we don't include the space id in the export or something?

We don't import into the right space in the first place because a space itself isn't exported; we only import into the active space: https://github.com/Kong/insomnia/pull/3795.

In theory we could also try and import directly into the correct remote space as long as some conditions are met. During import, if a user is signed in, we fetch the list of remote projects, match the workspace being imported, to then be placed into the right space.

The behavior as it stands in this PR currently will still be needed, in case a user imports while they are logged out. On log in and pull, the workspace should be moved under the expected space.

DMarby (approved):

PR #3876, opened 2021-08-04
Migrate collections into their remote space

develohpanda:
### Why

Let's talk about this migration!

In a world before Spaces existed in the application, all collections would show on the single dashboard. But now, we have spaces, which show only the collections contained within the selected space.

Previously, a `Workspace` did not need to have a `parentId`, because a `Workspace` was the top-most model in the hierarchy. But now, a `Space` is the top-most model, thus all `Workspace` models must have a `parentId` specified. This is achieved by a simple migration which already exists, which moves workspaces under the base space.

https://github.com/Kong/insomnia/blob/7abde2a01700f587179941b3231fb1078fcb1e41/packages/insomnia-app/app/models/workspace.ts#L161-L167

The exception lies with workspaces/collections which previously had sync enabled with Insomnia sync. These collections (projects) existed under a Space (team) on the server but the Space was, until now, mostly ignored by the application.

### What

**Because we now also mirror the remote Spaces from the API into the Insomnia application, we should move existing collections that have sync enabled into their respective remote space in the application.** Conceptually we need to  query the API to know which space (team) each collection should exist under, and assign that space as the parent of the collection.

### How

The migration logic will only query the API if the following conditions are met for a given collection:
1. the user is logged in
2. the collection has a local project (in the `<app-data>/version-control` directory, indicating that the collection has sync enabled)
3. the collection does not exist in a remote space

This migration check will run on launch, and then on a 30 minute interval, though every effort has been made to avoid querying the API unless absolutely necessary. This is enforced by unit tests.

### Where

There are three places we could trigger the initial migration to run (and set up an interval) and each place has it's pros and cons.

#### in the main process before the app launches [ref](https://github.com/Kong/insomnia/blob/6f445f86680462614f8e30c31d91771e754a5268/packages/insomnia-app/app/main.development.ts#L53-L64)  
➕ ensures collections are moved to the expected space before app renders
➖ not possible because the VCS network call requires a session id which is stored in `localStorage` and with the current VCS implementation not easy to access from the main process

#### in the render process before the app renders [ref](https://github.com/Kong/insomnia/blob/cecd95f0d7553a0f3126abceb07427be505df7d2/packages/insomnia-app/app/ui/index.tsx#L28-L51)
➕ ensures collections are moved to the expected space before app renders
➖ slow network call could result in a perceived slow launch / failed launch of the app when comparing to existing launch times, potentially need to add full-page loading UI?
➖ need to gracefully exit/skip if no network access

#### in the render process after the app renders (current implementation)
➖ collections will move after the app renders. Depending on the page being viewed, this could look funny. (eg. if viewing the dashboard of the base space, and a collection is moved to a remote space, it will seem to just disappear from the dashboard - see demo below). Potentially show an alert when one or more collections are moved?
➕ slow network won't slow down the app launch
➕ should not need collections to be moved very frequently after the first launch of 2021.5.

### Questions

- [ ] I have chosen the third option as per the current implementation, but **keen for any other feedback here**.
- [ ] Do we also need to check for collections which need to be moved, on [login state change](https://github.com/Kong/insomnia/blob/0178533e5284cbcaff2c39a4ba2d560b16422e4c/packages/insomnia-app/app/ui/redux/modules/index.ts#L24-L26)?

### Demo

In the following demo, I open the Insomnia Testing remote space and pull down a collection named INS-813. I then force it into the base space using the following helper function (this can only be run during development). This puts the app in a state where we have sync enabled on a collection sitting under a non-remote space (the same state a user will be in after upgrading to 2021.5).

```js
async function moveWorkspaceToSpace(workspaceId, spaceId) {
  const workspace = await models.workspace.getById(workspaceId);
  await models.workspace.update(workspace, { parentId: spaceId });
}
```

Once we are in this state, I trigger a refresh of the app (to simulate the user launching the app for the first time). After the app relaunches, we see the collection disappears from the dashboard and a message printed in the console indicating that the collection has been moved into it's correct space. If multiple collections need to be moved into different spaces, then multiple console logs will be printed, one for each collection.

Closes INS-760


packages/insomnia-app/app/sync/vcs/migrate-collections.ts
```diff
@@ -0,0 +1,66 @@
...
+export const migrateCollectionsIntoRemoteSpace = async (vcsDirectory: string) => {
+  const driver = new FileSystemDriver({ directory: vcsDirectory });
+  const vcs = new VCS(driver);
+  
+  const collections = (await models.workspace.all()).filter(isCollection);
+  const remoteSpaces = (await models.space.all()).filter(isRemoteSpace);
+    
+  // Are there any collections that have sync setup but are not in a remote space?
+  const isNotInRemoteSpace = (collection: Workspace) => !Boolean(remoteSpaces.find(space => space._id === collection.parentId));
+  const hasLocalProject = (collection: Workspace) => vcs.hasLocalProjectForId(collection._id);
```
DMarby, line 27: I think you'll want/need `vcs._getProjectByRootDocument` here instead, as currently it tries to find the Workspace ID in the vcs store, which it will never be able to do, as things are stored by their project IDs.

packages/insomnia-app/app/sync/vcs/vcs.ts
```diff
@@ -74,6 +74,10 @@ export class VCS {
     return this._project !== null;
   }
 
+  hasLocalProjectForId(id: string) {
+    return Boolean(this._store.hasItem(paths.project(id)));
```
DMarby: Might be worth noting that this will never return true if you pass it a collection id as I think is currently done in `migrateCollectionsIntoRemoteSpace`, as projects are stored using their project ID, not the collection id.
I think we'll need to use the approach that's used in `_getProjectByRootDocument` or `removeProjectsForRoot` to find out if there's any project for a given collection ID (`rootDocumentId`) instead.

develohpanda: That's a great catch yeah; since this isn't linked up to anything I haven't actually tried it yet. I wrote some tests with VCS mocks, but I think only the network requests should be mocked (`projectsInAnyTeam`); any local things should use the `MemoryDriver` instance, so I'll try to fix this with some unit tests.

So many ids, so I appreciate the second set of eyes across this 👀

packages/insomnia-app/app/sync/store/drivers/memory-driver.ts
```diff
@@ -42,9 +42,9 @@ export default class MemoryDriver implements BaseDriver {
       }
 
       const levels = key.split('/').length;
-      const isOnSameLevel = levels === baseLevels;
+      const isInBaseLevel = levels === baseLevels + 1;
```
develohpanda, line 45: This part of the memory driver implementation was very broken and my goodness it took an effort to figure out what was up 🙄

packages/insomnia-app/app/sync/vcs/vcs.ts
develohpanda: Updated in [`51be8da` (#3876)](https://github.com/Kong/insomnia/pull/3876/commits/51be8da674e60014625b063a5324736717061fcb) and added unit test in [`d6dbb5a` (#3876)](https://github.com/Kong/insomnia/pull/3876/commits/d6dbb5a671181d0f2bf82d32639c121d5c35f98c)

DMarby:
> Do we also need to check for collections which need to be moved, on login state change?

@develohpanda I think that would be a good idea, I think that we could maybe avoid running it on a timer entirely if we do it on startup + on login state change?

packages/insomnia-app/app/ui/hooks/use-sync-migration.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { getDataDirectory } from '../../common/electron-helpers';
+import FileSystemDriver from '../../sync/store/drivers/file-system-driver';
+import { migrateCollectionsIntoRemoteSpace } from '../../sync/vcs/migrate-collections';
+import { VCS } from '../../sync/vcs/vcs';
+import { invokeAsyncSynchronously } from './effect-helpers';
+
+const check = async () => {
+  const directory = path.join(getDataDirectory(), 'version-control');
+  const driver = new FileSystemDriver({ directory });
```
DMarby: Might be worth making a utility function here, so that that this and the other place we instantiate VCS in the app doesn't diverge

DMarby (approved):

wdawson (commented):
As I understand it, this means that existing collections will move to the base space. I think that is a "local only" space that cannot be synced. What happens if some of these workspaces are synced? Is that basically your demo? Do they get moved out and then moved back on the next login??

DMarby:
> As I understand it, this means that existing collections will move to the base space. I think that is a "local only" space that cannot be synced. What happens if some of these workspaces are synced? Is that basically your demo? Do they get moved out and then moved back on the next login??

When upgrading from previous versions of the app, all existing design documents/collections do indeed appear in the base space.
What this PR does is that it ensures that existing synced collections gets moved into their correct remote spaces (assuming the user is logged in)

packages/insomnia-app/app/sync/store/drivers/memory-driver.ts
dimitropoulos: wow. just wow.  I can only imagine.

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1516,6 +1517,8 @@ class App extends PureComponent<AppProps, State> {
     return (
       <KeydownBinder onKeydown={this._handleKeyDown}>
         <GrpcProvider>
+          <AppHooks />
```
dimitropoulos, line 1517: since `App` is traditionally one of the most simple components in a project (because it's often just a top-level simple container) I was about to ask "any reason we can't just make this hooks now instead of a hack like this? (clever, as the hack is!)" and then I saw the line number (`1520`) and then I saw the code....  NOPE.  RUN AWAY.  haha.  Now is most definitely not a time to crack that egg.

develohpanda:
> > Do we also need to check for collections which need to be moved, on login state change?
> 
> @develohpanda I think that would be a good idea, I think that we could maybe avoid running it on a timer entirely if we do it on startup + on login state change?

Yeah, I think you're right. This PR combined with https://github.com/Kong/insomnia/pull/3873 covers most of the (reasonable) scenarios I can think of 🤔 will replace the timer with a login state change hook.

Updated in [`f7813a7` (#3876)](https://github.com/Kong/insomnia/pull/3876/commits/f7813a74a70240f37a27898a4c3ac43df8d3cb13)

packages/insomnia-app/app/ui/hooks/use-sync-migration.ts
develohpanda: Let me know if [`030aefc` (#3876)](https://github.com/Kong/insomnia/pull/3876/commits/030aefc360940e1121c68d5f5fa738e5d160dc31) is along the lines of what you have in mind. If not, I can extract a standalone utility, but want to avoid pulling `getDataDirectory()` which exists in an electron context into `FileSystemDriver` which doesn't currently know about electron

packages/insomnia-app/app/ui/hooks/effect-helpers.ts
```diff
@@ -0,0 +1,6 @@
+export const invokeAsyncSynchronously = (callback: () => Promise<void>) => {
+  async function func() {
+    await callback();
+  }
+  func();
+};
```
dimitropoulos: how is this different from https://github.com/streamich/react-use/blob/master/docs/useAsync.md in practice?

https://github.com/streamich/react-use/blob/master/src/useAsync.ts

And to be clear, I'm asking why we couldn't just do

```ts
export const useSyncMigration = () => {
  // Check once on mount
  useAsync(check, []);
};
```

instead of the current:
```ts
export const useSyncMigration = () => {
  // Check once on mount
  useEffect(() =>  invokeAsyncSynchronously(check), []);
};
```

develohpanda: No pressing reasons, in practice the two implementations serve the same purpose. I did search for an async effect hook in `react-use`, so (at least to me) it wasn't immediately clear that `useAsync` can do what I expect (run an effect on mount) until looking through the source code.

`useAsync` maintains internal state and state changes will cause it's calling components to update. I think using `useAsync` would be more suitable if we want to show a loading state, or explicitly handle errors if a promise fails to resolve, though in this particular case the state changes and updates probably have no material effect (pun intended).

develohpanda: Updated in [`6c095bc` (#3876)](https://github.com/Kong/insomnia/pull/3876/commits/6c095bc401e6e3ec5d09af1db8b2de78b21120b5) and https://github.com/Kong/insomnia/pull/3888

packages/insomnia-app/app/ui/hooks/use-sync-migration.ts
DMarby: LGTM

wdawson (approved):

PR #3885, opened 2021-08-05
Display the settings modal for remote spaces and update the copy for …

gatzjames:
Enables the space settings for remote spaces.

The name input field is disabled and read-only.
There is also a helper tooltip next to the label that points you to the website if you want to rename the remote space.
Deleting the space removes any local collections/documents and changes that were not synced.
The copy of the confirmation dialog is also changed to try to explain that.

### Demo:

ins-790

### Concerns:
- [x] Is the tooltip copy helpful for users?
- [x] Does the link point to the correct place? Currently points to https://app.insomnia.rest/app/teams
- [x] Is the confirmation dialog copy clear enough for users?

Closes INS-790


wdawson:
This looks alright to me. But I wonder if the fact that the remote space will not be deleted remotely should be clearer. @erickrico any thoughts? And @falondarville , not sure if you want to be included on in-product copy stuff like this but tagging you for now.

packages/insomnia-app/app/ui/components/modals/space-settings-modal.tsx
```diff
@@ -48,10 +49,12 @@ class SpaceSettingsModal extends PureComponent<Props> {
 
   render() {
     const { space } = this.props;
-    if (isBaseSpace(space) || isRemoteSpace(space)) {
+    if (isBaseSpace(space)) {
```
develohpanda: I wonder if this should be using the following check. It's serving the same purpose conceptually, not just implementation

```suggestion
    if (!spaceHasSettings(space)) {
```

develohpanda (approved):
> Does the link point to the correct place? Currently points to app.insomnia.rest/app/teams

Yes I think so; @DMarby please confirm?

LGTM, will rely on @wdawson for copy feedback.

DMarby:
> Yes I think so; @DMarby please confirm?

Looks good to me

packages/insomnia-app/app/ui/redux/modules/space.ts
```diff
@@ -27,9 +27,13 @@ export const createSpace = () => dispatch => {
 };
 
 export const removeSpace = (space: Space) => dispatch => {
+  const message = isRemoteSpace(space)
+    ? `Deleting a remote space will delete all local copies and changes of ${strings.document.plural.toLowerCase()} and ${strings.collection.plural.toLowerCase()} within. All changes that are not synced will be lost. This cannot be undone. Are you sure you want to delete ${space.name}?`
```
wdawson: We should probably add a sentence about how the remote will not be deleted. And guidance for deleting the remote, go to this URL (I'm assuming that's possible? @DMarby @develohpanda ) Something like:

> Deleting a remote space will delete all local copies and changes of `${strings.document.plural.toLowerCase()}` and `${strings.collection.plural.toLowerCase()}` within. All changes that are not synced will be lost. The space will continue to exist remotely. To delete the remote space, visit `<a href="https://app.insomnia.rest/app/teams">the insomnia website.</a>`. Deleting this space locally cannot be undone. Are you sure you want to delete ${space.name}?

wdawson (commented):
Seems like we should improve the clarity of what happens to the remote space. Included a suggestion but feel free to change if we thnk there is a better way to put it.

packages/insomnia-app/app/ui/redux/modules/space.ts
develohpanda: ~~I think only an owner can delete the remote space, so while I might see this as a user on my client, going to the insomnia website it's unlikely I'll see an option to delete it. It's probably still fine to show it, but just a note~~ see below

DMarby: You can't delete remote Spaces at all currently, not even on the website, so I don't think we can include instructions for how to do so unfortunately.

wdawson: Is that difficult to add? If we don't want to add that capability right now, we can just remove the one sentence about visiting the website.

DMarby: @wdawson Not in particular, but would require some API and website changes, so I think it makes more sense to do at a later date, separately from this PR.

wdawson: Sounds good. We can remove that sentence from my suggestion for now and add it back when we add delete capabilities.

gatzjames: Added the copy without the deletion/link for the remote space. I think it's much clearer now!

packages/insomnia-app/app/ui/redux/modules/space.ts
```diff
@@ -27,9 +27,13 @@ export const createSpace = () => dispatch => {
 };
 
 export const removeSpace = (space: Space) => dispatch => {
+  const message = isRemoteSpace(space)
+    ? `Deleting a remote space will delete all local copies and changes of ${strings.document.plural.toLowerCase()} and ${strings.collection.plural.toLowerCase()} within. All changes that are not synced will be lost. The space will continue to exist remotely. Deleting this space locally cannot be undone. Are you sure you want to delete ${space.name}?`
```
wdawson: Do we have variables for the "space" word? Since we'll probably change it before release to "project" it would be good to use that.

gatzjames: Good one! Added the constant instead.

wdawson (approved):
⛵LGTM, thanks for the changes!

PR #3888, opened 2021-08-09
Use `useAsync`

develohpanda:
What it says on the tin


gatzjames (approved):
+💯 for the eslint rule for deps 😱

packages/insomnia-app/app/ui/hooks/theme.ts
```diff
@@ -19,13 +20,11 @@ export const useThemes = () => {
...
+    const pluginThemes = await getThemes();
+    setThemes(pluginThemes.map(({ theme }) => theme));
   // Reload themes if pluginConfig changes
-  useEffect(() => {
-    const func = async () => {
-      const pluginThemes = await getThemes();
-      setThemes(pluginThemes.map(({ theme }) => theme));
-    };
-    func();
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
dimitropoulos:
```suggestion
  // eslint-disable-next-line react-hooks/exhaustive-deps -- reason why this rule is turned off
```

By the way, https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/require-description.html exists (mystictea is one of the main eslint developers).  We are probably a little early to turn this rule on, but maybe eventually we can without having to do dozens of ` -- TODO` lines.

develohpanda: I'd say we switch it on with a warning right now. _something something perfect not be the enemy of good something something_

PR #3894, opened 2021-08-09
Optionally disable certificate validation for OAuth2 auth windows

### LINKED ISSUE #2778: OAuth 2.0 Authorization Window does not respect "Validate certificates" setting
**Describe the bug**
When fetching a new "authorization code" token with help from the OAuth 2.0 auth helper, if the "Authorization URL" uses a self-signed certificate, the Authorization Window appears, but is blank. This is regardless of the "Validate certificates" setting in the preferences window.

**To Reproduce**
1. Set the "Validate Certificates" setting in the preferences window to "false" (unchecked):
1. In an "OAuth 2.0" mode auth tab, enter valid settings including:
* The "grant type" field with a value of "Authorization Code"
* The "authorization URL" field pointing to an HTTPS URL which does not use a valid HTTPS certificate (such as `localhost` or `127.0.0.1`)
3. Click on 'Fetch Tokens'
1. Note the blank window appearing and no network traffic
1. Note the error in the DevTools console, despite the "Validate certificates" setting from earlier:
```
Uncaught (in promise) Error: ERR_CERT_AUTHORITY_INVALID (-202) loading 'https://127.0.0.1:5001/connect/authorize?response_type=code&client_id=insomnia&redirect_uri=https%3A%2F%2Ffake.url%2Foidc-callback&scope=openid%20profile%20email%20otherscopes&state=820813df-8ca6-4fd3-83c5-45fd3f1a5f7a'
```

**Expected behavior**
Assuming "Validate certificates" is set to false (unchecked), the Authorization Window should correctly send a request to the Authorization URL server and display the HTML response so the user can authenticate.

**Desktop:**
 - OS: macOS 10.15.7 (this might be important as this could be a macOS specific issue)
 - Installation Method: Manual Install (`.dmg` download from website)
 - App Version: Insomnia Core 2020.4.1

### LINKED ISSUE #3893: Disabling "Validate SSL" feature does not extend to opened OAuth2 windows
**Describe the bug**
When authenticating a request via OAuth2, Insomnia opens a new windows to perform the login flow. Currently, when presented with a certificate not recognized by the bundled certificate store, Insomia will simply show a blank white screen (tested on Ubuntu 20, with Chromium as the system default browser) in the new window opened to handle OAuth2 login regardless of whether the `Validate SSL` setting is checked.

**To Reproduce**
1. Enter an `Authorization URL` in the `OAuth2` tab which points to a server that presents a self-signed certificate.
2. Attempt to authenticate a request.
3. Note the blank screen.

**Expected behavior**
If the `Validate SSL` checkbox is unchecked, the page should fully render.

 - OS: Ubuntu 20.04.02 LTS
 - Installation Method: Compiled from source (commit be2c347)
 - App Version v2021.4.1

**Additional context**
Relates to https://github.com/Kong/insomnia/discussions/3617.

I have a fix locally, and will open a PR for discussion shortly.

schrodingersket:
Closes #3893, Closes #2778, Closes INS-912

I fully expect to have some feedback regarding better ways to grab the `validateSSL` setting or maybe a different place where this event handler should be, but I figured a functional fix would be a good place to start!


packages/insomnia-app/app/main.development.ts
```diff
@@ -157,6 +157,17 @@ function _launchApp() {
       requestHeaders: details.requestHeaders,
     });
   });
+  // Set up certificate error handler
+  app.on('certificate-error', async (e, _webContents, _url, _error, _certificate, callback) => {
+    e.preventDefault();
+    try {
+      if (!(await models.settings.getOrCreate()).validateSSL) {
+        return callback(true);
+      }
```
develohpanda: Following the example in [the docs](https://www.electronjs.org/docs/api/app#event-certificate-error), I think there are a few updates worth making here:

* `e.preventDefault()` called immediately before `callback(true)` within the if-block, not outside
* invoking the callback doesn't return anything so we can drop the return
* we also need to invoke `callback(false)` if we _should_ validate

packages/insomnia-app/app/main.development.ts
```diff
@@ -157,6 +157,17 @@ function _launchApp() {
...
+    } catch (err) {
+      console.error('[certificate-error] An error occurred: ', err);
```
develohpanda: Should this also log `_error`?

develohpanda (commented):
Thank you for the issue report and this PR!

Because this is specifically for auth, for security reasons I wonder if we should introduce a standalone setting (like `validateOAuthSSL`) rather than use the existing one. This way users have more control over disabling cert validation for regular requests but ensuring cert validation for authentication.

@wdawson / @DMarby / @johnwchadwick any opinions on using the existing flag or introducing a new one specifically for OAuth? The current flag `validateSSL` links to the blue box. We do have a section dedicated to security in preferences, so a custom flag like `validateOAuthSSL` could sit in there (red box).

image

packages/insomnia-app/app/main.development.ts
schrodingersket: You are absolutely correct. Do you think `_error` (which, since it will be used will become just `error`) should always be logged when a `certificate-error` event is emitted? Or only when we call `callback` with `false`?

packages/insomnia-app/app/main.development.ts
schrodingersket: > Following the example in [the docs](https://www.electronjs.org/docs/api/app#event-certificate-error), I think there are a few updates worth making here:
> 
> * `e.preventDefault()` called immediately before `callback(true)` within the if-block, not outside

That was my first thought as well, and what I tried first; however, because fetching the `settings` models is an `async` call, this unfortunately won't stop the event from propagating upward, which in turn causes the behavior described in #3893. Do you have any suggestions as to how to handle this? I see one of two options to do this right off the top of my head:

1) Fetch the `settings` model synchronously. I'm not sure how to do this, and I feel like forcing synchronous behavior in an event handler is probably less than ideal.
2) Re-emit the original event if `validateSSL` (or whatever feature flag is determined to be best to control this behavior) is `true`. I'm also not certain as to what best practices are for this.

Alternatively, we could simply stop event execution here no matter what, and just handle the event the same way other events in the application are handled.

> * invoking the callback doesn't return anything so we can drop the return

Yup. Had that in there simply to end function execution early. I'll refactor that if you prefer consistent function returns.

> * we also need to invoke `callback(false)` if we _should_ validate

100% :+1:

schrodingersket:
Thanks for getting the conversation started on this! Once the Insomnia team decides which direction to take this PR, I'm happy to help where I can (:

packages/insomnia-app/app/main.development.ts
develohpanda: I think, only when we invoke `callback` with `false` 🤔

packages/insomnia-app/app/main.development.ts
develohpanda: Great point about the async handler!

Something else just came to mind - I think it might be better to handle this event only on the OAuth window that opens, not on the entire app. The API is the same, except you add it onto the child window that is created, there are some other event handlers here already.

https://github.com/Kong/insomnia/blob/70d5797d94ac3a8bd83e218dd3f342d9a889bfa8/packages/insomnia-app/app/network/o-auth-2/misc.ts#L105-L105

We should be able to fetch settings/already have access to settings _before_ setting up the event handler and opening the window, so that the event handler can be synchronous and we can invoke preventDefault() right next to the callback. More importantly we restrict this behavior to only the OAuth window.

We can either send settings down to the [`authorizeUserInWindow`](https://github.com/Kong/insomnia/blob/70d5797d94ac3a8bd83e218dd3f342d9a889bfa8/packages/insomnia-app/app/network/o-auth-2/misc.ts#L56-L56) function as an argument, or fetch from settings (as as `await models.settings.getOrCreate()`) the same way you are. If we can avoid the database call that would be ideal, but see how you get on.

What do you think?

> Yup. Had that in there simply to end function execution early. I'll refactor that if you prefer consistent function returns.

Yes please! I'll add a lint rule to catch this in future (if there is one). Out lint rules are still going through refinement 🤗

DMarby: I definitely think that this should go on the child window rather then globally in the app @develohpanda 👍

DMarby:
> any opinions on using the existing flag or introducing a new one specifically for OAuth? The current flag validateSSL links to the blue box. We do have a section dedicated to security in preferences, so a custom flag like validateOAuthSSL could sit in there (red box).

@develohpanda I think it might make sense for this to be a separate option.
Given that some oauth flows also make requests using our network stack however, if the option to not validate certificates for oauth is enabled, but the one for requests is not, that could lead to the window being able to open, but the request made to complete the flow failing, so if we want to make a separate option for this we might also need to make that option be taken into account for those cases?
https://github.com/Kong/insomnia/blob/366a54f6302589def1224ea48d0959955d06ee7b/packages/insomnia-app/app/network/o-auth-2/grant-client-credentials.ts#L66
https://github.com/Kong/insomnia/blob/366a54f6302589def1224ea48d0959955d06ee7b/packages/insomnia-app/app/network/o-auth-2/grant-password.ts#L70
https://github.com/Kong/insomnia/blob/366a54f6302589def1224ea48d0959955d06ee7b/packages/insomnia-app/app/network/o-auth-2/grant-authorization-code.ts#L242
https://github.com/Kong/insomnia/blob/366a54f6302589def1224ea48d0959955d06ee7b/packages/insomnia-app/app/network/o-auth-2/refresh-token.ts#L64

develohpanda:
> > any opinions on using the existing flag or introducing a new one specifically for OAuth? The current flag validateSSL links to the blue box. We do have a section dedicated to security in preferences, so a custom flag like validateOAuthSSL could sit in there (red box).
> 
> @develohpanda I think it might make sense for this to be a separate option.
> Given that some oauth flows also make requests using our network stack however, if the option to not validate certificates for oauth is enabled, but the one for requests is not, that could lead to the window being able to open, but the request made to complete the flow failing, so if we want to make a separate option for this we might also need to make that option be taken into account for those cases?
> 

That's a great point! It looks like all those OAuth flows use `network.sendWithSettings`, and `sendWithSettings` is _only_ used by OAuth flows

image

`sendWithSettings` calls into `_actuallySend`, which contains the following 

https://github.com/Kong/insomnia/blob/7abde2a01700f587179941b3231fb1078fcb1e41/packages/insomnia-app/app/network/network.ts#L421-L428

So it should be fairly straightfoward to tell the networking stack to use `settings.validateOAuthSSL` if `_actuallySend` is called by `sendWithSettings`. 

Lots of ways to solve this, @schrodingersket I'll let you give it a crack, but this is the general code path! 😄 

To summarize the feedback:
- [ ] introduce a new setting named `validateOAuthSSL` alongside `validateSSL`
- [ ] add that setting to the "Security" section under General preferences
- [ ] move the event handler to individual OAuth 2.0 window instead of full app
- [ ] use settings/get settings before the window opens to have a synchronous handler
- [ ] update the networking layer to respect the `validateSSl` or `validateOAuthSSL` flag depending on the call path

packages/insomnia-app/app/main.development.ts
schrodingersket: Excellent! TIL that you can attach event handlers to just the child windows. That definitely feels like the right way to go.

schrodingersket:
> Lots of ways to solve this, @schrodingersket I'll let you give it a crack, but this is the general code path! 
> 
> To summarize the feedback:
> 
> * [ ]  introduce a new setting named `validateOAuthSSL` alongside `validateSSL`
> * [ ]  add that setting to the "Security" section under General preferences
> * [ ]  move the event handler to individual OAuth 2.0 window instead of full app
> * [ ]  use settings/get settings before the window opens to have a synchronous handler
> * [ ]  update the networking layer to respect the `validateSSl` or `validateOAuthSSL` flag depending on the call path

That all sounds perfectly doable. I'll take a swing at it this week :sunglasses:

Thanks for all the input everyone! Super cool to get to learn from you guys and to have the opportunity give a little back to a project I use every day :nerd_face:

wdawson:
Just as a naming comment, I wonder if it would be better to not include the authorization protocol (i.e. "OAuth") in the name for the flag. In case folks are using different protocols. Maybe just `validateAuthSSL` instead? It would apply to OIDC as well I assume.

develohpanda:
> Maybe just `validateAuthSSL` instead?

Yep! I'm good with that 👍🏽

schrodingersket:
@develohpanda - looks like we have a slight snag in restricting the scope of the new `validateAuthSSL` option to the child window created during the OAuth2 authorization flow.

The TL;DR is that the `certificate-error` event is thrown _both_ in the global `app` object, and the child window - and that a `event.preventDefault()` does not seem prevent a `certificate-error` event from being emitted in `app`. Specifically, calling `child.loadURL` throws an actual `Error` instance (which is the cause of the blank screen appearing to hang - when the error is thrown, `child.close()` never gets called):

https://github.com/Kong/insomnia/blob/a50d1539f26640ef787b0c45a3ecdefbe51205ea/packages/insomnia-app/app/network/o-auth-2/misc.ts#L149-L158

I pushed https://github.com/Kong/insomnia/commit/a50d1539f26640ef787b0c45a3ecdefbe51205ea which provides all the other functionality described above and in which you can reproduce what I'm running into if you'd like to poke at it. Any suggestions as to how to handle this would be greatly appreciated (: Adding the `certificate-error` event handler to the global `app` object certainly works, but I definitely want to try to get that behavior scope-limited to the child window before resorting to that as an "actual" fix.

Anyway, here are some screenshots of the new option in `Settings`, as well as what happens when an error is encountered in the authorization window when the `Fetch Tokens` button is clicked in the `OAuth 2` tab:

2021-08-10_22-19-11_new_security_settings

2021-08-10_23-02-25_fetch_tokens_err

schrodingersket:
Bingo! I found a workaround for the issue described above using the [setCertificateVerifyProc API](https://www.electronjs.org/docs/api/session#sessetcertificateverifyprocproc).

Should be all good to go with this now. I'll work on adding some unit tests and updating existing ones as appropriate tomorrow, but it should be all set for review at this point.

develohpanda:
Oh wow, that's impressive @schrodingersket 🙌🏽 👏🏽  I hadn't had a chance to debug through this yet but I'm stoked you found a way around it!

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -371,7 +371,12 @@ class General extends PureComponent<Props, State> {
 
         <div className="row-fill row-fill--top">
           <div>
-            {this.renderBooleanSetting('Validate certificates', 'validateSSL', '')}
+            {this.renderBooleanSetting(
+              'Validate certificates',
+              'validateSSL',
+              'Indicates whether SSL certificates should be validated for API requests.' +
+                'This does not affect SSL certificate validation during authentication.'
```
develohpanda: Nice

packages/insomnia-app/app/network/network.ts
```diff
@@ -904,7 +904,10 @@ export async function sendWithSettings(
     renderResult.request,
     renderResult.context,
     workspace,
-    settings,
+    {
+      ...settings,
+      validateSSL: settings.validateAuthSSL,
+    },
```
develohpanda: Instead of overriding this I'd be inclined to introduce a new parameter to `_actuallySend` for validating SSL, and the various code-paths can set it. Then we update the implementation of `_actuallySend` (line 422) to work with that parameter instead of the property on settings

packages/insomnia-app/app/network/network.ts
```diff
@@ -419,13 +419,13 @@ export async function _actuallySend(
...
-        setOpt(Curl.option.SSL_VERIFYPEER, 0);
-        addTimelineText('Disable SSL validation');
-      }
+      // if (settings.validateSSL) {
+      //   addTimelineText('Enable SSL validation');
+      // } else {
+      setOpt(Curl.option.SSL_VERIFYHOST, 0);
+      setOpt(Curl.option.SSL_VERIFYPEER, 0);
+      addTimelineText('Disable SSL validation');
+      // }
```
develohpanda: Accidentally commented out (I assume?)

develohpanda (commented):
Just an initial code review 🥳

packages/insomnia-app/app/network/network.ts
schrodingersket: You beat me to it by about 30 seconds. Updated along with the unit test fix (:

packages/insomnia-app/app/network/network.ts
schrodingersket: Sounds good to me. So we'd then ignore `settings.validateSSL` and `settings.validateAuthSSL` in `_actuallySend` altogether and only use the new function param, yeah?

develohpanda: Yep! We could even enforce that with TS by omitting those two values from the Settings type in the parameter

schrodingersket: Love it. So the new function signature would be:

```typescript
export async function _actuallySend(
  renderedRequest: RenderedRequest,
  renderContext: Record<string, any>,
  workspace: Workspace,
  settings: Omit<Settings, 'validateSSL'|'validateAuthSSL'>,
  environment?: Environment | null,
  validateSSL=true,
) {
  ...
}
```

...correct? Just want to confirm that I've got it right on that before pushing another commit on this branch. On a related note, does your team prefer that pull request commits are squashed into a single commit before merging, or would you rather have the commit history reflect the code review process as it happened?

develohpanda: Yeah that looks good.

Having new commits is nice because it helps with reviewing the changes as they happen. We typically squash merge into develop, so excess commits in the pr not a concern, and often beneficial 😊 in terms of updating your branch with develop, whether you merge or rebase that's your preference

schrodingersket: Perfect - function signature is updated :sunglasses:

...and sounds good! I'll definitely keep that in mind. (:

schrodingersket:
I think I've pretty much got this wrapped up with respect to the comments that have been added so far. Added unit tests where it seemed to make sense - let me know if y'all feel that more tests are necessary and if so, where those should be (:

Specifically, I added unit tests for the new `setCertificateVerifyProc` callback, as well as to `sendWithSettings` to verify that each of those functions respect the new `validateAuthSSL` option.

develohpanda:
@schrodingersket thank you! We have a release coming up so the team is briefly occupied there, but after that we'll look to get this and other PRs reviewed.

develohpanda (commented):
Suuuper impressed with the unit tests you've written. 💯 



I made an update in [`6ce9d69` (#3894)](https://github.com/Kong/insomnia/pull/3894/commits/6ce9d696db7bec064981a4acc1e8481a29d56678) to grab the type definition from Electron types. The test was also tricky to process initially but after I understood it more I made a few updates in the hopes of making the test itself easier to read. It should be the same though, and still passes, but worth a glance, to make sure I didn't break anything.

Overall, code looks good to me! I am holding off on approval until I do some high level QA, although I don't have an environment set up to directly reproduce this issue with OAuth. 

Assuming you have an environment set up to demonstrate SSL validation working/failing in the OAuth 2 window, would you please be able to share a gif or video of it behaving as expected with each setting?

schrodingersket:
Your comment just made my day. (:

And thanks for tracking down that typing on `setCertificateVerifyProc`! I knew it existed somewhere, but between being pretty new to Electron development and those being unit tests, I didn't spend _that_ much time looking for it. Definitely cleaner with that around though :D

As requested, here's a screencast:

https://user-images.githubusercontent.com/17204901/131025856-d2e22df8-338e-4c83-82d6-156d0bbd8cbf.mp4


Thanks again for all the guidance on this!

develohpanda:
No worries! Great work on this PR 👏🏽 

This might just be me but seems that the video is corrupt, I am not able to play it 😢  Could you please try re-uploading it?

image

dimitropoulos:
@develohpanda the video works for me now, maybe try again?

develohpanda:
@dimitropoulos Looks like it was a Firefox thing 😭 works in Chrome. I've never hit this before (still marked as corrupted on Firefox), how strange!

develohpanda (approved):
LGTM, thank you for the video! 🙌🏽

The error message does seem a bit verbose but I don't think we should parse it to render something different; rather show the entire reason why something failed (we do the same thing when failing to fetch a GrapqhQL schema, show the raw message)

dimitropoulos (approved):
@schrodingersket I would really love to have some pointers into how you get the test environment going that you used in the video you took (and in general in this PR).  I'm always on the lookout for more easily testing the OAuth2 features of the app.

The code looks good to me and the functionality in the video looks good as well.

schrodingersket:
> @schrodingersket I would really love to have some pointers into how you get the test environment going that you used in the video you took (and in general in this PR). I'm always on the lookout for more easily testing the OAuth2 features of the app.
> 
> The code looks good to me and the functionality in the video looks good as well.

@dimitropoulos sure thing! We use [Keycloak](https://www.keycloak.org/) at work for IAM and OAuth2, and I've been testing against that for this PR. OAuth2, JWT, and OIDC support are pretty well-supported in my experience with it.

If you'd like, I'd be happy throw together a super quick test repo for you based on their [Docker image](https://github.com/keycloak/keycloak-containers/blob/master/server/README.md) with SSL set up for you as a reference repository for testing if you feel it'd be worthwhile to have on hand and don't feel like rummaging through all the docs yourself (:

dimitropoulos:
I would really really love that and the whole team will appreciate it!  Thanks!!

schrodingersket:
@dimitropoulos @develohpanda - here you go:

https://github.com/schrodingersket/keycloak-oauth2-ssl-testbed

Everything should be configured and ready right out of the box for you there. I added a fairly comprehensive README with screenshots and such for getting everything running, but the TL;DR is (assuming you've installed Docker) that you just need to run `run.sh` in a terminal and import the included Insomnia export in that repo into your local instance and you should be all good to go. It's also configured as a `Git Sync` repo, so you can just import it from Git instead of as a File Import if you prefer.

I added the script I used to generate the certificates (via `openssl`) to the repository, so you can modify those certs however you need; that script generates a root certificate authority and signs the server certificates with it, since I expect it might be useful for y'all to have a convenient place to test out custom CA bundles as well (:

develohpanda:
Thanks a tonne @schrodingersket, this is bound to be a useful resource 🙌

dimitropoulos:
hey @schrodingersket just wanted to come back and say thank you (again) for the above.  we used it just now to work with another issue on the [insomnia stream](https://app.slack.com/client/T3C5CEYCD/C031YHPBA3G) and it worked very well!  we really appreciate it!

schrodingersket:
Whoohoo! Glad you found it to be a useful resource. (:

PR #3884, opened 2021-08-05
React ESLint rules

develohpanda:
Review commit-by-commit. Each commit adds a single rule then runs `npm run lint:fix`.

Open to changing any rule configs, but having a rule to make things consistent (and easier to dev) is a start.

PR #3897, opened 2021-08-10
Lint rules for equivalence, arrow func spacing and func returns

develohpanda:


.eslintrc.js
```diff
@@ -60,12 +60,15 @@ module.exports = {
     'camelcase': ['error', { allow: ['__export_format', '__export_date', '__export_source'] }],
     'comma-dangle': ['error', 'always-multiline'],
     'comma-spacing': 'error',
+    'consistent-return': 'warn',
```
develohpanda, line 63: Thoughts on this? Set to warning at the moment because it there are ~47 violations right now, but mostly older violations which we _should_ slowly fix and convert this to an error.

dimitropoulos: I'm conflicted.  On the one hand it's nice to force things to be specified, on the other hand I see this as a "this is how the language works at a very very basic level" thing.

Looking at what it catches, it does catch some interesting cases.  Here's an onClick that's (for no reason I can see) returning a value when really it should just not return (i.e. the final two returns should be removed).

Screenshot_20210810_130441


then there are others that are borderline in a different way (implicitly returning undefined and returning null is the same as far as react is concerned in this context):
Screenshot_20210810_130658

All and all, 47 doesn't seem like much.  If you want to turn this rule on and try it out I'm open minded, but we should fix these.  There are other rules that were recently added as warnings but not fixed (e.g. `prefer-stateless-function` and `no-array-index`) and I'd love to keep the number of active warnings pretty low.

develohpanda: > prefer-stateless-function

There are too many components to convert right now, but we will do them one by one, and then flip to an error. This rule will warn open source/new contributors from writing class components that can (and should) be written as functions.

> no-array-index

https://github.com/Kong/insomnia/pull/3884#discussion_r684529271

And it's the same with this rule I think; we can only fix things as we go if we can see what needs fixing. Warnings will show in the files we are working in, the same way TS shows warnings and errors in the files we are working in. I'd like to merge it in as a warning, so at least there are markers in our code to allow us to gradually fix (and we _are_ gradually fixing things)

develohpanda: > on the other hand I see this as a "this is how the language works at a very very basic level" thing.

This comment applies for the eqeqeq rule too, ensuring we use `===` instead of `==` 🤷🏽 just because we can because the language allows, IMO, doesn't mean we should

dimitropoulos: I don't disagree with https://github.com/Kong/insomnia/pull/3884#discussion_r684529271 at all - just that if there are exceptions we should be ignoring those lines and writing a description into those lines about why the situation is exceptional.


 ------
 
> There are too many components to convert right now, but we will do them one by one

Ok, I get that.. but whenabouts are you thinkin'?  What I'm ultimately commenting on is that we've enabled a few of these at once without doing the work (i.e. fixing or adding and documenting the ignores) of cleaning it up.

Don't get me wrong, I'm definitely not about to blockade a lint rule being added, haha, but I just want to make sure we're not adding more than we're keeping up with.

----

re: https://github.com/Kong/insomnia/pull/3897#discussion_r686337346 I think we're maybe talking about different things.  double vs triple equals are very different beasts with wildly different semantics.  `return` vs `return undefined` are literal equivalents.  it's a stylistic change at that point.  is that clearer?

develohpanda: > I don't disagree with #3884 (comment) at all - just that if there are exceptions we should be ignoring those lines and writing a description into those lines about why the situation is exceptional.

Noted, being more explicit. I like that approach, let's aim for it

> but I just want to make sure we're not adding more than we're keeping up with.

Personally I don't think we are, all wheels of the cart move together (features, ts fixes, lint fixes) but if we feel that it gets too noisy we can always dial it down again.

> re: #3897 (comment) I think we're maybe talking about different things. double vs triple equals are very different beasts with wildly different semantics. return vs return undefined are literal equivalents. it's a stylistic change at that point. is that clearer?

Yep! Clearer. I think we can adjust the rule too, I'm okay with `return` instead of `return undefined`, but want to avoid invoke-and-return in a single line when the invoked function (`bar()`) doesn't return anything. That becomes misleading, see following example. Thoughts?

```diff
const baz = (): void => {
	// ... do stuff but return nothing
}

const foo = () => {
	if (xyz) {
-		return bar();
+		bar();
+		return;
	}

	return;
}
```

dimitropoulos: > but want to avoid invoke-and-return in a single line when the invoked function (bar()) doesn't return anything.

I'm a big supporter of this!  let's do it!.  if it's bad... hey... we can always dial it back later.

----

> if we feel that it gets too noisy we can always dial it down again

To speak in these terms, then, I feel that it's too noisy at the moment this PR merges.  But I'm flexible and open minded.  Let's try out what you're suggesting and see how it goes.

dimitropoulos (approved):

.eslintrc.js
develohpanda: Sounds good. I agree, it'll be noise the instance this merges too, but if we aren't able to gradually reduce the noise and it becomes a _hindrance_ then we can revisit it

PR #3898, opened 2021-08-10
🧹 Use HelpTooltip from the component library and remove duplicate

develohpanda:
What it says on the tin, had some spare time while watching tv 😎 

The SVG Icon from `insomnia-components` doesn't quite look/size the same as the fontawesome icons, so this is draft until we have better icons


dimitropoulos (approved):
bonus points for the sfc conversion

PR #3903, opened 2021-08-11
Add segment metrics for plugin data context

develohpanda:
This is an initial attempt. When looking at other places segment is used, it seems like the even is a plain string, but please advise if the event should be something else.

We had to set up this fallback for the plugin API to avoid introducing a breaking change, but we should try to track how often this code-path is hit.

### DIFF
--- a/packages/insomnia-app/app/plugins/context/data.ts
+++ b/packages/insomnia-app/app/plugins/context/data.ts
@@ -1,3 +1,4 @@
+import { trackSegmentEvent } from '../../common/analytics';
 import { exportWorkspacesData, exportWorkspacesHAR } from '../../common/export';
 import type { ImportRawConfig } from '../../common/import';
 import { importRaw, importUri } from '../../common/import';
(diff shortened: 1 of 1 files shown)


DMarby (approved):

wdawson (approved):

packages/insomnia-app/app/plugins/context/data.ts
```diff
@@ -26,11 +27,15 @@ const buildImportRawConfig = (options: PluginImportOptions, activeSpaceId: strin
...
-  activeSpaceId
-    ? models.workspace.findByParentId(activeSpaceId)
-    : models.workspace.all();
+const getWorkspaces = (activeSpaceId?: string) => {
+  if (activeSpaceId) {
+    trackSegmentEvent('Plugin export loading workspaces for active space');
+    return models.workspace.findByParentId(activeSpaceId);
+  } else {
+    trackSegmentEvent('Plugin export loading all workspace');
+    return models.workspace.all();
```
dimitropoulos, line 36: @develohpanda should I take this to mean you'd not be interested in having https://eslint.org/docs/rules/no-else-return some day?

develohpanda: Not against adding that rule!

dimitropoulos: sweet.  didn't think so but thought I'd ask.

PR #3904, opened 2021-08-11
Introduce segment event enum

develohpanda:
Using this approach allows us to add, find usages, and modify the actual event string being sent in one location.


dimitropoulos:
> find usages

could't we just search `trackSegmentEvent`?  (not trying to nit-pick, just want to understand what your thought process was on that part)

(I'm a big fan of this change, btw)

packages/insomnia-app/app/common/analytics.ts
```diff
@@ -114,7 +114,22 @@ export async function getDeviceId() {
...
+  requestCreate = 'Request Created',
+  requestExecute = 'Request Executed',
+  testSuiteCreate = 'Test Suite Created',
+  testSuiteDelete = 'Test Suite Deleted',
+  unitTestCreate = 'Unit Test Created',
+  unitTestRunAll = 'Ran All Unit Tests',
+  unitTestRun = 'Ran Individual Unit Test',
+  unitTestDelete = 'Unit Test Deleted',
+  spaceLocalCreate = 'Local Space Created',
+  spaceLocalDelete = 'Local Space Deleted',
```
dimitropoulos:
```suggestion
  collectionCreate = 'Collection Created',
  documentCreate = 'Document Created',
  requestCreate = 'Request Created',
  requestExecute = 'Request Executed',
  spaceLocalCreate = 'Local Space Created',
  spaceLocalDelete = 'Local Space Deleted',
  testSuiteCreate = 'Test Suite Created',
  testSuiteDelete = 'Test Suite Deleted',
  unitTestCreate = 'Unit Test Created',
  unitTestDelete = 'Unit Test Deleted',
  unitTestRun = 'Ran Individual Unit Test',
  unitTestRunAll = 'Ran All Unit Tests',
```

any opposition to sorting these? :)

dimitropoulos (commented):
side note regarding the tests failing (`Cannot read property 'documentCreate' of undefined`):  I thought maybe this was to due with ts-jest having a hard time (perhaps because babel) with enums, but if you add `console.log(ForceToWorkspace.current);` before the line that's failing it works fine.  not sure if that's helpful to you but I didn't have more than a few seconds to look into it inbetween other work - but I thought I'd at least mention.

develohpanda:
> > find usages
> 
> could't we just search `trackSegmentEvent`? (not trying to nit-pick, just want to understand what your thought process was on that part)

Great question - not usages of the function, find usages of a particular event. Using an enum (for example) allows you to find everywhere a value from it is referenced, everywhere we send a particular event. Using a string union you will need to do a full-text search (AFAIK).

image

packages/insomnia-app/app/common/analytics.ts
develohpanda: Add a lint rule to automate it for me ;)

I don't like the wording either, but it means we can keep things grouped nicely as they get sorted. You win some, you lose some. I named them this way in anticipation of sorting haha.

develohpanda:
> side note regarding the tests failing (Cannot read property 'documentCreate' of undefined): I thought maybe this was to due with ts-jest having a hard time (perhaps because babel) with enums, but if you add console.log(ForceToWorkspace.current); before the line that's failing it works fine. not sure if that's helpful to you but I didn't have more than a few seconds to look into it inbetween other work - but I thought I'd at least mention.

Looks like it's (potentially) a result of [circular imports](https://github.com/typescript-eslint/typescript-eslint/issues/561), but if an enum is exported from it's own individual file as a top-level import, it's fine (and better than using const enums IMO)

I think lets cautiously roll with enums, unless there is a strong preference to _not_ use them

dimitropoulos:
ahh.  I thought it might be that.  I tried to look for what it might be but with as many imports as the source and destination files had I gave up pretty fast, heh.

I'm not a fan of enums since discriminated unions got so powerful.  I even think that if discriminated unions had been this powerful from the beginning that enums would never have happened.

Still, though, it's like... it's _fine_. haha.

Let's go with this for now - it's an improvement (I think we would all agree on that) - and see where it takes us.

_____

If there's anything I'd suggest it's that we consider making a zero-dependency package, `entities` (or something, not married to the name), that can contain just the types (and enums) we need throughout the monorepo.  That way there's never a circular dependency and we can also have a good place to share types.  Something like this enum would be a good candidate for that package.

We could, of course, take it a step further and (instead) make a (also zero dependency (well, zero _internal_ dependencies)) package, `common` (or whatever name we decide) that will have all the things I mentioned above _as well as_ any common code that we might want in other places.  E.g. all the helper functions and type guards like `isWorkspace` and all the constants and such.

Just a thought.

dimitropoulos (approved):

develohpanda:
> If there's anything I'd suggest it's that we consider making a zero-dependency package, `entities` (or something, not married to the name), that can contain just the types (and enums) we need throughout the monorepo. That way there's never a circular dependency and we can also have a good place to share types. Something like this enum would be a good candidate for that package.
> 
> We could, of course, take it a step further and (instead) make a (also zero dependency (well, zero _internal_ dependencies)) package, `common` (or whatever name we decide) that will have all the things I mentioned above _as well as_ any common code that we might want in other places. E.g. all the helper functions and type guards like `isWorkspace` and all the constants and such.
> 
> Just a thought.

Yep I'm on board with that and definitely think it would be useful

PR #3905, opened 2021-08-11
adds operator spacing rules

dimitropoulos:
I noticed the following line in a recent PR

```ts
const segmentEvent = design ? SegmentEvent.documentCreate: SegmentEvent.collectionCreate;
```

This PR would (autofix) the following:
```diff
- const segmentEvent = design ? SegmentEvent.documentCreate: SegmentEvent.collectionCreate;
+ const segmentEvent = design ? SegmentEvent.documentCreate : SegmentEvent.collectionCreate;
```

I ran `lint:fix` and there are presently no violations to these two rules in the project.

until the next eslint PR!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -81,6 +81,8 @@ module.exports = {
     'quotes': 'off',
     'semi': ['error', 'always'],
     'space-before-function-paren': ['error', { anonymous: 'never', named: 'never', asyncArrow: 'always' }],
+    'space-infix-ops': 'off', // successor('@typescript-eslint/space-infix-ops')
+    'space-unary-ops': 'error',
     'space-in-parens': 'error',
     'spaced-comment': ['error', 'always', {
       exceptions: ['/', '*', '-', '* '], // for ASCII art :)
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -81,6 +81,8 @@ module.exports = {
     'quotes': 'off',
     'semi': ['error', 'always'],
     'space-before-function-paren': ['error', { anonymous: 'never', named: 'never', asyncArrow: 'always' }],
+    'space-infix-ops': 'off', // successor('@typescript-eslint/space-infix-ops')
```
develohpanda, line 84: I noticed this in a couple of other places too (eg the `quotes` rule is off), but didn't realize that's because it is superseeded by a ts-eslint rule. This is one place where I'm not a fan of ordering, because contextually the rules can't be kept together. I like that in this instance you've added a comment to explain why it's off, but just an observation about some of the other rules that are off, and not immediately obvious as to why.

develohpanda (approved):
🚢

.eslintrc.js
dimitropoulos: I hear you, but it gets even worse than that.  some of the "extension" rules don't map 1:1 to the eslint rules.

beyond that, I think there's work afoot to get the typescript-specific stuff for the extension rules built into eslint core (but there's a lot of debate about it).

If you WANNNNNTTTT we can go with the approach of https://github.com/dimitropoulos/eslint-config-intense/blob/master/index.js#L6 which will cause us to be much more clear about _why_ a rule is off.  I'd be more than happy to make such a (noop) PR - but I didn't want to push it unless others here want it.  I know it's very "dimitri" to do it the way I've done it there.

develohpanda: That seems like a nice approach to force a reason for why something is off, let's try it 👀

PR #3908, opened 2021-08-12
Fix footer subtext alignment in modals

gatzjames:
This removes the `tall` className from the modal footer subtext to fix the vertical alignment of it.

After some investigation this was introduced by the Electron upgrade from v9 to v11 (and corresponding Chrome versions).
It seems that children of a flex parent with a % height were not working correctly before.
So applying the `height: 100%` property on the children had no effect and would result in the children being vertically aligned because of the parent's `align-items: center` property.
After the upgrade the children would occupy 100% height and the text would v-align on the top.

Had some time to explore all the UX flows for the modals changed so here are all the modals after the update:

CookiesModal:

CookiesModal

CodePromptModal:

CodePromptModal

GenerateClientCodeModal:

GenerateClientCodeModal

PromptModal:

PromptModal

EnvironmentsModal:

EnvironmentsModal

RequestCreateModal:

RequestCreateModal

SyncStagingModal:

SyncStagingModal

GitBranchesModal:

GitBranchesModal

GitStagingModal:

GitStagingModal

GitLogModal:

GitLogModal

GitRepositorySettingsModal:

GitRepositorySettingsModal

Closes INS-831


dimitropoulos (approved):
can confirm:

Screenshot_20210812_104949

PR #3910, opened 2021-08-12
Fix markdown-preview styling regression

gatzjames:
After upgrading the markdown-preview title we moved the compiled html into another element.
This changed how the styles get applied to the compiled html since the `markdown-preview__content` className
is not in the parent element anymore.

This solution introduces a new className `markdown-preview__content-title` which applies the same styles to the title so that it appears exactly the same as before.
I also moved the `markdown-preview__content` className to the compiled html's parent element for the styles to be applied correctly.

## Regression:

regression_840

## Demo:

fix_840

Closes INS-840


johnwchadwick (approved):
Ahhh, that makes sense. Whoops!

PR #3913, opened 2021-08-12
Update mock implementation and move the segment event enum

develohpanda:
Follow up to #3904


packages/insomnia-app/app/common/__mocks__/analytics.ts
```diff
@@ -1,4 +1,5 @@
 // WARNING: changing this to `export default` will break the mock and be incredibly hard to debug. Ask me how I know.
-const trackEvent = jest.fn();
-const trackSegmentEvent = jest.fn();
-module.exports = { trackEvent, trackSegmentEvent };
+const _analytics = jest.requireActual('../analytics');
+_analytics.trackEvent = jest.fn();
+_analytics.trackSegmentEvent = jest.fn();
+module.exports = _analytics;
```
dimitropoulos, line 5: at this point I almost wanna sorta ask if there's some way we can prevent ourselves from stepping on this kind of nail in the future.  even, like, documentation of why it's so funky would be cool I guess as a minimum.  I feel like it's been nothing but pain with the subtle intricacies of the jest mocks...

this is one thing that definitely got much worse in TypeScript (maybe, the only thing?) to grok

dimitropoulos (approved):

packages/insomnia-app/app/common/__mocks__/analytics.ts
develohpanda: I don't think it got any better or worse with TS, mocks have just always been a pain because it's so hard to have type checking there :(

dimitropoulos: I just mean that TS added another point of indirection / point of failure / thing to think about

PR #3918, opened 2021-08-16
add vscode debug configuration for insomnia-app

gatzjames:
Adds a vscode launch config to run insomnia from vscode and debug both the main and renderer processes.

debug

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,66 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Launch: Electron",
+      "type": "node",
+      "request": "launch",
+      "protocol": "inspector",
+      "sourceMaps": true,
+      "presentation": {
+        "hidden": false,
+        "group": "Insomnia",
+        "order": 1
+      },
+      "cwd": "${workspaceFolder}/packages/insomnia-app",
+      "runtimeExecutable": "${workspaceFolder}/packages/insomnia-app/node_modules/.bin/electron",
+      "runtimeArgs": ["--remote-debugging-port=9222", "."],
+      "outputCapture": "std",
+      "sourceMapPathOverrides": {
+        "webpack:///./*": "${workspaceFolder}/packages/insomnia-app/app/*"
+      },
+      "windows": {
+        "type": "node",
+        "request": "launch",
+        "name": "Launch: Electron",
+        "runtimeExecutable": "${workspaceFolder}/packages/insomnia-app/node_modules/.bin/electron.cmd"
+      },
+      "env": {
+        "TS_NODE_PROJECT": "tsconfig.webpack.json",
+        "NODE_ENV": "development",
+        "ELECTRON_IS_DEV": "1"
+      }
+    },
+    {
+      "name": "Attach: Electron Renderer",
+      "type": "chrome",
+      "request": "attach",
+      "presentation": {
+        "hidden": false,
+        "group": "Insomnia",
+        "order": 2
+      },
+      "port": 9222,
+      "sourceMapPathOverrides": {
+        "webpack:///./*": "${workspaceFolder}/packages/insomnia-app/app/*"
+      },
+      "webRoot": "${workspaceFolder}/packages/insomnia-app/app",
+      "timeout": 60000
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,66 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Launch: Electron",
```
develohpanda: Should this be called `Attach: Electron Main`? It's not immediately clear to me in which situation I would use each of these configurations

image

.gitignore
```diff
@@ -23,7 +23,6 @@ node_modules/
 .idea
 .DS_Store
 *test-plugins
-.vscode
 graphql.config.json
```
dimitropoulos: (line unrelated)

This is what I get with `Launch: Electron`.
Screenshot_20210817_163119

I'm using https://github.com/tj/n
Screenshot_20210817_163616

develohpanda:
Hey @gatzjames, what is the plan/status for this PR? Is it blocked on something at the moment?

.vscode/launch.json
gatzjames: I went with the name of the configuration type since the first one launches electron (and auto-attaches a debugger) while the second one attaches a debugger to a remote already running process.
Agree that it's not pretty clear but thinking that 'Attach: Electron Main' might be more confusing since it uses a launch configuration.
Perhaps we could ditch the `:` and make it a bit more clear e.g.
- `Launch Electron Main`
- `Attach to Electron Renderer`

develohpanda: I just worked through using this. What are your thoughts on actually hiding these two options? Realistically all we need is the compound configuration, maybe titled "Launch and Attach" or "Debug". And after hiding the individual tasks from the dropdown, we could name them `Electron: main` and `Electron: renderer`, so that while debugging they show as such.

Could you describe a workflow in which the `Launch Electron Main` and `Attach to Electron Renderer` configurations might be launched individually?

I made this change in 62894c32b5911e1ae67263b568b826efbe949442 (because it was easier to demonstrate with an example rather than describe here, feel free to remove)

.vscode/launch.json
```diff
@@ -0,0 +1,66 @@
...
+      },
+      "env": {
+        "TS_NODE_PROJECT": "tsconfig.webpack.json",
+        "NODE_ENV": "development",
+        "ELECTRON_IS_DEV": "1"
+      }
+    },
+    {
+      "name": "Attach: Electron Renderer",
+      "type": "chrome",
```
develohpanda: It looks like the chrome extension has been deprecated in favor of https://github.com/microsoft/vscode-js-debug which is now built in. This supports chrome, supposedly, but I couldn't get it to work 🤔 seems like setting this type to `pwa-chrome` worked, and that's one of the supported types registered by this built-in extension (https://github.com/microsoft/vscode-js-debug/blob/main/OPTIONS.md). Maybe the docs are out of date? How did you get this to work with `"type": "chrome"`?

```suggestion
      "type": "pwa-chrome",
```

develohpanda:
Just tested with the GraphQL and gRPC collections in the "Insomnia Testing" Project (which you should have access to); seems to be working well!

.gitignore
dimitropoulos: works for me now.  I'm trying on a computer that has node installed with `n` but a different configuration it seems:

```console
x1@x1:~/src/github.com/Kong/insomnia$ which node
/usr/local/bin/node
```

With my computer configured like this - it works great.  I think this is a "me" thing for why it didn't work before so we can proceed.  This isn't the only tool I have on that other computer (above) that has issues with extensions not being able to find node.

dimitropoulos (approved):
just ran though everything and it's definitely working!  I'm sure we'll have ideas/things to improve on this in the future but I'm really happy to be able to use this!  awesome!

.gitignore
dimitropoulos: in case it comes up again in the future, I got the original failure to succeed by running:

```shell
sudo ln --symbolic --force ~/n/bin/node /usr/local/bin/node
```

(i.e. symlinking where node was installed to where it was expected to install)

PR #3920, opened 2021-08-17
Mutate the error message instead of throwing a new error

develohpanda:
Catching and throwing a new error loses the stack trace, this PR mutates the error messages and re-throws the same error.

### DIFF
--- a/packages/openapi-2-kong/src/declarative-config/generate.ts
+++ b/packages/openapi-2-kong/src/declarative-config/generate.ts
@@ -25,7 +25,11 @@ export function generateDeclarativeConfigFromSpec(
     // see: https://github.com/Kong/studio/issues/93
     const result: DeclarativeConfigResult = JSON.parse(JSON.stringify(declarativeConfigResult));
     return result;
-  } catch (err) {
-    throw new Error('Failed to generate spec: ' + err.message);
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      err.message = `Failed to generate spec: ${err.message}`;
+      throw err;
+    }
+    throw err;
   }
 }


packages/openapi-2-kong/src/declarative-config/generate.ts
```diff
@@ -25,7 +25,11 @@ export function generateDeclarativeConfigFromSpec(
     // see: https://github.com/Kong/studio/issues/93
     const result: DeclarativeConfigResult = JSON.parse(JSON.stringify(declarativeConfigResult));
     return result;
-  } catch (err) {
-    throw new Error('Failed to generate spec: ' + err.message);
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      err.message = `Failed to generate spec: ${err.message}`;
+      throw err;
```
gatzjames, line 31: Looks good to me!
Do we need to throw the error here since we also throw it below?

develohpanda: Oh good point, we don't have to throw here!

dimitropoulos (commented):
compare that to:


on second thought.. maybe this doesn't do what we think it's doing.  Thoughts?

____

UPDATE

This, it turns out, is the _stack_ property.  See the following:


So what that means, then, is that if we rewrite `Error.message` we also need to rewrite `Error.stack` because otherwise it will contain the old information.

develohpanda:
@dimitropoulos that's a great point. I'm going to close this PR and will improve the error handling with INS-874. I don't think prepending `Failed to generate spec` is useful anyway, we should just throw and let the caller handle how to present an error to users.

PR #3923, opened 2021-08-17
Space Sync Error improvement

dimitropoulos:
closes INS-788

### DIFF
--- a/packages/insomnia-app/app/sync/vcs/util.ts
+++ b/packages/insomnia-app/app/sync/vcs/util.ts
@@ -1,6 +1,7 @@
 import clone from 'clone';
 import crypto from 'crypto';
 
+import { strings } from '../../common/strings';
 import { BaseModel } from '../../models';
 import { deleteKeys, resetKeys, shouldIgnoreKey } from '../ignore-keys';
 import { deterministicStringify } from '../lib/deterministicStringify';
@@ -552,3 +553,30 @@ export function describeChanges<T extends BaseModel>(a: T, b: T): string[] {
 
   return changes;
 }
+
+// delete collection, branches, push, pull
+export const interceptAccessError = async <T>(
+  {
+    callback,
+    action,
+    resourceName,
+    resourceType = strings.collection.singular.toLowerCase(),
+  }: {
+    callback: () => T | Promise<T>,
+    action: string;
+    resourceName: string;
+    resourceType?: string;
+  }
+) => {
+  try {
+    console.log('trying');
+    return await callback();
+  } catch (error: unknown) {
+    console.log('catching', error);
+    if (error instanceof Error && error.message.includes('invalid access')) {
+      // TODO make sure the stack trace remains
+      throw new Error(`You no longer have permission to ${action} the "${resourceName}" ${resourceType}.  Contact your team administrator if you think this is an error.`);
+    }
+    throw error;
+  }
+};
--- a/packages/insomnia-app/app/sync/vcs/util.ts
+++ b/packages/insomnia-app/app/sync/vcs/util.ts
@@ -554,7 +554,6 @@ export function describeChanges<T extends BaseModel>(a: T, b: T): string[] {
   return changes;
 }
 
-// delete collection, branches, push, pull
 export const interceptAccessError = async <T>(
   {
     callback,
@@ -569,12 +568,9 @@ export const interceptAccessError = async <T>(
   }
 ) => {
   try {
-    console.log('trying');
     return await callback();
   } catch (error: unknown) {
-    console.log('catching', error);
     if (error instanceof Error && error.message.includes('invalid access')) {
-      // TODO make sure the stack trace remains
       throw new Error(`You no longer have permission to ${action} the "${resourceName}" ${resourceType}.  Contact your team administrator if you think this is an error.`);
     }
     throw error;
--- a/packages/insomnia-app/app/sync/vcs/vcs.ts
+++ b/packages/insomnia-app/app/sync/vcs/vcs.ts
@@ -102,11 +102,11 @@ export class VCS {
   async switchProject(rootDocumentId: string) {
     const project = await this._getProjectByRootDocument(rootDocumentId);
 
-    if (project !== null) {
-      await this.setProject(project);
-    } else {
-      this._project = null;
+    if (project === null) {
+      return;
     }
+
+    await this.setProject(project);
   }
 
   async switchAndCreateProjectIfNotExist(rootDocumentId: string, name: string) {
--- a/packages/insomnia-app/app/sync/vcs/vcs.ts
+++ b/packages/insomnia-app/app/sync/vcs/vcs.ts
@@ -102,11 +102,11 @@ export class VCS {
   async switchProject(rootDocumentId: string) {
     const project = await this._getProjectByRootDocument(rootDocumentId);
 
-    if (project === null) {
-      return;
+    if (project !== null) {
+      await this.setProject(project);
+    } else {
+      this._project = null;
     }
-
-    await this.setProject(project);
   }
 
   async switchAndCreateProjectIfNotExist(rootDocumentId: string, name: string) {
--- a/packages/insomnia-app/app/ui/hooks/workspace.ts
+++ b/packages/insomnia-app/app/ui/hooks/workspace.ts
@@ -2,9 +2,7 @@ import { Reducer, useCallback, useMemo, useReducer } from 'react';
 import { useSelector } from 'react-redux';
 import { useAsync } from 'react-use';
 
-import { getDataDirectory } from '../../common/electron-helpers';
 import { isRemoteSpace } from '../../models/space';
-import FileSystemDriver from '../../sync/store/drivers/file-system-driver';
 import { Project } from '../../sync/types';
 import { ProjectWithTeam } from '../../sync/vcs/normalize-project-team';
 import { pullProject } from '../../sync/vcs/pull-project';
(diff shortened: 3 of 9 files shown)


packages/insomnia-app/app/sync/vcs/util.ts
```diff
@@ -552,3 +553,30 @@ export function describeChanges<T extends BaseModel>(a: T, b: T): string[] {
...
+  }
+) => {
+  try {
+    console.log('trying');
+    return await callback();
+  } catch (error: unknown) {
+    console.log('catching', error);
+    if (error instanceof Error && error.message.includes('invalid access')) {
+      // TODO make sure the stack trace remains
+      throw new Error(`You no longer have permission to ${action} the "${resourceName}" ${resourceType}.  Contact your team administrator if you think this is an error.`);
```
dimitropoulos, line 574: I'm thinkin' since we don't actually use the stack trace, there's nothing really being lost by the re-throw working as it does.  thoughts?

develohpanda: I'm ok with that in order to not block this PR. In the future it might be better to throw custom errors and send the existing error internally, following something like this example. In general I don't think it's a good idea to swallow stack traces just because they're not used _right now_

https://github.com/Kong/insomnia/blob/0ef92797cdad1606df31bbf34f9047670f0c609d/packages/insomnia-inso/src/errors.ts#L3-L22

packages/insomnia-app/app/sync/vcs/util.ts
```diff
@@ -552,3 +553,30 @@ export function describeChanges<T extends BaseModel>(a: T, b: T): string[] {
...
+    resourceType = strings.collection.singular.toLowerCase(),
+  }: {
+    callback: () => T | Promise<T>,
+    action: string;
+    resourceName: string;
+    resourceType?: string;
+  }
+) => {
+  try {
+    console.log('trying');
```
develohpanda:
```suggestion
```

packages/insomnia-app/app/sync/vcs/util.ts
```diff
@@ -552,3 +553,30 @@ export function describeChanges<T extends BaseModel>(a: T, b: T): string[] {
...
+    return await callback();
+  } catch (error: unknown) {
+    console.log('catching', error);
```
develohpanda:
```suggestion
```

packages/insomnia-app/app/sync/vcs/vcs.ts
```diff
@@ -102,11 +102,11 @@ export class VCS {
...
 
-    if (project !== null) {
-      await this.setProject(project);
-    } else {
-      this._project = null;
+    if (project === null) {
+      return;
     }
+
+    await this.setProject(project);
```
develohpanda: This change doesn't make sense; now, if a project is not found by `_getProjectByRootDocument` we exit instead of setting `this._project` to null...?

packages/insomnia-app/app/sync/vcs/util.ts
dimitropoulos: ugg. sorry.  I did a self review (before the squash) and removed a few of these but clearly I didn't get them all.

dimitropoulos: fixed in 285787a7948940524d3bc59451c0df83a86da6ce

packages/insomnia-app/app/sync/vcs/util.ts
dimitropoulos: 285787a7948940524d3bc59451c0df83a86da6ce

packages/insomnia-app/app/sync/vcs/vcs.ts
dimitropoulos: I don't understand your feedback, could you please rephrase?  Before it was doing "if it's not equal to null set it otherwise set it equal to null".  now it's saying "if it's already null then you're all done.  if not, then set it"

dimitropoulos: I'm not married to this change being in this PR, btw, if you wanna have it removed - just seemed like an obvious change to make when we were looking at it to increase reading comprehension of the thing we were trying to debug to make this PR happen.

develohpanda: but `project`, the constant in this scope being checked in the if condition, is different to `this._project`, the class property being set in the if-else block

develohpanda: The previous implementation was:

```ts
const foo = await this._getProjectByRootDocument();
if (foo !== null) {
	await this.setProject(foo);
} else {
	this._project = null;
}
```

But the updated implementation is

```ts
const foo = await this._getProjectByRootDocument();
if (foo === null) {
	return;
}

await this.setProject(foo);
```

The two are fundamentally different, there's nothing that is setting `this._project = null` in the updated implementation

develohpanda: (I posted this before but GH landed in a bugged state...)

The previous implementation was:

```ts
const foo = await this._getProjectByRootDocument();
if (foo !== null) {
	await this.setProject(foo);
} else {
	this._project = null;
}
```

But the updated implementation is

```ts
const foo = await this._getProjectByRootDocument();
if (foo === null) {
	return;
}

await this.setProject(foo);
```

The two are fundamentally different, there's nothing that is setting `this._project = null` in the updated implementation

DMarby: This definitely needs to be reverted, after this change, the state of the VCS is not correctly cleared in some cases when switching from a workspace with sync setup to a workspace that does not have sync setup:

packages/insomnia-app/app/ui/hooks/workspace.ts
```diff
@@ -89,7 +91,8 @@ export const useRemoteWorkspaces = (vcs?: VCS) => {
 
     try {
       // Clone old VCS so we don't mess anything up while working on other projects
-      const newVCS = vcs.newInstance();
+      const driver = FileSystemDriver.create(getDataDirectory());
+      const newVCS = new VCS(driver);
```
DMarby: Please revert this change as well.

While this change **should** work correctly, given how brittle our sync-related code is and the customer impact of things breaking, and that this PR needs to be included in the upcoming release, let's try to avoid any changes that are not strictly needed.

DMarby (changes requested):

packages/insomnia-app/app/ui/hooks/workspace.ts
dimitropoulos: sure: @develohpanda sent me this code and asked me to change it while we were working on it - happy to remove.  I think he was saying we are trying to move all of the new instance methods to something like this.

packages/insomnia-app/app/sync/vcs/vcs.ts
dimitropoulos: sorry, my brain crossed `this.project` and `project`.  definitely a bad change.  @develohpanda and I (but ultimately me) made this change in a live share when we were debugging like 4 things at once and I think that's where blindness came from.  neither of us noticed at the time - glad we caught it now.  thanks.  fixed in 75c750c0dc53e4b876cc438202ebf5b7e638a43e

packages/insomnia-app/app/ui/hooks/workspace.ts
dimitropoulos: removed in 5956617b069b2348d303062244908d27c730779a

packages/insomnia-app/app/sync/vcs/util.ts
dimitropoulos: since we were both working on this simultaneously should I wait for the solution you come up with after https://github.com/Kong/insomnia/pull/3920 or is now a good time to make the helper, myself?  Since it'll be a good few hours until you get a chance to read this I think I might try (in a separate PR).

develohpanda: That PR is unrelated to this area (that's to do with O2K) and there's no simultaneous work happening so no risk of conflict or overwrite. Up to you how you'd like to go about it :)

DMarby (changes requested):
Seems to work well for the operations that it's intercepting, it's not currently intercepting pulls or pushes as it should however:


To intercept that (as well as pulls/pushes directly from the sync menu), the following pull and push calls needs to be intercepted as well: https://github.com/Kong/insomnia/blob/0ef92797cdad1606df31bbf34f9047670f0c609d/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx#L229
https://github.com/Kong/insomnia/blob/0ef92797cdad1606df31bbf34f9047670f0c609d/packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx#L249

It's also not intercepting the loading of remote branches in the branches modal, but that's probably less important: 
image

packages/insomnia-app/app/sync/vcs/util.ts
dimitropoulos: see: https://github.com/Kong/insomnia/pull/3925

packages/insomnia-app/app/ui/hooks/workspace.ts
dimitropoulos: fixed in 5956617b069b2348d303062244908d27c730779a

develohpanda: Just for context here, because I didn't ask to make this improvement just because we're in the area.

We were debugging the class instance error we faced with the VCS class, but that ended up being related to closures and not how the instance was being created. Although this is something we do need to look into, it was no longer necessary for this PR so in the future we should remove our WIP changes

dimitropoulos:
@DMarby here's an example of the push error (pull was also protected):
Screenshot_20210819_155117


and here's an example of the remote ones
Screenshot_20210819_155043

check out 7b94cfee855db8914c60896ef4d976b7b3d33d46

develohpanda:
Code LGTM, haven't had a chance to QA

DMarby (approved):

PR #3928, opened 2021-08-19
Space -> Project rename & update code-references

develohpanda:
This PR is to rename "Spaces" to "Projects". It also updates code references.

Some things of note:
* the database id prefix has been changed (from `sp` to `proj`)
	* What this means is, while developing locally, if you had previously pulled remote spaces, they would now be duplicated (because the generated id has changed due to the prefix), but this is only going to impact those of us who have been actively developing. If working as an update from 2021.4.1 or a fresh install, users will not see the duplication 
	* I chose `proj` instead of `prj` to at least differentiate from "projects" on the backend (which represent a different entity) image
* the local storage keys have been updated (both changed keys are new since 2021.4.1)
* the segment events have been updated
* existing mentions of a "project" have been updated to "backendProject"
	* this is because in the current state a "project" in the backend is a collection in the app and not the same entity that is labeled a "Project" in the app - the two represent different things

Aside: yay for using `strings`, most places didn't need any extra change.

<details>
<summary>Screenshots</summary>

image

image

image

image

</details>


packages/insomnia-app/app/common/analytics.ts
```diff
@@ -121,8 +121,8 @@ export enum SegmentEvent {
   pluginExportLoadWorkspacesInSpace = 'Plugin export loading workspaces for active space',
```
DMarby:
```suggestion
  pluginExportLoadWorkspacesInSpace = 'Plugin export loading workspaces for active project',
```

packages/insomnia-app/app/ui/redux/modules/__tests__/space.test.js
```diff
@@ -80,7 +80,7 @@ describe('space', () => {
       } = getAndClearShowAlertMockArgs();
 
       expect(title).toBe('Delete Space');
```
DMarby: Not a blocker for this PR at all, but there's a bunch of references to `Space` in this file in UI/tracking asserts here still. Since the build still passes, I think something might be wrong with these tests?

packages/insomnia-app/app/models/space.ts
```diff
@@ -5,7 +5,7 @@ import type { BaseModel } from './index';
 
 export const name = 'Space';
 export const type = 'Space';
-export const prefix = 'sp';
+export const prefix = 'proj';
```
DMarby: Does the `name` down in the `init()` function for this model need to be updated as well, or is that never used/shown to the user?

DMarby (dismissed):
:shipit:

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -388,7 +388,7 @@ export const setActiveSpace = (spaceId: string) => {
 };
 
 export const setSpaceSortOrder = (sortOrder: SpaceSortOrder) => {
-  const key = `${LOCALSTORAGE_PREFIX}::space-sort-order`;
+  const key = `${LOCALSTORAGE_PREFIX}::dashboard-sort-order`;
```
dimitropoulos, line 392: In `global.tsx` exists the line:
```ts
export const SET_SPACE_SORT_ORDER = 'global/space-sort-order';
```

Should we, then, change that too?  I fixed it in 389b1a0cbd7719f6e42706476bab0d02d62b7079.

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -379,7 +379,7 @@ export const setActiveActivity = (activity: GlobalActivity) => {
 };
 
 export const setActiveSpace = (spaceId: string) => {
-  const key = `${LOCALSTORAGE_PREFIX}::activeSpaceId`;
+  const key = `${LOCALSTORAGE_PREFIX}::activeProjectId`;
```
dimitropoulos, line 383: I feel strongly that we should not let the key in localstorage fall out of sync with what's in the code.  it took me 4 `f12` renames and 50 seconds (I timed it) to rename the remaining `activeProjectId`s.  I pushed that change to 921b434246a299f065a791a222338776bdce0110.  I don't see what we stand to gain by not fixing this now, but I'm open minded to hear something I may not be considering.

packages/insomnia-app/app/ui/redux/modules/__tests__/space.test.js
develohpanda: Great question, it's possible this wasn't running because it's somehow a JS file. I have updated the file and will make sure the tests run and make sense

packages/insomnia-app/app/models/space.ts
develohpanda: It's never shown, but has also been updated now

packages/insomnia-app/app/ui/redux/modules/global.tsx
develohpanda: Now that code-references are being updated too, yes 👍🏽

packages/insomnia-app/app/ui/redux/modules/global.tsx
develohpanda: Now that code-references are being updated too, this is fine by me 👍🏽

>  it took me 4 f12 renames and 50 seconds (I timed it)

FWIW, my concern with renames is not how long it takes to rename, it's the increased scope of change and testing necessary as a result (esp in this case). Ideally these keys are all constants defined in one place and reused everwhere.

packages/insomnia-app/app/__jest__/redux-state-for-test.ts
```diff
@@ -9,8 +9,8 @@ export const reduxStateForTest = async (global: Partial<GlobalState> = {}): Prom
   global: {
     activeWorkspaceId: null,
     activeActivity: ACTIVITY_HOME,
-    activeSpaceId: BASE_SPACE_ID,
-    spaceSortOrder: 'modified-desc',
+    activeProjectId: BASE_PROJECT_ID,
+    projectSortOrder: 'modified-desc',
```
develohpanda: @dimitropoulos I had switched this from `spaceSortOrder` to `dashboardSortOrder` to intentionally decouple the model name, because the sorting is applied to the dashboard, not to the list of projects. I still think this should be `dashboardSortOrder` and will update in a commit, which we can remove should we choose to.

packages/insomnia-app/app/models/index.ts
```diff
@@ -42,8 +42,8 @@ export interface BaseModel {
   _id: string;
   type: string;
   // TSCONVERSION -- parentId is always required for all models, except 4:
-  //   - Stats, Settings, and Space, which never have a parentId
-  //   - Workspace optionally has a parentId (which will be the id of a Space)
+  //   - Stats, Settings, and Project, which never have a parentId
+  //   - Workspace optionally has a parentId (which will be the id of a project)
```
develohpanda:
```suggestion
  //   - Workspace optionally has a parentId (which will be the id of a Project)
```

packages/insomnia-app/app/sync/vcs/__tests__/initialize-backend-project.test.ts
```diff
@@ -5,30 +5,30 @@ import * as models from '../../../models';
 import { snapshotStateEntrySchema } from '../../__schemas__/type-schemas';
 import MemoryDriver from '../../store/drivers/memory-driver';
 import { Snapshot } from '../../types';
-import { initializeLocalProjectAndMarkForSync, pushSnapshotOnInitialize } from '../initialize-project';
+import { initializeLocalBackendProjectAndMarkForSync, pushSnapshotOnInitialize } from '../initialize-backend-project';
 import { VCS } from '../vcs';
 
 const snapshotStateBuilder = createBuilder(snapshotStateEntrySchema);
 
 describe('initialize-project', () => {
```
develohpanda:
```suggestion
describe('initialize-backend-project', () => {
```

packages/insomnia-app/app/sync/vcs/vcs.ts
```diff
@@ -491,11 +491,11 @@ export class VCS {
   }
 
   async shareWithTeam(teamId: string) {
```
develohpanda, line 493: Not removing in this PR, but some of these share methods are no longer used and can be safely removed

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
```diff
@@ -518,7 +518,7 @@ class UnconnectedSyncDropdown extends PureComponent<Props, State> {
             {syncMenuHeader}
             {remoteProjects.length === 0 && (
               <DropdownItem onClick={this._handleEnableSync}>
-                <i className="fa fa-plus-circle" /> Create Local Project
+                <i className="fa fa-plus-circle" /> Create Locally
```
develohpanda, line 534: If Insomnia doesn't enable sync automatically (eg. collection imported into a remote space), you have to follow a manual setup sync process (same as currently in the stable release). The first step of this manual setup process contains the terminology "Create Local Project", to initialize a local backend project. What should this be renamed to? For the time being I have renamed to "Create Locally".

** @wdawson should this be "Create Locally" or "Create Local Backend Project" or something else?**

#### This PR
image

#### 2021.4.1
image

packages/insomnia-app/app/ui/components/sidebar/sidebar-request-group-row.tsx
```diff
@@ -23,7 +23,7 @@ import { SidebarRequestRow } from './sidebar-request-row';
 type ReduxProps = ReturnType<typeof mapStateToProps>;
 
 const mapStateToProps = (state: RootState) => ({
-  activeSpace: selectActiveRequest(state),
+  activeProject: selectActiveRequest(state), // TODO this seems wrong (!?!)
```
develohpanda, line 26: Woah what the heck, good catch. Seems to not be used at all so at least we're safe on that front :thinking:

(Thanks for marking to change and not changing in this PR :heart:)

packages/insomnia-app/app/ui/redux/modules/__tests__/space.test.js
develohpanda: Fixed here: 9c73f4fb3f34f7c6593fac6dfa7c97a81bf05783

packages/insomnia-app/app/ui/components/dropdowns/sync-dropdown.tsx
wdawson: I would vote for something like: "Initialize [Collection|Design Document]" if possible. It's confusing to me that the only option would be to do something locally when the whole point of that button is to start a sync. So to me, "initialize" sounds like the first of a few steps and it's reminiscent of a new git repo.

develohpanda: I think the key note here is that it's only initializing locally and that nothing exists on the server until you create and push a snapshot. We can streamline this process a bit more in future to avoid this interim state.

Would a straightforward `Initialize` be sufficient? (This is only available with collections, not documents)

packages/insomnia-app/app/__jest__/redux-state-for-test.ts
develohpanda: 9d377d52e98748b21e024bdf2ccdeb9fc73afc81

PR #3931, opened 2021-08-20
adds markdown linting

dimitropoulos:
# UPDATE

This PR has been converted into a PR just focusing on Markdown*

*with a single line of scope creep to add the eslint plugin to the recommends

_______________________

I have been meaning to add this for a while.  What it does is that when someone opens our repo for the first time and does not have a recommended extension, it will show the user a prompt recommending they review and install the workspace recommendations.

read more about it here https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

I also added some basic in-project configuration for the markdown linter.  Hopefully very standard stuff we all already agree on but as always - I'd love to hear suggestions.


.vscode/extensions.json
```diff
@@ -0,0 +1,16 @@
+{
```
develohpanda, line 1: While this is probably okay we shouldn't assume all users on the team and in open-source will use VS Code. Because we have not added the `markdownlint-cli` it means any users not using VS Code won't be able lint (or fix) their markdown locally. Having these recommended extensions is awesome, and in addition I think we should add `markdownlint-cli` into lerna root and include it with the `lint` and `lint:fix` commands

.vscode/extensions.json
```diff
@@ -0,0 +1,16 @@
...
+  // See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+  // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+  // List of extensions which should be recommended for users of this workspace.
+  "recommendations": [
+    "yzhang.markdown-all-in-one",
+    "DavidAnson.vscode-markdownlint",
+    "dbaeumer.vscode-eslint",
+    "ms-vsliveshare.vsliveshare",
```
develohpanda: Personally I don't think we should add vsliveshare, most of the time it's not used (and not at all by open-source), and it's not necessary to dev on this codebase

.vscode/extensions.json
```diff
@@ -0,0 +1,16 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+  // Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+  // List of extensions which should be recommended for users of this workspace.
+  "recommendations": [
+    "yzhang.markdown-all-in-one",
```
develohpanda, line 7: Just had a look at this extension, pretty nifty!

develohpanda (changes requested):
Looks good, just a couple of notes!

.vscode/extensions.json
```diff
@@ -0,0 +1,16 @@
...
+    "DavidAnson.vscode-markdownlint",
+    "dbaeumer.vscode-eslint",
+    "ms-vsliveshare.vsliveshare",
+  ],
```
gatzjames, line 10: Since insomnia uses editorconfig this extension might be useful:

```suggestion
    "editorconfig.editorconfig"
  ],
```

.vscode/extensions.json
dimitropoulos: I _ALMOST_ did that, but it seems like a first step would be to actually lint the markdown.  This PR is just about sorta getting the wheels greased and at least surfacing these errors to our all-vscode-using team (who, until now, have evidently not been using or following a markdown linter).

.vscode/extensions.json
dimitropoulos: ok, I see what you're saying.  removed in 88f439224033e50d2dc1f5a4b50d3f5d45866dc5

.vscode/extensions.json
dimitropoulos: I didn't realize this was here: see https://github.com/Kong/insomnia/pull/3933

.vscode/extensions.json
dimitropoulos: From there, next steps might be to actually do a linting pass and then to add markdownlint-cli.

develohpanda: In that case, I would suggest extracting markdown linting entirely from this PR and create a new one titled "introducing markdown linting", which:
* adds the lint rules
* adds the recommended MD extensions
* adds to the npm script
* does a linting pass

This way, we won't be in an interim state where as we work in markdown files, linting fixes will come trickle through one by one, alongside actual documentation/README changes.

Is there an auto-fix for the rules introduced or is it manual? Linting and formatting is great but when first being introduced we should ideally do a standalone pass to bring it in line, with no feature work - similar to the array-type or import order lint rules. The trickle approach seems fine if it requires manual fixes, but if we have an auto-fix I vote to do it when first introducing markdownlint.

dimitropoulos: I didn't do that, like I said above, because I want to give team members a chance to see what the linting is like and give us a chance to form an opinion.

Consider that for many developers (again, this is one of the most popular of all vscode extensions) there's no "interim state" after this PR any more than there was before.  It's just a change about recommended extensions - for me (for example) the extension has been installed all along and I've been seeing these warnings for quite some time now anyway.

Still.. I'll timebox it to 30 minutes and see if I can just do it all in one gulp like you're asking for.  Not sure about how many of them have autofixers but I know some do (definitely not all).

dimitropoulos: (did in e714811c33bdfc12a07930393d58daf3b0309bb7)

packages/insomnia-importers/README.md
```diff
@@ -2,27 +2,30 @@
...
+
+- Insomnia v1, v2, v3, v4
 - Postman v2.0
-- cURL
+- Postman Environment
 - HTTP Archive Format 1.2 (HAR)
+- cURL
 - Swagger 2.0
 - OpenAPI 3.0
+- WSDL
```
dimitropoulos, line 17: not linting, but noticed this was out of date.  it's pretty minor so seemed fine.

packages/insomnia-importers/README.md
```diff
@@ -34,8 +37,8 @@ insomnia-import /path/to/har-export.json > insomnia-export.json
 
 ## Programmatic Usage
 
-```javascript
-const importers = require('insomnia-importers');
+```ts
+import { importers } from 'insomnia-importers';
```
dimitropoulos, line 42: this was the only remaining javascript markdown codefence.  it's already using async so seemed like switching to ts was appropriate since I was looking over all codefence language specifications.

.markdownlint.yaml
```diff
@@ -0,0 +1,27 @@
...
+extends: null
+
+# MD004 - Unordered list style
+# We want to use dashes consistently everywhere
+MD004:
+  style: dash
+
+# MD013 - Line length
+# We do not want to artificially limit line lengh in a language like markdown because of the nature of the language's handling of whitespace.  Line wrapping is just a fact of life in markdown and trying to impose the conventions of other languages on markdown is an impedance mismatch that leads to people thinking that whitespace in markdown is more representative of the rendered HTML than it really is.
+MD013: false
```
develohpanda, line 16: 🔥 I always found this frustrating

DEVELOPMENT.md
```diff
@@ -11,32 +11,26 @@ to provide access to operating system features.
...
 - [`React`](https://reactjs.org/) is the library used for all UI components.
-- [`styled-components`](https://styled-components.com/) and [`Less`](http://lesscss.org/) are used
-  for styling UI components.
-- [`Electron Builder`](https://github.com/electron-userland/electron-builder) is used to help build,
-  sign, and package Insomnia for distribution.
-- [`Flow`](https://flow.org/) is used for adding types to the codebase. Not everything is Flow but
-  all new code should be typed with Flow.
+- [`styled-components`](https://styled-components.com/) and [`Less`](http://lesscss.org/) are used for styling UI components.
+- [`Electron Builder`](https://github.com/electron-userland/electron-builder) is used to help build, sign, and package Insomnia for distribution.
+- [`Flow`](https://flow.org/) is used for adding types to the codebase. Not everything is Flow but all new code should be typed with Flow.
```
develohpanda, line 16: Heh 😅

package.json
```diff
@@ -13,6 +13,7 @@
   "scripts": {
     "build": "lerna run build --parallel",
     "lint": "lerna run lint --stream --no-bail",
```
develohpanda, line 15: Should this all-encompasing lint command be updated to? 

```suggestion
    "lint": "npm run lint:markdown && lerna run lint --stream --no-bail",
```

packages/insomnia-importers/README.md
```diff
@@ -2,27 +2,30 @@
...
-Insomnia v3 format.
+This repository contains converters to translate popular HTTP data formats to the latest Insomnia format.
 
-- Insomnia v1 and v2
+Supported import types include:
+
+- Insomnia v1, v2, v3, v4
 - Postman v2.0
-- cURL
+- Postman Environment
```
develohpanda, line 11: What is Postman Environment? I saw it in #3935 too and have never come across it before. It might be an internal importer of some sort?

packages/insomnia-inso/README.md
```diff
@@ -13,39 +14,49 @@
...
-  + [ `inso run test` ](#-inso-run-test-identifier)
-  + [ `inso export spec` ](#-inso-export-spec-identifier)
-  + [ `inso script` ](#-inso-script-name)
-* [Configuration](#configuration)
-* [Git Bash](#git-bash)
-* [Continuous Integration](#continuous-integration)
-* [Development](#development)
-
-# Data source
+- [Table of Contents](#table-of-contents)
```
develohpanda: I remember giving this feedback in another recent PR but can't seem to find it again...

Having the table of contents inside the table of contents is a bit messed up. Additionally, the entire table is too verbose now. Is there a way to exclude certain headings (like H3 or H4) from the auto-generation? For example, the examples under each command don't need to be in the table of contents.

The text for each command has been changed, but that is understandable because it is using the actual heading text now. I think that's OK; can we drop the opening `$` in the actual headings so it doesn't come through to the TOC? That was mostly for stylistic purposes.

I suggest, _in this PR_, we either:
1. disable the auto-generation for now and revert this if it was done with the `markdown-all-in-one` extension
2. update the actual headers so that the TOC looks like it did before after an auto-generator passes through it and figure out how to exclude certain sections from the TOC

### Develop
image

### This branch
image

packages/insomnia-inso/README.md
```diff
@@ -73,70 +84,71 @@ Similar to the Kong [Kubernetes](https://insomnia.rest/plugins/insomnia-plugin-k
 | `--output <path>` | `-o` |save the generated config to a file in the working directory|
 | `--tags <tags>` | |comma separated list of tags to apply to each entity|
 
-### Examples
+#### Generate Config Examples
 
-When running in the <a href="https://github.com/Kong/insomnia/tree/develop/packages/insomnia-inso/src/db/fixtures/git-repo">git-repo</a> directory
```
develohpanda, line 78: Funky, I'm not sure how an anchor ended up _here_

packages/insomnia-inso/README.md
```diff
@@ -73,70 +84,71 @@ Similar to the Kong [Kubernetes](https://insomnia.rest/plugins/insomnia-plugin-k
...
 
-## `$ inso lint spec [identifier]`
+### `$ inso lint spec [identifier]`
 
 Insomnia has the ability to lint and validate your OpenAPI specification as you write it. This command adds the same functionality to `inso` , in order to run linting during CI workflows. Lint results will be printed to the console, and `inso` will exit with an appropriate exit code.
 
 **`[identifier]`**: this can be a **document name, or id**.
 
-### Examples
+#### Lint Spec Examples
```
develohpanda, line 130: Why not keep this as `Examples`, given it is underneath the "lint spec" command header?

.vscode/extensions.json
develohpanda: > there's no "interim state" after this PR any more than there was before...for me (for example) the extension has been installed all along and I've been seeing these warnings for quite some time now anyway.

The interim state is that we now formally introduce something to lint (an extension with config) but don't fix what is fixable (most of the things in this PR seemed pretty straightforward and the linter appears to be happy). The fact that it was installed _only_ for you and raising warnings already isn't as big of a concern because it wasn't a "this project uses this config", it was "my personal workspace expects this config".

I also fully understand that we have added ES Lint rules that raise warnings (the need for Function components where possible, for example) which are currently in an interim state. There are changes that need to be tackled in smaller chunks (like FC over class components) and there are changes that should ideally go in as a sweep (this PR, majority of eslint rules - eg. import sorting, array types, etc).

--- 

We are no longer in an interim state with the updated PR, so I am happy to mark this conversation/concern as resolved.

develohpanda (changes requested):
Everything looks good, I only have one note (which I think I've made before as well) about the table of contents of the inso readme.

.markdownlint.yaml
dimitropoulos: I'm very glad to hear you say that.  SAME!

DEVELOPMENT.md
dimitropoulos: I saw this, but resisted the urge to change now due to scope creep.  yep.

packages/insomnia-inso/README.md
dimitropoulos: There's a rule, MD024, that requires not having duplicate headers, that's why this changed.

Screenshot_20210825_092117

packages/insomnia-inso/README.md
dimitropoulos: yeah, it is strange, and it was copy-pasted (it seems) so the same exact thing was in a few places.  oh well, haha.

packages/insomnia-inso/README.md
dimitropoulos: I tried https://github.com/dimitropoulos/insomnia/blob/aa8540277e6e3d2e7a97006d8f5976768c344067/packages/insomnia-inso/README.md as an example (basically wrapping the example sections in code blocks) but it just looked too strange.

I finally consulted the docs and was able to find exactly what you're looking for, I think.  Check out the commit description on https://github.com/dimitropoulos/insomnia/commit/d84ad844b6e980638966bd727fa22f69dd80bb54#diff-d97cd93da196fffe22c9d3cdd00d273f31502fc7a6721ef9cdc04e234de4c028 and lemme know what you think.

packages/insomnia-inso/README.md
dimitropoulos: But just to say it out loud, I think the main reason to follow this rule is that if you don't follow it you break (or make annoying) linking ability.  Say someone wanted to create a link to one of these examples it would create a link like https://github.com/Kong/insomnia/blob/develop/packages/insomnia-inso/README.md#examples-2 instead of https://github.com/Kong/insomnia/blob/develop/packages/insomnia-inso/README.md#lint-spec-examples which is an improvement.

packages/insomnia-inso/README.md
dimitropoulos: (please feel free, like always, to unresolve if you think there's more to be done here but it seems like the omit is exactly what you were looking for.  and, hey, by the way, thanks for pushing the topic because it's cool to be aware of.  I don't share your dissatisfaction with the TOC being larger, but I totally can understand how you might see it that way and I appreciate you bringing it up.)

packages/insomnia-importers/README.md
dimitropoulos: check out https://github.com/Kong/insomnia/blob/develop/packages/insomnia-importers/src/importers/postman-env.ts.  Basically we can _separately import a postman environment (see: https://www.kenst.com/2021/04/how-to-export-environments-from-postman/) which is different form the postman schema (see: https://schema.postman.com/).

You can see some examples in our test fixtures here: https://github.com/Kong/insomnia/tree/develop/packages/insomnia-importers/src/importers/fixtures/postman-env

develohpanda: Interesting - TIL

packages/insomnia-inso/README.md
develohpanda: Hmmm, yeah, the anchors being unique is a fair point. I think using the `omit in toc` option is a good enough compromise for de-duplicating these sections, so marking this as resolved.

packages/insomnia-inso/README.md
develohpanda: TY for digging into it! I think the dissatisfaction only arises from it becoming noisy - there's a lot of content and nested content but not everything belongs in the TOC. What you've done is great, I've got a separate question but I'll start a new thread for it

packages/insomnia-inso/README.md
```diff
@@ -13,39 +14,43 @@
   <img src="https://raw.githubusercontent.com/Kong/insomnia/develop/packages/insomnia-inso/assets/demo.gif" alt=""/>
   <br />
 </div>
+<!-- markdownlint-enable no-inline-html  -->
 
 Table of Contents
 =================
```
develohpanda: What are your thoughts on removing the TOC header entirely? It shows up in the actual table of contents as a top level entry, pushing everything down one level in hierarchy. This doesn't look great and isn't correct either.

image

develohpanda (commented):
One last note; rest LGTM. Thanks for working through it 🙌🏽

packages/insomnia-inso/README.md
dimitropoulos: I love this suggestion, good point!  check out 7400464d660296df3ce0a399493485f150a020c8.

I looked at https://github.com/facebook/jest/blob/master/README.md as an example.

Before:
Screenshot_20210826_102910

<br />
<br />
<br />
<br />
<br />

After:
Screenshot_20210826_102942

In retrospect, it's kinda funny that the (one and only) h1 for this npm package was "Table of Contents".

dimitropoulos: (I tried it without any `Table of Contents` header whatsoever, as well, but it felt strange/bare next to the image, and required a few lines of spacing, too, whereas with the h2 it comes by default.)

develohpanda: What you've got now looks awesome ☺️ h1 feels so in-yo-face haha

develohpanda (approved):

PR #3933, opened 2021-08-23
removes .editorconfig file

dimitropoulos:
I didn't realize there was an editorconfig file in this project.  There's no need for one.  The aim is for us to be leveraging eslint (or other specialized linting like markdownlint) for all of these things.

Other reasons to remove this file include:
- It means avoiding inconsistencies with the eslint tooling.
  - For example, eslint can handle windows `CRLF`, but the editorconfig requires `LF` line endings.  This just plain wrong in windows environments (despite how much I wish it wasn't.
  - For another example, the quote_type `single` is not flexible enough to handle the cases that the corresponding lint rule handles.  For example, we may use double quotes in strings with a `'` to avoid escape characters.  The eslint rule already covers this.
- It means we won't be enforcing rules that are not run in the CI (which, is always the goal).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,0 @@
-[*.{js,json,html,md,graphql,gql}]
-charset = utf-8
-indent_style = space
-indent_size = 2
-insert_final_newline = true
-trim_trailing_whitespace = true
-end_of_line = lf
-quote_type = single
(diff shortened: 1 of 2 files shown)


develohpanda (approved):
Addressing all of the file types covered by `editorconfig` right now:
* we have no graphql/gql files
* JSON files seem to format correctly even after getting rid of this
* JS files are being removed in favor of TS and enforced by ES Lint
* md will use markdownlint
* we have very few and rare touched plain html files

Yay for removing conflicting tooling!

PR #3935, opened 2021-08-23
removes hardcoded import names

dimitropoulos:
I noticed while QA'ing for the next release that these names are hardcoded.  I see no reason that doing so benefits us - and it's a 1-line change to update it.

before:
before

after:
image

### DIFF
--- a/packages/insomnia-app/app/ui/components/settings/import-export.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/import-export.tsx
@@ -66,7 +66,7 @@ export const ImportExport: FC<Props> = ({ hideSettingsModal }) => {
       <div className="no-margin-top">
         Import format will be automatically detected.
         <HelpTooltip className="space-left">
-          supported formats include: {importers.map(importer => importer.name).join(', ')}
+          Supported formats include: {importers.map(importer => importer.name).join(', ')}
         </HelpTooltip>
       </div>
       <p>
--- a/packages/insomnia-app/app/ui/components/settings/import-export.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/import-export.tsx
@@ -70,7 +70,7 @@ export const ImportExport: FC<Props> = ({ hideSettingsModal }) => {
         </HelpTooltip>
       </div>
       <p>
-        Don't see your format here? <Link href={docsImportExport}>Add Your Own</Link>.
+        Your format isn't supported? <Link href={docsImportExport}>Add Your Own</Link>.
       </p>
       <div className="pad-top">
         <Dropdown outline>
--- a/packages/insomnia-app/app/ui/components/settings/import-export.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/import-export.tsx
@@ -11,6 +11,7 @@ import { importClipBoard, importFile, importUri } from '../../redux/modules/impo
 import { selectActiveProjectName, selectActiveWorkspace } from '../../redux/selectors';
 import { Dropdown, DropdownButton, DropdownDivider, DropdownItem } from '../base/dropdown';
 import Link from '../base/link';
+import HelpTooltip from '../help-tooltip';
 import ExportRequestsModal from '../modals/export-requests-modal';
 import { showModal, showPrompt } from '../modals/index';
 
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { FC, useCallback } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 
+import { importers } from '../../../../../insomnia-importers/dist/src';
```
develohpanda: I don't think it's a good idea to import from `/dist`, might have been an oversight. I think it's totally reasonable to have a top-level export from `insomnia-importers` containing the supported formats/names and using that instead

develohpanda (changes requested):

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: TOTALLY a mistake.  Dang autoimporter getting ambitious.  I wonder if there's a way to tell it to never do this?

8da3c45082a6c83515eabadd2a59dcc90e627444

I double checked the line below, but didn't think about this one.

dimitropoulos: (good catch, thanks!)

develohpanda: Oh nice, it was already a top-level export!

> I wonder if there's a way to tell it to never do this?

[`no-relative-packages`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-relative-packages.md) seems very useful for us, it comes as part of `eslint-plugin-import`.

develohpanda (dismissed):

develohpanda:
> image


@wdawson should we display the supported formats in this much detail? For instance, we don't need to show older Insomnia formats. People don't usually know which format file they have, and should safely assume that all Insomnia export versions can be imported. And the Postman ones are now separated into "Postman" and "Postman Environment", and I'm not sure what the difference is between the two.

IMO we don't need to show the level of detail of supported formats; Insomnia will error if it's not able to import. Or we could de-clutter the tab and show the full (and potentially growing) list in a tool-tip, for instance.

develohpanda:
(@dimitropoulos I initially approved, and then disabled auto-merge because I wanted Wils's opinion. Then figured I'll dismiss my review because I have an outstanding question, so I re-enabled auto-merge again like you had done so initially 😅)

dimitropoulos:
Sure, it's no rush

wdawson:
I definitely wouldn't bold the list. I like the tooltip suggestion. People need some place to see what's supported (the full list). That could live in the docs eventually if things get too out of hand. But tooltip would work nicely for now. "Supported import formats include: x, y, z"

develohpanda (changes requested):
> I definitely wouldn't bold the list. I like the tooltip suggestion. People need some place to see what's supported (the full list). That could live in the docs eventually if things get too out of hand. But tooltip would work nicely for now. "Supported import formats include: x, y, z"

Let's go with this approach

dimitropoulos:
ok, I updated:
image

wdawson (approved):
⛵ ship it! Thanks!

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -61,12 +63,14 @@ export const ImportExport: FC<Props> = ({ hideSettingsModal }) => {
...
   return (
     <div>
-      <p className="no-margin-top">
-        Import format will be automatically detected (
-        <strong>Insomnia, Postman v2, HAR, Curl, Swagger, OpenAPI v3</strong>)
-      </p>
+      <div className="no-margin-top">
+        Import format will be automatically detected.
+        <HelpTooltip className="space-left">
+          supported formats include: {importers.map(importer => importer.name).join(', ')}
```
develohpanda:
```suggestion
          Supported formats include: {importers.map(importer => importer.name).join(', ')}
```

develohpanda (approved):

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: you know, haha, I did this at first and then changed it before committing because it isn't a sentence.  despite that being technically correct I didn't think to check what the app does in this situation anyway, and it seems very much the norm to cap the first letter - sentence or not.  good catch.

PR #3936, opened 2021-08-23
Inso -> Inso CLI (in docs)

dimitropoulos:
In preparation for work on the new doc site, we decided that it's a clearer name to call this "Inso CLI" in the docs (and urls).  `Inso` is hard to search for, among other reasons.

This PR is just a supporting change to that, updating some markdown files to be consistent.

Made in separate commits so that if we want to drop any one of the commits it's easy to do.


reynolek (approved):
LGTM

develohpanda (approved):

PR #3939, opened 2021-08-24
renames base project to default project in code and copy

dimitropoulos:
During standup today we discussed renaming the "base project" to "default project" to be more clear to users what it is and what it does.  This PR makes that change.  Individual commit notes are in the commit descriptions.

The main places this changes for users are in the tooltip over the icon for the default space.

before:
Screenshot_20210824_171348

after:
Screenshot_20210824_171248

### DIFF
--- a/packages/insomnia-app/app/models/project.ts
+++ b/packages/insomnia-app/app/models/project.ts
@@ -9,13 +9,13 @@ export const prefix = 'proj';
 export const canDuplicate = false;
 export const canSync = false;
 
-export const BASE_PROJECT_ID = `${prefix}_base-project`;
+export const DEFAULT_PROJECT_ID = `${prefix}_base-project`;
 
-export const isBaseProject = (project: Project) => project._id === BASE_PROJECT_ID;
-export const isNotBaseProject = (project: Project) => !isBaseProject(project);
+export const isDefaultProject = (project: Project) => project._id === DEFAULT_PROJECT_ID;
+export const isNotDefaultProject = (project: Project) => !isDefaultProject(project);
 export const isLocalProject = (project: Project): project is LocalProject => project.remoteId === null;
 export const isRemoteProject = (project: Project): project is RemoteProject => !isLocalProject(project);
-export const projectHasSettings = (project: Project) => !isBaseProject(project);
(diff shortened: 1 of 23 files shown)


packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.tsx
```diff
@@ -158,7 +159,7 @@ class ResponseHistoryDropdown extends PureComponent<Props> {
       activeEnvironment,
       ...extraProps
     } = this.props;
-    const environmentName = activeEnvironment ? activeEnvironment.name : 'Base';
+    const environmentName = activeEnvironment ? activeEnvironment.name : strings.defaultProject.singular;
```
dimitropoulos: so I'm doing a self review and this suddenly seems wrong.  can anyone confirm?

reynolek: This seems to be changing the environment name, so thats probably wrong.

develohpanda: Good catch, this shouldn't change

packages/insomnia-app/app/models/project.ts
```diff
@@ -9,13 +9,13 @@ export const prefix = 'proj';
 export const canDuplicate = false;
 export const canSync = false;
 
-export const BASE_PROJECT_ID = `${prefix}_base-project`;
+export const DEFAULT_PROJECT_ID = `${prefix}_base-project`;
```
develohpanda:
```suggestion
export const DEFAULT_PROJECT_ID = `${prefix}_default-project`;
```

packages/insomnia-app/app/ui/components/dropdowns/response-history-dropdown.tsx
develohpanda: Reverted in [`6015e83` (#3939)](https://github.com/Kong/insomnia/pull/3939/commits/6015e83ec33ef805560b68f9570af78ff589a8b5)

develohpanda (approved):
LGTM

image

image

PR #3949, opened 2021-08-26
🧹 Remove axios, use generics for send-request in insomnia-testing

develohpanda:
This has been a long time coming, but `insomnia-testing` does not actually use axios and the fact that the code paths and tests exist tend to add a lot of confusion and means we're not testing the right stuff.

Axios was initially used to send requests used by unit testing. But since that initial iteration, insomnia's networking library is exposed via `insomnia-send-request`, and that callback is sent to `insomnia-testing` via `runTests` or `runTestsCli`.

The option to send that callback to `insomnia-testing` was previously optional but is now required - there is no business-logic usage of `insomnia-testing` that _doesn't_ send `sendRequest` as an option, and making it required allows us to safely remove axios.

QA steps would be to make sure unit tests that hit a network work properly and the results (pass/fail) are shown correctly both with the app and with inso.


packages/insomnia-testing/src/run/insomnia.ts
```diff
@@ -1,38 +1,7 @@
...
-export interface Response {
-  status: number;
-  statusText: string;
-  data: {
-    [key: string]: unknown;
-  };
-  headers: {
-    [key: string]: string;
-  };
-}
```
dimitropoulos, line 29: I'm so so happy to be rid of these.  They always freak me out because they are duplicates in the intellisense menu and I worry that I might use the wrong one accidentally.

packages/insomnia-testing/src/run/run.ts
```diff
@@ -5,25 +5,26 @@ import Mocha, { Reporter, ReporterConstructor } from 'mocha';
...
+import { TestResults } from './entities';
+import { Insomnia, InsomniaOptions } from './insomnia';
 import { JavaScriptReporter } from './javascript-reporter';
 
 declare global {
   namespace NodeJS {
     interface Global {
-      insomnia?: Insomnia;
+      // Not sure how to type the generic type here
+      insomnia?: Insomnia<{}>;
```
dimitropoulos: how about:
```ts
export class Insomnia<TResponse = {}> {
```

packages/insomnia-testing/src/run/run.ts
```diff
@@ -5,25 +5,26 @@ import Mocha, { Reporter, ReporterConstructor } from 'mocha';
...
+const runInternal = async <TReturn, TNetworkResponse>(
   testSrc: string | string[],
-  options: InsomniaOptions,
+  options: InsomniaOptions<TNetworkResponse>,
   reporter: Reporter | ReporterConstructor = 'spec',
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- type not available, and postponing anyway until the impending move to all jest (and no mocha)
-  extractResult: (runner: { [key: string]: any }) => T,
-): Promise<T> => new Promise((resolve, reject) => {
+  extractResult: (runner: Mocha.Runner) => TReturn,
+): Promise<TReturn> => new Promise((resolve, reject) => {
```
dimitropoulos:
```suggestion
) => new Promise<TReturn>((resolve, reject) => {
```

dimitropoulos (approved):
Screenshot_20210826_100728

I'm very glad you did this because I'm sure I would have never (or, almost never) realized this on my own.

packages/insomnia-testing/src/run/run.ts
develohpanda: 🤯 that's a thing?!!?!?!?!?!

PR #3953, opened 2021-08-26
implements sort order for projects

dimitropoulos:
This PR does two things:
1. it sorts projects anywhere they are presented app wide such that the default project is always first, then local projects, then remote projects, and all are sorted by name within their group.
1. ~it changes the position and copy of the projects dropdown divider.  It used to read `All Projects` and be under the default project, and now it reads `Projects` and is at the top.~

| Before | After |
| - | - |
|  |  |
|   |  |


packages/insomnia-app/app/models/__tests__/project.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    const sortedProjects = [
+      { type: 'default', name: 'a', remoteId: null, id: DEFAULT_PROJECT_ID },
+      { type: 'local', name: 'a', remoteId: null },
+      { type: 'local', name: 'b', remoteId: null },
+      { type: 'remote', name: '0', remoteId: 'notNull' },
+      { type: 'remote', name: 'a', remoteId: 'notNull' },
+      { type: 'remote', name: 'b', remoteId: 'notNull' },
+    ] as Project[];
+    expect(result).toEqual(sortedProjects);
```
develohpanda: To improve readability and to help with debugging I suggest writing these as the following:

```js
const defaultProject = {};
const remoteA = {};
const localA = {};
const remoteB = {};
const LocalA = {};

const unsortedProjects = [
	remoteA,
	localB,
	defaultProject,
	remoteB,
	localA,
];

const sortedProjects = [
	defaultProject,
	localA,
	localB,
	remoteA,
	remoteB,
];
```

This is because between the sorted and unsorted arrays, the object never actually changes, so it would be much easier to grok if we work with alias to show what's _actually_ changing (the ordering).

You could also use fluent-builder to create these projects without the need for casting `as Project`. You can use [this](https://github.com/Kong/insomnia/blob/7abde2a01700f587179941b3231fb1078fcb1e41/packages/insomnia-app/app/models/__schemas__/model-schemas.ts#L41-L45) as an example to create a `Schema<Project>`.

packages/insomnia-app/app/models/project.ts
```diff
@@ -17,6 +19,13 @@ export const isLocalProject = (project: Project): project is LocalProject => pro
 export const isRemoteProject = (project: Project): project is RemoteProject => !isLocalProject(project);
 export const projectHasSettings = (project: Project) => !isDefaultProject(project);
 
+export const sortProjects = sortWith<Project>([
+  descend(isDefaultProject),
+  descend(isLocalProject),
+  descend(isRemoteProject),
+  ascend(prop('name')),
+]);
+
```
develohpanda: This is view related so I don't think it should exist in the actual `project` model. Is there a place to put ramda conditions like this, similar to how we created a place for react hooks?

Perhaps, in this case, inside the redux selector itself? The test you wrote could be used to write a test for the selector as well, because we should ensure that `selectProjects` always returns projects in the expected order.

packages/insomnia-app/app/models/__tests__/project.test.ts
```diff
@@ -0,0 +1,25 @@
+import { DEFAULT_PROJECT_ID, Project, sortProjects } from '../project';
+
+describe('sortProjects', () => {
+  it('returns the correct sort order', () => {
```
develohpanda:
```suggestion
  it('sorts projects by default > local > remote > name', () => {
```

packages/insomnia-app/app/ui/components/dropdowns/project-dropdown.tsx
```diff
@@ -98,11 +99,13 @@ export const ProjectDropdown: FC<Props> = ({ vcs }) => {
     />
   ), [setActive, activeProject._id]);
 
+  const [defaultProject, userProjects] = partition(isDefaultProject, projects);
```
develohpanda: 🔥

packages/insomnia-app/app/ui/components/dropdowns/project-dropdown.tsx
```diff
@@ -98,11 +99,13 @@ export const ProjectDropdown: FC<Props> = ({ vcs }) => {
...
+
   return (
     <Dropdown renderButton={button} onOpen={refresh}>
-      {projects.filter(isDefaultProject).map(renderProject)}
+      {defaultProject.map(renderProject)}
```
develohpanda: Naming is hard 😓 even though we know this should always return a single item, because we are doing a map, `defaultProject` is actually an array and it would be clearer to do `defaultProjects`. I was misled by this when looking just now, assuming `defaultProject` was singular while `userProjects` was an array.

develohpanda (changes requested):
Looks like ramda isn't actually needed by inso, left-over code from before?

Overall LGTM, couple of items of feedback.

I am SO EXCITED to finally be able to use ramda. Looking forward to it!

packages/insomnia-app/app/ui/components/dropdowns/project-dropdown.tsx
dimitropoulos: you know what's hilarious.... haha.  so originally I had it as `defaultProjects` (out of instinct) but I thought others might find that annoying so I changed it to singular.  I'm glad, though, that you prefer it this way because I do too!

dimitropoulos: 6bd3d8cb50dfbe7a4c13e62916d69db6c6577859

packages/insomnia-app/app/models/__tests__/project.test.ts
dimitropoulos: this is perfect.  love it.  I was trying to think of a way to make this easier to read but I wasn't feeling creative enough I guess, haha.  394bd4acae58fd936ce8e9d72918a393d9e0128b

dimitropoulos: Thanks for the suggestion re fluent-builder.  In this case I don't see a need to bring any other tool in because it's so neat and simple, but it's always nice to be thinking of ways to use fluent-builder.

packages/insomnia-app/app/models/project.ts
dimitropoulos: I'd like to better understand your frame of mind on this.  To me, this isn't view related any more than anything else.  It's data management.  Just like if we happened to use a SQL `SORT` on the backend, this (to me) has a lot more to do with data management than anything else.  Sure, you know, the result is that it shows up in the same order in the dropdown, but it's more than just that.  Otherwise, one could argue that almost every function of the application is in some way or another view related.  I know that you're not making that point or anything, but what I'm trying to say is that I don't see where the line gets drawn.

As for putting it in the selectors file, I don't really want to do that because it's not a selector.  It seems most prudent to me to keep it here because it's a utility specific to projects.

How about this: we could go the other direction with it and remove it from the selector and put it in the `all` function, maybe?

```diff
export async function all() {
  const projects = await db.all<Project>(type);

  if (!projects.find(c => c._id === DEFAULT_PROJECT_ID)) {
    await create({ _id: DEFAULT_PROJECT_ID, name: getAppName(), remoteId: null });
-    return db.all<Project>(type);
+    return sortProjects(db.all<Project>(type));
  }

-  return projects;
+  return sortProjects(projects);
}
```

dimitropoulos:
@develohpanda as for ramda in inso, check out the commit message I made in the commit that added it https://github.com/Kong/insomnia/pull/3953/commits/9ae2b123cf3500eee8480240e8d04ced275b3121

It seemed like it wasn't worth spending more time debugging since I know we want to have ramda available there, too.

____


update: I guess I was wrong.  attempted fix b7d43a538de87a1bf39a5bbf4b438023e12d81b1 since I was able to reproduce it locally

packages/insomnia-app/app/models/__tests__/project.test.ts
develohpanda: I guess the end goal I am suggesting is to remove the `as Project` cast.

We're sending either an incomplete object, or forcing TS to accept whatever we are passing it to be a `Project`. That means as the `Project` model changes TS will never inform us that the data being sent here is no longer valid.

Let's say we were to replace the `remoteId` field with something else in the future (which can be  a likely scenario), or we add a new field and add it to the sort criteria. This test will break, but we'll only be informed at runtime and not compile time, sort of defeating the purpose of having TS in the first place, since typechecking should be able to catch that error.

Fluent-builder is just a vessel to achieve that type-safety but my ultimate goal with this feedback is to remove the `as Project[]` cast and catch errors at compile-time rather than test runtime. An alternate approach to fluent-builder could be to use `await models.project.create()` which will also return a valid project, making this test dependent on a database - which is also fine, just slower.

I'm not too concerned about how we solve this (database, fluent-builder, something else) but would like to avoid encouraging casting as a practice especially if we have safer solutions. What do you think?

packages/insomnia-app/app/ui/components/dropdowns/project-dropdown.tsx
develohpanda: I was second-guessing about this feedback because it's "just naming" after all, but am glad I put it down 😄 thanks!

packages/insomnia-app/app/models/project.ts
develohpanda: That's a really good point you raise about sorting in SQL, though I think the context might be slightly different here because we don't gain any performance by sorting in the data layer the way you do with SQL. This is a loose example and I totally understand your POV with sorting in SQL.

I think largely we're on the same page. When I suggested putting it in the selectors file, I don't mean export from that file, I mean consume it in that selector itself so that the result of `selectProjects` is sorted.

Sorting the result of the `await models.projects.all()` also seems reasonable; it will be worth checking how frequently that code-path is hit vs sorting in the redux selector during regular use of the app. We obviously wouldn't want to sort unnecessarily, and the caching within redux selectors should help us there a bit. The sorted result only matters where we are rendering into the UI, and we only render to the UI through redux, so I think we're safe there.

The reason I don't think `sortProjects` should exist as a top-level export in `models/project.ts` is because as a general pattern, the large majority of the functions in `models/*.ts` proxy to `common/database.js`.

Seeing a `models.project.sortProjects()` function available could imply that it might sort and _persist_ in the database (`sortProjects` as a name implies an action rather than a check like `isProject`). In essence, `sortProjects` _is_ a utility, but `models/projects.ts` _is not_ a collection of utilities for the model, it's a proxy to the database.

To that end, that was my thought process for why `models/helpers/is-model.ts` existed, but I can see why it makes sense to separate all of those into the actual models.

I hope that helps to clarify my thought process? There are a few points:
* what a top level export in a `models/*.ts` means
* how frequently we are hitting a sorting code-path during regular use of the app

packages/insomnia-app/app/models/__tests__/project.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const sortedProjects = [
+      defaultProject,
+      localA,
+      localB,
+      remote0,
+      remoteA,
+      remoteB,
+    ] as Project[];
+    expect(result).toEqual(sortedProjects);
```
develohpanda: The changes to this test, even just visually without reading anything, shows what is being tested. 😍

develohpanda (commented):
QA'd this and it looks good to me! Thanks for removing from inso, it looks correct, in that send-request is the one that needs it and not inso.

Though I agree we'll want to use ramda everywhere!!

Haven't approved yet as there is still some discussion above.

packages/insomnia-app/app/models/__tests__/project.test.ts
dimitropoulos: well would you look at that! there was a significant error in the test because of the cast and we missed it.  I removed the cast and it caught it.  I was using `id` instead of `_id` for the default project.  Now, if we ever replace `remoteId` or whatever (I know you were just making an example) TS will catch it.

9c12cf3418f72b024b7ee823e6a3e5339254485c

re Fluent-builder, I'm totally not opposed, just don't quite see how to use it here (need to read the docs more).

dimitropoulos: re: 
> I guess the end goal I am suggesting is to remove the as Project cast.

since I removed the cast, I'm gonna resolve this conversation - please lemme know if I missed something!

develohpanda: LGTM!

packages/insomnia-app/app/models/project.ts
dimitropoulos: > When I suggested putting it in the selectors file, I don't mean export from that file

Ahh, thanks for clarifying, this is exactly what I (wrongly) understood from what you had said.

> is because as a general pattern, the large majority of the functions in models/*.ts proxy to common/database.js

This is a great point.  It's the result of another problem which is that the interface for models is not well defined.  It's just "whatever the file exports", which we have no way to enforce.

> Seeing a models.project.sortProjects() function available could imply that it might sort and persist in the database

I had also not yet thought about it this way.  I completely agree with you, now.  Given that the interface for models is so shaky, we need to make this change.  However, I don't think a refactor of the type we would need to solve your feedback is in scope for this project.  There are already other functions here that have exactly the same problem (e.g. `isDefaultProject`, `isNotDefaultProject`, `isLocalProject`, `isRemoteProject`, `projectHasSettings`) regarding the model interface being the module's exports.

so, for now, I moved it to `helpers/project.ts` in 20f08000f24277ffae859ca698ea5dbefecb16f9

note, though, that this will restrict us in the future, because if the project model _itself_ wants to use something from the helpers then we'll have a circular dependency (as it now stands).  for example if we were to put the sorting at the `all` method like we were talking.

develohpanda: All sounds good to me, thanks for asking for clarification :)

packages/insomnia-app/app/models/helpers/project.ts
```diff
@@ -0,0 +1,10 @@
+import { ascend, descend, prop, sortWith } from 'ramda';
+
+import { isDefaultProject, isLocalProject, isRemoteProject, Project } from '../project';
+
+export const sortProjects = <T extends Pick<Project, 'name' | 'remoteId' | '_id'>>(projects: T[]) => sortWith<T>([
```
develohpanda, line 5: 🚀 generics ftw

develohpanda (approved):
LGTM!

packages/insomnia-app/app/models/helpers/project.ts
dimitropoulos: haha :)

PR #3957, opened 2021-08-26
Minor Electon-related cleanup

dimitropoulos:
I noticed a few little dangling things that were relevant to the upgrade to electron v11 we just did.  They're all super small so I thought it'd cool if they were each separated to their own commit, but still in the same PR.  Read each commit individually for best review experience.


packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -1,4 +1,4 @@
-import electron, { BrowserWindow } from 'electron';
+import electron, { BrowserWindow, MenuItemConstructorOptions } from 'electron';
```
gatzjames, line 1: Good to see electron types getting better even if we need to patch them! 😁

packages/insomnia-app/app/ui/components/viewers/response-web-view.tsx
```diff
@@ -13,73 +13,73 @@ interface Props {
...
+  webview = createRef<HTMLWebViewElement>();
 
-  _handleSetWebViewRef(n: HTMLElement) {
-    this._webview = n;
-
-    if (this._webview) {
-      this._webview.addEventListener('dom-ready', this._handleDOMReady);
+  componentDidMount() {
+    if (!this.webview.current) {
+      // This is not supposed to be possible in react since the webview is not conditionally rendered (i.e. it's always rendered), but putting an error here just in case.
```
gatzjames, line 21: Agree but good to have it here in case someone changes this in the future.

gatzjames (approved):
Nice one 👍 🧹

PR #3961, opened 2021-08-27
fixes "Hide menu bar" setting

dimitropoulos:
Also, while I was there, I removed the requirement to restart the application.  Now you can change this value and immediately see the change take place - no restart required.

closes INS-908

### DIFF
--- a/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
@@ -12,13 +12,11 @@ export const BooleanSetting: FC<{
   setting: keyof BaseSettings;
   help?: string;
   forceRestart?: boolean;
-  callback?: () => void;
 }> = ({
   label,
   setting,
   help,
   forceRestart,
-  callback,
 }) => {
   const settings = useSelector(selectSettings);
 
@@ -31,8 +29,7 @@ export const BooleanSetting: FC<{
     await models.settings.patch({
       [setting]: checked,
     });
-    callback?.();
-  }, [callback, setting]);
+  }, [setting]);
 
   return (
     <div className="form-control form-control--thin">
--- a/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
@@ -37,7 +37,7 @@ export const BooleanSetting: FC<{
         {label}
         {help && <HelpTooltip className="space-left">{help}</HelpTooltip>}
         {forceRestart && (
-          <Tooltip message="Will restart app" className="space-left">
+          <Tooltip message="Will restart the app" className="space-left">
             <i className="fa fa-refresh super-duper-faint" />
           </Tooltip>
         )}
--- a/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
@@ -28,11 +28,11 @@ export const BooleanSetting: FC<{
 
   const onChange = useCallback(async (event: ChangeEvent<HTMLInputElement>) => {
     const { checked } = event.currentTarget;
-    await models.settings.update(settings, {
+    await models.settings.patch({
       [setting]: checked,
     });
     callback?.();
-  }, [callback, setting, settings]);
+  }, [callback, setting]);
 
   return (
     <div className="form-control form-control--thin">
--- a/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
@@ -2,13 +2,14 @@ import React, { ChangeEvent, FC, useCallback } from 'react';
 import { useSelector } from 'react-redux';
 
 import * as models from '../../../models/index';
+import { BaseSettings } from '../../../models/settings';
 import { selectSettings } from '../../redux/selectors';
 import HelpTooltip from '../help-tooltip';
 import Tooltip from '../tooltip';
 
 export const BooleanSetting: FC<{
   label: string;
-  setting: string;
+  setting: keyof BaseSettings;
   help?: string;
   forceRestart?: boolean;
   callback?: () => void;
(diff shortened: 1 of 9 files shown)


packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -217,17 +188,34 @@ class General extends PureComponent<Props, State> {
...
+              label="Reveal passwords"
+              setting="showPasswords"
+            />
+            {!isMac() && (
+              <BooleanSetting
+                label="Hide menu bar"
+                setting="autoHideMenuBar"
+                callback={() => {
+                  setMenuBarVisibility(settings.autoHideMenuBar);
+                }}
```
develohpanda: I reckon instead of this being an explicit callback here, it can be an effect powered by `settings.autoHideMenuBar` in `app-hooks.tsx`

```ts
export const useMenuBarVisibility = () => {
  const { autoHideMenuBar } = useSelector(selectSettings);

  useEffect(() => {
    setMenuBarVisibility(autoHideMenuBar);
  }, [autoHideMenuBar]);
}
```

Otherwise, it might be better to send the _updated_ settings object into the `callback()`, though I think an effect bound to the setting is a nicer approach.

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { selectSettings } from '../../redux/selectors';
+import HelpTooltip from '../help-tooltip';
+import Tooltip from '../tooltip';
+
+export const BooleanSetting: FC<{
+  label: string;
+  setting: string;
+  help?: string;
+  forceRestart?: boolean;
+  callback?: () => void;
```
develohpanda: Following a previous comment, I suggest we remove the callback here and create an effect at a higher level that depends on the setting value

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useSelector } from 'react-redux';
+
+import * as models from '../../../models/index';
+import { selectSettings } from '../../redux/selectors';
+import HelpTooltip from '../help-tooltip';
+import Tooltip from '../tooltip';
+
+export const BooleanSetting: FC<{
+  label: string;
+  setting: string;
```
develohpanda:
```suggestion
  setting: keyof Settings;
```

(If the current implementation is correct, there should be no errors raised by making this typing change. If an error is raised, I don't suggest fixing in this PR)

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
```diff
@@ -0,0 +1,55 @@
...
+    throw new Error(`Invalid boolean setting name ${setting}`);
+  }
+
+  const onChange = useCallback(async (event: ChangeEvent<HTMLInputElement>) => {
+    const { checked } = event.currentTarget;
+    await models.settings.update(settings, {
+      [setting]: checked,
+    });
+    callback?.();
+  }, [callback, setting, settings]);
```
develohpanda: This change will prevent the callbacks for all `BooleanSetting` instances changing because of a different setting. This is the same approach we used in the [themes](https://github.com/Kong/insomnia/blob/251287062d2f9e345c88041006ee380b1275faee/packages/insomnia-app/app/ui/hooks/theme.ts#L49), for the same reason.

```suggestion
  const onChange = useCallback(async (event: ChangeEvent<HTMLInputElement>) => {
    const { checked } = event.currentTarget;
    await models.settings.patch({
      [setting]: checked,
    });
    callback?.();
  }, [callback, setting]);
```

develohpanda (changes requested):
Couple of suggestions. Can you also please include a gif in the description? I only have a dev environment available on MacOS. If it works on Linux we can merge this, and validate Windows in the patch release candidate.

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
dimitropoulos: actually, it did cause an error but one that I think we want to catch - that we weren't casting to boolean for `checked`, which we should have been because otherwise it throws a runtime react error.  fixed in 0ccc891c54931e64236676dddf0bac072516445a.

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
```diff
@@ -0,0 +1,56 @@
...
+    callback?.();
+  }, [callback, setting]);
+
+  return (
+    <div className="form-control form-control--thin">
+      <label className="inline-block">
+        {label}
+        {help && <HelpTooltip className="space-left">{help}</HelpTooltip>}
+        {forceRestart && (
+          <Tooltip message="Will restart app" className="space-left">
```
gatzjames:
```suggestion
          <Tooltip message="Will restart the app" className="space-left">
```

packages/insomnia-app/app/ui/components/settings/general.tsx
gatzjames: I see that in the app we sometimes use ipcRenderer <-> ipcMain and others we access the electron.remote module to use some apis.
In this case would it make sense to listen for the update of the `settings. autoHideMenuBar` and run the `setMenuBarVisibility` from the main process? What do you think? 🤔 Curious how we handle these cases in general.

gatzjames (commented):
LGTM 👍
Left a small copy improvement suggestion and a question but nothing blocking!

packages/insomnia-app/app/ui/components/settings/general.tsx
develohpanda: @gatzjames I don't see why we couldn't do that? I'm not sure if that will require a restart of the app or if it can happen on the fly if we do so in the main process instead of the render process. 

I will add that there isn't much boilerplate in place to communicate between main <-> render in a safe and maintainable manner right now. There is the gRPC implementation where an attempt has been made using React.Context and confined to the gRPC context. I know we'll have to go down this route in future so there's definitely scope to be thinking about it. In general cases like these happen in the render process unless it _needs_ to happen in the main process.

I hope that helps with an explanation for the current state?

develohpanda (approved):

PR #3962, opened 2021-08-27
Move prompt hide function before the async onComplete handler

### LINKED ISSUE #3959: Fails to launch 2nd prompt window in templateTag since release 2021.5.0
We use the insomnia templateTags functionality to log in to our environment to obtain a JWT (which we need to store for subsequent REST requests)
Your latest release has broken this process because it doesn’t seem to support launching the 2nd dialog window.
See line 11 in the code snippet below (where we launch a dialog to set the password)
```
const jwt = require("jsonwebtoken");

const defaultJTI = "7670cade-d6a2-40dc-9b8d-ea3498467ca3";

async function loginUsingJsonContent(context, models, url, environment) {
  // Load up the previous (successful) username & password from the local store.
  let defaultUsername = await context.store.getItem(`${environment}.username`);
  let defaultPassword = await context.store.getItem(`${environment}.password`);
  // Prompt for username & password (with previous as defaults)
  let username = await context.app.prompt("Enter username", {defaultValue: (defaultUsername || "")});
  let password = await context.app.prompt("Enter password", {defaultValue: (defaultPassword || "")});
  var body = `{"username": "${username}","password": "${password}"}`
  fetch(url, {
    method: 'post',
    headers: {
      "Content-type": "application/json"
    },
    body: body
  })
  .then(response => {
    if (response.status != 200) throw `Failed to login to ${response.url}.  Received status ${response.status}.`;
    return response.json();
  })
  .then(function (data) {
    let decodedToken = jwt.decode(data.access_token, {json : true, complete : true});
    console.log(decodedToken);
    if (isNaN(decodedToken.payload.customer_id)) throw `Successfully logged in, but customer is not valid.  customer_id:${decodedToken.payload.customer_id}`;
    context.store.setItem(`${environment}.customer_id`, decodedToken.payload.customer_id);
    context.store.setItem(`${environment}.token`, data.access_token);
    context.store.setItem(`${environment}.username`, username);
    context.store.setItem(`${environment}.password`, password);
    context.app.alert(`Successfully logged in customer ${decodedToken.payload.customer_id}`);
  })
  .catch(function (error) {
    console.log('Request failed', error);
    context.app.alert('Request failed'+ error);
  });
}

  {
    label: "Login Local Dev",
    icon: "fa-star",
    action: async (context, models) => {
      await loginUsingJsonContent(context, models, "http://dev.127.0.0.1.nip.io/api/auth/v2/login", "localDev");
    },
  },
  {
    label: "Login Cloud Sim",
    icon: "fa-heart",
    action: async (context, models) => {
      await loginUsingJsonContent(context, models, "https://sim.nonprod.gbcloud.com.au/api/auth/v2/login", "cloudSim");
    },
  },
  {
    label: "Login Cloud Forrealz",
    icon: "fa-film",
    action: async (context, models) => {
      await loginUsingJsonContent(context, models, "http://forrealz.nonprod.gbcloud.com.au/api/auth/v2/login", "cloudForrealz");
    },
  },
  {
    label: "Login Local Non Container",
    icon: "fa-home",
    action: async (context, models) => {
      await loginUsingJsonContent(context, models, "http://dev.127.0.0.1.nip.io/api/auth/v2/login", "localNonContainer");
    },
  },
];

module.exports.templateTags = [
  {
    name: "loginAuth",
    displayName: "Login Auth Bearer Token",
    description: "Return the access token from the last login (prefixed with 'Bearer ').",
    args: [
      {
        displayName: 'environment',
        type: 'string',
        defaultValue: 'environment',
      },
    ],
    async run(context, environment) {
      if (!await context.store.hasItem(`${environment}.token`)) {
        throw `Please login to ${environment}`;
      }
      return "Bearer " + (await context.store.getItem(`${environment}.token`));
    },
  },
  {
    name: "loginCustId",
    displayName: "Login Customer Id",
    description: "Return the customer Id from the last login.",
    args: [
      {
        displayName: 'environment',
        type: 'string',
        defaultValue: 'environment',
      },
    ],
    async run(context, environment) {
      if (!await context.store.hasItem(`${environment}.customer_id`)) {
        throw `Please login to ${environment}`;
      }
      return await context.store.getItem(`${environment}.customer_id`);
    },
  },
];
```

gatzjames:
Fixes an issue where opening consecutive prompts only displays the first one.

Since the _done function in the prompt-modal became async, it is now possible for the onComplete function to trigger some logic that opens a new prompt before the old one gets hidden.

You can test that the fix works by generating a new plugin in Insomnia > Preferences > Plugins, and using the code provided in #3959 as an example.

**Before:**

before3959

**After:**

3959after

Closes #3959

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx
+++ b/packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx
@@ -76,8 +76,8 @@ class PromptModal extends PureComponent<{}, State> {
   async _done(rawValue: string) {
     const { onComplete, upperCase } = this.state;
     const value = upperCase ? rawValue.toUpperCase() : rawValue;
-    await onComplete?.(value);
     this.hide();
+    await onComplete?.(value);
   }
 
   _handleCancel() {


jorrocks (approved):

packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx
```diff
@@ -76,8 +76,8 @@ class PromptModal extends PureComponent<{}, State> {
   async _done(rawValue: string) {
     const { onComplete, upperCase } = this.state;
     const value = upperCase ? rawValue.toUpperCase() : rawValue;
-    await onComplete?.(value);
     this.hide();
+    await onComplete?.(value);
```
develohpanda, line 80: The modal hiding before the complete handler is finished seems counter-intuitive, doesn't it? https://github.com/Kong/insomnia/pull/3474 added in a loading state to cater for async event handlers, but maybe that approach needs to be adjusted rather than moving the callback invocation like this. Need to think about what the best solution here would be

develohpanda: I suggest this PR be super-seeded by https://github.com/Kong/insomnia/pull/3970, see the description in there for an explanation about the root cause.

gatzjames: Nice one! 👍

gatzjames:
Closing in favour of #3970

PR #3970, opened 2021-08-30
Don't resolve plugin prompt promise until modal has closed, to avoid a race condition

### LINKED ISSUE #3959: Fails to launch 2nd prompt window in templateTag since release 2021.5.0

develohpanda:
### The background

All modals in this codebase are callback based. That is, to process the result of a modal, you need to send an `onComplete` or `onCancel` callback into the `modal.show()` method.

The plugin API which exposes the prompt modal (`context.app.prompt`) _is not_ callback based, it is promise based. That is, if the prompt was cancelled, `context.app.prompt` should reject with an error, and if the prompt was submitted, `context.app.prompt` should resolve with the prompt value.

The callback-based prompt is converted to a promise-based helper in `packages/insomnia-app/app/plugins/context/app.tsx`.

> The alert modal "attempts" to make itself be both callback-based and promise-based, which is additional confusion, so this (and modals in general) do need a cleanup which is not done in this PR.

### The change

The `onComplete` callback for a prompt was previously synchronous, however an update in https://github.com/Kong/insomnia/pull/3474 introduced asynchronous handlers, adding a loading state and not closing the modal until the `onComplete` handler was finished.

### The race condition

This change introduced a race condition with the plugin API implementation translation. On clicking submit and triggering the `onComplete` callback, the plugin API (which resolves the wrapped promise - see previous implementation) would resolve the promise and tell a plugin that "the prompt is complete", while at the same time the application would be in the process of closing the modal. If the plugin immediately opened another prompt (while the application was still trying to close the first, because of a race condition), the _second_ prompt would close.

### The fix

> We already know that the modal APIs need to be improved, and ideally an attempt made to remove the difference in API (callback-based vs promise-based) to avoid mistakes in translation such as this one. This PR only makes a minimal fix for `context.app.prompt`. One approach to make this plugin stable would be to decouple the `PromptModal` used within the app from the plugin API and have a minimal `PromptModal` specifically for plugins with a restricted implementation matching the API available the plugin developers.

In this PR, when translating from the callback-based prompt modal to a promise-based helper for plugins, we only resolve the promise if:
1. the user pressed submit
1. after the modal has hidden

### The demo

This demo uses the example plugin in #3959 and also https://github.com/Vyoam/insomnia-plugin-postman-export, both of which were impacted by this race condition.

Closes #3959

### DIFF
--- a/packages/insomnia-app/app/plugins/context/app.tsx
+++ b/packages/insomnia-app/app/plugins/context/app.tsx
@@ -69,7 +69,11 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
           return Promise.resolve(options.defaultValue || '');
         }
 
+        // This custom promise converts the prompt modal from being callback-based to reject when the modal is cancelled and resolve when the modal is submitted
         return new Promise((resolve, reject) => {
+          let shouldResolve = false;
+          let resolveWith: string;
+
           showPrompt({
             title,
             ...(options || ({} as Record<string, any>)),
@@ -79,7 +83,15 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
             },
 
             onComplete(value: string) {
-              resolve(value);
+              shouldResolve = true;
+              resolveWith = value;
+            },
+
+            // don't resolve the overall promise until the modal has hidden after clicking submit
+            onHide() {
+              if (shouldResolve){
+                resolve(resolveWith);
+              }
             },
           });
         });
--- a/packages/insomnia-app/app/plugins/context/app.tsx
+++ b/packages/insomnia-app/app/plugins/context/app.tsx
@@ -89,7 +89,7 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
 
             // don't resolve the overall promise until the modal has hidden after clicking submit
             onHide() {
-              if (shouldResolve){
+              if (shouldResolve) {
                 resolve(resolveWith);
               }
             },
--- a/packages/insomnia-app/app/plugins/context/app.tsx
+++ b/packages/insomnia-app/app/plugins/context/app.tsx
@@ -69,10 +69,10 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
           return Promise.resolve(options.defaultValue || '');
         }
 
-        // This custom promise converts the prompt modal from being callback-based to reject when the modal is cancelled and resolve when the modal is submitted
-        return new Promise((resolve, reject) => {
+        // This custom promise converts the prompt modal from being callback-based to reject when the modal is cancelled and resolve when the modal is submitted and hidden
+        return new Promise<string>((resolve, reject) => {
           let shouldResolve = false;
-          let resolveWith: string;
+          let resolveWith: string | null = null;
 
           showPrompt({
             title,
@@ -89,7 +89,7 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
 
             // don't resolve the overall promise until the modal has hidden after clicking submit
             onHide() {
-              if (shouldResolve) {
+              if (shouldResolve && resolveWith !== null) {
                 resolve(resolveWith);
               }
             },
--- a/packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx
+++ b/packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx
@@ -25,6 +25,7 @@ interface State {
   cancelable?: boolean | null;
   onComplete?: (arg0: string) => Promise<void> | void;
   onCancel?: (() => void) | null;
+  onHide?: () => void,
   onDeleteHint?: ((arg0: string) => void) | null;
   currentValue: string;
   loading: boolean;
@@ -44,6 +45,7 @@ export interface PromptModalOptions {
   label?: string;
   hints?: string[];
   onComplete?: (arg0: string) => Promise<void> | void;
+  onHide?: () => void,
   onDeleteHint?: (arg0: string) => void;
   onCancel?: () => void;
 }
@@ -69,6 +71,7 @@ class PromptModal extends PureComponent<{}, State> {
     onComplete: undefined,
     onCancel: undefined,
     onDeleteHint: undefined,
+    onHide: undefined,
     currentValue: '',
     loading: false,
   }
@@ -162,6 +165,7 @@ class PromptModal extends PureComponent<{}, State> {
     onCancel,
     validate,
     onDeleteHint,
+    onHide,
   }: PromptModalOptions) {
     this.setState({
       currentValue: '',
@@ -181,6 +185,7 @@ class PromptModal extends PureComponent<{}, State> {
       validate,
       hints: hints || [],
       loading: false,
+      onHide,
     });
     this.modal?.show();
 
@@ -276,7 +281,7 @@ class PromptModal extends PureComponent<{}, State> {
       'form-control--outlined': inputType !== 'checkbox',
     });
     return (
-      <Modal ref={this._setModalRef} noEscape={!cancelable || loading} onCancel={this._handleCancel}>
+      <Modal ref={this._setModalRef} noEscape={!cancelable || loading} onCancel={this._handleCancel} onHide={this.state.onHide}>
         <ModalHeader>{title}</ModalHeader>
         <ModalBody className="wide">
           <form onSubmit={this._handleSubmit} className="wide pad">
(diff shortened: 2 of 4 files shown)


packages/insomnia-app/app/plugins/context/app.tsx
```diff
@@ -79,7 +83,15 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
...
 
             onComplete(value: string) {
-              resolve(value);
+              shouldResolve = true;
+              resolveWith = value;
+            },
+
+            // don't resolve the overall promise until the modal has hidden after clicking submit
+            onHide() {
+              if (shouldResolve){
```
gatzjames:
```suggestion
              if (shouldResolve) {
```

gatzjames (approved):
Nicely done and very well explained 🙏

packages/insomnia-app/app/plugins/context/app.tsx
```diff
@@ -69,7 +69,11 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
           return Promise.resolve(options.defaultValue || '');
         }
 
+        // This custom promise converts the prompt modal from being callback-based to reject when the modal is cancelled and resolve when the modal is submitted
         return new Promise((resolve, reject) => {
+          let shouldResolve = false;
+          let resolveWith: string;
```
dimitropoulos:
```suggestion
          let resolveWith: string | null = null;
```

this is going to need to be fixed when strict mode is on, best not to create more technical debt now.

packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx
```diff
@@ -69,6 +71,7 @@ class PromptModal extends PureComponent<{}, State> {
     onComplete: undefined,
     onCancel: undefined,
     onDeleteHint: undefined,
+    onHide: undefined,
```
dimitropoulos, line 74: why not just not define it (to match the type)? when we have https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types on we'll need to come back and fix this.

packages/insomnia-app/app/plugins/context/app.tsx
```diff
@@ -79,7 +83,15 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
             },
 
             onComplete(value: string) {
-              resolve(value);
+              shouldResolve = true;
+              resolveWith = value;
+            },
```
dimitropoulos, line 88: I'm guessing you probably agree, actually, but I'm worried about how confusing it is that calling an `onComplete` callback doesn't actually... well... haha... _complete_.  The promise doesn't actually resolve until the `onHide` function is called, and that's only called from within the a custom callback on a custom component on a custom wrapper...

```ts
        showModal(WrapperModal, {
          title,
          body: <HtmlElementWrapper el={body} onUnmount={options.onHide} />,
          tall: options.tall,
          skinny: options.skinny,
          wide: options.wide,
        });
```

I totally understand how in this moment it can make sense to do it this way, but I don't think this is an acceptable long-term solution because it adds even more complexity and cognitive-overhead to something that's already quite complex (hence, us introducing the original bug _in the first place_).

dimitropoulos (commented):
Overall, I want to see this fixed in production as soon as we can, so I don't think we should blockade this solution for now, but I definitely want to (if we merge this approach) see us fix the problem of global modals soon, which will (in itself) completely solve this issue.  I can give examples you want of codebases that accomplish non-global modals.

dimitropoulos (approved):
This needs to be fixed immediately.  The fix here works.  Therefore I approve.

However, I think the fact that we're having a patch release due to a gotcha in the overall architecture of modals and then we had to fix it by introducing what I consider to be a new gotcha (that onComplete only defers the actual completion logic to onHide) should cause us to consider improving how modals work in the app.  A necessary gotcha - I agree - given the current state of things.  But we can change the state of things and incrementally improve.  I think we should do so soon.

packages/insomnia-app/app/plugins/context/app.tsx
develohpanda: I think the snippet above is from something else, but fundamentally I understand and agree with your observation :raised_hands: All of this arises from the fact that we have single instance modals shared across all calls; we fix that, we fix many of these issues.

packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx
develohpanda: Because it shouldn't be initialized to anything; we should just leave it off the initial state here, but since the other handlers initialize as undefined I followed the same pattern here. I vote, in this instance, on coming back to fix it once that's enabled, because this isn't the only instance that will need fixing in this file

packages/insomnia-app/app/plugins/context/app.tsx
develohpanda: I think the snippet above is from something else, but fundamentally I understand and agree with your observation 🙌 All of this arises from the fact that we have single instance modals shared across all calls; we fix that, we fix many of these issues.

PR #3977, opened 2021-08-30
[Feature/Plugin API] Add clipboard wrapper methods to context app

jasonblalock:
- Adds clipboard object with `.readText` and `.writeText` methods using
underlying electorn clipboard methods

Closes INS-919

### DIFF
--- a/packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
+++ b/packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
@@ -20,7 +20,7 @@ describe('init()', () => {
       'prompt',
       'showGenericModalDialog',
       'showSaveDialog',
-    ]);
+    ].sort());
     expect(Object.keys(result.app.clipboard).sort()).toEqual([
       'clear',
       'readText',
--- a/packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
+++ b/packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
@@ -25,7 +25,7 @@ describe('init()', () => {
       'clear',
       'readText',
       'writeText',
-    ]);
+    ].sort());
   });
 });
 
--- a/packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
+++ b/packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
@@ -1,3 +1,6 @@
+import electron from 'electron';
+import { mocked } from 'ts-jest/utils';
+
 import { globalBeforeEach } from '../../../__jest__/before-each';
 import { RENDER_PURPOSE_SEND } from '../../../common/render';
 import * as modals from '../../../ui/components/modals';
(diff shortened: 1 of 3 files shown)


develohpanda (approved):
Nice! Verified with a small plugin I made to QA, which exports a workspace directly into your clipboard! Ever since you talked about the clipboard API being available to plugins, I've had so many quality-of-life ideas which can be built out as plugins 🙌🏽 

For anybody else looking at this, there is further discussion about this in the [Insomnia Slack](https://getinsomnia.slack.com/archives/CF69F4J7M/p1628795685006200).

```js
  {
    action: async (context, {workspace}) => {
      const exportContents = await context.data.export.insomnia(workspace);

      // This is new!!
      context.app.clipboard.writeText(exportContents);

      await context.app.alert('Exported to clipboard', `The workspace ${workspace.name} has been exported to your clipboard.`);
    },
    label: 'Export to clipboard',
  },
]
```

packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
```diff
@@ -11,12 +14,18 @@ describe('init()', () => {
     expect(Object.keys(result)).toEqual(['app', '__private']);
     expect(Object.keys(result.app).sort()).toEqual([
       'alert',
+      'clipboard',
       'dialog',
       'getPath',
       'prompt',
       'showGenericModalDialog',
       'showSaveDialog',
     ]);
```
jasonblalock: Curious how you feel about sorting both the expected and the result keys so that the test is indifferent to order since I don’t think that’s what we’re actually trying to test here? Could also use a linting rule if we're sorting more for the developer experience (let machines do what they do best).

```suggestion
    ].sort());
```

develohpanda: Heh, meet @dimitropoulos, who is also super keen on sorted properties 😝 I think we'll get there, eventually!

That sounds reasonable to me though, in this instance it would be better to use `.sort()` in code because the expect block also does a `.sort()`. Nice catch!

wdawson (approved):
Looks great to me! ⛵ Thanks for the work on this!

packages/insomnia-app/app/plugins/context/__tests__/app.test.ts
```diff
@@ -11,11 +14,17 @@ describe('init()', () => {
...
       'getPath',
       'prompt',
       'showGenericModalDialog',
       'showSaveDialog',
+    ].sort());
+    expect(Object.keys(result.app.clipboard).sort()).toEqual([
+      'clear',
+      'readText',
+      'writeText',
     ]);
```
develohpanda:
```suggestion
    ].sort());
```

DMarby (approved):

dimitropoulos (approved):
lgtm - thanks!  This oughta be useful!

PR #3985, opened 2021-08-31
Update README.md to appropriately describe product

dimitropoulos:
closes: https://github.com/Kong/insomnia/discussions/3834


wdawson (approved):
Looks good. Do we also support SOAP or is that just not worth calling out in the 2020s?

dimitropoulos:
not sure, but even if we do probably best to not mention it, haha.

PR #3991, opened 2021-09-02
fixes styling regression for autocomplete

dimitropoulos:
I noticed that the styling for the autocomplete has broken.  The background doesn't fill all the way across (and, it should).

## Before

https://user-images.githubusercontent.com/15232461/131869557-f2d948b7-c8ca-4722-a2df-ce01c5e3c043.mp4

## After

https://user-images.githubusercontent.com/15232461/131869780-cd272977-04b8-4a90-983d-623b95d73e29.mp4

There's a way in which, I guess, this isn't a regression, but it seems to have been uncovered after https://github.com/Kong/insomnia/pull/3983.  Visually, though, it is a regression.

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.ts
+++ b/packages/insomnia-app/app/ui/components/codemirror/extensions/autocomplete.ts
@@ -108,8 +108,9 @@ CodeMirror.defineOption('environmentAutocomplete', null, (cm, options) => {
           widget.close();
           return CodeMirror.Pass;
         },
-      }, // Good for debugging
-      // ,closeOnUnfocus: false
+      },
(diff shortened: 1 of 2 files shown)


develohpanda (approved):
Woah, this is intense.

image

Nice find for the background, thanks for fixing!

dimitropoulos:
yeah, so re: those strong tags, I too was amazed when when I first saw it but it appears to be a side-effect of how the fuzzy search highlighting works.  I looked into it and we could easily disable it.... but I didn't see a reason to try.  It's fairly benign.  If you wanna make a PR it's almost definitely a 3-5 line fix (basically, just skip the mapping with strongs when there's no search).

Issue #3990, opened 2021-09-02
ERROR all request and response in curly braces with no Tag menu - request chaining


theundertakerjr666:
### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for a bug that matches the one I want to file, without success.

### Insomnia Version

2021.5.2

### What operating system are you using?

macOS

### Operating System Version

macOS Catalina 11.5.2

### Installation method

download app from insomnia

### Expected Behavior

{
	"expr": [
		"a =[Response > Body attribute]"
	]
}

### Actual Behavior

{
	"expr": [
		"a ={% jsonata_response '', '', 'never' %}"
	]
}

### Reproduction Steps

1) create request 1, it should generate results such as a login token
2) refer to request chaining by typing in Response > body attribute , application would return {% jsonata_response '', '', 'never' %}"
3) all function is undefined

<img width="462" alt="Screenshot 2021-09-02 at 17 11 29" src="https://user-images.githubusercontent.com/4310314/131859341-d6de86fb-2189-4b3f-b377-e83766de5220.png">

### Additional Information

There is no drop down menu to select attribute or map chained responses

theundertakerjr666:
<img width="794" alt="Screenshot 2021-09-02 at 17 14 23" src="https://user-images.githubusercontent.com/4310314/131859693-47dada62-d141-4a2e-a34b-4c635ab7ccd4.png">

dimitropoulos:
Is it possible this was solved by https://github.com/Kong/insomnia/pull/3983 (merged just 10 hours ago)?

I just tried, and was unable to reproduce:

https://user-images.githubusercontent.com/15232461/131864612-17055763-fc2e-4f34-ac70-2c3dcb447a5c.mp4

any suggestions?

theundertakerjr666:
I have not fully resolved it yet, but it seems to be related to "Raw template syntax" option selected in Insomnia's preference

dimitropoulos:
Does it work if you change the setting of `Raw template syntax`?

develohpanda:
Hi @theundertakerjr666, the default Response -> Body Attribute raw template should be `{% response 'body', '', '', 'never', 60 %}`, not `{% jsonata_response '', '', 'never' %}` as you are experiencing.

Could you please share a gif/video to show exactly what you are experiencing?

[...]

PR #3992, opened 2021-09-02
fixes nunjucks-tag pointer

dimitropoulos:
I noticed while working on https://github.com/Kong/insomnia/issues/3990 that this bit of UI doesn't use the correct pointer.  If you don't know that this needs to be clicked for the UX flow to continue, the cursor can go a long way towards hinting that this but of UI needs to be clicked on.

| BEFORE | AFTER |
| - | - |
| image | image
|


develohpanda (approved):
We don't really use pointers anywhere else in the app so I will defer to @wdawson / @erickrico for this one. Personally, I think this is a good change.

wdawson:
Looks good to me! 👍

PR #3993, opened 2021-09-02
Improve codemirror typings

gatzjames:
Still work in progress 💪 but most of the @ts-expect-error comments are gone from the code-editor and the places it's being used at.


packages/insomnia-app/app/codemirror.d.ts
```diff
@@ -1,9 +1,19 @@
...
 import { HandleGetRenderContext, HandleRender } from './common/render';
+import { NunjucksParsedTag } from './templating/utils';
 
 type LinkClickCallback = (url: string) => void;
 
+declare module 'codemirror-graphql/jump' {
+  type ModifiedGraphQLJumpOptions = Omit<GraphQLJumpOptions, 'onClick'> & {
+    onClick: GraphQLInfoOptions['onClick'];
+  };
+}
```
dimitropoulos: since codemirror-graphql is a separate package, I think it'd be best if we kept with the pattern of having a separate file for each declaration.  I get that these come from the same monorepo and all that, so in this case I can see the motivation for having it in one place.  Still, it seems a lot cleaner to me to separate like we do everywhere else.  Your thoughts?

packages/insomnia-app/app/ui/components/codemirror/modes/nunjucks.ts
```diff
@@ -1,4 +1,11 @@
 import CodeMirror from 'codemirror';
+
+export function isNunjucksMode(
+  mode: CodeMirror.ModeSpec<unknown>
+): mode is CodeMirror.ModeSpec<{ baseMode: 'string' }> {
+  return 'baseMode' in mode;
```
dimitropoulos, line 6: I get really nervous about uses of the `in` operator because of how many gotchas there are regarding prototypes.  Is it possible that `Object.prototype.hasOwnProperty.call(mode, 'baseMode')` will work here?

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -163,12 +164,16 @@ class GraphQLEditor extends PureComponent<Props, State> {
     });
   }
 
-  _handleClickReference(reference: ActiveReference, event: React.MouseEvent) {
+  _handleClickReference(reference: Maybe<ActiveReference>, event: MouseEvent) {
```
dimitropoulos, line 177: lib.dom's `MouseEvent` (https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts#L10230) and React's `MouseEvent` (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1238) are different.  If this is fine, then cool, but just wanted to call it out.

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -701,19 +707,19 @@ class GraphQLEditor extends PureComponent<Props, State> {
...
               onClick: this._handleClickReference,
-            }}
-            jumpOptions={{
-              schema: schema || null,
+            } : undefined}
+            jumpOptions={schema ? {
+              schema,
               onClick: this._handleClickReference,
-            }}
+            } : undefined}
```
dimitropoulos: what about

```
...{schema ? {
  hintOptions: {
    schema,
    completeSingle: false,
  },
  infoOptions: {
    schema,
    renderDescription: GraphQLEditor.renderMarkdown,
    onClick: this._handleClickReference,
  },
  jumpOptions: {
    schema,
    onClick: this._handleClickReference,
  },
} : {}}

File: packages/insomnia-app/app/ui/components/modals/code-prompt-modal.tsx [hunk 5]
```diff
@@ -192,10 +190,8 @@ class CodePromptModal extends PureComponent<Props, State> {
                   onChange={this._handleChange}
                   nunjucksPowerUserMode={nunjucksPowerUserMode}
                   isVariableUncovered={isVariableUncovered}
-                  // @ts-expect-error -- TSCONVERSION appears to be a genuine error
-                  getRenderContext={enableRender ? handleGetRenderContext : null}
-                  // @ts-expect-error -- TSCONVERSION appears to be a genuine error
-                  render={enableRender ? handleRender : null}
+                  getRenderContext={handleGetRenderContext}
+                  render={handleRender}
```
dimitropoulos, line 194: was it intentional that the conditions were removed here?

packages/insomnia-app/app/ui/components/modals/generate-code-modal.tsx
```diff
@@ -36,9 +37,9 @@ interface Props {
 
 interface State {
   cmd: string;
-  request: null;
-  target: string;
-  client: string;
+  request?: Request;
+  target: HTTPSnippetTarget;
+  client: HTTPSnippetClient;
```
dimitropoulos, line 42: sweet.

packages/insomnia-app/app/ui/components/modals/generate-code-modal.tsx
```diff
@@ -13,8 +14,8 @@ import ModalFooter from '../base/modal-footer';
 import ModalHeader from '../base/modal-header';
 import CodeEditor from '../codemirror/code-editor';
 
-const DEFAULT_TARGET = availableTargets().find(t => t.key === 'shell');
-const DEFAULT_CLIENT = DEFAULT_TARGET.clients.find(t => t.key === 'curl');
+const DEFAULT_TARGET = HTTPSnippet.availableTargets().find(t => t.key === 'shell');
+const DEFAULT_CLIENT = DEFAULT_TARGET?.clients.find(t => t.key === 'curl');
```
dimitropoulos, line 18: right now, the type of this is `any`
```suggestion
const DEFAULT_CLIENT = DEFAULT_TARGET?.clients.find(t => t.key === 'curl') as HTTPSnippetClient;
```

dimitropoulos: right now, this is typed as `HTTPSnippetClient | undefined`, but we know it's defined

```suggestion
const DEFAULT_CLIENT = DEFAULT_TARGET?.clients.find(t => t.key === 'curl') as HTTPSnippetClient;
```

packages/insomnia-app/app/ui/components/modals/generate-code-modal.tsx
```diff
@@ -13,8 +14,8 @@ import ModalFooter from '../base/modal-footer';
 import ModalHeader from '../base/modal-header';
 import CodeEditor from '../codemirror/code-editor';
 
-const DEFAULT_TARGET = availableTargets().find(t => t.key === 'shell');
-const DEFAULT_CLIENT = DEFAULT_TARGET.clients.find(t => t.key === 'curl');
+const DEFAULT_TARGET = HTTPSnippet.availableTargets().find(t => t.key === 'shell');
```
dimitropoulos: right now, this is typed as `HTTPSnippetTarget | undefined`, but we know it's defined

```suggestion
const DEFAULT_TARGET = HTTPSnippet.availableTargets().find(t => t.key === 'shell') as HTTPSnippetTarget;
```

packages/insomnia-app/app/ui/components/modals/generate-code-modal.tsx
```diff
@@ -85,30 +84,31 @@ class GenerateCodeModal extends PureComponent<Props, State> {
...
-    // @ts-expect-error -- TSCONVERSION
     if (currentTarget.key === target.key) {
       // No change
       return;
     }
 
     const client = target.clients.find(c => c.key === target.default);
 
-    this._generateCode(this.state.request, target, client);
+    request && this._generateCode(request, target, client);
```
dimitropoulos: please use an `if` statement:

```ts
if (request) {
  const client = target.clients.find(c => c.key === target.default);
  this._generateCode(request, target, client);
}
```

or, in this case, maybe

```ts
if (!request) {
  return;
}

const client = target.clients.find(c => c.key === target.default);
this._generateCode(request, target, client);
 ```

File: packages/insomnia-app/app/ui/components/modals/generate-config-modal.tsx [hunk 10]
```diff
@@ -112,8 +112,6 @@ class GenerateConfigModal extends PureComponent<Props, State> {
           lineWrapping={settings.editorLineWrapping}
           mode={config.mimeType}
           nunjucksPowerUserMode={settings.nunjucksPowerUserMode}
-          // @ts-expect-error -- TSCONVERSION appears to be genuine
-          onChange={this._handleChange}
```
dimitropoulos, line 116: I'm sure you had a good reason, but could you share why this was removed?

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -146,6 +148,10 @@ interface State {
   filter: string;
 }
 
+function isMarkerRange(mark: CodeMirror.Position | CodeMirror.MarkerRange | undefined): mark is CodeMirror.MarkerRange {
+  return Boolean(mark && 'from' in mark);
```
dimitropoulos: can you please refactor this to be simpler and also not use the `in` operator?

```suggestion
  if (!mark) {
    return false;
  }
  return Object.prototype.hasOwnProperty.call(mark, 'from');
```

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -348,26 +354,23 @@ class CodeEditor extends Component<Props, State> {
       return;
     }
 
-    // @ts-expect-error this is a (Node & ParentNode) in the types, but I think it's actually supposed to be Element
-    this.codeMirror.getTextArea().parentNode?.setAttribute(name, value);
+    this.codeMirror.getTextArea().parentElement?.setAttribute(name, value);
```
dimitropoulos, line 359: I'm nervous about this change just because these kinds of subtle changes can be hard to debug.  are `parentElement` and `parentNode` always the same?

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -412,16 +415,20 @@ class CodeEditor extends Component<Props, State> {
 
     const marks = this.codeMirror
       .getAllMarks()
-      // @ts-expect-error -- TSCONVERSION
-      .filter(c => c.__isFold)
+      .filter(mark => mark.__isFold)
```
dimitropoulos, line 420: thank you for fixing the variable name

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -412,16 +415,20 @@ class CodeEditor extends Component<Props, State> {
...
-        const { from, to } = mark.find();
-        return {
-          from,
-          to,
-        };
+        const result = mark.find();
+
+        if (isMarkerRange(result)) {
+          return result;
+        } else {
```
dimitropoulos: I'll add https://eslint.org/docs/rules/no-else-return in a separate PR, but let's stick to it here preemptively

packages/insomnia-app/app/httpsnippet.d.ts
```diff
@@ -0,0 +1,33 @@
+// @TODO We should make a PR for improving the types of httpsnippet.
```
dimitropoulos, line 1: I'm starting to think we should make a PR to just convert the whole thing to native typescript

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -844,15 +851,18 @@ class CodeEditor extends Component<Props, State> {
     }
 
     // Strip of charset if there is one
-    Object.keys(options).map(key => {
-      this._codemirrorSmartSetOption(key, options[key]);
-    });
+    Object.keys(options).map(key =>
```
dimitropoulos, line 858: if you use ramda's `keys` it should "just work"

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -762,15 +769,15 @@ class CodeEditor extends Component<Props, State> {
...
 
     if (!hideGutters && options.lineNumbers) {
-      options.gutters.push('CodeMirror-linenumbers');
+      options.gutters?.push('CodeMirror-linenumbers');
     }
 
     if (!hideGutters && options.foldGutter) {
-      options.gutters.push('CodeMirror-foldgutter');
+      options.gutters?.push('CodeMirror-foldgutter');
     }
```
dimitropoulos: maybe it'd be better to wrap this whole thing in `if (options.gutters && !hideGutters)`

packages/insomnia-app/app/codemirror.d.ts
develohpanda: I agree with the above 👍🏽

packages/insomnia-app/app/ui/components/modals/generate-config-modal.tsx
gatzjames: Yeah! The editor is readOnly so it doesn't fire the onChange. Also there is not a _handleChange method in this component.

generate-config-modal

packages/insomnia-app/app/ui/components/modals/code-prompt-modal.tsx
gatzjames: Yes they are optional in the props so the check is happening inside the component.

packages/insomnia-app/app/httpsnippet.d.ts
develohpanda: Our team maintains this repository (as well)

dimitropoulos: Today I learned... Interesting

packages/insomnia-app/app/codemirror.d.ts
dimitropoulos: 305a2675072f00741b5c38a5dc4c56acc66c7154

packages/insomnia-app/app/ui/components/editors/body/graph-ql-editor.tsx
gatzjames: This event along with some focus events e.g. blur/focus are fired from Codemirror instead of React.

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
dimitropoulos: because nodes don't have attributes anyway, it's clearer to set attribute on elements

packages/insomnia-app/app/ui/components/modals/generate-code-modal.tsx
dimitropoulos: 6a911d21177f38df81ee7abbe45e186a73d5d8e6

packages/insomnia-app/app/ui/components/modals/generate-code-modal.tsx
dimitropoulos: cb6603e0ea8539dbf8e5e5530dcd41d87c9cef5c

packages/insomnia-app/app/ui/components/modals/generate-code-modal.tsx
dimitropoulos: cb6603e0ea8539dbf8e5e5530dcd41d87c9cef5c

dimitropoulos (approved):

PR #3995, opened 2021-09-02
fixes querystring updating regression

### LINKED ISSUE #3788: The URL preview does not update when working with queries
**Describe the bug**
The URL preview in query section does not refresh after switching parameters on and off

**To Reproduce**
1. Select one request
2. Go to Query section
3. Note the value of URL in URL Preview
4. Enable one more query

**Expected behavior**
After switching queries on and off, URL preview needs to me updated

**Screenshots**
image
In image the expected URL in preview should be `http://10.42.0.97:8060/launch/dev?contentId=5c5dc3a0e78ba537db1af5e8%7C5c5debbb2e1a27473d9caa3f&cacheclean=true`

 - OS: Arch 5.12.14-arch1-1
 - Installation Method: insomnia-bin from AUR
 - App Version v2021.4.0

**Additional info**
The preview is updated after moving focus to another request and then back.
I have reproduced the issue on appimage too. I have tried running fresh by removing my config files from ~/.config/Insomnia

dimitropoulos:
closes https://github.com/Kong/insomnia/issues/3788

I removed the debounce (as we have been, there's absolutely no need for it here) and that solved the bug.  I fixed the positioning while I was there (the text wasn't centered vertically before) as well as the fact that the copy button wasn't disabled when there was nothing to copy.

### DIFF
--- a/packages/insomnia-app/app/ui/components/rendered-query-string.tsx
+++ b/packages/insomnia-app/app/ui/components/rendered-query-string.tsx
@@ -1,104 +1,73 @@
-import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import { buildQueryStringFromParams, joinUrlAndQueryString, smartEncodeUrl } from 'insomnia-url';
-import React, { PureComponent } from 'react';
+import React, { FC, useState } from 'react';
+import { useAsync } from 'react-use';
+import styled from 'styled-components';
 
-import { AUTOBIND_CFG } from '../../common/constants';
 import { HandleRender } from '../../common/render';
-import CopyButton from './base/copy-button';
+import { Request } from '../../models/request';
+import { CopyButton as _CopyButton } from './base/copy-button';
+
+const Wrapper = styled.div({
+  display: 'flex',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+  overflow: 'auto',
+  position: 'relative',
+  width: '100%',
+});
+
+const CopyButton = styled(_CopyButton)({
+  alignSelf: 'start',
+});
 
 interface Props {
-  request: any;
+  request: Request;
   handleRender: HandleRender;
 }
 
-interface State {
-  string: string;
-}
+const defaultPreview = '...';
 
-@autoBindMethodsForReact(AUTOBIND_CFG)
-class RenderedQueryString extends PureComponent<Props, State> {
-  state: State = {
-    string: '',
-  };
+export const RenderedQueryString: FC<Props> = ({ request, handleRender }) => {
+  const [previewString, setPreviewString] = useState(defaultPreview);
 
-  _interval: NodeJS.Timeout | null = null;
-
-  async _debouncedUpdate(props: Props) {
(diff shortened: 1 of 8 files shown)


packages/insomnia-app/app/ui/components/rendered-query-string.tsx
```diff
@@ -1,104 +1,73 @@
...
-          <i className="fa fa-copy" />
-        </CopyButton>
-        {inner}
-      </div>
-    );
-  }
-}
-
-export default RenderedQueryString;
+  }, [request, handleRender]);
```
develohpanda: I would suggest changing this implementation so that the the only items from `request` needed in the dependency array are `request.url`, `request.settingEncodeUrl`, `request.parameters`.

I don't think it's necessary to invoke the templating render `handleRender` (which is why, my guess is, the debounce existed) when _any_ property on the `request` changes.

develohpanda (changes requested):
Ran locally and looks to be working as intended 🙌🏽 As with removing any debounce timeouts, do you know why it might have been there in the first place? Was there a valid reason for it? 

The above question is _not_ the requested change; rather the comment below. But I'm still curious if there is any background for this debounce to exist (even if it wasn't working as intended).

packages/insomnia-app/app/ui/components/rendered-query-string.tsx
dimitropoulos: cool - updated in 869c95923a128afb02e769d4e69646dbaad732a1

dimitropoulos: This component has basically been this way from the beginning with no explanation (that I can find).  I remember react back in 2016 and this kind of thing was in vogue back then because... well... haha, it was the beginning in a lot of ways and performance wasn't anywhere near what it is now.  People were looking for ways to squeeze performance out of react and this was one of the ones they came up with. 
 
https://github.com/Kong/insomnia/commit/bb6e3859716264dc443842d45dc829040de04a4b#diff-c9d2ff9244afd131e2c8d0be1635a2e3de6d63220bd739205e282b89463224d8 is the start "`LOts of good stuff`".  You can see in that commit that this wasn't the only file he did this to - which is more evidence that this wasn't something that had a product need per-se, but rather just a general approach he was trying out. 

The refactor to `debouncedUpdate` came later https://github.com/Kong/insomnia/commit/14640eafd6c395da8872c3a13a7c7479d3d48708 but the end result is essentially the same, in the `Some small fixes` commit.

gatzjames:
Works nice! 👍  Added some space between the button and the text and made the button sticky in https://github.com/Kong/insomnia/pull/3995/commits/237596e38b457921f8e4058933d353fa1835bbc5

sticky-button

gatzjames (approved):

dimitropoulos:
Wow great suggestion @gatzjames !  Love it!

DusanLesan:
Great. Thank you for fixing this. I cannot wait for release.

PR #3999, opened 2021-09-03
Update response headers viewer to FC

develohpanda:
Resolves an ES Lint warning, removes inline repetitive styling in favor of `styled-components` (but still using named CSS classes that exist), and removes index as key.

This is general cleanup and should have no behavior change.

image


reynolek (approved):
LGTM

gatzjames (approved):
🙌

PR #4002, opened 2021-09-04
[Docs] Add devel docs for package installation on Fedora

jasonblalock:
- Initial Dev Setup only included installing OS dependencies for Linux distributions that use apt-get (Ubuntu/Debian)
- This adds instructions for what deps are needed on Fedora using dnf
- Added OS headers for clearer separation of OS reqs/troubleshooting

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ npm test
 npm run app-start
 ```
 
+### Linux
+
 If you are on Linux, you may need to install the following supporting packages:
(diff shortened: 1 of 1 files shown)


reynolek (approved):
Thanks for taking the time to improve our README!

develohpanda (approved):
Thank you for adding this! 🙏🏽

PR #4003, opened 2021-09-05
[Maintenance] Update package-lock.json to lockfileversion 2

### LINKED ISSUE #4004: Update package-lock.json to lockversion 2 via npm v7
I was bootstrapping your insomnia project for the first time using npm v7. Noticed it went through and automatically updated all the package-lock.json files to the new lockfileversion 2. Thought it might be something that would be good to do, but if not, nbd. 
You can see the changes that occurred at #4003. More details about that here: https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile. 

> Changes to the lockfile
> 
> One change to take note of is the new lockfile format, which is backwards compatible with npm 6 users. The lockfile v2 unlocks the ability to do deterministic and reproducible builds to produce a package tree.
> 
> In prior versions, the yarn.lock files were ignored, the npm CLI can now use yarn.lock as the source of package metadata and resolution guidance. If a yarn.lock file is present, then npm will also keep it up-to-date with the contents of the package tree.
> 
> Running npm install with npm 7 in a project with a v1 lockfile will replace that lockfile with the new v2 format. To avoid this, you can run npm install --no-save.

jasonblalock:
- Using npm 7 automatically updates package-lock.json to new
  lockfileversion 2 on `npm install`
- This is backwards compatible with version 1 on npm 6
- Reference: https://github.blog/2021-02-02-npm-7-is-now-generally-available/#changes-to-the-lockfile

- Closes #4004


develohpanda:
Thanks a tonne for this! Because we don't have an immediate release coming up, CI is passing fine, and NPM 7 is generally available, I'm okay with merging this and seeing how it goes; we can always switch back if necessary. We as a team should look to switch to move to NPM 7 as well.

One thing I note from the blog is that peer dependencies get automatically installed now (yay!). This is exciting but means it'll be worth double checking all of our existing packages in this monorepo and make sure their peer deps are correctly defined.

## TODO:
- [x] Ensure no redundant peer dependencies
- [x] Ensure peer dependency versions are correct (fixed, I think?)

develohpanda (commented):
Looks like there was only one discrepancy, I fixed that in a666b85e0f9a1a3c71ea9c8c42271c4708a1adca. 🙌🏽 

When I ran `npm run bootstrap` it replaced the `package-lock.json` at the root with the V1 version, and on running `npm run clean && npm run bootstrap` it replaced all of them with the V1 versions.

We might want to wait for a little bit, or notify existing contributors somehow of the switch to NPM 7 to avoid these being committed. Is there a way to specify which version of NPM to use (similar to .nvmrc for the version of Node)? I tried to set the engines property in the root package.json but that didn't seem to do the trick in notifiying I was not on NPM 7.

jasonblalock:
Well that's unfortunate considering the whole point was that the new version was supposed to be compatible with the old version. So why would you convert it back. 

I'm going to poke at it to see if I can find a way to see if there's something similar to what you're doing with `scripts/check-version.js`. I can think of at least one way to have the warning come up, but I want to see if there's something more integrated in npm.

jasonblalock:
Opened #4020 separately to add the npm version checks. Could add it here if you'd prefer. But it might be easier to add all the package-lock.json files to the new one if you like one of the version check methodologies to avoid continuous merge conflicts since this touches all the lockfiles.

jasonblalock:
Per discussions in https://github.com/Kong/insomnia/pull/4020#discussion_r706984869 this will not be pursued until node version is updated such that npm v7 is the accompanying version

Issue #2050, opened 2020-04-24
[Bug] XPath filter returns no results


brad-young:
**Describe the bug**
I am trying to chain together a few API calls. Step #1, get a TOKEN, Step #2, use the token in subsequent calls.

Here is the sample response from the TOKEN API call:

```xml
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope
  xmlns:soap="http://www.w3.org/2003/05/soap-envelope"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <soap:Body>
    <LoginResponse
      xmlns="http://url/DataService/">
      <LoginResult>xxx-777-xxx-123</LoginResult>
    </LoginResponse>
  </soap:Body>
</soap:Envelope>

```

Then in my "next" API call, I am using the Insomnia RESPONSE functionality, by grabbing the TOKEN from the body.

Using the following XPath filter:
```
//*[local-name(.)='LoginResult']/text()
```

However, the UI in the modal window says "Returned no results".

The example does work in https://extendsclass.com/xpath-tester.html

It appears to be the xmlns value

**Screenshots**
attached 

**Desktop (please complete the following information):**
 - OS: macOS 10.14.6
![Screen Shot 2020-04-24 at 12 44 44 PM](https://user-images.githubusercontent.com/7274594/80238470-80742600-862c-11ea-8cba-fdb5ddfec356.png)
![Screen Shot 2020-04-24 at 1 07 25 PM](https://user-images.githubusercontent.com/7274594/80238512-9255c900-862c-11ea-96b4-9b20b7b92236.png)

gschier:
I think I've seen this reported before somewhere. It looks like the XPath library we're using doesn't like the `xmlns="http://url/DataService/"` attribute. Removing this makes the query `//LoginResult` work. However, the `text()` function also seems to be not supported, as `//LoginResult/text()` returns nothing.

Perhaps we should look at updating our XPath dependency or finding a more-complete one.

brad-young:
yeah, the hard part is not being able to control the TOKEN API call that adds the xmlns. And .... I don't think there's another way to do this, correct? If I want to chain together commands and grab the token value, this is really the only method.

gschier:
Not quite. You could write your own fork of the Response Tag plugin or maybe make use of this one https://insomnia.rest/plugins/insomnia-plugin-response-eval

[...]

PR #4012, opened 2021-09-08
Don't encode if string is empty in the tag editor

develohpanda:
When using a template tag string argument with encoding enabled (as per the [response => body attribute template tag](https://github.com/Kong/insomnia/blob/6a60e2bbbca6214270653be3bd235eec4a6d7e79/plugins/insomnia-plugin-response/index.js#L50-L50)), Insomnia encodes the filter value, but if the filter value is empty it should not attempt to encode and leave it blank.

In the following demo video, the first example is live, and second is this PR. This fix was inspired by this comment (https://github.com/Kong/insomnia/issues/2050#issuecomment-908826203), noticed while exploring XPath issues.

https://www.loom.com/share/7a3bbf3c52b048ce902e634be0bd9198

Closes INS-947

### DIFF
--- a/packages/insomnia-app/app/templating/utils.ts
+++ b/packages/insomnia-app/app/templating/utils.ts
@@ -246,8 +246,12 @@ export function getDefaultFill(name: string, args: NunjucksParsedTagArg[]) {
   return `${name} ${stringArgs.join(', ')}`;
 }
 
-export function encodeEncoding(value: string, encoding: 'base64') {
+export function encodeEncoding<T>(value: T, encoding?: 'base64' | 'unknown' | 'utf8') {
   if (typeof value !== 'string') {
+    return value;
+  }
+
+  if (value.length === 0) {
     return value;
   }
 
@@ -259,7 +263,7 @@ export function encodeEncoding(value: string, encoding: 'base64') {
   return value;
 }
 
-export function decodeEncoding(value: string) {
+export function decodeEncoding<T>(value: T) {
   if (typeof value !== 'string') {
     return value;
   }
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/templating/utils.ts
```diff
@@ -246,11 +246,15 @@ export function getDefaultFill(name: string, args: NunjucksParsedTagArg[]) {
   return `${name} ${stringArgs.join(', ')}`;
 }
 
-export function encodeEncoding(value: string, encoding: 'base64') {
+export function encodeEncoding<T>(value: T, encoding?: 'base64' | 'unknown' | 'utf8') {
```
develohpanda: I typed encoding this way because unit tests test these encoding values 🤔

gatzjames: Looks like it's only used by the tag-editor so we could even deprecate the other cases if we wanted:
https://github.com/Kong/insomnia/blob/7ffc39142807e5efe69ecdf798f20c72e64ad82b/packages/insomnia-app/app/ui/components/templating/tag-editor.tsx#L271

develohpanda: Also true: it's only used by one thing and it's hard-coded even though the plugin configuration might say 'base64'. I'll get rid of the extra types and the unit tests!

develohpanda: Resolved in [`93f4551` (#4012)](https://github.com/Kong/insomnia/pull/4012/commits/93f4551b8443d86b55c3a7d0a58a3f52af812997)

gatzjames (approved):
👍

PR #4017, opened 2021-09-09
Use the database change listener in the global tests

gatzjames:
In some of the global tests we fire an async action and then wait for an arbitrary amount of time before asserting against the database.
I've replaced that with using the db.onChange handler, making the tests more reliable.


develohpanda (approved):
VERY NICE!!!

DMarby (approved):

dimitropoulos (approved):
it's a little strange to me how `onDatabaseChange` sends `database.offChange` a reference to itself, but if it works it works! haha :)

PR #4019, opened 2021-09-10
fixes all react/prefer-stateless-function warnings

dimitropoulos:
There are two things to look out for in reviewing this PR:

1. a component that did `const { x, y } = this.props` at the top but then later did `this.props.z` in the meat of the component.  Thankfully this didn't appear to happy in any of the affected components because they were all so small.  I did a self review at the end and didn't find any of these.
1. there are duplicate components (Dropdown) in insomnia-app and insomnia-components.  It's possible that I messed up an import somewhere.  I did a self review and caught two of these.  One _really_ cool thing about this is that if this changes it will cross the "empty line internal vs. external dependency barrier" because of the import sorting!!! that's how I noticed it, first.  _very_ convenient.
1. check that every time `PureComponent` was used previously (in components with props), `memo` is used now.

Other than conversion to function components, there should be no logical changes in this PR.


packages/insomnia-app/app/__jest__/with-redux-store.tsx
```diff
@@ -1,13 +1,10 @@
...
-  class ReduxWrapper extends React.Component {
-    render() {
-      const { children } = this.props;
-      return <Provider store={store}>{children}</Provider>;
-    }
-  };
+// eslint-disable-next-line react/display-name -- There's not a good way to do with a FunctionComponent while also maintaining the display name.
+export const withReduxStore = (store: MockStoreEnhanced<RootState, {}>): FC => ({ children }) => (
+  <Provider store={store}>{children}</Provider>
+);
```
dimitropoulos, line 10: this was the only one that gave me trouble.  if you fix to sfc it complains that it doesn't have a display name, but I couldn't figure a way how to do it otherwise.

packages/insomnia-app/app/ui/components/base/dropdown/dropdown-right.tsx
```diff
@@ -1,20 +0,0 @@
-import classnames from 'classnames';
```
dimitropoulos: this component is not used anywhere in the monorepo and is not exported.   it was excluded from the index file (https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/base/dropdown/index.ts).

aside from being touched in global typescript and eslint refactors, this file appears not to have been touched in 5 years.

packages/insomnia-components/src/card-container.tsx
```diff
@@ -1,14 +1,7 @@
...
   display: flex;
   flex-wrap: wrap;
   padding-top: var(--padding-md);
 `;
-
-export class CardContainer extends PureComponent {
-  render() {
-    return <Cards {...this.props} />;
-  }
-}
```
dimitropoulos, line 14: I sat and thought about this to make sure that it is logically equivalent and I cannot see a way in which it is not.  always good to double check, though.  This is a "straight out of the docs" way to do styled components and while I can understand how someone might write code like this when we were first learning how to use styled-components (https://github.com/Kong/studio/commit/7c58e5907fdc1bcacbb52b0b97cfce7db40dfeba#diff-3562ac93155ce13a13bc5cb18b227c72bdcc674a8174e430060fc195fc1f4007) the extra components provides no value.

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-default-value.tsx
```diff
@@ -1,26 +1,23 @@
...
   field: GraphQLFieldWithParentName;
 }
 
-class GraphQLDefaultValue extends PureComponent<Props> {
-  render() {
-    const { field } = this.props;
-    // Make Flow happy :/
-    const fieldO: Record<string, any> = field;
+export const GraphQLDefaultValue: FC<Props> = memo(({ field }) => {
+  // Make Flow happy :/
```
jackkav, line 11: What flow? 😉

packages/insomnia-app/app/ui/components/base/dropdown/dropdown.tsx
```diff
@@ -8,9 +9,8 @@ import { hotKeyRefs } from '../../../../common/hotkeys';
 import { executeHotKey } from '../../../../common/hotkeys-listener';
 import { fuzzyMatch } from '../../../../common/misc';
 import KeydownBinder from '../../keydown-binder';
-import DropdownButton from './dropdown-button';
-import DropdownDivider from './dropdown-divider';
-import DropdownItem from './dropdown-item';
+import { DropdownDivider } from './dropdown-divider';
+import { DropdownItem } from './dropdown-item';
```
jackkav, line 13:
```suggestion
import { DropdownItem } from './dropdown-item';
import { DropdownButton }  from './dropdown-button';
```

packages/insomnia-app/app/ui/components/base/dropdown/dropdown.tsx
```diff
@@ -36,8 +36,18 @@ interface State {
...
 
+const isComponent = (match: string) => (child: ReactNode) => any(equals(match), [
+  // @ts-expect-error not sure
+  child.type.name,
+  // @ts-expect-error not sure
+  child.type.displayName,
+]);
+
+const isDropdownItem = isComponent(DropdownItem.name);
+const isDropdownDivider = isComponent(DropdownDivider.name);
```
jackkav, line 47:
```suggestion
const isDropdownDivider = isComponent(DropdownDivider.name);
const isDropdownButton = isComponent(DropdownButton.name);
```

jackkav (changes requested):

packages/insomnia-app/app/ui/components/graph-ql-explorer/graph-ql-default-value.tsx
jackkav: Looks awesome so far, I'm curious why you chose FC over plain functions in light of [this article](https://fettblog.eu/typescript-react-why-i-dont-use-react-fc/), but just a nitpick for my benefit.

I tested it locally and fixed the IsDropdownButton changes to make it run, and get lots of console logs complaining about `Dropdown needs exactly one DropdownButton!` and missing components
image

dimitropoulos: dang, I thought I got that in b01a15affbff41d7ab4540fe1cca7a14d2d9265b

dimitropoulos: (reverted all changes to dropdown components in https://github.com/Kong/insomnia/pull/4019/commits/318989242973b53ddcf400e79a887c7abbc39fd4).  There is neither the time nor do I have the requisite energy to fix it "for real" right now.

gatzjames (approved):
🎉 LGTM

PR #4030, opened 2021-09-14
prevent fetching spellcheck Dictionary

gatzjames:
Disables spellchecking in electron and sets the dictionary download URL to `https://00.00/` as a way to prevent fetching anything from Chromium's CDN.

TODO:
- [x] Validate the behaviour in Linux/Windows

Closes INS-891

### DIFF
--- a/packages/insomnia-app/app/main/window-utils.ts
+++ b/packages/insomnia-app/app/main/window-utils.ts
@@ -26,9 +26,8 @@ const { app, Menu, shell, dialog, clipboard, session } = electron;
 app.allowRendererProcessReuse = false;
 
 app.on('ready', () => {
-  // There's no option that prevents Electron from fetching spellcheck dictionaries from
-  // Chromium's CDN and passing a non-resolving URL is the only known way to prevent it from
-  // fetching.  https://github.com/electron/electron/issues/22995
+  // There's no option that prevents Electron from fetching spellcheck dictionaries from Chromium's CDN and passing a non-resolving URL is the only known way to prevent it from fetching.
+  // see: https://github.com/electron/electron/issues/22995
   // On macOS the OS spellchecker is used and therefore we do not download any dictionary files.
   // This API is a no-op on macOS.
   session.defaultSession.setSpellCheckerDictionaryDownloadURL('https://00.00/');
--- a/packages/insomnia-app/app/main/window-utils.ts
+++ b/packages/insomnia-app/app/main/window-utils.ts
@@ -20,10 +20,19 @@ import { clickLink, getDataDirectory, restartApp } from '../common/electron-help
 import * as log from '../common/log';
 import LocalStorage from './local-storage';
 
-const { app, Menu, shell, dialog, clipboard } = electron;
+const { app, Menu, shell, dialog, clipboard, session } = electron;
 // So we can use native modules in renderer
 // NOTE: This was (deprecated in Electron 10)[https://github.com/electron/electron/issues/18397] and (removed in Electron 14)[https://github.com/electron/electron/pull/26874]
 app.allowRendererProcessReuse = false;
+
+app.on('ready', () => {
+  // There's no option that prevents Electron from fetching spellcheck dictionaries from
+  // Chromium's CDN and passing a non-resolving URL is the only known way to prevent it from
+  // fetching.  https://github.com/electron/electron/issues/22995
+  // On macOS the OS spellchecker is used and therefore we do not download any dictionary files.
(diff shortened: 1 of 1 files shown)


packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -20,11 +20,20 @@ import { clickLink, getDataDirectory, restartApp } from '../common/electron-help
...
-const { app, Menu, shell, dialog, clipboard } = electron;
+const { app, Menu, shell, dialog, clipboard, session } = electron;
 // So we can use native modules in renderer
 // NOTE: This was (deprecated in Electron 10)[https://github.com/electron/electron/issues/18397] and (removed in Electron 14)[https://github.com/electron/electron/pull/26874]
 app.allowRendererProcessReuse = false;
 
+app.on('ready', () => {
+  // There's no option that prevents Electron from fetching spellcheck dictionaries from
+  // Chromium's CDN and passing a non-resolving URL is the only known way to prevent it from
+  // fetching.  https://github.com/electron/electron/issues/22995
```
dimitropoulos: nit: no reason to break lines unnecessarily:
Screenshot_20210914_202710

```suggestion
  // There's no option that prevents Electron from fetching spellcheck dictionaries from Chromium's CDN and passing a non-resolving URL is the only known way to prevent it from fetching.
  // see: https://github.com/electron/electron/issues/22995
```

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -20,11 +20,20 @@ import { clickLink, getDataDirectory, restartApp } from '../common/electron-help
...
+  // On macOS the OS spellchecker is used and therefore we do not download any dictionary files.
+  // This API is a no-op on macOS.
+  session.defaultSession.setSpellCheckerDictionaryDownloadURL('https://00.00/');
```
develohpanda, line 33: Would setting it this way also apply the same setting to OAuth windows that open, or just to the main window of the app?

What is the best way to validate this - through WireShark for instance, or were you able to detect it though electron itself?

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -85,6 +94,7 @@ export function createWindow() {
     acceptFirstMouse: true,
     icon: path.resolve(__dirname, appLogo),
     webPreferences: {
+      spellcheck: false,
```
develohpanda: Worth double checking where else we create new windows, there are a few, here's one for example:
https://github.com/Kong/insomnia/blob/ccac337c4f327329f32bb93ac1324cfe28562c2b/packages/insomnia-app/app/network/o-auth-2/misc.ts#L76-L83

dimitropoulos (approved):
before this PR, when running 

```
strace -fe trace=network -o network2.log npm run app-start
```

The following two lines were observed.

```
162898 sendto(26, ";\35\1 \0\1\0\0\0\0\0\1\nredirector\4gvt1\3com"..., 48, MSG_NOSIGNAL, NULL, 0) = 48
162898 recvfrom(26, ";\35\201\200\0\1\0\1\0\0\0\1\nredirector\4gvt1\3com"..., 1024, 0, {sa_family=AF_INET, sin_port=htons(53), sin_addr=inet_addr("127.0.0.53")}, [28->16]) = 64
```

After this PR, both lines were gone.

packages/insomnia-app/app/main/window-utils.ts
dimitropoulos: an update on this: @gatzjames, @johnwchadwick and I discussed this and determined that we actually want to leave it on.  Ultimately, the thing we are trying to do is prevent the fetch from happening.  But for MacOS users, there _is no_ fetch because chromium uses builtin functionality.  It's not like we want to turn off spellchecking as a product goal - we just want to disable the fetch.  If linux and windows both worked like mac, then we'd never be in this situation digging around.  So, based on that, there's nothing to lose from leaving this enabled.  If there's no dictionary then spellchecking will just not work, anyway, so there's no error or anything like that that will surface to the user.

packages/insomnia-app/app/main/window-utils.ts
dimitropoulos: we finally figured out an approach that was reliable: check out https://github.com/Kong/insomnia/pull/4030#pullrequestreview-756340122

develohpanda:
@dimitropoulos I feel like that's a valuable note to make in our Confluence space so that if/when we ever need to investigate again we know what to do!

PR #4036, opened 2021-09-17
[Feature] The proxy configuration values are now hidden by default

### LINKED ISSUE #3138: Feature request - Hide proxy id and password when provided in http proxy field.
As part of growing security needs and to conceal private information it is vital that we hide proxy id and password in the http proxy field. We can opt for a password-type proxy id and password text fields in this case. 

We can come up with a custom proxy configuration approach like what Postman provides.

image

Or just make the field show ***** once the user has typed in the text. 
image

**Additional context**

MrSnix:
**Feature request**
The information in the proxy settings field shouldn't be visible due to security concerns.

**Feature implementation**
The proxy input field is now hidden by default

**Showcase**

Sep-13-2021 20-45-24

Closes #3138


develohpanda (changes requested):
Thanks for working through this! It looks pretty good and appears to work as well. 🙌🏽 

I have a couple of bits of feedback for the direction we are trying to take the `General` preferences component in, which I will explain below.

Firstly, we are looking to move from the `renderBooleanSetting()` class method approach to a standalone FC approach - a [`BooleanSetting`](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx). We want to do this for the other ones too (`renderNumberSetting` and `renderTextSetting`) but haven't been able to yet. 

Since this PR introduces a new one (`renderPasswordSetting`), I suggest we refactor that to a `MaskedSetting` (renamed to be more generic) similar to the `BooleanSetting`. This way the new component also manages it's own state, allowing us to remove the `fieldsHidden` state property!

Secondly, at the top of the general preferences there is an option to `Reveal Passwords`, which controls the password visibility everywhere in the app. I recorded a [demo](https://www.loom.com/share/359f567daff5433182d14cb5ba51052d) of that to show how it works the the password field in the Basic Auth form. 

Ideally, this `MaskedSetting` component also adheres to that setting (which is fairly straightforward to achieve). The Basic Auth password field (amongst other places) uses the [`PasswordEditor`](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/editors/password-editor.tsx) component. This, under the hood, uses code-mirror so our templating engine works, but that is **not** necessary for preferences, which should just use a regular `input` as per your current implementation.

tl;dr:
1. Take inspiration from [`BooleanSetting`](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx) to create a `MaskedSetting` component
2. Take inspiration from [`PasswordEditor`](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/editors/password-editor.tsx) to connect the `MaskedSetting` component to the `Reveal Passwords` setting.

How does that sound? 😄

MrSnix:
**EDIT**: This was fixed

Hi dear @develohpanda, I applied all the requested changes 😄  but ... something went wrong 😞 (?) when I refactored the input field as a React component and I can't figure out the reason why 🧐 . After the component was created, it started suffering from **input lag**. I tried several options but none of them worked (e.g React.PureComponent, React.useMemo).
Seems like updating the setting value takes time and this causes the lag (?) (but it was great before I refactored it to a component). The input lag is strong when I keep pressing the backspace button 🔙  to delete the current value and gets worse and worse with time. (it gets progressively stuck more and more). I wonder if the `BooleanSetting` suffers from the same issue but we can't see it 

I can debounce the input ... but is it the right solution?

Here is an attached .gif of the behaviour:

Peek 26-09-2021 14-27

MrSnix:
My previous commit had some mistakes, **I fixed the input lag.**
I apologise.

I simply need some feedback about the current implementation

Thanks

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const [isHidden, setHidden] = useState(true);
+
+  const settings = useSelector(selectSettings);
+
+  if (!settings.hasOwnProperty(setting)) {
+    throw new Error(`Invalid masked setting name ${setting}`);
+  }
+
+  const [input, setInput] = useState(settings[setting] + '');
```
develohpanda: Seems unnecesary to keep a local copy of the input as well as within the database (via the `onInputChange` handler). We should be okay to remove this local state, and use `<input value={String(settings[setting])} ... />` further down.

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { ChangeEvent, FC, useState } from 'react';
+import { useSelector } from 'react-redux';
+
+import * as models from '../../../models';
+import { BaseSettings } from '../../../models/settings';
+import { selectSettings } from '../../redux/selectors';
+import { HelpTooltip } from '../help-tooltip';
+
+export const MaskedSetting: FC<{
```
develohpanda, line 10: A couple of notes on the props for this component. This component serves as a controlled wrapper around the `input` element specifically for text/password inputs. We should be able to use the parameters from `renderTextSetting` as props for this component.

- We can only realistically handle a `text`/`password` input type, so we can avoid exposing `type` as a prop and hardcode the input type further down.
- `showPasswords` doesn't need to be a prop. By using the settings selector in the component body already, we can also access `settings.showPasswords` from there directly

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <input
+          value={input}
+          type={!showPasswords && isHidden ? 'password' : type}
+          name={setting}
+          onChange={onInputChange}
+          {...props}
+        />
+        {!showPasswords && (
+          <button className={'form-control__right'} onClick={() => setHidden(!isHidden)}>
+            {isHidden ? <i className="fa fa-eye" /> : <i className="fa fa-eye-slash" />}
```
develohpanda: Looks like this ternary is flipped. `fa-eye` should be the icon when passwords are visible, and `fa-eye-slash` when passwords are hidden.

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
```diff
@@ -0,0 +1,65 @@
...
+}> = ({
+  label,
+  setting,
+  type,
+  showPasswords,
+  help,
+  props,
+}) => {
+
+  const [isHidden, setHidden] = useState(true);
```
develohpanda: Try `import { useToggle } from 'react-use';` for this :)

https://github.com/streamich/react-use/blob/master/docs/useToggle.md

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { selectSettings } from '../../redux/selectors';
+import { HelpTooltip } from '../help-tooltip';
+
+export const MaskedSetting: FC<{
+  label: string;
+  setting: keyof BaseSettings;
+  type: string;
+  showPasswords: boolean;
+  help?: string;
+  props?: Record<string, any>;
```
develohpanda:
```suggestion
  props?: React.HTMLProps<HTMLInputElement>;
```

packages/insomnia-app/app/ui/css/components/forms.less
```diff
@@ -9,7 +9,7 @@
     position: relative;
 
     input {
-      padding-right: 2em;
+      padding-right: 2em !important;
```
develohpanda: Is this still needed? I checked with and without and couldn't see any UI differences.

develohpanda (changes requested):
Making really good progress 🙌🏽 A couple more notes please!

packages/insomnia-app/app/ui/css/components/forms.less
MrSnix: Yes, IMO I like it this way... but if it doesn't follow the guide-lines I can remove it

**With** `!important` :
image

**Without**:
image

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
MrSnix: There is a specific reason why I duplicated the state.

**If I don't do it**, this happens because the key-stroke event is fired way before the update gets executed even if we `await` the database inside the `onChange` handler
Oct-04-2021 17-30-14

**If I do it**, everything works like a charm:
Oct-04-2021 18-02-46

Let me know what do you think

packages/insomnia-app/app/ui/css/components/forms.less
develohpanda: Ah I see, good catch. Seems to happen here as well
image

develohpanda: I removed the `!important` because it can be quite hard to debug and replaced it with a specificity fix: [`11866bc` (#4036)](https://github.com/Kong/insomnia/pull/4036/commits/11866bc77c6edd725f3a6332c810fd2339c2aae1)

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
develohpanda: I was experimenting with this a bit, and I think for right now the fix would be to leave the input as uncontrolled, so instead of setting `value` we set `defaultValue`.

This is how the `DebouncedInput` works mainly, as well as the [code-editor](https://github.com/Kong/insomnia/blob/d53ef39c9b76aefcfa71630078d838bb95eaad56/packages/insomnia-app/app/ui/components/editors/body/raw-editor.tsx#L53) itself.

[`114d396` (#4036)](https://github.com/Kong/insomnia/pull/4036/commits/114d396fd0e0dbd6f28d0ab20d87f56b65f5d3e4)

develohpanda:
@MrSnix I pushed a couple of commits, have a look and let me know if you have any questions 😄 if not, I'll do a final review

MrSnix:
@develohpanda LGTM!

Feel free to go :D

Thanks as always for the time spent on this

develohpanda:
Excellent, thank _you_! This is a great improvement 👏🏽

develohpanda (approved):
Looks and works great! This will probably get some more work put into it soon as we migrate the remaining settings into standalone components. Thanks for working through it!

https://www.loom.com/share/5c07f57a39e9452e91c3c8f8b0d7b119

packages/insomnia-app/app/ui/components/settings/masked-setting.tsx
```diff
@@ -0,0 +1,59 @@
...
+import * as models from '../../../models';
+import { Settings } from '../../../models/settings';
+import { selectSettings } from '../../redux/selectors';
+import { HelpTooltip } from '../help-tooltip';
+
+export const MaskedSetting: FC<{
+  label: string;
+  setting: keyof Settings;
+  help?: string;
+  props?: React.HTMLProps<HTMLInputElement>;
```
dimitropoulos, line 14: I think we should be very careful about using `props` (unqualified) as a prop name.  That creates a situation where we have `Component.props.props.someValue`.  Not saying this is the same thing since we're destructuring, but it's unusual to do this, all the same.

What I suggest is we do something like "inputProps".

Another approach would be to do something like [MUI does](https://mui.com/customization/unstyled-components/#anatomy-of-an-unstyled-component), `componentProps.input` could contain the overrides.

------

However, from another perspective, why is `disabled` and `placeholder` being wrapped in the first place?  They seem like normal props that are very much "on topic" for the MaskedSetting component.  From that perspective, we should just remove the `props` wrapper altogether.

That would be my vote, which, thankfully, sidesteps the entire issue of what to name the wrapper prop.

We're planning to loop back on these settings components soon, so we can make this adjustment then, it doesn't need to block this PR at this moment.

dimitropoulos (commented):
One non-blocking thing I noticed: it's good practice when making a password input like this to not leak the value of the input to the DOM if the password is hidden.  For example, the password fields for Okta, 1Password, GitHub, Amazon, and many more all work like this.  This is also what we do for `PasswordViewer`, if you wanna take a look, although it's a little easier there because it's not an input.

dimitropoulos:
as always, @MrSnix, thank you for the contribution!

MrSnix:
Hi @dimitropoulos, I am glad to see your feedback on this one.

> The password fields for Okta, 1Password, GitHub, Amazon, and many more all work like this. 
> This is also what we do for `PasswordViewer` if you wanna take a look.

I just took a look and it was enlightening, I wasn't aware of this or better I thought about something like this in the past but I didn't know it was a thing. **Thanks a lot**, I just learnt something new :D

> As always, @MrSnix, thank you for the contribution!

It is a pleasure to see that contributing has a positive impact on the community.
Even small changes, if I can address them, I try to.

arafattechstyle:
Hi, do we know if this is going to be part of which release?

develohpanda:
Hi @arafattechstyle, this will be part of a release coming this week!

PR #4039, opened 2021-09-18
[Feature] Introducing un-safe plugin installation

MrSnix:
**Issue**
Yarn prints warnings messages (e.g deprecated dependencies, custom library messages) on the stderr and Insomnia erroneously treat them as installation errors.

**What this PR aims for**
Resolve the **edge cases** which may prevent a plugin to be installed just for a couple of warnings.

**What is this "unsafe" mode all about**
Insomnia executes its own plugin installation as always ... but _skips_ the stderr check on its last step.

 **Implementation**
The user can only install un-safely when he gets an installation failure **that failed when the plugin reached the stderr check.**

**Showcase**
showcase

As always feedback are welcome.

### DIFF
--- a/packages/insomnia-app/app/plugins/install.ts
+++ b/packages/insomnia-app/app/plugins/install.ts
@@ -40,7 +40,7 @@ interface InsomniaPlugin {
   };
 }
 
-export default async function(lookupName: string) {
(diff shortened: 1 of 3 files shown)


hockeymikey:
Why was this closed?

PR #4040, opened 2021-09-19
[Bug] Improved HTTP custom methods detection while importing specs

MrSnix:
changelog(Fixes): circular references in OpenAPI specs should now be allowed

**Issue**
While I was working on #4015 I had this issue while importing large open-api specs where
insomnia-importer wrongly detects properties such as "description" or "parameters" as custom HTTP methods

**Implementation**
I reworked just a bit the filtering function, now it excludes anything that is not a plain object or custom open-api extensions.
The reason is pretty simple, usually, a method is represented as a plain object (as opposed to `parameters` which is an array) and the method name never starts with `x-` (open-api extensions). So I thought it would be acceptable to rewrite the function as stated, this should resolve all the possible issues instead of checking they do not match some given words (e.g 'parameters') as it is done now.

As you can see from the picture, all the methods are plain objects, parameters is an array and descriptive fields are plain strings.

Screenshot from 2021-09-18 23:41:06

**Before**
Screenshot from 2021-09-18 23:43:48

**After**
Screenshot from 2021-09-19 14:59:32

As always opinions, feedback and observations are welcome.


packages/insomnia-importers/src/importers/openapi-3.ts
```diff
@@ -134,13 +134,24 @@ const parseEndpoints = (document?: OpenAPIV3.Document | null) => {
...
-          path,
-          method,
-        }));
+      const isSpecExtension = (method: string): boolean => {
+        return method.indexOf('x-') === 0;
+      };
+
+      const isPlainObject = (input: object): boolean => {
+        return input && !Array.isArray(input) && typeof input === 'object';
+      };
```
develohpanda: We can define these outside the `parseEndpoints` function so that they're not recreated each time

packages/insomnia-importers/src/importers/openapi-3.ts
```diff
@@ -134,13 +134,24 @@ const parseEndpoints = (document?: OpenAPIV3.Document | null) => {
...
+
+      const methods = Object.entries(schemasPerMethod)
+        // We are filtering out anything that is not a plain object
+        // and custom open-api extensions
```
develohpanda:
```suggestion
        // Only keep entries that are plain objects and not spec extensions
```

develohpanda (approved):
Works as intended when I import a spec that uses a custom HTTP method (which is why this non-restrictive approach exists). Thanks!

image

packages/insomnia-importers/src/importers/openapi-3.ts
MrSnix: I always forget about this thing ... 😢 

Thanks

develohpanda (approved):

packages/insomnia-importers/src/importers/openapi-3.ts
```diff
@@ -110,6 +110,22 @@ const parseDocument = (rawData: string): OpenAPIV3.Document | null => {
...
+const isSpecExtension = (property: string): boolean => {
+  return property.indexOf('x-') === 0;
+};
+
+/**
+ * Checks if the given value is a plain object
+ * @param value The value to test
+ */
+const isPlainObject = (value: object): boolean => {
+  return value && !Array.isArray(value) && typeof value === 'object';
```
dimitropoulos: We have the `ban-types` rule turned off because of our widespread usage of `Function` during the TypeScript conversion, but if it were on it would have errorred on this line.  

Screenshot_20211110_095415

But, beyond that, this is function really takes something even higher level than `Record<..., ...>`, it ultimately takes `unknown` and _asserts_ that the value it passes is a `Record<string, unknown>`.  I would therefore suggest we use a [type-guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards):

```suggestion
const isPlainObject = (value: unknown): value is Record<string, unknown> => {
  return value !== null && !Array.isArray(value) && typeof value === 'object';
};
```

You'll notice one other change, the addition of `!== null` at the beginning.  This is required because since `unknown` is not assignable to boolean and thus can't be used with `&&`.  Also, though, it makes a clear distinction since the `typeof null` (very unfortunately...) returns `object`.

At this point, the first and last condition are a reimplementation of [`isObjectLike` from lodash](https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L11772).

Speaking of which, `ramda-adjunct` has `isPlainObject`, and **I suggest we use it in this situation**.  There are some tricky edge cases that this implementation doesn't cover and I'd like to avoid getting into it in our codebase.  For example, this function will return `true` for any DOM element (e.g. `document.getElementsByTagName('div')[0]`), but Lodash, Ramda, and others (reasonably so, I'd argue) return `false`.  For some more examples, the function here will return `false` for `Math` but `true` for `Error` when it should return `false` for both.  It will return `true` for `arguments`, as well.

Anyway, I don't mean to pick it apart - just thought you'd appreciate to hear some of the reasoning for not introducing a function like this into our codebase since, lol, like many things in JavaScript, it's way harder once you get into realms, proxies, builtins, and so on.

packages/insomnia-importers/src/importers/openapi-3.ts
```diff
@@ -110,6 +110,22 @@ const parseDocument = (rawData: string): OpenAPIV3.Document | null => {
   }
 };
 
+/**
+ * Checks if the given property name is an open-api extension
+ * @param property The property name
+ */
+const isSpecExtension = (property: string): boolean => {
```
dimitropoulos: if you wanted to, we could make this a type guard like this:

```suggestion
const isSpecExtension = (property: string): property is SpecExtension => {
```

We'd just need to add

```ts
type SpecExtension = `x-${string}`;
```
on the line above the tsdoc for this function.

packages/insomnia-importers/src/importers/openapi-3.ts
```diff
@@ -134,13 +150,15 @@ const parseEndpoints = (document?: OpenAPIV3.Document | null) => {
...
+        // Only keep entries that are plain objects and not spec extensions
+        .filter(([key, value]) => isPlainObject(value) && !isSpecExtension(key));
+
+      return methods.map(([method]) => ({
+        ...((schemasPerMethod as Record<string, OpenAPIV3.SchemaObject>)[method]),
+        path,
+        method,
+      }));
     })
     .flat();
```
dimitropoulos, line 157: {speaking about the state of the codebase before this PR} I find it _very_ difficult to understand what this function (`parseEndpoints`) is doing, and I think it would REALLY benefit from some tests.  I was going to suggest some updates to this function to make it easier to read, but I ran out of time.

But all the same, that has nothing to do with your work in this PR because it was already in that state.  You made it better, at least, and a little easier to read.

If it's any interest to you, here's where I got (bonus points because it allows for removing the cast):

```ts
  const endpointsSchemas = pipe(
    () => document.paths,
    mapObjIndexed((schemasPerMethod, path) => {
      if (!schemasPerMethod) {
        return [];
      }

      return toPairs(schemasPerMethod)
        // Only keep entries that are not spec extensions
        .filter(([method]) => !isSpecExtension(method))
        .map(([method, pathItemObject]) => ({
          ...(isPlainObject(pathItemObject) ? pathItemObject : {}),
          path,
          method,
        }));
    }),
    values,
    flatten,
  )();
```

While, according to TypeScript, this change works, I am afraid to touch it at all without tests.  To write tests would require understanding it and after spending a nearly-unacceptably-long-time looking at it I still don't feel like I understand what it does enough to write tests, heh.  cc: @gatzjames

dimitropoulos (approved):
none of the changes I mentioned are blocking.  thanks again!

packages/insomnia-importers/src/importers/openapi-3.ts
MrSnix: Hi @dimitropoulos, **you are absolutely right**, it's good practice to use strong and battle-proof library functions for this kind of thing. **I don't even know why I didn't think to use lodash in the first place.**

> Just thought you'd appreciate to hear some of the reasoning for not introducing a function like this into our codebase

I am 2000% with you on this one.

---

The current package doesn't have `ramda-adjunct` listed as a dependency but there is `lodash`.

If you want I can add it, otherwise I can use the lodash function, **let me know which one to use.**

dimitropoulos: so sorry I didn't see this sooner, I went ahead and added it in 4205ea84393dc602e2960c5d704916c296ef4bab to speed things up a bit.  apologies again.

packages/insomnia-importers/src/importers/openapi-3.ts
dimitropoulos: added in 960c47514d47c362d0606e78cca3b4b6bbf7325e just to get this PR closer to merging

PR #4042, opened 2021-09-21
Update axios-request to respect the validateSSL app setting

gatzjames:
The plugin api exposes the __private.axios method to allow plugins to send requests outside of the insomnia request lifecycle.
This PR checks the `validateSSL` app setting in order to skip validating certificates.

Tested using the `insomnia-plugin-kong-deploy` which uses the axios module to login to the kong-portal and upload the configuration:

deploy-to-portal

Closes INS-941

### DIFF
--- a/packages/insomnia-app/app/plugins/context/app.tsx
+++ b/packages/insomnia-app/app/plugins/context/app.tsx
@@ -2,10 +2,7 @@ import * as electron from 'electron';
 import React from 'react';
 
 import * as analytics from '../../../app/common/analytics';
-import {
-  axiosRequest as axios,
-  axiosRequest,
-} from '../../../app/network/axios-request';
+import { axiosRequest as axios } from '../../../app/network/axios-request';
 import { getAppPlatform, getAppVersion } from '../../common/constants';
 import type { RenderPurpose } from '../../common/render';
 import {
@@ -15,6 +12,7 @@ import {
 } from '../../common/render';
 import HtmlElementWrapper from '../../ui/components/html-element-wrapper';
 import { showAlert, showModal, showPrompt } from '../../ui/components/modals';
+import { PromptModalOptions } from '../../ui/components/modals/prompt-modal';
 import WrapperModal from '../../ui/components/modals/wrapper-modal';
 
 interface DialogOptions {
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/plugins/context/app.tsx
```diff
@@ -2,7 +2,10 @@ import * as electron from 'electron';
 import React from 'react';
 
 import * as analytics from '../../../app/common/analytics';
-import { axiosRequest as axios } from '../../../app/network/axios-request';
+import {
+  axiosRequest as axios,
+  axiosRequest,
+} from '../../../app/network/axios-request';
```
develohpanda: I'm guessing this double import was a mistake :)

packages/insomnia-app/app/plugins/context/app.tsx
```diff
@@ -14,11 +17,71 @@ import HtmlElementWrapper from '../../ui/components/html-element-wrapper';
...
+  skinny?: boolean;
+  wide?: boolean;
+}
+
+interface PromptOptions {
+  label?: string;
+  defaultValue?: string;
+  submitName?: string;
+  cancelable?: boolean;
+}
```
develohpanda: I don't think these should be defined explicitly because they're subsets of the actual options:

https://github.com/Kong/insomnia/blob/7ffc39142807e5efe69ecdf798f20c72e64ad82b/packages/insomnia-app/app/ui/components/modals/prompt-modal.tsx#L34-L51

So I would favor using `Pick` where possible

develohpanda (approved):
I don't have a local instance of Kong running in order to validate this but seeing your gif was useful, thanks!

Are there any mocked unit tests we can write to ensure this functionality doesn't break?

packages/insomnia-app/app/plugins/context/app.tsx
develohpanda: Also in future I actually want to expose the additional options to plugins, but for right now let's keep the API as-is

gatzjames: Agree! I wanted to expose/export them somehow but it would expand this PR too much.

develohpanda: Yeah totally. We need to think of a way to generate and distribute the plugin API typescript types

Issue #3786, opened 2021-07-05
Insomnia goes idle for 2 minutes before every request.


Raibo:
**Describe the bug**
We have to wait up to 2 minutes for a request to happen.
On my PC it happens to be about 130 seconds. It was less, about 20-40 seconds before, now it's really hard to use.

When I hit "Send" button, "Loading" title appears for about a second, then time count starts and Insomnia goes idle.

During this idle period:
- Insomnia ignores the "Cancel" button, therefore request cannot be canceled.
- Other menus are available (for example, "Preferences", "Manage Environments")
- Insomnia process starts to require 15-40% CPU, doing seemingly nothing. (Intel Core i5-4690K)
- Application, which is supposed to receive the request, registers nothing coming. (Unlike insomnia, browser receive responses immediately for same endpoints)

After the request is done:
- From the moment it actually "happens", everything is ok.
- Insomnia shows timing (next to request status) not including the "Idle" period, several milliseconds.
- In "Timeline" tab, in "Current time is " statement, it shows time after the idle period passed.

**To Reproduce**
1. Click 'Send' in any request. (It reproduces for any request)

**Expected behavior**
Request starts immediately or shortly, in a few seconds.

**Screenshots**
![image](https://user-images.githubusercontent.com/8547320/124468109-26cfc580-ddb2-11eb-8806-9cba6d5bfb02.png)
![image](https://user-images.githubusercontent.com/8547320/124471463-51238200-ddb6-11eb-901a-91d9be57a552.png)

**Desktop:**
 - OS: 
 1. Microsoft Windows [Version 10.0.18363.1621]
 2. MAC OS 2021.4.0
 - Installation Method: -
 - App Version: Insomnia 2021.4.0

DFSko:
Confirm. I also have this problem

costyn:
I also have a similar issue. Recently updated to 2021.4.1, and any New Requests I've added since then will time out after a minute with a 'Error: Server returned nothing (no headers, no data)' error. The exact same query in (for example) Postman instantly returns the correct data.  Old requests still in Insomnia work ok. And funnily enough, if I edit the parameters of an old request to match the new one, it also works. 

OS: Windows 10 Enterprise version 20H2 13/‎01/‎2021
App version: 2021.04.1

dimitropoulos:
I was unable to reproduce this.  Any tips?  I sent a request to `mockbin.org/delay/3000` and it started immediately and finished after the specified time.

[...]

PR #4043, opened 2021-09-21
adds ndjson language support extension

dimitropoulos:
When providing an answer in https://github.com/Kong/insomnia/issues/3786#issuecomment-924236677 I realized that we'd benefit from just including this extension.

Our database files are newline-delimited JSON (http://ndjson.org/), the [native format of NeDB](https://github.com/louischatriot/nedb/issues/518#issuecomment-322640935).

| BEFORE | AFTER |
| - | - |
| annoying prompt: <br />Screenshot_20210921_141455 | \~no annoying prompt\~ |
| no language support: <br /> Screenshot_20210921_141618 | language support: <br /> Screenshot_20210921_141546 |

It's often enough, at least for me, that I look at a data file, so I thought this would be useful to others as well.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "files.associations": {
+    "*.db": "ndjson",
+    "*.jsonl": "ndjson",
+    "*.njson": "ndjson"
+  },
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -7,6 +7,7 @@
     "yzhang.markdown-all-in-one",
     "DavidAnson.vscode-markdownlint",
     "dbaeumer.vscode-eslint",
+    "adrieankhisbe.vscode-ndjson",
   ],
   // List of extensions recommended by VS Code that should not be recommended for users of this workspace.
   "unwantedRecommendations": [


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "files.associations": {
+    "*.db": "ndjson",
+    "*.jsonl": "ndjson",
```
dimitropoulos, line 4: in case anyone else is/was curious https://github.com/ndjson/ndjson.github.io/issues/1#issuecomment-109935996 shows the relationship between JSON Lines and ndjson..  lol... what's that xkcd for this, again?

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "files.associations": {
```
dimitropoulos, line 2: this is a copy-pasta from what's in the docs for the plugin https://github.com/AdrieanKhisbe/vscode-ndjson

Arguably, the `.db` entry is the only one we need, but I figured it can't hurt to keep the other ones, too.

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
...
+    "*.db": "ndjson",
+    "*.jsonl": "ndjson",
+    "*.njson": "ndjson"
+  },
+}
```
johnwchadwick, line 7: nit: Missing newline.

(Could potentially avoid this in the future with `"files.insertFinalNewline": true`!)

develohpanda (approved):

.vscode/settings.json
dimitropoulos: see https://github.com/Kong/insomnia/pull/4057

PR #4046, opened 2021-09-23
change mocha timeout to be 1 minute

dimitropoulos:
closes INS-898

> 5s is not reasonable for a network test and often causes timeouts. Use 1 minute.


develohpanda (approved):
image

Works as expected, timing out after 60s

PR #4047, opened 2021-09-23
updates spectral due to CVE

dimitropoulos:
closes INS-975

immer 8.0.4, used by spectral 5.9.2 has CVE-2021-23436.  this is resolved in spectral 6.0.0.  However, spectral will not publish this version to NPM because they split the cli and core package https://github.com/stoplightio/spectral/issues/1832.  Since we're not using the cli (we're just using the TypeScript), we can just use the core package.

Beyond that, there were changes (https://github.com/stoplightio/spectral/issues/1799) that needed to be made because the `isOpenAPIv2` and `isOpenAPIv3` functions are no longer accessible (and are instead now built in).

The part I'm least sure about is what changed regarding loading rulesets.  It appears from the above to be picking up the ruleset automatically from the format?? Not feeling totally sure about that part - but let's see what the CLI thinks (assuming, like any CI run, that the present tests will fail if there's a problem with this area).

### DIFF
--- a/packages/insomnia-app/app/common/spectral.ts
+++ b/packages/insomnia-app/app/common/spectral.ts
@@ -1,12 +1,5 @@
-import { IRuleResult, isOpenApiv2, isOpenApiv3, Spectral } from '@stoplight/spectral';
+import { IRuleResult, Spectral } from '@stoplight/spectral-core';
 
-export const initializeSpectral = () => {
-  const spectral = new Spectral();
-  spectral.registerFormat('oas2', isOpenApiv2);
-  spectral.registerFormat('oas3', isOpenApiv3);
-  spectral.loadRuleset('spectral:oas');
-
-  return spectral;
-};
+export const initializeSpectral = () => new Spectral();
 
 export const isLintError = (result: IRuleResult) => result.severity === 0;
--- a/packages/insomnia-app/package-lock.json
+++ b/packages/insomnia-app/package-lock.json
@@ -1577,27 +1577,6 @@
 				"@sinonjs/commons": "^1.7.0"
 			}
 		},
-		"@stoplight/better-ajv-errors": {
-			"version": "0.0.4",
-			"resolved": "https://registry.npmjs.org/@stoplight/better-ajv-errors/-/better-ajv-errors-0.0.4.tgz",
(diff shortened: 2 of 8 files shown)


dimitropoulos:
I'm not sure why the CI is failing:

```
insomnia-inso: FAIL src/commands/lint-specification.test.ts
insomnia-inso:   ● Test suite failed to run
insomnia-inso:     Cannot find module 'nimma/legacy' from 'node_modules/@stoplight/spectral-core/dist/runner/runner.js'
insomnia-inso:     Require stack:
insomnia-inso:       node_modules/@stoplight/spectral-core/dist/runner/runner.js
insomnia-inso:       node_modules/@stoplight/spectral-core/dist/runner/index.js
insomnia-inso:       node_modules/@stoplight/spectral-core/dist/spectral.js
insomnia-inso:       node_modules/@stoplight/spectral-core/dist/index.js
insomnia-inso:       src/commands/lint-specification.ts
insomnia-inso:       src/commands/lint-specification.test.ts
insomnia-inso:       at Resolver.resolveModule (../../node_modules/jest-resolve/build/index.js:306:11)
insomnia-inso:       at Object.<anonymous> (node_modules/@stoplight/spectral-core/src/runner/runner.ts:9:1)
```

I did find https://meta.stoplight.io/docs/spectral/ZG9jOjg2MDIwMDM-spectral-v5-to-v6-migration-guide, and I suspect there may be we need to do regarding rulesets, but the documentation is all over the place and doesn't match the source code (at least with respect to how to configure the default oas rulesets).

I tried installing `'@stoplight/spectral-rulesets'` but that didn't seem to work when I pass it's `oas` export to `spectral.setRuleset`.

develohpanda:
Something we need to heavily consider is this conversation (you allude to it above but just adding a reference): https://kongstrong.slack.com/archives/CLDMTEH5M/p1631616930172600?thread_ts=1631614733.171100&channel=CLDMTEH5M&message_ts=1631616930.172600

dimitropoulos:
for those without access to kong's internal slack - what was discussed above was just that spectral adds / changes rule (defaults) in minor releases (not to mention major).  there's a good argument to be made that any such change is a breaking change for insomnia's users - even if spectral doesn't think so.

I mean... you shouldn't do a minor insomnia upgrade and then suddenly have failing tests that weren't failing before.

--

We can get around this by simply making our own config.  We might.... we might wanna do that.  That way we're never going to hit this problem, and not to mention things will be more explicit.  That'll have to be a separate PR that blocks this one, I think.

dimitropoulos:
we're gonna go with https://github.com/Kong/insomnia/pull/4054 instead, for now

PR #2891, opened 2020-12-02
Allow switching sub-environment inside Manage Environment modal

cobwebsonsale:
Closes #2761

- Added a button next to the sub-environment name to activate it
- Added a filled button next to the current active sub-environment
- Added helpful text in tooltips
- Renamed occurrences of `activeEnvironment` to `selectedEnvironment` in `WorkspaceEnvironmentsEditModal` (to avoid confusion) -- separated in another commit, for easier review/revert.

Caveat:
Could not re-render the currently selected environment in the right-pane after activation, so showing the activated environment instead. The re-render only makes sense when Base Environment is selected, as a different active environment changes the validations on variables. Tried:
- calling `_load`, 
- adding `activeEnvironmentId` to `state` inside `WorkspaceEnvironmentsEditModal` and setting state, and
- deriving `activeEnvironmentId` from `activeWorkspaceMeta` in `Wrapper`, as that is what gets updated in `handleChangeEnvironment`

Demo:
switch-environment

PR #4048, opened 2021-09-23
Environments order

### LINKED ISSUE #3822: Sub Environments are being re-ordered after leaving environment screen
**Describe the bug**
Sub Environments are being arbitrarily re-ordered after leaving environment screen

**To Reproduce**
1 - Open managed envs
2 - Order multiple sub environments in desired order
3 - close screen
4 - Envs are now moved to a seemingly arbitrary order.

**Expected behavior**
Order is persisted when changed

**Screenshots**

Initial Order
Screen Shot 2021-07-16 at 9 00 19 AM

Re-ordered to desired order

Screen Shot 2021-07-16 at 9 00 39 AM

Clicked 'done' and viewed envs, this was order

Screen Shot 2021-07-16 at 9 00 51 AM

 - OS: [e.g. macOS 10.15.1]
 - Installation Method: [e.g. homebrew]
 - App Version [e.g. v7.0.3]

**Additional context**

### LINKED ISSUE #3922: MetaSortKey of imsomnia export has wrong value
**Describe the bug**
When exporting my Insomnia Workspace, a enviroment gets a wrong MetaSortKey value (1).

**To Reproduce**
I have no idea how to reproduce, but will provide more info of someone asks for it. 

**Expected behavior**
I would expect the base enviroment to have a lower metaSortKey than any other enviroment.

My custom enviroment does have 1 as MetaSortKey, but my base enviroment has 1625240243045 (same es created value)

 - OS: Windows 10 Pro 20H2 
 - OS-Build: 19042.1165
 - Installation Method: windows installer
 - App Version 3032.4.1

**Additional context**
I too have an issue to reorder the enviroments in the envirement overview popup (it doesn't work, drag and drop reordering works, but when I select an enviroment the reordering is reverted)

vincendep:
Closes #3922
Closes #3822
Closes INS-954


dimitropoulos (approved):
Thanks! I tested using the steps in https://github.com/Kong/insomnia/issues/3822 and can confirm that this PR fixes what's described in those steps.

Here's what I tested with:
Screenshot_20211011_154712

vincendep:
Hi @dimitropoulos. I also include a change to address issue #3922, but I was not sure if it was a good idea since there were no feedback from you under that

dimitropoulos:
@vincendep is it something different than this fix?  if so, it'd be great if you knew the reproduction steps (and would open a new PR for that issue so we can handle separately).

vincendep:
It is quite related, since I had to update my previous changes.

packages/insomnia-app/app/models/environment.ts
```diff
@@ -76,6 +76,7 @@ export async function getOrCreateForParentId(parentId: string) {
       // Deterministic base env ID. It helps reduce sync complexity since we won't have to
       // de-duplicate environments.
       _id: `${prefix}_${crypto.createHash('sha1').update(parentId).digest('hex')}`,
+      metaSortKey: 0,
```
develohpanda: I'm just curious on what purpose this serves specifically? It's worth writing a comment here to explain why the `metaSortKey` for the base environment should be hard-coded to 0.

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.tsx
```diff
@@ -362,9 +362,9 @@ class WorkspaceEnvironmentsEditModal extends PureComponent<Props, State> {
     // Do this last so we don't block the sorting
     db.bufferChanges();
 
-    Promise.all(newSubEnvironments.map(environment => this._updateEnvironment(
+    Promise.all(newSubEnvironments.map((environment, idx) => this._updateEnvironment(
       environment,
-      { metaSortKey: 1 },
+      { metaSortKey: idx + 1 },
```
develohpanda: I think setting this to just `idx` should do the job. Looking at where this regression was introduced, that was the existing behavior: https://github.com/Kong/insomnia/pull/2891/files?authenticity_token=NotWiXgjksFGynH9ySdJfzuPFfvny1ipT861Ia0PCdxxQciOP08il4GzvolDtfnjCbgSLI7MzW3fm3gge3EmMQ%3D%3D&file-filters%5B%5D=.tsx#diff-83bf0bed6b835cb203957ba0b8b615b662856919617cab840b6ced94a9d02708L337-R375. During the refactor, `metaSortKey: i` was mistakenly changed to `metaSortKey: 1`.

https://github.com/Kong/insomnia/blob/ce9637e6e2f511605be6ca8bd84e5fbe0641c19a/packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.tsx#L337-L346

Is the `+ 1` strictly necessary?

develohpanda (commented):
On testing this seems to work well! A couple of notes however, would love your thoughts!

vincendep:
Hi @develohpanda, I introduced that changes after looking at issue #3922. It is to keep the base environment metaSortKey always lower than the sub environment ones.

develohpanda:
> Hi @develohpanda, I introduced that changes after looking at issue #3922. It is to keep the base environment metaSortKey always lower than the sub environment ones.

Gotcha, thanks! I think we should remove the hard-coded value there and stick with what was there before, for the base environment. The `metaSortKey` for the base environment holds no significance, so while it's likely harmless to change it it doesn't feel like there's a compelling reason to change it.

The extra note in #3922 mentions the re-ordering issue, which is fixed by this PR, but that's not directly related to the base environment `metaSortKey`, only the sub environment ones.

How does that sound?

vincendep:
It sounds good to me, indeed I was not sure of the utility to have the metaSotKey of the base environment lower than others. So I'll remove the hard coded one and set back the others to index instead of index plus 1

develohpanda (approved):

gatzjames (approved):
Good one! LGTM!

PR #4053, opened 2021-09-28
Use forwardRef for the Pane components

gatzjames:
Found a regression after changing class components to function due to refs not being forwarded.
The issue is that the panes and sidebar will not resize because the ref's are not registered.

Before:
resize-before

After:
resize-after

### DIFF
--- a/packages/insomnia-app/app/ui/containers/app.tsx
+++ b/packages/insomnia-app/app/ui/containers/app.tsx
@@ -1018,7 +1018,7 @@ class App extends PureComponent<AppProps, State> {
         });
       }
 
-      const sidebar = this._sidebarRef.current
+      const sidebar = this._sidebarRef.current;
 
       if (sidebar) {
         const currentPixelWidth = sidebar.offsetWidth;
--- a/packages/insomnia-app/app/ui/containers/app.tsx
+++ b/packages/insomnia-app/app/ui/containers/app.tsx
@@ -4,8 +4,7 @@ import fs from 'fs';
 import HTTPSnippet from 'httpsnippet';
 import * as mime from 'mime-types';
 import * as path from 'path';
-import React, { PureComponent, RefObject } from 'react';
-import ReactDOM from 'react-dom';
+import React, { createRef, PureComponent } from 'react';
 import { connect } from 'react-redux';
 import { Action, bindActionCreators, Dispatch } from 'redux';
 import { parse as urlParse } from 'url';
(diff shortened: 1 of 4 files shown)


packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1034,20 +1018,21 @@ class App extends PureComponent<AppProps, State> {
...
-      const currentPixelWidth = sidebar.offsetWidth;
-      // @ts-expect-error -- TSCONVERSION
-      const ratio = (e.clientX - sidebar.offsetLeft) / currentPixelWidth;
-      const width = this.state.sidebarWidth * ratio;
-      let sidebarWidth = Math.min(width, MAX_SIDEBAR_REMS);
-
-      if (sidebarWidth < COLLAPSE_SIDEBAR_REMS) {
-        sidebarWidth = MIN_SIDEBAR_REMS;
-      }
+      const sidebar = this._sidebarRef.current
```
dimitropoulos:
```suggestion
      const sidebar = this._sidebarRef.current;
```

dimitropoulos:
it's interesting to me, though, that the sidebar one is _not_ broken for me on develop (d53ef39c9b76aefcfa71630078d838bb95eaad56) but it was for you...

dimitropoulos (approved):
works great for me and nice to see the other on-topic fixes (i.e. the removal of ts-ignores)

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -978,20 +965,19 @@ class App extends PureComponent<AppProps, State> {
...
-      const requestPaneWidth = requestPane.offsetWidth;
-      // @ts-expect-error -- TSCONVERSION
-      const responsePaneWidth = responsePane.offsetWidth;
-      // @ts-expect-error -- TSCONVERSION
-      const pixelOffset = e.clientX - requestPane.offsetLeft;
-      let paneWidth = pixelOffset / (requestPaneWidth + responsePaneWidth);
-      paneWidth = Math.min(Math.max(paneWidth, MIN_PANE_WIDTH), MAX_PANE_WIDTH);
-
-      this._handleSetPaneWidth(paneWidth);
+      const requestPane = this._requestPaneRef.current;
```
develohpanda, line 968: Here's a thought (definitely not for this PR): is it possible to move the refs and this drag handling directly into `PageLayout`? I don't _think_ this strictly needs to be in `App`, it was just never moved out. That way we can remove the sending of refs down as we are, that might be a remnant

develohpanda (approved):
Thanks for fixing this up, works for me :+1:.

Once thing that is worth doing (it'll be tedious, but the PR to convert to FC was also large) is to check we don't have other instances where refs are being lost. Modals, dropdowns, code editors, etc, all rely on refs to an extent. These are just examples and I don't know if they were updated recently, but again, it's worth double checking this.

packages/insomnia-app/app/ui/containers/app.tsx
gatzjames: Agree I'll make a follow up to this to clean it up then!

PR #4049, opened 2021-09-24
Create single inso executable using pkg

develohpanda:
### GitHub Actions workflow changes

The release workflow has been updated to also create a GitHub release and publish Inso CLI binary artifacts into the release. The Insomnia app release is created against the `core@2021.2.3` tag, while the Inso CLI release is created agsinst the `lib@1.2.3` tag.

This is because Inso CLI is also published to NPM, using the `lib@1.2.3` tag generated from Lerna, and this way both distributions (what's available in NPM and what's available in GitHub Releases) for Inso CLI remain in sync.

The **Release** workflow will now:
* Run all tests (excluding smoke tests)
* Publish to NPM
* Assert if a release already exists against a `lib@x.y.z` tag.
	* Exists: skip creating and publishing Inso CLI binaries
	* Doesn't exist:
		* Create release
		* Create Inso CLI binaries on macOS, Windows and Ubuntu
		* Run CLI smoke tests
		* Publish the compressed binaries into the GitHub Release
			* `windows.zip`
			* `macos.tar`
			* `linux.tar.xz`
* Build Insomnia app for macOS, Windows, Ubuntu
* Publish Insomnia app artifacts into GitHub Release

The **Test/PR** workflow will now:
* Create Inso CLI binaries for macOS, Windows and Ubuntu
* Publish the compressed binaries into the CI artifacts (retained for ~5 days)
* Run CLI smoke tests

### Code changes

* Plugins loading now has a try-catch. In case a plugin fails to load, it will not crash anything and will print to console.
* `jsonpath` dependency updated to `jsonpath-plus`
* Moved `deep-equal` to a `devDependency`

### Smoke tests

* Updated `inso-nedb` fixture
* Added CLI smoke tests for more commands (`inso run test`, `inso generate config`, `inso export spec`, `inso lint spec`)
* CLI smoke tests also run with the compiled binary

### Plugins

* Updated `insomnia-plugin-jsonpath` to use `jsonpath-plus` instead of `jsonpath`
* Updated `insomnia-plugin-os` to use `jsonpath-plus` instead of `jsonpath`
* Updated `insomnia-plugin-response` to use `jsonpath-plus` instead of `jsonpath` (this impacts request chaining)

### Local dev

```
npm run clean
npm run bootstrap
npm run inso-package
npm run test:smoke:cli
```

PR #4056, opened 2021-09-29
Move jsonpath plugin to TS

develohpanda:
Initially I did this as part of https://github.com/Kong/insomnia/pull/4049, this PR just extracts it out.

We can drop webpack from it, it was added as part of experimentation, but it didn't seem worth to throw away this code.


plugins/insomnia-plugin-jsonpath/package.json
```diff
@@ -12,15 +12,31 @@
...
+    "lint:fix": "npm run lint -- --fix",
+    "clean": "tsc --build tsconfig.build.json --clean",
+    "postclean": "rimraf dist",
+    "build": "webpack --config webpack.config.js --display errors-only",
+    "watch": "npm run build -- --watch",
+    "prepublish": "npm run build"
+  },
+  "devDependencies": {
+    "@types/webpack": "^4.41.26",
+    "jsonpath-plus": "^6.0.1",
```
develohpanda, line 35: Should be a regular dependency, all others can be devDeps

dimitropoulos: resolved in 4c09f19c59f4b396facbdd0f391640b5841f0dc5

plugins/insomnia-plugin-jsonpath/tsconfig.build.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "dist",
+    "rootDir": "src",
+    "types": ["node"]
```
dimitropoulos, line 6: https://github.com/Kong/insomnia/blob/develop/tsconfig.base.json#L22 the base already includes this so I'm assuming you don't need to also specify it here?

plugins/insomnia-plugin-jsonpath/tsconfig.build.json
```diff
@@ -0,0 +1,16 @@
...
+  },
+  "include": [
+    "src"
+  ],
+  "exclude": [
+    "**/fixtures",
+    "**/test-helpers",
+    "**/*.test.ts"
+  ]
+}
```
dimitropoulos:
```suggestion
}

```

plugins/insomnia-plugin-jsonpath/tsconfig.build.json
develohpanda: Ya likely, I just copy-pasta'd from `insomnia-testing` IIRC

dimitropoulos: ahh, ok.  that makes sense.  `insomnia-testing` _needs_ to do this because of the conflict with globals in the mocha and jest types.

PR #4058, opened 2021-09-29
fix(import): Importing from the projects view should respect the forc…

### LINKED ISSUE #4034: Postman collection import successful, but not visible on any dashboard

### Insomnia Version

2021.5.2

macOS

### Operating System Version

macOS Big Sur 11.5.2

### Installation method

homebrew

none

### Expected Behavior

Collection to show up on dashboard after successful import.

### Actual Behavior

No collection is visible or accessible anywhere in Insomnia.

### Reproduction Steps

1. Open Insomnia
2. Under create, choose IMPORT FROM => "URL"
3. Fill in
    ```
    https://fivetran.com/assets-docs/postman/rest_api_collection.json
    ```
4. Press Fetch and Import
5.  "Import Succeeded" message appears with "You imported 6 Folders, 35 Requests!"
6. PROBEM: No new collections..

### Additional Information

I've tried this with Paw and Postman, which imports the collection without issues.
Can also be reproduced with Insomnia native collection format https://git.io/github-v3-rest-api-for-insomnia

Exact steps in screenshots below
CleanShot 2021-09-16 at 15 39 49@2x
CleanShot 2021-09-16 at 15 39 53@2x
CleanShot 2021-09-16 at 15 39 58@2x
CleanShot 2021-09-16 at 15 40 03@2x

gatzjames:
This PR fixes the issue where if we import curl/postman data in a new workspace the data do not appear in the app.

It appears this happens in the import logic:
Some importers currently return data that do not contain a workspace for example the curl importer will return a request object.
This is expected from the importer, but we currently do not create a new workspace and it results in the data being added in the database without a parent workspace reference.

In case that the imported data do not contain a workspace we now do the following:
From the Project view we ask the user if they want to import their data in an existing/new workspace.
From the Workspace view we import the data to the current workspace.

#### Import to an empty project:
import-to-empty-project

#### Import to existing workspace:
project-view import-to-existing

#### Import to new workspace:
project-view import-to-new-workspace

#### Import to the current workspace:
workspace-view import-to-current

#### Import an existing workspace (e.g. insomnia exported project)
workspace-view import-existing-workspace

- [x] Added smoke tests for the import flow
- [x] Added integration tests for the import util and the database

Closes #4034
Closes INS-906


wdawson:
These experiences seem flipped to me. 

When importing a curl request from the dashboard, it would be helpful to have a choice about importing to an existing workspace or a new one. The app should navigate to that workspace with that request selected as well.

When importing from within an existing workspace, I don't think we need to give the user a choice. It's clear from the context (IMO) that you're importing into the current workspace. The app should navigate the user to the newly imported request.

Anything I'm missing?

gatzjames:
@wdawson great feedback 💯

> When importing a curl request from the dashboard, it would be helpful to have a choice about importing to an existing workspace or a new one.

As a user would you expect to select from the active project's workspaces or would you expect a way to also pick the project?

> The app should navigate to that workspace with that request selected as well.
> When importing from within an existing workspace, I don't think we need to give the user a choice. It's clear from the context (IMO) that you're importing into the current workspace. The app should navigate the user to the newly imported request.

Sounds like a great UX improvement!

wdawson:
@gatzjames

> As a user would you expect to select from the active project's workspaces or would you expect a way to also pick the project?

I don't think we need a project picker too. Especially with some new designs from Erick that should give better context about the current project you're in.

packages/insomnia-app/app/common/import.ts
```diff
@@ -77,8 +77,7 @@ export async function importUri(uri: string, importConfig: ImportRawConfig) {
   if (error) {
     showError({
       title: 'Failed to import',
-      // @ts-expect-error -- TSCONVERSION appears to be a genuine error
-      error: error.message,
+      error: error,
```
dimitropoulos:
```suggestion
      error,
```
I suppose maybe now is the time to introduce https://eslint.org/docs/rules/object-shorthand

dimitropoulos: (well, not this PR, I just mean, like, some time these days)

packages/insomnia-app/app/common/__tests__/import.test.ts
```diff
@@ -12,12 +21,204 @@ describe('isApiSpecImport()', () => {
...
+      body: {
+        'text': '{\"email_id\": \"tem_123\"}',
+      },
+    });
+  });
+
+  it('should import a postman collection to a new workspace', async () => {
+    const rawFixture = fs
+      .readFileSync(
+        path.join(__dirname, '..', '__fixtures__', 'postman', 'aws-signature-auth-v2_0-input.json'),
```
dimitropoulos: it'd be nice to pull these `path.join`s (in all of these added tests) out into their own variables, that way, also, we won't have to break to multiple lines

packages/insomnia-app/app/common/import.ts
```diff
@@ -77,8 +77,7 @@ export async function importUri(uri: string, importConfig: ImportRawConfig) {
   if (error) {
     showError({
       title: 'Failed to import',
-      // @ts-expect-error -- TSCONVERSION appears to be a genuine error
-      error: error.message,
+      error,
```
dimitropoulos, line 81: thanks

packages/insomnia-app/app/ui/components/buttons/settings-button.tsx
```diff
@@ -24,7 +24,7 @@ const SettingsButton: FunctionComponent<Props> = ({ className, settings }) => (
       </>
     }
   >
-    <CircleButton className={className} onClick={showSettingsModal}>
+    <CircleButton data-testid="settings-button" className={className} onClick={showSettingsModal}>
```
dimitropoulos: I'm a big fan of this approach.  I hope we use it more.  It was strange at first, when I first saw this project, that we're using `react$` and css selectors.  Since then, I've had so so so many problems with `react$` and css selectors for tests that it's confirmed for me that regular ol' `data-testid` is still the best way to go

packages/insomnia-smoke-test/designer/app.test.ts
```diff
@@ -30,7 +30,7 @@ xdescribe('Application launch', function() {
 
     // Install plugin
     await settings.openWithKeyboardShortcut(app);
-    await settings.goToPlugins(app);
+    await settings.goToPluginsTab(app);
```
dimitropoulos, line 33: thank you

dimitropoulos (approved):
Thank you for the vigorous testing!

I functionally tested every case in the description (all worked well!) and aside from a small (non-blocking) suggestion, code looks good too.

packages/insomnia-app/app/common/__tests__/import.test.ts
```diff
@@ -1,9 +1,18 @@
...
 describe('isApiSpecImport()', () => {
-  it.each(['swagger2', 'openapi3'])('should return true if spec id is %o', (id: string) => {
-    expect(importUtil.isApiSpecImport({ id })).toBe(true);
-  });
+  it.each(['swagger2', 'openapi3'])(
+    'should return true if spec id is %o',
+    (id: string) => {
+      expect(importUtil.isApiSpecImport({ id })).toBe(true);
+    }
+  );
```
develohpanda, line 15: What do you have enabled in your dev environment causing these format changes? I'm not opposed to it at all, but I'd much prefer to add that config into the repo, and force all files to conform in a single PR, so that it doesn't increase the changeset in future PRs and everyone follows the same config.

Are you using prettier, and do we need to revive the prettier conversation of yore (post-release, of course)?

packages/insomnia-app/app/common/import.ts
```diff
@@ -171,6 +170,15 @@ export async function importRaw(
     importedDocs[model.type] = [];
   }
 
+  // Add a workspace to the resources if it doesn't exist
+  if (!data.resources.some(resource => resource._type === EXPORT_TYPE_WORKSPACE)) {
+    data.resources.push({
```
develohpanda, line 180: See https://github.com/gatzjames/insomnia/blob/72d70375a7347c3e366f846619fa50a89fdfdefc/packages/insomnia-app/app/common/import.ts#L166 (just a few lines above)

My question is, should the workspace stub be added to the start or end of the array? Once the answer is known, we should add into a comment explicitly stating this is added to the start or end.

packages/insomnia-app/app/common/import.ts
```diff
@@ -171,6 +170,15 @@ export async function importRaw(
...
+      ...models.workspace.init(),
+      _id: '__WORKSPACE_ID__',
+      _type: 'workspace',
```
develohpanda:
```suggestion
      _id: WORKSPACE_ID_KEY,
      _type: EXPORT_TYPE_WORKSPACE,
```

Magic strings are no gooooodddd

packages/insomnia-app/app/common/import.ts
```diff
@@ -171,6 +170,15 @@ export async function importRaw(
     importedDocs[model.type] = [];
   }
 
+  // Add a workspace to the resources if it doesn't exist
+  if (!data.resources.some(resource => resource._type === EXPORT_TYPE_WORKSPACE)) {
+    data.resources.push({
+      ...models.workspace.init(),
```
develohpanda: The issue with this is the object pushed into `data.resources` will look like the following:

```js
{
  _id: '__WORKSPACE_ID__',
  _type: 'workspace', // expected by the import logic
  type: 'Workspace', // comes from the init
  // ...
}
```

There exists code which maps `model.type` (in the database) to `model._type` as expected by the export file. We could remove the `model.type` when creating this object to push into `data.resources` but it doesn't feel like a very reliably approach to me.

In addition, this will implicitly force the import to always be a [collection](https://github.com/Kong/insomnia/blob/7f7b9f56144e275a68a85abfe1c91f1081a9eb1c/packages/insomnia-app/app/models/workspace.ts#L50) and never give the option. I think this is (probably) fine, but we should explicitly set the scope to be collection, rather than depend on `models.workspace.init()`.

We should also override the name to be something like "imported collection" (but note the bug about this in the root comment)

packages/insomnia-app/app/ui/components/buttons/settings-button.tsx
develohpanda: I think maybe you haven't used the `react$` selector enough and attributed the smoke-test pains to this selector.

The `react$` and `css` selectors also serve an important purpose in that they help you find things with specific attributes, props/state.

Two examples:
- modal footer buttons (which are totally configurable in use): https://github.com/gatzjames/insomnia/blob/cceedcbf77569b109e28b0e667bc5e346629b317/packages/insomnia-smoke-test/modules/modal.ts#L28-L30
- a request with a particular name, using react props (or even state): https://github.com/Kong/insomnia/blob/107a37fb920277abd806582b4bc8c2bc956fc13d/packages/insomnia-smoke-test/modules/debug.ts#L69-L71

In many cases, `data-testid` is redundant, such as this case, because we _already_ have a label (in the component name). We're just aliasing something that already has a unique name. Did `react$('SettingsButton')` not work?

> Note: no need to change this in this PR. We're having the convo here but I'm not requesting this to be changed right now.

packages/insomnia-smoke-test/modules/settings.ts
```diff
@@ -1,40 +1,47 @@
...
 import { clickTabByText } from './tabs';
 
-export const openWithKeyboardShortcut = async app => {
+export const openFromSettingsButton = async (app: Application) => {
+  await (await app.client.$('[data-testid="settings-button"]')).click();
+
+  await modal.waitUntilOpened(app, { modalName: 'SettingsModal' });
+};
+
+export const openWithKeyboardShortcut = async (app: Application) => {
```
develohpanda, line 14: It would be good to use a mix of both of these in our tests. Covers more ground, smoke tests tend to be less surgical because they take longer to run IMO

> Can be addressed in a separate PR

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -1,18 +1,68 @@
...
+            title: 'Import',
+            message: 'Do you want to import into an existing workspace or a new one?',
+            yesText: 'Existing',
+            noText: 'New Workspace',
+            onDone: (yes: boolean) => {
+              resolve(yes);
+            },
+          });
+        });
+      }
```
develohpanda: This can exist outside this function as well, right? Alongside `askToSelectExistingWorkspace`?

> Can be addressed in a separate PR

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -1,18 +1,68 @@
 import { strings } from '../../../common/strings';
```
develohpanda, line 1: Thanks for not converting the stuff in here to arrow-functions and leaving that potentially to a separate PR

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -1,18 +1,68 @@
...
+      // Return null if there are no available workspaces to chose from.
+      if (activeProjectWorkspaces.length === 0) {
+        return null;
+      }
+
+      function askToImportIntoNewWorkspace() {
+        return new Promise(resolve => {
+          showModal(AskModal, {
+            title: 'Import',
+            message: 'Do you want to import into an existing workspace or a new one?',
```
develohpanda: Need to use `${strings.workspace.singular.toLowerCase()}`

Also wording doesn't quite map well with the buttons:
- "Existing workspace" -> "Existing"
- "New" -> "New Workspace"

I think it's reasonable for the buttons to be "existing" and "new"

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -1,18 +1,68 @@
...
   new = 'new',
-  current = 'current'
+  current = 'current',
+  existing = 'existing'
 }
 
+export type SelectExistingWorkspacePrompt = Promise<string | null>;
+export function askToSelectExistingWorkspace(workspaces: Workspace[]): SelectExistingWorkspacePrompt {
+  return new Promise(resolve => {
+    const options = workspaces.map(workspace => ({ name: workspace.name, value: workspace._id }));
```
develohpanda, line 17: We should be using `workspace.name` for collections, and `apiSpec.fileName` for design documents, the way this selector works.

https://github.com/Kong/insomnia/blob/e793e6f166b7c7a62ec0867a535d6eb7ebf44d97/packages/insomnia-app/app/ui/redux/selectors.ts#L158

https://www.loom.com/share/e3ee3e14e4e143748c40b3b426d6e794

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -22,8 +72,7 @@ export function askToImportIntoWorkspace({ workspaceId, forceToWorkspace }: { wo
         return null;
 
       case ForceToWorkspace.current:
-        return workspaceId;
-
+        return workspaceId ?? null;
```
develohpanda: Why is this fallback necessary?

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -22,8 +72,7 @@ export function askToImportIntoWorkspace({ workspaceId, forceToWorkspace }: { wo
...
       default:
```
develohpanda: We should add `ForceToWorkspace.existing` as a switch-case, and mark it as unreachable (because of the if-condition above)

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      await onboarding.skipOnboardingFlow(app);
+
+      await home.documentListingShown(app);
+      await home.expectTotalDocuments(app, 0);
+
+      const buffer = await fs.promises.readFile(
+        `${__dirname}/../fixtures/swagger2.yaml`
+      );
+      const swagger2Text = buffer.toString();
+      await app.electron.clipboard.writeText(swagger2Text);
```
develohpanda: Please create a helper in one of the modules (or a new module) called, `readFileContentsIntoClipboard()`.

https://github.com/Kong/insomnia/tree/develop/packages/insomnia-smoke-test#readability

> Can be addressed in a separate PR

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -0,0 +1,204 @@
...
+import * as debug from '../modules/debug';
+import * as home from '../modules/home';
+import * as modal from '../modules/modal';
+import * as onboarding from '../modules/onboarding';
+import * as settings from '../modules/settings';
+
+describe('Import', function() {
+  jest.setTimeout(50000);
+
+  describe('from the Project View', () => {
```
develohpanda:
```suggestion
  describe('from the Dashboard', () => {
```

> Can be addressed in a separate PR

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      await home.expectTotalDocuments(app, 1);
+
+      // Import the swagger spec a second time
+      await home.importFromClipboard(app);
+
+      await home.expectTotalDocuments(app, 1);
+    });
+  });
+
+  describe('from the Settings View', () => {
```
develohpanda:
```suggestion
  describe('from Preferences', () => {
```

> Can be addressed in a separate PR

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -0,0 +1,204 @@
...
+    let app: Application;
+
+    beforeEach(async () => {
+      app = await launchApp();
+    });
+
+    afterEach(async () => {
+      await stop(app);
+    });
```
develohpanda: This can exist in the root `describe` block

> Can be addressed in a separate PR

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -0,0 +1,204 @@
...
+      await debug.clickSendRequest(app);
+
+      // Ensure 200
+      await debug.expect200(app);
+
+      await debug.goToDashboard(app);
+      await home.expectTotalDocuments(app, 1);
+    });
+
+    it('should update the existing workspace (e.g. Insomnia Exports)', async () => {
```
develohpanda: This test is lacking some assertions I think.

The `insomnia4.yaml` fixture only has a workspace. Ideally it should also have a request, so you can validate that the request is imported into the expected workspace. (This particular line of feedback applies to most of the smoke tests in this file.)

To double check, is the following behavior what you expect? My clipboard contains the contents in `/packages/insomnia-inso/src/db/fixtures/insomnia-v4/insomnia_v4.json`.

https://www.loom.com/share/1dbf1722a1dd428d86c4120fa51e3aeb

If this is the expected behavior you're trying to test, then the assertions at the end should check that the request exists in the expected workspace. If this is not what you're trying to test, there's different (existing) behavior causing this which I can talk about/describe.

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    beforeEach(async () => {
+      app = await launchApp();
+    });
+
+    afterEach(async () => {
+      await stop(app);
+    });
+
+    it('should directly import to the active workspace', async () => {
```
develohpanda, line 189: This test is specifically validating that swagger2 imports successfully

packages/insomnia-app/app/common/__tests__/import.test.ts
```diff
@@ -12,12 +21,204 @@ describe('isApiSpecImport()', () => {
...
+    expect(workspacesCount).toBe(1);
+
+    expect(curlRequests[0]).toMatchObject({
+      body: {
+        'text': '{\"email_id\": \"tem_123\"}',
+      },
+    });
+  });
+
+  it('should import a curl request to an existing workspace', async () => {
```
develohpanda, line 71: I'm a bit confused, this test looks identical to the one above?

develohpanda (changes requested):
Largely this looks good. There are some quirks I've found, identified below and in the comments.

The main issue, where Postman & cURL would fail to import from the dashboard, are both fixed. 🎉 

## Exploratory QA (just the failing cases)

<details>
<summary>🔴 Importing into the existing workspace renames that workspace</summary>

https://www.loom.com/share/f314af9460a34432a6ad18c5d312e2cd

</details>

<details>
<summary>🔴 Selector for importing into a workspace has the wrong name</summary>

https://www.loom.com/share/e3ee3e14e4e143748c40b3b426d6e794

</details>

Below the comments that can be addressed in a separate PR, I have mentioned as such. Those that don't have the message are requested changes in _this_ PR, in addition to resolving the bugs above.

packages/insomnia-app/app/ui/components/buttons/settings-button.tsx
gatzjames: I'd also prefer to use what we already have tbh.
The issue was that the `.react$(...)` selector would fail randomly where the `[data-testid=...]` seems to be more stable.

On a side note I do find that relying on implementation details (react selectors, component names, knowing what props have been passed to a function etc) is brittle and shouldn't happen on e2e tests. Ideally we should have selectors that check dom aria attributes like role or selecting by text. Changing a component name shouldn't affect the e2e tests imo.

develohpanda: That seems reasonable. When we have some downtime let's talk about that. We'll look through all the existing areas where the react selector is used, and see how we can best translate it to attributes

packages/insomnia-app/app/common/__tests__/import.test.ts
gatzjames: 😅 Not sure how this happened since I have autoformatting off.

+1 for prettier though

packages/insomnia-app/app/common/__tests__/import.test.ts
gatzjames: Correct! Nice catch!
It should test what happens when we add the curl request to an existing workspace. Added in https://github.com/Kong/insomnia/commit/1b9e9d8fa27261a646e65d24755ca02d3fcf4e5b

packages/insomnia-app/app/common/__tests__/import.test.ts
gatzjames: Good one!
Added in https://github.com/Kong/insomnia/commit/8dac8cd4066fc1924afd804c61ae49e62339b504

packages/insomnia-app/app/common/__tests__/import.test.ts
gatzjames: Reverted the format in https://github.com/Kong/insomnia/commit/8dac8cd4066fc1924afd804c61ae49e62339b504

packages/insomnia-app/app/common/import.ts
gatzjames: Afaik there isn't a clear contract with importers about resource order but I tested with all fixture types and found the convention.

Added a note in https://github.com/Kong/insomnia/commit/09007b05837f83f98da45feb4adec57dbfe58721

packages/insomnia-app/app/ui/redux/modules/helpers.ts
gatzjames: Good one!
Added a selector that returns the workspaces with the name resolved by scope in https://github.com/Kong/insomnia/commit/965486ca73602d0fd6b1fb7f685f101469c3c625

packages/insomnia-app/app/common/import.ts
gatzjames: Nice one!
Removing this also solves the issue that renames a workspace when you import!
Changed in https://github.com/Kong/insomnia/commit/ce4969eb9cc0e66b2b192e7b4be9171df83b9e04

packages/insomnia-smoke-test/core/import.test.ts
gatzjames: 👍 Moved it out in https://github.com/Kong/insomnia/commit/3e921770dca97e840bcce3919f7562222601c9d3

packages/insomnia-smoke-test/core/import.test.ts
gatzjames: I updated the title of the test to make it a bit more clear.
I agree that sending the request or checking that it exists will increase the confidence of the imported data structures.
The reason I didn't test this here is that this test is mostly focusing on the behaviour of where the workspace will be imported depending on where you are on the app e.g. in this specific test if you are in a workspace view and try to import an `insomnia-export` it will import it in the active project as a new workspace instead of merging it with the current.
We already test the imported data structures in unit tests so not sure about the granularity we'd like to test with here 🤔

gatzjames: So I ended up extending the tests with clicking on folders/requests and sending them to cover more of the imported resources.
Changed in https://github.com/Kong/insomnia/commit/6e835230d1ba53b018e4320287a80569cdc439a6

develohpanda: Ok cool.  I think it's ok to skip the actual sending of requests, but I think it's important to have at least a request or something in the import file in addition to the workspace itself, to make sure the contents/children get imported properly. What you've got now is great :) thanks

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
```diff
@@ -140,4 +140,50 @@ describe('selectors', () => {
...
+        name: 'designWorkspace.name',
+        scope: WorkspaceScopeKeys.design,
+      });
+
+      const newApiSpec = await models.apiSpec.getOrCreateForParentId(
+        newWorkspace._id
+      );
+
+      // The database will update the api spec with the workspace name
+      // That's why we need to explicitly update the ApiSpec name
```
develohpanda, line 162: This makes me 🤦🏽 but also, it is what it is. Thanks for adding a comment outlining explicitly why

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -136,6 +136,27 @@ export const selectAllApiSpecs = createSelector(
...
 );
 
+export const selectWorkspacesWithResolvedNameForActiveProject = createSelector(
+  selectWorkspacesForActiveProject,
+  selectAllApiSpecs,
+  (workspaces, apiSpecs) => {
+    return workspaces.map(workspace => {
+      if (isCollection(workspace)) {
+        return workspace;
+      }
```
develohpanda, line 146: Nice early exit to prevent unnecessary searching over `apiSpecs` 🔥

packages/insomnia-app/app/common/import.ts
```diff
@@ -171,6 +174,7 @@ export async function importRaw(
   }
 
   // Add a workspace to the resources if it doesn't exist
+  // NOTE: The workspace should be the last item of the resources
```
develohpanda, line 178: 👍🏽 nice

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -47,34 +47,41 @@ async function askToImportIntoNewWorkspace(): Promise<boolean> {
 export type ImportToWorkspacePrompt = () => null | string | Promise<null | string>;
 export function askToImportIntoWorkspace({ workspaceId, forceToWorkspace, activeProjectWorkspaces }: { workspaceId?: string; forceToWorkspace?: ForceToWorkspace; activeProjectWorkspaces: Workspace[] }): ImportToWorkspacePrompt {
   return function() {
-    if (forceToWorkspace === ForceToWorkspace.existing) {
-      // Return null if there are no available workspaces to chose from.
-      if (activeProjectWorkspaces.length === 0) {
+    switch (forceToWorkspace) {
```
develohpanda, line 50: So much clearer to follow 🔥

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -47,34 +47,41 @@ async function askToImportIntoNewWorkspace(): Promise<boolean> {
...
+      case ForceToWorkspace.existing: {
+        // Return null if there are no available workspaces to chose from.
+        if (activeProjectWorkspaces.length === 0) {
+          return null;
+        }
+
+        return new Promise(async resolve => {
+          const yes = await askToImportIntoNewWorkspace();
+          if (yes) {
+            const workspaceId = await askToSelectExistingWorkspace(activeProjectWorkspaces);
```
develohpanda, line 72: Something here seems odd 🤔 in pseudo-code:

```
ask to import into new workspace
if yes
	ask to select existing workspace
	return selected workspace id
if no
	return null
```

Are the conditions/variable names flipped? We ask to import into a new workspace, and if `yes` we ask to select an existing workspace.

I know you didn't change this so we don't have to fix in this PR but definitely would like to follow-up on it

packages/insomnia-smoke-test/modules/application.ts
```diff
@@ -111,3 +111,7 @@ export const takeScreenshot = async (app, name) => {
   const buffer = await app.browserWindow.capturePage();
   await fs.promises.writeFile(path.join('screenshots', `${name}.png`), buffer);
 };
+
+export const writeTextToClipboard = async (app, text) => {
+  app.electron.clipboard.writeText(text);
```
develohpanda, line 116: I think we're missing an await here, and all invocations of `writeTextToClipoard` also need to be awaited (right?)

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -24,139 +25,209 @@ describe('Import', function() {
...
+      // Launch the app
       await client.correctlyLaunched(app);
       await onboarding.skipOnboardingFlow(app);
 
       await home.documentListingShown(app);
       await home.expectTotalDocuments(app, 0);
-      const swagger2Text = await getFixtureContent('swagger2.yaml');
+
+      // Import the fixture from the clipboard
+      const swagger2Text = await loadFixture('swagger2.yaml');
```
develohpanda, line 36: Nice, I like it! The combination of `loadFixture` and `writeTextToClipboard` is much nicer than my suggestion 🔥

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -24,139 +25,209 @@ describe('Import', function() {
...
 
+      // It should update the existing document
       await home.expectTotalDocuments(app, 1);
+
+      // Open the new document and send a request
+      await home.openDocumentWithTitle(app, 'Insomnia');
+
+      await debug.clickFolderByName(app, 'Actions');
+
+      await debug.clickRequestByName(app, 'Get Sleep');
```
develohpanda, line 180: 😴

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -165,15 +236,38 @@ describe('Import', function() {
...
+      // Import the updated fixture from the preferences panel
+      const insomnia4UpdatedText = await loadFixture('insomnia4-update.yaml');
+
+      writeTextToClipboard(app, insomnia4UpdatedText);
       await settings.openFromSettingsButton(app);
       await settings.goToDataTab(app);
       await settings.importFromClipboard(app);
 
+      // NOTE: The document was imported directly to the
+      // previously imported workspace and the user didn't see a dialog
```
develohpanda, line 253: 🔥

packages/insomnia-smoke-test/modules/design.ts
```diff
@@ -0,0 +1,13 @@
+import { Application } from 'spectron';
+
+export const pageDisplayed = async (app: Application) => {
+  await app.client.react$('WrapperDesign').then(e => e.waitForDisplayed());
+};
+
+export const goToActivity = async (app: Application, activity: 'spec' | 'debug' | 'test' = 'spec') => {
+  // NOTE: We shouldn't need to click the span.
+  // TODO: Make the radio button group in the header keyboard accessible.
+  const debugRadioInput = await app.client.$(`input[name="activity-toggle"][value=${activity}] ~ span`);
```
develohpanda, line 10: Woah, TIL about `~`. And totally agree on the activity selector component needing an overhaul.

packages/insomnia-smoke-test/modules/home.ts
```diff
@@ -1,8 +1,9 @@
 import faker from 'faker';
+import { Application } from 'spectron';
 
 import * as debug from './debug';
+import * as design from './design';
```
develohpanda, line 5: Thanks for favoring consistency 💯

packages/insomnia-smoke-test/modules/home.ts
```diff
@@ -14,30 +15,38 @@ export const expectDocumentWithTitle = async (app, title) => {
   await app.client.waitUntilTextExists('.document-listing__body', title);
 };
 
-export const findCardWithTitle = async (app, text) => {
-  let card;
-  await app.client.waitUntil(async () => {
```
develohpanda, line 19: The change you've made is much nicer and improves readability. I'm not sure if we should still add the `waitUntil` in...maybe this change is fine.

What the `waitUntil` does is, if after the first time the callback is run we don't have a successful result, it'll keep trying until we time out. This could be useful if the components aren't on the page at the very start but take time to load. Dropdowns sometimes behave like that which is what inspired this approach IIRC.

If removal of `waitUntil` here causes flakes lets keep this change in mind, but I do think your change is safe. and soooo much clearer.

packages/insomnia-smoke-test/modules/home.ts
```diff
@@ -14,30 +15,38 @@ export const expectDocumentWithTitle = async (app, title) => {
...
+  const cardBadge = await card.$('.header-item.card-badge');
+  const isCollection = await cardBadge.getText() === 'Collection';
   await card.waitForDisplayed();
   await card.click();
 
-  await debug.pageDisplayed(app);
+  if (isCollection) {
+    await debug.pageDisplayed(app);
+  } else {
+    await design.pageDisplayed(app);
```
develohpanda, line 49: Probably fine for now but remember this isn't always the case, we cache the last active activity for a workspace. For example:

1. create design document
2. click design document on dashboard
3. click "test" tab
4. go to dashboard
5. click design document on dashboard
6. notice you're on the "test" tab

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
```diff
@@ -168,8 +168,14 @@ describe('selectors', () => {
         activeWorkspaceId: null,
       });
 
+      // NOTE: Sometimes the order of the entities loaded by the db is different.
+      const sortWorkspacesByName = (
+        workspaceA: Workspace,
+        workspaceB: Workspace
+      ) => workspaceA.created - workspaceB.created;
```
develohpanda: Can remove this and use `createdFirstSort` or `createdLastSort` from https://github.com/Kong/insomnia/blob/7135114768fdaea677a37b6c109e53687a2da277/packages/insomnia-app/app/common/sorting.ts#L28

develohpanda (commented):
Overall code LGTM!

## QA

Previous bugs have been fixed 🔥 

<details>
<summary>🔴 importing curl from dashboard into a new workspace has "undefined" in the prompt, and if importing as a design document, the name is "new collection"</summary>

Could the "undefined" be related to the bug we fixed in develop recently? Importing postman has the same "undefined" message

https://www.loom.com/share/29415c17f1a542158c58a63e1353a397

</details>

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
gatzjames: 👍 Nice one added in https://github.com/Kong/insomnia/commit/978676d276230e8cc8d8036b6866c6912780cc4c

packages/insomnia-smoke-test/modules/home.ts
gatzjames: 🤩 That's a cool feature! didn't know about this one!
I added a navigation to the 'spec' activity before checking if the page is displayed in https://github.com/Kong/insomnia/commit/f3cc35c4fdb3f8474645c9f394de7dc65ea09a64

packages/insomnia-smoke-test/modules/home.ts
gatzjames: 👍 Yep I tried to remove the waitUntil here but the tests would fail often so we should keep it for now!

packages/insomnia-app/app/ui/redux/modules/helpers.ts
gatzjames: Agree. In the current design we create resources in the database and then ask the user about where/how to import them. So if the user cancel's the transaction we rely on db cleanup for orphaned resources. Imo we could improve this by doing different passes e.g. parseImportedData -> getUserInput -> storeInDB.

gatzjames:
> 🔴 importing curl from dashboard into a new workspace has "undefined" in the prompt, and if importing as a design document, the name is "new collection"
> Could the "undefined" be related to the bug we fixed in develop recently? Importing postman has the same "undefined" message
> 
> https://www.loom.com/share/29415c17f1a542158c58a63e1353a397

👍 That's an actual issue:

When we import a curl request we create a new workspace without a name so it appears as undefined in the dialog and then gets the default value which is "New Collection".

When we ask the user about the scope (Collection/Document) we can't create a generic name like __New Document/Collection__ so perhaps we could default to __New Workspace__? @develohpanda @wdawson 

After we get the scope we could change the name but that might be confusing imo:

Defaulting to __New Workspace__ and then updating the name based on scope:
new_worskspace_scope_name

Defaulting to __New Workspace__
new_workspace-name

develohpanda:
> When we ask the user about the scope (Collection/Document) we can't create a generic name like New Document/Collection so perhaps we could default to New Workspace? @develohpanda @wdawson

Hmm, I'm not totally on board with this, we've been dropping the "Workspace" terminology from the UI as much as possible.

This whole process really needs a revisit because this bug (and some other implementations) show that the import is happening in a weird order. I wonder if as a stop-gap we just use "Imported" as the name or something (I really don't like my own suggestion through...)

wdawson:
> When we ask the user about the scope (Collection/Document) we can't create a generic name like New Document/Collection so perhaps we could default to New Workspace? @develohpanda @wdawson

yeah, I think we just need to ask, "Do you want to create a design document or a request collection?" .. no need to have a name there for this I don't think.

wdawson:
Also, would be great to go into that workspace instead of dumping you on the dashboard. I think it's fair to assume that the user will want to start working with the thing they are importing. Bulk import use-cases obviously suffer in that case, but I think that's something we haven't yet tackled and should tackle separately.

develohpanda:
That sounds like a reasonable idea, if we don't have a workspace name then just saying "Do you want to import as design doc or req coll", and if we do have a workspace name, we keep the existing wording?

wdawson:
> That sounds like a reasonable idea, if we don't have a workspace name then just saying "Do you want to import as design doc or req coll", and if we do have a workspace name, we keep the existing wording?

Yep, sounds good (without the abbreviations 😉 )

gatzjames:
> > That sounds like a reasonable idea, if we don't have a workspace name then just saying "Do you want to import as design doc or req coll", and if we do have a workspace name, we keep the existing wording?
> 
> Yep, sounds good (without the abbreviations 😉 )

Good one! Updated this in https://github.com/Kong/insomnia/pull/4058/commits/a0149e3af5da9c706eb5a3f000e818b47f896389

import_new_default_name

packages/insomnia-app/app/common/import.ts
```diff
@@ -268,6 +272,17 @@ export async function importRaw(
       // If workspace, check and set the scope and parentId while importing a new workspace
       if (isWorkspace(model)) {
         await updateWorkspaceScope(resource as Workspace, resultsType, getWorkspaceScope);
+
+        // If the workspace doesn't have a name, update the default name based on it's scope
+        if (!resource.name) {
+          const name =
+            resource.scope === 'collection'
```
develohpanda:
```suggestion
            (resource as Workspace).scope === 'collection'
```

To keep at least some semblance of type safety in this area, even though `as` casting can easily lead to false positives

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
```diff
@@ -168,8 +169,9 @@ describe('selectors', () => {
         activeWorkspaceId: null,
       });
 
+      // NOTE: Sometimes the order of the entities loaded by the db is different so we need to sort them.
       expect(
-        selectWorkspacesWithResolvedNameForActiveProject(state)
+        selectWorkspacesWithResolvedNameForActiveProject(state).sort(sortMethodMap['created-asc'])
```
develohpanda: Oh strange, I though `createdFirstSort` was exported. It probably should be exported and just used directly as `sort(createdFirstSort)` 😂 Some others are exported, like `metaSortKeySort` for instance

packages/insomnia-app/app/ui/redux/modules/helpers.ts
```diff
@@ -100,17 +100,21 @@ export function askToImportIntoWorkspace({ workspaceId, forceToWorkspace, active
 
 export type SetWorkspaceScopePrompt = (name: string) => WorkspaceScope | Promise<WorkspaceScope>;
 export function askToSetWorkspaceScope(scope?: WorkspaceScope): SetWorkspaceScopePrompt {
-  return function(name: string) {
+  return function(name?: string) {
```
develohpanda: `SetWorkspaceScopePrompt` (two lines above) has a required `name` argument, but our function return here has it as optional; strange that TS doesn't complain

develohpanda (approved):
LGTM 🚀 thank you for your patience and perseverance with this one, you're a legend!

We need to do a review of the unresolved comments in this PR, and either address them in a follow-up PR or create Linear tickets for them, but none are blocking to merging this. 🙌🏽

packages/insomnia-app/app/common/import.ts
gatzjames: Good one! I thought for some reason that isWorkspace narrows it down but it's the model not the resource 😅

packages/insomnia-app/app/ui/redux/modules/helpers.ts
gatzjames: 🤔 Good one.
Since (name?: string) is a subset of (name: string) that's ok for TS although not what we want.
The weird thing is that while the opossite is an error [example](https://www.typescriptlang.org/play?ts=4.0.5#code/C4TwDgpgBAyg9gWwgMQK4DsDGwCWd1QC8UAFOgIZID8AXFAM7ABOO6A5gJREB8UARnDgAbCOXQBuAFCSAZhmx4COAEoRgqJunhI0WXPhIc62lPP0EA3pKhQmajQTKUIdRi3ZdCvKzZt31mlAAQoIiYk5IHFI2AL6ScZIiwAyIpnqKRFAq9pomugoGUdL0qfnmJADkwBCMFRySJTpmioZAA) when I tried it in our codebase tsc doesn't complain? e.g. setting `type SetWorkspaceScopePrompt = (name?: string) => ...` and returning a function like `return (name: string) => {...}` should be an error.

packages/insomnia-smoke-test/modules/settings.ts
gatzjames: Agree! The tests would fail when I was using the keyboard shorcuts though in some OSes so leaving it for an update in tests!

packages/insomnia-app/app/ui/redux/modules/helpers.ts
gatzjames: Changed in https://github.com/Kong/insomnia/pull/4058/commits/6640a184a017087529498a8af40f89f352f25608

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
gatzjames: Cool wanted to do this but wasn't sure if there's a reason about the exports there!
Changed it in https://github.com/Kong/insomnia/pull/4058/commits/adf07e19a9b0bc2db6455fd2d6ea532079dff083

packages/insomnia-app/app/ui/redux/modules/helpers.ts
develohpanda: It could relate to strict mode in our code base as strict mode is off 🤔

packages/insomnia-app/app/common/import.ts
develohpanda: Yeah and `isWorkspace(model)` is very _very_ wrong, too. Can you see why?

`model` in this case is the same as `import * as model from '../models/workspace'`,  but the `isWorkspace` type-guard converts `model` to be the `Workspace` interface.

It's a small distinction but problematic. `model` contains the Workspace entity helper functions, but the `isWorkspace` typeguard converts `model` to be typed as a `Workspace` entity

packages/insomnia-app/app/ui/redux/modules/helpers.ts
develohpanda: @gatzjames it's related to `strictFunctionTypes` which is related to `strictMode` by extension. In the playground, I got the error to disappear by disabling `strictFunctionTypes`

image

wdawson:
⛵ LGTM

PR #4060, opened 2021-09-29
unifies getting of workspace name in displayment

dimitropoulos:
closes INS-1008

This PR has a restricted scope: There are deeper things at play regarding the handling of api specs, but right now the thing we want to immediately address is to make sure that the logical path for everywhere the user sees what they would consider the workspace name (although, under the hood, it is actually the apispec.filename that is used for design documents), it is the same.


packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -135,6 +133,30 @@ export const selectActiveWorkspaceMeta = createSelector(
...
+export const selectActiveWorkspaceName = createSelector(
+  selectActiveWorkspace,
+  selectActiveApiSpec,
+  (activeWorkspace, activeApiSpec) => {
+    if (!activeWorkspace) {
+      // see above, but since the selectActiveWorkspace selector really can return undefined, we need to handle it here.
+      return undefined;
+    }
+
+    return isCollection(activeWorkspace) ? activeWorkspace.name : activeApiSpec.fileName;
```
develohpanda: Can we either A) use or B) get rid of the helper? I vote option A.

https://github.com/Kong/insomnia/blob/b5fe4357108f0e347c02918fc37edf381e2b16c7/packages/insomnia-app/app/models/helpers/get-workspace-name.ts#L4-L4

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -135,6 +133,30 @@ export const selectActiveWorkspaceMeta = createSelector(
...
+);
+
+export const selectActiveWorkspaceName = createSelector(
+  selectActiveWorkspace,
+  selectActiveApiSpec,
+  (activeWorkspace, activeApiSpec) => {
+    if (!activeWorkspace) {
+      // see above, but since the selectActiveWorkspace selector really can return undefined, we need to handle it here.
+      return undefined;
+    }
```
develohpanda, line 165: Seems reasonable. An API spec cannot/should not exist without a workspace.

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1103,9 +1105,9 @@ class App extends PureComponent<AppProps, State> {
 
     if (activity === ACTIVITY_HOME || activity === ACTIVITY_MIGRATION) {
       title = getAppName();
-    } else if (activeWorkspace && activeApiSpec) {
+    } else if (activeWorkspace && activeWorkspaceName) {
```
develohpanda, line 1108: As a fresh reader this would seem so counter intuitive 😂 (ie. why can't I just check `activeWorkspace?.name`)

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: abso-toot-ly e78b17337dc6036338851bf907161f48eb3f59ac

great suggestion.  I _THOUGHT_ I saw that helper somewhere but when I searched for it I couldn't find it.  in retrospect it was because I was searching for `isCollection` and `workspace.name` and therefore the single letter variable caught me.

packages/insomnia-app/app/ui/containers/app.tsx
dimitropoulos: I know right.... well... one step at a time, heh.

packages/insomnia-app/app/models/helpers/workspace-operations.ts
```diff
@@ -3,20 +3,23 @@ import type { ApiSpec } from '../api-spec';
...
   }
 }
 
-export async function duplicate(w: Workspace, { name, parentId }: Pick<Workspace, 'name' | 'parentId'>) {
-  const newWorkspace = await db.duplicate(w, {
+export async function duplicate(
+  workspace: Workspace,
+  { name, parentId }: Pick<Workspace, 'name' | 'parentId'>,
+) {
+  const newWorkspace = await db.duplicate(workspace, {
```
gatzjames, line 22: 🥺

gatzjames (approved):

dimitropoulos:
please take another look.  take note of the commit message of [`03e4355` (#4060)](https://github.com/Kong/insomnia/pull/4060/commits/03e4355bfab7dc3ebd017728cf38b9a805577e9c)

If anyone can think of another way to avoid the sfc conversation (or pull out to a separate PR) I'd be grateful, but nothing else I tried worked (including setting displayName which has always worked in every other similar situation so far as I'm aware, and setting of `function.name` is not possible due to javascript use-strict disallowing .name mutations).

packages/insomnia-app/app/models/helpers/__tests__/get-workspace-name.test.ts
```diff
@@ -1,19 +0,0 @@
-import * as models from '../../../models';
-import { WorkspaceScopeKeys } from '../../workspace';
-import getWorkspaceName from '../get-workspace-name';
-
-describe('getWorkspaceName', () => {
```
develohpanda, line 5: It's not nice to see tests being removed (instead of moved) after a refactor, for behavior we intend to keep.

Could you please translate these tests across to where-ever the logic now lives?

packages/insomnia-app/app/ui/components/modals/workspace-settings-modal.tsx
```diff
@@ -539,3 +543,9 @@ export class WorkspaceSettingsModal extends PureComponent<Props, State> {
     );
   }
 }
+
+const mapStateToProps = (state: RootState) => ({
+  activeWorkspaceName: selectActiveWorkspaceName(state),
+});
+
+export const WorkspaceSettingsModal = connect(mapStateToProps, null, null, { forwardRef: true })(UnconnectedWorkspaceSettingsModal);
```
develohpanda, line 551: :+1:

packages/insomnia-app/app/ui/components/workspace-page-header.tsx
```diff
@@ -39,24 +35,16 @@ export const WorkspacePageHeader: FunctionComponent<Props> = ({
...
-    { id: 'home', node: strings.home.singular, onClick: homeCallback },
-    { id: 'workspace', node: <Fragment key="workspace-dd">{workspace}</Fragment> },
+    {
+      id: 'home',
+      node: strings.home.singular,
+      onClick: homeCallback,
+    },
+    {
+      id: 'workspace',
+      node: <WorkspaceDropdown />,
```
develohpanda: YES!!!!!

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -135,6 +132,30 @@ export const selectActiveWorkspaceMeta = createSelector(
...
 
+export const selectAllApiSpecs = createSelector(
+  selectEntitiesLists,
+  entities => entities.apiSpecs,
+);
+
+export const selectActiveApiSpec = createSelector(
+  selectAllApiSpecs,
+  selectActiveWorkspace,
+  (apiSpecs, activeWorkspace) => apiSpecs.find(apiSpec => apiSpec.parentId === activeWorkspace?._id) as ApiSpec
```
develohpanda: Please remove the cast and handle explicitly. Casting here is basically saying "ignore the error".

The same way it's possible for `selectActiveWorkspace` to return `undefined`, it's a totally valid state where there is no `activeApiSpec`.

Think about it: if there's no `activeWorkspace`, should there be an `activeApiSpec`? And, if there _is_ an `activeWorkspace`, we know as a business rule that there must be an `activeApiSpec`, so instead of casting we should throw an error knowing it's unreachable, and return `undefined` when we know it should be `undefined`.

Something like:

```ts
if (!activeWorkspace) {
    // There should never be an active api spec without an active workspace
    return undefined;
}

const activeSpec = apiSpecs.find(apiSpec => apiSpec.parentId === activeWorkspace._id);

if (!activeSpec) {
    // This case should never be reached; an api spec should always exist for a given workspace.
    throw new Error('xyz');
}

return activeSpec;
```

develohpanda (changes requested):
> please take another look. take note of the commit message of [`03e4355` (#4060)](https://github.com/Kong/insomnia/pull/4060/commits/03e4355bfab7dc3ebd017728cf38b9a805577e9c)
> 
> If anyone can think of another way to avoid the sfc conversation (or pull out to a separate PR) I'd be grateful, but nothing else I tried worked (including setting displayName which has always worked in every other similar situation so far as I'm aware, and setting of `function.name` is not possible due to javascript use-strict disallowing .name mutations).

The way around this is to use the unconnected component name in spectron. See the `SidebarRequestRow` component for an example: https://github.com/Kong/insomnia/blob/f48cef02bd14c18dcbf3c7e8393081f5d57743c6/packages/insomnia-smoke-test/modules/component-names.ts#L2

As such, could you please extract the conversion to a FC into a separate PR and connect with redux using `connect()` for the time being?

packages/insomnia-app/app/ui/redux/selectors.ts
gatzjames: > Think about it: if there's no `activeWorkspace`, should there be an `activeApiSpec`? And, if there _is_ an `activeWorkspace`, we know as a business rule that there must be an `activeApiSpec`

🤔 afaik if the workspace scope is 'collection' it wouldn't have an ApiSpec associated with it?

develohpanda: In the current state the model is still always created, to avoid the discrepancy between design/collection from existing in the data layer and further complicating that problem. It's just that m in the UI, there is no way to see the apiSpec (no tab switcher at the top) in a request collection.

However to your point, that can be controlled in the redux selector too - check the active workspace scope and selectively return an api spec :)

dimitropoulos:
@develohpanda I tried that at first and it also fails.  Since this was already approved (well all but this one change) I'm going to wait and see if we can find any potential workaround to this one issue, as to avoid expanding the scope beyond what is critically necessary.

Screenshot_20211004_090911

packages/insomnia-app/app/models/helpers/workspace-operations.ts
dimitropoulos: yeah, I saw that too.  welp.  one step at a time.

packages/insomnia-app/app/models/helpers/__tests__/get-workspace-name.test.ts
dimitropoulos: thanks for bringing it up.  I didn't realize we had tests for selectors and the thought crossed my mind but doing the requisite work to bootstrap it all seemed out of scope.  thankfully, though, upon _actually looking_ (haha) it seems like `reduxStateForTest` does the job: 60a87fc886d16bc6df37983085a2a1511ff9774c

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: removing the cast I'm totally game for - and I did so in 60a87fc886d16bc6df37983085a2a1511ff9774c.

But this change to throw the error isn't really related to the scope of this ticket.  I've done what's strictly asked for up here in 495560ea4a659a07ee48120bc074783f10ab2920 but I'd ask that if we want to expand scope further that we do so in a separate ticket.  I know there's some quirky stuff going on with API specs - but we can't do it all at once and this ticket is about how a workspace name appears.  It's definitely unfortunate how the workspaces and apispecs are conflated and I'm glad we're taking steps towards fixing it -> but I'd like to cap it here for this PR if we can.

develohpanda: This particular example is not expanding scope; we introduced a new selector to achieve the desired behavior and that new selector logic doesn't handle all the logical codepaths.

Moreover, if that error state _is_ being hit, we should expect things to explode in our testing, because as mentioned in the comment that case should never be reached.

Thanks for adding it in, there's no more to be done for this selector.

packages/insomnia-app/app/ui/redux/selectors.ts
develohpanda: @dimitropoulos could you please write a test for this case?

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -135,6 +134,43 @@ export const selectActiveWorkspaceMeta = createSelector(
   },
 );
 
+export const selectAllApiSpecs = createSelector(
+  selectEntitiesLists,
+  entities => entities.apiSpecs,
+);
+
+export const selectActiveApiSpec = createSelector(
```
develohpanda, line 139: Could you please add tests for the various codepaths in here?

packages/insomnia-app/app/ui/redux/selectors.ts
```diff
@@ -100,25 +101,23 @@ export const selectAllWorkspaces = createSelector(
...
 );
 
 export const selectActiveWorkspace = createSelector(
   selectWorkspacesForActiveProject,
   (state: RootState) => state.global.activeWorkspaceId,
   (state: RootState) => state.global.activeActivity,
   (workspaces, activeWorkspaceId, activeActivity) => {
+    if (activeActivity && !isWorkspaceActivity(activeActivity)) {
+      throw JSON.stringify(activeActivity);
+    }
```
develohpanda: Curious, could you please provide any background here, or a comment to outline why this exists?

I've also never come across the `throw 'string'` syntax before, but I'm assuming it doesn't just wrap that into an error and we're missing stack traces?

develohpanda (changes requested):
Almost there, we have the opportunity to write more unit tests

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: Sorry, meant to remove this before pushing. Can ignore.  More tests is the top priority now for sure, yep.

dimitropoulos: removed in 1bc930928180220e62a992586c6a44cc61f7d349

what happened is I was at a loss for why my tests weren't working and trying to get to the bottom of it and it turned out I needed to have a different activity.  I added this to debug and was so relieved when I figure out the problem re: activities that I didn't realize I needed to come remove it after.

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: sure: 91cabe01f6c2c7696434aed7293da108509ab7f0

packages/insomnia-app/app/ui/redux/selectors.ts
dimitropoulos: yep :)

0dd56d3820afcb7d9a13c6f597b917019f38b643

dimitropoulos:
(all tests should now be in)

packages/insomnia-app/app/ui/redux/__tests__/selectors.test.ts
```diff
@@ -44,4 +46,104 @@ describe('selectors', () => {
...
+      });
+
+      expect(selectActiveApiSpec(state)).toMatchObject({
+        contentType: 'yaml',
+        contents: '',
+        fileName: 'apiSpec.fileName',
+        parentId: workspace._id,
+        type: 'ApiSpec',
+      });
+    });
```
develohpanda: Curious: why not store the result of `updateOrCreateForParentId` and check for the entire object here?

Conceptually it tells you that the object created above is the very object that should be in the assertion list. It connects the setup + assertion really explicitly.

```suggestion
    it('will return the apiSpec for a given workspace', async () => {
      const workspace = await models.workspace.create({
        name: 'workspace.name',
        scope: WorkspaceScopeKeys.design,
      });
      const spec = await models.apiSpec.updateOrCreateForParentId(
        workspace._id,
        { fileName: 'apiSpec.fileName' },
      );

      const state = await reduxStateForTest({
        activeActivity: ACTIVITY_DEBUG,
        activeWorkspaceId: workspace._id,
      });

      expect(selectActiveApiSpec(state)).toEqual(spec);
    });
```

develohpanda (approved):

PR #4082, opened 2021-10-06
Inso binaries codesign & notarize on macOS

develohpanda:
This PR extends the test workflow to create a macos installer package. This installer will install inso into `/usr/local/bin`. 

Currently the notarization and stapling steps are enabled in the test workflow, but the intention is for these to only be enabled in the release workflow. There is no need to spam apple and notarize artifacts from every CI run.

Next tasks (in separate PRs):
- change the `compressed` folder name to `artifacts`
- make the inso version configurable (#4086) so we can:
	- append a run-number for CI artifacts
	- tell `pkgbuild` that we're creating a new version of the installer/package
- update the release workflow with these steps 
- explore breaking the test workflow down into smaller bits, to reduce iteration time (low priority)

Resources & links to outline how to package, sign, notarize and staple an installer for macOS are included at the top of `packages/insomnia-inso/src/scripts/macos-pkg.sh`.

Grab artifacts from the latest CI run.


packages/insomnia-inso/src/scripts/codesign.entitlements
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" …>
+<plist version="1.0">
+<dict>
+    <key>com.apple.security.cs.disable-executable-page-protection</key>
```
develohpanda: Explore whether a more specific entitlement can be used: https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-executable-page-protection

johnwchadwick: I'm guessing this is specifically going to be `com.apple.security.cs.allow-unsigned-executable-memory`. Modern versions of V8 probably do follow `MAP_JIT` restrictions, but I doubt the version used in Node v12 does. So this is probably something that we're hitting due to hardened runtime.

develohpanda: Updated in [`de16eb9` (#4082)](https://github.com/Kong/insomnia/pull/4082/commits/de16eb9e63b90716731c99e023e5af40d4140876) and confirmed updating to this entitlement works

johnwchadwick (approved):
Overall, LGTM. Very good work. :+1:

I think we should probably track getting rid of `com.apple.security.cs.disable-library-validation`. Looks like we wind up needing it for Electron, too:

```
% codesign -d --entitlements :- /Applications/Insomnia.app|grep key|cut -c10-|rev|cut -c7-|rev
Executable=/Applications/Insomnia.app/Contents/MacOS/Insomnia
com.apple.security.cs.allow-jit
com.apple.security.cs.allow-unsigned-executable-memory
com.apple.security.cs.allow-dyld-environment-variables
com.apple.security.cs.disable-library-validation
```

Apparently, if we knock that out, upgrading to Electron 12 will let us knock out `com.apple.security.cs.allow-unsigned-executable-memory` as well, and hopefully eventually we'll be able to get it down to just `com.apple.security.cs.allow-jit`, the only logically necessary entitlement. But that's a bit off probably.

develohpanda:
> I think we should probably track getting rid of `com.apple.security.cs.disable-library-validation`

I have a feeling that if we sign the `node-libcurl.node` file (I'm not sure of the exact file name), we can get rid of that entitlement, both from Electron and from the Inso installer package. That shouldn't be too hard to do, we can try it when we have some downtime.

PR #4085, opened 2021-10-07
Add smoke test for XML response and filter

develohpanda:
20 successful runs

image

Test itself should be fairly readable

Test will fail until the next release (alpha/beta/stable - any)


packages/insomnia-smoke-test/core/app.test.ts
```diff
@@ -105,7 +106,37 @@ describe('Application launch', function() {
     await expect(csvViewer.getText()).resolves.toBe('a b c\n1 2 3');
   });
 
-  it('sends PDF request and shows rich response', async () => {
+  it.only('sends dummy.xml request and shows raw response', async () => {
```
gatzjames:
```suggestion
  it('sends dummy.xml request and shows raw response', async () => {
```

gatzjames (approved):

packages/insomnia-smoke-test/core/app.test.ts
develohpanda: My bad!

PR #4086, opened 2021-10-08
Allow specifying a custom version for Inso

develohpanda:
This PR is a follow up to #4082.

This allows a user (and the macos `pkgbuild` command) to differentiate between versions of inso. Currently it is only used in the test workflow for build artifacts, but part of this will be necessary in the release workflow as well.

image

image


johnwchadwick (approved):
LGTM.

PR #4098, opened 2021-10-12
WIP: Spellchecking dictionaries test

gatzjames:
WIP: Testing to see what events the spellcheking module fires on different OSes

### DIFF
--- a/packages/insomnia-app/app/common/electron-helpers.ts
+++ b/packages/insomnia-app/app/common/electron-helpers.ts
@@ -1,4 +1,5 @@
 import * as electron from 'electron';
+import { session } from 'electron';
 import mkdirp from 'mkdirp';
 import { join } from 'path';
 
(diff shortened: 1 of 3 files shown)


develohpanda (commented):
This looks like a pretty good approach if it works! Much better than what I had in mind. It's worth double-checking the failing case (ie. we don't call `disableSpellcheckerDownload`) and expect the test to fail.

It looks like the `spellcheck-dictionary-initialized` is always fired, but I see a `spellcheck-dictionary-download-failure` event on ubuntu which is promissing.

Also as a complete side-note; are there any parameters to the initialize/begin event callback where we can cancel the download, instead of changing the download URL like our current approach?

gatzjames:
> This looks like a pretty good approach if it works! Much better than what I had in mind. It's worth double-checking the failing case (ie. we don't call `disableSpellcheckerDownload`) and expect the test to fail.

+1

> It looks like the `spellcheck-dictionary-initialized` is always fired, but I see a `spellcheck-dictionary-download-failure` event on ubuntu which is promissing.

The 'initialized' event runs on windows/Mac because they provide a native spellchecker!

So we should be checking that either the 'initialised' fired without the 'download-begin' or that there is a 'download-failure' event imo.

Also worth to note that there isn't a way to access the session object from spectron that's why I went with the logs instead. I'd prefer to test this from the outside if we can so any ideas welcome!

> Also as a complete side-note; are there any parameters to the initialize/begin event callback where we can cancel the download, instead of changing the download URL like our current approach?

The docs for the events are not clear as to what's possible https://www.electronjs.org/docs/api/session#event-spellcheck-dictionary-initialized so I'll try to prevent the events and see what happens 👀

PR #4102, opened 2021-10-13
Output Inso binary with .exe extension for Windows

develohpanda:
`pkg` does not automatically append an extension on Windows, so this will change the output to be `inso.exe` when run on Windows.


develohpanda:
When compressing with `tar`, Windows Explorer fails to extract the zip.

image

But using tar to extract contents works fine. The following is from the previous alpha, meaning this PR is not necesasry.

image

PR #4104, opened 2021-10-13
Version env variable for inso artifact name in test workflow

develohpanda:
The test workflow creating run artifacts should suffix the run number onto artifacts, but that wasn't happening because the environment variable was not correctly set. Previously, the `inso-package` and `inso-package:artifacts` steps were combined, but on separation this was missed.

This screenshot is current behavior. The zip file should also have the `-run.1234` suffix.

image


develohpanda:
Superseeded by #4108

PR #4108, opened 2021-10-13
Notarize raw binary

develohpanda:
Changes:
- Use `ditto` to create a zip archive on MacOS
- Notarize the raw binary in addition to notarizing and stapling the installer. The raw binary cannot be stapled.
- Use `-a -cf` tar options for Windows

Grab artifacts from https://github.com/Kong/insomnia/actions/runs/1340523788 to test if necessary


packages/insomnia-inso/src/scripts/macos-pkg.sh
```diff
@@ -43,17 +43,16 @@ security import certificate.p12 -k "$KEYCHAIN" -P "$MACOS_CERTIFICATE_PWD" -T /u
 # New requirement for MacOS 10.12+
 security set-key-partition-list -S apple-tool:,apple:,codesign:,pkgbuild: -s -k "$KEYCHAIN_PASSWORD" "$KEYCHAIN"
 
+# Sign the binary
```
develohpanda, line 46: Sign before copying over to the staging area, because we want to notarize the signed binary in addition to the installer.

The staging area is used for collecting files to be part of the macos installer. Copying over to the staging area is somewhat redundant at the moment because we only create one binary on macos. But this way we ensure that regardless of what files exist under `packages/insomnia-inso/binaries`, only `packages/insomnia-inso/binaries/inso` will be added to the macos installer.

packages/insomnia-inso/src/scripts/artifacts.ts
```diff
@@ -11,39 +11,63 @@ const isMac = () => platform === 'darwin';
...
-  return new Promise<void>(resolve => {
-    const cwd = path.join(__dirname, '../../artifacts');
-    mkdirp.sync(cwd);
+const startProcess = (cwd: ProcessEnvOptions['cwd']) => {
+  const name = getName();
 
-    const tarName = isWindows() ? 'tar.exe' : 'tar';
-    const process = spawn(tarName,
+  if (isMac()) {
+    return spawn('ditto',
```
develohpanda, line 35: Followed the `build a zip archive` steps from https://developer.apple.com/forums/thread/128166. We can consolidate this in the future, to see if just using tar works for all OS's...

.github/workflows/release-core.yml
```diff
@@ -170,12 +170,21 @@ jobs:
...
         if: matrix.os == 'macos-latest'
         uses: devbotsxyz/xcode-staple@v1
         with:
           product-path: packages/insomnia-inso/artifacts/${{ steps.inso-variables.outputs.pkg-name }}.pkg
 
+      - name: Notarize binary
+        if: matrix.os == 'macos-latest'
+        uses: devbotsxyz/xcode-notarize@v1
+        with:
+          product-path: packages/insomnia-inso/binaries/inso
```
develohpanda, line 183: This action creates a temporary `.zip` automatically for submitting to the notarization endpoint

.github/workflows/release-core.yml
```diff
@@ -170,12 +170,21 @@ jobs:
...
+          primary-bundle-id: ${{ steps.inso-variables.outputs.bundle-id }}-binary
```
develohpanda, line 184: Thoughts on the bundle id? Docs say it just needs to be unique, so I appended `-binary` to the end of it... 🤷🏽

johnwchadwick: You know, I don't really know for sure if this matters. I think raw mach-o binaries don't have a specific bundle ID, though, so it probably can be any valid value.

packages/insomnia-inso/src/scripts/artifacts.ts
johnwchadwick: Tar should work on any system where `tar`==`bsdtar` (macOS, Windows.) You can also install `bsdtar` on Ubuntu and other Linux distros, but I think they tend to come with the more limited GNU tar by default. Of course, if we don't need to make `.zip` files on Linux, this is fine.

.github/workflows/release-core.yml
develohpanda: I tried removing this, but the build failed, so just adding _something_ works fine

johnwchadwick (approved):
LGTM.

.github/workflows/release-core.yml
johnwchadwick: Ahhh, yeah; that seems fine to me.

PR #4124, opened 2021-10-15
Fix dropdown search crashing on child node checks

gatzjames:
Fixes an issue where searching on a dropdown results in a crash.

### Before:
Searching on the dropdown will result in an error:
dropdown-before

### After:
Searching on the dropdown will filter out the items.
I used the `hidden` class instead of not rendering the items. Otherwise this breaks the search functionality that uses the rendered DOM elements to find matches.
dropdown-after

Closes INS-897


dimitropoulos (approved):
tested and can confirm the break is gone, now.

PR #4132, opened 2021-10-19
block starting the app if there's an invalid config

dimitropoulos:
closes INS-1030

This PR will block starting the app if there's an invalid insomnia config, and report the raw error (and file location) to the user.

Before this PR, the app would happily start up, despite an invalid config.  After this PR, it shows a message like this to the user:

for a config like

```json
{
  "insomniaConfig": "1.0.0",
  "settings": {
    "allowNotificationRequests": 1,
    "disablePaidFeatureAdz": false
  }
}
```

and then kills the app process.

### DIFF
--- a/packages/insomnia-app/app/models/helpers/settings.ts
+++ b/packages/insomnia-app/app/models/helpers/settings.ts
@@ -1,6 +1,7 @@
 import { readFileSync } from 'fs';
 import { Settings } from 'insomnia-common';
 import { InsomniaConfig, validate } from 'insomnia-config';
+import { ValidationResult } from 'insomnia-config/dist/validate';
 import { resolve } from 'path';
 import { mapObjIndexed, once } from 'ramda';
 import { omitBy } from 'ramda-adjunct';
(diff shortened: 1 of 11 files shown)


develohpanda (approved):
For QA, I ran `npm run app-package:smoke` locally. I then placed an invalid config file in the app data folder, and launched the packaged app. I see the following on macos.

As an MVP, this is fine IMO, because of time constraints

image

packages/insomnia-app/app/models/helpers/settings.ts
```diff
@@ -1,6 +1,7 @@
 import { readFileSync } from 'fs';
 import { Settings } from 'insomnia-common';
 import { InsomniaConfig, validate } from 'insomnia-config';
+import { ValidationResult } from 'insomnia-config/dist/validate';
```
develohpanda: We should fix this, but fixing in this PR will involve disabling smoke tests

dimitropoulos: this was a mistake.  we should fix it indeed.

dimitropoulos: fixed in 5f7409a32c583e36cc8a5310f33c668c057594e1

PR #4133, opened 2021-10-20
human readable errors for insomnia config validation

dimitropoulos:
closes INS-1032

This PR introduces the following quality-of-life improvement to insomnia config errors:

| BEFORE | AFTER |
| - | - |
| Screenshot_20211019_171720 | Screenshot_20211019_212250 |

### DIFF
--- a/packages/insomnia-app/app/models/helpers/settings.ts
+++ b/packages/insomnia-app/app/models/helpers/settings.ts
@@ -1,6 +1,6 @@
 import { readFileSync } from 'fs';
 import { Settings } from 'insomnia-common';
-import { ErrorResult, InsomniaConfig, isErrorResult, validate } from 'insomnia-config';
+import { ErrorResult, InsomniaConfig, isErrorResult, validate } from 'insomnia-config/dist';
 import { resolve } from 'path';
 import { mapObjIndexed, once } from 'ramda';
 import { omitBy } from 'ramda-adjunct';
--- a/packages/insomnia-app/app/models/helpers/settings.ts
+++ b/packages/insomnia-app/app/models/helpers/settings.ts
@@ -1,6 +1,6 @@
 import { readFileSync } from 'fs';
 import { Settings } from 'insomnia-common';
-import { InsomniaConfig, validate, ValidationResult } from 'insomnia-config/dist';
+import { ErrorResult, InsomniaConfig, validate } from 'insomnia-config/dist';
 import { resolve } from 'path';
 import { mapObjIndexed, once } from 'ramda';
 import { omitBy } from 'ramda-adjunct';
@@ -71,7 +71,8 @@ interface ConfigError {
   error: {
     configPath?: string;
     insomniaConfig: unknown;
-    errors: ValidationResult['errors'];
+    errors: ErrorResult['errors'];
+    humanErrors: ErrorResult['humanErrors'];
   };
 }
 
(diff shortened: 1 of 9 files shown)


packages/insomnia-app/app/models/helpers/settings.ts
```diff
@@ -71,7 +71,8 @@ interface ConfigError {
   error: {
     configPath?: string;
     insomniaConfig: unknown;
```
develohpanda, line 76: I've noticed this a couple of times and am not asking for any changes here but just want to understand, why is it that this needs to be typed as `unknown`?

packages/insomnia-config/src/validate.ts
```diff
@@ -22,12 +23,13 @@ export const ingest = (input: string | InsomniaConfig | unknown) => {
 export interface ValidResult {
   valid: true;
   errors: null;
-  humanError: null;
+  humanErrors: null;
```
develohpanda: Why `null` instead of omitting the property (both `errors` and `humanErrors`) altogether?

develohpanda (dismissed):
Error on macOS, LGTM

image

In the `[Error 1]` section, I see `.settings/insomnia.config.json`. I don't understand what that syntax means 🤔

packages/insomnia-config/src/validate.ts
dimitropoulos: the reason for this approach is to avoid having to make typeguards, so that (ideally) typechecking will automatically flow when you do a membership check (i.e. with the `in` operator or something).  now that there's a typeguard in your recent commit, then, yes, it's not needed anymore.  removed in f80d69150108b8e6d9f6ed873f9b0d79084bc177

packages/insomnia-app/app/models/helpers/settings.ts
dimitropoulos: Think about it this way: if there's an error, then by definition the thing provided is _not_ an `InsomniaConfig`.  But what is it?  Well, we don't know.  Hence, `unknown`.  We can't make any assumptions about the shape of this object, so we just have to take it piece by piece (if we need pieces of the data), which is what `unknown` requires you to do.

packages/insomnia-config/src/validate.ts
```diff
@@ -21,17 +22,19 @@ export const ingest = (input: string | InsomniaConfig | unknown) => {
...
 export interface ValidResult {
   valid: true;
-  errors: null;
-  humanError: null;
 }
 
 export interface ErrorResult {
   valid: false;
   errors: ErrorObject[];
+  humanErrors: ValidationError[];
```
gatzjames: This name leads me to think these are errors made by humans/users.
humanReadableErrors or validationErrors would be more precise imo

packages/insomnia-config/src/validate.ts
dimitropoulos: Screenshot_20211020_113250

Ok, actually, I think I take it back, and for the same reason I mentioned above.  I thought that it wouldn't be that hard in the code, and, indeed, it's not _that_ hard, but it makes testing much much more of a burden.

dimitropoulos: added back in 9dfd8354573f1741006805f6bbebe5b883ac7789

gatzjames (commented):
~~Having an invalid json in the `insomniaDataDirectory` path still opens the app for me:~~ Addressed in another PR

If I provide an incorrect value I see an error as expected:
insomnia-config-error

packages/insomnia-app/app/models/helpers/settings.ts
```diff
@@ -39,6 +38,10 @@ export const getLocalDevConfigFilePath = () => (
   isDevelopment() ? '../../packages/insomnia-app/app' as string : undefined
```
gatzjames, line 38: Not related to this PR but should we use the path util instead:
```suggestion
  isDevelopment() ? path.join(__dirname, '..', '..', 'packages', 'insomnia-app', 'app') as string : undefined
```

packages/insomnia-config/src/validate.ts
dimitropoulos: ok, I see where you're coming from, I'll update now

dimitropoulos: updated in ad398ee033f12c5019391f942108751371933774

packages/insomnia-app/app/models/helpers/settings.ts
dimitropoulos: I think it's related _enough_ since this literal path appears on the screen to users as part of the change to the modal made in this PR.  it's not going to ever fail since it goes through `resolve`, but yes, you're right, this should be fixed.

gatzjames (approved):
🚀 LGTM

packages/insomnia-app/app/models/helpers/settings.ts
dimitropoulos: update, heh.  it didn't work.. not sure why.  but yes, I agree, we can handle later

PR #4137, opened 2021-10-20
hides ad in account tab when `disablePaidFeatureAds` is `true`

dimitropoulos:
closes INS-1070

By mistake, when implementing `disablePaidFeatureAds` we missed an ad that exists in the account tab.  Now, when `disablePaidFeatureAds` is `true`, you will not see this ad.

| BEFORE | AFTER |
| - | - |
| before | after |


develohpanda (approved):
LGTM, this approval doesn't have QA however screenshots and code look good

PR #4139, opened 2021-10-20
fails validation if the config has a parse error

dimitropoulos:
closes INS-1046

this will present an error to the user if the config

Before this PR, invalid JSON was just skipped.

After this PR, users are presented with a message like this:

and then the app exits.

The above can be tested with a bad insomnia config, e.g.
```json
{
  "insomniaConfig": "1.0.0",
  settings": {}
}
```


gatzjames (approved):
Works nice!
Screenshot 2021-10-20 at 21 12 43

PR #4147, opened 2021-10-22
Handle human readable errors correctly

develohpanda:
There exists an underlying bug with isConfigError, where it was checking for a field that does not exist, because of `any` types. This causes a false positive with the unit tests, and means the fallback `json.stringify` error message is shown, rather than the humanReadable one (see snapshot change).

Using the following config file, 

```json
{
  "insomniaConfig": "1.0.0",
  "settings": {
    "abc": 123
  }
}
```

2021.6.0-alpha.6 shows
image

This PR shows
image

Follow-up to https://github.com/Kong/insomnia/pull/4134. Closes INS-1092.

### DIFF
--- a/packages/insomnia-app/app/common/__tests__/validate-insomnia-config.test.ts
+++ b/packages/insomnia-app/app/common/__tests__/validate-insomnia-config.test.ts
@@ -8,7 +8,7 @@ const getConfigSettings = mocked(_getConfigSettings);
 
 describe('validateInsomniaConfig', () => {
   it('should return error if there is a parse error', () => {
-  // Arrange
+    // Arrange
     const errorReturn = {
       error: {
         syntaxError: new SyntaxError('mock syntax error'),
@@ -27,7 +27,7 @@ describe('validateInsomniaConfig', () => {
   });
 
   it('should return error if there is a config error', () => {
-  // Arrange
+    // Arrange
     const errorReturn: ConfigError = {
       error: {
         errors: [],
@@ -52,7 +52,7 @@ describe('validateInsomniaConfig', () => {
   });
 
   it('should return error if there is an unexpected error', () => {
-  // Arrange
+    // Arrange
     const errorReturn = {
       error: {
         errors: [],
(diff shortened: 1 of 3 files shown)


packages/insomnia-app/app/common/__tests__/validate-insomnia-config.test.ts
```diff
@@ -27,7 +27,32 @@ describe('validateInsomniaConfig', () => {
   });
 
   it('should return error if there is a config error', () => {
-    // Arrange
+  // Arrange
```
reynolek: Intention white space change?

packages/insomnia-app/app/common/__tests__/validate-insomnia-config.test.ts
```diff
@@ -1,14 +1,14 @@
...
-import { getConfigSettings as _getConfigSettings  } from '../../models/helpers/settings';
+import { ConfigError, getConfigSettings as _getConfigSettings  } from '../../models/helpers/settings';
 import { validateInsomniaConfig } from '../validate-insomnia-config';
 
 jest.mock('../../models/helpers/settings');
 const getConfigSettings = mocked(_getConfigSettings);
 
 describe('validateInsomniaConfig', () => {
   it('should return error if there is a parse error', () => {
-    // Arrange
```
reynolek: Intentional white space change?

packages/insomnia-app/app/common/__tests__/validate-insomnia-config.test.ts
develohpanda: Accidental during copy-pasta. All fixed.

gatzjames (approved):
👍🚀

reynolek (approved):
:shipit:

reynolek:
The only interesting thing I see (and this imo isnt a blocker) is that the previous error showed an Insomnia logo in the screenshot, and the new error shows an electron logo.

develohpanda:
Ah, that's because the screenshot from this pr is from dev mode and not after packaging the app. While running in dev mode, the electron logo is used.

reynolek:
> Ah, that's because the screenshot from this pr is from dev mode and not after packaging the app. While running in dev mode, the electron logo is used.

Oh sweet, so basically no big deal and it will be fixed in the release bundle. So non-issue. Carry on! :)

PR #4148, opened 2021-10-22
Exit app with fail exit code if config failed validation

develohpanda:
This came up in discussion during review of #4134. This is NOT a blocker for release, raised a separate PR for discussion if necessary.

This PR
image

2021.6.0-alpha.6
image


jackkav (approved):
Lgtm

reynolek (approved):
I think this is the right approach since we are hitting a critical failure preventing launch.

Issue #1160, opened 2018-09-13
[Feature Request] JSON Input - Minify Button


barell:
- Insomnia Version: 6.0.1
- Operating System: Ubuntu 16.04

## Details

At the moment when I send a request using JSON, there is a button "Beautify JSON" at the bottom left corner. It is very useful and quick when I paste unformatted JSON to make it readable.

I think it would be nice to have opposite feature, eg. button called "Minify JSON" which basically means no new lines and spaces between JSON tokens. I guess, similar thing could be done to XML request type as I can see there is "Beautify XML" button.

dungmv:
I created pr minify json and xml, hope that useful #4154

dimitropoulos:
could you please provide some context as to why you need this feature?

would a plugin that does it not be sufficient in some way?

dungmv:
a few days ago, my co-worker ask me Curl. I was generating Curl from Insomnia, it look awful. so I think, it will be better if the body minify before generating Curl, And sometimes I want to send a body to co-workers

dimitropoulos:
Are you aware of the `Copy as Curl` feature?

### XML

![Screenshot_20211102_081643](https://user-images.githubusercontent.com/15232461/139844369-220d8846-57dd-4d7a-9cdb-043d4ad9afde.png)


```bash
curl --request POST \
  --url http://mockbin.org/echo \
  --header 'Content-Type: application/xml' \
  --data '<slideshow>
	<author>Yours Truly</author>
	<title>Sample Slide Show</title>
</slideshow>'
```

### JSON

![](https://user-images.githubusercontent.com/15232461/139843691-8c96312e-f112-4246-bc97-b1328645b493.png)

```bash
curl --request POST \
  --url http://mockbin.org/echo \
  --header 'Content-Type: application/json' \
  --data '{
  "slideshow": {
    "author": "Yours Truly",
    "title": "Sample Slide Show"
  }
}'
```

Does this solve your use-case (and more)?

dungmv:
@dimitropoulos  I know how to generate curl from insomnia,
but you should try with a large body, it looks very bad, and when I try to send it to my co-worker by telegram, It breaks into multi-message and copies very hard
for example
```
curl --request POST \
  --url http://localhost/order_bundle/tmdt_export_order \
[...]

PR #4154, opened 2021-10-26
minify json and xml

dungmv:
add button minify xml and json
Screen Shot 2021-10-26 at 22 53 08

Screen Shot 2021-10-26 at 22 54 10


develohpanda:
Hi @dungmv, thank you for the PR! Could you please provide any information to the use case for a minify function?

Also, as an alternate solution, have you explored using a [request hook plugin](https://docs.insomnia.rest/insomnia/hooks-and-actions#request-and-response-hooks) to intercept the request and minify the body before sending?

dungmv:
@develohpanda I explained some cases I have met here https://github.com/Kong/insomnia/issues/1160#issuecomment-952455390

PR #4159, opened 2021-10-28
Reduce AppBar sizing and add Project Name to breadcrumbs

dimitropoulos:
> Note: this PR is still awaiting design decisions on the color of the background bar (and thus, some subtle color changes of the multiswitch)

closes INS-845

This PR implements a reduction in size of the app's top bar.

| BEFORE | AFTER |
| - | - |
| Screenshot_20211028_120101 | Screenshot_20211028_115518 |


dimitropoulos (commented):
image

By they way, it crossed my mind to also make sure to update the sizing of the bar as it's represented in the theme selector preview.  But then, when I looked at it, I found that it's actually been more narrow all along.  It's a loose representation of the app, anyway, so it doesn't _deeply_ matter that a tweak of the sort that's implemented in this PR, but I just wanted to mention that I checked and it looks good already.

packages/insomnia-components/src/breadcrumb.tsx
```diff
@@ -24,19 +24,28 @@ const StyledBreadcrumb = styled.ul`
   a,
   a::before,
   li > ::before {
-    color: var(--color-font);
+    color: var(--color-font) !important;
```
develohpanda, line 27: Is there a way we can avoid `!important`? What is taking prcedence over this resulting in the need for `!important`? Not a blocker but want to avoid this if we can.

packages/insomnia-app/app/ui/components/app-header.tsx
```diff
@@ -0,0 +1,54 @@
...
+  breadcrumbProps: { crumbs, isLoading },
+  gridCenter,
+  gridRight,
+}) => {
+  return (
+    <Header
+      gridLeft={(
+        <Fragment>
+          <img src={coreLogo} alt="Insomnia" width="28" height="28" />
+          <Breadcrumb crumbs={crumbs} isLoading={isLoading}/>
```
develohpanda: Why not expand `breadcrumbProps` into this? `<Breadcrumb {...breadcrumbProps} />`

packages/insomnia-app/app/ui/css/components/dropdown.less
```diff
@@ -7,6 +7,10 @@
   &:focus {
     outline: none;
   }
+
+  & button {
+    cursor: pointer;
```
develohpanda, line 12: This only works on some dropdowns. Eg on the dashboard, the card context menu dropdown has a pointer cursor, but the project dropdown, pull dropdown, create dropdown all still have the default cursor

develohpanda (changes requested):
Insomnia Sync button on a request collection inside a remote project is broken

image

packages/insomnia-components/src/breadcrumb.tsx
dimitropoulos: oh, I really wanted to, but I couldn't find another way.  ~it's already `!important`ed higher up, so there isn't much option without ripping it up a lot more than I was trying to do in this PR.~ correction, it's not `!important`ed higher up, but it is overridden with a higher specificity.

This is what happens without the addition of ` !important`.



As an aside, this seems to be something we need to do fairly often.  CSS in JS solves this because there's never any specificity fighting that goes on, but maybe we can come up with some interim solution (outside of this PR).

packages/insomnia-app/app/ui/components/app-header.tsx
dimitropoulos: sure, 81182def3ff043f2316a6f879bf3b8364129c3a1.

the answer to "why" is that it's a habit of mine from using TypeScript and React in the very beginning.  It was a while before TypeScript could really reliably handle spread props, and due to that I don't think to do it very often out of habit, even in situations like this where it makes total sense and even these days when TypeScript can handle it just fine.

packages/insomnia-app/app/ui/css/components/dropdown.less
dimitropoulos: good to know.  yeah, we'll get to them all eventually I hope, but I just wanted to stay focused on the design that's in scope for this job.  I'd love it if you could open a PR to change the rest, I'm not actually sure where they all are.

if you'd like some change to take place for this PR, please unresove and let me know.

dimitropoulos:
d4d46a5ed7f657bd1284296493f2aae83ea4101c fixes the dropdown sync button

packages/insomnia-components/src/breadcrumb.tsx
develohpanda: Ah all good then, thanks for trying to

packages/insomnia-app/app/ui/components/app-header.tsx
develohpanda: Awesome!

wdawson:
Is the `Project name` breadcrumb still clickable? It looks like it wouldn't be because it's grey. I think it should be to take you to the dashboard for that project.

develohpanda (approved):
LGTM  🚀 

> Is the Project name breadcrumb still clickable? It looks like it wouldn't be because it's grey. I think it should be to take you to the dashboard for that project.

@wdawson yep still clickable: https://www.loom.com/share/69c139125b08481b983115cadab65239

PR #4161, opened 2021-10-29
add vscode launch config for debugging inso

gatzjames:
Adds a vscode launch configuration for inso making it easy to debug from the editor.

inso_debug

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -60,8 +60,7 @@
       "detail": "Compile Renderer (Watch) | Compile Main",
       "dependsOn": [
         "Insomnia: Compile Renderer (Watch)",
-        "Insomnia: Compile Main",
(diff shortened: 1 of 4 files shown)


.vscode/tasks.json
```diff
@@ -61,7 +61,50 @@
       "dependsOn": [
         "Insomnia: Compile Renderer (Watch)",
         "Insomnia: Compile Main",
+        "Components: Compile (Watch)"
```
develohpanda: Was this change intentional?

develohpanda (approved):
🚀 I left the second argument/prompt empty and inso prompted me as expected

image

.vscode/tasks.json
gatzjames: 😅 No forgot it cause I'm working on a launch config for when working with the components pkg 😁

PR #4162, opened 2021-10-29
inso: use --src for loading a git dir

gatzjames:
The `--src` flag is now also used by the gitAdapter to check if the given path is a git directory as expected by the inso cli. 

inso-src

Todo
- [ ] Corresponding PR in Docs

Closes INS-786


falondarville:
@gatzjames FYI We're going to move all the main Inso CLI docs to docs.insomnia.net so we don't have to maintain both the docs site and the README. I'll add a link to the README after I've finalized the Inso CLI docs on the site.

packages/insomnia-inso/src/__snapshots__/inso-snapshot.test.ts.snap
```diff
@@ -7,7 +7,7 @@ A CLI for Insomnia!
```
develohpanda: Interesting, how did these snapshots change, given there's no change to where the `--workingDir` description is set? In any case, `--workingDir` isn't being deprecated (only `--appDataDir`).

packages/insomnia-inso/src/db/index.ts
```diff
@@ -56,14 +56,14 @@ export const loadDb = async ({
   let db: Database | null = null;
 
   // try load from git
-  if (!appDataDir && !src) {
-    const dir = workingDir || '.';
+  if (!appDataDir) {
+    const dir = src || workingDir || '.';
```
develohpanda, line 60: 👍🏽

develohpanda (approved):
LGTM. Like mentioned above I think we can skip on updating the README docs as we'll remove them in favor of https://github.com/Kong/insomnia-docs/pull/53 before next release.

image

PR #4165, opened 2021-11-01
Remove unused core plugin metadata option

develohpanda:
Core plugins are listed here, and according to https://github.com/Kong/insomnia-website/pull/17  the option isn't used by anything, therefore it is redundant.

https://github.com/Kong/insomnia/blob/841df1cff4221bba381c82ac3b1a3b595f4320f3/packages/insomnia-app/config/config.json#L24-L39


gatzjames (approved):
🧹🚀

PR #4170, opened 2021-11-02
inso: try to run inso from a Docker container

filfreire:
This PR is just an experiment, there's no official ticket yet, to try and run `inso` from a Docker container, without requiring `inso` users to have a local node/npm environment setup, and possibly in the future so folks can easily just run an `inso` container as part of their own CI stages, or abstract on top of it.

## How to run
1. Clone the repository (this branch in specific).
2. Install [Docker](https://docs.docker.com/get-docker/).
3. Run `docker build -t insomnia-inso:latest -f inso.Dockerfile .`.
4. Create an alias for the containerised version of `inso`, e.g. `alias inso-docker="docker run -it --rm insomnia-inso:latest`.
5. Try to run a typical `inso` command using the container, e.g. `inso-docker help`

UPDATE:
In order to run Insomnia specs in a container, you'll need to mount the specs folder on your most machine to a given folder on the container. Examples:

- Mounting an Insomnia git sync repository folder to a folder on the container:

```shell
cd <your-git-sync-repo-folder>
docker run -it --rm -v $(pwd):/var/temp insomnia-inso:latest run test -w /var/temp
```

- Mounting the Insomnia Application Data folder:

```shell
# On macOS:
docker run -v $HOME/Library/Application\ Support/Insomnia:/var/temp -it --rm insomnia-inso:latest run test --src /var/temp
# On Linux:
docker run -v $HOME/.config/Insomnia:/var/temp -it --rm insomnia-inso:latest run test --src /var/temp
# On Windows (using Docker for Windows and WSL):
docker run -v /mnt/c/Users/<your_username>/AppData/Roaming/Insomnia:/var/temp -it --rm insomnia-inso:latest run test --src /var/temp
```

- Mounting the folder where you keep an Insomnia v4 export:

```shell
cd <some-folder>
docker run -it --rm -v $(pwd):/var/temp insomnia-inso:latest run test -w /var/temp/Insomnia_YYYY-MM-DD.json
```

## What hasn't been tested

- ~~Still need to work out the quirks of mounting directories so folks can actually run their insomnia models.~~ Added detailed instructions on how folks can mount and run specs in commit https://github.com/Kong/insomnia/commit/fabec5fc09b81cd8626645e59385bc5b35154e71
- ~~The `bootstrap`/`build` stage (that runs `webpack --config webpack/webpack.config.development.js`) fails when running `docker build (...)` on some `tslint` rules, hence the `@ts-ignore` additions.~~  UPDATE: The failures were due to the fact `typescript` inside the container was defaulting to version `4.4.0`, causing new TSLint failures. We've applied workarounds for those failures in this PR, so we don't need to pin typescript to an older version.


packages/insomnia-inso/src/commands/generate-config.ts
```diff
@@ -81,6 +81,7 @@ export const generateConfig = async (
       result = await generate(fileName, conversionType, generationTags);
     }
   } catch (e) {
+    // @ts-ignore: weird docker build fail with TS2345
```
develohpanda: Could it be that you also need to copy `.eslintrc.js` from the lerna root? `packages/insomnia-inso/.eslintrc.js` extends `.eslintrc.js`, same with `.eslintignore`

filfreire: Copying `.eslint(...)` files didn't seem to do the trick. 

We might be missing something else that happens when we run `npm run bootstrap` on the lerna root level that isn't happening if we run `npm run bootstrap` on the `packages/insomnia-inso` level alone.

I can reproduce this locally outside of docker if I do:
```shell
# on root of the project
npm run clean

# cd into inso package
cd packages/insomnia-inso
npm install
npm install typescript rimraf webpack webpack-cli -g
npm link typescript
npm run bootstrap
```

Then I get the same kind of errors as in the docker build:

```
ERROR in /Users/someuser/insomnia/packages/insomnia-inso/src/write-file.ts
./src/write-file.ts 19:62-63
[tsl] ERROR in /Users/someuser/insomnia/packages/insomnia-inso/src/write-file.ts(19,63)
      TS2345: Argument of type 'unknown' is not assignable to parameter of type 'Error | undefined'.
  Type 'unknown' is not assignable to type 'Error'.

(...)
```

develohpanda: Oh this appears to be a valid TS error, I see it in VS Code when navigating to `packages/insomnia-inso/src/write-file.ts`. Same error exists in a few other places too

image

develohpanda: I think the implementation of `InsoError` probably needs to change to the following. @gatzjames how does the `error.cause` polyfill type the `cause` property?

```diff
export class InsoError extends Error {
- cause?: Error | null;
+ cause?: unknown;

- constructor(message: string, cause?: Error) {
+ constructor(message: string, cause?: unknown) {
    super(message);
    this.name = 'InsoError';
    this.cause = cause;
  }
}
```

filfreire: @develohpanda it seems pinning `typescript` to `4.2.3` avoids these TSlint errors for now, including the dockerfile.

Screenshot 2021-11-15 at 09 52 39


cc @gatzjames @dimitropoulos

gatzjames: > I think the implementation of `InsoError` probably needs to change to the following. @gatzjames how does the `error.cause` polyfill type the `cause` property?
> 
> ```diff
> export class InsoError extends Error {
> - cause?: Error | null;
> + cause?: unknown;
> 
> - constructor(message: string, cause?: Error) {
> + constructor(message: string, cause?: unknown) {
>     super(message);
>     this.name = 'InsoError';
>     this.cause = cause;
>   }
> }
> ```

Yep if we want to be close to the spec then unknown is the way to go. 👍
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/error-cause/Error.d.ts

develohpanda: Sweet 👍 I'd vote to make this change rather than pinning the version of TS though interesting that pinning doesn't throw errors

packages/insomnia-inso/src/commands/lint-specification.ts
```diff
@@ -45,7 +45,11 @@ export async function lintSpecification(
       return false;
     }
   } catch (e) {
-    logger.fatal(e.message);
+    let errorMessage = "Failed to load linting specification."
```
filfreire: Fixes 
```
#22 5.321 ./src/commands/lint-specification.ts 48:17-18
#22 5.321 [tsl] ERROR in /packages/insomnia-inso/src/commands/lint-specification.ts(48,18)
#22 5.321       TS2571: Object is of type 'unknown'.
```

packages/insomnia-inso/src/commands/generate-config.ts
filfreire: I've applied the above fix and unpinned the TS version.

filfreire:
@gatzjames @develohpanda @dimitropoulos - added instructions to the README of how folks can both build and run inso from the container against their exports, insomnia appdata dirs, and git sync folders. Would appreciate a review 🙏 

In this PR we're not pushing any official inso container image to Docker Hub just yet - we can take care of that in another PR - but this way any folks that want to work with `inso` in a Docker ecosystem (and avoid local node setup) can do so.

inso.Dockerfile
```diff
@@ -0,0 +1,23 @@
+# syntax=docker/dockerfile:1
```
develohpanda: Just a note on the naming for this file; maybe we should use `inso-dev.Dockerfile` to avoid any confusion with using inso in production with a dockerfile (INS-1177)?

inso.Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY lerna.json .
+
+WORKDIR /packages/insomnia-inso/
+
+RUN npm install
+RUN npm install typescript rimraf webpack webpack-cli -g
+
+RUN npm link typescript
+
+RUN npm run bootstrap
```
develohpanda: Because we only copy `packages/insomnia-inso`, how does it link as expected with `packages/insomnia-send-request` and `packages/insomnia-testing` (for example)? Fairly unfamiliar with docker so this might be a dumb question :)

packages/insomnia-inso/README.md
```diff
@@ -26,3 +26,45 @@ See the [open-source Inso CLI documentation](https://docs.insomnia.rest/inso-cli
...
+
+## Run with Docker
+
+> Note: this feature is still experimental / in active development.
+
+1. Clone the repository.
+2. Install [Docker](https://docs.docker.com/get-docker/).
+3. Run `docker build -t insomnia-inso:latest -f inso.Dockerfile .`.
+4. (Optional) Create an alias for the containerised version of `inso`, e.g. `alias inso-docker="docker run -it --rm insomnia-inso:latest"`.
+5. (Optional) Try to run an `inso` command, e.g. `inso-docker help`
```
develohpanda: For the tag, should we use something like `insomnia-inso:dev` to make it really clear this is just running in dev mode, and to differentiate from the production dockerfile which would use the single executable from github releases?

develohpanda (commented):
> In order to run Insomnia specs in a container, you'll need to mount the specs folder on your most machine to a given folder on the container.

I'm guessing this will be the case with using the single executable via docker as well, right? See INS-1177

packages/insomnia-inso/src/commands/lint-specification.ts
develohpanda:
```suggestion
    let errorMessage = 'Failed to load linting specification.';
```

To fix up the lint error

inso.Dockerfile
filfreire: Agreed, renamed it.

inso.Dockerfile
filfreire: I'm not entirely sure, but I suspect it is because both are only used in tests - and in this particular docker build we don't run the `test` script, we only run `bootstrap` which only runs  webpack.

packages/insomnia-inso/README.md
filfreire: Agreed, renamed it to `insomnia-inso:dev` 👍

filfreire:
> > In order to run Insomnia specs in a container, you'll need to mount the specs folder on your host machine to a given folder on the container.
>
> I'm guessing this will be the case with using the single executable via docker as well, right? See INS-1177

@develohpanda, yes, this will also be the case if we use single executable in a docker container.

packages/insomnia-inso/src/commands/lint-specification.ts
filfreire: Fixed it, thanks 👍

inso.Dockerfile
develohpanda: insomnia-testing contains the unit testing functionality that we have in design documents, on the test tab and in `inso run test`, not specifically for unit tests of the codebase. It's a bit meta and confusing 😅

filfreire:
@develohpanda @wdawson @wongstein - gonna go ahead and close this PR as it is "stalled".
We'll pick back up the docker topic with https://linear.app/insomnia/issue/INS-1238/publish-inso-cli-docker-images-as-part-of-the-release-workflow

PR #4190, opened 2021-11-08
reorders github issue template

dimitropoulos:
closes INS-1160

The goal of this ticket is just to rearrange ~(but not modify)~* the things in the GitHub issue template to make the final output easier to read for developers and contributors.

I added b66c244479c3d1ea244897fac41a945fea8089a8 as a separate commit so we can drop it individually if we decide we don't want to have the "duplicate issue check" at the top.  Seems like there's good reason to do it both ways.

*in https://github.com/Kong/insomnia/pull/4190#discussion_r745144302 there was one other very minor change made

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -2,6 +2,13 @@ name: Bug Report
 description: Report an Insomnia bug
 labels: [B-bug, S-unverified]
 body:
+- type: checkboxes
+  attributes:
+    label: Is there an existing issue for this?
+    description: Please search to see if an issue already exists for the bug you encountered.
+    options:
+      - label: I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for a bug that matches the one I want to file, without success.
+        required: true
 - type: textarea
   attributes:
     label: Expected Behavior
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -9,6 +9,34 @@ body:
...
+      2. Click on '....'
+      3. Scroll down to '....'
+      4. See error
+- type: textarea
+  attributes:
+    label: Additional Information
+    description: |
+      If your problem needs further explanation, please add more information here.
+
+      Tip: You can attach images or log files by clicking this area to highlight it and then dragging files in.
```
wdawson: Maybe we should move this tip to the first `textarea`?

wdawson (approved):
LGTM ⛵

develohpanda (approved):
Totally on board with this, it'll help in slack previews as well 🙌🏽

.github/ISSUE_TEMPLATE/bug_report.yml
dimitropoulos: cool: talked with @wdawson on how best to achieve this and we came up with [`4796f85` (#4190)](https://github.com/Kong/insomnia/pull/4190/commits/4796f852a7b7f97c01abb0e3e3e066f8747c249a)

PR #4196, opened 2021-11-09
fix: consistently use `index`

dimitropoulos:
closes: INS-1049

just a simple copy/consistency change we noticed in the context of other work.  looping back to update.


develohpanda (approved):
Thanks for not fixing the react-array-index lint warning here 🙌🏽

PR #4197, opened 2021-11-09
allow for an empty string as a workspace name in the header

gatzjames:
Currently when creating/updating a workspace with an empty name the application ends up in the document view and the navigation disappears.

empty-workspace-name

Initially I thought there was something happening in the database but it turns out that the header is expecting a name otherwise it doesn't get rendered. I also tested how empty names get handled in a remote project so I ended up allowing an empty name in the page header!

With this fix:

empty-wrk-name-fix

How workspaces with empty names are displayed in a remote project:
RemoteWorkspaceEmpty

Closes INS-1150

### DIFF
--- a/packages/insomnia-app/app/ui/components/dropdowns/workspace-dropdown.tsx
+++ b/packages/insomnia-app/app/ui/components/dropdowns/workspace-dropdown.tsx
@@ -32,7 +32,7 @@ import { WorkspaceSettingsModal } from '../modals/workspace-settings-modal';
 interface Props {
   activeEnvironment: Environment | null;
   activeWorkspace: Workspace;
-  activeWorkspaceName: string;
+  activeWorkspaceName?: string;
   activeApiSpec: ApiSpec;
   activeProject: Project;
   hotKeyRegistry: HotKeyRegistry;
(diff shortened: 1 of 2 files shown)


develohpanda (approved):
How counter-intuitive but this looks good to me.

Alongside this (in another PR or within this one), we should probably update the `WorkspaceSettingsModal`, and the rename function from the dashboard, to prevent an empty workspace name.

jackkav (approved):
Looks good! 👍

gatzjames:
> How counter-intuitive but this looks good to me.
> 
> Alongside this (in another PR or within this one), we should probably update the `WorkspaceSettingsModal`, and the rename function from the dashboard, to prevent an empty workspace name.

@develohpanda Agree. I went trough that thought process but got stuck on what the default name should be then. As a user that created the empty name workspace I would be surprised to see a `swagger-api.yaml` getting created. So I thought why not eliminate the problem completely? 🤔 Might be a wild idea but I like that in most apps you create something and it becomes `Untitled Document 7` without requiring to type a name first and auto-increments the name to avoid confusion. That way there is no need for a modal with default values etc and the user has one less thing to choose when creating a new doc.
This still leaves the problem of renaming. Most apps won't allow you to save with an empty name (e.g. disable the save button) but in our case the changes on the workspace settings are saved while the user types on the fields 😅 There are a couple of UX issues with this approach imo but changing how this works with an empty name as it is will be difficult and hard to test.

That said I'd suggest moving the `default name` to another issue and merging the fix for now.

develohpanda:
Yes totally!! Let's get this in as-is 😀

PR #4202, opened 2021-11-10
Use explicit filename with tar

develohpanda:
A couple of issues with using tar on Windows...

* I cannot find documentation for what the `-a` flag does, and the archive cannot be opened by Windows Explorer if we archive _without_ the `-a` flag.
* `tar -C ../binaries -a -cf output.zip inso.exe` can be extracted fine by Windows Explorer, but replace `inso.exe` with `.` and the archive fails to extract via Windows Explorer. Is there some different syntax I should be using to say "archive everything in this directory" on Windows?

In any case, this should work fine now. Needs to be validated using the build artifacts created from CI.

Archiving can use a revisit: we use `ditto` on macOS and `tar` for Windows and Linux. We currently produce a zip archive for Windows and macOS, and a `tar.xz` for Linux. We don't need to change any of this until the need for it arises.

Closes INS-1056

### DIFF
--- a/packages/insomnia-inso/src/scripts/artifacts.ts
+++ b/packages/insomnia-inso/src/scripts/artifacts.ts
@@ -11,7 +11,7 @@ const isMac = () => platform === 'darwin';
 const isLinux = () => platform === 'linux';
 const isWindows = () => platform === 'win32';
 
-const getName = () => {
+const getArchiveName = () => {
   const version = getVersion();
   if (isMac()) {
     return `inso-macos-${version}.zip`;
@@ -29,7 +29,7 @@ const getName = () => {
 };
 
 const startProcess = (cwd: ProcessEnvOptions['cwd']) => {
-  const name = getName();
+  const name = getArchiveName();
 
   if (isMac()) {
     return spawn('ditto',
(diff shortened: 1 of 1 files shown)


johnwchadwick:
Note that Windows does not use GNU tar, but bsdtar. You can see the relevant documentation on the [bsdtar man page](https://www.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&format=html).

> -a, --auto-compress
> Use the archive suffix to decide a set of the format and the compressions.

In other words: `-a` tells `bsdtar` to read the filename, and then it implies a `--format` option using the suffix.

As for why `.` is different from `inso.exe`... that requires a bit more effort.

When you compress `.` with `tar`, you get an archive that contains the current folder. Note that this is NOT the same as an archive that contains the _files_ in the current folder. If you peak into the structures of the resulting archive, you will see it legitimately has a `.` folder, and in that `.` folder it has the files from the current directory.

This makes sense with `tar` archives which can contain relative paths, and those relative paths can be expanded on extraction. However, with `zip` it is strange and uncommon. The ancient zip handling code in Windows Explorer likely filters `.` and `..` as it is somewhat common to do this in directory listing code (as `.` and `..` were, historically, actual links stored on disk, and still today are synthesized when listing directories) and this backfires as it hides the folder that actually contains the files. (Note: this is somewhat speculation, as I do not know how the zip handling code in Windows works.) However, many other tools will probably have no issue unzipping the archive, especially on UNIX-likes, since it is common with other archive formats anyways.

I don't believe there is an alternative to passing a list of files.

johnwchadwick (approved):

develohpanda:
Thank you for the detailed explanation!

PR #4219, opened 2021-11-17
Leave binary ad-hoc signed and don't manually sign and notarize

develohpanda:
We don't manually sign, therefore just leave as ad-hoc signed
Closes INS-1073


develohpanda:
image

PR #4220, opened 2021-11-17
Codesign with library option

develohpanda:
Inspired by https://stackoverflow.com/a/60642917

Closes INS-1073


develohpanda:
Fails to run unit tests, try codesigning node-libcurl

image

PR #4231, opened 2021-11-19
fix nedb flakiness with the created-at property

gatzjames:
It appears that the created at property is not reliable when creating models in order.
Since the selector doesn't rely on ordering we now check for the existence of the workspaces and the name resolution instead.


jackkav (approved):
LGTM 🚀

PR #4235, opened 2021-11-21
Plugin's filtered by 'enabled' in OpenAPI config

### LINKED ISSUE #3153: Plugin extension 'enabled' not respected in OpenAPI config
## To Reproduce

Add a plugin extension similar to the following to an OAI spec:

```yaml
      x-kong-plugin-cors:
        enabled: false
        config:
          origins: 
            - "*"
          credentials: false
          preflight_continue: false
```

Then run `inso generate config path/to/file.yml`. Type is not important, as both output formats have the same issue.

## Expected

Since `enabled: false` is present, it would either be present and disabled or not present in the generated output at all.

## Actual

The same output is generated, regardless of whether `enabled` is `true` or `false`.

HarryET:
Closes #3153

### DIFF
--- a/packages/openapi-2-kong/src/declarative-config/services.ts
+++ b/packages/openapi-2-kong/src/declarative-config/services.ts
@@ -135,7 +135,7 @@ export function generateService(server: OA3Server, api: OpenApi3Spec, tags: stri
         tags,
       });
 
-      const plugins = [...regularPlugins, ...securityPlugins];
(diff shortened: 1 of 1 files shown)


develohpanda:
Closing this PR as the root issue has already been resolved but wasn't closed. I really appreciate your efforts, though - please see https://github.com/Kong/insomnia/issues/3153#issuecomment-984164092! 🙌🏽

PR #4244, opened 2021-11-24
Add gRPC Headers (INS-362) (#3667)

filfreire:
Closes #3667 and [INS-362](https://linear.app/insomnia/issue/INS-362/grpc-headersmetadata)

This PR adds initial support for gRPC "Headers" (aka. Metadata).

### How to test

1. Checkout this branch 
2. Run `npm run bootstrap && npm run app-start`
3. Clone https://github.com/filfreire/grpc-example-python and follow instructions to build/run
4. Import this [Insomnia collection](https://raw.githubusercontent.com/filfreire/grpc-example-python/main/grpc_insomnia.json) (URL is `https://raw.githubusercontent.com/filfreire/grpc-example-python/main/grpc_insomnia.json`) (<a href="insomnia://app/import?uri=https://raw.githubusercontent.com/filfreire/grpc-example-python/main/grpc_insomnia.json">Open in Insomnia</a>)

5. (In case you need) you can find the [proto file here](https://github.com/filfreire/grpc-example-python/blob/main/proto/helloworld.proto)
6. For testing client, server and bidi streams, use this [other proto file instead](https://github.com/filfreire/grpc-example-python/blob/main/route_example/route_guide.proto) and use the [route_guide example](https://github.com/filfreire/grpc-example-python#build-and-run-route_guide-example).

### Example recording

https://user-images.githubusercontent.com/11976836/143292010-52b29127-bf8c-458f-b76b-584286ad8ed7.mp4

### Problems

- ~~Only tested unary gRPC requests. Still need to add support and test metadata for server, client and bidirectional streams.~~ **UPDATE**: All types of requests are supported now. 
- ~~There's a tiny UI bug, that I suspect might be general to wherever `KeyValueEditor` is used - if the user switches between environments - and is using (and disabling) variables from two environments, the value field stops being in an error state, but the color stays red (check the video above near the end).~~ UPDATE: Bug has been reported on Linear, see [INS-1245](https://linear.app/insomnia/issue/INS-1245/error-state-styling-issues-in-header-keyvalueeditor)
- ~~There's 2 tests failing, not sure yet how to fix them just yet (maybe @dimitropoulos or @develohpanda you can help me in this case). More details:~~ **UPDATE**: this has been fixed, we had to mock `grpc.Metadata` class similarly to how we mocked other `grpc` lib calls.

## UPDATE, 30 Nov 2021

- Added support to all types of gRPC comms that we support, not just unary streams, but client, server and bidirectional as well.

https://user-images.githubusercontent.com/11976836/144087183-810adcb8-edad-46dd-b401-91364792912e.mp4

### DIFF
--- a/packages/insomnia-app/app/common/common-headers.ts
+++ b/packages/insomnia-app/app/common/common-headers.ts
@@ -1,0 +1,27 @@
+
+import allCharsets from '../datasets/charsets';
+import allMimeTypes from '../datasets/content-types';
+import allEncodings from '../datasets/encodings';
+import allHeaderNames from '../datasets/header-names';
+import { RequestHeader } from '../models/request';
+
+export const _getCommonHeaderValues = (pair: RequestHeader): any[] => {
+  switch (pair.name.toLowerCase()) {
+    case 'content-type':
+    case 'accept':
+      return allMimeTypes;
+
+    case 'accept-charset':
+      return allCharsets;
+
+    case 'accept-encoding':
+      return allEncodings;
+
+    default:
+      return [];
+  }
+};
+
+export const _getCommonHeaderNames = (): any[] => {
+  return allHeaderNames;
+};
--- a/packages/insomnia-app/app/common/common-headers.ts
+++ b/packages/insomnia-app/app/common/common-headers.ts
@@ -5,7 +5,7 @@ import allEncodings from '../datasets/encodings';
 import allHeaderNames from '../datasets/header-names';
 import { RequestHeader } from '../models/request';
 
-export const _getCommonHeaderValues = (pair: RequestHeader): any[] => {
+export const getCommonHeaderValues = (pair: RequestHeader): any[] => {
   switch (pair.name.toLowerCase()) {
     case 'content-type':
     case 'accept':
@@ -22,6 +22,6 @@ export const _getCommonHeaderValues = (pair: RequestHeader): any[] => {
   }
 };
 
-export const _getCommonHeaderNames = (): any[] => {
+export const getCommonHeaderNames = (): any[] => {
   return allHeaderNames;
 };
--- a/packages/insomnia-app/app/network/grpc/index.ts
+++ b/packages/insomnia-app/app/network/grpc/index.ts
@@ -4,7 +4,7 @@ import { ServiceClient } from '@grpc/grpc-js/build/src/make-client';
 
 import { SegmentEvent, trackSegmentEvent } from '../../common/analytics';
 import * as models from '../../models';
-import type { GrpcRequest } from '../../models/grpc-request';
+import type { GrpcRequest, GrpcRequestHeader } from '../../models/grpc-request';
 import callCache from './call-cache';
 import type { GrpcMethodDefinition } from './method';
 import { getMethodType, GrpcMethodTypeEnum } from './method';
@@ -37,6 +37,7 @@ const _makeUnaryRequest = (
     method: { path, requestSerialize, responseDeserialize },
     client,
     respond,
+    metadata,
   }: RequestData,
   bodyText: string,
 ): Call | undefined => {
@@ -50,12 +51,15 @@ const _makeUnaryRequest = (
     return;
   }
 
-  // Make call
+  const grpcMetadata = _parseMetadata(metadata);
+
+  // eslint-disable-next-line consistent-return
   return client.makeUnaryRequest(
     path,
     requestSerialize,
     responseDeserialize,
     messageBody,
+    grpcMetadata,
     callback,
   );
 };
@@ -121,6 +125,7 @@ interface RequestData {
   respond: ResponseCallbacks;
   client: ServiceClient;
   method: GrpcMethodDefinition;
+  metadata: GrpcRequestHeader[];
 }
 
 export const start = async (
@@ -128,6 +133,7 @@ export const start = async (
   respond: ResponseCallbacks,
 ) => {
   const requestId = request._id;
+  const metadata = request.metadata;
   const method = await protoLoader.getSelectedMethod(request);
 
   if (!method) {
@@ -152,6 +158,7 @@ export const start = async (
     client,
     method,
     respond,
+    metadata,
   };
   let call;
 
@@ -278,4 +285,18 @@ const _parseMessage = (
     respond.sendError(requestId, e);
     return undefined;
   }
+};
+
+const _parseMetadata = (
+  metadata: GrpcRequestHeader[]
+): grpc.Metadata => {
+  const grpcMetadata = new grpc.Metadata({});
+  if (metadata) {
+    for (const entry of metadata) {
+      if (!entry.disabled) {
+        grpcMetadata.add(entry.name, entry.value);
+      }
+    }
+  }
+  return grpcMetadata;
 };
--- a/packages/insomnia-app/app/network/grpc/index.ts
+++ b/packages/insomnia-app/app/network/grpc/index.ts
@@ -69,12 +69,15 @@ const _makeClientStreamRequest = ({
   method: { path, requestSerialize, responseDeserialize },
   client,
   respond,
+  metadata,
 }: RequestData): Call | undefined => {
   // Create callback
   const callback = _createUnaryCallback(requestId, respond);
 
+  const grpcMetadata = _parseMetadata(metadata);
+
   // Make call
-  return client.makeClientStreamRequest(path, requestSerialize, responseDeserialize, callback);
+  return client.makeClientStreamRequest(path, requestSerialize, responseDeserialize, grpcMetadata, callback);
 };
 
 const _makeServerStreamRequest = (
@@ -83,6 +86,7 @@ const _makeServerStreamRequest = (
     method: { path, requestSerialize, responseDeserialize },
     client,
     respond,
+    metadata,
   }: RequestData,
   bodyText: string,
 ): Call | undefined => {
@@ -93,16 +97,20 @@ const _makeServerStreamRequest = (
     return;
   }
 
+  const grpcMetadata = _parseMetadata(metadata);
+
   // Make call
   const call = client.makeServerStreamRequest(
     path,
     requestSerialize,
     responseDeserialize,
     messageBody,
+    grpcMetadata,
   );
 
   _setupServerStreamListeners(call, requestId, respond);
 
+  // eslint-disable-next-line consistent-return
   return call;
 };
 
@@ -111,9 +119,12 @@ const _makeBidiStreamRequest = ({
   method: { path, requestSerialize, responseDeserialize },
   client,
   respond,
+  metadata,
 }: RequestData): Call | undefined => {
+  const grpcMetadata = _parseMetadata(metadata);
+
   // Make call
-  const call = client.makeBidiStreamRequest(path, requestSerialize, responseDeserialize);
+  const call = client.makeBidiStreamRequest(path, requestSerialize, responseDeserialize, grpcMetadata);
 
   _setupServerStreamListeners(call, requestId, respond);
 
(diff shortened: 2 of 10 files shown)


develohpanda:
@filfreire I think the tests are likely failing because the grpc dependency is mocked [here](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/__mocks__/%40grpc/grpc-js.ts). I'd be happy to pair next week on this, sorry I couldn't get to it today!

filfreire:
> @filfreire I think the tests are likely failing because the grpc dependency is mocked [here](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/__mocks__/%40grpc/grpc-js.ts). I'd be happy to pair next week on this, sorry I couldn't get to it today!

@develohpanda I've since managed to fix it in [my last commit](https://github.com/Kong/insomnia/pull/4244/commits/b08164173a3466a4d0e27db5b71dcb778eb42621) after some debugging to try and understand how the mocks were being used.

packages/insomnia-app/app/network/grpc/index.ts
```diff
@@ -279,3 +297,17 @@ const _parseMessage = (
     return undefined;
   }
 };
+
+const _parseMetadata = (
+  metadata: GrpcRequestHeader[]
+): grpc.Metadata => {
+  const grpcMetadata = new grpc.Metadata({});
+  if (metadata) {
```
develohpanda: I think we can skip this if-condition because `metadata` isn't optional

packages/insomnia-app/app/network/grpc/index.ts
```diff
@@ -279,3 +297,17 @@ const _parseMessage = (
     return undefined;
   }
 };
+
+const _parseMetadata = (
+  metadata: GrpcRequestHeader[]
+): grpc.Metadata => {
+  const grpcMetadata = new grpc.Metadata({});
```
develohpanda: According to the TS types, the argument for `grpc.Metadata` is optional so `new grpc.Metadata()` is also valid. Is the empty object as a parameter necessary here or can be removed?

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.tsx
```diff
@@ -132,7 +135,21 @@ export const GrpcRequestPane: FunctionComponent<Props> = ({
               />
             </TabPanel>
             <TabPanel className="react-tabs__tab-panel">
-              <h4 className="pad">Coming soon! 😊</h4>
+              <ErrorBoundary key={uniquenessKey} errorClassName="font-error pad text-center">
```
develohpanda: I'd love to hear why you chose to use an `ErrorBoundary` here

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.tsx
```diff
@@ -132,7 +135,21 @@ export const GrpcRequestPane: FunctionComponent<Props> = ({
...
+                  sortable
+                  namePlaceholder="header"
+                  valuePlaceholder="value"
+                  descriptionPlaceholder="description"
+                  pairs={activeRequest.metadata}
+                  isVariableUncovered={isVariableUncovered}
+                  handleRender={handleRender}
+                  handleGetRenderContext={handleGetRenderContext}
+                  handleGetAutocompleteNameConstants={RequestHeadersEditor._getCommonHeaderNames}
+                  handleGetAutocompleteValueConstants={RequestHeadersEditor._getCommonHeaderValues}
```
develohpanda: Are the autocomplete name constants and value constants valid for gRPC?

If so, I think we should create a standalone function and consume that in this component and in `RequestHeadersEditor`, instead of invoking the static method on `RequestHeadersEditor` from this component.

Feels strange to be calling into `RequestHeadersEditor` from here

develohpanda:
This PR is awesome @filfreire!!

> There's a tiny UI bug, that I suspect might be general to wherever KeyValueEditor is used - if the user switches between environments - and is using (and disabling) variables from two environments, the value field stops being in an error state, but the color stays red (check the video above near the end).

Yeah this is an existing bug in CodeEditor and the [nunjucks-tags](https://github.com/Kong/insomnia/blob/d53ef39c9b76aefcfa71630078d838bb95eaad56/packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.ts) extension. It's pretty nasty because that extension does DOM manipulation because of how CodeMirror works, so it could use some love for sure.

One thing that might help is identifying specific reproduction steps and creating a ticket in Linear Triage for it, and we can tidy things up bit by bit

develohpanda (changes requested):
### QA

I tried to follow the instructions [here](https://github.com/filfreire/grpc-example-python#how-to-build-and-run) but build fails with an error. Do I need python installed as well?

image

<details>
<summary>🔴 Overflow doesn't scroll</summary>

Upon looking at other uses of KeyValueEditor it likely needs to be wrapped with a div containing the `scrollable` class

https://www.loom.com/share/ec90b36d3fdd4e1aada6a706f385a65a

</details>

<details>
<summary>🟢 Template tags in the metadata value</summary>

https://www.loom.com/share/6f79cc4f02af47d597166496a718dcac

</details>

<details>
<summary>🔴 Illegal characters in metadata key</summary>

https://www.loom.com/share/a956753d917a4a90ab5193d57eee32b2

</details>

<details>
<summary>🔴 (non-blocking) Version control sync shows -metadata instead of +metadata</summary>

When viewing the changes in the create snapshot modal, the metadata field should say `+metadata` instead of `-metadata`, I think. This bug is likely unrelated to this PR, but is a bug nonetheless.

image

</details>

packages/insomnia-app/app/network/grpc/index.ts
filfreire: True 👍

packages/insomnia-app/app/network/grpc/index.ts
filfreire: Indeed it is not needed, I left it there by chance when I was debugging failed tests, good catch 👍

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.tsx
filfreire: I noticed we were using it as well on Requests Pane (e.g. https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/panes/request-pane.tsx) on the `RequestHeadersEditor` so I decided to follow a similar structure. During development it prevented Insomnia from freezing with an exception when I opened the gRPC Headers section, printing the exception instead.

Not sure in this case, if we should keep it or remove it?

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.tsx
filfreire: @develohpanda I couldn't find a lot of examples of common "headers" for gRPC, so I did the change you mentioned, moved both functions into `packages/insomnia-app/app/common/common-headers.ts` that way we're not depedent on `RequestHeadersEditor` and we can create different common headers for newer implementations.

filfreire:
> This PR is awesome @filfreire!!
> 
> > There's a tiny UI bug, that I suspect might be general to wherever KeyValueEditor is used - if the user switches between environments - and is using (and disabling) variables from two environments, the value field stops being in an error state, but the color stays red (check the video above near the end).
> 
> Yeah this is an existing bug in CodeEditor and the [nunjucks-tags](https://github.com/Kong/insomnia/blob/d53ef39c9b76aefcfa71630078d838bb95eaad56/packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.ts) extension. It's pretty nasty because that extension does DOM manipulation because of how CodeMirror works, so it could use some love for sure.
> 
> One thing that might help is identifying specific reproduction steps and creating a ticket in Linear Triage for it, and we can tidy things up bit by bit

@develohpanda I've identified some steps we can do to reproduce, and reported on  Linear (see [INS-1245](https://linear.app/insomnia/issue/INS-1245/error-state-styling-issues-in-header-keyvalueeditor)

filfreire:
@develohpanda regarding the above review https://github.com/Kong/insomnia/pull/4244#pullrequestreview-819820051

<details><summary>✅   Fixed the Overflow doesn't scroll</summary>

Kapture 2021-12-02 at 17 13 25

</details>

<details><summary>✅   Fixed the Illegal characters in metadata key, now it should show an error similar to when the request body is invalid</summary>
Screenshot 2021-12-02 at 16 37 03
</details>

<details><summary> ❓   Couldn't reproduce the control sync shows `-metadata` instead of `+metadata` issue, in my case it didn't show a `-` or `+` character 🤔 </summary>
Screenshot 2021-12-02 at 16 46 17
</details>

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/index.tsx
develohpanda: I was looking for that as an example and somehow missed it 😢 let's keep it!

packages/insomnia-app/app/common/common-headers.ts
```diff
@@ -0,0 +1,27 @@
+
+import allCharsets from '../datasets/charsets';
+import allMimeTypes from '../datasets/content-types';
+import allEncodings from '../datasets/encodings';
+import allHeaderNames from '../datasets/header-names';
+import { RequestHeader } from '../models/request';
+
+export const _getCommonHeaderValues = (pair: RequestHeader): any[] => {
```
develohpanda: Was there a particular reason you chose the underscore prefix? If not, we're trying to move away from it and it feels reasonable to just use `getCommonHeaderValues` here :)

Would love to see some simple unit tests for this function now that it's just a standalone utlity too (but can be follow-up PR, doesn't have to block this PR)

packages/insomnia-app/app/network/grpc/index.ts
```diff
@@ -279,3 +313,22 @@ const _parseMessage = (
...
+  requestId: string,
+  respond: ResponseCallbacks,
+): grpc.Metadata | undefined => {
+  const grpcMetadata = new grpc.Metadata();
+  for (const entry of metadata) {
+    if (!entry.disabled) {
+      try {
+        grpcMetadata.add(entry.name, entry.value);
+      } catch (err) {
+        respond.sendError(requestId, err);
```
develohpanda, line 328: I think we can write some mocked tests for this, let's pair and do so in a separate PR to not block this one

develohpanda (approved):
Full reproduction steps of the `-metadata` bug: https://www.loom.com/share/edbd6eb17cca40a7a4a82396f32bdaae. It happens because we've introduced a new property on the data model that a previously synced collection doesn't know about. You should have access to the project and collection I'm using in the video there. (To reiterate though I don't think this is a blocker for this PR as I think it's a more underlying problem in the diff-ing)

Looks good to me! I wasn't able to test metadata itself actually flowing through to a backend (the example one in your description wouldn't work on my machine), but the functionality within the app is 👌🏽

packages/insomnia-app/app/network/grpc/index.ts
filfreire: Ok 👍

packages/insomnia-app/app/common/common-headers.ts
filfreire: I've fixed the naming, and also added some simple unit tests 👍

jackkav (approved):
Tested with `docker build --tag grpc-example:latest . && docker run -it --rm -p 50051:50051 --name grpc-example grpc-example:latest`
and the hello world sample, looks great!

develohpanda:
Hey @filfreire! Sorry about the conflicts, they will have come up after merging #4248. Should be fairly straightforward to resolve

filfreire:
> Hey @filfreire! Sorry about the conflicts, they will have come up after merging #4248. Should be fairly straightforward to resolve

@develohpanda No worries, I've resolved the conflicts 👍

PR #4248, opened 2021-11-25
Move Nunjucks rendering into React Context

develohpanda:
> The final approach to this PR has diverged _very_ slightly from this description. Namely, the `NunjucksRenderFunctionContext` has been dropped.

This PR is a follow up to #4243 in an effort to reduce prop-drilling into `CodeEditor`.

## Some background

Nunjucks is a templating framework. Insomnia uses it for templating for plugins and environment variables for example. With any template you give it some context (like the selected environment or workspace or request) and “render” the templated text into a final resolved text before consuming it. These render functions are currently drilled from the top-most component (App), but being used by the bottom-most component (CodeEditor).

### Single prop
Currently, `CodeEditor` (and a few other areas of the code) use the presence of a `handleRender` and `handleGetRenderContext` function in order to enable or disable Nunjucks. Let's call this the single prop approach. For example:

```jsx
<CodeEditor 
  renderNunjucks={enableNunjucks ? handleRenderNunjucks : undefined} 
/> 
```

The single prop approach is used in this code-base (mainly with nunjucks related things). Using this approach _can be useful_ and although personally I feel that it adds confusion, undoing that is a lot of work.

The reason why I feel this approach is confusing is because the function handler isn't _actually_ optional, invoking it is. To use the example above, the `renderNunjucks` callback always exists and is always the same across all uses of the `CodeEditor`, however _some_ instances of `CodeEditor` either a) need nunjucks disabled, or b) need nunjucks configurable.

### Multiple props

An alternative approach is to use multiple props, but that still doesn't remove the `renderNunjucks` prop drilling.

```jsx
<CodeEditor 
  renderNunjucks={handleRenderNunjucks} 
  enableNunjucks={enableNunjucks}
/>
```

## Prop drilling and overwriting

Because of the single prop approach, removing the nunjucks handlers isn't only a prop drilling issue, it's a prop _overwriting_ issue too. This in itself isn't bad, but problem arises from the fact that all instances of `CodeEditor` need to use the same nunjucks handlers, and instead of prop drilling from `App` all the way down to the `CodeEditor` (sometimes through 5-6 components), the `CodeEditor` should be able to access it directly and _separately_, be told not to use it.

## The desired solution

In the single prop approach, by being dependent on the presence of `renderNunjucks`, each `CodeEditor` becomes opt-in. This means if we remove the prop drilling, we need to introduce a prop to make rendering nunjucks opt-in, which leads us to the simple API below.

```jsx
// Enable nunjucks
<CodeEditor enableNunjucks />

// Disable nunjucks
<CodeEditor />
```

## A note on PR scope

This PR doesn't _change_ how `CodeEditor` works with the render functions; `CodeEditor` still expects the single prop approach, however an abstraction has been created in which the `useGatedNunjucks` (more on this below) hook determines if the render functions should be provided to `CodeEditor` or not.

This way we achieve our desired API above (remove prop drilling and make nunjucks opt-in) without changing how the `CodeEditor` consumes the props _just yet_, in order to reduce the surface area.

## How the prop drilling was removed

tl;dr React Context FTW: see more notes [here](https://kentcdodds.com/blog/how-to-use-react-context-effectively) on the approach used. Although verbose, it is safer.

There are two React Contexts for Nunjucks; `NunjucksEnabledContext` and `NunjucksRenderFunctionContext`.

### `NunjucksEnabledContext`

This context stores and shares a value down the React component tree indicating whether Nunjucks is blanket enabled or disabled. This is useful with the `BodyEditor` for example, where we can tell a request to disable nunjucks for the request body. To do this, we simply wrap the body the context provider.

```jsx
const BodyEditor: FC = () => {
  // ...
  return <NunjucksEnabledProvider disable={disableNunjucksInRequestBody}>
    {body}
  </NunjucksEnabledProvider>;
}
```

The `useGatedNunjucks` hook uses this (among other factors) to determine whether nunjucks should be enabled or not.

### `NunjucksRenderFunctionContext`

This contains the render functions themselves, `handleRender` and `handleGetRenderContext`. Both are extracted from `App` and converted to hooks.

`useRenderFunctions` provides the callbacks (and changes the callbacks as dependencies such as the environment, request or workspace change), and these callbacks are inserted into the `NunjucksRenderFunctionContext` to be used by components via  the`useNunjucks` and `useGatedNunjucks` hooks..

### `useNunjucks`

The `useNunjucks` hook is useful where you need guaranteed access to render nunjucks _outside_ of a `CodeEditor`. This is used in-line by components like `OAuth2Auth`, `CookiesModal` and `MarkdownPreview`. Given these are all class components, we can use a wrapper function component to access the hook and render function.

```jsx
export const OAuth2Auth: FC<Omit<Props, 'handleRender'>> = props => {
  const { handleRender } = useNunjucks();
  return <OAuth2AuthInternal {...props} handleRender={handleRender} />;
};

```

### `useGatedNunjucks`

The `useGatedNunjucks` hook is useful to provide conditional access to the render functions. It currently uses two parameters to determine whether it should return the render functions:
1. Whether Nunjucks is enabled via a `NunjucksEnabledProvider` (see `BodyEditor` for an example)
2. Whether Nunjucks is enabled (not disabled) by the hook props

https://github.com/Kong/insomnia/blob/db4b0bec5165bb8401eb8b523de736dc3446bec1/packages/insomnia-app/app/ui/context/nunjucks/use-nunjucks.ts#L9-L43

This hook is currently used by `CodeEditor` and `MarkdownPreview`, and the hook prop is derived from the component prop, where an instance of `CodeEditor` explicitly chooses to opt-in to Nunjucks.

https://github.com/Kong/insomnia/blob/2308f89f9c6ccc8e6b72af94068eccc701e28336/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx#L1322-L1334

## tl;dr

The above explains the guts of it all. To summarise:

1. There is no need to drill `handleRender` and `handleGetRenderContext` anymore
2. A `NunjucksProvider` has been added to `App`, which initializes both Nunjucks contexts (see above)
3. `CodeEditor` has a prop to opt-in enable Nunjucks, called `enableNunjucks`
4. To blanket disable Nunjucks for a child tree of components, use `<NunjucksEnabledProvider disable>{children}</NunjucksEnabledProvider>` anywhere
5. Full-text search result of "HandleRender" goes from 329 results (in 51 files) on develop to 98 results (in 17 files) in this PR

```jsx
<NunjucksEnabledProvider>
  <NunjucksRenderFunctionProvider>
    
    <CodeEditor enableNunjucks />  // enabled
    <CodeEditor />  // disabled

    // Blanket disable nunjucks
    <NunjucksEnabledProvider disable>
      <CodeEditor enableNunjucks />  // disabled
      <CodeEditor />  // disabled
    </NunjucksEnabledProvider>

  </NunjucksRenderFunctionProvider>
</NunjucksEnabledProvider>
```

Closes INS-1228


packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -772,11 +777,11 @@ export class UnconnectedCodeEditor extends Component<Props, State> {
     if (this.props.render) {
       mode = {
         name: 'nunjucks',
-        baseMode: CodeEditor._normalizeMode(rawMode),
+        baseMode: UnconnectedCodeEditor._normalizeMode(rawMode),
```
jackkav, line 795: There seems to be a convention mismatch with 
- CodeEditor + UnconnectedCodeEditor
- OAuth2Auth + OAuth2AuthInternal
Both appear to be signifying a react component and a wrapper usually injecting props, but using different language.

There are probably a few patterns for this in the codebase, which one should this PR use? There are a few that come to mind.
- UnconnectedMyButton + MyButton
- MyButton + MyButtonInternal
- MyButton + MyButtonComponent
- MyButtonWrapper + MyButton
- MyButtonHOC + MyButton
- MyButtonWithRedux + MyButton

https://charlypoly.com/publications/react-components-naming-convention

This article cites the third one, but I'm not fussy which naming pattern I just like consistency. It's also great that each are co-located beside each other so it's not too much trouble to figure out what they mean.

packages/insomnia-app/app/ui/context/nunjucks/index.ts
```diff
@@ -0,0 +1,4 @@
+export { useNunjucks, useGatedNunjucks } from './use-nunjucks';
+export { NunjucksProvider, DisableNunjucks } from './nunjucks-provider';
+export { NunjucksEnabledProvider } from './nunjucks-enabled-context';
+export { useNunjucksEnabled } from './nunjucks-enabled-context';
```
dimitropoulos: I'd like to hear more about your rationale for this index file.  To me, it adds a layer of complexity (literally and figuratively) that has very few benefits for something not exported by the package.

If these functions are so related that we want them to come from the same place (i.e. `app/ui/context/nunjucks`) then why not make a file (i.e. `app/ui/context/nunjucks.ts`) and put them there?  All 4 of these are small enough that it seems reasonable to do that.  Creating the index file makes go-to-definition more complicated, and it makes auto imports difficult, and it also creates a situation for ambiguous imports in the rest of the app where some places go directly to the file and some use the index file (which, we've seen a lot and been undoing with Buttons, for a recent example).

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -92,13 +91,17 @@ const BASE_CODEMIRROR_OPTIONS: CodeMirror.EditorConfiguration = {
...
+  enableNunjucks?: boolean;
   ignoreEditorFontSettings?: boolean;
 }
 
-const mapStateToProps = (state: RootState, { ignoreEditorFontSettings }: OwnProps) => {
+const useDerivedProps = ({ enableNunjucks, ignoreEditorFontSettings }: FCProps) => {
+  const {
+    handleRender,
+    handleGetRenderContext,
+  } = useGatedNunjucks({ disabled: !Boolean(enableNunjucks) });
```
dimitropoulos:
```suggestion
  } = useGatedNunjucks({ disabled: !enableNunjucks });
```

I might be mistaken, but I believe we don't need the Boolean cast here

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
dimitropoulos: It's a pattern in the app that we use this `UnconnectedXYZ` terminology.  I chuckle when I see it because (from my perspective) it's exactly the opposite case.  When you do `export const MyComponent = connect(mapStateToProps)(UnconnectedComponent)` it's actually `UnconnectedComponent` that is _connected_ to redux.  I see it the other way too, that `Unconnected` might have been intended to mean "the thing that needs to be used with `connect` in order to work".  Anyway, hopefully that provides a bit more context as to why we do this.

Screenshot_20211130_132034

___

And besides, my guess is that @develohpanda would like to just pull those static methods out of this component altogether, but he avoided doing so as to restrict scope.  We could also solve the problem here by just using `this`, e.g. `this._isJSON`.  I'm not personally convinced the pseudo-private static method (lol) does a lot for us here.

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -92,13 +91,17 @@ const BASE_CODEMIRROR_OPTIONS: CodeMirror.EditorConfiguration = {
...
-type ReduxProps = ReturnType<typeof mapStateToProps>;
-
-interface OwnProps {
+interface FCProps {
+  enableNunjucks?: boolean;
   ignoreEditorFontSettings?: boolean;
 }
 
-const mapStateToProps = (state: RootState, { ignoreEditorFontSettings }: OwnProps) => {
+const useDerivedProps = ({ enableNunjucks, ignoreEditorFontSettings }: FCProps) => {
```
dimitropoulos: If you can think of a way to inline `useDerivedProps` into `CodeEditorFCWithRef` I think it would aid in readability.  Presently you have to jump up and down the file a bit to have context as to what is actually going on.  I can see from fiddling with it (before making this comment) that it's likely the reason you split it out was so that we could do `type Props = RawProps & ReturnType<typeof useDerivedProps>`.  We're in weird territory here, aren't we, because if we were just using regular classes `typeof mapStateToProps` would be plenty, and if we were using hooks then the type of the selector would also be plenty, but because we to translate from one to the other it gets a little strange.

All the same - I'm cool with what's here but just wanted to point out that if you can think of a better way, it'd be cool.

packages/insomnia-app/app/ui/components/rendered-text.tsx
```diff
@@ -1,15 +1,18 @@
...
-  render: (...args: any[]) => any;
+  render: HandleRender;
 }
 interface State {
   renderedText: string;
   error: string;
 }
 
-export class RenderedText extends PureComponent<Props, State> {
+class InternalRenderedText extends PureComponent<Props, State> {
```
dimitropoulos:
```suggestion
class RenderedTextInternal extends PureComponent<Props, State> {
```

there are 8 or so others like this that all put `Internal` at the end.  If we go with this approach for this PR I think we should keep the name consistent here with those others.

packages/insomnia-app/app/ui/components/editors/auth/asap-auth.tsx
```diff
@@ -33,7 +34,7 @@ interface Props {
 }
 
 @autoBindMethodsForReact(AUTOBIND_CFG)
-export class AsapAuth extends PureComponent<Props> {
+class AsapAuthInternal extends PureComponent<Props> {
```
dimitropoulos, line 37: If we move forward with the approach for getting access to the nunjucks hook, I'd ask for an immediate followup ticket for `AsapAuthInternal`, `OAuth1AuthInternal`, `OAuth2AuthInternal`, `KeyValueEditorRowInternal`, `MarkdownPreviewInternal`, `CookiesModalInternal`, `InternalRenderedText`, `TagEditorInternal`, and `VariableEditorInternal` all being converted to proper function components and for this wrapping layer to be removed.

packages/insomnia-app/app/ui/components/wrapper.tsx
```diff
@@ -539,40 +534,31 @@ export class Wrapper extends PureComponent<WrapperProps, State> {
...
-              handleRender={handleRender}
-              handleGetRenderContext={handleGetRenderContext}
               workspaces={workspaces}
               isVariableUncovered={isVariableUncovered}
             />
 
             {activeWorkspace ? <>
               {/* TODO: Figure out why cookieJar is sometimes null */}
               {activeCookieJar ? <>
+                {/* TODO: Why isn't CookiesModal registered? :O */}
```
dimitropoulos: indeed: looks like you removed it by accident in [`e7cff3c` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/e7cff3c612252c75d78ea1c36b0fb01ff107c176#diff-af7e9001ce392681816ece64171c85c0ae80449a3d55018201e61295859376adL584)

all of the other ones used `handleGetRenderContext` so were two line deletes, but this one didn't so the 2nd line happened to be the ref.

packages/insomnia-app/app/ui/components/modals/workspace-environments-edit-modal.tsx
```diff
@@ -581,8 +576,6 @@ export class WorkspaceEnvironmentsEditModal extends PureComponent<Props, State>
                 key={`${this.editorKey}::${selectedEnvironment ? selectedEnvironment._id : 'n/a'}`}
                 environmentInfo={environmentInfo}
                 didChange={this._didChange}
-                render={render}
-                getRenderContext={getRenderContext}
                 isVariableUncovered={isVariableUncovered}
```
dimitropoulos, line 579: I'm so happy, as an aside, to be getting rid of inconsistencies like this.  As the saying goes...

> the best API is no API

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
```diff
@@ -0,0 +1,50 @@
...
+export const useRenderFunctions = () => {
+  const getRenderContextPromiseCache = useMemo(() => ({}), []);
+
+  const environmentId = useSelector(selectActiveEnvironment)?._id;
+  const request = useSelector(selectActiveRequest);
+  const workspace = useSelector(selectActiveWorkspace);
+
+  const fetchRenderContext = useCallback(async () => {
+    const ancestors = await getRenderContextAncestors(request || workspace);
+    return getRenderContext({
```
dimitropoulos: this isn't a criticism of your efforts in this PR, or this line in particular because I know 90% of the code in this file was simply ported from `App.tsx`, but this line exactly is where I, as a reader, get lost, as it calls out to some other big complex thing (i.e. `getRenderContext`) that has a bunch of its own things going on.  There's so many layers of abstraction going on here.  This PR makes the situation better, MOST DEFINITELY, but I wonder what else we can do to make such a critical path of the app make more intuitive sense.

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -488,48 +483,6 @@ class App extends PureComponent<AppProps, State> {
...
-    const context = await this._fetchRenderContext();
-    const keys = getKeys(context, NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME);
-    return {
-      context,
-      keys,
-    };
-  }
-
-  /**
-   * Heavily optimized render function
```
dimitropoulos, line 511: so... about this comment.  It's hard to know exactly what this is ultimately getting at, but I just wanted to ask (since it's been removed) if there's any more context you have on what's here to describe what this is referring to.

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -488,48 +483,6 @@ class App extends PureComponent<AppProps, State> {
...
-   *
-   * @param text - template to render
-   * @param contextCacheKey - if rendering multiple times in parallel, set this
```
dimitropoulos, line 514: it would be great to preserve these two notes

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
jackkav: The name Unconnected seems to couple the redux HOC with the React component being composed with it. I get that it's conventional now, and should be too concerned within in this PR scope, but it still reads weird to me. Probably distracting conversation away from the point of the PR, I'll resolve it.

develohpanda: Thanks for raising it, it super valid because a) this component is no longer using the redux connect function, b) you're right that it couples the component name with redux and c) inconsistent with the "internal" terminology too.

In a follow-up PR I'll address all of them and make consistent. I like the "internal" word because it's explicitly says its internal and shouldn't be exported. So something like `ComponentInternal` (my preference) or `InternalComponent`.

packages/insomnia-app/app/ui/context/nunjucks/nunjucks-enabled-context.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const NunjucksEnabledProvider: FC<Props> = ({ disable, children }) => {
+  const [enabled, setEnabled] = useState(!Boolean(disable));
+
+  useEffect(() => {
+    setEnabled(!Boolean(disable));
+  }, [disable]);
+
+  return (
+    <NunjucksEnabledContext.Provider value={{ enabled }}>
```
dimitropoulos: One thing that jumps out at me when first reading this is to make this suggestion to just remove the need for secondary state (i.e. with `useState` and `useEffect` watching that state) altogether.

```suggestion
    <NunjucksEnabledContext.Provider value={{ enabled: !disable }}>
```

If I had to guess, the rationale might have been to avoid a re-render, but I don't think that's what will happen (unless I missed something).

Here's the react docs https://reactjs.org/docs/context.html#contextprovider

> All consumers that are descendants of a Provider will re-render whenever the Provider’s value prop changes. The propagation from Provider to its descendant consumers (including .contextType and useContext) is not subject to the shouldComponentUpdate method, so the consumer is updated even when an ancestor component skips an update.

The next part is significant, too:

> Changes are determined by comparing the new and old values using the same algorithm as Object.is.

So, in this case:

```ts
Object.is({ enabled: true }, { enabled: true }) // => false
```

Thoughts?  Do you know of a way we could test whether using `useState`+`useEffect` here actually does anything?

packages/insomnia-app/app/ui/context/nunjucks/use-nunjucks.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  return {};
+};
+
+const shouldEnableNunjucks = ({ enabledByProvider, enabledByProp }: { enabledByProvider: boolean; enabledByProp: boolean }) => {
+  // context: disabled, prop: disabled -> should disable
+  // context: disabled, prop: enabled  -> should disable
+  // context: enabled,  prop: disabled -> should disable
+  // context: enabled,  prop: enabled  -> should enable
```
dimitropoulos: would love to see some tests on this

packages/insomnia-app/app/ui/context/nunjucks/use-nunjucks.ts
```diff
@@ -0,0 +1,43 @@
+import { useNunjucksEnabled } from './nunjucks-enabled-context';
+import { useNunjucksRenderFuncs } from './nunjucks-render-function-context';
+
+/**
+ * Access functions useful for Nunjucks rendering
+ */
+export const useNunjucks = () => useNunjucksRenderFuncs();
```
dimitropoulos: is this re-call here because of the module-level state (i.e. `NunjucksRenderFunctionContext`) in `./nunjucks-render-function-context`?  If so take a look at https://github.com/Kong/insomnia/pull/4248/files?w=0#r759373287.  I think this is a concrete example of complexity introduced by separating things into files that are closely coupled anyway.  Like always, lemme know if I missed something.

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
develohpanda: Addressed in [`48f214d` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/48f214dc07fab2b7812ab17515f91fd874dfe464) by moving the FC closer to `useDerivedProps`

packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
develohpanda: Nice catch [`bc423ee` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/bc423eec98900cdda8112a2ffddd7096a4fbccca)

packages/insomnia-app/app/ui/components/editors/auth/asap-auth.tsx
develohpanda: https://linear.app/insomnia/issue/INS-1235#comment-36ee9bd2

packages/insomnia-app/app/ui/components/wrapper.tsx
develohpanda: Oops! Thanks for looking into it <3

packages/insomnia-app/app/ui/components/rendered-text.tsx
develohpanda: 47318dd30362a626ca684f3584da56bd552e75b6

packages/insomnia-app/app/ui/components/wrapper.tsx
develohpanda: 6765bbe52ceea5f332ed2406f1cd7f024ee34bdf

packages/insomnia-app/app/ui/containers/app.tsx
develohpanda: Yep that's my bad, good call. 44d19b55689ad3a60145dbb5e7a8b55402b55bca

packages/insomnia-app/app/ui/context/nunjucks/nunjucks-enabled-context.tsx
develohpanda: This is a great catch. We can (and I will) test this, but we don't need `useState` and `useEffect` here. f8c168e4f145d2dc698f632d2dd1517794c0df00

packages/insomnia-app/app/ui/context/nunjucks/use-nunjucks.ts
develohpanda: Fixed in 19f541e72bb995cf0da95d72af03c6546793a914

packages/insomnia-app/app/ui/context/nunjucks/index.ts
develohpanda: Fixed in 19f541e72bb995cf0da95d72af03c6546793a914. Not resolving this yet as I intend to write a readme (similar to the PR description here).

packages/insomnia-app/app/ui/context/nunjucks/nunjucks-enabled-context.tsx
develohpanda: Test added in [`958c7db` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/958c7db54743a034b468001ab188a612b7d8f7a5)

packages/insomnia-app/app/ui/context/nunjucks/use-nunjucks.ts
develohpanda: Added in [`1433836` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/143383614bae8ac11ce58ba7d867eb590d1edd6b)

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
```diff
@@ -0,0 +1,63 @@
+import { useCallback, useMemo } from 'react';
+import { useSelector } from 'react-redux';
+
+import { getRenderContext, getRenderContextAncestors, HandleGetRenderContext, HandleRender, render } from '../../../common/render';
+import { NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME } from '../../../templating';
+import { getKeys } from '../../../templating/utils';
+import { selectActiveEnvironment, selectActiveRequest, selectActiveWorkspace } from '../../redux/selectors';
+
+/**
+ * @deprecated this hook is intended to only be used by `NunjucksRenderFunctionProvider`
```
gatzjames: 🤔 Curious why this is deprecated before it's even created?

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Heavily optimized render function
+   *
+   * @param text - template to render
+   * @param contextCacheKey - if rendering multiple times in parallel, set this
+   * @returns {Promise}
+   * @private
+   */
+  const handleRender: HandleRender = useCallback(async <T>(obj: T, contextCacheKey: string | null = null) => {
```
gatzjames: We could use undefined instead of null here
```suggestion
  const handleRender: HandleRender = useCallback(async <T>(obj: T, contextCacheKey?: string) => {
```
 or even better not accept an undefined cache key in the function

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!contextCacheKey || !getRenderContextPromiseCache[contextCacheKey]) {
+      // NOTE: We're caching promises here to avoid race conditions
+      // @ts-expect-error -- TSCONVERSION contextCacheKey being null used as object index
+      getRenderContextPromiseCache[contextCacheKey] = fetchRenderContext();
+    }
+
+    // Set timeout to delete the key eventually
+    // @ts-expect-error -- TSCONVERSION contextCacheKey being null used as object index
+    setTimeout(() => delete getRenderContextPromiseCache[contextCacheKey], 5000);
```
gatzjames: I think we should reevaluate the usage of setTimeout here and the promise cache invalidation.
One problem I see with the current implementation is that the timeout is not reset if we store the same `contextCacheKey` which can lead to race condition trouble here.
Imo if we need the ability to invalidate records on the cache it would be better if that functionality is on the cache layer itself.
e.g.
```ts
// Imaginary cache api that handles timeouts and resets the timeout if the same key is stored
cache.set(key, data, { timeout: 5000 })

// A few moments later (<5s)
let contextEarly = await cache.get(key) // data

// A few moments later than before (>5s)
let contextLate = await cache.get(key) // null
```

As a quick solution for this we could store the timeoutId with the contextCacheKey and clear the timeout if it's being reused:
e.g.
```ts
const contextPromiseTimeoutCache = new Map<string, string>();

const previousTimeoutId = contextPromiseTimeoutCache.get(contextCacheKey);

if (previousTimeoutId) {
  clearTimeout(previousTimeoutId)
}

const timeoutId = setTimeout(() => delete getRenderContextPromiseCache[contextCacheKey], 5000);

contextPromiseTimeoutCache.set(contextCacheKey, timeoutId)
```

packages/insomnia-app/app/ui/context/nunjucks/use-gated-nunjucks-render-functions.ts
```diff
@@ -0,0 +1,38 @@
...
+ *  1. Nunjucks is not diabled via the hook props
+ *  2. Nunjucks is not disabled by the last NunjucksEnabledProvider in the React tree
+ *
+ * For ungated access, use `useNunjucksRenderFunctions` instead
+ */
+export const useGatedNunjucksRenderFunctions = (props: { disabled?: boolean } = {}): Partial<ReturnType<typeof useNunjucksRenderFunctions>> => {
+  const { enabled } = useNunjucksEnabled();
+  const funcs = useNunjucksRenderFunctions();
+
+  const shouldEnable = shouldEnableNunjucks({
```
gatzjames: 🤔 I think that this function could be replaced with an inline check:
```suggestion
const isNunjucksTemplatingEnabled = enabled && !props.disabled
```

packages/insomnia-app/app/ui/components/sidebar/sidebar-request-row.tsx
```diff
@@ -332,6 +333,11 @@ const mapStateToProps = (state: RootState) => ({
   activeEnvironment: selectActiveEnvironment(state),
 });
 
-const source = DragSource<Props, DnDDragProps, DragObject>('SIDEBAR_REQUEST_ROW', dragSource, sourceCollect)(UnconnectedSidebarRequestRow);
+export const UnconnectedSidebarRequestRowWithRender: FC<Omit<Props, 'handleRender'>> = props => {
```
gatzjames: Tested this and seems like the drag'n'drop api expects a ref to be passed otherwise reordering won't work 😅. Using forwardRef and passing the ref to the class component seems to do the magic!

gatzjames (changes requested):
👏 This is some awesome work!
Found an issue with forwarding refs so we should double check and make sure that there's no ref expectation for the components that we moved to functions or if there is that we forward the ref properly.

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
develohpanda: I agree with all of this ❤️ It relates to one of dimitri's comments above as well where this whole logic is a bit confusing: https://github.com/Kong/insomnia/pull/4248#discussion_r759592758.

I mainly copy-pasted it from app.tsx and don't want to change it (in this PR), but let's address in a follow-up PR? I'll create a ticket for us to revisit this logic: INS-1224.

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
develohpanda: Same as https://github.com/Kong/insomnia/pull/4248#discussion_r760670168, just refactored as-is from app.tsx

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
develohpanda: @jackkav & @dimitropoulos and I talked about this on a call yesterday. It's because we need to export this to test it and use it in `NunjucksRenderFunctionProvider`, but it should _only_ be used by `NunjucksRenderFunctionProvider`, as the context provider provides a "cached" instance to all consumers via `useNunjucksRenderFunctions` and `useGatedNunjucksRenderFunctions`.

I'd love to chat about any other suggestions you have for how we can improve this.

_something something in js all exports are created equal_ :(

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
develohpanda: Follow-up INS-1224.

I want to write _some_ tests for this at least for the current state, I'll document any findings, but the ticket above is a larger follow-up to address some of James' comments below too.

How does that sound?

packages/insomnia-app/app/ui/containers/app.tsx
develohpanda: INS-1224. As mentioned above, I'll document what I find as I write some tests still.

packages/insomnia-app/app/ui/context/nunjucks/use-gated-nunjucks-render-functions.ts
develohpanda: I addressed this in [`778ef70` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/778ef70d8958888dcbdd35599b9d0d58be0cfd02), let me know what you think!

packages/insomnia-app/app/ui/components/sidebar/sidebar-request-row.tsx
develohpanda: Wow that's not misleading at all 🤯 Thanks for being so diligent in QA for this. I'll double check other components, but at least passing a ref _manually_ into a FC actually fails 🎉 

```tsx
const SomeComponent: FC = () => <div />;

// Not allowed
<SomeComponent ref={...} />
```

develohpanda: At first I did [`4e17fa0` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/4e17fa06250727a100258c80744ac57ef026e410) but opted for [`17eeeb4` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/17eeeb483e905096b418adbcfaf025504da00759) instead. This way we avoid manually working with the ref.

I also removed the generic prop definitions because TS resolves them correctly (and better than manually typing them).

Really good find :)

develohpanda:
> we should double check and make sure that there's no ref expectation for the components that we moved to functions or if there is that we forward the ref properly.

The `Row` component is the only other component that uses DnD  and has a change in this PR and it forwards the refs as expected. (`RequestSidebarGroupRow` also uses DnD and works fine).

packages/insomnia-app/app/ui/context/nunjucks/__tests__/use-render-functions.test.ts
```diff
@@ -0,0 +1,203 @@
...
+  getRenderContextAncestors: jest.fn(),
+}));
+
+const middlewares = [thunk];
+const mockStore = configureMockStore<RootState>(middlewares);
+
+const mockAncestors: PromiseValue<ReturnType<typeof getRenderContextAncestorsMock>> = [];
+const mockContext: PromiseValue<ReturnType<typeof getRenderContextMock>> = { foo: 'bar' };
+
+describe('useRenderFunctions', () => {
```
develohpanda: image

packages/insomnia-app/app/ui/context/nunjucks/__tests__/nunjucks-enabled-context.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { renderHook } from '@testing-library/react-hooks';
+
+import { NunjucksEnabledProvider, useNunjucksEnabled } from '../nunjucks-enabled-context';
+
+describe('NunjucksEnabledProvider', () => {
```
develohpanda, line 5: image

packages/insomnia-app/app/ui/context/nunjucks/__tests__/nunjucks-provider.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { withReduxStore } from '../../../../__jest__/with-redux-store';
+import { RootState } from '../../../redux/modules';
+import { useNunjucksEnabled } from '../nunjucks-enabled-context';
+import { NunjucksProvider } from '../nunjucks-provider';
+import { useNunjucksRenderFunctions } from '../nunjucks-render-function-context';
+
+const middlewares = [thunk];
+const mockStore = configureMockStore<RootState>(middlewares);
+
+describe('NunjucksProvider', () => {
```
develohpanda: image

packages/insomnia-app/app/ui/context/nunjucks/__tests__/nunjucks-render-function-context.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    handleRender: jest.fn().mockResolvedValue(true),
+    handleGetRenderContext: jest.fn().mockResolvedValue(true),
+  };
+
+  return ({
+    useRenderFunctions: () => renderFunctions,
+  });
+});
+
+describe('NunjucksRenderFunctionProvider', () => {
```
develohpanda: image

packages/insomnia-app/app/ui/context/nunjucks/__tests__/use-gated-nunjucks-render-functions.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    handleRender: jest.fn(),
+    handleGetRenderContext: jest.fn(),
+  };
+
+  return ({
+    useRenderFunctions: () => renderFunctions,
+  });
+});
+
+describe('useGatedNunjucksRenderFunctions', () => {
```
develohpanda: image

develohpanda (commented):
Screenshots of tests so they're easier to process

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
jackkav: I think it's a code smell to indicate to developers which functions to use and not to use, especially misleading them with this new deprecated convention. Instead consider the reason for the indication and resolve that. In this case it appears to be the complexity of the code paths moved over from app.tsx if they were simplifed would it be necessary to house them in their own, "don't use me functions"?

develohpanda: It's not the complexity of the functions having moved out of app.tsx requiring a "don't use me". It's the fact this hook is preparing data for react context and we want consumers to pull from context.

I've tried to in-line the contents of this hook within `NunjucksRenderFunctionProvider` and also within the file itself, to prevent an export entirely, to no avail.

In-lining also adds some degree of indirection when unit testing, but this is less of a concern.

Please share any concrete suggestions for how you might see this working and I'll adapt, otherwise I'll give it another go at in-lining tomorrow but will likely need to pair.

> especially misleading them with this new deprecated convention

Lastly I'd like to add that using a deprecated flag is what we had decided on a call two days ago between @jackkav and @dimitropoulos, so this comment comes as a surprise

gatzjames: Got it!
I think that using `@deprecated` for marking internal functions feels a bit off.
Does this stop the function from being imported elsewhere?
If not I think a comment about it's usage should be enough or we could use an index file in the nunjucks folder and specify which functions etc get exported from there?

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
gatzjames: Definitely for another PR! This is going to be quite some work plus testing etc!

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
develohpanda: I changed this up in [`91d453f` (#4248)](https://github.com/Kong/insomnia/pull/4248/commits/91d453f4ec9c51dedb472892a0ae6da75a7a5693), have a look at the end result (as the commit itself will be a bit verbose). How does it feel now?

I've removed the `NunjucksRenderFunctionProvider` and moved the cache to be a singleton, so anything that needs to render functions only needs to use `useNunjucks` and `useGatedNunjucks`. There's no depreated "don't use me" function anymore.

I've also added a test which will fail if the cache _isn't_ a singleton (the last one here)

image

packages/insomnia-app/app/ui/context/nunjucks/__tests__/use-gated-nunjucks.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    handleRender: jest.fn(),
+    handleGetRenderContext: jest.fn(),
+  };
+
+  return ({
+    useNunjucks: () => funcs,
+  });
+});
+
+describe('useGatedNunjucks', () => {
```
develohpanda, line 19: image

packages/insomnia-app/app/ui/context/nunjucks/__tests__/use-nunjucks.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  getRenderContextAncestors: jest.fn(),
+}));
+
+const middlewares = [thunk];
+const mockStore = configureMockStore<RootState>(middlewares);
+
+const mockAncestors: PromiseValue<ReturnType<typeof getRenderContextAncestorsMock>> = [];
+const mockContext: PromiseValue<ReturnType<typeof getRenderContextMock>> = { foo: 'bar' };
+
+describe('useNunjucks', () => {
```
develohpanda, line 32: image

packages/insomnia-app/app/ui/context/nunjucks/use-render-functions.ts
gatzjames: Looks great! I like that it removes the problem of reusing this fn completely!

gatzjames (approved):
👏 🎉 🙏

PR #4254, opened 2021-11-28
Import multiple envs when importing OpenAPI3

sixdouglas:
changelog(Improvements): Each OpenAPI server is now imported as a sub-environment

When the OpenApi file to import contains multiple servers, import each one as an environment.

Closes #3627


ryanholden8:
Really looking forward to using this, thank you!

packages/insomnia-importers/src/importers/openapi-3.ts
```diff
@@ -110,6 +107,43 @@ const parseDocument = (rawData: string): OpenAPIV3.Document | null => {
...
+  return document.servers
+      .map(server => {
+        const currentServerUrl = getServerUrl(server);
+        const protocol = currentServerUrl.protocol || '';
+
+        // Base path is pulled out of the URL, and the trailing slash is removed
+        const basePath = (currentServerUrl.pathname || '').replace(/\/$/, '');
+        const openapiEnv: ImportRequest = {
+          _type: 'environment',
+          _id: 'env___BASE_ENVIRONMENT_ID___sub',
```
develohpanda: This will cause all of the sub environments to have the same id and insomnia won't see them as unique.

In other areas we append a hash of the dependent property to make the id deterministic and unique. What would be a dependent property here - the URL? Ie. if the server URL changes, a _new_ sub environment will be created.

packages/insomnia-importers/src/importers/openapi-3.ts
```diff
@@ -110,6 +107,43 @@ const parseDocument = (rawData: string): OpenAPIV3.Document | null => {
...
+          parentId: baseEnv._id,
+          name: 'OpenAPI env',
```
develohpanda: Same note as above, the name for each sub environment will be `OpenAPI env`. What would you expect the environment created from each new server to be named?

packages/insomnia-importers/src/convert.test.ts
```diff
@@ -10,3 +10,92 @@ describe('Import errors', () => {
     }
   });
 });
+
+describe('Import OpenApi', () => {
```
develohpanda: Any particular reason you chose to add tests like this instead of adding to the fixtures in packages/insomnia-importers/src/importers/fixtures/openapi3?

develohpanda:
Thank you for the PR! A couple of notes, namely the `_id` and `name` generation for the sub-environments.

develohpanda (changes requested):
Thank you for the PR! A couple of notes, namely the `_id` and `name` generation for the sub-environments.

(The CI errors are linting, which should be fixed up with `npm run lint:fix`).

packages/insomnia-importers/src/importers/openapi-3.ts
sixdouglas: Nice catch. I'll fix it.

packages/insomnia-importers/src/importers/openapi-3.ts
sixdouglas: I'll fix it too

packages/insomnia-importers/src/convert.test.ts
sixdouglas: No particular reason, just didn't understand the test mechanism. So I wrote this one.
I'll do it in the same way as the other tests.

Fishbowler:
This would be really useful for getting Insomnia into my organisation. Is there something I can do to help get this finished? I'm not sure what's left to do.

antonioturdo:
I also hope to have this improvement in Insomnia.

dgmora:
This feels very necessary to work with insomnia when our primary source is OpenAPI. Is there a problem with the PR?

filfreire (approved):
Sorry for the long wait @sixdouglas

Rebased with latest develop and fixed lint and tests.

Tested locally, LGTM! Waiting for someone else from @Kong/team-insomnia to give the 👍

PR #4262, opened 2021-12-01
Make note about servers on a path or operation being ignored clearer

develohpanda:
What it says on the tin. Docs about the overrides that are currently not supported by O2K: https://swagger.io/docs/specification/api-host-and-base-path/#overriding

### DIFF
--- a/packages/openapi-2-kong/README.md
+++ b/packages/openapi-2-kong/README.md
@@ -166,7 +166,7 @@ The service name is set to the following
 
 Upstreams and targets are generated from the `servers` root property.
 
-> Note: In OpenAPI, `servers` can be overridden on the `operation` or `path` object. These overrides are currently not supported and will be ignored while generating config.
+> Note: In OpenAPI, `servers` can be overridden on the `operation` or `path` object. These overrides are currently not supported by openapi-2-kong and will be ignored while generating config.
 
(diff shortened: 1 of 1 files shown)


packages/openapi-2-kong/README.md
```diff
@@ -164,7 +164,9 @@ The service name is set to the following
 
 ### `$.upstreams`
 
-Upstreams and targets are generated from the `servers` root property (note that `servers` defined on `operation` or `path` objects are ignored).
+Upstreams and targets are generated from the `servers` root property.
+
+> Note: In OpenAPI, `servers` can be overridden on the `operation` or `path` object. These overrides are currently not supported and will be ignored while generating config.
```
falondarville:
```suggestion
> Note: In OpenAPI, `servers` can be overridden on the `operation` or `path` object. These overrides are currently not supported in Insomnia and will be ignored while generating config.
```

falondarville: That clarifies the distinction between, "you can do this as per OpenAPI but over here, it won't work as you expect"

develohpanda:
```suggestion
> Note: In OpenAPI, `servers` can be overridden on the `operation` or `path` object. These overrides are currently not supported by openapi-2-kong and will be ignored while generating config.
```

falondarville (approved):
LGTM

PR #4266, opened 2021-12-02
componentize settings renderers

dimitropoulos:
closes: INS-1047

In the initial PR where BooleanSettings was introduced, it was introduced out of present necessity (to be able to use a hook) and the other settings renderers were skipped due to being out of scope of the immediate need.  This PR cleans that up.


packages/insomnia-app/app/ui/components/settings/plugins.tsx
```diff
@@ -167,8 +167,10 @@ export class Plugins extends PureComponent<Props, State> {
   }
 
   async _handleUpdatePluginConfig(pluginName: string, config: PluginConfig) {
-    const { updateSetting, settings } = this.props;
-    await updateSetting('pluginConfig', { ...settings.pluginConfig, [pluginName]: config });
+    const { settings } = this.props;
+    await models.settings.update(this.props.settings, {
+      pluginConfig: { ...settings.pluginConfig, [pluginName]: config },
+    });
```
dimitropoulos, line 173: admittedly, the removal of these two lines is not strictly necessary, but it's what allowed for the removal of `updateSetting={this._handleUpdateSetting}` in the parent component (`SettingsModal`) and since that's exactly what I was already doing, 2 lines in the context of the same thing being done in the rest of this PR seemed ok.  I'm happy to pull out to a separate PR if anyone disagrees.

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
```diff
@@ -23,16 +22,16 @@ const Descriptions = styled.div({
 export const BooleanSetting: FC<{
   /** each element of this array will appear as a paragraph below the setting describing it */
   descriptions?: string[];
-  forceRestart?: boolean;
   help?: string;
-  label: string;
+  label: ReactNode;
   setting: keyof Settings;
+  inputProps?: InputHTMLAttributes<HTMLInputElement>;
```
develohpanda: Because we don't expand `inputProps` into the input, and the only thing we use is the `onChange` handler, I'm hesitant of adding it as a prop here, as it's very misleading.

In my opinion we should have a single `onChange?: InputHTMLAttributes<HTMLInputElement>['onChange']` prop.

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
```diff
@@ -47,19 +46,15 @@ export const BooleanSetting: FC<{
     await models.settings.patch({
       [setting]: checked,
     });
-  }, [setting]);
+    inputProps?.onChange?.(event);
+  }, [setting, inputProps]);
```
develohpanda: If `inputProps` is replaced for a single `onChange` prop, this comment becomes redundant, but I'll share anyway.

Because only `inputProps.onChange` is used by this hook, my understanding is that the dependency array should avoid being dependant on the entire `inputProps` object, and only be dependant on `inputProps.onChange`.

packages/insomnia-app/app/ui/components/settings/enum-setting.tsx
```diff
@@ -0,0 +1,53 @@
...
+  help?: ReactNode;
+  label: string;
+  setting: keyof Settings;
+  values: {
+    name: string;
+    value: T;
+  }[];
+}
+
+export const EnumSetting = <T extends string = string>({
```
develohpanda: What's the rationale behind the default `= string`?

packages/insomnia-app/app/ui/components/settings/text-setting.tsx
```diff
@@ -0,0 +1,70 @@
...
+}> = ({
+  help,
+  label,
+  setting,
+  inputProps = {},
+}) => {
+  const settings = useSelector(selectSettings);
+
+  if (!Object.prototype.hasOwnProperty.call(settings, setting)) {
+    throw new Error(`Invalid number setting name ${setting}`);
```
develohpanda:
```suggestion
    throw new Error(`Invalid setting name ${setting}`);
```

packages/insomnia-app/app/ui/components/settings/text-setting.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const onChange = useCallback(async (event: ChangeEvent<HTMLInputElement>) => {
+    const { value, min, max } = event.currentTarget;
+    let updatedValue: string | number | null | undefined = value;
+
+    if (inputProps.type === 'number') {
```
develohpanda: I realize this comes from the previous implementation, but is there some unnecessary abstraction here? I'll let you decide how/where you'd like to resolve it.

`NumberSetting` uses `TextSetting` changing only one thing, while `TextSetting` contains special-cased code specific for `NumberSetting`. Feels like that special-cased code should exist within `NumberSetting`.

Is there enough overlap between `TextSetting` and `NumberSetting` to justify not having unique implementations inside each component?

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -173,41 +79,46 @@ class General extends PureComponent<Props> {
...
+                  Raw template syntax{' '}
+                  <Tooltip message="Will restart the app" className="space-left">
+                    <i className="fa fa-refresh super-duper-faint" />
+                  </Tooltip>
+                </Fragment>
+              )}
               setting="nunjucksPowerUserMode"
-              forceRestart
+              inputProps={{
+                onChange: restartApp,
```
develohpanda: Why not have this restart be an effect (similar to the font change effects), rather than overriding the `onChange` handler?

packages/insomnia-app/app/ui/containers/app-hooks.tsx
```diff
@@ -1,11 +1,12 @@
...
 
 import { useMenuBarVisibility } from '../hooks/settings-hooks';
+import { useSettingsSideEffects } from '../hooks/use-settings-side-effects';
 import { useSyncMigration } from '../hooks/use-sync-migration';
 
 export const AppHooks: FC = () => {
   useSyncMigration();
   useMenuBarVisibility();
-
+  useSettingsSideEffects();
```
develohpanda, line 8: The `useMenuBarVisibility()` hook is also a settings side-effect. Because these new effects are specific to font settings, I'd vote to keep it restricted - `useFontSettings()`, and placed inside `../hooks/settings-hooks`, or something.

tl;dr theres some divergence in patterns here and I'd like to consolidate them please.

packages/insomnia-app/app/ui/hooks/use-settings-side-effects.ts
```diff
@@ -0,0 +1,26 @@
...
+import { useSelector } from 'react-redux';
+
+import { selectSettings } from '../redux/selectors';
+
+const updateFontStyle = (key: string, value: string | null) => document?.querySelector('html')?.style.setProperty(key, value);
+
+export const useSettingsSideEffects = () => {
+  const settings = useSelector(selectSettings);
+
+  useLayoutEffect(() => {
```
develohpanda, line 35: Is there a particular reason to use `useLayoutEffect` instead of `useEffect`? I know the difference between the two, but want to understand why you chose `useLayoutEffect` here.

packages/insomnia-app/app/ui/components/settings/plugins.tsx
develohpanda: Seems fine to me

develohpanda (changes requested):
Look at how clean that `General` component is now 😍 super excited for this!

### `inputProps`

Now that we're refactoring this work, I have reservations of using the `inputProps` approach. It seemed appropriate in the previous implementation, but not in the new approach. I see the `*Settings` components as proxies into the underlying inputs. By exposing the entire `inputProps` object, we're leaking implementation details out of the component (which is an accepted approach in some cases, but doesn't seem justified or necessary here).

For example, the only props actually used are `min` & `max` (for `NumberSetting`), `placeholder` & `disabled` for `TextSetting` (and `MaskedSetting` by extension), and `onChange` in _one_ place (which can be converted to a side-effect). All of these would fit as standalone props just as well.

> `inputProps.type` is also used by `NumberSetting`, but have a look at https://github.com/Kong/insomnia/pull/4266#discussion_r762654822 regarding the `NumberSetting` and `TextSetting` implementation

I really think that the approach we should use is to _not_ use `inputProps` at all, rather expose individual props that are used by the component, and greatly simplify the interface. If in the future we need the expose the entire props object we can.

The use of `inputProps` is also inconsistent;

1. `MaskedSetting` calls it `props`
2. `TextSetting` and `NumberSetting` defaults it to `{}`
3. `BooleanSetting` doesn't default to anything

What are your thoughts regarding the above?

### `MaskedSetting`

Were you planning on updating `MaskedSetting` to be based on `TextSetting`? This PR or future PR is fine, just double checking in case that was missed.

### Unit tests

I'd love to see some unit tests as part of the new work that's been done in this PR. In particular, I can see:
1. React tests for the components themselves making sure the inputs are configured as expected
2. Tests for the change handlers within each settings component (eg `TextSetting` has several conditionals), ensuring the db is patched as expected
3. React Hook tests for the font side-effects

Happy to pair on any unit tests

### QA

<details>
<summary>🔴 UI jumps around when changing the font settings</summary>

This applies to the font family inputs as well, which was introduced newly since the last release, but we should be using onBlur for those as well. In fact, should _any_ inputs use `onChange` (other than boolean/enum)?

https://www.loom.com/share/549f519ee66246b784ca6195ffd5a12d

</details>

<details>
<summary>🔴 Spacing inconsistencies</summary>

image

</details>

<details>
<summary>🟢 Plugin config works as expected</summary>

Related to the self-review comment

https://www.loom.com/share/3298bf1f357548c693ca82f56160435c

</details>

packages/insomnia-app/app/ui/components/settings/text-setting.tsx
dimitropoulos: :+1: 

fixed in 69b464a8cd7a3dddc1273426d299b8085ffd2c9f

packages/insomnia-app/app/ui/components/settings/enum-setting.tsx
dimitropoulos: interesting.  it was out of two things: API convenience and resisting expanding scope to rename a bunch of things since we had two of the 4 usages of this component that were not actually using enums.  But.. I took a second look and it turns out there's exactly one line of "scope creep" in code editor to switch something to be actual enums.  It's a quite small changeset to make it required, so I did so in b0467354b2cb660443d9321dbce3de48db8f43f1.

Thanks.  Wouldn't have thought to dig into this.

packages/insomnia-app/app/ui/components/settings/text-setting.tsx
dimitropoulos: see bb9f52dd8d4976369b6888af526bf0838fd79b80 and 3dd90b3dbf6668883c6d06ef717e32d90c71442c

This is a great suggestion because it forced me to reevaluate what the overlap really is.  Turns out, it's really the other way around - there's like... almost half... of TextSetting that is specifically concerned with just number stuff.  Now, both are much simpler and only focus on their own thing.  Also, it gave us an opportunity to actually verify that string settings are passed to the string setting thing, and so forth.  Really great.

packages/insomnia-app/app/ui/hooks/use-settings-side-effects.ts
dimitropoulos: @gatzjames was adamant about it, and I just remember that he had a really solid reason at the time but I don't recall it exactly.  @gatzjames please comment.

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
dimitropoulos: totally, see https://github.com/Kong/insomnia/pull/4266#discussion_r763221128

packages/insomnia-app/app/ui/components/settings/boolean-setting.tsx
dimitropoulos: yep :) see https://github.com/Kong/insomnia/pull/4266#discussion_r763221128

packages/insomnia-app/app/ui/containers/app-hooks.tsx
dimitropoulos: > or something

Yep, I totally see where you're coming from.  check out 7d7e8e24ce41c8e8164e92589df81a44f4e58582.  Now it's all consolidated.  Although, yes, the font ones feel more _the same_ than other things we'll add to this hook, I think the ease of reading and simplicity is really nice just having them all in line like this, so seems good.  My next move will be to add the `restartApp` thing there, too, so it's looking like this pattern is working well.  as always: please feel totally free to unresolve if you'd like to talk more into it.

develohpanda: I think these one-liner effects are straightforward enough (and my goodness are they beautiful) to be all together. After all, they are settings side-effects! I like what you've done, natural evolution of the pattern. 😊

packages/insomnia-app/app/ui/components/settings/general.tsx
dimitropoulos: the reason was that I was hoping to follow up on this (i.e. removing the restart requirement altogether) in a future PR.  the other reason was that it's a little more tricky than the others because if you do something like `useEffect(restartApp, [settings.nunjucksPowerUserMode])` you put the app in a restart loop.  I gave it a shot in c5a89329c0b41183f19ccea4a9fdb8a8b5e640cf.  Lemme know your thoughts.

dimitropoulos:
the styling fix was in e68077b69aaf19fcb97bb4ef2bbc58d93327c2e8.  sorry bout that.  since that section isn't shown in Linux I didn't see it while doing my self-review.

dimitropoulos:
re: `onBlur`, sorry, I hadn't updated, but I checked in with @erickrico and we are going to make it `onChange` like every other input

packages/insomnia-app/app/ui/components/settings/general.tsx
develohpanda: Hmm, it seems non-trivial, maybe we can just keep as you had it before (without the effect) as the end goal is actually to remove the need altogether.

> the reason was that I was hoping to follow up on this (i.e. removing the restart requirement altogether) in a future PR

The changes I've got in https://github.com/Kong/insomnia/pull/4276 will remove the need for the restart, so follow-up PR is already on it's way (planned into a future sprint as INS-1245)! 🎉 Speeeeeeeeddd

develohpanda:
Checking that the placeholder was dropped intentionally?

This PR
image

2021.7.1
image

> re: `onBlur`, sorry, I hadn't updated, but I checked in with @erickrico and we are going to make it `onChange` like every other input

I just want to confirm that this is the behavior we are opting for by using `onChange` instead of `onBlur`: https://www.loom.com/share/db90c827ff984da492bcd8916f513029. Doesn't seem very user friendly. How do other tools work (as my opinion might be an exception)? If not `onBlur`, can we use a debounce, or something else?

packages/insomnia-app/app/ui/hooks/use-settings-side-effects.ts
gatzjames: The main reason is that useEffect will run after react has rendered (committed the DOM mutations) so that would mean we might be doing an extra paint creating flicker in the ui. It takes more than a gh comment to explain this so happy to have a chat about it. In general I use this rule of thumb:
Use useLayoutEffect for DOM related updates and useEffect for everything else :)

dimitropoulos:
no, that was not intentional.  after your feedback (which, again, I completely agree with!) and the changes to remove `inputProps` props, this broke because the props were being spread before (and therefor missed in the refactor because it was never written explicitly in the body of the component which, actually, was exactly the kind of bug your feedback suggested would happen! haha.  that didn't take long!! heh).

I still need to take some time and go through this and I wasn't expecting you to review it again before I requested a re-review from you.  Please hold off on further reviews until I request another - I'm hoping to add some tests, too.

__

Great news, though, I went through each of the components again to make sure all the props are used (`disabled` and `placeholder` were missed).  I corrected those, but while looking at it more objectively I discovered that we no longer, in any case for any of these components, need `onChange` props.  That's great news because it's a signal that the side effects are handled by the machinery of the project (i.e. app-hooks) rather than being handled as special-cases in a view component.  Cool.  Check out a7a40ef5c942ff5b59dadc9364799c13f9d464e6.

packages/insomnia-app/app/ui/components/settings/general.tsx
dimitropoulos: soooo... haha.. since d26e9d2b0fe8d247e0be2e6d2afd20b6c22cbea3 a few days ago we removed all `onChange` or `inputProps` props altogether in all of these, and it'd be cool to leave it that way.  beyond that, e8526e49d087f0090decb08c82211787c33d8220 happened after rebasing this on https://github.com/Kong/insomnia/pull/4277 which generalized the code, at least, to not care about which mode it is that it's run in.  since they'll both (i.e. 'nunjucksPowerUserMode', and 'showVariableSourceAndValue') be removed the same way, now, would you mind if we just left it so as to avoid re-implementing the `inputProps` or `onChange` handlers (since, we know we're gonna want to remove those too)?

develohpanda: Ahaha fair. Yeah all good

packages/insomnia-app/app/ui/components/settings/__tests__/boolean-setting.test.tsx
```diff
@@ -0,0 +1,42 @@
+import { render } from '@testing-library/react';
```
develohpanda, line 1: Could you make a follow-up linear ticket for us to go through and add tests for the change handlers? We don't have to in this PR, but should come back and complete it.

packages/insomnia-app/app/ui/components/settings/__tests__/enum-setting.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  it('should render help text', async () => {
+    const { getByText } = render(enumSetting, container);
+    expect(getByText(help)).toBeInTheDocument();
+  });
+
+  it('should be disabled when passed a disabled prop', async () => {
+    const { getByText } = render(enumSetting, container);
+    expect(getByText(values[0].name)).toBeInTheDocument();
+    expect(getByText(values[1].name)).toBeInTheDocument();
+  });
```
develohpanda: This test doesn't seem relevant to this component

packages/insomnia-app/app/ui/components/settings/__tests__/text-setting.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it('should render help text', async () => {
+    const { getByText } = render(textSetting, container);
+    expect(getByText(help)).toBeInTheDocument();
+  });
+
+  it('should be disabled when passed a disabled prop', async () => {
+    const { getByLabelText } = render(textSetting, container);
+    const input = getByLabelText(label).closest('input');
```
develohpanda, line 53: TIL about `closest`. Interesting!

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -1,150 +1,63 @@
...
 import { HelpTooltip } from '../help-tooltip';
 import { BooleanSetting } from './boolean-setting';
+import { EnumSetting } from './enum-setting';
 import { MaskedSetting } from './masked-setting';
+import { NumberSetting } from './number-setting';
+import { TextSetting } from './text-setting';
+
+/**
+ * We are attempting to move the app away from needing settings changes to restart the app.
+ * For now, this component is a holdover until such a time as we are able to fix the underlying cases.
```
develohpanda:
```suggestion
 * For now, this component is a holdover until such a time as we are able to fix the underlying cases. (INS-1245)
```

packages/insomnia-config/src/generated/schemas/insomnia.schema.json
```diff
@@ -10,21 +10,25 @@
       ]
     },
     "settings": {
-      "$ref": "#/definitions/Partial<Pick<Settings,\"allowNotificationRequests\"|\"disableUpdateNotification\"|\"enableAnalytics\"|\"disablePaidFeatureAds\"|\"incognitoMode\">>"
+      "$ref": "#/definitions/Partial<Pick<Settings,\"allowNotificationRequests\"|\"disablePaidFeatureAds\"|\"disableUpdateNotification\"|\"enableAnalytics\"|\"incognitoMode\">>"
```
develohpanda, line 13: Was this change intentional? It's strange that this changed

develohpanda (changes requested):
Code LGTM, comments are non-blocking and minor.

The only thing blocking is the following comment from an earlier review.

> > re: `onBlur`, sorry, I hadn't updated, but I checked in with @erickrico and we are going to make it `onChange` like every other input
> 
> I just want to confirm that this is the behavior we are opting for by using `onChange` instead of `onBlur`: [loom.com/share/db90c827ff984da492bcd8916f513029](https://www.loom.com/share/db90c827ff984da492bcd8916f513029). Doesn't seem very user friendly. How do other tools work (as my opinion might be an exception)? If not `onBlur`, can we use a debounce, or something else?

packages/insomnia-config/src/generated/schemas/insomnia.schema.json
dimitropoulos: it was autogenerated, yes, but it's because they were sorted, is all.

packages/insomnia-app/app/ui/components/settings/__tests__/enum-setting.test.tsx
dimitropoulos: good call: the name was just wrong -> 7849feac1ed81daf232beaeb9551f3de602bb9b6 fixes it.  the test is there just to make sure that it renders the enum options as passed.

dimitropoulos:
> I just want to confirm that this is the behavior we are opting for by using onChange instead of onBlur

confirmed

> Doesn't seem very user friendly.

I know what you mean, but also -> I showed 5 people this input and asked them to change the font size and not one of them was able to do it with the onBlur functionality.  It just doesn't make sense and is not natural. 

For one thing, consider that it was broken before in ways, too.   For example, if you increase the font size and then hit `Data`.  It doesn't switch tabs!

https://user-images.githubusercontent.com/15232461/146014632-fc7e4d80-10a5-42f0-b8a9-3fc81c2df68b.mp4

For another thing, none of the other inputs work like this.  Including the font change, which, can cause such a large change in line height that the input you're working with can easily go off the screen:

https://user-images.githubusercontent.com/15232461/146015030-5d4440ae-b5d7-4dbd-9dc5-051e25d7d735.mp4

> How do other tools work (as my opinion might be an exception)?

I looked, but it's hard to find a tool that offers an _interface_ font size independent of the zoom (including other electron-based API clients).  Even vscode doesn't do this as far as I am aware.  They offer a zoom (which we also offer) which has the same purpose but keeps everything more proportional.

> If not onBlur, can we use a debounce

Yes, I did this in 4854d45ca07a7c97407e9d7df228fe9f7f3dbe99, but it stutters the app on launch.  I reverted it in 04a663e17ef4842c5baeb81532aeca9a1638a103.

> or something else?

I also tried to throttle, but that can come with the same stuttering as debounce on app boot.

packages/insomnia-app/app/ui/components/settings/__tests__/boolean-setting.test.tsx
dimitropoulos: INS-1219

dimitropoulos:
(merged based on `Code LGTM, comments are non-blocking and minor.` above)

PR #4276, opened 2021-12-07
Update Nunjucks Tags in real time

develohpanda:
Super hacky fix to at least get things working and avoid memory leaks (but the memory leaks needs validation)

Demos

https://www.loom.com/share/40ce3de971314a3e8a95fd204680eb7a
https://www.loom.com/share/5af3d16215a84b20ba941ba4ad3c2792

TODO:
- [ ] add some notes

Closes INS-1245

Issue #754, opened 2018-02-12
[Feature Request] Show the source of a variable replacement when hover and click


rocketraman:
- Insomnia Version: 5.14.6
- Operating System: Linux Fedora 27

## Details

I use variables extensively. I have variables at the environment level, a base shared environment, and at the folder level.

This sometimes gets a bit confusing. It would be helpful if, when hovering over a variable placeholder, in addition to the URL a little bit of text like `(from env Foo)` or `(from folder xxx)` was appended. The same information could be shown in the popup when one clicks on the variable.

gschier:
It took some thinking to figure out if this is even possible, but I think it is. 

Here's how I would approach this problem, in case someone wants to submit a PR:

- Add a `getKeysMeta` function to environment that returns a map of path to source name, where source name the name of the environment or folder. See [this](https://github.com/getinsomnia/insomnia/blob/b3e473071824f0e48e7c26507163182acbe01703/packages/insomnia-app/app/common/render.js#L206-L209) for example
- Use `getKeysMeta` function to figure out the source when [setting the `title` attribute](https://github.com/getinsomnia/insomnia/blob/b3e473071824f0e48e7c26507163182acbe01703/packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.js#L250) on the tag

PR #4277, opened 2021-12-07
Allow users to configure what is shown on environment variable template tags

develohpanda:
changelog(Improvements): Now you can configure whether values are shown for environment variable template tags

### Demo

https://www.loom.com/share/0fee8e05315741c5bfe3f8c8c4d19dce

### Summary

This PR does:
* Add a boolean in Preferences > General to uncover variables
* Persist the setting in settings instead of in-memory
* Remove all the prop drilling
* Will swap what is shown in the tool-tip when the option is enabled vs disabled (see demo recording below)

This PR doesn't:
* Resolve the issue where changes to nunjucks settings requires a restart or re-render; #4276 will resolve that issue

In the future:
* We could give users more control over what they want to see in the template tag. The three bits of information we have are:
	* the source (root, or the environment name, or the folder name)
	* the variable name (`foo`, `host`, etc)
	* the value (`bar`, `https://insomnia.rest`)

The initial request for this came from https://github.com/Kong/insomnia/issues/754, was resolved in #1274, but it got lost somewhere. Seems like a valuable thing to have. The only place this is configurable right now is via a hotkey, but I added a setting for it.

TODO:
- [x] QA
- [x] Setting name
- [x] Confirm setting help text

Closes INS-1226


packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
```diff
@@ -1317,7 +1316,6 @@ export class UnconnectedCodeEditor extends Component<Props, State> {
           onMouseLeave={onMouseLeave}
         >
           <textarea
-            key={isVariableUncovered ? 'foo' : 'bar'}
```
dimitropoulos, line 1320: haha, happy to see this go

develohpanda: Lol it was actually causing a bug! During unmount, we do `this.codeMirror.toTextEditor()`, which converts the CodeMirror instance back to a `textarea`. But because of this key, the text area was being forced to change. What's more, this forced change using the `key` actually didn't fix the underlying issue 😂 I'm guessing that's why this thing is such a hidden feature, hidden behind a keyboard shortcut right now.

develohpanda:
@wdawson / @falondarville here's a demo of this feature (which already existed, I just added a setting for it so users can access it).

https://www.loom.com/share/0fee8e05315741c5bfe3f8c8c4d19dce

Here's what the title and tooltip currently show

image

Could you please let me know if you'd like these reworded?

wdawson:
@develohpanda 
> Could you please let me know if you'd like these reworded?

Is "Show variable values" clearer for the title? The tooltip could read: "Shows the source and value of the variables instead of the variable name on the template tags"

CC - @falondarville

falondarville:
@develohpanda @wdawson thoughts?

If checked, reveals the environment variable source and value in the template tag. Otherwise, hover over the template tag to see the source and value.

develohpanda:
I like that too, the fallback! It's informative. @falondarville Does that use Wils's title suggestion of "Show variable values"?

falondarville:
Should it be "Show variable source and values"? I lean towards wanting to be explicit and we have some longer check box titles, so it won't look off

wdawson:
> Should it be "Show variable source and values"? I lean towards wanting to be explicit and we have some longer check box titles, so it won't look off

Agree! As long as there's the space to do it.

develohpanda:
image

Looks good now!

packages/insomnia-app/app/ui/components/editors/grpc-editor.tsx
```diff
@@ -13,7 +13,6 @@ export const GRPCEditor: FunctionComponent<Props> = ({
   content,
   handleChange,
   readOnly,
-  isVariableUncovered,
```
dimitropoulos, line 16: missed this being also removed from `Props` above

dimitropoulos: (fixed in c9c22da0f8118643658af239b3018bc885f37acc if you wanna just pull it)

dimitropoulos (commented):
noticed we're missing typings in the nunjucks extension for isVariableUncovered.  very tempted to add the handle render types as well, but I didn't.. heh.  b87a87c3c399800955a900a5894c434f521e4b06 did it if you wanna take a look (it's outside of the line scope of this changeset on GitHub so I can't comment on it here)

packages/insomnia-app/app/ui/components/key-value-editor/key-value-editor.tsx
```diff
@@ -427,7 +427,6 @@ export class KeyValueEditor extends PureComponent<Props, State> {
       valuePlaceholder,
       namePlaceholder,
       descriptionPlaceholder,
-      isVariableUncovered,
```
dimitropoulos, line 430: the type in `Props` at the top of the file needs to go too, did it in 03dd1bfcf969a0da176be141ee3bc4e90d3a95d6 if you wanna pull it

packages/insomnia-app/app/ui/components/key-value-editor/row.tsx
```diff
@@ -274,7 +274,6 @@ class KeyValueEditorRowInternal extends PureComponent<Props, State> {
       forceInput,
       descriptionPlaceholder,
       pair,
-      isVariableUncovered,
```
dimitropoulos, line 277: (same as others, e6ff2c3924f6932964164889d39e9a99eb5aaa9b)

packages/insomnia-app/app/ui/components/viewers/grpc-tabbed-messages.tsx
```diff
@@ -35,7 +35,6 @@ export const GrpcTabbedMessages: FunctionComponent<Props> = ({
   handleCommit,
   handleStream,
   uniquenessKey,
-  isVariableUncovered,
```
dimitropoulos, line 38: (same as others 940af34e5539cf52ce749a111f8f8a3cbe9b24c2)

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -347,12 +345,7 @@ class App extends PureComponent<AppProps, State> {
         },
       ],
       [hotKeyRefs.PLUGIN_RELOAD, this._handleReloadPlugins],
-      [
-        hotKeyRefs.ENVIRONMENT_UNCOVER_VARIABLES,
-        async () => {
-          await this._updateIsVariableUncovered();
-        },
-      ],
+      [hotKeyRefs.ENVIRONMENT_UNCOVER_VARIABLES, this._updateIsVariableUncovered],
```
dimitropoulos: so, looking at this, this won't totally work, yet, because we need to do some work to get the current restart requirement to go away, right?  I'm on board with _not_ fixing that in this PR -> just wanted to make sure I'm understanding it correctly.  If that's the case, can you please make a ticket for fixing that part so we can triage for the future?

packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.ts
```diff
@@ -309,8 +309,11 @@ async function _updateElementText(render, mark, text, renderContext, isVariableU
       const con = context.context.getKeysContext();
       const contextForKey = con.keyContext[cleanedStr];
       // Only prefix the title with context, if context is found
-      title = contextForKey ? `{${contextForKey}}: ${title}` : title;
-      innerHTML = isVariableUncovered ? title : cleanedStr;
+      const valueAndContext = contextForKey ? `{${contextForKey}}: ${title}` : title;
+
+      // Swap what's shown in the tooltip vs the innerHTML
+      innerHTML = isVariableUncovered ? valueAndContext : cleanedStr;
```
dimitropoulos: I wish there was some way to not have to do all the drilling inside of this extension code.  There's no "react" here so a hook selector makes no sense... so it is what it is I guess.  curious, though, if you've thought of anything (even, that we could potentially try in the future) to get this extension code to be less of an ugly duckling in terms of its capabilities compared to the rest of our codebase.

packages/insomnia-app/app/ui/components/codemirror/one-line-editor.tsx
```diff
@@ -347,7 +347,6 @@ export class OneLineEditor extends PureComponent<Props, State> {
       placeholder,
       onPaste,
       getAutocompleteConstants,
-      isVariableUncovered,
```
dimitropoulos, line 350: (same as others, see 3be981695e9cec8bb7699bec3dbfbdd4189f1258)

dimitropoulos (approved):
it's really incredible.  TWO HUNDRED and THIRTY NINE usages in the app before, when all we ever needed was:

- 1 usage for the settings tab in preferences (which, haha, we didn't bother having before, thanks for adding!!)
- 1 usage for the hotkey execution logic
- 2 usages in the nunjucks-tags extension

and.... that's it!

wonderful! wonderful wonderful!

___

I pushed some commits to https://github.com/dimitropoulos/insomnia/commits/feature/ins-1226-remove-isvariableuncovered-prop-drilling referenced in the feedback.

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -160,6 +160,13 @@ class General extends PureComponent<Props> {
               setting="forceVerticalLayout"
               help="Stack application panels (e.g. request / response) vertically instead of horizontally."
             />
+
+            <BooleanSetting
+              label="Show variable source and values"
+              help="If checked, reveals the environment variable source and value in the template tag. Otherwise, hover over the template tag to see the source and value."
+              setting="isVariableUncovered"
```
jackkav: nice to have a variable name that matches the new label somewhat, but the migration may not be worth the benefit now its usage is simpler to understand. Besides that, nice job.

jackkav (approved):

packages/insomnia-app/app/ui/containers/app.tsx
develohpanda: That's right, using the hotkey exhibits the same behavior as only changing the option in preferences and not restarting the app. I've got https://github.com/Kong/insomnia/pull/4276 as a spike (backed by INS-1245 in cycle 59) to address removing the restart requirement by fixing the underlying bug.

packages/insomnia-app/app/ui/components/codemirror/extensions/nunjucks-tags.ts
develohpanda: I'm planning to review the extension code in #4276, and will make this nicer to use. INS-1245

develohpanda:
@dimitropoulos thank you for the commits! Have pulled them in 💯

packages/insomnia-app/app/ui/components/settings/general.tsx
develohpanda: Great point 👍🏽 renamed to `showVariableSourceAndValue`. The good thing is because this didn't previously exist in the database, no migration was necessary, just a find and replace in code. 2e5d63952123d0da94f1fa902196712d7ee412a9

dimitropoulos: GREAT point @jackkav.  speaking personally, I think my head was too deep in the current falling sands of low-level concepts to notice this high-level thing.

PR #4280, opened 2021-12-07
adds more documentation to codemirror addon (and fixes type error)

dimitropoulos:
whilst reviewing https://github.com/Kong/insomnia/pull/4275 I uncovered the reason the dependnecy was in this package.  Yes, yes, it's true that it was written there before, but hopefully this time it'll be more clear.  @jackkav and I both did not understand the previous comment in sufficient detail to know that it's a _global_ required by a dependency that's at play here.  Hopefully the added documentation speaks for itself - if not, we need to adjust it.


jackkav (approved):
Thanks for clearing this up!

PR #4290, opened 2021-12-09
gRPC IPC action handler unit tests for working with IPC

develohpanda:
Inspired by a recent bug where the wrong event was being fired for cURL IPC


packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+jest.mock('electron');
+jest.mock('../../../../../network/grpc/prepare', () => {
+  const module = {
+    prepareGrpcMessage: jest.fn(),
+    prepareGrpcRequest: jest.fn(),
+  };
+
+  return module;
+});
```
dimitropoulos: Screenshot_20211210_101923

Besides one (pictured on the right above) which seems to need to do it so it can type the object more ergonomically, this would be the only module where we don't just return an object literal.  It'd be good to favor consistency
```suggestion
jest.mock('../../../../../network/grpc/prepare', () => ({
  prepareGrpcMessage: jest.fn(),
  prepareGrpcRequest: jest.fn(),
}));
```

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
```diff
@@ -0,0 +1,78 @@
...
+const messageParamsBuilder = createBuilder(grpcIpcMessageParamsSchema);
+const requestParamsBuilder = createBuilder(grpcIpcRequestParamsSchema);
+
+describe('useActionHandlers', () => {
+  it('should send start', async () => {
+    const requestParams = requestParamsBuilder.build();
+
+    prepareGrpcRequest.mockResolvedValue(requestParams);
+
+    const { result } = renderHook(() => useActionHandlers(requestParams.request._id, 'env', GrpcMethodTypeEnum.unary, mockGrpcDispatch));
```
dimitropoulos:
```suggestion
    const { result } = renderHook(() => useActionHandlers(
      requestParams.request._id,
      'env',
      GrpcMethodTypeEnum.unary,
      mockGrpcDispatch
    ));
```

One suggestion was going to make was that it get's hard to read out there this long, but, ultimately, the real feedback would be to give `useActionHandlers` named arguments (i.e. an object).  That's clearly out of scope for this PR so I'll just mention it and maybe we can update in the future.

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    const { result } = renderHook(() => useActionHandlers(messageParams.requestId, 'env', GrpcMethodTypeEnum.client, mockGrpcDispatch));
+
+    await result.current.stream();
+
+    expect(ipcRenderer.send).toHaveBeenCalledWith(GrpcRequestEventEnum.sendMessage, messageParams);
+    expect(mockGrpcDispatch).toHaveBeenCalledWith(expect.objectContaining({ type: GrpcActionTypeEnum.requestMessage }));
+  });
+
+  it('should send cancel', () => {
+    const requestId = 'req';
```
dimitropoulos, line 68: here's the code of `_sendCancel`:
```ts
const _sendCancel = (reqId: string) => {
  if (!reqId) {
    return;
  }

  ipcRenderer.send(GrpcRequestEventEnum.cancel, reqId);
};
```

In this test we're testing the 2nd branch, but it'd be cool to also test the first one if we can (where if the requestId is an empty string it won't `ipcRenderer.send`.  That goes for all of the actions (send, cancel, stream, and commit).

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    expect(ipcRenderer.send).toHaveBeenCalledWith(GrpcRequestEventEnum.start, requestParams);
+    expect(mockGrpcDispatch).toHaveBeenCalledWith(expect.objectContaining({ type: GrpcActionTypeEnum.clear }));
+  });
+
+  it('should send stream', async () => {
+    const messageParams = messageParamsBuilder.build();
+
+    prepareGrpcMessage.mockResolvedValue(messageParams);
+
+    const { result } = renderHook(() => useActionHandlers(messageParams.requestId, 'env', GrpcMethodTypeEnum.client, mockGrpcDispatch));
```
dimitropoulos, line 56: regarding `GrpcMethodTypeEnum`, this file only tests `client` and `unary`.  Is it in scope to also test `bidi` or `server`?

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
```diff
@@ -0,0 +1,78 @@
+import { createBuilder } from '@develohpanda/fluent-builder';
+import { renderHook } from '@testing-library/react-hooks';
+import { ipcRenderer } from 'electron';
```
dimitropoulos: not saying to change this in this PR per se, but it still throws me off that some things are mocked implicitly like this (which, I do not so much enjoy) and some things are mocked explicitly (as below the imports).

develohpanda: It's not entirely an implicit mock (though I agree it is misleading), because we call `jest.mock('electron');` further down.

There is a case of mocks files instead of in-line mocks, and the biggest one is when the module is complex and/or is mocked in many places. Some examples from our codebase are electron, @grpc/grpc-js, node-forge and node-libcurl.

That said, there's no winning with jest mocks, they're confusing as heck regardless of where they're written 🙄

Hopefully fb83da462a8817fa2ad0c6e796a822b58e92ad43 makes it nicer to consume.

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
develohpanda: The action handlers don't differ between the different values, but I suppose that's a good test in itself.

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
develohpanda: Good idea

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
develohpanda: A linter should identify when something should be moved onto new lines and automatically fix. I agree, in a future PR we could (and should) use an option bag

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
develohpanda: 9e9515e25b450347c7c4db78ca9ea7c86281e708

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
develohpanda: 4decbb4b3767ff5c2ffe61bbf6bbe1f0914dc23d

packages/insomnia-app/app/network/grpc/__schemas__/grpc-ipc-request-params-schema.ts
```diff
@@ -1,8 +1,8 @@
-import type { Schema } from '@develohpanda/fluent-builder';
+import { createBuilder, Schema } from '@develohpanda/fluent-builder';
```
jackkav, line 1: this could be `@kong/fluent-builder` or pulled into the repo in order to avoid bus factor. It's also not clear what it's for, from the package name.

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import { mocked } from 'ts-jest/utils';
+
+import { GrpcRequestEventEnum } from '../../../../../common/ipc-events';
+import { grpcIpcMessageParamsSchema } from '../../../../../network/grpc/__schemas__/grpc-ipc-message-params-schema';
+import { grpcIpcRequestParamsSchema } from '../../../../../network/grpc/__schemas__/grpc-ipc-request-params-schema';
+import { GrpcMethodTypeEnum } from '../../../../../network/grpc/method';
+import { prepareGrpcMessage as _prepareGrpcMessage, prepareGrpcRequest as _prepareGrpcRequest } from '../../../../../network/grpc/prepare';
+import { GrpcDispatch } from '../../../../context/grpc';
+import { GrpcActionTypeEnum } from '../../../../context/grpc/grpc-actions';
+import useActionHandlers from '../use-action-handlers';
```
jackkav: having so many imports and mocks makes this test brittle and possibly an obstacle in the likely refactoring of ipc code in future.

packages/insomnia-app/app/network/grpc/__schemas__/grpc-ipc-request-params-schema.ts
develohpanda: `fluent-builder` is a package I've maintained separately for a long time so probably not something to pull in under the Kong org. Regarding name, `fluent-builder` is inspired by [fluent-assertions](https://fluentassertions.com/), and implements the [builder design pattern](https://sourcemaking.com/design_patterns/builder) for TS.

But also what's in a name: [`chokidar`](https://www.npmjs.com/package/chokidar) which is used by webpack/jest in watch mode is a name derived from the word Hindi word ["Chowkidar"](https://github.com/paulmillr/chokidar#also).

> Chowkidar is a transliteration of a Hindi word meaning 'watchman, gatekeeper', चौकीदार. This ultimately comes from Sanskrit _ चतुष्क_ (crossway, quadrangle, consisting-of-four).

packages/insomnia-app/app/ui/components/panes/grpc-request-pane/__tests__/use-action-handlers.test.ts
develohpanda: Because the feedback doesn't outline any action points, I'll treat it as a prompt for discussion and share my view points. There are few things to unpack from the comment, so I'll address each individually but briefly because I don't think they're totally connected. Apologies if it's a bit long, but I'd like to address each item.

#### Number of imports

We could do some refactoring, or use barrel exports to reduce the lines of imports, but the number of imports doesn't necessarily translate to complexity nor does it inherently contribute to brittleness

#### Number of mocks

This is a very valid point. It's possible to go _too far_ with mocks, leaving you in a situation where you are simply testing the mocks. In this particular example, I'm not sure where the concern of "so many mocks" comes from, given we only mock `electron` and `network/grpc/prepare` (the later is thoroughly [tested](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/network/grpc/__tests__/prepare.test.ts) and requires a bunch of setup).

But the depth of mocking depends on the end goal of the test being written and the nature of the test - unit, integration, e2e. In lieu of formal dependency injection, I've used mocks here to significantly reduce the test surface area to be only the contents of `use-action-handlers`; to test one _unit_ of work.

More specifically, these tests are making sure that for each action handler, we send the expected _ipc event_ and the expected _args_. The implementation within the handlers can change, but the event and args shouldn't change.

> If we don't use the mocks we have, we'll have to use deeper mocks (see the example I linked above), and add setup content into these tests for areas of code that aren't being targeted by these tests, which will ultimately lead to _more_ complexity in the given suite.

> Notice that these tests are decoupled from the database, whereas many tests require writing to/reading from the database just for setting up the necessary data.

#### Possible obstacle in future refactoring

I respectfully disagree with this having any influence. Unit tests (or tests in general) can either be viewed as obstacles or safety nets. If something needs refactoring in the future, then the tests either become the way to ensure we haven't broken something in the refactor, or we update the tests to match the new code and review the changes in spec. Potential future work in the area is rarely a reason to _not_ write the necessary tests now, even if they are complex because there are mocks/setup involved with testing one specific aspect.

In addition, gRPC was largely written over this IPC bridge, I am not anticipating any significant refactors to this area.

---

Would you have any suggestions for how these tests should change?

jackkav: Thanks for sharing your viewpoint, this is a great chance to discuss.

> I am not anticipating any significant refactors to this area.

That is what inspired me to voice concerns. I found that the renderer code for grpc uses the nodejs module rather than the web module so some refactoring will need to happen for contextIsolation. 

My previous comment didn't give provide enough context and so weren't fair, thats my bad. So I'll just share some ideas and you can do what you like with them.

My viewpoint on unit testing is:
- Multiline Arrange or test setup and mocks more complex than a stub are a smell that our domain logic is coupled with a library.
- Testing library interactions is always useful for learning about a library but has limited return on investment because they tend to be hard for others to read without others going through the same learning experience it took to write them.

This viewpoint is often hard to apply but it has paid off for me. I don't expect you to agree with me, or do anything differently in this PR just wanted to take this opportunity to share my viewpoint and perhaps engage further discussion in a higher bandwidth medium.

TLDR: not a blocker.

packages/insomnia-app/app/network/grpc/__schemas__/grpc-ipc-request-params-schema.ts
jackkav: - builder pattern is used for solving object creation problems in OO, I don't see why this couldn't be achieved in a more idiomatic way in pure javascript, or better yet refactor the code under test to eliminate the need to solve object creation problems.
- naming is hard.
- IMHO single maintainer dependencies are short sighted no matter who maintains them.

I found this recent article insightful in relation to my single maintainer opinion. https://blog.yossarian.net/2021/12/12/Calculating-my-open-source-blast-radius

develohpanda:
Closing in favor of gRPC e2e tests (INS-1341) which I know @filfreire has been thinking about.

PR #4307, opened 2021-12-14
removes onboarding activity and associated handling code

dimitropoulos:
closes INS-129
changelog(Improvements): First launch experience is now simpler

This PR has a very restricted scope to removing the `WrapperOnboarding` from the first launch flow as part of a project to simplify the first launch experience.

b2377a3b-e3c9-46b8-be7b-afe3fb459812


packages/insomnia-app/app/ui/components/wrapper.tsx
```diff
@@ -728,10 +726,8 @@ export class Wrapper extends PureComponent<WrapperProps, State> {
 
           {activity === ACTIVITY_MIGRATION && <WrapperMigration wrapperProps={this.props} />}
 
-          {activity === ACTIVITY_ANALYTICS && <WrapperAnalytics wrapperProps={this.props} />}
-
-          {(activity === ACTIVITY_ONBOARDING || activity === null) && (
-            <WrapperOnboarding wrapperProps={this.props} />
+          {(activity === ACTIVITY_ANALYTICS || activity === null) && (
```
dimitropoulos, line 729: I left this here in the name of not breaking anything, but I would like to have more context as to the situations in which the activity is `null`, if anyone knows.

packages/insomnia-app/app/ui/redux/modules/__tests__/global.test.ts
```diff
@@ -151,9 +125,9 @@ describe('global', () => {
   });
 
   describe('goToNextActivity', () => {
-    it('should go from onboarding to home', async () => {
-      const settings = createSettings(false, false, true);
-      const activeActivity = ACTIVITY_ONBOARDING;
+    it('should go from analytics to home', async () => {
```
dimitropoulos, line 128: I didn't do it because I want to keep scope down, but I think we should remove these tests that use `goToNextActivity` in the future and actually test the component itself (in this case, `WrapperAnalytics`).  The reason is because the component itself _does not use_ the `goToNextActivity` utility _anyway_.  So here in these tests we're testing something that the app doesn't even do.

Thoughts?

jackkav (approved):

packages/insomnia-app/app/ui/components/wrapper-onboarding.tsx
```diff
@@ -1,189 +0,0 @@
-import 'swagger-ui-react/swagger-ui.css';
```
develohpanda, line 1: So glad this page is gone

packages/insomnia-app/app/ui/components/wrapper.tsx
develohpanda: Before onboarding/analytics were activities in their own right, it was "null" which specified onboarding/analytics, so it was kept as a fallback. It should be in the history if you follow it far enough

packages/insomnia-app/app/ui/redux/modules/__tests__/global.test.ts
develohpanda: `goToNextActivity` is basically a finite state machine, moving from one to the next in a logical manner and doesn't know where it's being used.

I _think_ it's still relevant. My suggestion would be to either a) consume this function where it should be but isn't right now, or b) keep the function and take away the code paths (and tests) that are no longer hit. Instead of merging the state machine with the component itself. I imagine this state machine concept will soon become redundant though (yay)

If I'm missing something let's chat about this before making a larger refactor in the future PR

develohpanda (commented):
Initial code review, still need to QA

filfreire (approved):
@dimitropoulos marking these as approved just to let you know I did run some tests, and the change seems to be ok.
Here's some of the things I tried:
- Spinning up the app with previous insomnia-data - to check if we would randomly get the analytics activity having completed it a long time ago
- Clean up the insomnia-data on two different occasions, skipping and accepting analytics, and reopened the app just to check if we would still see that activity after accepting/skipping, we didn't ✅ 
- Tried to spin up the app from a few weird states:
  - If you close it when the analytics stuff is still not resolved
  - If you had closed it on another version for some reason on the onboarding activity. It'll show you the app and detect that you had skipped analytics. I was wondering if it would be left in a weird state, but I was able to use the app fine.
  
 I suspect we had a "bug" before though - if folks would accept/skip analytics, but quit the app with the onboarding activity on - we would show them the analytics activity again instead of picking back up from the onboarding one. But I guess it wouldn't matter now after this change and I would admit it's not a show-stopper.
 
 PS.: I suspect it might be worth testing this out on different OSs after we merge and bake a proper build, specially on Windows since it tends to have a bit more surprises when we boot up the app 🤷

develohpanda (approved):
Approving on the grounds of @filfreire QA above, thanks for running through the thorough QA!

PR #4310, opened 2021-12-15
removes analytics activity and create a modal

dimitropoulos:
closes INS-1286

This PR makes the analytics activity a modal.


develohpanda (approved):
🚢 

(I didn't QA this as part of my review because of time constraints, but we pair programmed and we QA'd during that, so I am satisfied with this!)

PR #4314, opened 2021-12-16
create background resources on first launch

dimitropoulos:
closes INS-1304

As part of our updates to the onboarding experience, when the app launches for the first time, we need to create (and activate) a new request in a new design document.


packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1416,9 +1419,27 @@ class App extends PureComponent<AppProps, State> {
...
   }
 
+  async _handleFirstLaunch() {
+    const { handleActivateWorkspace, stats } = this.props;
+
+    if (stats.launches > 1) {
+      return;
+    }
+
+    const { _id: workspaceId } = await models.workspace.create({ scope: 'design', parentId: DEFAULT_PROJECT_ID });
```
gatzjames: There's quite some async work happening in this function.
What would happen if any of these fails/throws for some reason? Will it be handled by an Error Boundary?

packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -1416,9 +1419,27 @@ class App extends PureComponent<AppProps, State> {
...
+    const request = await models.request.create({ parentId: workspaceId });
+    await models.workspaceMeta.getOrCreateByParentId(workspaceId);
+    await models.workspaceMeta.updateByParentId(workspaceId, { activeRequestId: request._id, activeActivity: ACTIVITY_DEBUG });
+
+    handleActivateWorkspace({ workspaceId });
```
gatzjames: 🤔 Is it important that `handleActivateWorkspace` is called before the settings update?
Thinking if it would be helpful to move these database operations into a function and return a cleanup function that cleans the db if any error happens?

packages/insomnia-smoke-test/modules/onboarding.ts
```diff
@@ -1,17 +1,22 @@
...
+
+const analyticsMessageShown = async (app: Application) => {
   await app.client.waitUntilTextExists(
     'p strong',
     'Share Usage Analytics with Kong Inc',
   );
 };
 
-const clickDontShare = async app => {
+const clickDontShare = async (app: Application) => {
```
gatzjames: 🙏

packages/insomnia-smoke-test/core/import.test.ts
```diff
@@ -10,7 +10,7 @@ import * as modal from '../modules/modal';
 import * as onboarding from '../modules/onboarding';
 import * as settings from '../modules/settings';
 
-describe.only('Import', function() {
+describe('Import', function() {
```
gatzjames: 👏

packages/insomnia-app/app/main.development.ts
```diff
@@ -70,13 +70,8 @@ app.on('ready', async () => {
   // Init some important things first
   await database.init(models.types());
   await _createModelInstances();
-  await errorHandling.init();
-  await windowUtils.init();
-  // Init the app
-  const { launches } = await _trackStats();
-  if (launches === 1) {
-    await handleFirstLaunch();
```
gatzjames, line 78: 🤔 Would this affect any other behaviour of the app since we now run this on render?
As far as I understand the app would not launch before doing the check and updating the settings.
With the current changes the app will launch and update the settings later on.
Is there any other part of the app that relies on this setting?

gatzjames (commented):
Looks great!
Added some comments about error handling and database atomicity let me know what you think!

packages/insomnia-app/app/main.development.ts
dimitropoulos: I started off doing this here, but in the end, it turns out there are too many things (e.g. side effects in localstorage, another is dispatching redux actions) that we can't do here.  `localStorage` is not available from this code.  not on `window.localStorage`, `global.localStorage`, `global.window.localStorage`, etc.  That causes a change to do it in App, where, by the way, it used to be done before we removed the creation of a starter workspace a while back.

As far as I am aware, there is no difference in the two.

packages/insomnia-smoke-test/core/import.test.ts
dimitropoulos: :)

johnwchadwick (approved):
Seems to work for me.

PR #4319, opened 2021-12-16
Cleanup extra 'v' in version of the document card

wdawson:
changelog(Fixes): An extra `v` will no longer appear in the dashboard for a Design Document's version if the version already starts with `v`

Took the quick fix path for this issue for now. As noted in the issue, we want to improve this eventually to more clearly show that it's metadata and pull the value directly. But this is the solution we want for now.

https://www.loom.com/share/5971e8b2100c4a96a114fc0ecc04f7b9?sharedAppSource=team_library

Resolves INS-1277


wdawson:
I couldn't easily find anywhere to add a test. Is there anything I can add to?

packages/insomnia-app/app/ui/components/workspace-card.tsx
```diff
@@ -85,13 +85,17 @@ export const WorkspaceCard: FC<WorkspaceCardProps> = ({
...
 
-  const version = spec?.info?.version || '';
+  let version = spec?.info?.version || '';
   let label: string = strings.collection.singular;
   let format = '';
   let labelIcon = <i className="fa fa-bars" />;
   let defaultActivity = ACTIVITY_DEBUG;
   let title = workspace.name;
 
+  if (version.length > 0 && !version.startsWith('v')) {
```
filfreire: Tested locally, this seems ok for strings like `v1.0` or `1.0`.

This would still have a problem with version names where folks decide to start with other letters, e.g. `alpha-1.0` or `something`. We could probably also check if it starts with any non digit character (a-zA-Z).

Screenshot 2021-12-17 at 13 22 16

filfreire (changes requested):

packages/insomnia-app/app/ui/components/workspace-card.tsx
wdawson: As per the approach in the ticket, we don't want to do that for now. Without additional context for what the string is, we want to keep some indication that it's a version. Perhaps we can add a space or something to make it clearer and look less like a mistake. 

We'll still have issues, per the current implementation, if the string starts with a `v` though.

wdawson: I may be the only one concerned about the lack of context, however. In which case, we could drop the `v` entirely.

dimitropoulos: <discussed in standup that since we all agree that what's in this PR is at least _better_ than what we have right now, we're going to merge as is and keep an eye open for the future on making this even better>

dimitropoulos (approved):

wdawson:
@filfreire , seems like the team is aligned that this is better than what we have and we will follow up later with additional improvements. Does that sound good to you?

dimitropoulos:
<discussed and the team agreed in standup that since we all agree that what's in this PR is at least better than what we have right now, we're going to merge as is and keep an eye open for the future on making this even better>

wdawson:
Thanks for the tests @dimitropoulos !

PR #4335, opened 2022-01-03
Vulnerability fix: CVE-2021-3777

iPedreroIBM:
Scanned using WhiteSource.:

nodejs-tmpl is vulnerable to Inefficient Regular Expression Complexity

Updated from 1.0.4 to 1.0.5.

URL: https://vuln.whitesourcesoftware.com/vulnerability/CVE-2021-3777

(Also removed local GitHub action that automatically updates the fork and whitesource config)


gatzjames (approved):

dimitropoulos (approved):

dimitropoulos (changes requested):
I'd love to get this merged soon, but I've just noticed there's a problem: the integrity fields were not updated.  That means that if we ran `npm install` after this PR that it will simply bump it right back to `1.0.4`.

I don't actually know what the right way is to do this, but here's an example of dependabot doing it: https://github.com/Kong/insomnia/pull/4359/files#diff-d75077b306a489aef01200210f58d2b45ea25bb4f106ec120db4446fa7125dc1L25020-R26095

johnwchadwick:
Closing in favor of #4716 which has the integrity fields updated properly.

PR #4338, opened 2022-01-04
AutoFocus URL bar when onboarding/switching/opening a request

dimitropoulos:
https://user-images.githubusercontent.com/15232461/148108232-2ba76bf0-e99f-4bac-acae-7353839e9af0.mp4

There are three behaviors this PR changes:

1. when onboarding, after clicking through the analytics modal the url bar will focus
2. after creating a new request, the URL bar will focus
3. after switching to a new request, the URL bar will focus (note, that this (logically speaking) also means that it will focus when switching to a new request when there wasn't one previously, such as from the dashboard).

closes INS-1308
changelog(Improvements): URL bar will now autofocus when interacting with new requests or switching between requests

### DIFF
--- a/packages/insomnia-app/app/ui/components/panes/request-pane.tsx
+++ b/packages/insomnia-app/app/ui/components/panes/request-pane.tsx
@@ -1,7 +1,8 @@
 import classnames from 'classnames';
 import { deconstructQueryStringToParams, extractQueryStringFromUrl } from 'insomnia-url';
-import React, { FC, useCallback } from 'react';
+import React, { FC, useCallback, useEffect, useRef } from 'react';
 import { Tab, TabList, TabPanel, Tabs } from 'react-tabs';
+import { useMount } from 'react-use';
 
 import { getAuthTypeName, getContentTypeName } from '../../../common/constants';
 import * as models from '../../../models';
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/ui/components/panes/request-pane.tsx
```diff
@@ -118,255 +135,239 @@ export class RequestPane extends PureComponent<Props> {
...
-    const hotKeyRegistry = settings.hotKeyRegistry;
+  const requestUrlBarRef = useRef<RequestUrlBar | null>(null);
+  useMount(() => {
+    requestUrlBarRef.current?.focusInput();
+  });
+  useEffect(() => {
+    requestUrlBarRef.current?.focusInput();
+  }, [
+    request?._id, // happens when the user switches requests
+    settings.hasPromptedAnalytics, // happens when the user dismisses the analytics modal
```
gatzjames, line 148: This isn't scalable since for every feature/modal etc that affects this the RequestPane should know about it and re-focus itself.
Since we want to revisit modals and focus in the app in general I'd say it's okay for now but the next time we need to explicitly handle focus like that we should invest in a more scalable solution in the app.

gatzjames (approved):
LGTM! Great improvement!

packages/insomnia-app/app/ui/components/panes/request-pane.tsx
dimitropoulos: I was thinking exactly the same thing, yeah.  it sorta is what it is at this point so I at least left a comment to make it less mysterious.

PR #4339, opened 2022-01-04
removes (now-unnecessary) OnboardingContainer

dimitropoulos:
closes INS-1307

now that the onboarding container is only used by migration, we can move it here, which will make the upcoming deletion of the migration code that much easier.

there should be no logical changes to the function of the app in this PR


gatzjames (approved):
👍

PR #4340, opened 2022-01-04
[DOCU-2110] update Preferences tooltips and title casing

falondarville:
Update Preferences tooltips and title casing. Uses active voice, sentence casing for subtitles, and ends sentences with periods.

### Reason

Clarifies functionality and makes tooltips easier to read and comprehend.

Closes INS-1273
Closes [DOCU-2110](https://konghq.atlassian.net/browse/DOCU-2110)

### DIFF
--- a/packages/insomnia-app/app/ui/components/settings/general.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/general.tsx
@@ -199,7 +199,7 @@ export const General: FC<Props> = ({ hideModal }) => {
             placeholder="-- System Default --"
           />
           <NumberSetting
-            label="Interface font size (PX)"
+            label="Interface font size (px)"
             setting="fontSize"
             min={MIN_INTERFACE_FONT_SIZE}
             max={MAX_INTERFACE_FONT_SIZE}
@@ -323,7 +323,7 @@ export const General: FC<Props> = ({ hideModal }) => {
           min={-1}
         />
         <NumberSetting
-          label="Max timeline chunk size (KB)"
+          label="Max timeline chunk size (KiB)"
           setting="maxTimelineDataSizeKB"
           help="Enter the maximum size in kilobytes to show on the response timeline. Decrease the number for less detailed responses."
           min={0}
--- a/packages/insomnia-app/app/ui/components/settings/general.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/general.tsx
@@ -325,7 +325,7 @@ export const General: FC<Props> = ({ hideModal }) => {
         <NumberSetting
           label="Max timeline chunk size (KiB)"
           setting="maxTimelineDataSizeKB"
-          help="Enter the maximum size in kilobytes to show on the response timeline. Decrease the number for less detailed responses."
+          help="Enter the maximum size in kibibytes to show on the response timeline. Decrease the number for less detailed responses."
           min={0}
         />
       </div>
--- a/packages/insomnia-app/app/ui/components/settings/general.tsx
+++ b/packages/insomnia-app/app/ui/components/settings/general.tsx
@@ -118,7 +118,7 @@ export const General: FC<Props> = ({ hideModal }) => {
           <BooleanSetting
             label="Use vertical layout"
             setting="forceVerticalLayout"
-            help="Stack application panels (e.g. request / response) vertically instead of horizontally."
+            help="If checked, stack request and response panels vertically."
           />
           <BooleanSetting
             label={<RestartTooltip message="Show variable source and value" />}
@@ -146,8 +146,8 @@ export const General: FC<Props> = ({ hideModal }) => {
 
       <div className="row-fill row-fill--top pad-top-sm">
         <EnumSetting<EnvironmentHighlightColorStyle>
-          label="Environment Highlight Style"
-          help="Configures the appearance of environment's color indicator"
+          label="Environment highlight style"
+          help="Select the sub-environment highlight area. Configure the highlight color itself in your environment settings."
           setting="environmentHighlightColorStyle"
           values={[
             { value:'sidebar-indicator', name: 'Sidebar indicator' },
@@ -162,7 +162,7 @@ export const General: FC<Props> = ({ hideModal }) => {
         <NumberSetting
           label="Autocomplete popup delay"
           setting="autocompleteDelay"
-          help="Configure the autocomplete popup delay in milliseconds (0 to disable)"
+          help="Delay the autocomplete popup by milliseconds. Enter 0 to disable the autocomplete delay."
           min={0}
           max={3000}
         />
@@ -193,13 +193,13 @@ export const General: FC<Props> = ({ hideModal }) => {
       <div className="form-row pad-top-sm">
         <div className="form-row">
           <TextSetting
-            label="Interface Font"
+            label="Interface font"
             setting="fontInterface"
-            help="Comma-separated list of fonts. If left empty, takes system defaults."
+            help="Enter a comma-separated list of fonts. If left empty, uses system defaults."
             placeholder="-- System Default --"
           />
           <NumberSetting
-            label="Interface Font Size (px)"
+            label="Interface font size (PX)"
             setting="fontSize"
             min={MIN_INTERFACE_FONT_SIZE}
             max={MAX_INTERFACE_FONT_SIZE}
@@ -209,9 +209,9 @@ export const General: FC<Props> = ({ hideModal }) => {
 
       <div className="form-row">
         <TextSetting
-          label="Text Editor Font"
+          label="Text editor font"
           setting="fontMonospace"
-          help="Comma-separated list of monospace fonts. If left empty, takes system defaults."
+          help="Enter a comma-separated list of monospace fonts. If left empty, uses system defaults."
           placeholder="-- System Default --"
         />
         <NumberSetting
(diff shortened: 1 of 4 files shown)


packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
```diff
@@ -312,7 +312,7 @@ export const ThemePanel: FC = () => {
         <label className="inline-block">
           Use OS color scheme
           <HelpTooltip className="space-left">
-            Pick your prefered themes for light and dark
+            Select different themes for when you’re using light versus dark color schemes on your OS. Check this box, then hover over a theme and select either sun (light) or moon (dark).
```
filfreire, line 315: I don't think it's a critical problem but the right-leaning single quote `’` (instead of the simpler single quote `'`) will be picked up by [vscode-gremlins](https://github.com/nhoizey/vscode-gremlins/) and similar tools. 

I wouldn't bother handling it in this PR though - as we have a ton more similar cases in the codebase of right-leaning single quote `’` and other kind of characters.

filfreire (approved):

packages/insomnia-app/app/ui/components/settings/theme-panel.tsx
falondarville: Oh good to know! Thank you!

packages/insomnia-app/app/ui/components/settings/import-export.tsx
```diff
@@ -72,10 +71,7 @@ export const ImportExport: FC<Props> = ({ hideSettingsModal }) => {
   return (
     <div data-testid="import-export-tab">
       <div className="no-margin-top">
-        Import format will be automatically detected.
-        <HelpTooltip className="space-left">
-          Supported formats include: {importers.map(importer => importer.name).join(', ')}
-        </HelpTooltip>
+        Import format will be automatically detected. Supported formats include: {importers.map(importer => importer.name).join(', ')}
```
dimitropoulos, line 74: this was done by product decision of @wdawson.  please check in with him.  It used to list them and then we changed it.

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -251,36 +250,36 @@ class General extends PureComponent<Props> {
...
-            label="Response History Limit"
+            label="Response history limit"
             setting="maxHistoryResponses"
-            help="Number of responses to keep for each request (-1 for infinity)"
+            help="Enter the number of responses to keep for each request. Enter -1 to keep all response history."
             min={-1}
           />
           <NumberSetting
-            label="Max Timeline Chunk Size (KB)"
+            label="Max timeline chunk size (KB)"
```
dimitropoulos: @wdawson do you want `(KB)` and `(PX)` (above) or `(kb)` and `(px)`?

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -308,48 +307,37 @@ class General extends PureComponent<Props> {
...
-            help="Indicates whether SSL certificates should be validated during authentication flows"
+            help="If checked, validates SSL certificates during authentication flows."
           />
         </div>
 
         <hr className="pad-top" />
 
-        <h2>
-          HTTP Network Proxy
-          <HelpTooltip
```
dimitropoulos: great change!

packages/insomnia-app/app/ui/components/settings/import-export.tsx
wdawson: @dimitropoulos I'd ideally like to retain the supported versions in the actual editor too, but only if a user enters an invalid version or none at all. As it is today, they see the error before they have a chance to do anything about it. But from what I understood, those errors are out of our control in swagger UI, right?

Still good to have this help text I think here, though 👍

packages/insomnia-app/app/ui/components/settings/general.tsx
wdawson: pixels are commonly abbreviated with the lowercase `px` so we should probably change to that.

For the size, it depends on what we mean (@dimitropoulos , do you know?):
- If it's [kilobytes](https://en.wikipedia.org/wiki/Kilobyte) (1000 bytes or 8000 bits), it's `kB`.
- If it's [kilobits](https://en.wikipedia.org/wiki/Kilobit) (125 bytes or 1000 bits), it's `kbit`.
- If it's [kibibytes](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units) (1024 bytes or 8192 bits), it's `KiB`.
- If it's [kibibits](https://en.wikipedia.org/wiki/Kibibit) (128 bytes or 1024 bits), it's `Kibit`.

packages/insomnia-app/app/ui/components/settings/import-export.tsx
dimitropoulos: The difference is that it will be spelled out on the screen for the user rather than hidden behind a tooltip now.  Sorry, I'm on my way out for the day or I'd send a before/after screenshot.

wdawson: Oh yep. I'm good with it as long as it looks good on the screen.

dimitropoulos: | BEFORE | AFTER |
| - | - |
| Screenshot_20220111_132626 | Screenshot_20220111_132446 |

packages/insomnia-app/app/ui/components/settings/general.tsx
dimitropoulos: turns out it's kibibytes! https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/network/network.ts#L322

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -193,13 +193,13 @@ export const General: FC<Props> = ({ hideModal }) => {
...
-            label="Interface Font"
+            label="Interface font"
             setting="fontInterface"
-            help="Comma-separated list of fonts. If left empty, takes system defaults."
+            help="Enter a comma-separated list of fonts. If left empty, uses system defaults."
             placeholder="-- System Default --"
           />
           <NumberSetting
-            label="Interface Font Size (px)"
+            label="Interface font size (PX)"
```
falondarville:
```suggestion
            label="Interface font size (px)"
```

packages/insomnia-app/app/ui/components/settings/import-export.tsx
wdawson: OHHH... I remember this now. Hmm. I get now how it doesn't really make sense in a tooltip (especially given the following line). But as a single line it's really hard to read (long list of things). I think listing in bullet form might be too long also. I wonder if there's a way to abbreviate? Example of what I mean:

> The format for import will be automatically detected. Supported formats include: Insomnia exports, some Postman exports, OpenAPI 3.0, and [more](https://docs.insomnia.rest/insomnia/import-export-data). 

Though I think we need to update that documentation if we go that route.

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -296,36 +296,36 @@ export const General: FC<Props> = ({ hideModal }) => {
...
-          label="Response History Limit"
+          label="Response history limit"
           setting="maxHistoryResponses"
-          help="Number of responses to keep for each request (-1 for infinity)"
+          help="Enter the number of responses to keep for each request. Enter -1 to keep all response history."
           min={-1}
         />
         <NumberSetting
-          label="Max Timeline Chunk Size (KB)"
+          label="Max timeline chunk size (KB)"
```
falondarville:
```suggestion
          label="Max timeline chunk size (KiB)"
```

packages/insomnia-app/app/ui/components/settings/import-export.tsx
falondarville: Yes, can we make this a separate task and PR to update the content that's being populated (as this is not hard-coded here).  @wdawson

falondarville: FYI created a Linear task to update the importers list https://linear.app/insomnia/issue/INS-1346/update-generated-importer-list

packages/insomnia-app/app/ui/components/settings/general.tsx
```diff
@@ -296,36 +296,36 @@ export const General: FC<Props> = ({ hideModal }) => {
...
-          help="Number of responses to keep for each request (-1 for infinity)"
+          help="Enter the number of responses to keep for each request. Enter -1 to keep all response history."
           min={-1}
         />
         <NumberSetting
-          label="Max Timeline Chunk Size (KB)"
+          label="Max timeline chunk size (KiB)"
           setting="maxTimelineDataSizeKB"
-          help="Maximum size in kilobytes to show on timeline"
+          help="Enter the maximum size in kilobytes to show on the response timeline. Decrease the number for less detailed responses."
```
dimitropoulos:
```suggestion
          help="Enter the maximum size in kibibytes to show on the response timeline. Decrease the number for less detailed responses."
```

dimitropoulos (approved):

PR #4342, opened 2022-01-05
EmptyState replacement for body editor

dimitropoulos:
This PR implements the new behavior for the request pane help text.

non gRPC requests:
| BEFORE | AFTER |
| - | - |
| Screenshot_20220105_162556 | Screenshot_20220105_162304 |

gRPC requests:
| BEFORE | AFTER |
| - | - |
| Screenshot_20220105_162543 | Screenshot_20220105_162258 |

closes: INS-1284


packages/insomnia-app/app/ui/components/panes/zero-state-pane.tsx
```diff
@@ -0,0 +1,100 @@
...
+}) => {
+  return (
+    <Panel>
+      <Wrapper>
+        <Icon>{icon}</Icon>
+        <Title>{title}</Title>
+        <Divider />
+        <SecondaryAction>{secondaryAction}</SecondaryAction>
+        <DocumentationLinks>
+          {map(({ title, url }) => (
```
gatzjames: Could we replace this with the native array.map? Seems like we don't need the extra abstraction cost.

packages/insomnia-app/app/ui/css/css-in-js.ts
```diff
@@ -0,0 +1,15 @@
+export const faint = {
```
gatzjames, line 1: Since these styles are only used by the empty state pane they could live in that file.
I'd also try to improve the names to be a bit more intuitive 😅

packages/insomnia-app/app/ui/components/panes/zero-state-pane.tsx
```diff
@@ -0,0 +1,100 @@
...
+  '& hover': {
+    textDecoration: 'none',
+  },
+});
+
+const LinkIcon = styled(SvgIcon)({
+  paddingLeft: 'var(--padding-sm)',
+});
+
+export const ZeroStatePane: FC<{
```
gatzjames: Not sure what zero state means? Could we rename this to EmptyStatePane instead? Afaik it's widely used naming for when there's nothing to display on the screen.

gatzjames (changes requested):
Nice one! Left some comments for naming and I believe the styles could stay on the component file!

packages/insomnia-app/app/ui/css/css-in-js.ts
dimitropoulos: sorry, for the lack of context.  they are direct copies of the utility files we use throughout the project https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/css/layout/base.less#L356

That's also why I put them here.  We've done a bit of copying this stuff around in the past (or, just not using the utilities) and at this point I think I want to stop copying around as much.

This is absolutely _not_ "CSS-in-JS done right", in my opinion, but it's the least friction necessary to get us off of the .less stuff and keep things familiar until we have an actual theme system like a mature ui library might.

dimitropoulos: Notice, too, that the prior implementation uses the `faint` class.  That's really where this ultimately comes from.

packages/insomnia-app/app/ui/components/panes/zero-state-pane.tsx
dimitropoulos: absolutely :)

apologies for my muscle memory -> a8f7ef591d7819f3fe02bf395d0de28f7127b37f

packages/insomnia-app/app/ui/components/panes/zero-state-pane.tsx
dimitropoulos: I originally called it that, but I changed it to ZeroState because that's the name of the feature and how we're referring to it.  let's chat about this.  I'm 100% cool with whatever.

dimitropoulos: upon chatting we agree that "empty state" is much more common for developers, so, even if product people / designers / whoever are using different terms, we don't really need to diverge from the most common terminology on the code.

15e528fa69dc854314f2a4b868606426ce10c0cf

packages/insomnia-app/app/ui/components/panes/empty-state-pane.tsx
```diff
@@ -0,0 +1,99 @@
...
+});
+
+const Wrapper = styled.div({
+  display: 'flex',
+  flexDirection: 'column',
+  alignItems: 'center',
+  justifyContent: 'center',
+  padding: 'var(--padding-md)',
+  textAlign: 'center',
+  ...superFaint,
```
gatzjames, line 22: I think we should in-line these since they aren't used anywhere else and the naming is not clear. Not sure what the benefit of a separate file is 🤔

gatzjames: This file is used in order to keep parity with the css classnames for now.
Something to look into when updating the styles/theming of the app since the current solution doesn't scale.

gatzjames (approved):

PR #4345, opened 2022-01-07
Empty State for Design Tab

dimitropoulos:
In this PR I implemented the empty state for the Design Tab

| BEFORE | AFTER |
| - | - |
| Screenshot_20220110_152245 | Screenshot_20220110_151918 |

This video shows it in action and explains some of the focus behavior:

https://user-images.githubusercontent.com/15232461/148833393-76853cf4-bd5f-4def-b11c-30374f95dac3.mp4

I separated out a commit [`c2ed656` (#4345)](https://github.com/Kong/insomnia/pull/4345/commits/c2ed6568b3abfad89bf6f0cb5840b262971e16cf) for the solution for the cache invalidation.  Without this commit, if you take one of the actions (import from file, import from url, start from example) the rest of the UI will update, but not the code editor itself.  I'm not very happy with this solution, but upon quite lengthy conversation with the rest of the team about alternatives I'm convinced it's the best available option.

closes: INS-1288


packages/insomnia-app/app/ui/components/design-empty-state.tsx
```diff
@@ -0,0 +1,160 @@
...
+      onComplete: async (uri: string) => {
+        const response = await window.fetch(uri);
+        if (!response) {
+          return;
+        }
+        const contents = await response.text();
+        await updateApiSpecContents(contents);
+      },
+    });
+    onUpdateContents();
```
develohpanda: Should this be here, or inside the `onComplete` handler?

packages/insomnia-app/app/ui/components/design-empty-state.tsx
```diff
@@ -0,0 +1,160 @@
...
+    return null;
+  }
+
+  return (
+    <Wrapper>
+      <EmptyStatePane
+        icon={<SvgIcon icon="drafting-compass" />}
+        documentationLinks={[
+          { title: 'Working with Design Documents', url: 'https://docs.insomnia.rest/insomnia/get-started' },
+          { title: 'Introduction to Insomnia', url: 'https://docs.insomnia.rest/insomnia/get-started' },
```
develohpanda: Suggestion to put the docs in the existing `common/documentation.ts` file, similar to how we have a strings file.

Speaking of, the "Design Documents" should be using the strings object unless you're choosing to deviate intentionally. There are  a few other places I have noticed this in this PR as well.

develohpanda (commented):
Using the `forceRefreshKey` is almost [unintentionally designed](https://github.com/Kong/insomnia/blob/dc56a993f9d75c3091cd12f81e20b6d3597b0a79/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx#L250-L251) behavior. This is because the CodeEditor is _not_ a controlled component as far as I can tell, so the `defaultValue` is used on mount, but the value is persisted by CodeMirror/our CodeEditor itself so changing the `defaultValue` prop doesn't do anything unless we force an update using the uniquenessKey.

This is kind of scary; there is a chance for the db and the component to fall out of sync momentarily, if file writing (for example) is slow when writing into NeDB. There's a whole host of ways this issue could surface though, not just file I/O.

This is a problem similar in principle (prop updates not propagating through the rendered instance) to the issue I'm resolving in https://github.com/Kong/insomnia/pull/4276.

packages/insomnia-app/app/ui/components/design-empty-state.tsx
dimitropoulos: GREAT catch.  it should definitely be in the onComplete handler.

91d60036311d249774acf260a4fdb75ec019f493

packages/insomnia-app/app/ui/components/design-empty-state.tsx
dimitropoulos: great point, I fixed the documentation.ts part here 37fc74f7bc6b87046ce9df8dff43301806e32c05

dimitropoulos: Regarding strings, I think maybe in this case since these strings are not actually "from" our source code, but rather copies of the names of the things on the docs (that we don't have a way to link), we want to keep them in sync with the docs.  Even though, yes, they happen to be the same as our own constants in the app _right now_, a change in the app doesn't necessitate a change in the docs title, you know what I mean?

cc: @falondarville just to be aware of this if we're ever adjusting the names.  They are (after the above commit) in just one place in `documentation.ts` now.

Screenshot_20220111_084627

packages/insomnia-app/app/ui/components/wrapper-design.tsx
```diff
@@ -180,6 +204,14 @@ export class WrapperDesign extends PureComponent<Props, State> {
       return null;
     }
 
+    if (!activeApiSpec.contents) {
+      return (
+        <EmptySpaceHelper>
+          Documentation for your OpenAPI spec will render here
```
develohpanda, line 210: The word `render` here feels like coding/front-end jargon. For an everyday user, something like `appear` or `show` seems friendlier?

packages/insomnia-app/app/ui/components/design-empty-state.tsx
```diff
@@ -0,0 +1,161 @@
...
+  return (
+    <Wrapper>
+      <EmptyStatePane
+        icon={<SvgIcon icon="drafting-compass" />}
+        documentationLinks={[
+          documentationLinks.workingWithDesignDocs,
+          documentationLinks.introductionToInsomnia,
+        ]}
+        secondaryAction={<SecondaryAction onUpdateContents={onUpdateContents} />}
+        title="Enter an OpenAPI specification here"
```
develohpanda, line 157: Instead of "enter", should this be "type or paste", or just "type"?

Also, "here" in the sentence "Enter an openapi spec here" doesn't really refer to anything, because when you click anywhere near the word "here", the cursor shows far away and there is a lot of whitespace in between. 

@erickrico / @wdawson / @falondarville your input here will be valuable, please!

Screenshot for reference
image

packages/insomnia-app/app/ui/components/design-empty-state.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+import { documentationLinks } from '../../common/documentation';
+import { selectFileOrFolder } from '../../common/select-file-or-folder';
+import * as models from '../../models';
+import { faint } from '../css/css-in-js';
+import { selectActiveApiSpec } from '../redux/selectors';
+import { showPrompt } from './modals';
+import { EmptyStatePane } from './panes/empty-state-pane';
+
+const Wrapper = styled.div({
```
develohpanda, line 15: I've seen a `Wrapper` styled component a few times and each time I think it's our big `Wrapper` component. May I suggest we use a `Container` for components like these, for future?

develohpanda:
The above can be addressed in this PR or a future PR. Disabling auto-merge so that I can approve but so that the comments are still looked at before merging, and others have a chance to review.

develohpanda (approved):

packages/insomnia-app/app/ui/components/design-empty-state.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const contents = String(await fs.promises.readFile(filePath));
+    await updateApiSpecContents(contents);
+  }, [updateApiSpecContents]);
+
+  const handleImportUri = useCallback(async () => {
+    showPrompt({
+      title: 'Import document from URL',
+      submitName: 'Fetch and Import',
+      label: 'URL',
+      placeholder: 'https://gist.githubusercontent.com/<user>/<id>/raw/<id>/openapi.yaml',
```
develohpanda: Having a valid placeholder feels much more practical here rather than the gist format. @wdawson and preferences?

Here is postman, for example

image

wdawson: Yeah, it would be great to have a real example here. Can definitely borrow from OpenAPI docs

wdawson: Also, let's go with json 😉

packages/insomnia-app/app/common/documentation.ts
```diff
@@ -6,10 +6,24 @@ export const docsTemplateTags = insomniaDocs('/article/171-template-tags');
...
+export const documentationLinks = {
+  introductionToInsomnia: {
+    title: 'Introduction to Insomnia',
+    url: docsIntroductionInsomnia,
+  },
+  workingWithDesignDocs: {
+    title: 'Working with Design Documents',
+    url: docsWorkingWithDesignDocs,
+  },
+} as const;
```
jackkav, line 29: can you explain? this doesn't seem conventional.

packages/insomnia-app/app/ui/components/wrapper-design.tsx
```diff
@@ -9,20 +10,31 @@ import { ACTIVITY_HOME, AUTOBIND_CFG } from '../../common/constants';
 import { initializeSpectral, isLintError } from '../../common/spectral';
 import type { ApiSpec } from '../../models/api-spec';
 import * as models from '../../models/index';
+import { superFaint } from '../css/css-in-js';
```
jackkav, line 13: :suspect:

packages/insomnia-app/app/ui/components/wrapper-design.tsx
```diff
@@ -145,25 +159,35 @@ export class WrapperDesign extends PureComponent<Props, State> {
     }
   }
 
+  _onUpdateContents() {
+    const { forceRefreshCounter } = this.state;
+    this.setState({ forceRefreshCounter: forceRefreshCounter + 1 });
```
jackkav, line 164: forceRefreshCounter + uniquenessKey is smelly react. Hope we can make this better,

jackkav (commented):
Since this PR hits the CodeMirror tech debt, I would approve it in the condition that addressing the debt was planned.

packages/insomnia-app/app/ui/components/wrapper-design.tsx
dimitropoulos: @develohpanda this copy was as directed by @wdawson.  @wdawson after these three PRs are in we can always loop back - take a look at @develohpanda's feedback and let me know what you think and we can follow-up.

packages/insomnia-app/app/ui/components/design-empty-state.tsx
dimitropoulos: @develohpanda this copy was as directed by @wdawson. @wdawson after these three PRs are in we can always loop back - take a look at @develohpanda's feedback and let me know what you think and we can follow-up.

packages/insomnia-app/app/ui/components/design-empty-state.tsx
dimitropoulos: Indeed, I used to use `Container` as well, but the thing is that term is globally overloaded (I mean, more than just in this codebase) because of the "Container Component vs Presentation Component" terminology that's ubiquitous in the react world.  Due to that, many CSS-in-JS projects I've seen started using the `Wrapper` lingo, instead to avoid confusion or implication of being a "Container Component".  So, I usually follow that.

re: WrapperHome and family, yeah, I know what you mean.  My thinking is that it shouldn't matter in the long run because those components are going away anyway.  We already deleted on recently (WrapperOnboarding) and we are going to delete another very soon (WrapperMigration), and the remaining 3 (design/debug/test)+1(home) are definitely on their way to being removed as well since we all agree that abstraction is unnecessary.

happy to chat more whenever on this

packages/insomnia-app/app/ui/components/design-empty-state.tsx
dimitropoulos: 873246cbfe25d2697553eb6196ea77fd81378ba9

packages/insomnia-app/app/ui/components/wrapper-design.tsx
dimitropoulos: I feel the same way, but I don't want to introduce magic numbers or new terminology: this is taken from https://github.com/Kong/insomnia/blob/914bcf30ed6af0775147551349f377530d7c6add/packages/insomnia-app/app/ui/css/layout/base.less#L356-L374

packages/insomnia-app/app/common/documentation.ts
dimitropoulos: It's fairly common (in our project, and others):


It's a [const assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions).

consider if we typed it like this:
code

if we then use it like this, TypeScript is happy.
Screenshot_20220113_084514

but you see the bug?
Screenshot_20220113_084543

The way around this is to not type it at all.  Then, you get this error:
Screenshot_20220113_084732

In most cases where this comes up, you _need_ to add the const assertion in order to get this behavior because the data you're working with has already been downcast (to something like string, as in my first example) and you need to make it more specific.  In order to make the intent clear to the reader that this is a const assertion (rather than relying on TypeScript's fallback for handling a literal) I added the `as const`, and also to prevent the bug of the type I showed above from appearing in the future if this data gets moved.

bonus points: we get intellisense if it's a const assertion:
Screenshot_20220113_085128

packages/insomnia-app/app/ui/components/wrapper-design.tsx
dimitropoulos: I couldn't agree more.  Like I said in the PR description: I'm not happy about this, but it's "where we're at" and I absolutely do not want to much around in CodeEditor for the purposes of this PR to get it to work as a controlled component, which is what would be required.

PR #4346, opened 2022-01-07
Update response-timeline-viewer to show response body.

p-m-j:
Whenever I have an issue with oauth in insomnia it takes an awful lot of effort to work out what has actually gone wrong.

Have to make use of debugging proxies, or add breakpoints in dev tools.

Adding the following (or similar if this isn't appropriate) would make life a lot easier for me.

image

image

changelog(Improvements): Insomnia will now show the response body for OAuth2 requests when there's an error

### DIFF
--- a/packages/insomnia-app/app/models/response.ts
+++ b/packages/insomnia-app/app/models/response.ts
@@ -7,6 +7,7 @@ import zlib from 'zlib';
 
 import { database as db, Query } from '../common/database';
 import { getDataDirectory } from '../common/electron-helpers';
+import { LIBCURL_DEBUG_MIGRATION_MAP } from '../network/network';
 import type { BaseModel } from './index';
 import * as models from './index';
 
@@ -228,8 +229,37 @@ export const getBodyBuffer = <TFail = null>(
     readFailureValue,
   );
 
-export function getTimeline(response: Response) {
-  return getTimelineFromPath(response.timelinePath || '');
+export function getTimeline(response: Response, showBody?: boolean) {
+  const { timelinePath, bodyPath } = response;
+
+  // No body, so return empty Buffer
+  if (!timelinePath) {
+    return [];
+  }
+
+  try {
+    const rawBuffer = fs.readFileSync(timelinePath);
+    const timeline = JSON.parse(rawBuffer.toString());
+    let body: ResponseTimelineEntry[] = [];
+    if (showBody) {
+      const name = LIBCURL_DEBUG_MIGRATION_MAP.DataOut;
+      body = [
+        {
+          name,
+          timestamp: Date.now(),
+          value: fs.readFileSync(bodyPath).toString(),
+        },
+      ];
+    }
+    const output = [
+      ...timeline,
+      ...body,
+    ] as ResponseTimelineEntry[];
+    return output;
+  } catch (err) {
+    console.warn('Failed to read response body', err.message);
+    return [];
+  }
 }
 
 function getBodyStreamFromPath<TFail extends Readable>(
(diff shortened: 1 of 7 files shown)


dimitropoulos:
Hi @p-m-j

We took a look at this PR today on the [Insomnia Stream](https://app.slack.com/client/T3C5CEYCD/C031YHPBA3G)

Screenshot_20220215_142546

Please feel free to verify, but I think we found a way to do exactly what you want while avoiding another problem that would come if we put the response body in the timeline at all times.

@gatzjames @jackkav @DMarby please take a look if you can.  https://github.com/schrodingersket/keycloak-oauth2-ssl-testbed will be a good thing to set up while you do, but check out the stream if you want to see how we tested it.  Basically you can just follow the instructions and just give keycloak a bad passoword to see an error.

p-m-j:
Hey @dimitropoulos, Would love to have a gander but that stream link just takes me to the slack "Sign in to your workspace" page.

dimitropoulos:
| BEFORE | AFTER |
| - | - |
| Screenshot_20220215_145204 | Screenshot_20220215_145049 |

I also put in a tiny commit to make the response timeline debug output more easily accessible.  before, you had to click on that little bug icon and literally nothing else.  in my opinion, it was effectively _hidden_ before, so felt like as good a time as any to fix that.

dimitropoulos:
@p-m-j that's because it's a link to a channel in the insomnia community chat.  You can sign up [here](https://getinsomnia.slack.com/).

In case you don't want to: [here](https://youtu.be/4pANlbmqppU?t=386) is a timestamped link to the YouTube video for the stream.  We're still in early-days with the stream, so apologies for the poor bitrate (we're evaluating tools to use for this (it's only the second stream we've done, the first being just last week) and all of them require you to pay to get any better quality).

p-m-j:
Couldn't sign up there but found https://chat.insomnia.rest/ which sends out an invite.

Cheers

dimitropoulos:
ohh, interesting.. hmm.. I'll have to learn how that all works, then.  thanks!

packages/insomnia-app/app/models/response.ts
```diff
@@ -228,8 +230,30 @@ export const getBodyBuffer = <TFail = null>(
...
+  try {
+    const rawBuffer = fs.readFileSync(timelinePath);
+    const timeline = JSON.parse(rawBuffer.toString()) as ResponseTimelineEntry[];
+    const body: ResponseTimelineEntry[] = showBody ? [
+      {
+        name: LIBCURL_DEBUG_MIGRATION_MAP.DataOut,
+        timestamp: Date.now(),
+        value: fs.readFileSync(bodyPath).toString(),
+      },
+    ] : [];
```
dimitropoulos, line 250: @jackkav since you were asking: this `body` variable, right here, is the core of what this PR adds.  Everything else is machinery leading up to this one change.

packages/insomnia-app/app/models/response.ts
```diff
@@ -228,8 +230,30 @@ export const getBodyBuffer = <TFail = null>(
...
+  if (!timelinePath) {
+    return [];
+  }
+
+  try {
+    const rawBuffer = fs.readFileSync(timelinePath);
+    const timeline = JSON.parse(rawBuffer.toString()) as ResponseTimelineEntry[];
+    const body: ResponseTimelineEntry[] = showBody ? [
+      {
+        name: LIBCURL_DEBUG_MIGRATION_MAP.DataOut,
```
filfreire, line 246: I'm not entirely sure if (and how) we should add some logic here to accommodate for the case where the body is supposedly `SslDataOut`, but I guess we can just leave it like that and maybe add a note here.

filfreire (approved):
@dimitropoulos marking this one as approved - since for the most part we checked it during the stream - only left 1 comment bellow. 

Best to wait for feedback from either @gatzjames @jackkav  @DMarby before merge.

packages/insomnia-app/app/models/response.ts
dimitropoulos: I looked into it and it turns out for whatever reason this type is explicitly ignored https://github.com/Kong/insomnia/blob/24017885093d08a527176f04159b8d37b34d3df1/packages/insomnia-app/app/network/network.ts#L314

DMarby (approved):

PR #4347, opened 2022-01-07
removes hugging face emoji

dimitropoulos:
We are trying to rid the app of emojis of this sort.  This is the only one of its kind that could find.

Screenshot_20220107_105327


wdawson (approved):
⛵

PR #4349, opened 2022-01-10
Allow overriding "coreAnnotations" on Kubernetes Ingress objects

ttyS0e:
The return value appender for the Ingress object annotations map is set up in the wrong order.

I would like to override the default `kubernetes.io/ingress.class: 'kong'` in an API Spec, like this:

```
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Customers MDM API
  x-kubernetes-ingress-metadata:
    annotations:
      kubernetes.io/ingress.class: 'mdm-workspace-ingress-class'
```

but I can't, because when `coreAnnotations` is appended to the map, it clobbers my value from the OAS.

changelog(Inso CLI): You can now override annotations on Kubernetes Ingress objects

### DIFF
--- a/packages/openapi-2-kong/src/kubernetes/generate.ts
+++ b/packages/openapi-2-kong/src/kubernetes/generate.ts
@@ -139,9 +139,9 @@ export const generateMetadataAnnotations = (
 
     const originalAnnotations = metadata?.annotations || {};
     return {
+      ...coreAnnotations,
       ...originalAnnotations,
       ...customAnnotations,
-      ...coreAnnotations,
     };
   }
 


develohpanda (approved):
I think the `coreAnnotations` was expanded last so as to ensure the required annotation(s) always exist, but if a user wants to override one they should have the ability to IMO. Thanks for the PR!

dimitropoulos (approved):
looks good to me as well.  been thinking about it all day and I think of what I would test manually beyond your use-case.  If we have to revisit later with a deep merge, then we can (to get more granular).

PR #4350, opened 2022-01-10
Adds helper in dev-only settings for launch counter

dimitropoulos:
Visibility, as well as the ability to update the launch counter is really helpful when we're working on features that look for particular values of stats.launches.  This PR adds a little helper to do so in the development-only settings section

Screenshot_20220110_171201

Screenshot_20220110_171220


develohpanda (approved):
💯 reviewed/tested on call, code LGMT

image

Hiding whitespace changes was very useful for review

PR #4351, opened 2022-01-11
updates support.insomnia.rest to docs.insomnia.rest

dimitropoulos:
closes INS-1351

It appears that this URL was updated, so we can point all links in the app to the right thing by just changing the factory.

### DIFF
--- a/packages/insomnia-app/app/common/documentation.ts
+++ b/packages/insomnia-app/app/common/documentation.ts
@@ -1,12 +1,12 @@
 const insomniaDocs = (slug: string) => `https://docs.insomnia.rest${slug}`;
 
 export const docsBase = insomniaDocs('/');
-export const docsGitSync = insomniaDocs('/article/193-git-sync');
-export const docsTemplateTags = insomniaDocs('/article/171-template-tags');
-export const docsVersionControl = insomniaDocs('/article/165-version-control-sync');
-export const docsPlugins = insomniaDocs('/article/173-plugins');
-export const docsImportExport = insomniaDocs('/article/172-importing-and-exporting-data');
-export const docsKeyMaps = insomniaDocs('/article/203-key-maps');
+export const docsGitSync = insomniaDocs('/insomnia/git-sync');
+export const docsTemplateTags = insomniaDocs('/insomnia/template-tags');
+export const docsVersionControl = insomniaDocs('/insomnia/version-control-sync');
+export const docsPlugins = insomniaDocs('/insomnia/introduction-to-plugins');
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/common/documentation.ts
```diff
@@ -1,4 +1,4 @@
-const insomniaDocs = (slug: string) => `https://support.insomnia.rest${slug}`;
+const insomniaDocs = (slug: string) => `https://docs.insomnia.rest${slug}`;
 
 export const docsBase = insomniaDocs('/');
 export const docsGitSync = insomniaDocs('/article/193-git-sync');
```
develohpanda: We should also update the slugs. The redirects are defined [here](https://github.com/Kong/insomnia-docs/blob/main/docs/vercel.json). This line for example, would become

```suggestion
export const docsGitSync = insomniaDocs('/insomnia/git-sync');
```

develohpanda (changes requested):

packages/insomnia-app/app/common/documentation.ts
dimitropoulos: good catch!  @falondarville could you go through them? if not lemme know

falondarville: Yep, I can pull this work and go through all of these!

falondarville: @dimitropoulos permission denied to push to your fork. Adding the fixes here til we can resolve permission issues: (Edit: file is `documentation.ts`)

```
export const docsBase = insomniaDocs('/');
export const docsGitSync = insomniaDocs('/insomnia/git-sync');
export const docsTemplateTags = insomniaDocs('/insomnia/template-tags');
export const docsVersionControl = insomniaDocs('/insomnia/version-control-sync');
export const docsPlugins = insomniaDocs('/insomnia/introduction-to-plugins');
export const docsImportExport = insomniaDocs('/insomnia/import-export-data');
export const docsKeyMaps = insomniaDocs('/insomnia/key-maps');
export const docsIntroductionInsomnia = insomniaDocs('/insomnia/get-started');
export const docsWorkingWithDesignDocs = insomniaDocs('/insomnia/design-documents');
```

falondarville (approved):
LGTM!

packages/insomnia-app/app/common/documentation.ts
dimitropoulos: (update: got it sorted out)

develohpanda (approved):
Woohoo TY!

PR #4362, opened 2022-01-13
Empty State for Test Tab

dimitropoulos:
This PR adds an Empty State for the Test Tab:

| BEFORE | AFTER |
| - | - |
| Screenshot_20220113_102252 | Screenshot_20220113_102148 |

And also creates an example test suite upon first launch.  Note that you can easily test this by using the new stats.launches helper added in https://github.com/Kong/insomnia/pull/4350.  Set it to `1` and do a soft refresh (`Shift`+`F5`)

Closes INS-1291


packages/insomnia-app/app/common/documentation.ts
```diff
@@ -26,4 +28,12 @@ export const documentationLinks = {
     title: 'Working with Design Documents',
     url: docsWorkingWithDesignDocs,
   },
+  unitTesting: {
+    title: 'Unit Testing',
+    url: docsUnitTesting,
```
develohpanda, line 33: Having the `docsUnitTesting` variable seems a bit verbose, why not just `url: insomniaDocs('/inso-cli/introduction')`? 😅

packages/insomnia-app/app/ui/redux/modules/global.tsx
```diff
@@ -741,9 +741,15 @@ export const initFirstLaunch = () => async (dispatch, getState) => {
...
+  const unitTestSuite = await models.unitTestSuite.create({
+    parentId: workspaceId,
+    name: 'Example Test Suite',
+  });
+
   await models.workspaceMeta.updateByParentId(workspaceId, {
     activeRequestId: request._id,
     activeActivity: ACTIVITY_DEBUG,
+    activeUnitTestSuiteId: unitTestSuite._id,
   });
```
develohpanda, line 745: There's a lot of stuff we expect to happen during first launch as defined in `initFirstLaunch`. It would be nice to add a unit test specifically for this action creator, to make sure the expected data is being created and code path reached. Any changes here that are not caught by a test and reliant on manual testing would have a detrimental effect on the onboarding UX.

develohpanda (changes requested):
This PR seems incomplete according to what I read in the title and description. If there's no test suite, there's no empty state. Ie. this PR only adds an empty state for when a suite is selected and has no tests.

image

Are there future tasks to add/update the empty states for where there is no request selected, or no unit test suite?

### QA

<details>
<summary>🟢 Creates expected data on initial launch</summary>

https://www.loom.com/share/929c5e4a7ba7434bafa6fbf1d9205236

</details>

<details>
<summary>🔴 Creating a new document/collection is different to my onboarding experience</summary>

As a user, this feels inconsistent because the UX we made them familiar with is not what they are faced with when starting a new document/collection.

https://www.loom.com/share/05d09ecb2ef94141b9934beea4fdd7b6

</details>

<details>
<summary>🔴 Overflow issue in test tab center pane</summary>

https://www.loom.com/share/05b49ff937b34b629e8fdd3b7c155df3

</details>

packages/insomnia-app/app/common/documentation.ts
dimitropoulos: to stay consistent with all other links, which all have their own variables.  also, to avoid potential duplication in the future if someone is looking for that link and doesn't see it in our existing list.

dimitropoulos:
@develohpanda 313ad0c2e73a292098d0fb02efa6e2a4c14f02f7 fixes the overflow issue.  it's been tricky to do without diverging too far from the other Pane usages, but thankfully this solution seems to work in all the situations:

develohpanda (approved):
Looks like overflow is fixed up now, thank you!

dimitropoulos:
Although this PR didn't do 90% of `initFirstLaunch`, I agree that it's time to have tests for it, and I'd like me/us to be accountable to that.  So, I created INS-1368, which will be a follow-up to all this.

packages/insomnia-app/app/ui/redux/modules/global.tsx
dimitropoulos: totally agree: https://github.com/Kong/insomnia/pull/4362#issuecomment-1015873136

PR #4364, opened 2022-01-13
removes DEPRECATED_ACTIVITY_INSOMNIA

dimitropoulos:
closes: INS-1314

per the existing tests for activity fallback, the usual fallbacks will apply if the app is booted with this activity, which was deprecated a year ago (https://github.com/Kong/insomnia/commit/26fb78ab99f3d16858fd4afe76d9ca3ba8fe7d92#diff-a5ade3251e8fc7740bbb8bdf5c47676cf76d7d066e76d4f56de0cd9763bd4cb7R159)

### DIFF
--- a/packages/insomnia-app/app/models/workspace-meta.ts
+++ b/packages/insomnia-app/app/models/workspace-meta.ts
@@ -1,9 +1,7 @@
 import {
-  ACTIVITY_DEBUG,
   DEFAULT_PANE_HEIGHT,
   DEFAULT_PANE_WIDTH,
   DEFAULT_SIDEBAR_WIDTH,
-  DEPRECATED_ACTIVITY_INSOMNIA,
 } from '../common/constants';
 import { database as db } from '../common/database';
 import type { BaseModel } from './index';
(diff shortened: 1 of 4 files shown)


packages/insomnia-app/app/models/workspace-meta.ts
```diff
@@ -63,15 +62,6 @@ export function init(): BaseWorkspaceMeta {
 }
 
 export function migrate(doc: WorkspaceMeta) {
```
dimitropoulos, line 63: it seemed fine, in the context of the yet-uncodefied api of a model, to just leave this noop function here, but if it's best to remove it then we can also remove it.

gatzjames (approved):
Great to see this being cleaned up! 👏
Not sure what the impact of removing the migrate function will be but if we decide to keep it as a noop we should make a ticket to keep track of the work that needs to happen next.

packages/insomnia-app/app/models/workspace-meta.ts
dimitropoulos: update: it is necessary

dimitropoulos:
update: I'm glad I followed my instinct not to remove it (despite no references) because actually it's called unconditionally in [`initModel`](https://github.com/Kong/insomnia/blob/4e9efb4d7d8223420a1f5deaea4fa69caf6118c9/packages/insomnia-app/app/models/index.ts#L194) for all models.  Turns out, many of the others have a noop `migrate`, including `ApiSpec`, `ClientCertificate`, `Environment`, `GitRepository`, `GrpcRequestMeta`, `GrpcRequest`, `OAuth2Token`, and many more.

so, that settles it: it definitely needs to stay, at least in the scope of this PR.

PR #4365, opened 2022-01-14
Git Sync Events

wdawson:
Instruments several events in the Git Sync flow so we can get a baseline for usage before trying to optimize.

I used generic `VCS _verb_` strings as the event type so we can keep our options open in the future if/when we have a VCS plugin interface. To that end, I've added properties to those events (with a function wrapper) to add more detail when needed. I'm sure this could be optimized but my goal was to get something done before we release the next version.

I've kept these events at the UI layer on purpose as the business logic underneath may be used in other applications and the analytics system may be different. I noticed that there were several places where business logic had crept into the UI code (e.g. `_handleMerge` function in [git-branches-modal.tsx](https://github.com/Kong/insomnia/blob/d53ef39c9b76aefcfa71630078d838bb95eaad56/packages/insomnia-app/app/ui/components/modals/git-branches-modal.tsx#L122-L129)). That made making clean calls to fire analytics events difficult but I did my best to work around that. Perhaps that's something that @gatzjames can tackle as a non-functional requirement of the Git Sync work (clear separation of a UI layer).

I also cleaned up some plugin events that we added in #3903. I added a comment to indicate my findings, but it seems like we have adequate defensive programming for Projects. I kept the conditional in the code but it's not clear to me whether or not we need it. I'll leave that to a future person to clean up, however.

Finally, I'm not sure if there are any tests worth adding here. As far as I can tell, they would be some unit tests that expect function calls and fail if those calls aren't there, which are my least favorite kinds of tests and I think would only cause more harm than good here.

Feel free to review in commit order if you'd like:
1. Cleanup for old plugin events
2. Add main Git Sync setup events
3. Add Git operation events
4. cleanup lint
5. cleanup tests

Closes INS-1355


packages/insomnia-app/app/plugins/context/data.ts
```diff
@@ -29,10 +28,11 @@ const buildImportRawConfig = (options: PluginImportOptions, activeProjectId: str
 
 const getWorkspaces = (activeProjectId?: string) => {
   if (activeProjectId) {
-    trackSegmentEvent(SegmentEvent.pluginExportLoadWorkspacesInProject);
     return models.workspace.findByParentId(activeProjectId);
   } else {
-    trackSegmentEvent(SegmentEvent.pluginExportLoadAllWokspace);
+    // This code path was kept in case there was ever a time when the app wouldn't have an active project.
+    // In over 5 months of monitoring in production, we never saw this happen.
+    // Keeping it for defensive purposes, but it's not clear if it's necessary.
```
wdawson, line 35: @develohpanda FYI

develohpanda: Thanks @wdawson! I think this is the right approach, keep the fallback for defensive purposes (it'll keep TS happy), with the comment. Thanks for removing the events.

packages/insomnia-app/app/common/analytics.ts
```diff
@@ -129,6 +127,17 @@ export enum SegmentEvent {
   unitTestDelete = 'Unit Test Deleted',
   unitTestRun = 'Ran Individual Unit Test',
   unitTestRunAll = 'Ran All Unit Tests',
+  vcsSyncStart = 'VCS Sync Started',
+  vcsSyncComplete = 'VCS Sync Completed',
+  vcsAction = 'VCS Action Executed',
+}
+
+export function vcsSegmentEventProperties(type: String, action: String, error?: string): Record<string, any> {
```
develohpanda: We can remove the return type and TS will automatically interpret it from the implementation 🎉 we can also be more specific with the allowed strings for each parameter so that we have type-safety in case of typos etc.

Two further suggested changed below to keep TS happy

```suggestion
type PushPull = 'push' | 'pull';

export function vcsSegmentEventProperties(
  type: 'git',
  action: PushPull | `force_${PushPull}` | 'create_branch' | 'merge_branch' | 'delete_branch' | 'checkout_branch' | 'commit' | 'stage_all' | 'stage' | 'restore' | 'update' | 'setup' | 'clone',
  error?: string
) {
```

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.tsx
```diff
@@ -186,6 +188,7 @@ class GitSyncDropdown extends PureComponent<Props, State> {
 
     try {
       await vcs.push(gitRepository.credentials, force);
+      trackSegmentEvent(SegmentEvent.vcsAction, vcsSegmentEventProperties('git', force ? 'force_' : '' + 'push'));
```
develohpanda:
```suggestion
      trackSegmentEvent(SegmentEvent.vcsAction, vcsSegmentEventProperties('git', force ? 'force_push' : 'push'));
```

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.tsx
```diff
@@ -203,6 +206,7 @@ class GitSyncDropdown extends PureComponent<Props, State> {
           title: 'Error Pushing Repository',
           error: err,
         });
+        trackSegmentEvent(SegmentEvent.vcsAction, vcsSegmentEventProperties('git', force ? 'force_' : '' + 'push', err.message));
```
develohpanda:
```suggestion
        trackSegmentEvent(SegmentEvent.vcsAction, vcsSegmentEventProperties('git', force ? 'force_push' : 'push', err.message));
```

packages/insomnia-app/app/ui/components/modals/git-staging-modal.tsx
```diff
@@ -120,6 +122,7 @@ export class GitStagingModal extends PureComponent<Props, State> {
       newItems[p].staged = doStage || forceAdd;
     }
 
+    trackSegmentEvent(SegmentEvent.vcsAction, vcsSegmentEventProperties('git', 'stage_all'));
```
develohpanda: This action toggles between all staged and all unstaged. The event probably needs an `unstage_all` equivalent, otherwise this will show double the data.

packages/insomnia-app/app/ui/components/modals/git-staging-modal.tsx
```diff
@@ -134,6 +137,7 @@ export class GitStagingModal extends PureComponent<Props, State> {
     }
 
     newItems[gitPath].staged = !newItems[gitPath].staged;
+    trackSegmentEvent(SegmentEvent.vcsAction, vcsSegmentEventProperties('git', 'stage'));
```
develohpanda: Same as above, this toggles one item between staged and unstaged, probably needs an `unstage` equivalent event

packages/insomnia-app/app/ui/components/modals/git-staging-modal.tsx
```diff
@@ -303,6 +307,7 @@ export class GitStagingModal extends PureComponent<Props, State> {
         status: i.status,
       }));
     await gitRollback(vcs, files);
+    trackSegmentEvent(SegmentEvent.vcsAction, vcsSegmentEventProperties('git', 'restore'));
```
develohpanda: If you look at the two usages of this function, one of the usages only sends a single item in the array (to `restore` one item), and the other sends all items (to `restore_all`).

FYI each of these five buttons will fire this event

image

packages/insomnia-app/app/ui/redux/modules/git.tsx
```diff
@@ -196,6 +208,7 @@ export const cloneGitRepository = ({ createFsClient }: {
         if (!(await containsInsomniaWorkspaceDir(fsClient))) {
           dispatch(noDocumentFound(repoSettingsPatch));
           dispatch(loadStop());
+          trackSegmentEvent(SegmentEvent.vcsSyncComplete, vcsSegmentEventProperties('git', 'clone', 'no workspace found'));
```
develohpanda: This happens when the `.insomnia/workspace` dir cannot be found

packages/insomnia-app/app/ui/redux/modules/git.tsx
```diff
@@ -205,12 +218,14 @@ export const cloneGitRepository = ({ createFsClient }: {
         if (workspaces.length === 0) {
           dispatch(noDocumentFound(repoSettingsPatch));
           dispatch(loadStop());
+          trackSegmentEvent(SegmentEvent.vcsSyncComplete, vcsSegmentEventProperties('git', 'clone', 'workspaces empty'));
```
develohpanda: This happens when the `.insomnia/workspace` dir is empty, or no workspace can be found

develohpanda (commented):
🎉 

> I noticed that there were several places where business logic had crept into the UI code

Unfortunately this has historically been the norm for a large majority of the code-base. This lack of separation is one aspect of what leads to poor test coverage. I would be hesitant to make this refactor a non-functional requirement (feels like scope-creep, because it would increase the review and QA overhead), but as a standalone tech debt item it would be great to tackle.

> Finally, I'm not sure if there are any tests worth adding here. 

Seems fine to me, thanks for checking!

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.tsx
wdawson: Ha I knew there was a better way as I was writing this. Thanks 🤦‍♂️

packages/insomnia-app/app/ui/redux/modules/git.tsx
wdawson: Want me to add a comment to the code or change the message?

packages/insomnia-app/app/ui/redux/modules/git.tsx
wdawson: Hmm... Is there a reason why we're checking that twice then? Seems like this and line 211 were for different cases. 

Want me to add a comment to the code or change the message?

packages/insomnia-app/app/ui/components/modals/git-staging-modal.tsx
wdawson: I was trying to cover that with the `restore` events since it seemed like that's the action we provide. But fair point about trying to differentiate between `restore` and `restore_all`.

But maybe I'm misunderstanding your feedback?

wdawson: Oh I think I see now. Since this function is a toggle, I will need to add in the restore events here too. Will do.

EDIT: Oh except this is a bit different than the "rollback" thing (even if `git restore ...` will do them both) ... hmm... Now I'm wondering if it's worth having staging events at all... seems like it could be noisy.

wdawson: Decided to keep them and change `restore` to `rollback` which should help remove any git-centric implementation too. Refactored a bit to get what we want here.

packages/insomnia-app/app/ui/components/modals/git-staging-modal.tsx
wdawson: Yeah originally I was fine with that but decided to separate with a small refactor.

packages/insomnia-app/app/ui/redux/modules/git.tsx
develohpanda: It's not checking twice, it's checking two different things. First we're checking if the directory exists. If it doesn't we exit, and if it does we try read it. After reading it, if we don't find what we expect (this check), then we exit, otherwise we continue.

Could probably update it but yeah, slightly different cases. I think we should update the message to be "no workspaces" or similar, instead of "empty workspace".

packages/insomnia-app/app/ui/redux/modules/git.tsx
develohpanda: This kind of represents a malformed `.insomnia` directory, so the message could be "malformed directory"? I'll leave that to you

develohpanda (approved):
Refactor and changes LGTM, couple of comments regarding the empty workspace events, but happy to approve 🚢

packages/insomnia-app/app/ui/components/modals/git-staging-modal.tsx
develohpanda: Cool! I like what you've done now, looks good 👍🏽

packages/insomnia-app/app/ui/components/dropdowns/git-sync-dropdown.tsx
develohpanda: Heh it's okay! :)

gatzjames (approved):
Code looks great and really helpful for the upcoming changes!
Agree with @develohpanda on the messages and kudos for the great explanation of how things work there!

wdawson:
Updated the last messages to address nits: https://github.com/Kong/insomnia/pull/4365/commits/6616898d7190f831f02acc58c4cd48545457ce37

And then will merge once tests pass again.

PR #4380, opened 2022-01-18
suggestion: allow override of the kubernetes resource name

ttyS0e:
A client has highlighted two things:

1. If you have multiple API specifications with similar plugins, the Kubernetes resource names (for KongPlugin) collide
2. The Kube resource names don't mean much to their support team, as they are given incremental names based on the actual Lua plugin name itself

So I had added a simple feature that allows us to override the generated Kubernetes KongPlugin object name, for example this API specification snippet:

```yaml
paths:
  /pets:
    x-kong-plugin-rate-limiting-advanced:
      name: rate-limit-advanced
      resourceName: us-customers-30-per-minute
      enabled: true
      config:
        limit:
        - 30
        window_size:
        - 60
```

will generate this Kube config snippet:

```yaml
apiVersion: configuration.konghq.com/v1                                                                                                                                                
kind: KongPlugin
metadata:
  name: add-us-customers-30-per-minute-p0
plugin: rate-limit-advanced
config:
  limits:
    - 30
  window_sizes:
    - 60
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: swagger-petstore-0
  annotations:
    konghq.com/plugins: add-us-customers-30-per-minute-p0
    kubernetes.io/ingress.class: kong
spec:
  rules:
...
...
```

Please let me know your thoughts on:

1. Is this the right way to go about this feature, and is this okay to releasea
2. If so, should the "resourceName" field be something different?

Thank you!


dimitropoulos:
Apologies for not seeing this PR sooner.  This seems pretty important for us to have considering how important kubernetes resource names are.  Do you think you'd be able to write a unit test for this (could be, just, like, the example you gave in the PR description above)?  If not or if you don't have time please let me know and I'll make time.

Thanks!

PR #4382, opened 2022-01-19
Spike: Run nightly schedule job if there's a recent commit (INS-1349)

filfreire:
This is a small experiment I'm doing for [INS-1349](https://linear.app/insomnia/issue/INS-1349/create-nightly-releases-rfc) - the idea is we would trigger nightly build related jobs on a schedule (e.g. every day at 0:30 UTC, we can discuss the date in the RFC), but only if there has been a recent commit.

cc @wdawson @wongstein @develohpanda


filfreire:
Closing this spike for now, we'll get back to Nightly releases work soon 😉

PR #4383, opened 2022-01-19
Segment Queueable Events for App Started

dimitropoulos:
This PR starts by cleaning up some existing errors in our analytics handling code, specifically for queued events, and using named arguments instead of a chain of 5 unnamed arguments (some of which were booleans) for internal tracking functions.

Once that baseline is established, ea6328ea113b4a81d14edac9d94dc9950be0044e switches the implementation of the "Launched", "First Launch", and "Updated" events to use Segment, all with the "App Started" event (now a member of the `SegmentEvents` enum).

Then, since the queuing system was _only_ used for these events alone, it ports the basic logic of that system to the segment events.  We're going to be moving away from google analytics entirely, so this is a nice two-for-one change, that these google analytics-specific codedpaths weren't used anywhere else.

Added in 31def5a2e230a5935027c72cc40a37354d60fce1 is a way of keeping the backdated timestamp to the events, something we didn't have before but would benefit from.  I followed https://segment.com/docs/connections/spec/common to determine how to do this, and it appears sending as a regular parameter is the way to go, but please double check my reading of the docs.

closes INS-1326


packages/insomnia-app/app/common/analytics.ts
```diff
@@ -143,16 +116,56 @@ export function vcsSegmentEventProperties(
   error?: string
 ) {
   return {
-    'type': type,
```
wdawson: I left these labels explicitly because if the variable names change, we don't want the keys to change. Is it a problem to keep them?

packages/insomnia-app/app/common/analytics.ts
```diff
@@ -143,16 +116,56 @@ export function vcsSegmentEventProperties(
...
-    'error': error,
+    type,
+    action,
+    error,
   };
 }
 
-export async function trackSegmentEvent(event: SegmentEvent, properties?: Record<string, any>) {
+interface QueuedEvent {
+  event: SegmentEvent;
```
wdawson, line 126: should we add a timestamp property to this so we can replay based on when the event actually occurred?

wdawson (commented):
Looks good with two minor comments

packages/insomnia-app/app/common/analytics.ts
develohpanda: This refactor is probably fine but regardless we should write a unit test to make sure the expected keys exist

packages/insomnia-app/app/common/analytics.ts
```diff
@@ -163,16 +116,56 @@ export function vcsSegmentEventProperties(
...
+let queuedEvents: QueuedEvent[] = [];
+
+async function flushQueuedEvents() {
+  console.log(`[segment] Flushing ${queuedEvents.length} queued events`);
+  const params = [...queuedEvents];
+
+  // Clear queue before we even start sending to prevent races
+  queuedEvents = [];
+
+  console.log('[segment] Flushing queued event', params);
```
develohpanda: Do you mean to put this console log inside the `params.map` call?

packages/insomnia-app/app/main.development.ts
```diff
@@ -200,21 +200,14 @@ async function _trackStats() {
...
-
-  if (firstLaunch) {
-    trackNonInteractiveEventQueueable('General', 'First Launch', stats.currentVersion);
-  } else if (justUpdated) {
-    trackNonInteractiveEventQueueable('General', 'Updated', stats.currentVersion);
-  } else {
-    trackNonInteractiveEventQueueable('General', 'Launched', stats.currentVersion);
-  }
+
+  trackSegmentEvent(SegmentEvent.appStarted, {}, { queueable: true });
```
develohpanda, line 204: We're intentionally dropping the version here? cc: @wdawson

wdawson: We automatically track the version in all segment events so we don't need it duplicated.

That's done [here](https://github.com/Kong/insomnia/pull/4383/files/1374cb8df2b259a6bd2ec4a795e484628d30a5ef..ea6328ea113b4a81d14edac9d94dc9950be0044e#diff-d980598f2592967d530bb1501b5f2844c5e5af99db85dd841a0c854fdfe01a64R196)

packages/insomnia-app/app/common/analytics.ts
dimitropoulos: first off: this is out of scope of this PR in my opinion, and I would have caught and removed it in my first self-review if I had done one (i.e. had not been in such a hurry to get this posted).  for that reason alone, I pulled it out in d239337660570fc368c816d2ccbddefc15c2b38e

secondly, though, it's almost by random chance that we don't already have https://eslint.org/docs/rules/object-shorthand and we absolutely should have a test to cover this because I suspect we'll be turning that rule on sooner than later.  I've been meaning to, recently, actually.

packages/insomnia-app/app/common/analytics.ts
dimitropoulos: great call, yes.  that was missing in the prior implementation.  easy improvement if I did it right: 31def5a2e230a5935027c72cc40a37354d60fce1

packages/insomnia-app/app/common/analytics.ts
dimitropoulos: looking at it, we really have no need for it anymore because we're already logging the first line, we can just add this there pretty simply: lemme know what you think of 966e4edcba6b235649a38b64ad885f3ea76db7a1 but I think the way that commit does it should be good.  wouldn't want to blow up the console with 1000 log lines if someone toggles analytics after an hour or whathaveyou

packages/insomnia-app/app/common/analytics.ts
```diff
@@ -187,9 +211,13 @@ export async function trackSegmentEvent(event: SegmentEvent, properties?: Record
...
       },
+    }, error => {
+      if (error) {
+        console.warn('[analytics] Error sending segment event', error);
+      }
     });
-  } catch (err) {
-    console.warn('[analytics] Error sending segment event', err);
+  } catch (error: unknown) {
+    console.warn('[analytics] Error sending segment event', error);
```
develohpanda: I think it'd be good to differentiate these two error messages _somehow_ so that when the error is seen we are able to determine which error handler threw the error. The stack trace would probably be from minified files so would be little help

develohpanda (approved):
LGTM

packages/insomnia-app/app/common/analytics.ts
develohpanda:
```suggestion
    console.warn('[analytics] Unexpected error while sending segment event', error);
```

dimitropoulos: https://github.com/Kong/insomnia/pull/4383/commits/34573839bdfa85eb3403cfeb248f735e36913b95

wdawson (approved):
Thanks for the work!

PR #4384, opened 2022-01-20
Add resize for recording menu item

develohpanda:
This helper is useful for resizing the application to a consistent size. Currently this is useful for twitter recordings for insomnia thursdays, but it could be useful for docs and PRs in the future as well.

Currently hardcoded for recording, could make it configurable in future.

image


packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -274,6 +274,27 @@ export function createWindow() {
...
+        click: () =>
+          mainWindow?.setBounds({
+            x: 100,
+            y: 100,
+            width: DEFAULT_WIDTH,
+            height: DEFAULT_HEIGHT,
+          }),
+      },
+      {
+        label: `Resize for Recordin${MNEMONIC_SYM}g`,
```
wdawson: Assuming this will be visible to everyone who uses the app, we should rename this something else that is clearer. We use that size for recording but other people may not want to. Something like "Resize to Small" for example

develohpanda: I like that!

develohpanda:
```suggestion
        label: `Resize to ${MNEMONIC_SYM}Small`,
```

wdawson (approved):
Looks fine to me from the product side

gatzjames:
TIL about mnemonics in windows and linux!
LGTM 👍 Really handy feature!

gatzjames (approved):

PR #4385, opened 2022-01-20
add test for URL being focused on first time user experience

filfreire:
> Note: the tests for this are failing because they are reproducing a bug - on first time user experience we are not focusing the url input element.

closes: INS-1388

cc @dimitropoulos


packages/insomnia-smoke-test/tests/app.test.ts
```diff
@@ -1,6 +1,17 @@
+import { expect } from '@playwright/test';
+
 import { loadFixture } from '../playwright/paths';
 import { test } from '../playwright/test';
 
+test('url field is focused for first time users', async ({ page }) => {
+  const textArea = ':nth-match(textarea, 2)';
+  // Uncomment to click on and force element to be focused:
+  // const urlInput = "//div[contains(@class, 'editor__container input editor--single-line')]";
+  // page.click(urlInput);
```
dimitropoulos: is there anywhere else we can have this or incorporate it into a test?  What I'm trying to avoid is commented code which will fall out of date.

packages/insomnia-smoke-test/tests/app.test.ts
```diff
@@ -1,6 +1,17 @@
+import { expect } from '@playwright/test';
+
 import { loadFixture } from '../playwright/paths';
 import { test } from '../playwright/test';
 
+test('url field is focused for first time users', async ({ page }) => {
```
gatzjames, line 6: Since this is testing the url bar only should we move this inside another test that tests a bigger user flow?

filfreire: @gatzjames I wanted to split the focus here in case only the focus was broken, not impacting the other "sub" tests.

Perhaps we could use something like [playwright's test steps](https://playwright.dev/docs/api/class-test#test-step)?

```ts
// imports

test('first time user experience', async ({ page }) => {
  await test.step('url input is foccused', async () => {
    // ...
  });
  await test.step('can send requests', async () => {
    // ...
  });
});
```

packages/insomnia-smoke-test/tests/app.test.ts
filfreire: In this case we can delete it. Already pushed a fix.

packages/insomnia-smoke-test/tests/app.test.ts
gatzjames: The main concern was adding too many small tests might lead to a slow e2e suite cause of the setup/teardown costs.
Not sure about the benefits of the `.step` api but looks handy and something to investigate!
I'd say we can add this and keep an eye on small e2e tests in general.

DMarby (approved):

PR #4404, opened 2022-01-25
removes insomnia designer migration code

dimitropoulos:
closes INS-1311
changelog(Breaking Changes): We removed the ability to migrate from Insomnia Designer

This PR removes all functionality (or mention) of migration from Insomnia Designer from documentation, tests, source code, user settings, etc.


jackkav (approved):
🥳

packages/insomnia-inso/src/db/index.ts
```diff
@@ -73,17 +73,6 @@ export const loadDb = async ({
...
-    if (!db && !appDataDir && !src) {
-      const designerDir = getAppDataDir('Insomnia Designer');
-      db = await neDbAdapter(designerDir);
-      db &&
-        logger.debug(
-          `Data store configured from Insomnia Designer app data directory at \`${path.resolve(
-            designerDir,
-          )}\``,
-        );
-    }
```
dimitropoulos, line 86: @falondarville @wdawson I just want to make sure you're aware of this: it means that `inso` will also not have any knowledge of anything related to insomnia designer.  It makes sense to remove here since literally all other mention of designer is being removed as well, but I just wanted to call it out since I don't think we'd talked about it prior.

PR #4410, opened 2022-01-25
convert WrapperDesign to a function component

dimitropoulos:
closes INS-1389

This PR converts the WrapperDesign component to a function component and tries to change as little else as possible.  3 of the 4 sections were pretty straightforward, but the Editor one had some fidgeting that was necessary to keep the debounce behavior intact.  There are no intended behavioral changes as a result of this PR.

I encourage any reviewer to read commit by commit.  With the exception of the very last commit, which contains a small fix, I tried to keep the scope of each commit constrained to what's in the commit message.

### DIFF
--- a/packages/insomnia-app/app/ui/components/wrapper-design.tsx
+++ b/packages/insomnia-app/app/ui/components/wrapper-design.tsx
@@ -8,7 +8,6 @@ import SwaggerUI from 'swagger-ui-react';
 import { parseApiSpec, ParsedApiSpec } from '../../common/api-specs';
 import type { GlobalActivity } from '../../common/constants';
 import { initializeSpectral, isLintError } from '../../common/spectral';
-import type { ApiSpec } from '../../models/api-spec';
 import * as models from '../../models/index';
 import { superFaint } from '../css/css-in-js';
 import previewIcon from '../images/icn-eye.svg';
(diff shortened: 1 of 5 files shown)


packages/insomnia-app/app/ui/components/wrapper-design.tsx
```diff
@@ -31,291 +30,285 @@ const EmptySpaceHelper = styled.div({
...
 
-    return (
+  return (
+    <div id="swagger-ui-wrapper">
       <ErrorBoundary
         invalidationKey={activeApiSpec.contents}
         renderError={() => (
           <div className="text-left margin pad">
-            <h4>An error occurred while trying to render your spec's navigation.</h4>
+            <h3>An error occurred while trying to render Swagger UI</h3>
```
gatzjames, line 194: Should we leak about Swagger UI in the UI?
I think the previous wording was more general

dimitropoulos: I changed the wording???

dimitropoulos: ok, @gatzjames just look a little closer.  wow, had me worried there!  this is just git being strange.  I didn't change any wording anywhere (and don't want to in the context of the scope of this PR).  It's just git being confused about lines.  This is just very similar to a different message we show, which is also preserved.

gatzjames: 🙈 That was so close though 😂

gatzjames (approved):
Awesome! LGTM!

PR #4412, opened 2022-01-26
feat(ci): bump homebrew formulae

curiositycasualty:
On pushes to develop (as when merging PRs), attempt to open version bump PRs for the `insomnia`, `inso`, and `inso-beta` Homebrew formulae.

### DIFF
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -1,0 +1,52 @@
+name: Update Homebrew
+
+on:
+  schedule:
+    # weekdays at 00:00
+    - cron: '0 0 * * 1-5'
+
+  workflow_dispatch:
+
+jobs:
+  update_homebrew:
+    name: Update Insomnia Formula
+    # must be macos, linux-brew doesn't have casks
+    runs-on: macos-latest
+    steps:
+      - name: Set up Homebrew
+        id: set-up-homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+
+      - name: Cache Homebrew Bundler RubyGems
+        id: cache
+        uses: actions/cache@v1
+        with:
+          path: ${{ steps.set-up-homebrew.outputs.gems-path }}
+          key: ${{ runner.os }}-rubygems-${{ steps.set-up-homebrew.outputs.gems-hash }}
+          restore-keys: ${{ runner.os }}-rubygems-
+
+      - name: Install Homebrew Bundler RubyGems
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: brew install-bundler-gems
+
+      - name: Add "Versions" Tap
+        run: brew tap homebrew/homebrew-cask-versions
+
+      # Update Homebrew's Inso(mnia) formulae
+      # https://github.com/Homebrew/actions/tree/master/bump-formulae
+      - name: Bump Inso (Beta) Formula
+        uses: Homebrew/actions/bump-formulae@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          formulae: --open-pr inso-beta
+      - name: Bump Inso Formula
+        uses: Homebrew/actions/bump-formulae@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          formulae: --open-pr inso
+      - name: Bump Insomnia Formula
+        uses: Homebrew/actions/bump-formulae@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          # Bump only these formulae if outdated
+          formulae: --open-pr insomnia
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -37,16 +37,16 @@ jobs:
       - name: Bump Inso (Beta) Formula
         uses: Homebrew/actions/bump-formulae@master
         with:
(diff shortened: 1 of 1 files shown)


develohpanda:
Should these happen on merge to develop or be a part of the release workflow? We don't publish anything on merge to develop.

curiositycasualty:
@develohpanda Ideally these would be part of workflows that are run when a new version of either Insomnia or Inso are released. And ideally, those would be 2 separate release workflows, one per. But the current structure of the "release-core" workflow, coupled with the whole "2 projects in 1 repo" thing, make trying to tie the bumping of the Homebrew versions into the existing stuff kind of a nightmare.

This "cron" approach also ensures that PRs for/releases of inso(mnia) won't be blocked due to unrelated Homebrew issues should they arise.

develohpanda:
Fair enough. I don't like how with GitHub actions you can't separate workflows into individual files and reference/reuse them (like in azure devops). Really bloats each workflow.

I like that separating this workflow saves the release workflow in case of failure. Misread from the description that this will run on each merge to develop. Looks like it's a cron job or manual dispatch 🎉

develohpanda (approved):
Approving, but please wait on @filfreire, @johnwchadwick or @DMarby to approve and merge 😊

johnwchadwick:
I've added `HOMEBREW_PR_GH_TOKEN` to the secrets, because there is no such secret `GITHUB_TOKEN` :) Please switch to using that secret instead.

I'm currently looking into the documentation for this PR to validate that it works how we expect it to. For example, I want to ensure that it is smart enough to know _when_ to open a PR. If it isn't, we probably need to move this from running on cron to on specific tag events or something along those lines.

(edit: Or as David has suggested to me, possibly even just part of the Release workflow we already run.)

.github/workflows/homebrew.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Add "Versions" Tap
+        run: brew tap homebrew/homebrew-cask-versions
+
+      # Update Homebrew's Inso(mnia) formulae
+      # https://github.com/Homebrew/actions/tree/master/bump-formulae
+      - name: Bump Inso (Beta) Formula
+        uses: Homebrew/actions/bump-formulae@master
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
johnwchadwick: Change `secrets.GITHUB_TOKEN` to `secrets.HOMEBREW_PR_GH_TOKEN`, otherwise LGTM.

johnwchadwick (changes requested):
Looked into it and it seems like it's OK to run on cron because it relies on livecheck. However, before it is merged the GH token secret name needs to be updated to the newly-minted one. LGTM as soon as that's done.

curiositycasualty:
> I don't like how with GitHub actions you can't separate workflows into individual files and reference/reuse them (like in azure devops) ~ @develohpanda 

You absolutely can: https://docs.github.com/en/actions/using-workflows/reusing-workflows see also: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows
I didn't undertake the work to create better/cleaner/leaner workflows in this PR because it seemed like too much scope-creep, combined with the fact that I was optimistically hoping that the splitting of the repo (into separate Insomnia and Inso repos) might be in discussion elsewhere. In which case, each repo would similarly have separate workflows/github actions config.

> I'm currently looking into the documentation for this PR to validate that it works how we expect it to. For example, I want to ensure that it is smart enough to know when to open a PR. If it isn't, we probably need to move this from running on cron to on specific tag events or something along those lines. ~ @johnwchadwick 

My understanding from the docs are that the action does nothing if no PR is required: https://github.com/Homebrew/actions/blame/master/bump-formulae/README.md#L11

In essence though, these github actions equivalent to just running `brew bump --open-pr inso-beta` -- as one would locally. The logic for what version to update Homebrew to is actually all handled in the casks/forumlae themselves (via the `livecheck` block). These are the `livecheck` blocks for our formulae:
  - `insomnia`: https://github.com/Homebrew/homebrew-cask/blob/master/Casks/insomnia.rb#L11-L15
  - `inso`: https://github.com/Homebrew/homebrew-cask/blob/master/Casks/inso.rb#L11-L15
  - `inso-beta`: https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/inso-beta.rb#L11-L15

You can get a sense of how these `livecheck` blocks work: `brew livecheck --debug inso{,-beta,mnia}`. The benefit of the way Homebrew designed this process to work is that automations like this one don't need to try to figure out versions or github releases or any other fiddly semantic versioning logic.

You can also view the workflow runs that happened while I was developing these on my fork: https://github.com/curiositycasualty/insomnia/actions/runs/1749756124

johnwchadwick:
Yep, SGTM. I did look into it and it seems like livecheck is going to do the right thing. I just wanted to validate it myself rather than make the assumption, since it wasn't immediately obvious to me.

curiositycasualty:
Even with the fork, it's been kinda tough to test this without Homebrew being out-of-sync with the Insomnia/Inso releases. So I expect some tweaking will be required upon the next release effecting any of the homebrew formulae.

johnwchadwick (approved):

DMarby (approved):

filfreire (approved):

PR #4428, opened 2022-02-01
updates concurrently, lerna, eslint

dimitropoulos:
I noticed there are some super-simple things we can do that will improve our daily development experience.

1. `concurrently` was many years out of date, which caused colors to be dropped from output.  all that needed to be done was update the dependency.
2. `lerna` can be upgraded to latest with no breaking changes
3. `eslint` can be easily upgraded to the next major with very little effort.

And, best of all, upgrading eslint shaves 45.7 seconds off our lint time (i.e. running on every CI): from 56 seconds down to 11.

| BEFORE | AFTER |
| - | - |
| Screenshot_20220201_122803 | Screenshot_20220201_122301 |


filfreire (approved):
I suspect for the `eslint` time reduction, the difference might not be there.

I tried two clean clones, one from latest `develop` and another from this branch. The initial `npm run lint` would take around ~80 seconds for both branches.

This branch: 

Screenshot 2022-02-02 at 15 28 31


Latest `develop`:

Screenshot 2022-02-02 at 15 29 09

Screenshot 2022-02-02 at 15 29 12


The subsequent `npm run lint` runs were less than 10 seconds for each.

PR #4429, opened 2022-02-01
Add GitHub OAuth for git repos

gatzjames:
Adds a flow for users to sign in with Github and use the generated token to perform git operations.

TODO:
- [x] Use the actual insomnia endpoints for requests
- [x] Cache GraphQL API requests to GitHub or use the REST search API
- [x] Error handling
- [x] e2e test
- [x] Security Audit

Closes INS-138
gh-oauth

changelog(Improvements): Added GitHub login for git sync


wdawson:
We should also get Design to provide input on any UI changes before those get released

packages/insomnia-app/package.json
```diff
@@ -96,7 +97,6 @@
     "framer-motion": "^1.11.1",
     "fs-extra": "^5.0.0",
     "fuzzysort": "^1.1.1",
-    "graphql": "^14.5.8",
```
DMarby: I think this is probably unintentional, we still use this in other parts of the app right?

gatzjames: Yep nice one removed the example with relay and removed graphql too.
Tests still passed though 😅

packages/insomnia-app/app/sync/git/github-oauth-provider.ts
```diff
@@ -0,0 +1,65 @@
...
+        code,
+      },
+    }).then(result => {
+      statesCache.delete(state);
+      setAccessToken(result.data.access_token);
+    });
+  } else {
+    throw new Error('Invalid state parameter');
+  }
+}
```
jackkav: the intent here is unclear, what state?

packages/insomnia-app/app/sync/git/utils.ts
```diff
@@ -10,30 +12,58 @@ export const translateSSHtoHTTP = (url: string) => {
...
+        };
+      }
+      default: {
+        return {
+          username: credentials.username,
+          password: credentials.token,
+        };
+      }
+    }
+  }
```
jackkav: this is subtly supporting an unnamed oauth2format, I could use some hand holding here to know what case this is supporting, or for the code to express that.

packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal/custom-repository-settings-form-group.tsx
```diff
@@ -0,0 +1,114 @@
...
+export interface Props {
+  onInputChange: (e: React.SyntheticEvent<HTMLInputElement>) => void;
+  isGitRepository: boolean;
+  inputs: {
+    uri?: string;
+    authorName?: string;
+    authorEmail?: string;
+    username?: string;
+    token?: string;
+    oauth2format?: 'custom' | 'github';
```
jackkav: 'custom' could use some explanation somewhere i'm guessing its a catch all for bitbucket and gitlab, but it wasn't clear from the PR description that was scoped into this PR. "Add Github OAuth for git repos"

packages/insomnia-components/src/combobox/combobox.tsx
```diff
@@ -0,0 +1,141 @@
...
+  ComboboxOption as _ComboboxOption,
+  ComboboxOptionProps,
+  ComboboxOptionText,
+  ComboboxPopover as _ComboboxPopover,
+  ComboboxPopoverProps,
+  ComboboxProps,
+  escapeRegexp,
+  useComboboxContext,
+  useComboboxOptionContext,
+} from '@reach/combobox';
```
jackkav: Consider putting this somewhere besides insomnia-components

jackkav (commented):
The exploration of alternative git oauth platforms is useful to avoid painting ourselves into a corner but i would advise against including them in this PR but keep the abstractions they necessitated. That way it is much easier to test, review and merge. ✌️

packages/insomnia-smoke-test/playwright/test.ts
```diff
@@ -12,14 +12,22 @@ import {
...
+  INSOMNIA_API_URL: string;
+  INSOMNIA_APP_WEBSITE_URL: string;
+  INSOMNIA_GITHUB_API_URL: string;
 }
 
 export const test = baseTest.extend<{
   app: ElectronApplication;
 }>({
   app: async ({ playwright, trace }, use, testInfo) => {
+    const webServerBaseUrl = 'http://127.0.0.1:4010';
```
DMarby: I think we can possibly get this from the `baseURL` fixture instead? https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests

packages/insomnia-smoke-test/tests/git-sync.test.ts
```diff
@@ -0,0 +1,57 @@
+// import { expect } from '@playwright/test';
```
DMarby: Duplicate import here?

packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
```diff
@@ -0,0 +1,343 @@
...
+          setUser(data.viewer);
+        }
+      });
+    }
+
+    return () => {
+      isMounted = false;
+      controller.abort();
+    };
+  }, [githubToken, onChange, setUser]);
```
jackkav: it's not clear why AbortController is used here.

jackkav (approved):
Tested LGTM 🚀

packages/insomnia-app/app/sync/git/utils.ts
```diff
@@ -10,28 +12,46 @@ export const translateSSHtoHTTP = (url: string) => {
...
+    password: credentials.password || credentials.token,
+  };
+};
 
 export const gitCallbacks = (credentials?: GitCredentials | null) => ({
   onMessage,
   onAuthFailure,
   onAuthSuccess,
-  onAuth: onAuth(credentials || undefined),
+  onAuth: onAuth(credentials ?? undefined),
```
jackkav, line 56: since 0 or '' is not a valid credential, nullish coalescence isn't a value adding change here, I'm tempted to suggest leaving this as it is or handling null input more explicitly.

packages/insomnia-app/app/sync/git/github-oauth-provider.ts
```diff
@@ -0,0 +1,71 @@
...
+  }).toString();
+
+  return url.toString();
+}
+
+export async function exchangeCodeForToken(input: {
+  code: string;
+  state: string;
+}) {
+  const { code, state } = input;
```
johnwchadwick: nit: Since input is not used other than to destructure it, you may as well just destructure input in the arguments instead of here.

```typescript
export async function exchangeCodeForToken({code, state}: {
  code: string;
  state: string;
}) {
```

packages/insomnia-app/app/sync/git/git-vcs.ts
```diff
@@ -15,17 +15,27 @@ export interface GitRemoteConfig {
...
 }
 
-interface GitCredentialsToken {
+interface GitCredentialsOAuth {
+  /**
+   * Supported OAuth formats.
+   * This is needed by isomorphic-git to be able to push/pull using an oauth2 token.
+   * https://isomorphic-git.org/docs/en/authentication.html
+  */
+  oauth2format?: 'github';
```
johnwchadwick, line 29: Why is this optional? Is there supposed to be a valid `GitCredentialsOAuth` with `oauth2format: undefined`?

(In the future, it would be better if this type was a proper discriminated union, but I'm not sure if this optional here is intentional.)

If this is meant to handle the token case, then it is a bit confusing.

packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal/custom-repository-settings-form-group.tsx
```diff
@@ -0,0 +1,114 @@
...
+export interface Props {
+  onInputChange: (e: React.SyntheticEvent<HTMLInputElement>) => void;
+  isGitRepository: boolean;
+  inputs: {
+    uri?: string;
+    authorName?: string;
+    authorEmail?: string;
+    username?: string;
+    token?: string;
+    oauth2format?: 'github';
```
DMarby: This doesn't seem to be used here, it should only exist on the `github-repository-form-group`, right?

packages/insomnia-app/app/sync/git/utils.ts
```diff
@@ -10,28 +12,46 @@ export const translateSSHtoHTTP = (url: string) => {
...
+      username: credentials.token,
+      password: 'x-oauth-basic',
+    };
+  }
+
+  console.log('[git-event] Using basic auth credentials');
+  return {
+    username: credentials.username,
+    // @ts-expect-error -- TSCONVERSION this needs to be handled better if credentials is undefined or which union type
+    password: credentials.password || credentials.token,
```
johnwchadwick, line 48: The `credentials.token` here doesn't seem to be necessary. As far as I can tell, `credentials.token` can't be set here because the OAuth codepath already eliminates that possibility. Am I misunderstanding?

johnwchadwick (approved):
Overall I approve though I have some comments that are non-critical.

packages/insomnia-app/app/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
```diff
@@ -0,0 +1,360 @@
...
+
+    if (githubToken && !user) {
+      fetchGraphQL<{ data: GitHubUserInfoQueryResult }>({
+        query: GitHubUserInfoQuery,
+        headers: {
+          Authorization: `Bearer ${githubToken}`,
+        },
+        url: GITHUB_GRAPHQL_API_URL,
+      })
+        .then(({ data }) => {
```
DMarby: Do we need to handle the graphql query returning an error here at all? (As opposed to axios throwing because the server returned an http error code, which is handled below)

DMarby (approved):

packages/insomnia-app/app/sync/git/utils.ts
DMarby: This is still needed for the case where a user configures custom credentials (username + token), where we don't fetch the token from GitHub:
image


What is confusing things further is that we never have a case where `password` is not null for a credential as far as I'm aware, as we don't expose a way to set that.
Not sure if we ever did/that's something we need to be backwards compatible with, if not we could likely drop `GitCredentialsBase` entirely and simplify this to just `password: credentials.token`.

packages/insomnia-app/app/sync/git/git-vcs.ts
DMarby: This is pretty confusing indeed, this is used by the `GitRepository` model, and there's already username + token credentials that are stored in the database without an `oauth2format`, as that's the case for when you configure custom credentials as opposed to automatically fetching them using this new feature, which is why this is optional I'd assume.

johnwchadwick: Acknowledged. Renaming this to `GitCredentialsOAuth` seems quite wrong with this in mind, considering it is still used for other kinds of tokens, but it is not critical enough to hold up this review process.

packages/insomnia-app/app/sync/git/utils.ts
johnwchadwick: OK. I think our typings are not correct, then, but I guess they already weren't correct before either.

PR #4430, opened 2022-02-01
Speed up the tests workflow

DMarby:
Speeds up the tests workflow a bit by running platform independent steps on a single platform only. Ubuntu was chosen as it is the fastest one, and should be the most strict (since it has a case sensitive fs).

This PR also removes the inso .dmg build step, which was non-functional on most of our PRs as secrets do not get passed to action runs originating from a fork, and for testing having the inso macOS binary as an artifact (same as on the other platforms) should be sufficient for now.

Lastly, this PR switches to a newer version of the `checkout` action, which should speed up the git checkout as it only checks out the latest commit.

I've opted to not touch the release/publish workflows in this PR (which is also using the older checkout action, amongst other things), but plan on going over those separately in the future.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,12 +50,15 @@ jobs:
         run: npm run bootstrap
 
       - name: Lint
+        if: matrix.os == 'ubuntu-latest'
         run: npm run lint
 
       - name: Lint markdown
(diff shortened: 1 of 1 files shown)


johnwchadwick (approved):

gatzjames (approved):
💯

PR #4439, opened 2022-02-02
Adds tracing with Playwright

gatzjames:
The goal is to have a quicker feedback loop for failed playwright tests.
This PR enables us to get a visual interactive trace for failed tests.

- [x] Use the [testInfo](https://playwright.dev/docs/api/class-testinfo) to check if we should start tracing/store the trace or not.
We now use the playwright.config values to decide if we capture the trace.
- [x] ~~Create links for generated artifacts to download/open in `https://trace.playwright.dev/?trace=${generated_artifact-url}`~~
This is more complicated than originally planned so we can move it as an enhancement to what we currently have.


dimitropoulos:
do we have an example of on such URL?  that, and, could you document somewhere how to make one.  or (better?) maybe we can figure out some way to include it somewhere automatically?  Maybe as an "annotation" here? 
image

filfreire (approved):
Tried this out locally, super awesome stuff ❤️ 🎉 

Screenshot 2022-02-04 at 12 30 39

filfreire:
@gatzjames feel free to re-request review if you add anything related to the WIP points.

I would suggest one alternative, so we don't hammer right away a trace.

We could for now just add a note to the readme (and maybe the smoke-test log in the end) that folks can run the trace viewer locally by downloading the file and running `npx playwright show-trace trace.zip`

Or, we use something like what's mentioned in the docs about [viewing remote traces](https://playwright.dev/docs/trace-viewer#viewing-remote-traces) 

Not sure how we would work out a download link for the build artifact though, as @dimitropoulos mentioned above, if we just use something like:
```
https://github.com/Kong/insomnia/suites/5163984636/artifacts/156371121
```
It won't work right away, it gets resolved into a link like:
```
https://pipelines.actions.githubusercontent.com/<REPLACE>/_apis/pipelines/1/runs/7413/signedartifactscontent?artifactName=windows-latest-smoke-test-screenshots-5801&urlExpires=2022-02-04T12%3A50%3A51.9180508Z&urlSigningMethod=HMACV2&urlSignature=SOMESIGNATURE%3D
```
Which is dependant on our login. Both these don't work on the local traceviewer.


Another alternative is if we use the [github api](https://docs.github.com/rest/reference/actions#download-an-artifact), something like:
```
curl \
  -H "Accept: application/vnd.github.v3+json" \
  https://api.github.com/repos/Kong/insomnia/actions/artifacts/156371121/zip
```
We would need to give it a token that has `actions` scope permissions. We would need to add it to the repo secrets, otherwise we get:
```
{
  "message": "You must have the actions scope to download artifacts.",
  "documentation_url": "https://docs.github.com/rest/reference/actions#download-an-artifact"
}
```

This would generate a url we could use for download, but

> Gets a redirect URL to download an archive for a repository. This URL expires after 1 minute. Look for Location: in the response header to find the URL for the download. The :archive_format must be zip. Anyone with read access to the repository can use this endpoint. 

Only thing close to an already available solution that could work as alternative for us to use / adapt was https://github.com/defcronyke/github-artifact 🤷

packages/insomnia-smoke-test/playwright.config.ts
```diff
@@ -7,6 +7,14 @@ const config: PlaywrightTestConfig = {
     timeout: 120 * 1000,
     reuseExistingServer: !process.env.CI,
   },
+  use: {
+    trace: {
+      mode: 'retain-on-failure',
+      screenshots: true,
+      snapshots: true,
+    },
+  },
```
jackkav, line 16: From what I have tried the use block in config doesn't do anything with electron apps beside pass properties. So this isn't required. But if it's useful to toggle it could be simplified to just true false or "on"

PR #4456, opened 2022-02-08
updates redux dependencies

dimitropoulos:
This PR updates our redux dependencies for the following reasons:

- react-redux v8 is ready, but [waiting](https://github.com/reduxjs/react-redux/releases/tag/v8.0.0-beta.3) for react 18 to launch.  when that time comes, it'd be great to be as close to that release as possible to ease the transition.
- react-redux contains some performance enhancements (mainly, the removal of an [unnecessary selector call on monunt](https://github.com/reduxjs/react-redux/releases/tag/v7.2.5), and some [unnecessary calling](https://github.com/reduxjs/react-redux/releases/tag/v7.2.4) after a dispatch (which is a somewhat common case))
- react-redux [now includes](https://github.com/reduxjs/react-redux/releases/tag/v7.2.3) the types, so we can drop the @types/react-redux dependency on our end
- our redux stack will now be compatable with [react 17](https://github.com/reduxjs/react-redux/releases/tag/v7.2.2)


gatzjames (approved):
👍

PR #4463, opened 2022-02-08
Update TypeScript version.

johnwchadwick:
This makes the TypeScript version match across other codebases, to facilitate better code sharing.


dimitropoulos (approved):
sweet.  tested the app and all looks to be working as intended.



I wasn't sure if ts-node needed to be updated, so also tested the ts-node scripts in `smoke-test` and `config` and they worked well, so looks like we're all good.  The same goes for ts-jest, but clearly that's working well too, so, great!

PR #4464, opened 2022-02-08
changelog generator script

dimitropoulos:
creates changelog generator script.  see the newly committed [readme](https://github.com/Kong/insomnia/pull/4464/files#diff-6c561485fa1c567441bb832e2452b64fd91eeab2084ea2fa3c9f297dd7727d6c) in the package for more context on what it does (and why).  

Going against usual tradition: I suggest you do not read commit by commit because it's almost entirely just new code.  I left the commits intact from my history when working on it just in case I needed to timetravel for a bit to investigate a decision or change.


filfreire (commented):
I say this with peace and love: when looking at the README, my first thought was "this is too much text" 😄 

I was expecting the README to just tell me, straightforward:

- How to build (and setup GH token)
- How to run (+ example usage)
- FAQ

I think most of the `General Principles` content and the `Other approaches` content we could keep as sort of a blog entry on the [repo's Wiki](https://github.com/Kong/insomnia/wiki)

@dimitropoulos what do you think?

dimitropoulos:
that's great feedback.  I think I'll make a "TLDR;" section.  lemme take a swing at it. :)

dimitropoulos:
I wanted to say a little more about why this was extracted into its own package.  Originally (if you check the commit history) it was just in `insomnia-common` as a script.  [48daa06](https://github.com/Kong/insomnia/pull/4464/commits/48daa06966fe03060d7549c0912b3192f1a2ff93) moved it to it's own package.  The reason was because I want an end user to be able to use `npx` to run it, i.e. `npx insomnia-changelog --base blah --head blah`.  This UX is not possible without it being in its own package.  I'll add that once I did this (check out the commit to see what I'm talking about) I found a lot of dependencies (esrunner, for example) and soft-"hacks" (like a bespoke tsconfig) that were in place to make it work with the "script in a folder somewhere" approach.  Putting it in its own package made the code itself a lot more obvious and typical.

This is a pretty general purpose script.  Could this eventually be moved somewhere else (i.e. `github.com/Kong/changelog-generator` or something) where other people can have more ownership over it without having to know or use anything related to insomnia?  Yes.  Yes it could.  But for now, keeping it in the monorepo means we're much more easily able to manage it and keep it to our code standards as they improve.

dimitropoulos:
pulled out into https://github.com/Kong/changelog-generator and thus this PR should be closed

PR #4468, opened 2022-02-08
all dependabot PRs as of 20220802

dimitropoulos:
at this point, we're gonna have to start using timestamps because we're having multiple of these per day, as we update.


filfreire (approved):
Ran a few tests locally, seems to be ok 👍 

@dimitropoulos as for the fact we get these a Lot, let's see if we get more the next few days/weeks, and check with the team what would be a hassle-free way of dealing with any dependabot PRs.

PR #4474, opened 2022-02-09
Recurring releases of insomnia app (INS-1429)

filfreire:
## What is changed

This PR adds a new workflow that'll pre-bake artifacts when there's changes on PRs or on develop.

These artifacts can be used for all our major platforms to some exploratory/deep testing on a work in progress PR or latest develop without the need to build those locally.

Also changed:
- Update default electron-builder targets
- Allow setting a build ref. Changes the default version to `0.0.1-dev+unknown`, so that build/packaging can be easily ran locally without setting any environment variables.

## How to test

You should now be able to just run `npm run app-package` without specifying git SHAs or Github references environment variables.

You should be able to run `smoke-test` related package command without any issues locally.

You can also mimic what happens on the new Github Actions workflow, locally, running something like:
```shell
BUILD_REF='1234' npm run app-package
```

This `BUILD_REF` env variable is for now intended to be used only in Github actions/ CI. 

It serves to add info to the version field on `About Insomnia`  menu, in order to help folks be sure of the build they are testing. 

In the case of Github Actions artifacts, it can for example be git's short SHA, like `BUILD_REF='$(git rev-parse --short HEAD)'`:

Screenshot 2022-02-09 at 16 16 37

cc @wongstein @wdawson - after this gets merged, it should enable us to easily download builds for a PR or latest develop if we want to test out something - effectively doing what we initially planned with nightlies.


.github/workflows/release-recurring.yml
```diff
@@ -0,0 +1,49 @@
...
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+         - os: "macos-latest"
+           build-targets: "zip"
+         - os: "windows-latest"
+           build-targets: "portable"
+         - os: "ubuntu-latest"
+           build-targets: ".tar.gz"
```
DMarby:
```suggestion
           build-targets: "tar.gz"
```

DMarby (approved):
🎉

PR #4503, opened 2022-02-16
fix(brew): de-duplicate --open-pr flag in GHA

curiositycasualty:
As it turns out, `Homebrew/actions/bump-formulae` [already supplies this flag](https://github.com/Homebrew/actions/blob/ead7c26a38d02a1a5e5f82958738769f4a526a7a/bump-formulae/action.yml#L17):


DMarby (approved):
Nice, thank you! 👍 

https://github.com/Kong/insomnia/runs/5209051939?check_suite_focus=true#step:6:6 It seems like the `inso-beta` cask is throwing an error ever since this workflow was initially merged, not sure if the tap for it needs to be explicitly given somehow, or if it's still working as expected despite the error?

curiositycasualty:
You're right. Let me get that looked at.

On a related note: I've opened:
  - https://github.com/Homebrew/homebrew-cask-versions/pull/13338
  - https://github.com/Homebrew/homebrew-cask/pull/119280
  - https://github.com/Homebrew/homebrew-cask/pull/119281

To try and address the issue where a release is so old that it falls onto "page 2" of the releases page and thus can't be used for `livecheck`-- which is what `brew bump` and this workflow rely on.

PR #4515, opened 2022-02-22
(Partial) removal of WrapperProps (part 3)

dimitropoulos:
This PR relates to (but does not close) the wrapper props ticket, INS-1058. This continues the work started in #4431 and #4510.

This PR focuses on wrapper props passed to modals.  Unfortunately, since most of the modals are still class components (i.e. not compatible with hooks and therefore direct usage of `useSelector`), some of them required continuing our "UnconnectedXyz" pattern.  I don't love this pattern, but at this point at least it's highly regular throughout the app.  This job to remove WrapperProps is already so highly complicated that I've concluded it's best not to expand scope at all.

### DIFF
--- a/packages/insomnia-app/app/ui/components/modals/cookie-modify-modal.tsx
+++ b/packages/insomnia-app/app/ui/components/modals/cookie-modify-modal.tsx
@@ -46,16 +46,14 @@ export class UnconnectedCookieModifyModal extends PureComponent<Props, State> {
     // Dunno why this is sent as an array
     cookie = cookie[0] || cookie;
     const { activeCookeJar } = this.props;
-    const oldCookie = activeCookeJar?.cookies.find(c => c.id === cookie.id);
+    const prevCookie = activeCookeJar?.cookies.find(c => c.id === cookie.id);
 
-    if (!oldCookie) {
+    if (!prevCookie) {
       // Cookie not found in jar
       return;
     }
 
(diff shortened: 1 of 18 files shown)


packages/insomnia-app/app/ui/components/modals/cookie-modify-modal.tsx
```diff
@@ -43,8 +45,8 @@ export class CookieModifyModal extends PureComponent<Props, State> {
   async show(cookie: Cookie) {
     // Dunno why this is sent as an array
     cookie = cookie[0] || cookie;
-    const { cookieJar } = this.props;
-    const oldCookie = cookieJar.cookies.find(c => c.id === cookie.id);
```
gatzjames, line 47: old is a bit confusing perhaps we could name these prevCookie/currentCookie?

gatzjames (approved):
🚀 Works great!

packages/insomnia-app/app/ui/components/modals/cookie-modify-modal.tsx
dimitropoulos: I hear you.  It was already using that terminology (which is why I didn't change it) but now's as good a time as any to use more standard terminology like you suggest: f1d763c57a5d49845734dea7b2e15d5b490bbf8f.

PR #4520, opened 2022-02-24
strengthens types of key command usage in keyboard-keys

akivajgordon:
(see history on this GitHub discussions for context as to the original motivations for this PR)

This PR adds stronger typing for keyboard keys and improves the strictness for accessing these helpers.


dimitropoulos (approved):

dimitropoulos (changes requested):
on second thought, maybe it'd be best to find some other package that does this job (validation) so we don't maintain it at all, rather than furthering our own abstraction for something that already exists (KeyboardEvent.code).

dimitropoulos (approved):

dimitropoulos (approved):
ok, after looking at it again I found a simpler way (that didn't involve hardcoded duplication of the keys or using `any`) to solve the same problem.  let's wait for @jackkav @gatzjames or @marckong to take a look before merging.

filfreire (approved):
Tested it locally, seems to be working ok 👍

marckong (approved):

PR #4539, opened 2022-03-02
updates usages of jest timers to prepare for jest 27

dimitropoulos:
Here's a quote from the [Jest 27 changelog](https://jestjs.io/blog/2021/05/25/jest-27):

> Another default that we are changing affects Fake Timers aka [Timer Mocks](https://jestjs.io/docs/timer-mocks). We introduced an opt-in "modern" implementation of Fake Timers in Jest 26 accessed transparently through the same API, but with much more comprehensive mocking, such as for Date and queueMicrotask.
This modern fake timers implementation will now be the default. If you are among the unlucky few who are affected by the subtle implementation differences too heavily to migrate, you can get back the old implementation using jest.useFakeTimers("legacy") or, if you are enabling fake timers globally via [configuration](https://jestjs.io/docs/configuration#timers-string), "timers": "legacy".

If we try to upgrade to Jest 27 we hit this problem in two files.  When it errors, it looks like this:

This PR simply updates the two places where we were depending on this older style to use the new [Timer Mocks](https://jestjs.io/docs/timer-mocks).

That way, we aren't blocked by this when we update Jest.

### DIFF
--- a/packages/insomnia-app/app/common/__tests__/local-storage.test.ts
+++ b/packages/insomnia-app/app/common/__tests__/local-storage.test.ts
@@ -4,12 +4,12 @@ import path from 'path';
 import { globalBeforeEach } from '../../__jest__/before-each';
 import LocalStorage from '../../main/local-storage';
 
+jest.spyOn(global, 'setTimeout');
+
 describe('LocalStorage()', () => {
   beforeEach(async () => {
     await globalBeforeEach();
     jest.useFakeTimers();
-    // There has to be a better way to reset this...
-    setTimeout.mock.calls = [];
   });
 
   afterEach(() => {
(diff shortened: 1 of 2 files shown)


packages/insomnia-app/app/common/__tests__/local-storage.test.ts
```diff
@@ -4,12 +4,12 @@ import path from 'path';
...
 import LocalStorage from '../../main/local-storage';
 
+jest.spyOn(global, 'setTimeout');
+
 describe('LocalStorage()', () => {
   beforeEach(async () => {
     await globalBeforeEach();
     jest.useFakeTimers();
-    // There has to be a better way to reset this...
-    setTimeout.mock.calls = [];
```
dimitropoulos, line 12: note that the `afterEach` calls `jest.clearAllTimers();` which is, ironically, exactly the "better way to reset" it.

dimitropoulos:
note that we can't just switch to the legacy, not just for moral reasons (i.e. development overhead) but also because other things we use will require it, e.g. the next version of testing-library: https://github.com/testing-library/dom-testing-library/releases/tag/v8.0.0

DMarby (approved):

PR #4559, opened 2022-03-08
hotfix: invalid OpenAPI spec versions causing the app to crash

dimitropoulos:
OpenAPI specs do not support the `number` datatype for `$.info.version`.  Only `string` is supported.  
Screenshot_20220308_152427

In Insomnia, if you change this field to a number, you will see a lint error.
Screenshot_20220308_150027

Unfortunately, if you make this change and then flip to the dashboard, the following error appears:
Screenshot_20220308_150050

This leaves your application in a totally broken state and there is literally nothing you can do in the app, or do about it to fix it.
Screenshot_20220308_150138

This change was introduced in https://github.com/Kong/insomnia/pull/4319 and would be preventable by spec validation (i.e. AJV) and/or functional TypeScript typings.

Our code is correct in the sense that it follows the spec.  The problem is there's a failure-mode for an error case that is quite breaking.

Fixes INS-1477


dimitropoulos:
update: I figured out a temporary workaround while we get a patch released.  see it here: https://youtu.be/D6yT20GPjZI

bstncartwright (approved):

filfreire (approved):

gatzjames (approved):
💯 Tested and works great!

PR #4574, opened 2022-03-10
Changelog generation for the automated releases

filfreire:
## Changes

This PR adds a new `Generate Changelog` workflow ( `changelog.yml`) that folks can run to add/update an auto-generated changelog comment in the automated release pull-requests.

The workflow runs on the background https://github.com/Kong/changelog-generator for the generation of the text.

We also automatically show pull-requests not covered by the release [that are missing a `changelog(Something): ...` entry](https://github.com/Kong/changelog-generator#double-checking-commits-without-changelogs).

### How to run

This new workflow is triggered:
- Automatically on each `release/<something>` PR opened/synchronised;
- Manually via the UI:
Screenshot 2022-03-22 at 12 48 01

- Manually via the [Github CLI](https://cli.github.com/), example:
```
gh workflow run changelog.yml -f version=<version> --repo Kong/insomnia
```

### End result

> Note: we don't create more than one comment, we always update the existing one.

Example of how the comment looks like:

Screenshot 2022-03-22 at 12 51 18

### Possible problems

- While testing I got rate-limited quite a bit, regardless of using a personal access token or `GITHUB_TOKEN`, we easily break a limit, 'x-ratelimit-limit': '30', if the diff between latest stable release and the version we are trying to generate the diff log is too spaced out.


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,104 @@
+name: Generate Changelog
+
+on:
+  # TODO - edit workflow to support from push
+  workflow_dispatch:
+    inputs:
+      version:
+        required: true
+        description: Release version (e.g. 2022.1.0 or 2022.1.0-beta.0)
+      onlyShowMissing:
```
DMarby: Instead of this being an option, I think it'd be better to always run both and show both the outputs in the comment?

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,104 @@
+name: Generate Changelog
+
+on:
+  # TODO - edit workflow to support from push
```
DMarby: I think rather then push, having this trigger automatically on PR open/update for release branches would be good as well, that together with `workflow_dispatch` for triggering updates should be a good start?

Adding something like this should do the trick:
```
  pull_request:
    types:
      - opened
      - synchronize
```

And on the job itself:
```
    if: "startsWith(github.head_ref, 'release/')"
```

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          npx @kong/changelog-generator generate \
+            --owner ${OWNER} \
+            --repo insomnia \
+            --base ${{ steps.latest_release.outputs.release }} \
+            --head ${HEAD} \
+            --releaseName "${RELEASE_NAME}" ${SHOW_MISSING} > ./temp_changelog.out
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OWNER: Kong
```
DMarby:
```suggestion
          OWNER: ${{ github.repository_owner }}
```

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,104 @@
...
+env:
+  RELEASE_BRANCH: release/${{ github.event.inputs.version }}
+  RELEASE_VERSION: ${{ github.event.inputs.version }}
+  SHOW_MISSING: ${{ github.event.inputs.onlyShowMissing == true && '--onlyShowMissing' || '' }}
+
+jobs:
+  generate-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout branch
```
DMarby: We probably don't need to check out the branch at all?
We can skip passing `node-version-file` to the setup-node action below, it doesn't really matter for running the changelog generator.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,104 @@
...
+          branch: ${{ env.RELEASE_BRANCH }}
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Get latest release for Changelog
+        id: latest_release
+        uses: pozetroninc/github-action-get-latest-release@master
```
DMarby: I think we need to find an action that can filter releases when fetching them, so that we can exclude alphas and inso releases.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,104 @@
...
+        type: boolean
+        required: false
+        description: Enable --onlyShowMissing flag of Kong/changelog-generator
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  RELEASE_BRANCH: release/${{ github.event.inputs.version }}
```
DMarby: To be able to support both `workflow_dispatch` and `pull_request`, we could make this fallback to `github.head_ref` I think?
For `RELEASE_VERSION` we'd need to add a separate step like we did in `build-start.yml`:
https://github.com/Kong/insomnia/blob/17e2aefd3e79cdf989cc080b94ed0dd20b2c9f17/.github/workflows/release-build.yml#L119-L126

.github/workflows/changelog.yml
filfreire: Done, thank you 👍

.github/workflows/changelog.yml
filfreire: Done, thank you 👍

.github/workflows/changelog.yml
filfreire: Done, thank you 👍

.github/workflows/changelog.yml
filfreire: Done, check the new screenshot to see how it looks like now
Screenshot 2022-03-22 at 12 51 18

.github/workflows/changelog.yml
filfreire: Done, thank you 👍

.github/workflows/changelog.yml
filfreire: Good point. 
Alphas if I'm not mistaken, when I was trying this out on my fork, they are excluded. I'll look into excluding the inso releases, as the way it is right now it is always picking up the latest `lib@x.y.z` one, instead of the `core@...` one.

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,126 @@
+name: Generate Changelog
+
+on:
+  # TODO - edit workflow to support from push
```
johnwchadwick: Is this TODO now stale?

.github/workflows/changelog.yml
johnwchadwick: Could always fork the action if needed, I'm sure. Maybe upstream would even accept a branch regex filter?

johnwchadwick (approved):

.github/workflows/changelog.yml
DMarby: https://github.com/marketplace/actions/git-get-release-action I think we can use this action instead, it has a release name regex filter

filfreire: @DMarby @johnwchadwick I pushed the change to use https://github.com/marketplace/actions/git-get-release-action instead.

Now we filter for:
```
          draft: false
          prerelease: false
          releaseNameRegEx: Insomnia .*
```

which should solve the problem and always point to latest stable release.

.github/workflows/changelog.yml
filfreire: Yeah, it is no longer applicable 👍

DMarby (approved):

PR #4577, opened 2022-03-10
updates `graphql` dependency

dimitropoulos:
The new version of NPM showed me that we are, indeed, using an unsupported graphql package version:
Screenshot_20220309_091548

closes ins-1479


gatzjames (approved):
Tested and works great!
Also double checked the deprecationReason prop and it seems that this is the way to check if a field is deprecated now!
🚀

PR #4578, opened 2022-03-10
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.


dimitropoulos (approved):
Screenshot_20220310_135113

looks good to me!

just curious: how'd you notice this?

CommanderRoot:
My IDE showed me a deprecation warning when working on a different repo. Since then everytime I use a software I check if they also might still use the deprecated function and if so I send a PR to fix it.

dimitropoulos:
cheers.  that's very kind of you.  thank you.  we appreciate it.

and hey, if there's anything else you need while you explore, don't hesitate to drop us a message!

jackkav (approved):

PR #4579, opened 2022-03-10
markdown generation now sanitizes with dompurify

dimitropoulos:
closes INS-1365

Presently, we get these errors:

Here's what the docs for `marked` say to do:
badcbe58-c758-472f-a28a-06ac4561ba08

And, indeed, this PR does exactly that.  `insane` seems to be left for dead, and `sanitize-html` has half as many weekly downloads as `dompurify`, so I used `dompurify` like they suggest.


filfreire (approved):
LGTM 👍 


Screenshot 2022-03-10 at 18 15 42

The warnings are also gone.


> ℹ️  Sidenote: I noticed something I didn't know we support, when we are editing markdown, we can also do `Ctrl+space` / `Cmd+space` and we can define timestamps or hashes in the markdown doc itself. Seems to be still working after the change:
> 
>  Screenshot 2022-03-10 at 18 15 24

PR #4580, opened 2022-03-10
chore(dep): update aws4

dimitropoulos:
This dependency is out of date (and the types are worse, at 5.3 years out of date), and I'm going through some of our dependencies in reverse order.

My plan was to write tests for this _before_ updating the dependency in order to make sure that the version bump doesn't affect anything: but great news!: it was [already tested](https://github.com/Kong/insomnia/blob/d3b4322bf2e2600e396a47e4cca903ead5fdef7c/packages/insomnia-app/app/network/__tests__/network.test.ts#L752).


packages/insomnia-app/app/network/network.ts
```diff
@@ -984,33 +984,30 @@ export function _getAwsAuthHeaders(
...
-        'content-type': contentTypeHeader.value,
-      }
-      : {},
+    ...(parsedUrl.path ? { path: parsedUrl.path } : {}),
+    headers: contentTypeHeader ? { 'content-type': contentTypeHeader.value } : {},
   };
   const signature = aws4.sign(awsSignOptions, credentials);
+  if (!signature.headers) {
+    return [];
+  }
```
jackkav, line 1005: How are error cases handled in this code? Will this new case cause an error to be ignored? I'm not familiar with it, but seems like it could use some illumination.

dimitropoulos: TLDR; we're just using this for signing (i.e. pure-functional input/output) and not making any requests as soon as I am aware.

here's the codepath:
image
image
image
image

jackkav: ok so its just a transform and doesn't handle errors or exceptions, cool

jackkav (approved):

PR #4582, opened 2022-03-10
chore(deps): mime-types update (and bug fix)

dimitropoulos:
We were ignoring a genuine (typescript) error, before.

https://github.com/jshttp/mime-types/blob/master/test/test.js#L117 shows that when passed something it doesn't understand, it will return false.  So, we need to handle that separately.  Aside from fixing the imports to not be unnecessary global imports, nothing else has changed other than the (non breaking) upgrade.


packages/insomnia-app/app/ui/containers/app.tsx
```diff
@@ -729,8 +729,8 @@ class App extends PureComponent<AppProps, State> {
         responsePatch.statusCode >= 200 &&
         responsePatch.statusCode < 300
       ) {
-        // @ts-expect-error -- TSCONVERSION contentType can be undefined
-        const extension = mime.extension(responsePatch.contentType) || 'unknown';
+        const fallback = 'unknown';
+        const extension = responsePatch.contentType ? mimeExtension(responsePatch.contentType) || fallback : fallback;
```
jackkav: x ? y || z : z
seems unintentional here

dimitropoulos: ugg.  I spent forever trying to decide on the best way to do this.  The problem is that you actually can't pass null or undefined to `y`, which is missing in your example.  I read the source code.

We _cannot_ pass `fallback` to `mimeExtension`, but we want the fallback to exist one way or the other.

another way to write this would be:
```ts
const fallback = 'unknown';
let extension = unknown;
if (responsePatch.contentType) {
  const calculatedExtension = mimeExtension(responsePatch.contentType);
  if (calculatedExtension) {
    extension = calculatedExtension;
  }
}
```

but that seems like _a lot_ of lifting

jackkav:
```ts
const sanitizedContentType= responsePatch.contentType && mimeExtension(responsePatch.contentType);
const extension = sanitizedContentType ||  fallback;
```
ta da 🎉

dimitropoulos: da5e9ad923cf9bb520ee9db55122774f0ce1a798

filfreire (approved):
Tested this locally seems ok. Didn't seem to break anything specially with multipart requests.

Screenshot 2022-03-18 at 17 41 15
Screenshot 2022-03-18 at 17 41 00
Screenshot 2022-03-18 at 17 38 49

PR #4585, opened 2022-03-10
Remove package-lock for insomnia-url

DMarby:
Not entirely sure if this is the correct approach, but this `package-lock.json` file was outdated and not valid json (missing comma after `lockFileVersion`), and after resolving that or removing it entirely, `lerna bootstrap` does not recreate it as this package does not have any dependencies in it's `package.json`.

Running `npm install` in the directory does create a blank `package-lock.json`, but that does not get correctly updated on lerna version bumps since `bootstrap` doesn't modify it, so that's probably not preferable?


gatzjames (approved):
Tested and works fine!
The dependencies in the lockfile seem outdated and perhaps used for previous way of building as far as i can telll so wasn't expecting any surprises.
It makes sense to not have a package-lock here,
since we don't install any dependency inside this package.

PR #4586, opened 2022-03-10
Fix response error colour

DMarby:


dimitropoulos (approved):
Here's `Request Error`, `100`, `200`, `300`, `400`, `500`:

| BEFORE | AFTER |
| - | - |
| Screenshot_20220310_164128 | Screenshot_20220310_163954

PR #4599, opened 2022-03-16
Simplify the test workflow

DMarby:
To make it more consistent with `release-build.yml` and avoid the extra job.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,22 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  get_version:
-    name: Get version
-    runs-on: ubuntu-latest
-    outputs:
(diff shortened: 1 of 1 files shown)


filfreire (approved):

.github/workflows/test.yml
```diff
@@ -15,22 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  get_version:
-    name: Get version
-    runs-on: ubuntu-latest
-    outputs:
-      inso-version: ${{ steps.get-package-version.outputs.current-version }}
-    steps:
-      - name: Checkout branch
```
filfreire, line 24: @DMarby In this case don't we still need the `actions/checkout@v3`?

DMarby: Nope, since that's done by the OS job below anyway on line 25

PR #4603, opened 2022-03-16
remove zprint-clj and EDN beautifying

dimitropoulos:
### What's the big picture.

We want to move all prettifying to https://www.npmjs.com/package/prettier.  That will mean that anything that can be prettified using prettier (or a prettier plugin) (i.e. GraphQL, JSON, YAML, etc.) will be supported by Insomnia.  In general, if we have outreach from users to add EDN prettifying back: we can think about it from that direction (i.e. a prettier plugin for EDN).

### zprint-clj

We use the [`zprint-clj`](https://www.npmjs.com/package/zprint-clj) package for only one thing: prettifying EDN.  [Here](https://learnxinyminutes.com/docs/edn/)'s a little thingy about what EDN is.  It's not... shall we say... common.  That's not, in itself, a reason _not_ to support prettifying it, please read on:

When upgrading this dependency I could not find source code for this npm, nor could I find a changelog.  Still.. it appeared to be a straightforward thing to update.  It takes EDN (https://github.com/edn-format/edn) and prettifies it.

But then I tried upgrading it and I got:
Screenshot_20220316_160045

I thought I could solve it quickly, but I couldn't. There's a shebang at the top of the new release, and there's a few different ways to solve it (babel plugin, webpack loader, etc.) and I can't decide which is best, and none of them are awesome. I did, as I said, test it, but by just deleting the shebang manually.

| BEFORE | AFTER |
| - | - |
| Screenshot_20220317_111958 | Screenshot_20220317_111652 |

### Why remove prettifying for EDN

This was introduced in https://github.com/Kong/insomnia/pull/1176.  If we see anything from users suggesting a way to keep using it, we can do so.

Note that this doesn't have any impact on EDN syntax highlighting in the code editor.

changelog(Removal): Removed EDN prettifying ahead of a move to prettier for all prettifying

### DIFF
--- a/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
+++ b/packages/insomnia-app/app/ui/components/codemirror/code-editor.tsx
@@ -14,7 +14,6 @@ import React, { Component, CSSProperties, forwardRef, ForwardRefRenderFunction, 
 import { useSelector } from 'react-redux';
 import { unreachable } from 'ts-assert-unreachable';
 import vkBeautify from 'vkbeautify';
-import zprint from 'zprint-clj';
 
 import {
   AUTOBIND_CFG,
@@ -725,14 +724,6 @@ export class UnconnectedCodeEditor extends Component<CodeEditorProps, State> {
     }
   }
 
-  static _prettifyEDN(code: string) {
-    try {
-      return zprint(code, null);
-    } catch (e) {
-      return code;
-    }
-  }
-
   _prettifyXML(code: string) {
     if (this.props.updateFilter && this.state.filter) {
       try {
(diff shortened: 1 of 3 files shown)


dimitropoulos:
update: I thought I could solve it quickly, but I couldn't.  There's a shebang at the top of the new release, and there's a few different ways to solve it (babel plugin, webpack loader, etc.) and I can't decide which is best.  I did, as I said, test it, but by just deleting the shebang manually.
Screenshot_20220316_163329

dimitropoulos:
update: I thought I could solve it quickly, but I couldn't. There's a shebang at the top of the new release, and there's a few different ways to solve it (babel plugin, webpack loader, etc.) and I can't decide which is best. I did, as I said, test it, but by just deleting the shebang manually.
Screenshot_20220316_160045

wdawson:
Looks like this is the least frequently used body type in Insomnia (less than 10 times per day from what we can see). So I'm good with removing for now.

jackkav (approved):
👏

Issue #4572, opened 2022-03-10
SecuritySchemes does not inject issuer and scopes into openIdConnect Plugin with OpenAPI-2-Kong


dmortim:
### Expected Behavior

When using securitySchemes the scopes and the openIdConnectUrl should be injected into the Kong OpenIdConnect plugin configuration. 

### Actual Behavior

When using security schemas with x-kong-security element, the scopes and issuer are not applied to the routes and it results in a schema violation when applying the configuration.  

```
while processing event: {Update} plugin openid-connect for route XXXXX failed: HTTP status 400 (message: "schema violation (config.issuer: required field missing)")
```

### Reproduction Steps

``` yaml
components:
  securitySchemes:
    openId:
      type: openIdConnect
      openIdConnectUrl: https://idp-endpoint.example.com/.well-kown # This should be injected into the kong-plugin configuration as the issuer
      x-kong-security-openid-connect:
        config:
          auth_methods:
            - bearer
        enabled: true
        protocols:
        - http
        - https

paths:
  /example/path:
    get:
      security:
        - openId:
            - example-scope # This should be injected into the kong-plugin configuration as the scopes required
``` 

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Additional Information

It seems this occurs because of this line. The plugin configuration is replacing the security configuration. 

https://github.com/Kong/insomnia/blob/develop/packages/openapi-2-kong/src/declarative-config/security-plugins.ts#L114

### Insomnia Version

2.4.1

### What operating system are you using?

Other Linux

### Operating System Version

redhat

### Installation method

binary

dmortim:
Hi I think I have identified what is causing this issue and have a potential solution:

https://github.com/Kong/insomnia/blob/develop/packages/openapi-2-kong/src/declarative-config/security-plugins.ts#L114

Instead of 
``` javascript
plugin.config = kongSecurity.config
``` 

Use something like this:
``` javascript
plugin.config = Object.assign(plugin.config, kongSecurity.config)
plugin.enabled = kongSecurity.enabled
```

This way we can specify other configurations from the openid-connect plugin like auth_methods, extra_jwks_uris etc

filfreire:
[...]

PR #4607, opened 2022-03-18
Concating x-kong-security-openid-connect config with security schemas…

dmortim:
Related to #4572

This will enable us to set more configuration properties that is available from the OIDC plugin in Kong. Currently, the OAS3 security schema object for openIdconnect configuration is replaced by the x-kong-security-openid-connect config, which removes scopes_required and the issuer that are set by the security property (on operations) and the openIdConnectUrl respectively. 

The documentation states the following, which does not seem to be happening currently [Security Plugins Docs](https://www.npmjs.com/package/openapi-2-kong#security-plugins)
```
The additional properties that Kong supports on its plugins can be 
configured by using custom extensions. The custom extensions are
x-kong-security-<plugin-name>.
``` 
 
Note: I don't write javascript / typescript so there may be a better way of achieving this for all x-kong-security extension configs.

Changelog(OpenAPI-2-Kong): Fixed an issue where we were not concatenating `x-kong-security-openid-connect` config with security schemas when generating declarative Kong configs


filfreire:
@dmortim thank you for contributing this!

Me and @dimitropoulos paired reviewed this in our latest [Insomnia live stream](https://www.youtube.com/watch?v=-5GH2SAgvS4).

We added a small improvement, and a unit test.

filfreire (approved):

johnwchadwick (approved):

PR #4615, opened 2022-03-21
chore(deps): updates redux-thunk

dimitropoulos:
This probably should have been done back in February in https://github.com/Kong/insomnia/pull/4456

Here's the release notes: https://github.com/reduxjs/redux-thunk/releases

Basically they updated a lot of TypeScript stuff and rewrote the thing in TypeScript.  While there's quite a lot of our typings for redux-thunk that I'd like to improve (we ignore some genuine errors, today) this PR isn't the time or place so I didn't touch it.

The null command thing (the only actual code change in this PR) may not be new, but it seems like we shouldn't be console.logging there.


johnwchadwick (approved):
LGTM, tested on Linux.

PR #4620, opened 2022-03-21
chore(deps): reselect

dimitropoulos:
QUITE a lot has changed with reselect between 4.0.0 and 4.1.5, believe it or not! https://github.com/reduxjs/reselect/releases but the good news is, it's almost exclusively internal improvements for TypeScript. 

We did have to make some upgrades now that the inferencing improved, but nothing more than a few exports for previously unexported interfaces.  Here's an example:
Screenshot_20220321_150342


johnwchadwick (approved):
LGTM, tested on Linux.

PR #4639, opened 2022-03-25
fix event propagation issue affect keydown events

dimitropoulos:
We noticed that when selecting the url bar and hitting Ctrl+Enter, the request send hotkey was not firing.  All the others still worked (e.g. F5, Ctrl+R).

We bisected down to the commit that upgraded React, and found the cause to be the (purportedly very rare) issue described here: https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues

changelog(Fixes): Fixed an issue where some keyboard shortcuts to send requests were not working


wdawson:
I haven't done enough investigation but thought I'd chime in after a cursory read. Just want to make sure I'm understanding what's going on here.

Based on the [react changelog link](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues) that you provided above, it looks like event propagation now works more like the DOM with react 17. It also seems like this change would "reveal bugs" in our event code. But in this case, it seems like we want the event listeners to happen during the capture phase (I don't know exactly what that is), which will make it so that we don't "miss events". 

I guess my question is: Are we confident that this is the right way to solve the problem? Or is this exposing a bug (or bugs)?

filfreire (approved):

dimitropoulos:
[THIS](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/codemirror/one-line-editor.tsx#L250) line is the reason why enter key was being stopped, but not others.  I couldn't find it in the initial search because we're using the literal number, 13, not `.enter` like (almost) everywhere else.

wdawson:
> [THIS](https://github.com/Kong/insomnia/blob/develop/packages/insomnia-app/app/ui/components/codemirror/one-line-editor.tsx#L250) line is the reason why enter key was being stopped, but not others. I couldn't find it in the initial search because we're using the literal number, 13, not .enter like (almost) everywhere else.

That looks ... ugly XD. Makes me wonder if we would need this PR if we did the submit in a different way. But I trust your fix.

gatzjames:
Before react 17 if we prevented the event from propagating it would still fire these event listeners.
Now it correctly prevents the event from propagating (which we do in the one-line-editor) which means we need to explicitly capture the event if we want to handle it like that.

Went through all the keyboard events and double checked that they work as expected with @dimitropoulos so we're confident it won't introduce new bugs!

PR #4643, opened 2022-03-25
Upgrades electron to 18.1.0

dimitropoulos:
Figured it can't hurt to give this a swing.  As you might have expected, we're blocked on

```
[app] Error: The module '/home/dimitri/src/github.com/Kong/insomnia/packages/insomnia-app/node_modules/@getinsomnia/node-libcurl/lib/binding/node_libcurl.node'
[app] was compiled against a different Node.js version using
[app] NODE_MODULE_VERSION 101. This version of Node.js requires
[app] NODE_MODULE_VERSION 103. Please try re-compiling or re-installing
```

but, at least it's a start.

changelog(Improvements): Upgraded to Electron 18.1.0 (which contains Chromium 100.0.4896.127)


filfreire (commented):
@dimitropoulos should we update this to [18.0.1](https://www.electronjs.org/releases/stable#18.0.1)?

jackkav (approved):
Groovy! :man_dancing:

dimitropoulos (commented):
Screenshot_20220427_111724

seems to work and I didn't see anything in the 18.x changelogs that affects us

also tested the oauth flow (to make sure a separate window comes up fine) and that seemed to be good.

PR #4650, opened 2022-03-28
Remove flow (and webpack) from portal plugin, replace with TS and ESBuild

dimitropoulos:
Please read commit by commit to see what happened in this PR (and why).


packages/insomnia-app/app/plugins/context/app.tsx
```diff
@@ -215,6 +216,7 @@ export function init(renderPurpose: RenderPurpose = RENDER_PURPOSE_GENERAL): {
     __private: {
       axios,
       analytics,
+      insomniaComponents: globalThis.document ? require('insomnia-components') : {},
```
gatzjames, line 219: Great find! I saw above that we use the [canShowDialogs](https://github.com/Kong/insomnia/pull/4650/files#diff-402269e89f59227a418f61eb5bcee5a8898091e26026955a394e401518a87bdbR74) variable to check if dialogs can be rendered.
Should we rely on it here too?

dimitropoulos: yikes.  that looks complicated.  I'd have to look more into what those are and what the overlap is.  I think I'd rather keep this very specific to what it does for now rather than risk breaking the build again in a tough-to-debug manner.

jackkav (approved):
Other notable changes outside PR title
- plugin build script convention setting, source, types, concurrently
- insomnia-components are injected rather than npm installed into package 
- typed plugin api convention setting

Curious what the consequences of typing and injection will be for other plugins, happy for discussion to happen after merging this PR though.

Looks great! and -10k lockfiles!

tomanagle (approved):

PR #4657, opened 2022-03-30
Reproduce and Fix #4432

### LINKED ISSUE #4432: Different behaviour between insomnia-inso from npm and binary release
### Expected Behavior
Generate the same declarative config.

### Actual Behavior
When running `inso generate` the two inso versions create different declarative configs.

### Reproduction Steps

Example OpenAPI spec:

```yaml
openapi: 3.0.0
info:
  version: 1.0.0
  title: Sample-API
servers:
  - url: https://example.com
paths:
  /list:
    get:
      description: Returns a list of stuff
      responses:
        '200':
          description: Successful response
```
 Run `inso generate config api-docs.yml -o kong-api-docs.yml` with both inso cli types and compare the result.

- [x]  I have searched the issue tracker for this problem.

### Additional Information
When using hyphens in `info.title` (here "Sample-API") insomnia-inso from npm keeps the title as service name and the binary release converts every hyphen to underscore. 

### Insomnia cli version
2.4.1

Linux in Kubernetes

### Installation method
Downloaded using curl and installed using npm

The version from npm is working

filfreire:
Changelog(OpenAPI-2-Kong): Fixed an issue where Kong declarative config generation had different behaviour for name delimiters between Inso CLI's npm package and binary release

## The problem

TL;DR: We have an issue where generating a kong declarative config, we replace dashes in original title of openapi specs with underscores. It's reproducible on tests, on `inso-cli` installed via github releases, and also on Insomnia plugin.

E.g.
```diff
<   - name: Sample-API
---
>   - name: Sample_API
13c13
<         name: Sample-API-list-get
---
>         name: Sample_API-list-get
```

## Investigation

After pairing with @DMarby and @gatzjames we found out some things:

- Folks that used `inso-cli` installed via npm (e.g. `npm i insomnia-inso@2.8.0`) faced a weird behavior where declarative kong configs that had names and some strings with stuff using `-` dashes stayed with dashes.
- BUT, folks that used the generated the declarative config using the `inso cli` from Github, or in the Insomnia Plugin, the `-` dash characters were replaced with underscore.

This was happening due to a difference in versions of `slugify`.

The packaged inso/insomnia was using an older `slugify` version that caused `-` dashes to be wrongly replaced.
Folks using `inso` from NPM had the latest `slugify` version (e.g. `1.6.5`) which didn't have this behavior.

Ultimately this difference in what `slugify` version was used on packaged inso vs. on inso via npm caused the overall weirdness that folks reported in #4432 and also in support tickets.

In this PR we pin `slugify` to `1.6.5` so the behavior stays consistent between packaged and npm `inso`, and we fixed and added tests to check the behavior is consistent.

## TODO
- [x] add fixtures to reproduce the issue
- [x] add a fix


DMarby (approved):

gatzjames (approved):
Tested and works nice!

PR #4667, opened 2022-04-04
Chore: Update extensions/v1beta1 to networking.k8s.io/v1 on OpenAPI-2-Kong

filfreire:
This is in order to follow the deprecation guide of past k8s version +1.22

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122

> Migrate manifests and API clients to use the networking.k8s.io/v1 API version, available since v1.19.
> All existing persisted objects are accessible via the new API
> Notable changes:
>  - `spec.backend` is renamed to `spec.defaultBackend`
>  - The backend `serviceName` field is renamed to `service.name`
>  - Numeric backend `servicePort` fields are renamed to `service.port.number`
>  - String backend `servicePort` fields are renamed to `service.port.name`
>  - `pathType` is now required for each specified path. Options are Prefix, Exact, and ImplementationSpecific. To match the undefined v1beta1 behavior, use ImplementationSpecific.

changelog(OpenAPI-2-Kong): Generated ingresses for KIC are now using the networking.k8s.io/v1 API version


tomanagle (approved):

sidwebworks (approved):

filfreire:
@wdawson for awareness:

- once merged this is **breaking change** for folks using kubernetes bellow v1.19.
- kubernetes 1.21 enters maintenance mode on 2022-04-28, End of Life is 2022-06-28
- kubernetes 1.22 enters maintenance mode on 2022-08-28, End of Life is 2022-10-28
- not merging means on the other hand that the experience is broken for folks using kubernetes above and including 1.22.


Source: https://kubernetes.io/releases/patch-releases/

wdawson:
Let's make sure we document this really well but sounds like we need to make a major version bump for sure.

DMarby (approved):

PR #4669, opened 2022-04-04
chore(deps): styled-components

dimitropoulos:
We have some pretty important work on moving to Vite (https://github.com/Kong/insomnia/pull/4329) that's being blocked by styled-components

Screenshot_20220404_174415

see: https://github.com/microsoft/TypeScript/issues/37597#issuecomment-628149946
____

The other we need to upgrade styled-components just for the sake of it.  We're pretty far behind.


gatzjames (approved):
🎉

PR #4677, opened 2022-04-07
Fix wsdl imports

DMarby:
changelog(Fixes): Fixed an issue where importing WSDL files into a request collection or design document was not working


filfreire (approved):
Seems to be back working 🎉 
Screenshot 2022-04-07 at 17 04 18

PR #4689, opened 2022-04-12
expose react and react-dom in plugin private context

gatzjames:
Fix an issue with requiring react from the insomnia-plugin-kong-portal.

In latest beta the plugin throws an error when selecting `Deploy To Portal` on a design document (OAS).
The issue is that while on dev or local production mode since we still use node's module resolution the plugin is able to find the same react/react-dom module that the render requires.
When we package the app this changes and creates issues like `You might be using  multiple versions of React` etc.

This PR exposes React and ReactDOM alongside insomnia-components so that we don't have to deal with these issues anymore.

### DIFF
--- a/plugins/insomnia-plugin-kong-portal/src/document-actions.tsx
+++ b/plugins/insomnia-plugin-kong-portal/src/document-actions.tsx
@@ -1,8 +1,8 @@
 import axios from 'axios';
-import React from 'react';
-import ReactDOM from 'react-dom';
+import type React from 'react';
+import type ReactDOM from 'react-dom';
 
-import { DeployToPortal } from './deploy-to-portal';
+import { getDeployToPortalComponent } from './deploy-to-portal';
 
 // This is a temporary shim until insomnia-app exports plugin types that plugin authors can use
 export interface Spec {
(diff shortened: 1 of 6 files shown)


plugins/insomnia-plugin-kong-portal/src/document-actions.tsx
```diff
@@ -24,7 +24,11 @@ export interface Context {
     analytics: {
       trackSegmentEvent: (event: string, properties?: Record<string, any>) => any;
     };
-    insomniaComponents: any;
+    loadRendererModules: () => Promise<{
+      insomniaComponents: any;
+      ReactDOM: typeof ReactDOM;
+      React: typeof React;
+    }>;
```
dimitropoulos, line 31: I really look forward to us not having to duplicate this..

plugins/insomnia-plugin-kong-portal/src/document-actions.tsx
```diff
@@ -35,15 +39,19 @@ export const documentActions = [
   {
     label: 'Deploy to Dev Portal',
     hideAfterClick: true,
-    action(context: Context, spec: Spec) {
+    async action(context: Context, spec: Spec) {
```
dimitropoulos: I don't think we can change this so easily.  Can we try to just do promise chains?  I realize it isn't pretty, but I really don't want to mess with the plugin API surface.

dimitropoulos: I did this in 0f3838d9dddd85407bb0cdb58e23c37dd523b1c7 to show you what I mean.  feel free, as always, to drop the commit if you have another idea

dimitropoulos (approved):

PR #4694, opened 2022-04-13
chore: bracket lint rules

dimitropoulos:
We have been unofficially following the rules in this PR _ALMOST_ 100%.  I noticed in a [recent PR](https://github.com/Kong/insomnia/pull/4687/files#diff-753915334846f8acd14467777fe41ddaaece4c2d3ec2f33f60837d6d872f6c3bR122) and was surprised it wasn't an error.  So I thought it would be good to make it official.

and in case you're wondering if this really matters: see how a little thing like skipping braces (and not having sufficient static analysis) can cost millions of dollars: https://dwheeler.com/essays/apple-goto-fail.html


marckong (approved):
Looks good to me, and great background knowledge sharing!

PR #4702, opened 2022-04-15
fix: environment modal fix

marckong:
This actually fixes the issue in the environment modal without changing behaviours in the current develop branch, and this


dimitropoulos (approved):
@marckong and I went through every shortcut and verified no change from current production.  then, at the same time, this fixes the bug in modals -> so win-win.

PR #4706, opened 2022-04-18
removes duplicated "config.longName"

dimitropoulos:
while reviewing https://github.com/Kong/insomnia/pull/4704 I noticed that `longName` and `productName` are exactly the same value and _ALMOST_ everywhere in the app `productName` was used.

I spent longer than I probably should have trying to verify _ANY_ justification for this, but I turned up with nothing.  Before there was a config file these values were just hardcoded https://github.com/Kong/insomnia/blob/c79d7fdd7757b062278131f817af7120848ded13/.electronbuilder.  Then later the config file first appeared in the (private) Insomnia Designer repo.  But even there, it had duplicated values for `productName` and `longName` (`Insomnia Designer`):

So I can only conclude that this was a premature optimization that never found a purpose.

Beyond that, I also searched electron-builder to make sure these aren't special values in some way.  So far as I can tell, they are not.

Second of that, since I already went through the diligence of figuring all of this out, I went ahead and made the config.json key match what we actually call it in the app: `appName`:

### DIFF
--- a/packages/insomnia-app/app/main.development.ts
+++ b/packages/insomnia-app/app/main.development.ts
@@ -39,9 +39,7 @@ const envDataPath = process.env.INSOMNIA_DATA_PATH;
 if (envDataPath) {
   app.setPath('userData', envDataPath);
 } else if (!isDevelopment()) {
-  // Explicitly set userData folder from config because it's sketchy to
-  // rely on electron-builder to use productName, which could be changed
-  // by accident.
+  // Explicitly set userData folder from config because it's sketchy to rely on electron-builder to use apName, which could be changed by accident.
   const defaultPath = app.getPath('userData');
   const newPath = path.join(defaultPath, '../', appConfig.userDataFolder);
   app.setPath('userData', newPath);
--- a/packages/insomnia-app/app/main.development.ts
+++ b/packages/insomnia-app/app/main.development.ts
@@ -39,7 +39,7 @@ const envDataPath = process.env.INSOMNIA_DATA_PATH;
 if (envDataPath) {
   app.setPath('userData', envDataPath);
 } else if (!isDevelopment()) {
-  // Explicitly set userData folder from config because it's sketchy to rely on electron-builder to use apName, which could be changed by accident.
+  // Explicitly set userData folder from config because it's sketchy to rely on electron-builder to use productName, which could be changed by accident.
   const defaultPath = app.getPath('userData');
   const newPath = path.join(defaultPath, '../', appConfig.userDataFolder);
   app.setPath('userData', newPath);
--- a/packages/insomnia-app/app/main/window-utils.ts
+++ b/packages/insomnia-app/app/main/window-utils.ts
@@ -7,6 +7,7 @@ import { pathToFileURL } from 'url';
 import {
   changelogUrl,
   getAppBuildDate,
+  getAppVersion,
   getProductName,
   isDevelopment,
   isLinux,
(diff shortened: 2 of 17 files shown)


packages/insomnia-app/scripts/build.ts
```diff
@@ -150,8 +150,7 @@ const generatePackageJson = async (relBasePkg: string, relOutPkg: string) => {
   const appPkg = {
     name: appConfig.name,
     version: appConfig.version,
-    productName: appConfig.productName,
-    longName: appConfig.longName,
+    appName: appConfig.appName,
```
DMarby: The `productName` property within package.json is used by electron-builder to allow for different characters/case then the `name` property, in our case `Insomnia` (capitalised), so we should set `productName` here, not `appName`. 
https://www.electron.build/configuration/configuration.html#configuration

DMarby (changes requested):

packages/insomnia-app/scripts/build.ts
dimitropoulos: great find! I searched the wrong thing in the electron-builder docs (appName rather than productName).  see 8b6bead2c97a79a261c58d81eb1632605b1ba74e.  now, it's back to the original property name in the config, but the rest of the app doesn't rename it

packages/insomnia-app/app/main.development.ts
```diff
@@ -39,9 +39,7 @@ const envDataPath = process.env.INSOMNIA_DATA_PATH;
 if (envDataPath) {
   app.setPath('userData', envDataPath);
 } else if (!isDevelopment()) {
-  // Explicitly set userData folder from config because it's sketchy to
-  // rely on electron-builder to use productName, which could be changed
-  // by accident.
+  // Explicitly set userData folder from config because it's sketchy to rely on electron-builder to use apName, which could be changed by accident.
```
DMarby:
```suggestion
  // Explicitly set userData folder from config because it's sketchy to rely on electron-builder to use productName, which could be changed by accident.
```

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -378,7 +376,7 @@ export function createWindow() {
     const ok = 'OK';
     const buttons = isLinux() ? [copy, ok] : [ok, copy];
     const detail = [
-      `Version: ${getAppLongName()} ${getAppVersion()}`,
+      `Version: ${getProductName()} ${getProductName()}`,
```
DMarby:
```suggestion
      `Version: ${getProductName()} ${getAppVersion()}`,
```

packages/insomnia-app/app/main/window-utils.ts
```diff
@@ -7,9 +7,7 @@ import { pathToFileURL } from 'url';
 import {
   changelogUrl,
   getAppBuildDate,
-  getAppLongName,
-  getAppName,
-  getAppVersion,
+  getProductName,
```
DMarby, line 11:
```suggestion
  getAppVersion,
  getProductName,
```

DMarby (changes requested):

packages/insomnia-app/app/main/window-utils.ts
dimitropoulos: wow, what an embarrassing mistake ba64f42feddbead7bc4f8073ad82cca56ba988c6

Thanks!

packages/insomnia-app/app/main/window-utils.ts
dimitropoulos: ba64f42feddbead7bc4f8073ad82cca56ba988c6

packages/insomnia-app/app/main.development.ts
dimitropoulos: ba90cd2640b271463aef474c0f5cfe1d9de8fa43

DMarby (approved):

PR #4709, opened 2022-04-19
catches JSON.parse possible error in oauth2 identity token field

dimitropoulos:
see https://youtu.be/PSUys7A0iVE?t=2298 for a demo of what can break and how this PR fixes it

changelog(Fixes): Fixes potential issue where typing an invalid JWT into some of the OAuth2 fields can crash the OAuth2  interface


gatzjames (approved):
👏 🚀

PR #4711, opened 2022-04-20
fix(styled-components): add specificity selector

gatzjames:
Fixes specificity issues with styled components that rely on import order.

The order of css/styled imports can affect which styles get applied to an element.
When we extend a styled component we usually override it's styles like so:
```tsx
const StyledBytton = styled(Button)({
  marginLeft: 5
})
```
The intend here is to override the marginLeft value for the Button. This will only happen if the className with the marginLeft override loads last in the Document.
To guarantee that the styles will override the previous ones we need to [increase the specificity](https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity)


filfreire (approved):
Tested this change locally, didn't notice anything apparent broken 👍

DMarby (approved):

PR #4718, opened 2022-04-22
4688 Header ending with -bin need buffer values

### LINKED ISSUE #4688: gRPC Request: Headers ending with -bin does not convert value to Buffer object.
### Expected Behavior

Adding a new header ending with '-bin' should convert the 'Value' column value into Buffer object.

### Actual Behavior

When sending a request, the following error is shown:

image

### Reproduction Steps

1. Create 'New Request'
2. Select 'gRPC'
3. Select folder/file containing protos
4. Enter valid url pointing to grpc service
5. Select method to execute
6. Selected 'Headers' tab
7. Add new header, make sure it ends with '-bin'.  Example: 'x-custom-bin'
8. Supply base64 value
9. Click 'Send'

### Additional Information

Reproduced on macOS as well.

I did some digging and this is required by grpc-js package.  Even though it isn't the 'right' fix, it did seem fairly easy to detect that header ends with '-bin' and converting the base64 value into a Buffer object.  Then everything works great (in my own tests)

### Insomnia Version

2022.2.1

Windows

### Operating System Version

Windows 10

### Installation method

Download from insomnia.net

_No response_

jevooq:
For metadata entries where the name ends with -bin the values must be
Buffers instead of strings.

The solution checks if the metadata entry should be a buffer and, when
yes, constructs a Buffer from the value. The value is split into the
encoding part and the real value part. If no encoding is specified the
Buffer will be constructed with the default encoding.

The default encoding when none is provided to the `Buffer.from` call is `utf8`.

The solution accepts empty values, as well as values without the encoding part.

The encoding part is not restricted to the current encoding values, so it should
be ok if in future new values are added. It is restricted though to some characters:
* a-z
* A-Z
* 0-9
* underscore (_)
* hyphen (-)

Closes #4688.

### DIFF
--- a/packages/insomnia/src/network/grpc/index.ts
+++ b/packages/insomnia/src/network/grpc/index.ts
@@ -323,7 +323,7 @@ const _parseMetadata = (
   for (const entry of metadata) {
     if (!entry.disabled) {
       try {
-        grpcMetadata.add(entry.name, entry.value);
+        grpcMetadata.add(entry.name, _parseMetadataValue(entry));
       } catch (err) {
         respond.sendError(requestId, err);
         return undefined;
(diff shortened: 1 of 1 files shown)


codedjinn:
Is there anyway to get some more traction on this fix ?

jevooq:
:thinking: Is there anything I can do to improve the chances of pushing this forward?

filfreire:
Hi @jevooq thanks for contributing this one!

We're going to close this for now and tackle the -bin issue when we work on gRPC later on as part of our roadmap.

Feel free to report any issues you find on gRPC as those will be useful for us, similar to the #4688 issue.

PR #4721, opened 2022-04-22
Update highlight.js and only include common languages

DMarby:
@gatzjames found out that this substantially reduces our bundle size as we were including support for some very esoteric languages.


gatzjames (approved):
🚀

filfreire (approved):

PR #4722, opened 2022-04-22
Publish Inso CLI docker images (INS-1238)

filfreire:
This PR adds needed functionality for us to start publishing `inso-cli` docker images.

### How to test docker image locally:

- Checkout this branch
- Download the .tar of  inso CLI Linux artifact into the `packages/insomnia-inso/artifacts/` folder.
- Build container:
```
docker build --tag insomnia-inso:latest ./packages/insomnia-inso/
```
- Run typical `inso` commands by mounting on folder where you have `.insomnia` data, e.g.:
```bash
docker run -it --rm -v $(pwd):/var/temp insomnia-inso:latest generate config -w /var/temp -f json
```

changelog(Improvements): We now publish Docker images for Inso CLI at https://hub.docker.com/r/kong/inso


.github/workflows/release-publish.yml
```diff
@@ -155,6 +155,22 @@ jobs:
...
+          docker tag insomnia-inso:temp ${DOCKER_REGISTRY}/${DOCKER_IMAGE}:${DOCKER_LATEST_TAG}
+          docker push ${DOCKER_REGISTRY}/${DOCKER_IMAGE}:${{ env.INSO_VERSION }}
+          docker push ${DOCKER_REGISTRY}/${DOCKER_IMAGE}:${DOCKER_LATEST_TAG}
+
+        env:
+          DOCKER_REGISTRY_TOKEN: ${{ secrets.DOCKER_REGISTRY_TOKEN }}
+          DOCKER_REGISTRY_USER: ${{ secrets.DOCKER_REGISTRY_USER }}
+          DOCKER_REGISTRY: docker.io
+          DOCKER_IMAGE: kongcloud/insomnia-inso
+          DOCKER_LATEST_TAG: ${{ env.IS_PRERELEASE == 'false' && 'latest' || contains(github.event.inputs.version, 'alpha') && 'alpha' || 'beta' }}
```
filfreire, line 172: We use this for the `latest`-like tags besides the specific version tags.

- If it's a stable release - we'll update the `latest` tag
- If it's a beta or alpha release - we'll update `beta` or `alpha` tags respectively (acting as a `latest` tag for beta/alpha track

cc @DMarby @johnwchadwick

dimitropoulos:
tell me if this is out of scope: but I'm worried that this will break in the future and we won't know.  is there some way to do a quick sanity check on the completed image to verify it works?

.github/workflows/release-publish.yml
johnwchadwick: SGTM.

johnwchadwick (approved):
LGTM

packages/insomnia-inso/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# It assumes inso-linux-VERSION.tar.xz exists in /packages/insomnia-inso/artifacts
+# You can run `npm run inso-package:artifacts` on a linux host OR
+# `curl -LO "https://github.com/Kong/insomnia/releases/download/lib%40<version>/inso-linux-<version>.tar.xz"`
+
+FROM docker.io/alpine AS fetch
+
+COPY ./artifacts/inso-linux-*.tar.xz /tmp/inso.tar.xz
+RUN tar -C /usr/bin -xvf /tmp/inso.tar.xz
+
+FROM docker.io/alpine as binary
```
johnwchadwick: Nit: It might? be a good idea to pin an exact alpine release here. I don't anticipate this will ever be an issue, though.

DMarby:
> tell me if this is out of scope: but I'm worried that this will break in the future and we won't know. is there some way to do a quick sanity check on the completed image to verify it works?

I agree with this sentiment in general. I think it's out of scope for publishing/this PR, we intentionally don't run tests in these workflows to keep them simple, quick and error-free, but it would be good to create a linear ticket to follow up on this, as we have similar issues for other publishing methods as well, eg ensuring installing inso and running it from `npm` actually works correctly, so in the longer term we need to figure out an approach how to test those cases more robustly (whether that's by implementing additional checks in our existing `Test` pipeline running on every PR, or some other pipeline running more infrequently, etc)

filfreire:
@DMarby @dimitropoulos I've added a new linear issue to triage to follow-up on doing sanity checks for the inso docker image - INS-1525

@johnwchadwick I agree with pinning alpine to a proper version, will edit the PR.

DMarby (approved):

PR #4720, opened 2022-04-22
remove pdfjs-dist in favor of native chromium pdf viewer

gatzjames:
Changelog(Improvements): PDF responses are now previewed using Chromium's native PDF viewer

Remove pdfjs and use the native pdf viewer instead.

pdfjs is a big part of our bundle and is only used in the PDF preview. Since electron supports rendering PDFs natively we can remove it and save some kb!

pdf

TODO:
Validate the behaviour in all supported platforms
- [x] macOS
- [x] linux
- [x] windows

PR #4723, opened 2022-04-22
Fix pdf rendering

DMarby:
Fixes PDF rendering by setting the workerSrc.
This is not ideal but will tie us over until we can switch to the native PDF viewer in #4720


filfreire (approved):
This seems to do the trick. LGTM
Screenshot 2022-04-22 at 16 46 57

gatzjames (approved):

marckong (approved):

PR #4724, opened 2022-04-22
Improve experience when debugging with vscode

gatzjames:
Currently when stopping the debug session the tasks remain open.

This PR uses the postDebugTask to stop all running tasks with it but there's some more work to make it robust.
e.g. when reloading etc.


gatzjames:
To be revisited

PR #4749, opened 2022-05-02
Added error handling for github auth urls

Wazbat:
Changelog(Improvements): Improved error messages for Github Sync

I'm not sure how best show these errors to the user, but I feel that instead of these two `console.error`'s that I added there should be some kind of error modal or something like that, but I'm not sure how best to do that in this project

### DIFF
--- a/packages/insomnia/src/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
+++ b/packages/insomnia/src/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
@@ -362,7 +362,7 @@ const GitHubSignInForm = ({ token }: GitHubSignInFormProps) => {
               try {
                 parsedURL = new URL(link);
               } catch (e) {
-                setError('It appears the URL that you entered is invalid');
+                setError('Invalid URL');
                 return;
               }
 
--- a/packages/insomnia/src/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
+++ b/packages/insomnia/src/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
@@ -325,6 +325,7 @@ interface GitHubSignInFormProps {
 }
 
 const GitHubSignInForm = ({ token }: GitHubSignInFormProps) => {
+  const [error, setError] = useState('');
   const [authUrl, setAuthUrl] = useState(() => generateAuthorizationUrl());
   const [isAuthenticating, setIsAuthenticating] = useState(false);
   const dispatch = useDispatch();
@@ -361,8 +362,7 @@ const GitHubSignInForm = ({ token }: GitHubSignInFormProps) => {
               try {
                 parsedURL = new URL(link);
               } catch (e) {
-                // setError('It appears the URL that you entered is invalid');
-                console.error('Provided url is invalid');
+                setError('It appears the URL that you entered is invalid');
                 return;
               }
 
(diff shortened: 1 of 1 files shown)


jackkav (commented):
you'll need to copy the error setState hook and html in the same places as `GitHubRepositoryForm`

`const [error, setError] = useState('');`

```
          {error && (
            <p className="notice error margin-bottom-sm">
              <button className="pull-right icon" onClick={() => setError('')}>
                <i className="fa fa-times" />
              </button>
              {error}
            </p>
          )}
```
tuck it just inside the closing form tag

Wazbat:
> you'll need to copy the error setState hook and html in the same places as `GitHubRepositoryForm`
> 
> `const [error, setError] = useState('');`
> 
> ```
>           {error && (
>             <p className="notice error margin-bottom-sm">
>               <button className="pull-right icon" onClick={() => setError('')}>
>                 <i className="fa fa-times" />
>               </button>
>               {error}
>             </p>
>           )}
> ```
> 
> tuck it just inside the closing form tag

I appreciate that. I've gone ahead and added the code. I tested it locally and both errors show and appear to look fine.

Not sure if you'd like me to change the text, as it might be better to go with a more clear definitive `The URL you entered is invalid` instead of the vague mysterious `It appears...` message

image

image

packages/insomnia/src/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
```diff
@@ -357,7 +358,14 @@ const GitHubSignInForm = ({ token }: GitHubSignInFormProps) => {
             const formData = new FormData(e.currentTarget);
             const link = formData.get('link');
             if (typeof link === 'string') {
-              const parsedURL = new URL(link);
+              let parsedURL: URL;
+              try {
+                parsedURL = new URL(link);
+              } catch (e) {
+                setError('It appears the URL that you entered is invalid');
```
jackkav:
```suggestion
                setError('Invalid URL');
```

packages/insomnia/src/ui/components/modals/git-repository-settings-modal/github-repository-settings-form-group.tsx
```diff
@@ -368,20 +376,30 @@ const GitHubSignInForm = ({ token }: GitHubSignInFormProps) => {
                 });
 
                 command(dispatch);
+              } else {
+                setError('It appears the URL that you entered is incomplete');
```
jackkav:
```suggestion
                setError('Missing URL');
```

Wazbat: Is "Missing URL" correct? It will only show this error if there is a valid url, howevever the `state` or `code` query params are missing

Perhaps `Incomplete URL` would be better?

jackkav: Ah yes I had misread because of the nested if statements perhaps you can use guard pattern to clear up the misunderstanding.
```
const hasCodeAndState = typeof code === 'string' && typeof state === 'string'
if(!hasCodeAndState) {
  setError('Incomplete URL');
  return;
}
const command = newCommand(COMMAND_GITHUB_OAUTH_AUTHENTICATE, {
  code,
  state,
});

command(dispatch);
```

Wazbat: Gotcha. I've moved that logic up as requested and changed the error text, though I implemented it without the `hasCodeAndState` variable.

Is it still readable or would you prefer an additional variable to make it easier to understand when glancing over the code?

filfreire (approved):

PR #4756, opened 2022-05-04
Correctly handle per-operation scopes when importing from OpenAPI 3

### LINKED ISSUE #4755: Import from OpenAPI 3 does not correctly handle per-request scopes
### Expected Behavior

When importing an OpenAPI 3 Schema, Insomnia should correctly handle the OAuth2 scopes that are required for each request ("operation" in OpenAPI speak). The relevant documenation can be found here: https://swagger.io/docs/specification/authentication/oauth2/#scopes-extra (note point 2 about the required scopes for each operation). See also the [specification](https://swagger.io/specification/#security-requirement-object):
>  If the security scheme is of type "oauth2" or "openIdConnect", then the value is a list of scope names required for the execution, and the list MAY be empty if authorization does not require a specified scope.

### Actual Behavior

Insomnia ignores the scopes that are defined as required for each operation and instead lists _all_ known scopes for every request. The result of this is that I have to adjust every imported operation manually, because the OAuth client I am testing with might not have permission to access _all_ scopes.

### Reproduction Steps

1. Import this [example schema](https://gist.github.com/diesieben07/582f2b5f3f46ea7fdb26cfb61255e152).
2. Observe _both_ requests being imported with `read write` in the "Scope" field.

### Additional Information

_No response_

### Insomnia Version

2022.3.0

Other Linux

### Operating System Version

`Linux 5.4.0-109-generic #123-Ubuntu SMP Fri Apr 8 09:10:54 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

### Installation method

Download .deb from insomnia.rest

_No response_

diesieben07:
This fixes #4755.
I have also corrected the involved tests, as they were actually testing this wrong behavior.

changelog(Fixes): Fixed an issue with how OAuth2 scopes were handled when importing from OpenAPI 3 specs


packages/insomnia-importers/src/importers/fixtures/openapi3/oauth2-output.json
```diff
@@ -44,7 +44,7 @@
           "authorizationUrl": "https://api.server.test/v1/auth",
           "grantType": "authorization_code",
           "redirectUrl": "{{ oauth2RedirectUrl }}",
-          "scope": "read:something write:something",
+          "scope": "",
```
filfreire, line 47: @diesieben07 can you add 2 other endpoints where
- we only define one of these two scopes
- we define an entirely different scope (not `read:something` and `write:something`)?

filfreire (commented):
@diesieben07 thank you for contributing this 😊 🙏 

I noticed only after writing my comment on #4755 that you had opened a PR with a potential fix.

I wasn't aware that [by the RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3) we could send scopes space-delimited, so it was good to clear that app.

I just have one small comment bellow, let me know if you're up for doing the improvement, if not one of the members of team can help out and finish this PR when possible.

packages/insomnia-importers/src/importers/fixtures/openapi3/oauth2-output.json
diesieben07: Thank you for the review.
I have a question about what you mean by "entirely different scope". Do you want a test that covers a Schema where an endpoint specifies a scope that is not listed in the security scheme? If so, that would technically not be a valid OpenAPI Schema anymore, if I understand the specification correctly. I can add a test for it, but the question is how such a scope should be handled:
- Should it be ignored, since technically it is invalid here?
- Should it be gracefully added to this endpoint anyways, even though it is not specified in the security scheme?

diesieben07: I've now updated the test to test no scopes, only one scope as well as both scopes on the endpoint level.
Please let me know how to proceed regarding my question above.

filfreire: @diesieben07 you are right, the spec would no longer be valid, and in theory in this case the authorization server would have to reply back with an `invalid_scope` [according to spec](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1). 

The question becomes, do we want to allow folks to purposely setup invalid scopes in their specs for testing purposes... 🤷 

For now the safest bet is probably the first option you mention: it should be ignored.

tagging @Kong/team-insomnia in case someone else might have another take on this.

filfreire (approved):

PR #4760, opened 2022-05-05
Start the grpc smoke-test server alongside the http server

DMarby:


packages/insomnia-smoke-test/server/grpc.ts
```diff
@@ -24,17 +24,21 @@ function sayHello(call, callback) {
...
-    console.log('Listening at grpc://0.0.0.0:50051');
-    server.start();
-  });
-}
+export const startGRPCServer = (port: number) => {
+  return new Promise<void>((resolve, reject) => {
+    const server = new grpc.Server();
+    // @ts-expect-error generated from proto file
+    server.addService(helloProto.HelloService.service, { sayHello: sayHello });
+    server.bindAsync(`localhost:${port}`, grpc.ServerCredentials.createInsecure(), (error) => {
```
DMarby: It is not generally desirable to listen on all interfaces (`0.0.0.0`) from a security perspective, hence the change to localhost here. This also means it's consistent with the express server.

packages/insomnia-smoke-test/server/index.ts
```diff
@@ -42,6 +44,8 @@ app.get('/', (_req, res) => {
   res.status(200).send();
 });
 
-app.listen(port, () => {
-  console.log(`Listening at http://localhost:${port}`);
+startGRPCServer(grpcPort).then(() => {
+  app.listen(port, () => {
```
DMarby, line 48: The reason for waiting until the gRPC server is started before we start the express server here, rather than just starting them both at the same time, is that Playwright only checks if the express server is alive before starting tests.
This way we hopefully avoid race conditions where the http server is started but the grpc server is not yet.

jackkav (approved):
nice simple solution :+1:

PR #4761, opened 2022-05-05
chore: add metrics tracking for design preview button

marckong:
We would like to make better designs with more metrics in some area of this project. We need to add tracking logic to this button click event.
- added trackSegmentEvent in wrapper-design.tsx
- added a new enum key in analytics.ts

### DIFF
--- a/packages/insomnia/src/common/analytics.ts
+++ b/packages/insomnia/src/common/analytics.ts
@@ -84,9 +84,6 @@ export enum SegmentEvent {
  * ourselves to make product decisions
  */
 interface SegmentEventProperties {
-  /**
-   * a sub category of an event, useful for analytics querying
-   * */
   type: string;
 
   /**
@@ -95,7 +92,7 @@ interface SegmentEventProperties {
   action: string;
 
   /**
-   * any error message as a consequence of the event
+   * a description of an error that occured as a result of the user action
    * */
   error?: string;
 }
--- a/packages/insomnia/src/common/analytics.ts
+++ b/packages/insomnia/src/common/analytics.ts
@@ -76,7 +76,7 @@ export enum SegmentEvent {
   buttonClick = 'Button Clicked',
 }
 
-interface SegmentEventProperties {
+export interface SegmentEventProperties {
   /**
    * TODO: what do we mean by type?
    * */
@@ -93,15 +93,6 @@ interface SegmentEventProperties {
   error?: string;
 }
 
-export type BaseButtonAction = 'show' | 'hide';
-export function buildEventProperties<T extends string>(
-  type: string,
-  action: T,
-  error?: string
-): SegmentEventProperties {
-  return { type, action, error };
-}
-
 type PushPull = 'push' | 'pull';
 type VCSAction = PushPull | `force_${PushPull}` |
   'create_branch' | 'merge_branch' | 'delete_branch' | 'checkout_branch' |
@@ -112,7 +103,7 @@ export function vcsSegmentEventProperties(
   action: VCSAction,
   error?: string
 ): SegmentEventProperties {
-  return buildEventProperties<VCSAction>(type, action, error);
+  return { type, action, error };
 }
 
 interface QueuedSegmentEvent {
--- a/packages/insomnia/src/common/analytics.ts
+++ b/packages/insomnia/src/common/analytics.ts
@@ -73,6 +73,7 @@ export enum SegmentEvent {
   vcsSyncStart = 'VCS Sync Started',
   vcsSyncComplete = 'VCS Sync Completed',
   vcsAction = 'VCS Action Executed',
+  designOutputPreviewToggle = 'Design Output Preview Toggled',
 }
 
 type PushPull = 'push' | 'pull';
--- a/packages/insomnia/src/common/analytics.ts
+++ b/packages/insomnia/src/common/analytics.ts
@@ -76,8 +76,24 @@ export enum SegmentEvent {
   buttonClick = 'Button Clicked',
 }
 
-type PushPull = 'push' | 'pull';
+interface SegmentEventProperties {
+  /**
+   * TODO: what do we mean by type?
+   * */
+  type: string;
 
+  /**
+   * TODO: what do we mean by action?
+   * */
+  action: string;
+
+  /**
+   * TODO: what is this error for and where is it coming from?
+   * */
+  error?: string;
+}
+
+type PushPull = 'push' | 'pull';
 export function vcsSegmentEventProperties(
   type: 'git',
   action: PushPull | `force_${PushPull}` |
@@ -85,7 +101,24 @@ export function vcsSegmentEventProperties(
     'commit' | 'stage_all' | 'stage' | 'unstage_all' | 'unstage' | 'rollback' | 'rollback_all' |
     'update' | 'setup' | 'clone',
   error?: string
-) {
+): SegmentEventProperties {
+  return {
+    'type': type,
+    'action': action,
+    'error': error,
+  };
+}
+
+export enum BaseButtonEvent {
+  show = 'show',
+  hide = 'hide',
+}
+
+export function buildEventProperties<T extends string>(
+  type: string,
+  action: T,
+  error?: string
+): SegmentEventProperties {
   return {
     'type': type,
     'action': action,
--- a/packages/insomnia/src/common/analytics.ts
+++ b/packages/insomnia/src/common/analytics.ts
@@ -80,8 +80,7 @@ export enum SegmentEvent {
  * You may refer to this Segment API Document - https://segment.com/docs/connections/spec/track/#properties
  * to understand what 'event properties' is for.
  *
- * The following properties are custom defined attributes for our own metrics purpose to help
- * ourselves to make product decisions
+ * The following properties are custom defined attributes for our own metrics purpose to help ourselves to make product decisions
  */
 interface SegmentEventProperties {
   type: string;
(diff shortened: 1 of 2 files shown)


packages/insomnia/src/common/analytics.ts
```diff
@@ -73,6 +73,7 @@ export enum SegmentEvent {
   vcsSyncStart = 'VCS Sync Started',
   vcsSyncComplete = 'VCS Sync Completed',
   vcsAction = 'VCS Action Executed',
+  designOutputPreviewToggle = 'Design Output Preview Toggled',
```
wdawson: I don't think we want a specific event for this particular toggle. We probably want a "Button Clicked" event with a type and/or location parameter. The reasoning is that there are often limits on the number of event names you can have.

wdawson (changes requested):

packages/insomnia/src/ui/components/wrapper-design.tsx
```diff
@@ -50,6 +51,14 @@ const RenderPageHeader: FC<Pick<Props,
 
     const workspaceId = activeWorkspace._id;
     await models.workspaceMeta.updateByParentId(workspaceId, { previewHidden: !previewHidden });
+
+    trackSegmentEvent(
+      SegmentEvent.buttonClick,
+      {
+        type: 'user_event',
```
marckong: maybe adding/defining segment event payload type would be a good idea?

packages/insomnia/src/ui/components/wrapper-design.tsx
```diff
@@ -50,6 +51,14 @@ const RenderPageHeader: FC<Pick<Props,
...
+        action: `preview_button_click__previewHidden_${!previewHidden}`,
+        location: 'wrapper-design',
```
dimitropoulos:
```suggestion
```

We definitely do not want to associate name of a code component to the analytics.  also, if we ever change the name of this component (which, I am actually planning on in this case) this would likely fall out of sync.

dimitropoulos: I tried to look for where `location` came from, but I don't see it elsewhere in the code and I don't see it in segment's docs.

packages/insomnia/src/ui/components/wrapper-design.tsx
wdawson: Sorry, I should have been more prescriptive in my suggestion. Here's the extra properties I think we should go with:

```js
{
  type: 'design preview toggle'
  action: 'show | hide' // should be `show` or `hide` depending on which toggle state occurs as a result of the click
}
```

This will reuse the properties we already have and remain human readable (that's the goal here, not machine readable).

packages/insomnia/src/ui/components/wrapper-design.tsx
wdawson: Yeah, this was my bad from my previous suggestion. I wasn't clear.

@marckong it might be worth adding something like [this](https://github.com/Kong/insomnia/blob/develop/packages/insomnia/src/common/analytics.ts#L80) but for button click properties:
```js
export function buttonClickSegmentEventProperties
```

wdawson (changes requested):

packages/insomnia/src/ui/components/wrapper-design.tsx
marckong: I think the confusion for my end came from lack of documentation or explanation about the interface for Segment Event Properties and how we are using it. I observed there are several formats. I wonder if we need to consolidate it into an interface like this:
```ts
interface SegmentEventProperties {
  /**
   * type - TODO: good to describe - what do we exactly mean by type? is this actually event type? and the SegmentEvent enum represents the event name?
   * */
  type: string;

  /**
   * description of action - TODO: good to describe - what do we mean by action? Maybe we are confusing between "event type", "event name", "user action"?
   * */
  action: string;

  /**
   * error message - TODO: good to describe - why do we need this? where is it coming from?
   * */
  error?: string;
}
```

wdawson: Could be helpful to type it if we want. I think we hadn't gone that far yet because we weren't sure if these properties were going to work or not as we expand the data collection that we do, so wanted a bit more flexibility for now. But I'm good either way.

marckong: Would this metrics [in the Segment UI] outcome be suitable for now?
```js
analytics.track({
  anonymousId: 'deleted-on-purpose',
  event: 'Button Clicked',
  properties: {
    action: 'hide',
    type: 'design preview toggle'
  }
});
```

```js
{
  "anonymousId": "deleted-on-purpose",
  "context": {
    "app": {
      "name": "Insomnia",
      "version": "2022.4.0-beta.0"
    },
    "library": {
      "name": "analytics-node",
      "version": "6.0.0"
    },
    "os": {
      "name": "mac",
      "version": "11.6.3"
    }
  },
  "event": "Button Clicked",
  "integrations": {},
  "messageId": "deleted-on-purpose",
  "originalTimestamp": "2022-05-05T22:22:38.157Z",
  "properties": {
    "action": "hide",
    "type": "design preview toggle"
  },
  "receivedAt": "2022-05-05T22:22:48.591Z",
  "sentAt": "2022-05-05T22:22:48.159Z",
  "timestamp": "2022-05-05T22:22:38.589Z",
  "type": "track",
  "writeKey": "deleted-on-purpose"
}
```

packages/insomnia/src/common/analytics.ts
```diff
@@ -73,23 +73,46 @@ export enum SegmentEvent {
...
+  action: string;
+
+  /**
+   * TODO: what is this error for and where is it coming from?
+   * */
+  error?: string;
+}
+
+export type BaseButtonAction = 'show' | 'hide';
+export function buildEventProperties<T extends string>(
```
marckong: This function may not serve a good purpose other than type checking in the dev time.

packages/insomnia/src/ui/components/wrapper-design.tsx
wdawson: Uh... no I think we want to have all the other data in there. You should use [this function](https://github.com/Kong/insomnia/blob/develop/packages/insomnia/src/common/analytics.ts#L131) in a similar way to other places in the app. 

**Edit:** Maybe I'm misunderstanding. We'd want anonymousId to be populated, not blank. We're also missing other properties like userId, os, version, etc. that should be included in the event that we send.

packages/insomnia/src/common/analytics.ts
```diff
@@ -73,23 +73,46 @@ export enum SegmentEvent {
   vcsSyncStart = 'VCS Sync Started',
   vcsSyncComplete = 'VCS Sync Completed',
   vcsAction = 'VCS Action Executed',
+  buttonClick = 'Button Clicked',
 }
 
-type PushPull = 'push' | 'pull';
+interface SegmentEventProperties {
+  /**
+   * TODO: what do we mean by type?
```
marckong: Let's take time to define what these actually mean.

packages/insomnia/src/ui/components/wrapper-design.tsx
marckong: Just to clarify, i deleted ids on purpose and this is the same way vcs tracking method executes the segment tracking API, so in terms of what are captured, they are same kinds. I could be wrong.

wdawson (approved):
LGTM from the product side. I'll let the devs also comment of the code etc.

dimitropoulos (commented):
I'm sorry.  I'm confused.  Why can't we just add this single line to wrapper-design:

```ts
    trackSegmentEvent(SegmentEvent.buttonClick, {
      type: 'design toggle preview',
      action: previewHidden ? 'show' : 'hide',
    });
```

packages/insomnia/src/common/analytics.ts
```diff
@@ -73,23 +73,46 @@ export enum SegmentEvent {
...
+  error?: string;
+}
+
+export type BaseButtonAction = 'show' | 'hide';
+export function buildEventProperties<T extends string>(
+  type: string,
+  action: T,
+  error?: string
+): SegmentEventProperties {
+  return { type, action, error };
```
dimitropoulos: this code will always send an `error` property, even when there is none.  not sure if that's the best thing to do.  I see that this helper function did this before, but at least we weren't calling it everywhere, you know?

filfreire (approved):

packages/insomnia/src/common/analytics.ts
```diff
@@ -73,23 +73,41 @@ export enum SegmentEvent {
   vcsSyncStart = 'VCS Sync Started',
   vcsSyncComplete = 'VCS Sync Completed',
   vcsAction = 'VCS Action Executed',
+  buttonClick = 'Button Clicked',
 }
 
-type PushPull = 'push' | 'pull';
+/**
+ * You may refer to this Segment API Document - https://segment.com/docs/connections/spec/track/#properties
+ * to understand what 'event properties' is for.
```
dimitropoulos:
```suggestion
 * You may refer to this Segment API Document - https://segment.com/docs/connections/spec/track/#properties to understand what 'event properties' is for.
```

packages/insomnia/src/common/analytics.ts
```diff
@@ -73,23 +73,41 @@ export enum SegmentEvent {
...
+ *
+ * The following properties are custom defined attributes for our own metrics purpose to help
+ * ourselves to make product decisions
```
dimitropoulos:
```suggestion
 * The following properties are custom defined attributes for our own metrics purpose to help ourselves to make product decisions
```

no need to introduce arbitrary line wraps

dimitropoulos (approved):

Issue #4715, opened 2022-04-20
Oauth2 Credentials Setting Doesn't Toggle


mttrhll:
### Expected Behavior

When using Oauth2 authentication with Client Credentials users can switch between "As Basic Auth Header" and "In Request Body".

### Actual Behavior

Requests are always sent in request body regardless of setting.

### Reproduction Steps

1. Create a new request
2. Select "Oauth 2" as authentication
3. For "GrantType" select "Client Credentials"
4. Fill in valid client ID, client secret, scope
5. Leave "Credentials" as default value of "As Basic Auth Header"
6. Retrieve a token (this should work so far)
7. Switch "Credentials" to "In Request Body" (many oauth providers will not return a token when the body contains credentials, this must fail to get a token to proceed with this use case)
8. Switch "Credentials" back to "As Basic Auth Header" - this will no longer work, it should go back to headers and succeed.

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Additional Information

Further details:

I grep'd through my insomnia data dir and found that the value in "insomnia.Request.db" for "credentialsInBody" seems to toggling correctly. But whatever underlying system gets/needs it doesn't change its behavior.

Also, if you duplicate the request that doesn't work the copy will have this behavior. But if you create an all new request and  leave the default it will retrieve a token successfully.

**My version/system info is below:**
Version: Insomnia 2022.2.1
Release date: 3/21/2022
OS: Darwin x64 21.4.0
Electron: 17.1.0
Node: 16.13.0
V8: 9.8.177.11-electron.0
Architecture: x64

### Insomnia Version

2022.2.1

### What operating system are you using?

macOS

### Operating System Version

macOs Monterey 12.3.1 (21E258)

### Installation method

insomnia.rest DMG

dimitropoulos:
We looked at this today [on the insomnia stream](https://youtu.be/n5nsA_fJ4vI?t=432).  In short.  YEP.

Sorry this is happening and we definitely need to look into it.  If nothing else, this spawned https://github.com/Kong/insomnia/pull/4768 which will at least give us some means to test this behavior as we devise a fix.

Apologies for taking so many days to get back to you on this.

reinierkleipool:
I have the same problem. CREDENTIALS drop down does not toggle between "As Basic Auth Header (default)" and "In request Body"
Token requests are always sent "In request body" regardless of selection.
**Seen in**:
[...]

PR #4768, opened 2022-05-10
always show debug oauth2 request timeline button

dimitropoulos:
_if this doesn't get the award for "best one-liner" I don't know what will, haha_

## Context

When you do OAuth2 "stuff" (in general) there are separate out-of-band calls that happen sometimes, for example to a `/token` endpoint to fetch your token.  Of course, when using OAuth2 Authentication in Insomnia, Insomnia handles this for you.  Great!

What's not great, though, is that you can't actually _see_ that request.  If you look at the timeline you see the timeline for the request corresponding to what's in your URL bar, which in this example is `/client-credentials`.

Screenshot_20220510_164927

## The behavior before this PR

Today [on the stream](https://youtu.be/n5nsA_fJ4vI?t=432) we were debugging https://github.com/Kong/insomnia/issues/4715 but had a pretty hard time because prior to now, you could not see successful requests to the `/token` endpoint and the bug we were looking into related to this OAuth-specific request.  Hilariously, the stream is a good example of exactly the kind of problems our users face when trying to debug such things on their own.

However, we located something of an escape hatch.  You see, when you have an _error_ in a request, just recently we added the functionality to see the timeline for the error.

It looks like this:
Screenshot_20220510_164206

## What this PR changes

To solve this, we tried (and failed at a few things), but eventually noticed that if we simply render the button at all times, it will correctly show the `/token` request whenever it has been made (including if it's cached) and not otherwise.

Notice the button on the `AFTER` that reads "Response Timeline"
| BEFORE | AFTER |
| - | - |
| Screenshot_20220510_164148 | Screenshot_20220510_164130 |

While from a code perspective this is a pretty minor (literal one line) change... this has pretty cool implications for users trying to investigate the OAuth calls that Insomnia makes.

The result of this is that when a user clicks this new (now, always present) button, they can see the full timeline whenever they wish:

Screenshot_20220510_210641

changelog(Improvements): adds a "Response Timeline" so that now you can see a raw timeline output of all OAuth 2 token requests

### DIFF
--- a/packages/insomnia/src/ui/components/editors/auth/o-auth-2-auth.tsx
+++ b/packages/insomnia/src/ui/components/editors/auth/o-auth-2-auth.tsx
@@ -417,7 +417,7 @@ const OAuth2Error: FC = () => {
       </div>
     );
   }
-  return null;
+  return debugButton;
 };
 
 const useActiveOAuth2Token = () => {


filfreire (approved):
image

PR #4781, opened 2022-05-13
refactor: rename the git sync submission label

marckong:
## Description
Currently, the Git Repository Settings Modal renders its submission button with "Done" label, which misleads its intention; closing the modal as its  complete or closing it as a dismissal.

After some discussion, instead of disabling the Done button when form is invalid and changing the label to "Sync", we just decided to move ahead with renaming the label only at this time until we refactor modal and form pattern.

image


packages/insomnia-smoke-test/tests/git-sync.test.ts
```diff
@@ -44,5 +44,5 @@ test('Sign in with GitHub', async ({ app, page }) => {
     .locator('input[name="uri"]')
     .fill('https://github.com/insomnia/example-repo');
 
-  await page.locator('text=Done').click();
+  await page.locator('button[name="git-sync-done"]').click();
```
dimitropoulos: please use `data-testid` for this kind of thing.  it's a common pattern, but you can read about it a little more here https://testing-library.com/docs/queries/bytestid/

dimitropoulos: and for playwright, here https://playwright.dev/docs/selectors#id-data-testid-data-test-id-data-test-selectors

dimitropoulos (changes requested):
please use `data-testid` for test ids, see comment

packages/insomnia-smoke-test/tests/git-sync.test.ts
marckong: Do we rip off data-testid for the production build in the babel setup or this gets included?

marckong: Our e2e test doesn't happen in the browser, so not sure keeping the data-testid in the production build makes sense if we want to rely on data-testid for selecting components. I generally think it's good to have accessibility attributes used for components and selecting them in tests. However, I have no issue using data-testid as long as we strip it off.

dimitropoulos (approved):

Issue #1694, opened 2019-10-02
[Bug] – Ability to comment-out lines in GraphQL query lost from 6.6.2


fivetwelve:
- Insomnia Version: 7.0.0
- Operating System: MacOS

## Details

Hi, recent adopter of Insomnia here. I've used Postman in the past and have been doing a lot of GraphQL recently and came across Insomnia and it's been fabulous for that.

Given I need to create and test GraphQL queries, I use Insomnia's query editor.

When I want to comment-out a line, I invoke the shortcut, `CMD-/` (much like VS Code and other IDE editors).

Then the line should be prepended with the hashmark, `#`. In version 7.0.0 this does not appear to work.

I was surprised and pleased this was possible in 6.6.2 (being new I can't speak for earlier versions) so I endeavoured to look for its setting in preferences but could not find one. I've rolled back my installation to 6.6.2 just to have this ability back.

AdrianCarreno:
I have the same issue in windows. I really got used to comment the lines I don´t want to excetue in the request. I hope they fix it soon,

SergioSuarezGil:
I have found this issue looking for documentation on a shortcut to comment / uncomment one or several lines in graphql, with a shortcut (because there is no context menu) is it possible?

AdrianCarreno:
> I have found this issue looking for documentation on a shortcut to comment / uncomment one or several lines in graphql, with a shortcut (because there is no context menu) is it possible?

 on Windows you can comment using `CTRL`+`}` (I have a Latinamerican keyboard, the actual key may be different for you). As OP said, the key binding is pretty much the same as VSCode, so you should try that.

It would be nice to have this shortcut detailed in the hel menu. Maybe submit a feature request?

SergioSuarezGil:
In macos is not working as vscode :(  CMD+SHIFT+7 (/)
<img width="1181" alt="Captura de pantalla 2021-10-13 a las 21 12 22" src="https://user-images.githubusercontent.com/3636232/137203076-4516ee53-2745-4493-8ea6-fbff823654c8.png">

PR #4786, opened 2022-05-16
Allow for the toggleComment call to be configurable on hotkeys

filfreire:
Follow-up to #1694 and to [this slack thread](https://getinsomnia.slack.com/archives/C5BKWBAMN/p1652475869314609)

This PR adds the ability for folks to edit or add their own custom hot key to trigger the `toggleComment` functionality of CodeMirror. Up until now we have it hard-coded to `Ctrl+/` or `Cmd+/` which isn't available as a working command for all kind of keyboard layouts, not allowing folks, for example, to comment out sections of a GraphQL query they're building.

I'm split on this change: 
- it opens a small precedent for exposing keyboard shortcuts specifically to deal with Code Editor functionality, where-as up until now we only had the `Beautify Active Code Editors` option exposed and that was it. 
- On the other hand it helps unblock folks that want to have better interaction with the code editors views we have but can't due keyboard layout differences, and the fact we hardcode several `extraKeys` combos.

Let me know what you think and if it's worth to merge this, or it's opening a can of worms 😅 

Screenshot 2022-05-16 at 12 10 06

https://user-images.githubusercontent.com/11976836/168580191-f05533ff-15ed-47e5-960f-a577ffec2a8e.mp4


gatzjames (approved):
Works nice!
I think it's a common pattern to have editor shortcuts exposed to the app's shortcut config.
As a user I would expect that to be the case and it can open up the use-cases you mentioned!

marckong (approved):
this looks good to me! It wouldn't hurt to add a playwright test though it's understandable that it's hard to do this now. I am also making some changes in this area and it's challenging :(

packages/insomnia/src/common/hotkeys.ts
```diff
@@ -250,6 +251,10 @@ const defaultRegistry: HotKeyRegistry = {
     keyComb(false, false, true, true, keyboardKeys.i.keyCode),
     keyComb(true, false, true, false, keyboardKeys.i.keyCode),
   ),
+  [hotKeyRefs.TOGGLE_COMMENT_CODE.id]: keyBinds(
+    keyComb(false, false, false, true, keyboardKeys.semicolon.keyCode),
+    keyComb(true, false, false, false, keyboardKeys.semicolon.keyCode),
```
dimitropoulos, line 256: does this change the standard keyboard shortcut for this?  for me it's not semicolon, but rather `/`

dimitropoulos: if so, I am not OK with us removing this very very standard shortcut.  VSCode, for example, uses this.  how do they accomplish it if it has remapping problems with operating systems?

dimitropoulos (changes requested):
see my comment

packages/insomnia/src/common/hotkeys.ts
filfreire: Hi @dimitropoulos 👋 

1) This doesn't change the standard one, it adds an extra one which folks can configure, besides `Ctrl+/`.

2) In VSCode they allow folks with a non US keyboard to change the default one, like from using `/` to using `;`, or `ç`, ... because we don't generally have the `/` available with a single keypress.


I wasn't sure how to propagate the change from and to the CodeMirror `extraKeys.toggleComment` variable that is hardcoded when the app starts, so I went ahead and create an alternate key combination (e.g. using `;`).

There's a side effect I didn't mention in the initial PR description: if we put the key combination here the same as CodeMirror `extraKeys.toggleComment` variable, which makes it so toggleComment is triggered twice in a row, basically doing and undoing the commented code in one go.

dimitropoulos: I hate to rain on the parade but if we're going to do this we need to do it right.  I'm glad you clarified that the double-triggering with codemirror is the rationale for the approach you selected because there is where the problem is.

Also, I don't think it's an acceptable UX to have two shortcuts for the same thing, one that may overlap with system shortcuts but is not configurable and another that is.  Not that I have a problem with multiple shortcuts for the same thing in general, but it seems like we're not really approaching this with the UX in top of mind.


Here's what I'd suggest:
1. configure the app with the native shortcut (`Ctrl`/`Cmd`+`/`) by default, just like it is now
1. configure CodeMirror to use this shortcut
1. ensure that changing the shortcut in Insomnia will change CodeMirror mapping as well (with the original shortcut it is today no longer functional)

filfreire:
Closing this for now, will revisit it if there's time in the future.

PR #4793, opened 2022-05-18
chore(deps) updates (all but insomnia) to use @jest/globals

dimitropoulos:
This work was scoped out of the [recent PR](https://github.com/Kong/insomnia/pull/4783) to move to jest 28 for the sake of time and complexity.  I am therefore following up with it here.

There are enough tests in the insomnia package that I opted to do all _but_ that package in this PR and hopefully get this PR merged quickly then move on to just the insomnia package as to reduce rebase-workload.

for anyone interested to learn more about `@jest/globals` https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44365 is a decent primer.  the TLDR; is that it allows us to use jest without globals (which, for us, means that we can't then accidentally call jest globals from our non-test source code).


marckong (approved):
LGTM!

jackkav (approved):
Take a bow :clap:

PR #4800, opened 2022-05-19
fix: prettify graphql query variables

SethFalco:
Before the "Prettify GraphQL" button only prettified the query, but not the variables.
Now the variables are prettified too! (While respecting the tab/indent-size settings!)

I'm honestly not really sure what it was doing with the variables before. 🤔
As in, it was calling prettify on the JSON before as well, but it just wasn't updating the object in the GUI. Not sure if that was a bug or maybe I'm missing something?

### Before
old

### After
new

---

On the side this also makes 2 minor edits in the context of this PR:
* Trim whitespace on one of the fixtures.
* Update the link to JSON Lint (and use the correct capitalization) as the domain expired and was bought by someone else it seems. 🤔

### DIFF
--- a/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
+++ b/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
@@ -65,7 +65,7 @@ const matchesOperation = (operationName: string | null | undefined) => ({ name }
 
 interface GraphQLBody {
   query: string;
-  variables?: Record<string, any>;
+  variables: string;
   operationName?: string;
 }
 
@@ -101,7 +101,8 @@ export class GraphQLEditor extends PureComponent<Props, State> {
   _disabledOperationMarkers: TextMarker[] = [];
   _documentAST: null | DocumentNode = null;
   _isMounted = false;
-  _queryEditor: null | CodeMirror.Editor = null;
+  _queryEditor:  null | CodeMirror.Editor = null;
+  _queryVariableEditor: null | CodeMirror.Editor = null;
   _schemaFetchTimeout: NodeJS.Timeout | null = null;
 
   constructor(props: Props) {
@@ -279,6 +280,10 @@ export class GraphQLEditor extends PureComponent<Props, State> {
     const { query, variables, operationName } = this.state.body;
 
     this._handleBodyChange(query, variables, operationName);
+  }
+
+  _handleQueryVariableEditorInit(codeMirror: CodeMirror.Editor) {
+    this._queryVariableEditor = codeMirror;
   }
 
   async _fetchAndSetSchema(rawRequest: Request) {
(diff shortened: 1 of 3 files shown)


SethFalco:
Actually, just converted this to a draft.
After further testing, this is flawed. 🤔 

If the query variables editor is modified, so it's malformed, then prettify is clicked, it reverts it back to the last valid input.

SethFalco:
Alright, I believe we should be all good now!
Tests are passing, formatting works, doesn't erase user input, JSON is still being validated, and GraphQL related errors are still displayed on hover!

I'm not too sure on how to use Nunjucks in Insomnia to be honest, everything I do leads to malformed tags. (Even on the stable version.) However, it does format at least. ^-^'

chabou:
I would love to have variables prettified too! ❤️

dimitropoulos:
let's look at this on the stream this coming week: come join us on the #stream channel over on the Insomnia community slack: https://chat.insomnia.rest.  The next one will be 10:30am ET on Tuesday May 24th.

SethFalco:
Since the stream was cancelled, any chance of getting feedback here, or shall I just wait for next week?

There's no urgency for the PR, but at the same time just want to make sure things are cool if you have any qualms with it. 👍🏾

filfreire:
Hi @SethFalco, sorry for yesterday's late notice.

We should be able resume the streams and to look at it next week on the next stream 🙇

dimitropoulos:
very sorry.  it was our first cancellation (I was super sick yesterday with the regular flu (of all things).  we're 15/16 weeks, now, so hopefully that means that next week should be all set.  If it's ok with you, I'd prefer to wait until the steam on 5/31 just because this kind of thing is so so much easier to work through and explain.

I don't think I have any qualms with it!  Just gotta figure out what the best UX should be :)

Right now I absolutely expect this PR to get merged in some form, for what it's worth.

SethFalco:
Absolutely no problem at all! Just wanted to know the plan!
High hopes for next week then. \\(^-^)/

SethFalco:
Ahh, as a small heads-up, I can't make it to next week's stream live, so I'll just watch the VOD later to hear what's up!

packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -573,10 +586,11 @@ export class GraphQLEditor extends PureComponent<Props, State> {
     }
 
     const query = obj.query || '';
-    const variables = obj.variables || null;
+    const variables = obj.variables || '';
```
gatzjames, line 593: Switching this to empty string results in double quotes "" being set as a default value to variables.
This is not ideal for users since the variables editor expects the user to define a JSON object instead of a string and makes the query invalid.

image

gatzjames:
@SethFalco great work!
I did some testing and found some issues we could look at the stream:
- The default value for variables being an empty string breaks new queries
- There is a flash of content when the variables editor contains template variables.

gatzjames (commented):
@SethFalco great work!
I did some testing and found some issues we could look at the stream:

The default value for variables being an empty string breaks new queries
There is a flash of content when the variables editor contains template variables.

packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
SethFalco: I should've tested a new query, sorry about that. 🤦🏾 
I'll look into it and fix both these issues on Saturday!

dimitropoulos:
(by the way, here's what I tested with on the stream)

```
https://rickandmortyapi.com/graphql
```

```graphql
query MyQuery($name: String!) {
  characters(page: 2, filter: { name: $name }) {
    info {
      count
    }
    results {
      name
    }
  }
  location(id: 1) {
    id
  }
  episodesByIds(ids: [1, 2]) {
    id
  }
}

query MyQuery2($name: String!) {
  characters(page: 2, filter: { name: $name }) {
    info {
      pages
    }
  }
}

```

```json
{
  "name": "rick"
}
```

SethFalco:
Thanks for the heads-up! I'll review this and the stream on Friday, if you don't already fix the branch and merge it yourself anyway. (Thanks for looking at it meanwhile!)

Sorry for being a bit slow there, work was doing an on-site last week, so I was away from my PC for a while, and now I'm sick.

dimitropoulos:
no worries at all! :)

jackkav:
Closing, stale.

PR #4804, opened 2022-05-20
Remove unused end trial modal

DMarby:
This is an unused modal that can only be triggered from the API, which currently never occurs, as we have other means to alert of trial ends/etc. It's also using an API endpoint which is being removed.


marckong (approved):
Wouldn't say no to getting rid of the dead code!

dimitropoulos (approved):
double checked and didn't see anything missed.  great!

PR #4809, opened 2022-05-20
docs: update links and minor typo

SethFalco:
* Makes links consistent, you used relative with no `./` in front most of the time, so I just did that everywhere.
* Makes it so links should resolve in the IDE/code-editor if in one, or the GitHub page if on GitHub. (Only exception is `/issues` which will not resolve to GitHub correctly even if clicks in a code-editor.
* Fixes 2 typos `Codemirror` → `CodeMirror` and `criticl` → `critical`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ Good pull requests (patches, improvements, new features) are a fantastic help. T
 
 **Please ask first** before embarking on any significant pull request (e.g. implementing features, refactoring code, porting to a different language), otherwise, you risk spending a lot of time working on something that might not get accepted into the project.
 
-If you plan on helping with development, please read the [Develop Insomnia](./README.md#develop-insomnia) section of the README for how to get started. The [Development Overview](./DEVELOPMENT.md) may also prove useful for a general overview of the application architecture.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -34,7 +34,7 @@ Good pull requests (patches, improvements, new features) are a fantastic help. T
 
 **Please ask first** before embarking on any significant pull request (e.g. implementing features, refactoring code, porting to a different language), otherwise, you risk spending a lot of time working on something that might not get accepted into the project.
 
-If you plan on helping with development, please read the [Develop Insomnia](./README.md#develop-insomnia) section of the README for how to get started. The [Development Overview](./DEVELOPMENT.md) may also prove useful for a general overview of the application architecture.
+If you plan on helping with development, please read the [Develop Insomnia](README.md#develop-insomnia) section of the README for how to get started. The [Development Overview](DEVELOPMENT.md) may also prove useful for a general overview of the application architecture.
```
juzhiyuan, line 37: Is it necessary to do this change?

SethFalco: Nope, this is just for consistency with the README.
I'm the one that PR-ed that line in recently, but during this clean-up, noticed that most other links don't prefix `./` to relative URLs.

I prefer `./` though, which is why they're there in the first place.

filfreire (approved):

PR #4817, opened 2022-05-26
INS-1501: Sentry integration.

johnwchadwick:
This is a basic Sentry integration for Insomnia, to send error data.

As far as privacy goes, the Sentry integration is designed to respect the user's existing data collection consent. It does not intentionally send or retain any PII with the error data, and only retain data sent to Sentry for up to 90 days. For more information on Sentry's security and privacy posture, visit their [Security](https://sentry.io/security/) page.

When data collection is disabled, Sentry will still be initialized, but it will only retain information locally in the app data folder.

We have [opened an issue](https://github.com/getsentry/sentry-electron/issues/489) with Sentry to try to provide more "first class" support for explicitly operating offline. Until then, we have a custom transport that uses a small workaround to block all outgoing requests. I do not believe any requests are being made outside of this, but please note that you can't use the typical Web Inspector to view these requests, because they happen on the main thread using Electron's net package.

### DIFF
--- a/packages/insomnia/src/account/fetch.ts
+++ b/packages/insomnia/src/account/fetch.ts
@@ -43,7 +43,14 @@ export async function getJson<T = any>(path: string, sessionId: string | null): 
   return response;
 }
 
-async function _fetch<T = any>(method: 'POST' | 'PUT' | 'GET', path: string, obj: unknown, sessionId: string | null, compressBody = false, retries = 0): Promise<T | string> {
+async function _fetch<T = any>(
+  method: 'POST' | 'PUT' | 'GET',
+  path: string,
+  obj: unknown,
+  sessionId: string | null,
+  compressBody = false,
+  retries = 0
+): Promise<T | string> {
   if (sessionId === undefined) {
     throw new Error(`No session ID provided to ${method}:${path}`);
   }
--- a/packages/insomnia/src/account/fetch.ts
+++ b/packages/insomnia/src/account/fetch.ts
@@ -15,19 +15,35 @@ export function onCommand(callback: Function) {
   _commandListeners.push(callback);
 }
 
-export async function post(path, obj, sessionId, compressBody = false) {
-  return _fetch('POST', path, obj, sessionId, compressBody);
+export async function post<T = any>(path: string, obj: unknown, sessionId: string | null, compressBody = false): Promise<T | string> {
+  return _fetch<T>('POST', path, obj, sessionId, compressBody);
 }
 
-export async function put(path, obj, sessionId, compressBody = false) {
-  return _fetch('PUT', path, obj, sessionId, compressBody);
+export async function postJson<T = any>(path: string, obj: unknown, sessionId: string | null, compressBody = false): Promise<T> {
+  const response = await post<T>(path, obj, sessionId, compressBody);
+  if (typeof response === 'string') {
+    throw new Error('Unexpected plaintext response');
+  }
+  return response;
 }
 
-export async function get(path, sessionId) {
-  return _fetch('GET', path, null, sessionId);
+export async function put<T = any>(path: string, obj: unknown, sessionId: string | null, compressBody = false): Promise<T | string> {
+  return _fetch<T>('PUT', path, obj, sessionId, compressBody);
 }
 
-async function _fetch(method, path, obj, sessionId, compressBody = false, retries = 0) {
+export async function get<T = any>(path: string, sessionId: string | null): Promise<T | string> {
+  return _fetch<T>('GET', path, null, sessionId);
+}
+
+export async function getJson<T = any>(path: string, sessionId: string | null): Promise<T> {
+  const response = await get<T>(path, sessionId);
+  if (typeof response === 'string') {
+    throw new Error('Unexpected plaintext response');
+  }
+  return response;
+}
+
+async function _fetch<T = any>(method: 'POST' | 'PUT' | 'GET', path: string, obj: unknown, sessionId: string | null, compressBody = false, retries = 0): Promise<T | string> {
   if (sessionId === undefined) {
     throw new Error(`No session ID provided to ${method}:${path}`);
   }
@@ -59,7 +75,7 @@ async function _fetch(method, path, obj, sessionId, compressBody = false, retrie
     config.headers['X-Session-Id'] = sessionId;
   }
 
-  let response;
+  let response: Response | undefined;
 
   const url = _getUrl(path);
 
--- a/packages/insomnia/src/common/sentry.ts
+++ b/packages/insomnia/src/common/sentry.ts
@@ -1,0 +1,82 @@
+import * as Sentry from '@sentry/electron';
+import type * as SentryElectronMain from '@sentry/electron/main';
+import type { SentryRequestType, Transport, TransportClass } from '@sentry/types';
+
+import { getAccountId, onLoginLogout } from '../account/session';
+import { database as db } from '../common/database';
+import * as models from '../models/index';
+import { isSettings } from '../models/settings';
+import { getAppEnvironment, getAppVersion } from './constants';
+
+let enabled = false;
+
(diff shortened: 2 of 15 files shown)


packages/insomnia/src/common/sentry.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Initialize Sentry. Sentry needs to be initialized as soon as possible; it
+// will fail if the Electron ready event fires before it is initialized.
+Sentry.init({
+  dsn: 'https://aaec2e800e644070a8daba5b7ad02c16@o1147619.ingest.sentry.io/6311804',
+  environment: getAppEnvironment(),
+  release: getAppVersion(),
+  transport,
+});
```
jackkav: There is are a number of benefits to clearing a path to separate main and renderer into different packages and enabling contextIsolation. To that end it would be advantageous to split up polymorphic files into main and renderer files. This would help take something off the pile of work to do unpicking at a later date. I'm happy to help out!

johnwchadwick: Sentry for Electron is "isomorphic" in a sense, in that it detects which environment it's running in and automatically IPCs back to the main/browser thread from renderer threads. Therefore, for the most part, the `common/sentry.ts` file is isomorphic.

That said, I gave a shot at splitting things up anyways. It seems to work, and it does make it clearer for the couple of things that aren't fully isomorphic. (And we don't have to care about the kill switch for the renderer thread anyways.)

packages/insomnia/src/account/fetch.ts
```diff
@@ -15,19 +15,35 @@ export function onCommand(callback: Function) {
   _commandListeners.push(callback);
 }
 
-export async function post(path, obj, sessionId, compressBody = false) {
-  return _fetch('POST', path, obj, sessionId, compressBody);
+export async function post<T = any>(path: string, obj: unknown, sessionId: string | null, compressBody = false): Promise<T | string> {
```
marckong, line 18: Would you be able to help me understand why we are assigning any here?

marckong (approved):
LGTM

packages/insomnia/src/common/sentry.ts
```diff
@@ -0,0 +1,9 @@
+import type { ElectronOptions } from '@sentry/electron';
+
+import { getAppEnvironment, getAppVersion } from './constants';
+
+export const SENTRY_OPTIONS: Partial<ElectronOptions> = {
+  dsn: 'https://aaec2e800e644070a8daba5b7ad02c16@o1147619.ingest.sentry.io/6311804',
+  environment: getAppEnvironment(),
```
dimitropoulos, line 7: ok, got it.  so just to double-check: sending all events at all times to the same dsn is preferable to having a "prod" and "staging/whatever" place to send them (like we do with segment or analytics).

Is that right?

packages/insomnia/src/common/sentry.ts
```diff
@@ -0,0 +1,9 @@
+import type { ElectronOptions } from '@sentry/electron';
+
+import { getAppEnvironment, getAppVersion } from './constants';
+
+export const SENTRY_OPTIONS: Partial<ElectronOptions> = {
+  dsn: 'https://aaec2e800e644070a8daba5b7ad02c16@o1147619.ingest.sentry.io/6311804',
```
dimitropoulos: it's not a huge huge deal: but we don't do anything else like this in the app this way.  what we would do is put the `dsn` in `config.json` and then make a selector in `./constants` like you have here.  If we do that, then we don't need this file at all, because then we could remove the indirection and (duplicate) `environment` and `release` in `src/main/sentry.ts` and `src/ui/sentry.ts`.

I'd be happy to do that for you if you want or if what I wrote isn't clear.  It can also stay this way... no biggie.  Just a little strange given everything else in the codebase like this (including what we're adding for GitLab soon).

packages/insomnia/src/main/sentry.ts
```diff
@@ -0,0 +1,48 @@
...
+import type { SentryRequestType } from '@sentry/types';
+
+import { database as db } from '../common/database';
+import { SENTRY_OPTIONS } from '../common/sentry';
+import * as models from '../models/index';
+import { isSettings } from '../models/settings';
+
+let enabled = false;
+
+// Watch setting for changes. This must be called after the DB is initialized.
```
dimitropoulos: We use TSDoc for this kind of thing:

| BEFORE | AFTER |
| - | - |
| Screenshot_20220527_144947 | Screenshot_20220527_144912 |

see: d696e34e8ed0bf772f03b5bedc2a526411c988ef

packages/insomnia/src/main/sentry.ts
```diff
@@ -0,0 +1,48 @@
...
+        enabled = doc.enableAnalytics;
+      }
+    }
+  });
+}
+
+// TODO(johnwchadwick): We are vendoring ElectronOfflineNetTransport just to be
+// able to control whether or not sending is allowed, because we don't have a
+// choice right now. We should work with the upstream library to get similar
+// functionality upstream. See getsentry/sentry-electron#489.
```
dimitropoulos: minor nit: we don't do these kinds of comments in this project (for a few reasons, but one of which is to avoid: Screenshot_20220527_144606)

see: d696e34e8ed0bf772f03b5bedc2a526411c988ef

packages/insomnia/src/account/fetch.ts
johnwchadwick: The typing is being introduced incrementally. If we didn't default to `any` we'd need to update every single callsite, but we don't have proper typings for most of them yet.

packages/insomnia/src/common/sentry.ts
johnwchadwick: Yep

packages/insomnia/src/ui/index.tsx
```diff
@@ -56,18 +59,6 @@ if (isDevelopment()) {
   window.db = db;
 }
 
-// Catch uncaught errors and report them
-if (window && !isDevelopment()) {
-  window.addEventListener('error', e => {
-    console.error('Uncaught Error', e.error || e);
-    trackSegmentEvent(SegmentEvent.criticalError, { detail: e?.message });
```
DMarby, line 63: Might make sense to remove `criticalError` from `SegmentEvent` as well now that it's no longer used?

DMarby (approved):

packages/insomnia/src/ui/index.tsx
johnwchadwick: Yep, done.

packages/insomnia/src/main/sentry.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  db.onChange(async changes => {
+    for (const change of changes) {
+      const [event, doc] = change;
+      if (isSettings(doc) && event === 'update') {
+        enabled = doc.enableAnalytics;
+      }
+    }
+  });
+}
```
jackkav, line 27: I'd prefer this wasn't a change listener, but its not too trivial to trigger it over ipc on setting change. It's also used this way to get the first app start event in segment, so it would be out of scope to address in this PR. Just a note.

jackkav (approved):
Non-blocking: I'm curious about the generic typed fetch methods and how they fit into this PR scope.

johnwchadwick:
> Non-blocking: I'm curious about the generic typed fetch methods and how they fit into this PR scope.

I didn't want to write new code that did not have typing, so I plumbed the typings for the session all the way through. Plus, I already had experience with this code because we actually have a near-duplicate of it on the (closed source) website code, and I had converted that to strict TypeScript already.

packages/insomnia/src/main/sentry.ts
johnwchadwick: Yep, this follows the same pattern as the Segment integration. We should change it for both of them, probably.

PR #4819, opened 2022-05-27
Spike: Adding mock WS server for the smoke tests (INS-1536).

filfreire:
Related ticket: INS-1536

This PR will add a WS server to our smoke-tests mock server that we'll be able to use for automation/smoke tests (similarly to what we did for gRPC and for oauth).

## Todo

- [x] try express-ws (problems - adding an unnecessary dev dependency to the smoke tests)
- [ ] try plain ws
- [ ] extract WS setup into it's separate .ts file and include it on `server/index.ts`

## Notes:

- We'll likely want to echo Headers, Params, and any other info we might need for debugging purposes
- We might want an endpoint where, once we send a message, we get back X amount of responses sequentially, spaced out in small random time intervals
- ...


filfreire:
Closing this one for now. Once we have Websockets implemented we can re-add the WS express server alongside a Playwright check.

PR #4821, opened 2022-05-27
FIX: check if bubble is a child of Tooltip container before removing it

### LINKED ISSUE #3986: Render Failure: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.

### Insomnia Version

2021.5.1

Windows

### Operating System Version

Windows 10 Version 20H2

### Installation method

Windows install

_No response_

### Expected Behavior

When swapping between gRPC and REST calls I expect them to render properly.

### Actual Behavior

The entire section fails to render until insomnia is restarted

### Reproduction Steps

I am having issues reproducing this after first opening insomnia but am running into it constantly as I work moving old rest calls into gRPC calls 

1. Open an existing collection with post REST calls and gRPC calls
2. Use software for a little while
3. Right click collection and Create a new gRPC request
4. Name Request and click create
4. select proto file and click save
5. Click on a separate POST or GET request in the collection and error will pop up

### Additional Information

insomniaerror

gabrielrra:
Closes #3986

There was a bug when removing a node from DOM using `removeChild` because it was trying to remove a node that was not a child.
I added a check to see if the Tooltip's bubble is a child of its container.
On issue #3986 there are a little more details about the bug

changelog(Fixes): Fixed an error when swapping between gRPC and HTTP requests


victorradael:
This error is really very inconvenient, the people of my company will be very happy with this correction.

marckong:
Hi @gabrielrra thank you for contributing! It's awesome to see contributors like yourself helping insomnia keep going. I reviewed your PR and seems your fix is good. Just one thing that we would need to check is how this would affect the other tooltips. We are now removing the tooltip element conditionally based on _getContainer() result, which seems to pretty much guarantee the existence of the container element, but we may want to be safe, so let us double check on it. Thank you so much again for helping us improve it!

marckong:
@victorradael Thank you for adding another voice here!

gabrielrra:
Thanks for checking it out @marckong. Glad I could help

filfreire (approved):
LGTM

marckong:
I tested this against the latest develop branch. This change sufficiently fixes the issue addressed at https://github.com/Kong/insomnia/issues/3986 whereas the develop branch still has the bug occur. Tooltip seems to behave fine in other areas too.

marckong (approved):
LGTM!

marckong:
Thought I had after reviewing and learning about tooltip component is that we could possibly turn this into a functional component. I was initially suggesting a unit test to cover this case, but it may be better to do so when we refactor this component into functional component.

PR #4841, opened 2022-06-07
refactor: add segment tracking for git sync

marckong:
- adding a oauth provider name type
- adding a util function to return oauth provider naem
- modifying the segments for git flow to add provider name

for github,
image

for gitlab,
image

### DIFF
--- a/packages/insomnia/src/sync/git/utils.ts
+++ b/packages/insomnia/src/sync/git/utils.ts
@@ -1,5 +1,6 @@
 import { AuthCallback, AuthFailureCallback, AuthSuccessCallback, GitAuth, MessageCallback } from 'isomorphic-git';
 
+import { OauthProviderName } from '../../models/git-repository';
 import type { GitCredentials } from './git-vcs';
 import { getAccessToken as getGitHubAccessToken } from './github-oauth-provider';
 import { getAccessToken as getGitlabAccessToken, refreshToken as refreshGitlabToken } from './gitlab-oauth-provider';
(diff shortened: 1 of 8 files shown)


packages/insomnia/src/sync/git/utils.ts
```diff
@@ -95,6 +96,14 @@ const onAuth = (credentials?: GitCredentials): AuthCallback => (): GitAuth => {
   };
 };
 
+export const getOauth2FormatName = (credentials?: GitCredentials | null): OauthProviderName | undefined => {
+  if (credentials && 'oauth2format' in credentials) {
+    return credentials.oauth2format;
+  }
+
+  return undefined;
```
marckong: Or we could return null but if we return null then we have to null check in the consumer functions.

johnwchadwick: If you want it to look a bit more ergonomic, a `return;` also returns undefined.

marckong: Good point! was initially returning null and just dumb-replaced null with undefined during a zoom call lol

wdawson (approved):
Looks good to me, thanks @marckong ! Did we also confirm that these events are firing in all cases where a user can do a git clone (e.g. from the dashboard, from an existing document, etc.)?

filfreire (approved):

filfreire (approved):
Aproved. `type-checks` are failing but I think we can merge once you fix that failure.

PR #4842, opened 2022-06-07
Improve typing in various UI components.

johnwchadwick:
Removes a bunch of implicit `any`s and `Function`s to replace them with more specific types. Some TODOs are added where things seem completely broken. This PR is not intended to change any behavior or fix any bugs.

### DIFF
--- a/packages/insomnia/src/ui/components/key-value-editor/row.tsx
+++ b/packages/insomnia/src/ui/components/key-value-editor/row.tsx
@@ -2,7 +2,7 @@
 import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import classnames from 'classnames';
 import React, { forwardRef, ForwardRefRenderFunction, PureComponent } from 'react';
-import { ConnectDragPreview, ConnectDragSource, ConnectDropTarget, DragSource, DropTarget } from 'react-dnd';
+import { ConnectDragPreview, ConnectDragSource, ConnectDropTarget, DragSource, DropTarget, DropTargetMonitor } from 'react-dnd';
 import ReactDOM from 'react-dom';
 
 import { AUTOBIND_CFG } from '../../../common/constants';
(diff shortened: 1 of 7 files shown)


marckong (approved):

packages/insomnia/src/ui/components/key-value-editor/row.tsx
```diff
@@ -543,27 +552,27 @@ const dragSource = {
...
-      props.onMove(monitor.getItem().pair, props.pair, 1);
+      props.onMove?.(monitor.getItem().pair, props.pair, 1);
     } else {
-      props.onMove(monitor.getItem().pair, props.pair, -1);
+      props.onMove?.(monitor.getItem().pair, props.pair, -1);
     }
   },
 
-  hover(_props, monitor, component) {
+  hover(_props: Props, monitor: DropTargetMonitor, component: any) {
```
marckong, line 575: is it ComponentType here?

marckong:
```js
React.ComponentType<Props>
```

johnwchadwick: This is the verbatim typing from React DND. We might need to upgrade it or improve typings upstream if we want to have this field be typed.

filfreire (approved):

DMarby (approved):

dimitropoulos (commented):
So I noticed we had a few single letter variable names that could fix and I set a 15 minute timer and did the ones I could find.... and it turned out to be a lot.  I think I restricted scope pretty well, but all the same if you'd like me to pull my commit out into its own PR please let me know and I'll do so immediately without question.  Don't want to hijack your PR, haha.

johnwchadwick:
> So I noticed we had a few single letter variable names that could fix and I set a 15 minute timer and did the ones I could find.... and it turned out to be a lot. I think I restricted scope pretty well, but all the same if you'd like me to pull my commit out into its own PR please let me know and I'll do so immediately without question. Don't want to hijack your PR, haha.

Yeah, I kind of do want you to pull that out of this. For one thing, I'm not even sure you should bother with it. Changing `e` to `event` or `error` when it's very obvious based on context doesn't greatly improve the code quality, if at all. It feels a bit like changing `i` to `iterator`.

It also has nothing to do with typings and balloons this PR from covering 7 files to 132.

dimitropoulos:
say no more, apologies.  I see you already removed it: sounds great.  I blame the LSP rename for making this so easy, haha.

dimitropoulos (approved):

johnwchadwick:
Not really a big deal, but while I absolutely do not mind when someone improves my PR, I do think that adding significant non-overlapping scope to an existing PR is generally undesirable. So I would prefer if you did not do that :P

PR #4845, opened 2022-06-07
test: adding unit tests to AccountDropdown

marckong:
@kreosus and I have done this exercise together to add some unit tests to AccountDropdown component.
- changed from ```dropdownsContainer``` to ```document.querySelector('#dropdowns-container')```. dropdownsContainer was not available during the run time for testing.
- added unit tests coverages for some of the component behaviours

*NOTE*
Initially, we added test.todo('calls session.logout callback when "Logout" button is clicked') to continue on the test path but will skip this one for now.


packages/insomnia/src/ui/components/dropdowns/account-dropdown/account-dropdown.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+  test('renders "Log In" label without paid feature ads', async () => {
+    const container = getDropdownContainer();
+    const store = await createMockStore({ disablePaidFeatureAds: true });
+    render(
+      <Provider store={store}>
+        <AccountDropdownButton />
+      </Provider>, { container }
+    );
+    expect(screen.getByText('Log In')).toBeDefined();
+    expect(screen.queryByText('Upgrde Now')).toBeNull();
```
DMarby: Typo here I think?

```suggestion
    expect(screen.queryByText('Upgrade Now')).toBeNull();
```

marckong: Good catch! Thanks for this comment, I discovered the redux wire up wasn't done correctly in the testing. Thank you so much.

filfreire (approved):

filfreire:
@marckong I think we could merge this one after fixing lint issues.

packages/insomnia/src/ui/index.tsx
```diff
@@ -40,6 +40,7 @@ document.title = getProductName();
   // Create Redux store
   const store = await initStore();
 
+  console.log(store);
```
dimitropoulos:
```suggestion
```

packages/insomnia/src/ui/components/dropdowns/account-dropdown/account-dropdown.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+    entities: {
+      settings: [settings],
+    },
+  });
+};
+
+describe('<AccountDropdownButton />', () => {
+  const getDropdownContainer = () => {
+    const container = document.createElement('div');
+    container.setAttribute('id', 'dropdowns-container');
```
dimitropoulos: how is this any different than data-testid?

dimitropoulos: also setting `id`s is problematic because you can't have multiple elements on a page with the same `id`

packages/insomnia/src/ui/components/base/dropdown/dropdown.tsx
```diff
@@ -445,7 +445,7 @@ export class Dropdown extends PureComponent<DropdownProps, State> {
             </div>
           </div>,
           // @ts-expect-error -- TSCONVERSION
-          dropdownsContainer,
+          document.querySelector('#dropdowns-container'),
```
dimitropoulos: this feels like a "magic string".  if we were to do the usual refactor for this (like https://twitter.com/kentcdodds/status/841429439380615169) we'd see that it's details for testing that leaked out into shipped implementation, wouldn't we?

packages/insomnia/src/ui/components/dropdowns/account-dropdown/account-dropdown.test.tsx
marckong: 'id' is what is used to select the container in the dropdown component. I find this to be trivial detail to be in the testing code.
FYI -
https://github.com/Kong/insomnia/blob/a58eb9fdd7d196bdc708f858c5a6368f03dda478/packages/insomnia/src/ui/components/base/dropdown/dropdown.tsx#L14

packages/insomnia/src/ui/components/base/dropdown/dropdown.tsx
marckong: This was not related to testing. If we were talking about data-testid, we can perhaps parse it out during the build time but I am not sure how this is related to this change.

marckong: Addressed

PR #4848, opened 2022-06-08
chore: single letter variables

dimitropoulos:
I noticed some single letter variable names being added (in types) in https://github.com/Kong/insomnia/pull/4842 and set a 15 minute timer to remove some... and quite a few got touched.

This is a quite focused scope, so there should be no logical changes from this PR whatsoever.

We had been trying for some time to fix all of these but never got around to a PR like this (we did a lot of them in the initial typescript conversion, but ran out of energy to do them all).

There may be more, but I set a 15 minute timebox on this commit, so progress is progress.


marckong (approved):
LGTM! more clear and indicative!

PR #4852, opened 2022-06-10
Improve typing in some API code and components.

johnwchadwick:
This is just another batch of fixes for implicit any and other typing issues across Insomnia. The improvements done to API and SRP were already present in the app.insomnia.rest codebase. The remaining fixes are mostly minor issues. Some of them appear to just be a result of accidental any occurring because we still have `noImplicitAny` disabled.


marckong (approved):
LGTM. Thank you for doing this work. It will only help us move forward safer!

DMarby (approved):

PR #4857, opened 2022-06-13
Allow using ID Token to retrieve schema

### LINKED ISSUE #3013: GraphQL schema documentation doesn't load if an OAuth2 access token tag is used.
**Describe the bug**
If a GraphQL request is configured to use OAuth 2 authentication via an OAuth 2 token tag, then Insomnia will always fail to load the schema documentation. 

It doesn't matter if the token has been refreshed and is valid, the schema documentation loading request will always fail. An error message appears below the request body which reads "Failed to fetch schema: No OAuth 2.0 tokens found for request".

It appears to be caused specifically by the tag that is set on the request header. If I manually paste the token value into the Authorization header, then the schema documentation will start to load successfully.

It's also worth noting that the GraphQL request itself _does_ work when the OAuth 2.0 token tag is used. It's just the schema documentation fetching that seems to be affected.

**To Reproduce**
1. Create a GraphQL request against an OAuth 2.0 protected GraphQL API
2. Configure OAuth2 authentication for the request. Make sure that a token has been successfully loaded.
3. Set the Authorization header of the request to use the "Request -> OAuth 2.0 Token - access token" tag
4. Make the GraphQL request and confirm that it works correctly.
5. Go back to the GraphQL body tab. Click schema -> refresh schema
6. Observe the error appears at the bottom of the tab
7. Go to the Headers tab and remove the OAuth2 token tag. Paste in the access token manually.
8. Go back and refresh the schema documentation again.
9. This time the schema will load correctly.

**Expected behavior**
Schema documentation should load if the request is configured with a valid access token. It should fail to load if that token is invalid for some reason.

**Screenshots**
Error message :
image

Authorization header config:
image

 - OS: Ubuntu
 - Installation Method: snap
 - App Version 2020.5.2

danelowe:
#3211 Allowed for using an identity token for auth rather than auth token. When using that feature, the GraphQL schema could not be fetched due to `throw new Error('No OAuth 2.0 identity tokens found for request');`. This commit moves the normalization of the request ID further up the chain so it can be used where this error would otherwise be thrown.

Closes #3013

### DIFF
--- a/packages/insomnia/src/models/o-auth-2-token.ts
+++ b/packages/insomnia/src/models/o-auth-2-token.ts
@@ -70,7 +70,7 @@ export function remove(token: OAuth2Token) {
 }
 
 export function getByParentId(parentId: string) {
-  return db.getWhere<OAuth2Token>(type, { parentId });
+  return db.getWhere<OAuth2Token>(type, { parentId: normaliseRequestId(parentId) });
 }
 
 export async function getOrCreateByParentId(parentId: string) {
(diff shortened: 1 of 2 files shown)


filfreire:
Hi @danelowe thanks for submitting this PR!

This is a bit of a rough problem to solve, we'll to get back to this issue asap. In the meantime you may want to check out the recording of last week's [Insomnia Stream](https://www.youtube.com/watch?v=GvS1z5w-uPE) where @jackkav and @dimitropoulos have looked at this PR.

We'll also have a stream later today, feel free to drop by. You can find more details on our Slack (chat.insomnia.rest) on the `#stream` channel.

PR #4858, opened 2022-06-13
Fix #4856 UI spacing on deploy to dev portal modal

### LINKED ISSUE #4856: Minor UI Misalignment  -  Deployment to Dev Portal Pop-up Box
### Expected Behavior

A message will shows that the yaml document is successfully deployed and available at the Dev Portal.

### Actual Behavior

The message intent is correct, just that there is a slight UI spacing misalignment as shown below in the screenshot.

Screenshot 2022-06-13 at 3 52 52 PM

### Reproduction Steps

1. From the Insomnia Dashboard, click the three dot menu on a Document. 
2. Select Deploy to Dev Portal and a modal will open. 
3. Enter the following information.
    - Kong API URL: http://localhost:8001 (admin api path)
    - Kong Workspace Name: Workspace name
    - Kong RBAC Token (optional): Leave it Blank if it is not set

### Additional Information

I took a look at the code, and i think the its just a matter of edit in the following file.  Didn't have time to set-up my local env properly for development,so i thought i should highlight here first

plugins/insomnia-plugin-kong-portal/src/deploy-to-portal.tsx

```
   <div className="pad">
            <p className="no-pad no-margin-top">
              The Document is now available on
              {kongPortalLegacyMode ? (
                'Dev Portal'
              ) : (
                <a href={kongPortalUrl}>Dev Portal</a>
              )}
            </p>
            <div>
              <Button data-close-modal="true">Close</Button>
            </div>
  </div>
```

### Insomnia Version

2022.4.0

macOS

### Operating System Version

macOS Monterey 12.4

### Installation method

Downloaded from insomnia.rest

NA

filfreire:
Fix for #4856 

Before:

Screenshot 2022-06-13 at 10 35 28

After:

Screenshot 2022-06-13 at 11 10 58

### DIFF
--- a/plugins/insomnia-plugin-kong-portal/src/deploy-to-portal.tsx
+++ b/plugins/insomnia-plugin-kong-portal/src/deploy-to-portal.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable no-trailing-spaces */
 import { AxiosError, AxiosRequestConfig } from 'axios';
 import type ReactLib from 'react';
 import { type SyntheticEvent } from 'react';
@@ -431,7 +432,7 @@ export function getDeployToPortalComponent(options: {
         return (
           <div className="pad">
             <p className="no-pad no-margin-top">
-              The Document is now available on
+              The Document is now available on 
               {kongPortalLegacyMode ? (
                 'Dev Portal'
               ) : (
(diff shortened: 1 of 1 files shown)


marckong (approved):
It looks good to me!

There are other ways to achieve this as well. If we use space in the text, it may be a good idea to check this following [css property](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) and its application in our app style sheet or plugin. It looks like there isn't any implication, so its all good in my observations.

Another way could be just adding this too.
```
&nbsp;
```

dimitropoulos (changes requested):
this is error prone because the intent is not clear and these kinds of changes can be forgotten or ignored.  please instead style the anchor tag to have a left margin.

yes, we do this kind of thing in places, but we shouldn't, and we're trying to make styling like this clearer.  if we're in a hurry we could do a `style` object directly on the anchor, and wrap the literal string with a span with the same style.

filfreire:
Closing this for now, will revisit it if there's time in the future.

PR #4860, opened 2022-06-13
Follow-up: rename shadowing variable r.

johnwchadwick:
This is a follow-up to #4853 to rename the shadowing variable `r`. Sorry!


dimitropoulos (approved):
seemed fine to me in the first place, but sure.

what will we rename the first `r` to when the time comes for it?  seems much better to me to just embrace that shadowing is a thing and live with it, you know..

all the same! it's no big thing so sure

PR #4864, opened 2022-06-14
Enable noImplicitAny.

johnwchadwick:
In general I've tried to avoid any changes that actually change the code output; this should almost only change the TypeScript typings. There is exactly one place where I intended for the code to change at all, and that is in `app.tsx:350`, where `activeRequest` is used instead of `this.props.activeRequest`, as the usage of the latter defeated type narrowing for seemingly no good reason (likely an accident.)

- In places where types are missing but the typing is sound and verifiable, proper typings are added.
- In places where keys are mapped over using `Object.keys`, casting is used to tell TypeScript that the `Object.keys` return value is a `keyof` the type. This can't be statically verified, but in theory it is still type sound.
- In places where keys are mapped over in an unsound or unverifiable way, `@ts-expect-error` is used.
- In places where soundness is violated due to the possibility of null or undefined, a non-null assertion is used, plus an eslint exception, plus a TODO. Some of these may be fixable with a null navigation operator, but it is out of scope for this PR.
- In places where types are missing and there is no easily inferred correct typing, an explicit `any` is inserted.


packages/insomnia/src/common/constants.ts
```diff
@@ -244,7 +244,7 @@ const previewModeMap = {
   [PREVIEW_MODE_SOURCE]: ['Source', 'Source Code'],
   [PREVIEW_MODE_RAW]: ['Raw', 'Raw Data'],
 };
-export const PREVIEW_MODES = Object.keys(previewModeMap);
+export const PREVIEW_MODES = Object.keys(previewModeMap) as (keyof typeof previewModeMap)[];
```
dimitropoulos, line 247: man, I hate that they put their foot down on this specific thing.  one of the benefits of using ramda is that you don't have to do stuff like this for `keys` and `values` and `map` and whatnot.

packages/insomnia/src/common/constants.ts
```diff
@@ -300,7 +300,7 @@ const authTypesMap = {
   [AUTH_AWS_IAM]: ['AWS', 'AWS IAM v4'],
   [AUTH_ASAP]: ['ASAP', 'Atlassian ASAP'],
   [AUTH_NETRC]: ['Netrc', 'Netrc File'],
-} as const;
```
dimitropoulos, line 303: to anyone else reviewing: usually this is very intentional that we use `as const` like this, but it doesn't seem to be necessary in this case so removing the `as const` is fine.   this is not exported by this file directly or indirectly and the usages below in `getAuthTypeName` just widen to `string` anyway.

this is the error you get without this change: 
image

packages/insomnia/src/common/constants.ts
```diff
@@ -382,15 +382,15 @@ export function getContentTypeName(contentType?: string | null, useLong = false)
...
+export function getAuthTypeName(authType: string, useLong = false) {
   if (authTypesMap.hasOwnProperty(authType)) {
     return useLong ? authTypesMap[authType][1] : authTypesMap[authType][0];
   } else {
     return '';
   }
 }
 
-export function getContentTypeFromHeaders(headers, defaultValue: string | null = null) {
+export function getContentTypeFromHeaders(headers: any[], defaultValue: string | null = null) {
```
dimitropoulos, line 395: the type here is actually `RequestHeader[] | ResponseHeader[]` but if you add it it surfaces more errors that will require logical changes, so seems not worth it right now in the scope of this PR.

image

packages/insomnia/src/jsonlint.d.ts
```diff
@@ -0,0 +1,2 @@
+declare module 'jsonlint-mod-fixed' {
```
dimitropoulos, line 1: to anyone else wondering: this npm is a peer dependency of codemirror linting functionality and we can essentially consider it "not our code" and it doesn't matter what the type is from our perspective.

packages/insomnia/src/network/certificate-url-parse.ts
```diff
@@ -17,7 +17,10 @@ export default function certificateUrlParse(url) {
   }
 }
 
-function _reinstateWildcards(string) {
+function _reinstateWildcards(string: string): string;
+function _reinstateWildcards(string: string | null): string | null;
+
```
dimitropoulos:
```suggestion
```

dimitropoulos: (looks like a typo extra line)

packages/insomnia/src/network/o-auth-1/get-token.ts
```diff
@@ -69,22 +69,18 @@ export default async function(
...
   }
 
   if (authentication.timestamp) {
-    // @ts-expect-error -- TSCONVERSION needs type widening
     requestData.data.oauth_timestamp = authentication.timestamp;
   }
 
   if (authentication.verifier) {
-    // @ts-expect-error -- TSCONVERSION needs type widening
     requestData.data.oauth_verifier = authentication.verifier;
```
dimitropoulos, line 84: nice.

packages/insomnia/src/preload.js
```diff
@@ -2,23 +2,23 @@ const { contextBridge, ipcRenderer } = require('electron');
 
 const main = {
   restart: () => ipcRenderer.send('restart'),
-  authorizeUserInWindow: options => ipcRenderer.invoke('authorizeUserInWindow', options),
-  setMenuBarVisibility: options => ipcRenderer.send('setMenuBarVisibility', options),
-  installPlugin: options => ipcRenderer.invoke('installPlugin', options),
-  curlRequest: options => ipcRenderer.invoke('curlRequest', options),
-  cancelCurlRequest: options => ipcRenderer.send('cancelCurlRequest', options),
-  writeFile: options => ipcRenderer.invoke('writeFile', options),
+  authorizeUserInWindow: (/** @type {any} */ options) => ipcRenderer.invoke('authorizeUserInWindow', options),
```
dimitropoulos, line 5: good thinking.  I wasn't aware we were typechecking `.js` files.  That's cool, though.  and necessary, considering we want to take the types here and expose them to users sooner than later.

packages/insomnia/src/sync/git/routable-fs-client.ts
```diff
@@ -1,6 +1,8 @@
 import * as git from 'isomorphic-git';
 import path from 'path';
 
+type Methods = 'readFile' | 'writeFile' | 'unlink' | 'readdir' | 'mkdir' | 'rmdir' | 'stat' | 'lstat' | 'readlink' | 'symlink';
```
dimitropoulos, line 4: wow this one is really helpful, I've wanted to do this before but never got to it.

packages/insomnia/src/ui/components/check-for-updates-button.tsx
```diff
@@ -23,14 +23,14 @@ export class CheckForUpdatesButton extends PureComponent<Props, State> {
     updateAvailable: false,
   };
 
-  _listenerCheckComplete(_e, updateAvailable: true, status: string) {
+  _listenerCheckComplete(_e: Electron.IpcRendererEvent, updateAvailable: true, status: string) {
```
dimitropoulos, line 26:
```suggestion
  _listenerCheckComplete(_event: Electron.IpcRendererEvent, updateAvailable: true, status: string) {
```

JUST KIDDING!

dimitropoulos: (seriously, just kidding, haha)

packages/insomnia/src/ui/components/codemirror/base-imports.ts
```diff
@@ -46,7 +46,7 @@ import 'codemirror/addon/lint/lint';
 /**/
 // for the code that uses this json parser, see https://github.com/codemirror/CodeMirror/blob/master/addon/lint/json-lint.js
 import * as jsonlint from 'jsonlint-mod-fixed';
-global.jsonlint = jsonlint;
+(global as any).jsonlint = jsonlint;
```
dimitropoulos: it's not worth chasing down because `jsonlint` is such a low priority but I was surprised to see this because we have

```ts

declare namespace NodeJS {
  interface Global {
    __DEV__: boolean;
    /** this is required by codemirror/addon/lint/json-lint */
    jsonlint: any;
    /** this is required by codemirror/addon/lint/yaml-lint */
    jsyaml: any;
  }
}
```
in `global.d.ts`.

oh well.  I don't care.

packages/insomnia/src/ui/components/editors/body/body-editor.tsx
```diff
@@ -101,7 +101,7 @@ export const BodyEditor: FC<Props> = ({
           Do you want set the <span className="monospace">Content-Type</span> header to{' '}
           <span className="monospace">{newContentType}</span>?
         </p>,
-        onDone: saidYes => {
+        onDone: (saidYes: any) => {
```
dimitropoulos: picked up 4 of these that were any in `987fa5e094af6bed0dc072b1a6507eb99e170de2` (you can check the modal callback type to see what the value is... normally I'd say we should improve this but the reality is that we're going to remove this entire system anyway

packages/insomnia/src/ui/components/modals/git-staging-modal.tsx
```diff
@@ -255,7 +255,7 @@ export class GitStagingModal extends PureComponent<Props, State> {
       }
 
       items[gitPath] = {
-        type,
+        type: type as any,
```
dimitropoulos, line 258: that's a neat trick.  I haven't seen this one before (to not use the shortand so you can cast right there in the object literal).

packages/insomnia/src/ui/components/wrapper-home.tsx
```diff
@@ -404,7 +404,7 @@ const mapStateToProps = (state: RootState) => ({
   workspacesForActiveProject: selectWorkspacesForActiveProject(state),
 });
 
-const mapDispatchToProps = dispatch => {
+const mapDispatchToProps = (dispatch: Dispatch<AnyAction>) => {
```
dimitropoulos, line 407: thanks.  I always get a headache when I try to type these redux things in our app.

packages/insomnia/src/ui/containers/app.tsx
```diff
@@ -1270,7 +1273,7 @@ class App extends PureComponent<AppProps, State> {
           message: 'Are you sure you want to clear all models? This operation cannot be undone.',
           yesText: 'Yes',
           noText: 'No',
-          onDone: async yes => {
+          onDone: async (yes: any) => {
```
dimitropoulos: got the two remaining of these that were any in 43b894bcd3c1ee1e6ae8732615a92d92b17861a5

packages/insomnia/src/ui/dnd-backend.ts
```diff
@@ -2,7 +2,7 @@ import HTML5Backend from 'react-dnd-html5-backend';
 
 // @ts-expect-error -- TSCONVERSION ignoring until we can update react-dnd-html5-backend to get accurate upstream types
 class DNDBackend extends HTML5Backend {
-  handleTopDragEndCapture(event) {
+  handleTopDragEndCapture(event: any) {
```
dimitropoulos, line 5: I'm very glad you didn't try to "go harder" on these because we're going to rip out our DND implementation soon, after what we learned this week with the create-request PR, and update it like 10 major versions (TLDR; all of this goes away and it's just regular hooks now).

packages/insomnia/src/ui/redux/modules/workspace.ts
```diff
@@ -41,7 +41,7 @@ export const createWorkspace = ({ scope, onCreate }: {
   scope: WorkspaceScope;
   onCreate?: OnWorkspaceCreateCallback;
 }) => {
-  return (dispatch, getState) => {
+  return (dispatch: any, getState: any) => {
```
dimitropoulos, line 44: my thoughts exactly... honestly...

packages/insomnia/tsconfig.build.json
```diff
@@ -5,6 +5,7 @@
     "rootDir": ".",
     "resolveJsonModule": true,
     "strict": false,
+    "noImplicitAny": true,
```
dimitropoulos, line 8: image

for anyone wondering "why not put this in the `tsonfig.base.json` it's because that file already has `strict: true` and strict mode includes noImplicitAny so it's already on for every other package after this PR

dimitropoulos (approved):
such a great improvement.  wonderful.

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -102,12 +102,14 @@ export class GenerateCodeModal extends PureComponent<Props, State> {
...
-    this._generateCode(request, target, client);
+    // TODO: remove non-null assertion
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    this._generateCode(request, target, client!);
   }
 
   async _generateCode(request: Request, target: HTTPSnippetTarget, client: HTTPSnippetClient) {
     // Some clients need a content-length for the request to succeed
-    const addContentLength = (TO_ADD_CONTENT_LENGTH[target.key] || []).find(c => c === client.key);
+    const addContentLength = Boolean((TO_ADD_CONTENT_LENGTH[target.key] || []).find(c => c === client.key));
```
johnwchadwick, line 112: Oh yeah, this is another line that changes the code slightly, and not just pure type changes only. I think this is truly the only other one.

johnwchadwick:
OK. I've read over the full diff one more time. Dimitri has reviewed my changes. I've reviewed Dimitri's changes. I think we're good to go.

Apologies for pushing this out relatively quickly, but since it seemed like we had reached an agreement that this approach was acceptable, I hope nobody will wind up _surprised_ by what is in this PR. In addition, after reading it multiple times and ensuring that tests pass and doing a bit of testing manually, I have relatively good confidence that these changes won't break anything despite the sweeping scope.

With that having been said, direct any hatemail about your merge conflicts at me :)

PR #4867, opened 2022-06-15
Add graphQL to smoke tests (INS-1592).

filfreire:
## Changes

As part of INS-1592, this PR adds GraphQL to our smoke test express server. It also attempts to add a smoke test scenario that would cover both basic usage as well as going through the schema documentation. This latter bit is useful to help us signal bugs like INS-1582.

## TODO:

- [x] add graphql to smoke test server
- [x] Improve playwright script selectors
- [x] Fix assertions and playwright flakiness when trying to open the graphQL schema


gatzjames (approved):
🚀

jackkav (approved):
bearer -> ringBearer is a bit clearer but not blocking

PR #4869, opened 2022-06-15
Enable strict TypeScript in insomnia.

johnwchadwick:
All trivial changes. The only non-type change I have here is that `_fetch` seems to erroneously call `this._fetch` in its retry loop, likely crashing. I'm pretty confident insomnia-api never returns 502 in its current form, so this codepath is likely not called (there's no evidence of it in Sentry.)

This one was a lot easier/quicker to do than the other one (only a couple hours,) so I am not worried about needing to rebase it a few times. That said, I'm eager to see it merged so we can catch more bugs, and fix some of the soundness errors that are uncovered with it.

As with the last one, bugfixes and almost any change that would modify the behavior are out of scope here. I only included the fetch change because there is literally no way it isn't already broken + the solution is very trivial.

### DIFF
--- a/packages/insomnia/src/ui/components/codemirror/extensions/clickable.ts
+++ b/packages/insomnia/src/ui/components/codemirror/extensions/clickable.ts
@@ -6,7 +6,7 @@ import { AllHtmlEntities } from 'html-entities';
 import { FLEXIBLE_URL_REGEX } from '../../../../common/constants';
 const entities = new AllHtmlEntities();
 
-CodeMirror.defineExtension('makeLinksClickable', function(handleClick: CodeMirrorLinkClickCallback) {
+CodeMirror.defineExtension('makeLinksClickable', function(this: CodeMirror.Editor, handleClick: CodeMirrorLinkClickCallback) {
   // Only add the click mode if we have links to click
   this.addOverlay({
     token: function(stream: any) {
--- a/packages/insomnia/tsconfig.build.json
+++ b/packages/insomnia/tsconfig.build.json
@@ -4,8 +4,6 @@
     "outDir": "./build",
     "rootDir": ".",
     "resolveJsonModule": true,
-    "strict": true,
-    "noImplicitAny": true,
     "isolatedModules": true,
     "strictNullChecks": true,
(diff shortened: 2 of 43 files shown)


packages/insomnia/src/ui/components/codemirror/extensions/clickable.ts
```diff
@@ -6,7 +6,7 @@ import { AllHtmlEntities } from 'html-entities';
 import { FLEXIBLE_URL_REGEX } from '../../../../common/constants';
 const entities = new AllHtmlEntities();
 
-CodeMirror.defineExtension('makeLinksClickable', function(handleClick: CodeMirrorLinkClickCallback) {
+CodeMirror.defineExtension('makeLinksClickable', function(this: CodeMirror.Editor, handleClick: CodeMirrorLinkClickCallback) {
```
dimitropoulos, line 9: for some reason I thought shadowing `this` had some edge-case or gotcha, but upon searching I couldn't find one, so cool I guess.

packages/insomnia/tsconfig.build.json
```diff
@@ -4,7 +4,7 @@
     "outDir": "./build",
     "rootDir": ".",
     "resolveJsonModule": true,
-    "strict": false,
+    "strict": true,
```
dimitropoulos: it's no biggie but now that every package is `strict` it'd be cool to remove this entry from all `tsconfig` and `tsconfig.build` files and just leave it there on the base.  we can do it later if you don't want to do it now, but twas the plan once we get the last package into strict mode (which this now would be the last one, right?).

dimitropoulos (approved):
only saw one very minor thing re the tsconfigs (non blocking, for sure).  looks great! woo hoo!

packages/insomnia/src/ui/components/codemirror/extensions/clickable.ts
johnwchadwick: Actually, shadowing `this` is special typescript syntax to set the `this` type of a function. IIRC it also causes the function to have type-checked `apply`/`bind` which is kind of cool.

PR #4872, opened 2022-06-16
fix: adding a fix to the openapi2kong circular deps issue

marckong:
# Description
This PR is to fix the circular dependency issue when parsing a deeply nested openapi spec document into a Kong declarative config. I find this openapi-2-kong package and its plugin to be somewhat redundant as it may have some overlapping with Deck and Kong CLI.

- added openapi-types to the custom OAS types
- converted deference method from the swagger parser into bundle method to get only $ref object if there is $ref property
- map through parameter schema and body options to resolve first one or two nesting with json schema as this is required by Kong and dump the entire "components" definitions to the object to allow Kong parser to catch references

Dumping the entire components is very dumb and should be avoided but this effort seems to not really fitting into the Insomnia repository domain and that for needs to be done outside of Insomnia in my opinion. Therefore, I didn't care too much to resolve only used references from components and attach them in order to stop dumping my hours.

*Some unit tests are broken but still testable.*

## How to test
1. spin up your own Kong environment following this [instructions](https://docs.konghq.com/gateway/latest/install-and-run/docker/) with enterprise license (please refer to the internal document -to be posted-)
2. take the circular yaml file in the PR JIRA ticket (linked from the Linear ticket)
3. run this branch to start a local Insomnia instance
4. create a new design document
5. import the yaml file
6. click the document title dropdown
7. click "Declarative Config" dropdown item to open a modal
8. click "Copy to clipboard" button
9. create a json file to paste the clipboard ```touch testing.json```
10. copy the json file to your docker container that runs the Kong gateway ``` docker cp testing.json ${container_id}:/testing.json```
11. run kong cli command ```docker exec -it ${container_id} kong config parse testing.json```

## Testing video
https://www.loom.com/share/c7d083ab687a4e18ae2e33a787f536bb

## Some notes about difficulty faced during this fix
There were several difficult moments when fixing this.
1. no prior knowledge of how Insomnia plugin should work with Kong 
13. no clear expected outcome of Kong declarative config generation out of a deeply nested spec document
14. the provided spec doc is about 3000 lines of code and making minimally bug-reproducible yaml file was very challenging as it needed to contain that circular dependency of references
15. no clear instructions or information of how to test and validate
16. existing openapi-2-kong package implementation seems to be mismatched with openapi-types and it has custom typings extending
17. there are many type castings and sometimes it violates the openapi spec typings
18. unit test mock data are basically repetition of helper function executions through iterations, which are very difficult to understand what needs to be asserted
19. no set up for symlinking the monorepo packages, so having to re-run bootstrap over and over (my next effort after this will be adding symlinking to the packages with lerna hoist option)


filfreire (approved):
LGTM!

I tested this locally on a handful of scenarios, with both `inso` and `Insomnia` packaged.

Validated these:
- declarative config created from Inso, both json and yaml formats
- declarative config created from Insomnia App
- example customer config 

Using [deck](https://github.com/kong/deck) `validate` and using  `docker exec -ti $KONG_CONTAINER kong config parse <config>`

Screenshot 2022-06-23 at 09 38 52

Screenshot 2022-06-23 at 09 45 10

Screenshot 2022-06-23 at 09 42 00

packages/openapi-2-kong/src/declarative-config/generate.ts
```diff
@@ -26,9 +26,16 @@ export function generateDeclarativeConfigFromSpec(
...
+     * There was an [issue](https://github.com/Kong/studio/issues/93) that required us to stringify and parse the declarative config object containing circular dependencies.
+     * However, that fix didn't seem to clear the issue of the circular dependencies completely.
+     *
+     * It is attempted to resolve the circular issue by bundling the openapi spec using SwaggerParser.bundle() method, which resolves all the schemas into $ref instead of dereferencing them.
+     * Then, we would just dump the components part of it with $schema property, so any JSON parsing logic can refer to the components object.
+     *
+     * Therefore, JSON.parse(JSON.stringify(result)) doesn't seem to be needed any more.
+     */
+    const result: DeclarativeConfigResult = declarativeConfigResult;
     return result;
```
gatzjames, line 39: Since these 2 are the same now we could change this to:
```suggestion
return declarativeConfigResult;
```

packages/openapi-2-kong/src/declarative-config/plugins.ts
```diff
@@ -36,46 +39,88 @@ const generatePlugin = (tags: string[]) => ([key, value]: Entry<PluginItem>): DC
...
   query: 'form',
   path: 'simple',
 };
 
-const generateParameterSchema = (operation?: OA3Operation) => {
+const resolveParameter = ($refs: SwaggerParser.$Refs, parameter: OpenAPIV3.ParameterObject | OpenAPIV3.ReferenceObject): OpenAPIV3.ParameterObject => {
+  if ('$ref' in parameter) {
+    const components = $refs.get('#/components');
+    const dereferenced = $refs.get(parameter.$ref);
+    const param = omit(parameter, ['$ref']);
```
gatzjames: We could remove the omit dep here:

```suggestion
const { $ref, ...param } = parameter;
```

gatzjames:
Wanted to say thanks for the great work on this!!! 🚀 Especially on the tests and documenting the risks, unknowns and future work!
The code is much better than it was and 💯 for improving the types!
Left some suggestions but don't have anything blocking rather than updating the e2e tests!

gatzjames (approved):
🚀 Awesome! LGTM!

Tieske:
Had a look at this, but this is not ok imho.

See this test output: https://github.com/Kong/insomnia/pull/4872/files#diff-851bcc7412f8c4f53bf99005320e7847469d74b5ada91dbb114561a44c7fa219R44

The body schema now results in:
```json
{
  "type": "object",
  "title": "request-body-nullable",
  "description": "This object is sent in post application request.",
  "required": [
    "redirectUri"
  ],
  "properties": {
    "redirectUri": {
      "type": [
        "string",
        "null"
      ],
      "nullable": true
    }
  },
  "components": {
    "schemas": {
      "request-body-nullable": {
        "type": "object",
        "title": "request-body-nullable",
        "description": "This object is sent in post application request.",
        "required": [
          "redirectUri"
        ],
        "properties": {
          "redirectUri": {
            "type": [
              "string",
              "null"
            ],
            "nullable": true
          }
        }
      }
    }
  },
  "$schema": "http://json-schema.org/schema"
}
```

Despite there being no references in the schema, it includes the `components` section, which actually has a complete copy. This will most likely blow up the size of the generated configs.

marckong:
@Tieske fair. I noted that we are just dumping the entire "components" part. We can have a second iteration to actually tree walk and resolve only what is "referred". However, that wouldn't be a trivial effort and we may look for some help from other teams enhance this part. We needed to fix this issue to unblock the customer. This is certainly not fitting into our Insomnia domain in my honest opinion.

Issue #4195, opened 2021-11-09
Edit Cookie shows Expires as an Invalid Date


mountainash:
### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for a bug that matches the one I want to file, without success.

### Insomnia Version

2021.6.0

### What operating system are you using?

macOS

### Operating System Version

macOS Big Sur v11.6.1

### Installation method

Download

### Expected Behavior

Expires field to show the value as it was originally entered

### Actual Behavior

When (re)editing a cookie the **Expires** field is transformed from a JavaScript date string to a UTC timestamp.

### Reproduction Steps

When clicking **Edit** on an existing cookie in the **Manage Cookie** modal, the **Expires** field value is shown as UTC (eg. `1636720197980`) instead of the originally entered (and valid) JavaScript time type string (eg. `2021-11-12T12:29:57.980Z`). Focussing the field will correctly show `Invalid Date` - but only if focused, clicking save allows the invalid value to be stored, and then will go on to cause confusion when sending requests as most‡ receiving servers won't accept it.

‡ _likely all_

wdawson:
Thanks for the report @mountainash 

We have some ideas for revamping the cookie experience generally and this is one of the things we want to fix. Stay tuned.

filfreire:
Was able to reproduce this on latest `2021.7.2` app version, and also the previous one `2021.6.0`:

<img width="702" alt="Screenshot 2021-12-13 at 12 45 27" src="https://user-images.githubusercontent.com/11976836/145816337-bd9d43f9-0e13-4234-8209-d6f3a8ce8d0d.png">
<img width="175" alt="Screenshot 2021-12-13 at 12 45 20" src="https://user-images.githubusercontent.com/11976836/145816341-83894d32-0f5c-4c9d-9bf6-b614b000bd74.png">

Issue #4035, opened 2021-09-16
Cookie jar cached when chaining 


cskwrd:
### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for a bug that matches the one I want to file, without success.

### Insomnia Version

2021.5.3

### What operating system are you using?

Windows

### Operating System Version

Windows 10 Pro Version 20H2

### Installation method

Exe

### Expected Behavior

1. Response-dependent template tags are evaluated.
2. Decision to make preceding request is made.
3. Cookie jar updated as needed.
4. Dependent request succeeds.

### Actual Behavior

Dependent request always fails.

### Reproduction Steps

1. Open [gist](https://gist.github.com/cskwrd/3511d99f436fd099e02df11cd949cbdb).
2. Downloads files.
3. Execute [`build-run.sh`](https://gist.github.com/cskwrd/3511d99f436fd099e02df11cd949cbdb#file-build-run-sh).
4. Import [`insomnia_stale_cookie_bug.json`](https://gist.github.com/cskwrd/3511d99f436fd099e02df11cd949cbdb#file-insomnia_stale_cookie_bug-json) collection into Insomnia.
5. Send `Check Sum (Cookie Jar)` between 1 and _a lot_ of times.
6. Experience pain and sadness.
7. Send `Check Sum (Force New Cookie)` between 1 and _a lot_ of times.
8. Experience pleasure and happiness.

filfreire:
Hi @cskwrd, thank you for reporting this issue!

I was able to reproduce it thanks to your example. For some reason, regardless of the `PHPSESSID` being "renewed" by `/setup.php` we're not updating it.

For some reason the `Send cookies automatically` option seems to ignore the cookie that is on the cookie jar and not send it for single requests:

<img width="845" alt="Screen Shot 2022-05-10 at 12 01 35" src="https://user-images.githubusercontent.com/11976836/167614964-cb7bd75c-66d0-41cf-8783-5e46ff035f99.png">

<img width="403" alt="Screen Shot 2022-05-10 at 12 00 58" src="https://user-images.githubusercontent.com/11976836/167614942-44bc2402-c741-4af7-ad1e-3d80778be3fe.png">
 
<img width="943" alt="Screen Shot 2022-05-10 at 12 00 48" src="https://user-images.githubusercontent.com/11976836/167614880-6fdd11a5-7ae5-4bca-a157-2da961796ebb.png">

[...]

Issue #2750, opened 2020-10-20
Unable to create tag for Cookie


svaldetero-envoc:
**Describe the bug**
I have cookies in the cookie jar. When I try to create a tag that "references a cookie value from the cookie jar", it tells me there are no cookies for that url.

**To Reproduce**
- Add a tag
- Select Cookie
- Type any url that matches something in the cookie jar

**Expected behavior**
I should be able to choose a cookie from that jar

**Screenshots**
![cookies](https://user-images.githubusercontent.com/54991761/96605499-e7148c80-12bb-11eb-9903-55ebc74db945.png)
![tag](https://user-images.githubusercontent.com/54991761/96605531-ee3b9a80-12bb-11eb-86a3-99e790774a50.png)

**Desktop (please complete the following information):**
 - OS: Windows
 - Installation Method: installer
 - App Version Core 2020.4.1

**Additional context**
I'm trying to get the XSRF token from aspnet core site. We call /me and it is returned as a cookie on the response. Then subsequent calls need to use that cookie.

jgiovaresco:
@svaldetero 
I've built a simple test case on my side and I noticed that I should use `http://localhost` in `Cookie Url` (instead of `localhost`) to be able to find cookies.

nijikokun:
Seems like we might need a better description, or error message detailing what format is expected.

Issue #1730, opened 2019-10-14
[Bug] Cookie Tag is replaced by fix value


vekunz:
- Insomnia Version: 7.0.1
- Operating System: Windows 10

## Details

We have a "Request A", which returns a value in a cookie. Then we have a "Request B" with a different domain, which needs the same cookie with the exact same value. So I used the "Cookie" Tag to inject the Cookie-Value from Request A to the Cookie-Value of Request B.
The problem is, that this tag is replaced by the value after a short period of time. I don't know exactly why and when, but it happends several times a day.

gschier:
Can you specify what you mean (eg. screenshot) when you say "this tag is replaced by the value after a short period of time"?

vekunz:
This is how it should look like:
![2019-10-16 10_39_13-d ecs architecture (ecm vkunad local) – CheckFileInfo](https://user-images.githubusercontent.com/5724298/66902612-3d902080-f001-11e9-85db-5d3599c6f439.png)

And this is how it looks like after a period of time:
![2019-10-16 10_38_34-d ecs architecture (ecm vkunad local) – CheckFileInfo](https://user-images.githubusercontent.com/5724298/66902670-50a2f080-f001-11e9-8088-395cc04f4f0a.png)

vekunz:
I just have discovered, when the Tag is replaced by the value. If I reexecute Request A, which updates the Cookie (referenced by the tab), then the Tag is replaced by the old value of the referenced cookie.

vekunz:
Are there any updates?

PR #4897, opened 2022-06-24
Fix edit cookie shows expires as an Invalid Date

tangweikun:
#4195 Fix Edit Cookie shows Expires as an Invalid Date

Suggestion：Maybe can use DatePicker instead of Input in Expires field 

changelog(Fixes): Fixed an issue where editing a cookie wrongfully displayed expiry dates as invalid

### DIFF
--- a/packages/insomnia/src/ui/components/modals/cookie-modify-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/cookie-modify-modal.tsx
@@ -188,7 +188,7 @@ export class UnconnectedCookieModifyModal extends PureComponent<Props, State> {
         <label>
           {capitalize(field)}
           {field === 'expires' ?
-            <input type="datetime-local" value={localDateTime} onChange={value => this._handleChange(field, value)} /> :
+            <input type="datetime-local" defaultValue={localDateTime} onChange={value => this._handleChange(field, value)} /> :
             <OneLineEditor
               defaultValue={val || ''}
               onChange={value => this._handleChange(field, value)}
--- a/packages/insomnia/src/ui/components/modals/cookie-modify-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/cookie-modify-modal.tsx
@@ -1,5 +1,6 @@
 import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import clone from 'clone';
+import { isValid } from 'date-fns';
 import { cookieToString } from 'insomnia-cookies';
 import React, { PureComponent } from 'react';
 import { connect } from 'react-redux';
@@ -107,12 +108,8 @@ export class UnconnectedCookieModifyModal extends PureComponent<Props, State> {
     const cookie = clone(nextCookie);
     // Sanitize expires field
     const expires = new Date(Number(cookie.expires))?.getTime();
-    const isInValidExpires = (val: Cookie['expires']) => val === null || val === '' || new Date(Number(val)).toString() === 'Invalid Date';
-    if (isInValidExpires(cookie.expires)) {
-      cookie.expires = null;
-    } else {
-      cookie.expires = expires;
-    }
+    const isValidDate = (val: Cookie['expires']) => val !== null && val !== '' && isValid(Number(val));
+    cookie.expires = isValidDate(cookie.expires) ? expires : null;
 
     // Clone so we don't modify the original
     const cookieJar = clone(prevCookieJar);
@@ -173,13 +170,18 @@ export class UnconnectedCookieModifyModal extends PureComponent<Props, State> {
     }
   }
 
-  _renderInputField(field: string, error: string | null = null) {
+  _renderInputField(field: string) {
     const { cookie } = this.state;
-    const {
-    } = this.props;
+    let error: string | null = null;
 
     if (!cookie) {
       return null;
+    }
+
+    if (field === 'expires') {
+      if (cookie.expires !== null && !isValid(Number(cookie.expires))) {
+        error = 'Invalid Date';
+      }
     }
 
     // @ts-expect-error -- mapping unsoundness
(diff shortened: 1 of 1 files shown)


filfreire (approved):
LGTM!

Before - date would show up as invalid, and upon closing cookie modal would dissappear:
Screenshot 2022-06-24 at 12 26 32


Now - it at least doesn't show up as invalid, and isn't deleted:
Screenshot 2022-06-24 at 12 29 15

It doesn't solve all our problems with cookies, but at least gets rid of the invalid date/disappear behaviour, so from my side I think we can merge. Thank you @tangweikun for contributing this 👍

tangweikun:
> 



> LGTM!
> 
> Before - date would show up as invalid, and upon closing cookie modal would dissappear: Screenshot 2022-06-24 at 12 26 32
> 
> Now - it at least doesn't show up as invalid, and isn't deleted: Screenshot 2022-06-24 at 12 29 15
> 
> It doesn't solve all our problems with cookies, but at least gets rid of the invalid date/disappear behaviour, so from my side I think we can merge. Thank you @tangweikun for contributing this 👍

What are the other related issues about cookies? You can show me the issues, I can take a deeper look another day. @filfreire

filfreire:
@tangweikun if you wish to have a look, these are some that are open:
- https://github.com/Kong/insomnia/issues/2750 
- https://github.com/Kong/insomnia/issues/4035
- https://github.com/Kong/insomnia/issues/1730

packages/insomnia/src/ui/components/modals/cookie-modify-modal.tsx
```diff
@@ -106,9 +106,9 @@ export class UnconnectedCookieModifyModal extends PureComponent<Props, State> {
 
     const cookie = clone(nextCookie);
     // Sanitize expires field
-    const expires = new Date(cookie.expires || '').getTime();
-
-    if (isNaN(expires)) {
+    const expires = new Date(Number(cookie.expires))?.getTime();
+    const isInValidExpires = (val: Cookie['expires']) => val === null || val === '' || new Date(Number(val)).toString() === 'Invalid Date';
+    if (isInValidExpires(cookie.expires)) {
```
jackkav: Its better to use positive naming here `isValidDate`
 also consider using date-fns for validation as we have it installed https://date-fns.org/v2.28.0/docs/isValid

packages/insomnia/src/ui/components/modals/cookie-modify-modal.tsx
```diff
@@ -227,7 +227,7 @@ export class UnconnectedCookieModifyModal extends PureComponent<Props, State> {
                   </div>
                   {this._renderInputField(
                     'expires',
-                    isNaN(new Date(cookie.expires || 0).getTime()) ? 'Invalid Date' : null,
+                    cookie.expires !== null && new Date(Number(cookie.expires)).toString() === 'Invalid Date' ? 'Invalid Date' : null,
```
jackkav: since this isn't react i'd consider moving it into a condition inside of `_renderInputField`
if(type==='expires'){
}

jackkav:
After testing this PR. It is an improvement on the original but could afford to be more clear about what values are acceptable inputs to the expires field.
It seems currently only a datetime expressed as timeSinceEpoch is accepted. eg. 1672531200000 
Which is inconvenient as a user experience.

I'm considering if we should either add support for typing in a GMT/UTCString eg. Thu, 21 Oct 2021 07:28:00 GMT
or perhaps just force the input to use a datetime input.
According to https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#write_a_new_cookie the tough-cookie library will always create a GMTstring but it can take string | number | Date | null
```
;max-age=max-age-in-seconds (e.g., 60*60*24*365 or 31536000 for a year)
;expires=date-in-GMTString-format If neither expires nor max-age specified it will expire at the end of session.
```
For now I think we can ignore the `never` state, but I've pushed a commit that addresses the above. What do we think? @tangweikun @filfreire 

image

tangweikun:
> After testing this PR. It is an improvement on the original but could afford to be more clear about what values are acceptable inputs to the expires field. It seems currently only a datetime expressed as timeSinceEpoch is accepted. eg. 1672531200000 Which is inconvenient as a user experience.
> 
> I'm considering if we should either add support for typing in a GMT/UTCString eg. Thu, 21 Oct 2021 07:28:00 GMT or perhaps just force the input to use a datetime input. According to https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#write_a_new_cookie the tough-cookie library will always create a GMTstring but it can take string | number | Date | null
> 
> ```
> ;max-age=max-age-in-seconds (e.g., 60*60*24*365 or 31536000 for a year)
> ;expires=date-in-GMTString-format If neither expires nor max-age specified it will expire at the end of session.
> ```
> 
> For now I think we can ignore the `never` state, but I've pushed a commit that addresses the above. What do we think? @tangweikun @filfreire
> 
> image
LGTM!
The interactive experience is better than before.At the same time I think expires row should be made shorter, now the user has to click on the far right to pop up the datepicker, like this:

coo

jackkav:
@tangweikun I agree, but I don't think we have a half width convention anywhere else yet.

filfreire (approved):
LGTM!

⚠️ Fyi @jackkav @tangweikun found a few issues:

- Pressing the `Clear` button doesn't visible delete the date from the input field (though in the background it does seem to delete it. Pressing `Clear` button a second time seems clear up the date.

1st clear attempt:
Screenshot 2022-07-04 at 16 15 51

2nd clear attempt:
Screenshot 2022-07-04 at 16 15 57

Finally cleared:
Screenshot 2022-07-04 at 16 16 03

- When I try to type a date in one of the input parts, like only the year part, I can't seem to write all 4 digits without that part being reset (e.g. try to write `2022`).
image

PR #4906, opened 2022-06-27
preload all the dynamic chunks for the app

gatzjames:
WIP: Move dynamic module loading to the left


gatzjames:
Closing as this will probably happen in the bundler with routes as the lazy boundaries

PR #4908, opened 2022-06-27
tsconfig cleanup

dimitropoulos:
This PR starts from a combination of a notice of redundant tsconfig options in https://github.com/Kong/insomnia/pull/4869#pullrequestreview-1009650868 and a recent spike of typescript project references which uncovered that it's been a long time since anyone took a cleanup pass on the configs.

The git history is pretty clean, so it should go pretty dang quickly if you review commit by commit.

No logical changes in this PR.


tsconfig.base.json
```diff
@@ -20,6 +19,7 @@
     "sourceMap": true,
     "strict": true,
     "target": "ES2019",
-    "types": ["node"]
-  }
+    "types": ["node"],
+    "useUnknownInCatchVariables": false,
```
marckong, line 23: I am wondering why we are turning this off? Is there any reason we wouldn't benefit from this - https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables. I suppose, we want to ensure ```catch(err: unknown) {}```?

> With the flag useUnknownInCatchVariables enabled, then you do not need the additional syntax (: unknown) nor a linter rule to try enforce this behavior.

dimitropoulos: we turned it off because we weren't in a position at the time (during a typescript upgrade) to update everything.  I can do in a followup PR, but didn't want to touch any code in this PR -> just the configs.

if and when we fix this, it will definitely require logical changes to the codebase (probably harmless if-statements and sanity checks and whatnot, but still).  turn it to `true` and see the errors.

marckong (approved):

PR #4909, opened 2022-06-27
Upload sourcemaps to Sentry on publish

DMarby:
Tested here: https://github.com/Kong/insomnia/runs/7078908702?check_suite_focus=true

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -41,6 +41,7 @@ jobs:
         shell: bash
         run: npm run app-package
         env:
+          NODE_OPTIONS: '--max_old_space_size=6144'
           APPLE_ID: ${{ matrix.os == 'macos-latest' && secrets.DESIGNER_APPLE_ID || '' }}
           APPLE_ID_PASSWORD: ${{ matrix.os == 'macos-latest' && secrets.DESIGNER_APPLE_ID_PASSWORD || '' }}
           CSC_LINK: ${{ matrix.csc_link_secret != '' && secrets[matrix.csc_link_secret] || '' }}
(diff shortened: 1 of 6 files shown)


.github/workflows/release-build.yml
```diff
@@ -41,6 +41,7 @@ jobs:
         shell: bash
         run: npm run app-package
         env:
+          NODE_OPTIONS: '--max_old_space_size=6144'
```
DMarby, line 44: This is now needed as enabling sourcemaps cause Node to OOM in CI otherwise.

johnwchadwick (approved):
I worry a bit about the NODE_OPTIONS stuff, but otherwise everything looks good to me.

marckong (approved):
I reserve the view as @johnwchadwick 's. I wonder what would be path forward correcting it if needed.

.github/workflows/release-build.yml
```diff
@@ -118,6 +119,15 @@ jobs:
...
             packages/insomnia-inso/artifacts/*
 
+      - name: Upload source assets for Sentry
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.os }}-sentry
+          path: |
+            packages/insomnia/build/*.js
+            packages/insomnia/build/*.map
+            !packages/insomnia/build/yarn-standalone.js
```
filfreire, line 129: What does the `!` character do here?

filfreire (approved):
LGTM! 👍

.github/workflows/release-build.yml
DMarby: It negates the path so that yarn-standalone doesn't get copied even tho it's covered by the globs above

PR #4910, opened 2022-06-27
undoes prop drilling for creating request groups

dimitropoulos:
This PR is a continuation of the groundwork in https://github.com/Kong/insomnia/pull/4862, except it stays focused on creating request groups (rather than creating requests themselves).

### Before
image

### After
image

To test this PR: there are exactly 3 situations where this code is executed:
1. create a new request via the hotkey in a workspace with no requests (shift+meta+n on mac, ctrl+shift+n on win/linux) and observe that a new folder is created in the workspace
    image
1. click the plus button and select "New Folder"
    image
1. create a new folder from the dropdown of an existing folder
    image

### DIFF
--- a/packages/insomnia/src/ui/hooks/create-request-group.ts
+++ b/packages/insomnia/src/ui/hooks/create-request-group.ts
@@ -1,0 +1,22 @@
+import * as models from '../../models';
+import { showPrompt } from '../components/modals';
+
+export const createRequestGroup = (parentId: string) => {
+  showPrompt({
+    title: 'New Folder',
+    defaultValue: 'My Folder',
+    submitName: 'Create',
+    label: 'Name',
+    selectText: true,
+    onComplete: async name => {
+      const requestGroup = await models.requestGroup.create({
+        parentId,
+        name,
+      });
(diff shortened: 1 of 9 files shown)


packages/insomnia/src/ui/hooks/create-request-group.ts
```diff
@@ -0,0 +1,22 @@
+import * as models from '../../models';
+import { showPrompt } from '../components/modals';
+
+export const createRequestGroup = (parentId: string) => {
+  showPrompt({
```
marckong, line 5: I suppose we can't find another way to do it other than this way because of how modals are implemented for now.

marckong (approved):
LGTM since this PR unblocks us from moving forward more functional. I hope we get to see the day of having showPrompt like methods only within React components or hooks soon!

filfreire (approved):
LGTM 👍

packages/insomnia/src/ui/hooks/create-request-group.ts
dimitropoulos: yep

PR #4913, opened 2022-06-29
refactor: disable URI input for github & gitlab

marckong:
- Adding logic to disable the form fields if the git repository is already cloned unless reset. This is to prevent any miscellaneous conflicts, such as pulling a Github repository in the Gitlab authenticated & cloned form, which creates a conflict.


filfreire (approved):
LGTM!

gatzjames (approved):
🚀

Issue #4894, opened 2022-06-23
“Generate code” for request with AWS IAM v4 not having Authorization Header


cloud-j-luna:
### Expected Behavior

When I click “Generate Code” in a request with authentication set to AWS IAM v4 (with all values set and working if I send the request through Insomnia) I expected to see a Authorization Header in the generated code.

### Actual Behavior

No header is there and so the request fails as unauthorised.
Generated code is simply `curl —request GET —url <the url I’m requesting>`

### Reproduction Steps

Create a request that uses AWS IAM v4 and click Generate Code.

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Insomnia Version

2022.4.2

### What operating system are you using?

Ubuntu

### Operating System Version

20.04

### Installation method

Ubuntu Software Store

filfreire:
Hi @joao-luna-98 thank you for reporting this!

We're able to reproduce it as well:
![image](https://user-images.githubusercontent.com/11976836/176442123-0f03ff98-3055-48be-b30c-96ab0e5b2302.png)

![image](https://user-images.githubusercontent.com/11976836/176442168-919844cf-94e1-40e4-97fb-9cb636065fe4.png)

Doesn't seem to be happening for other authorization methods we have, e.g. bearer or basic:
![image](https://user-images.githubusercontent.com/11976836/176442282-01a68978-ce91-4f6d-9c63-fb2ac74b45bf.png)


@dimitropoulos any idea if this could be an issue on https://github.com/Kong/httpsnippet?

EDIT/UPDATE: we've been investigating some more and seems that it is happening even before the conversion when we call httpsnippet. If we try to export the request as HAR, it is missing the Auth header. Other authentication methods when exported to HAR, work and show the Auth header.

Seems we don't get the AWS auth header, because it was never added [to the getAuthHeader method](https://github.com/Kong/insomnia/blob/develop/packages/insomnia/src/network/authentication.ts#L23=)

cloud-j-luna:
Could there be an option to either show it as an `Authorization` header or as a `--aws-sigv4` which was added on version 7.75 of curl ([more details in the manpage](https://curl.se/docs/manpage.html)).

marko-stimac:
Auth details are missing in some other scenarios. I don't see them when using Microsoft NTLM auth with PHP Curl, data should include

CURLOPT_HTTPAUTH => CURLAUTH_NTLM,
[...]

PR #4916, opened 2022-06-29
Add debug log for missing AWS auth in getAuthHeader, #4894

filfreire:
While triaging issue #4894 with @kreosus we found that we are not getting AWS auth header likely due to the fact it's not implemented on `getAuthHeader` method, which is used on HAR request export, and also to generate code snippet.

Opening a draft now so we can tackle it asap.


filfreire:
Closing this for now until we have proper time to revisit it.

PR #4917, opened 2022-06-29
Feature/ins 1573 duplicating request makes original

gatzjames:
- Uses `capture: false` on global keypress events to respect bubbling (don't handle keypresses that other components already handled)
- Modals stop propagation for keypress events since they run on their own focus context.

Concerns:
- Code editor handles the `Esc` key in the default keymap and there's no way to move focus outside with the keyboard.
- There seems to be a need for really `global` keypress handlers which requires going through all the events and defining their focus scope.


filfreire (approved):
LGTM.

I tested this locally and like @gatzjames mentioned we now see a lot of cases where bubbling is taken into account, which means things work a bit different from what we're used to.

gatzjames:
Closing in favor of https://github.com/Kong/insomnia/pull/4937

PR #4918, opened 2022-06-29
fixes ContentTypeDropdown and AuthDropdown

dimitropoulos:
previously not rendering because of a bug with dropdown that hadn't been ported from insomnia-components, despite the same exact fix happening there.


jackkav (approved):
Looks fine to me, implementation drift and duplication is 😢  though. Hope we can agreed on a way solve this class of problem in future.

dimitropoulos:
totally agree.  really sucks to be bitten by this YET AGAIN with duplicating the dropdowns... sad day

PR #4934, opened 2022-07-06
useCallback with async is ok, after all

dimitropoulos:
in https://github.com/Kong/insomnia/pull/4904#discussion_r909461737 I had changed some useCallback calls to not use `async`.  Turns out I was overgeneralizing a rule that _is_ legit about `useEffect`, but doesn't apply to useCallback.  Oh well.  This PR fixes it to be more idiomatic and read more clearly.

The scope was restrained pretty far, so the scope should be nice and contained (with no logical changes).


jackkav (approved):
looks good to me

filfreire (approved):

marckong (approved):

PR #4955, opened 2022-07-12
use the window event to get the oauth refresh page

gatzjames:
Instead of relying on the `BrowserContext.on('page', ...)` event this uses the `ElectronApplication.on('window', ...)` to capture the new page in the OAuth flow.


jackkav (approved):
waiting periods seem to be shorter. would be nice to have an explanation, but I'm willing to bet this is more likely to make it less flakey than more.

PR #4964, opened 2022-07-13
Move login flow to accounts website.

johnwchadwick:
# UI
The UI from the application side looks like this.

### Initial State
| User action | Session expiration |
| --- | --- |
|  screenshot of the login modal when it is first opened by manual action; copy indicates browser has already opened | screenshot of the login modal when it is opened automatically; copy indicates that you must click a button to open the browser |

### Enter Manual Token
login modal displaying form to manually enter token

# Rationale

Our intent with this PR is to unify the SRP handshake and key derivation in one place. To spell it out explicitly, the intent of this PR is not to obfuscate the login process from the public, but just to make the app more flexible so that we can implement additional features to login without needing to duplicate it between the app and the accounts website.

## Security

The security model of this new login process has some implications. The general flow works like this:

1. On app startup, an X25519 keypair is generated for login delegation purposes.
2. When the login modal is opened, a new URL is opened linking into the accounts website, containing the public key.
3. The login process is continued in the browser, including SRP and key derivation. The accounts website always requires a password, even if the user is already logged in.
4. The accounts website running in the browser invokes an `insomnia://` deep-link containing a [sealed box](https://libsodium.gitbook.io/doc/public-key_cryptography/sealed_boxes) with the derived key and session token.
5. The app grabs the account keys and uses the derived key to decrypt them.

Though this removes both SRP and key derivation from the app side of the login flow, some code in the app still needs to deal with SRP and the key derivation process for the time being; this will be dealt with in the future, allowing us to remove a significant amount of code and some dependencies. For now, the remainder of session code is unchanged.

There are a few assumptions:

1. The end user device is trusted and not compromised; if it were not trusted, the current login flow would still be insecure, as it could be defeated with a simple key logger.
2. The app is not compromised; if the app were compromised, the existing login flow would also be unsafe, since either the raw credentials or the long-term session could be stolen.
3. The browser is trusted and not compromised; **This is a new assumption from this PR**. However, if the browser _was_ compromised, many other account-related actions would be insecure, as they can only be performed in the browser.

This login flow is assumed to be secure with those assumptions.

Here are some example scenarios:

- Phishing/token theft: An attacker (e.g. running in another browser tab) could try to start the app authentication flow and get the token from it somehow. The following mitigations apply:
    - Deep links: An attacker would need to register the `insomnia://` protocol to be able to steal the tokens directly out of the automatic flow. This is unlikely in cases where the computer is not compromised. (They could still try to convince the user to copy and paste the token into an attacker-controlled input.)
    - Password requirement: Other security issues require the user to explicitly input their password to finish the login process. Because of this, the login flow itself could never facilitate an automated attack, even in the case where the computer is compromised.
    - We could also improve this by implementing the login token as a one-time exchangeable token where the API call to exchange the token is inaccessible via browsers as a result of Same-Origin policy. This would increase the complexity of a phishing attack and make it easier to detect.
- Keymatter/ciphertext leakage: An attacker could try to intercept the public key or authentication box somehow.
    - The public key and cipher text are both useless without the private key, which never leaves the app's memory. Assuming the app is not compromised, this flow is already very secure in this regard.
    - We could still improve it by implementing the one-time token mechanism discussed above: then, the session token in the box would become useless after it is used.
- Accounts website XSS: An attacker could gain (possibly persistent) code execution on the accounts website, allowing the attacker to directly exercise code in the context of app authentication.
    - Due to the requirement to enter a password to create the app session, we are happily not introducing additional risk here.
    - If an attacker has code execution on the accounts website, they can potentially steal the derived key, OR they can potentially read the user's credentials as they are typing them. With regards to app authentication, while the attacker could now steal a long-term token out of the app auth flow, this is effectively not different from stealing the momentarily entered credentials/passphrase. And since the app authentication requires the credentials on the API side, there's no additional risk here.
    - Again, this can be improved by implementing a one-time token exchange mechanism, as it would make it impossible to get the token from the browser context alone.

changelog(Improvements): Login flow is now done using the [app.insomnia.rest](https://app.insomnia.rest/) website

### DIFF
--- a/packages/insomnia/src/index.html
+++ b/packages/insomnia/src/index.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta
       http-equiv="Content-Security-Policy"
-      content="font-src 'self' data:; default-src * insomnia://*; img-src blob: data: * insomnia://*; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: * insomnia://*;"
+      content="font-src 'self' data:; connect-src * data:; default-src * insomnia://*; img-src blob: data: * insomnia://*; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: * insomnia://*;"
     />
   </head>
   <body>
--- a/packages/insomnia/src/ui/components/modals/login-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/login-modal.tsx
@@ -1,5 +1,6 @@
 import { decodeBase64, encodeBase64 } from '@getinsomnia/api-client/base64';
 import { keyPair, open } from '@getinsomnia/api-client/sealedbox';
+import * as Sentry from '@sentry/electron/renderer';
 import React, { Dispatch, FormEvent, forwardRef, MutableRefObject, RefObject, SetStateAction, useCallback, useEffect, useImperativeHandle, useRef, useState } from 'react';
 
 import * as session from '../../../account/session';
@@ -61,9 +62,10 @@ export class LoginModalHandle {
       }
       const decoder = new TextDecoder();
       const box: AuthBox = JSON.parse(decoder.decode(boxData));
-      session.absorbKey(box.token, box.key);
+      await session.absorbKey(box.token, box.key);
       this.setState({ ...this.stateRef.current, state: 'done' });
     } catch (e) {
+      Sentry.captureException(e);
       this.setState({ ...this.stateRef.current, state: 'error', error: `Error loading credentials: ${String(e)}` });
     }
   }
--- a/packages/insomnia/src/ui/components/modals/login-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/login-modal.tsx
@@ -1,6 +1,6 @@
 import { decodeBase64, encodeBase64 } from '@getinsomnia/api-client/base64';
 import { keyPair, open } from '@getinsomnia/api-client/sealedbox';
-import * as Sentry from '@sentry/electron/renderer';
+import * as Sentry from '@sentry/electron';
 import React, { Dispatch, FormEvent, forwardRef, MutableRefObject, RefObject, SetStateAction, useCallback, useEffect, useImperativeHandle, useRef, useState } from 'react';
 
 import * as session from '../../../account/session';
@@ -24,7 +24,7 @@ const sessionKeyPair = keyPair();
 export let currentLoginModalHandle: LoginModalHandle | null = null;
 
 interface State {
-  state: 'ready' | 'error' | 'token-entry' | 'loading-session' | 'done';
+  state: 'ready' | 'error' | 'token-entry' | 'loading-session';
   url: string;
   error: string;
 }
@@ -63,7 +63,6 @@ export class LoginModalHandle {
       const decoder = new TextDecoder();
       const box: AuthBox = JSON.parse(decoder.decode(boxData));
       await session.absorbKey(box.token, box.key);
-      this.setState({ ...this.stateRef.current, state: 'done' });
     } catch (e) {
       Sentry.captureException(e);
       this.setState({ ...this.stateRef.current, state: 'error', error: `Error loading credentials: ${String(e)}` });
@@ -194,9 +193,6 @@ export const LoginModal = forwardRef<LoginModalHandle, {}>(function LoginModal({
       case 'error':
         return <div className="notice error">An error occurred: {state.error}</div>;
 
-      case 'done':
-        return <div className="notice surprise">Success! You are now authenticated.</div>;
-
       default:
         return <div className="notice error">Unknown state.</div>;
     }
@@ -214,8 +210,6 @@ export const LoginModal = forwardRef<LoginModalHandle, {}>(function LoginModal({
           <button className="btn" onClick={reset}>Retry</button> : null}
         {state.state === 'ready' ?
           <button className="btn" onClick={goToTokenEntry}>Manually Enter Token</button> : null}
-        {state.state === 'done' ?
-          <button className="btn" onClick={modalRef.current?.hide}>Close</button> : null}
       </ModalFooter>
     </Modal>
   );
--- a/packages/insomnia/src/ui/components/modals/login-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/login-modal.tsx
@@ -1,7 +1,7 @@
 import { decodeBase64, encodeBase64 } from '@getinsomnia/api-client/base64';
 import { keyPair, open } from '@getinsomnia/api-client/sealedbox';
 import * as Sentry from '@sentry/electron';
-import React, { Dispatch, FormEvent, forwardRef, memo, MutableRefObject, RefObject, SetStateAction, useCallback, useEffect, useImperativeHandle, useRef, useState } from 'react';
+import React, { Dispatch, FormEvent, forwardRef, memo, RefObject, SetStateAction, useCallback, useImperativeHandle, useRef, useState } from 'react';
 
 import * as session from '../../../account/session';
 import { getAppWebsiteBaseURL } from '../../../common/constants';
@@ -23,6 +23,10 @@ const sessionKeyPair = keyPair();
  */
 export let currentLoginModalHandle: LoginModalHandle | null = null;
 
+session.onLoginLogout(() => {
+  currentLoginModalHandle?.hide();
+});
+
 interface State {
   state: 'ready' | 'error' | 'token-entry' | 'loading-session';
   url: string;
@@ -43,7 +47,6 @@ interface AuthBox {
 export class LoginModalHandle {
   constructor(
     private readonly modalRef: RefObject<Modal>,
-    private readonly stateRef: Readonly<MutableRefObject<State & Options>>,
     private readonly setState: Dispatch<SetStateAction<State & Options>>,
   ) {}
 
@@ -53,12 +56,12 @@ export class LoginModalHandle {
   }
 
   async submitAuthCode(code: string) {
-    this.setState({ ...this.stateRef.current, state: 'loading-session' });
+    this.setState(state => ({ ...state, state: 'loading-session' }));
     try {
       const rawBox = await decodeBase64(code.trim());
       const boxData = open(rawBox, sessionKeyPair.publicKey, sessionKeyPair.secretKey);
       if (!boxData) {
-        this.setState({ ...this.stateRef.current, state: 'error', error: 'Invalid authentication code.' });
+        this.setState(state => ({ ...state, state: 'error', error: 'Invalid authentication code.' }));
         return;
       }
       const decoder = new TextDecoder();
@@ -66,7 +69,7 @@ export class LoginModalHandle {
       await session.absorbKey(box.token, box.key);
     } catch (e) {
       Sentry.captureException(e);
-      this.setState({ ...this.stateRef.current, state: 'error', error: `Error loading credentials: ${String(e)}` });
+      this.setState(state => ({ ...state, state: 'error', error: `Error loading credentials: ${String(e)}` }));
     }
   }
 
@@ -108,22 +111,9 @@ export const LoginModal = memo(forwardRef<LoginModalHandle, {}>(function LoginMo
     url: '',
   });
 
-  const stateRef = useRef(state);
-
-  useEffect(() => {
-    stateRef.current = state;
-  }, [state]);
-
   useImperativeHandle(ref, () => {
-    currentLoginModalHandle = new LoginModalHandle(modalRef, stateRef, setState);
+    currentLoginModalHandle = new LoginModalHandle(modalRef, setState);
     return currentLoginModalHandle;
-  }, []);
-
-  // Clean up global login modal handle on unmount.
-  useEffect(() => {
-    return () => {
-      currentLoginModalHandle = null;
-    };
   }, []);
 
   const reset = useCallback(() => {
--- a/packages/insomnia/src/ui/components/modals/login-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/login-modal.tsx
@@ -1,7 +1,7 @@
 import { decodeBase64, encodeBase64 } from '@getinsomnia/api-client/base64';
 import { keyPair, open } from '@getinsomnia/api-client/sealedbox';
 import * as Sentry from '@sentry/electron';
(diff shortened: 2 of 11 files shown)


packages/insomnia/src/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta
       http-equiv="Content-Security-Policy"
-      content="font-src 'self' data:; default-src * insomnia://*; img-src blob: data: * insomnia://*; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: * insomnia://*;"
+      content="font-src 'self' data:; connect-src * data:; default-src * insomnia://*; img-src blob: data: * insomnia://*; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; media-src blob: data: mediastream: * insomnia://*;"
```
johnwchadwick, line 7: Note: This is to allow for fast ArrayBuffer base64 computation via `fetch` and blob. It is a bit of a hack, but the approach is very fast in both Electron and the browser. I sincerely believe there is no security implication of adding `data:` to connect-src`, but I am not 100% certain of this.

dimitropoulos:
(oh, sorry, I didn't realize this was still in draft: well all the same added a super minor commit 624aa5002bdf0ee8158842bcaa0d1ab34746887e to reflect what we do all other places in the app.  the goal is to get onto jsx-transform (https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) at which point we'd have no need for the default `React` thingy at all except for cases where there are collisions (such as React.MouseEvent vs MouseEvent from lib.dom.  generally we aim to treat the react npm imports like any other npm.)

johnwchadwick:
> (oh, sorry, I didn't realize this was still in draft: well all the same added a super minor commit [624aa50](https://github.com/Kong/insomnia/commit/624aa5002bdf0ee8158842bcaa0d1ab34746887e) to reflect what we do all other places in the app. the goal is to get onto jsx-transform (https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) at which point we'd have no need for the default `React` thingy at all except for cases where there are collisions (such as React.MouseEvent vs MouseEvent from lib.dom. generally we aim to treat the react npm imports like any other npm.)

It actually *is* like any other import; namespaces and default imports are valid ES and not a special case. We use default imports and namespace imports in other places.

I don't understand how this is an improvement.

packages/insomnia/src/ui/redux/modules/global.tsx
```diff
@@ -304,6 +305,14 @@ export const newCommand = (command: string, args: any) => async (dispatch: Dispa
       break;
     }
 
+    case COMMAND_FINISH_AUTHENTICATION: {
+      if (submitAuthCode) {
```
DMarby: This should be `currentLoginModalHandle?.submitAuthCode` or similar I think?
Currently doesn't work, tested by changing the protocol handler in the website to `insomniadev://` as that's what the app registers when in dev mode.

packages/insomnia/src/ui/components/modals/login-modal.tsx
```diff
@@ -1,132 +1,224 @@
...
+        {state.state === 'done' ?
+          <button className="btn" onClick={modalRef.current?.hide}>Close</button> : null}
+      </ModalFooter>
+    </Modal>
+  );
+});
+
+LoginModal.displayName = 'LoginModal';
+
+export const showLoginModal = () => showModal(LoginModalHandle);
```
DMarby, line 234: Seems like all calls to `showModal(LoginModal)` are now broken, not sure if they need to be switched out to reference this instead or something?

This seems to break the protocol handler for `COMMAND_LOGIN`, as well as opening the login modal from Settings -> Account -> Log In

packages/insomnia/src/ui/components/modals/login-modal.tsx
```diff
@@ -1,132 +1,224 @@
...
-      });
+      const rawBox = await decodeBase64(code.trim());
+      const boxData = open(rawBox, sessionKeyPair.publicKey, sessionKeyPair.secretKey);
+      if (!boxData) {
+        this.setState({ ...this.stateRef.current, state: 'error', error: 'Invalid authentication code.' });
+        return;
+      }
+      const decoder = new TextDecoder();
+      const box: AuthBox = JSON.parse(decoder.decode(boxData));
+      session.absorbKey(box.token, box.key);
```
DMarby: This should probably be awaited since does a `_whoami` call, in-case it errors or such?

packages/insomnia/src/ui/components/modals/login-modal.tsx
```diff
@@ -1,132 +1,224 @@
...
+      this.setState({ ...this.stateRef.current, state: 'done' });
+    } catch (e) {
```
DMarby, line 71: Maybe worth logging the full error here, and/or pass it to Sentry, to aid in debugging if something breaks for a user?

packages/insomnia/src/ui/components/modals/login-modal.tsx
```diff
@@ -1,132 +1,224 @@
...
+      const rawBox = await decodeBase64(code.trim());
+      const boxData = open(rawBox, sessionKeyPair.publicKey, sessionKeyPair.secretKey);
+      if (!boxData) {
+        this.setState({ ...this.stateRef.current, state: 'error', error: 'Invalid authentication code.' });
+        return;
+      }
+      const decoder = new TextDecoder();
+      const box: AuthBox = JSON.parse(decoder.decode(boxData));
+      session.absorbKey(box.token, box.key);
+      this.setState({ ...this.stateRef.current, state: 'done' });
```
DMarby: Seems like something causes this to momentarily flash and the entire modal to disappear upon sign-in, which is a fairly confusing experience. Might have to do with the hooks in `absorbKey` causing a re-render somewhere up the tree?

DMarby (changes requested):

packages/insomnia/src/ui/redux/modules/global.tsx
johnwchadwick: It was meant to be `currentLoginModalHandle`... I'm not sure how that happened.

I don't currently have a way to test this because, as it turns out, the Electron function to register the protocol handler doesn't do anything on Linux.

packages/insomnia/src/ui/components/modals/login-modal.tsx
johnwchadwick: Fixed. (I was aware of this one, but I hadn't fixed it yet. I didn't expect this PR to be the first one to be reviewed.)

packages/insomnia/src/ui/redux/modules/global.tsx
johnwchadwick: (Should be fixed, but not tested.)

packages/insomnia/src/ui/components/modals/login-modal.tsx
johnwchadwick: Done.

packages/insomnia/src/ui/components/modals/login-modal.tsx
johnwchadwick: Added Sentry.captureException.

packages/insomnia/src/ui/components/modals/login-modal.tsx
johnwchadwick: Fair enough; I just got rid of the done state, since the old login modal didn't have it anyway.

packages/insomnia/src/ui/redux/modules/global.tsx
DMarby: Works like a charm :shipit:

packages/insomnia/src/ui/components/modals/login-modal.tsx
```diff
@@ -1,132 +1,220 @@
...
-      message,
+      state: 'ready',
+      title: title ?? '',
+      message: message ?? '',
+      url,
     });
-    this.modal?.show();
-    setTimeout(() => this._emailInput?.focus(), 100);
+    this.modalRef.current?.show();
+    clickLink(url);
```
DMarby: Not sure what to do about/whether to do anything about this yet UX wise, but leaving it as a note so I don't forget:

The fact that we automatically open this in the browser becomes a bit weird in certain situations, eg when this modal is automatically opened every time a sync action is done and you're not authenticated (including listing projects on the dashboard)

packages/insomnia/src/ui/redux/modules/global.tsx
```diff
@@ -304,6 +305,14 @@ export const newCommand = (command: string, args: any) => async (dispatch: Dispa
       break;
     }
 
+    case COMMAND_FINISH_AUTHENTICATION: {
+      if (currentLoginModalHandle) {
+        currentLoginModalHandle?.submitAuthCode(args.box);
+      } else {
+        console.log(`Received auth box, but no login modal... ${command}`);
+      }
```
DMarby, line 314: nit: Should probably add a break here, lest we accidentally fall-through to the next command added in the future

packages/insomnia/src/ui/components/modals/login-modal.tsx
johnwchadwick: You know, I thought about this one. Maybe we should make it explicit only in those cases? I could add another step for that.

We should probably improve the 'broken session' UX at some point, as it's currently extremely annoying. (A great start would be having a 'log out' option in the modal when you come from one of these cases.)

DMarby: I think making it explicit only for those cases probably makes sense, yeah.

Agreed that we definitely need to do something about the invalid session ux in general, log out option in the modal does sound reasonable to me as a start as well

packages/insomnia/src/ui/redux/modules/global.tsx
johnwchadwick: Done.

packages/insomnia/src/ui/components/modals/login-modal.tsx
johnwchadwick: I've fixed it now, and also added a Log Out button that displays on the left side. So now it should be less annoying.

wdawson:
Does the session expiration flow still say "...continue using Insomnia Plus"? If so, "Insomnia Plus" isn't a thing anymore so we should figure out different language.

johnwchadwick:
> Does the session expiration flow still say "...continue using Insomnia Plus"? If so, "Insomnia Plus" isn't a thing anymore so we should figure out different language.

This is from the API, not this PR, so it will be fixed whenever the change is deployed in the backend.

wdawson (approved):
LGTM

DMarby (approved):

packages/insomnia/src/ui/components/modals/login-modal.tsx
```diff
@@ -1,132 +1,236 @@
...
+
+  const submitToken = useCallback((event: FormEvent) => {
+    event.preventDefault();
+    currentLoginModalHandle?.submitAuthCode(tokenInputRef.current?.value ?? '');
+  }, []);
+
+  const copyUrl = useCallback(() => {
+    urlInputRef.current?.select();
+    document.execCommand('copy');
+  }, []);
```
dimitropoulos: issue: this implementation of copying broke for me when I tried it, causing me to not be able to log in.

the below fixes it (including the `import { clipboard } from 'electron'`):


```suggestion
  const copyUrl = useCallback(() => {
    clipboard.writeText(state.url);
  }, [state.url]);
```


My reproduction steps were:
- clicked `Log In` in the Account dropdown
- it redirects to a new open tab in a Chrome workspace that I didn't want it to (i.e. one that I don't have access to my insomnia password anyway)
- so I went back and clicked "Copy URL" and it did not copy anything

Also, document.execCommand is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) anyway.  Check out `copy-button.tsx` to see us using the clipboard export from electron in the same way.

Also, this implementation is dependent on the selection, whereas the suggested fix is not.  When first trying to figure out what was broken I tried selecting from the input, but it kept just copying `loginKey` to my clipboard.  Later I realized when I looked at the implementation that it was because of the selection being used, rather than the url.

packages/insomnia/src/ui/components/modals/login-modal.tsx
```diff
@@ -1,132 +1,236 @@
...
-            </div>
-            <button type="submit" className="btn">
-              {loading ? <i className="fa fa-spin fa-refresh margin-right-sm" /> : null}
-              Log In
+export const LoginModal = memo(forwardRef<LoginModalHandle, {}>(function LoginModal({ ...props }, ref) {
+  const modalRef = useRef<Modal>(null);
+  const tokenInputRef = useRef<HTMLInputElement>(null);
+  const urlInputRef = useRef<HTMLInputElement>(null);
+
+  const [state, setState] = useState<State & Options>({
```
dimitropoulos, line 105: question: could you share your motivations for using a state object like this?

We don't use this pattern anywhere else in the app, and it's not very common in react codebases for useState hooks.  I don't know the full implications of doing this in the future as we move to react 18 with batching and whatnot.  I am aware that at least [some people](https://dev.to/thoughtspile/are-many-usestates-better-than-usestateobject-j3l) have experimented with this approach and I wanted to get your take first.

johnwchadwick: It's just following the same state paradigm as the component it replaces. Truth be told, there's really no reason it *has* to be done this way, but it really does seem like an unnecessary PITA to do it any other way, and I just can't see it as an improvement over what I have here. As an example, having multiple different setState functions would make it more difficult to deal with the modal handle kerfuffle, since I need to plumb those. Are you aware of any downsides to this approach? If not, I think I'm just going to favor merging with this. Obviously, it can be refactored later, as it's basically just a simpler and lower boilerplate equivalent to splitting the object up into individual useStates.

packages/insomnia/src/ui/components/modals/login-modal.tsx
johnwchadwick: Thanks, but: you didn't need to write all of this; It was only a mistake. I copied the same code from the browser portion, which has no choice but to do this, because browsers have intentionally limited clipboard APIs.

johnwchadwick: (also fixed)

packages/insomnia/src/ui/components/modals/login-modal.tsx
dimitropoulos: ok, maybe we'll revisit later then.  thanks for the explanation.

dimitropoulos (approved):

johnwchadwick:
OK, I could probably get more reviews to be safe, but I actually feel pretty confident with the state we got it in, so I'm going to just go for the merge. Thanks everyone.

PR #4978, opened 2022-07-18
fix: return the input focus back after sending a request

marckong:
This PR closes [INS-1650](https://linear.app/insomnia/issue/INS-1650/regression-send-button-stays-pushed-in-after-clicking-figure-out-if) by adding focus back to the input when sending a request is executed as it was the original behaviour. We can discuss further on what we may find conforming to the accessibility standards in the followup discussion.

Before:
image

After:
image

### DIFF
--- a/packages/insomnia/src/ui/components/base/dropdown/dropdown.tsx
+++ b/packages/insomnia/src/ui/components/base/dropdown/dropdown.tsx
@@ -19,7 +19,7 @@ import { hotKeyRefs } from '../../../../common/hotkeys';
 import { executeHotKey } from '../../../../common/hotkeys-listener';
 import { fuzzyMatch } from '../../../../common/misc';
 import { KeydownBinder } from '../../keydown-binder';
-import { DropdownButton } from './dropdown-button';
+import { DROPDOWN_BUTTON_DISPLAY_NAME } from './dropdown-button';
 import { DropdownDivider } from './dropdown-divider';
 import { DropdownItem } from './dropdown-item';
 
@@ -46,7 +46,7 @@ const isComponent = (match: string) => (child: ReactNode) =>
   ]);
 
 const isDropdownItem = isComponent(DropdownItem.name);
-const isDropdownButton = isComponent(DropdownButton.name);
+const isDropdownButton = isComponent(DROPDOWN_BUTTON_DISPLAY_NAME);
 const isDropdownDivider = isComponent(DropdownDivider.name);
 
 // This walks the children tree and returns the dropdown specific components.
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/ui/components/request-url-bar.tsx
```diff
@@ -120,7 +121,7 @@ export const RequestUrlBar = forwardRef<RequestUrlBarHandle, Props>(({
   }, [handleUpdateDownloadPath, request._id]);
   const handleClearDownloadLocation = () => handleUpdateDownloadPath(request._id, null);
 
-  const handleKeyDown = useCallback(async (event: KeyboardEvent) => {
+  const handleKeyDown = useCallback((event: KeyboardEvent) => {
```
marckong, line 124: async isn't used here.

filfreire (approved):
The behavior of the send button itself seems to be better.

⚠️ Note that the dropdown right next to the send button still has the same behavior though:

Screen Shot 2022-07-19 at 08 41 45

marckong:
> The behavior of the send button itself seems to be better.
> 
> ⚠️ Note that the dropdown right next to the send button still has the same behavior though:
> 
> Screen Shot 2022-07-19 at 08 41 45

added a fix for the dropdown trigger button staying pushed in.

packages/insomnia/src/ui/components/base/dropdown/dropdown.tsx
```diff
@@ -46,7 +46,7 @@ const isComponent = (match: string) => (child: ReactNode) =>
   ]);
 
 const isDropdownItem = isComponent(DropdownItem.name);
-const isDropdownButton = isComponent(DropdownButton.name);
+const isDropdownButton = isComponent(DROPDOWN_BUTTON_DISPLAY_NAME);
```
dimitropoulos: question: why create an entirely new concept rather than the approach for doing these before using `Component.name`?

gatzjames: 🤔 Not sure how this can be achieved.
Assigning to DropdownButton.name is a no-op since it's readonly.

gatzjames (approved):
🚀 Tested and it works great!
Made a ticket to follow up on the edge cases of focus when we send with delays etc.

packages/insomnia/src/ui/components/base/dropdown/dropdown.tsx
marckong: This was definitely tricky. Indeed, it is read only, and returned as undefined. The provided React doc gave an idea, but its example didn't work in this case.

A quick hack was needed to use Object.assign, which I am not in favor of as it feels like a workaround. I would argue displayName would be a better fit here if anything. However, If there is another way of doing so, by all means, please feel free to contribute. Otherwise, I will merge it with Object.assign way.

dimitropoulos: (just an update: we talked about this in standup today and I passed along the react documentation on how to achieve this https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools)

marckong: Tried with the approach suggested on the doc:
image

Error occurs as the name is still undefined:
image

PR #4983, opened 2022-07-19
Prompt button to fc component

### LINKED ISSUE #4977: Convert Prompt Button component into a function component.
### Expected Behavior

Hello Insomnia,

I'll try to convert the prompt-button component into a function component.

### Actual Behavior

It's a Class component.

### Reproduction Steps

_No response_

### Additional Information

_No response_

### Insomnia Version

2022.5.0-beta.2

macOS

### Operating System Version

macOS Monterey

### Installation method

Download from insomnia.rest

_No response_

pavkout:
This PR has the following change:

* Convert prompt-button into function component.

Closes #4977

### DIFF
--- a/packages/insomnia/src/ui/components/base/prompt-button.tsx
+++ b/packages/insomnia/src/ui/components/base/prompt-button.tsx
@@ -1,4 +1,11 @@
-import React, { MouseEvent, PropsWithChildren, useCallback, useEffect, useRef, useState } from 'react';
+import React, {
+  MouseEvent,
+  PropsWithChildren,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 
 import { Button } from './button';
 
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/ui/components/base/prompt-button.tsx
```diff
@@ -1,150 +1,136 @@
...
 
-import { AUTOBIND_CFG } from '../../../common/constants';
 import { Button } from './button';
 
-type States =
-  | typeof STATE_DEFAULT
-  | typeof STATE_ASK
-  | typeof STATE_DONE
-  ;
+enum States {
```
marckong: nit pick - we may prefer to use a union type instead since enum will actually include the js object translation into the bundle.

packages/insomnia/src/ui/components/base/prompt-button.tsx
```diff
@@ -1,150 +1,136 @@
...
-    ).trim();
-    const finalDoneMessage = doneMessage || 'Done';
-    let innerMsg;
-
-    if (state === STATE_ASK && addIcon) {
-      innerMsg = (
-        <span className="warning" title="Click again to confirm">
-          <i className="fa fa-exclamation-circle" />
-          {finalConfirmMessage ? <span className="space-left">{finalConfirmMessage}</span> : ''}
+  const getChildren = useCallback(() => {
```
marckong: I wonder if this was necessary. If getChildren is to return different node based on props/states, why not just creating another functional component that renders them accordingly?

gatzjames (approved):
🚀 Tested and it works nice!

PR #4997, opened 2022-07-22
fix: (first pass) clear token with flush listener

marckong:
First pass to solve the oauth e2e test failure
- Added an event listener to the db changes to set loading false. This is a workaround.

### DIFF
--- a/packages/insomnia/src/ui/components/editors/auth/o-auth-2-auth.tsx
+++ b/packages/insomnia/src/ui/components/editors/auth/o-auth-2-auth.tsx
@@ -1,5 +1,5 @@
 import { Button } from 'insomnia-components';
-import React, { ChangeEvent, FC, ReactNode, useCallback, useMemo, useState } from 'react';
+import React, { ChangeEvent, FC, ReactElement, ReactNode, useCallback, useMemo, useState } from 'react';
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/ui/components/editors/auth/use-oauth2-token.ts
```diff
@@ -0,0 +1,42 @@
...
+      return;
+    }
+
+    setLoading(true);
+    models.oAuth2Token.remove(token);
+  }, [token]);
+
+  useEffect(() => {
+    function listener(changes: ChangeBufferEvent[]): void {
+      for (const change of changes) {
```
marckong, line 29: not really happy with this

packages/insomnia/src/ui/components/editors/auth/o-auth-2-auth.tsx
```diff
@@ -421,16 +422,10 @@ const OAuth2Error: FC = () => {
 };
 
 const useActiveOAuth2Token = () => {
```
marckong, line 424: @TODO: merge this with useOAuth2TokenClear hook in the similar manner

marckong:
didn't fix.

PR #4998, opened 2022-07-22
OAuth test checks

gatzjames:


packages/insomnia-smoke-test/tests/oauth.test.ts
```diff
@@ -54,7 +54,9 @@ test('can make oauth2 requests', async ({ app, page }) => {
   await page.locator('button:has-text("Clear OAuth 2 session")').click();
   await page.locator('button:text-is("Clear")').click();
   await page.locator('button:has-text("Click to confirm")').click();
```
filfreire: I think we need to remove this one as well if we change from `PromptButton` to `Button`:
```suggestion
```

gatzjames: 😁 thought I did

gatzjames:
Closing in favour of #4999

PR #5005, opened 2022-07-26
Chore/response viewer fc

gatzjames:
Updates the ResponseViewer to a function component.

Highlights:
- Removes the need for a ref (and process.nextTick hacks) to refresh the underlying components by using the key prop.
- `shouldComponentUpdate` was trying to check the equality of Buffers but we use getBody since a long time now so it became obsolete.
- Reading/parsing the data based on the type on the client side seems wrong and heavy. It could benefit from an api that allows us to handle this outside of the renderer process.

### DIFF
--- a/packages/insomnia/src/ui/components/viewers/response-viewer.tsx
+++ b/packages/insomnia/src/ui/components/viewers/response-viewer.tsx
@@ -1,9 +1,15 @@
-import { autoBindMethodsForReact } from 'class-autobind-decorator';
 import iconv from 'iconv-lite';
-import React, { Component, Fragment } from 'react';
+import React, {
+  forwardRef,
+  Fragment,
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react';
 
 import {
-  AUTOBIND_CFG,
   HUGE_RESPONSE_MB,
   LARGE_RESPONSE_MB,
   PREVIEW_MODE_FRIENDLY,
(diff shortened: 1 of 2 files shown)


packages/insomnia/src/ui/components/viewers/response-viewer.tsx
```diff
@@ -24,6 +29,10 @@ import { ResponseWebView } from './response-web-view';
 
 let alwaysShowLargeResponses = false;
```
marckong, line 30: this line of code is totally off this PR changes, but I wonder why this was declared outside of the component and used inside to affect what the component renders (for instance setting a state)

packages/insomnia/src/ui/components/viewers/response-viewer.tsx
```diff
@@ -280,23 +227,7 @@ export class ResponseViewer extends Component<ResponseViewerProps, State> {
...
-      filterHistory,
-      previewMode,
-      responseId,
-      updateFilter,
-      url,
-    } = this.props;
-    const { bodyBuffer, error: parseError } = this.state;
-    const error = responseError || parseError;
-
+  function renderResponseView() {
```
marckong, line 230: nit pick - in the ideal world it would be great if we can break this into actual components instead of having it as a render method.

marckong (approved):
Left minor comments that aren't really related to this PR changes. I ran a quick smoke test not covering all cases but this PR looks good to me for code changes. Good work tackling another challenging conversion!

packages/insomnia/src/ui/components/viewers/response-viewer.tsx
gatzjames: +1
I tried splitting this into another component but there's quite some coupling with the refs and would prefer to refactor in a follow up.

packages/insomnia/src/ui/components/viewers/response-viewer.tsx
gatzjames: Afaik this keeps the user's choice cached for all responses (for the app's lifetime).
So if a user switches to another large response it will use their previous option.
🤔 Worth investigating if it should be stored in user's settings or local storage for persistence.

marckong: Agreed! Thank you for explanation!

filfreire (approved):

PR #5006, opened 2022-07-26
Fix an issue with URL encoding on cURL imports

brxxn:
Fixes an issue where `application/x-www-form-urlencoded` cURL body parameters are left URL encoded when being imported, causing them to be double-encoded (for example, body parameters that contain an encoded `=` were imported as `%3D`, which would be sent as `%253D` rather than just `%3D`).

This fix was implemented by using `decodeURIComponent` on the name/value pair since the insomnia form body expects it to be decoded and encodes the values when they are sent.

changelog(Fixes): Fixed an issue with cURL imports that caused URL-encoded parameters to be encoded twice

### DIFF
--- a/packages/insomnia-importers/src/importers/curl.ts
+++ b/packages/insomnia-importers/src/importers/curl.ts
@@ -172,7 +172,7 @@ const importCommand = (parseEntries: ParseEntry[]): ImportRequest => {
     const pair = pairsByName[paramName];
 
     if (pair && pair.length) {
-      textBodyParams = textBodyParams.concat(paramName === 'data-urlencode' ? encodeURIComponent(pair) : pair);
+      textBodyParams = textBodyParams.concat(paramName === 'data-urlencode' ? pair.map(item => encodeURIComponent(item)) : pair);
     }
   }
 
--- a/packages/insomnia-importers/src/importers/curl.ts
+++ b/packages/insomnia-importers/src/importers/curl.ts
@@ -172,7 +172,7 @@ const importCommand = (parseEntries: ParseEntry[]): ImportRequest => {
     const pair = pairsByName[paramName];
 
     if (pair && pair.length) {
-      textBodyParams = textBodyParams.concat(pair);
+      textBodyParams = textBodyParams.concat(paramName === 'data-urlencode' ? encodeURIComponent(pair) : pair);
     }
   }
 
--- a/packages/insomnia-importers/src/importers/curl.ts
+++ b/packages/insomnia-importers/src/importers/curl.ts
@@ -223,8 +223,8 @@ const importCommand = (parseEntries: ParseEntry[]): ImportRequest => {
     body.params = textBody.split('&').map(v => {
       const [name, value] = v.split('=');
       return {
-        name: name || '',
-        value: value || '',
+        name: decodeURIComponent(name || ''),
+        value: decodeURIComponent(value || ''),
       };
     });
   } else if (textBody) {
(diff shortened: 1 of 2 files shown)


packages/insomnia-importers/src/importers/curl.ts
```diff
@@ -223,8 +223,8 @@ const importCommand = (parseEntries: ParseEntry[]): ImportRequest => {
     body.params = textBody.split('&').map(v => {
       const [name, value] = v.split('=');
       return {
-        name: name || '',
-        value: value || '',
+        name: decodeURIComponent(name || ''),
+        value: decodeURIComponent(value || ''),
```
johnwchadwick, line 227: One problem with this is that we naively merge all of the different data parameters (including `data-urlencode` which specifies the data the way Insomnia wants it.)

Still, I think this is actually the correct place to do `decodeURIComponent` at. So to make this accurate, above, only when handling `data-urlencode` parameters, we should probably first _encode_ with `encodeURIComponent`, before adding it to the array.

WDYT?

packages/insomnia-importers/src/importers/curl.ts
```diff
@@ -172,7 +172,7 @@ const importCommand = (parseEntries: ParseEntry[]): ImportRequest => {
     const pair = pairsByName[paramName];
 
     if (pair && pair.length) {
-      textBodyParams = textBodyParams.concat(pair);
+      textBodyParams = textBodyParams.concat(paramName === 'data-urlencode' ? encodeURIComponent(pair) : pair);
```
johnwchadwick: Unfortunately, this doesn't _quite_ work because `pairsByName` is a map of arrays. So `encodeURIComponent` would need to be applied to each element in `pair` rather than `pair` itself. (This is why the build/tests are failing.)

brxxn:
Should be resolved by latest commit on my branch

johnwchadwick (approved):
LGTM.

dimitropoulos (approved):
it's a sticky wicket, this one is.  recently we had something similar in httpsnippet to deal with https://github.com/Kong/httpsnippet/pull/198

cURL treats `-data` as being raw data but insomnia's curl importer parses it as form data if the content type is set to `application/x-www-form-urlencoded` - and insomnia will internally url encode the data that is input here, so if it is already url encoded it will encode it twice. the solution of encoding `-data-urlencode` arguments may seem a bit weird, but you can interleave `-data` and `-data-urlencode` so it needs to be normalized in some form ahead of time and this seems to be the way that makes the most sense.

this should make our handling of `-data` and `-data-urlencode` correct so that the cURL importer is able to correctly construct more request bodies.

@johnwchadwick and I wrote some tests for the preexisting behavior in https://github.com/Kong/insomnia/pull/5006/commits/43556c10876f48fa292c27555a44f607f121ebbe and then updated the tests in https://github.com/Kong/insomnia/pull/5006/commits/714a6b26a8f5d0c6cb3a6f0f53bb12a121d73164 to reflect the new changes this PR brought about

johnwchadwick:
As a future step, I think it'd be nice to fall back to raw text body for the request in the event that the cURL body cannot be correctly parsed as form data. (Example scenarios: `a=` couldn't be losslessly parsed since it has no value, so it will be missing the equals sign in the parsed form; `a=1=2` couldn't be parsed since one of the equals signs would wind up being url encoded no matter where you split the key/value pair; `%` couldn't be parsed because it is not url encoded, so no possible form data could result in a single raw % being sent.)

PR #5009, opened 2022-07-26
[INS-1517] adds GitHubStarsButton

dimitropoulos:
This PR adds a GitHub star tracker, as commonly seen in many applications.

### Some notes
- if incognito mode is enabled, it will not make any network requests
- app-wide unnecessary rerendering influenced the design:
  - when using `useEffect` the component was benchmarked to rerender 8 times (despite no props changes)
  - when using `useMount` it is improved (as, it only fetches on mount) to when changing insomnia activities (broadly speaking)

### App Dashboard
| BEFORE | AFTER |
| - | - |
| Screenshot_20220726_192216 | Screenshot_20220726_191908 |

### Request Collection
| BEFORE | AFTER |
| - | - |
| Screenshot_20220726_192705 | Screenshot_20220726_192634 |

### Design Document
| BEFORE | AFTER |
| - | - |
| Screenshot_20220726_192234 | Screenshot_20220726_192744 |

### Unit Test
| BEFORE | AFTER |
| - | - |
| Screenshot_20220726_192322 | Screenshot_20220726_191938 |

### Also works well in themes

| state | |
| - | - |
| neutral state | Screenshot_20220726_192519 |
| hover "Star" | Screenshot_20220726_192456 |
| hover count | Screenshot_20220726_192433 |

### Segment
As with the rest of the application, segment events will not fire if "send usage statistics" is false (including by being set by incognito mode).  If true, here is what the events look like (cc: @wdawson):

changelog(Improvements): Adds a GitHub stargazer button to the app header

### DIFF
--- a/packages/insomnia/src/ui/components/github-stars-button.tsx
+++ b/packages/insomnia/src/ui/components/github-stars-button.tsx
@@ -8,8 +8,12 @@ import { SegmentEvent, trackSegmentEvent } from '../../common/analytics';
 import { selectSettings } from '../redux/selectors';
 
 const Wrapper = styled.div({
-  marginLeft: 'var(--padding-md)',
+  fontSize: 'var(--font-size-xs)',
+  zIndex: 0,
+  marginLeft: -10,
   display: 'flex',
+  height: '16px',
+  alignSelf: 'center',
   '& a': {
     textDecoration: 'none !important',
     fontWeight: 'normal !important',
@@ -22,10 +26,11 @@ const Wrapper = styled.div({
 });
 
 const Star = styled.a({
-  backgroundColor: 'var(--hl-sm)',
-  padding: 'var(--padding-xxs)',
+  width: 42,
+  justifyContent: 'flex-end',
+  backgroundColor: 'var(--hl-xs)',
+  padding: '2px',
   alignItems: 'center',
-  height: '100%',
   display: 'flex',
   border: '1px solid var(--hl-sm)',
   color: 'var(--hl) !important',
@@ -40,11 +45,14 @@ const Icon = styled(SvgIcon)({
 
 const Counter = styled.a({
   alignItems: 'center',
-  padding: '0 var(--padding-xs)',
+  paddingLeft: 4,
+  paddingRight: 5,
+  padding: '0 3px',
   display: 'flex',
   fontVariantNumeric: 'tabular-nums',
   border: '1px solid var(--hl-sm)',
   borderLeft: 'none',
+  borderRadius: '0 10px 10px 0',
   ':hover': {
     color: 'var(--color-surprise) !important',
   },
--- a/packages/insomnia/src/ui/components/github-stars-button.tsx
+++ b/packages/insomnia/src/ui/components/github-stars-button.tsx
@@ -1,0 +1,87 @@
+import { SvgIcon } from 'insomnia-components';
+import React, { useState } from 'react';
+import { useMount } from 'react-use';
+import styled from 'styled-components';
+
+const Wrapper = styled.div({
+  marginLeft: 'var(--padding-md)',
+  display: 'flex',
+  '& a': {
+    textDecoration: 'none !important',
+    fontWeight: 'normal !important',
+    color: 'var(--hl) !important',
+  },
+  ':hover': {
+    color: 'var(--color-font) !important',
+    cursor: 'pointer',
+  },
+});
+
+const Star = styled.a({
+  backgroundColor: 'var(--hl-sm)',
+  padding: 'var(--padding-xxs)',
+  alignItems: 'center',
+  height: '100%',
+  display: 'flex',
+  border: '1px solid var(--hl-sm)',
+  color: 'var(--hl) !important',
+  ':hover': {
+    borderColor: 'var(--hl-xl)',
+  },
+});
+
+const Icon = styled(SvgIcon)({
+  marginRight: 'var(--padding-xxs)',
+});
+
+const Counter = styled.a({
+  alignItems: 'center',
+  padding: '0 var(--padding-xs)',
+  display: 'flex',
+  fontVariantNumeric: 'tabular-nums',
+  border: '1px solid var(--hl-sm)',
+  borderLeft: 'none',
+  ':hover': {
+    color: 'var(--color-surprise) !important',
+  },
+});
+
+export const GitHubStarsButton = () => {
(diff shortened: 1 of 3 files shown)


wdawson:
Why are there two different events?

packages/insomnia/src/ui/components/github-stars-button.tsx
```diff
@@ -0,0 +1,110 @@
...
+  ':hover': {
+    borderColor: 'var(--hl-xl)',
+  },
+});
+
+const Icon = styled(SvgIcon)({
+  marginRight: 'var(--padding-xxs)',
+});
+
+const Counter = styled.a({
```
marckong, line 46: Nice example of styled component pattern!

packages/insomnia/src/ui/components/github-stars-button.tsx
```diff
@@ -0,0 +1,110 @@
...
+  ':hover': {
+    color: 'var(--color-surprise) !important',
+  },
+});
+
+export const GitHubStarsButton = () => {
+  const { incognitoMode } = useSelector(selectSettings);
+  const [starCount, setStarCount] = useState(21700);
+  const [error, setError] = useState<Error | null>(null);
+  const org = 'Kong';
```
marckong: nit pick - I wonder if they need to be declared as local variables if they are static values.

packages/insomnia/src/ui/components/github-stars-button.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const repo = 'insomnia';
+
+  useMount(() => {
+    if (incognitoMode) {
+      return;
+    }
+
+    fetch(`https://api.github.com/repos/${org}/${repo}`)
```
marckong: non-blocker - I wonder what we would do during the idle time. Possible outcomes of fetching are data, loading, error. We have data and error, but not having loading. If we don't show the loading state of this button fetching, then it's probably not related.

marckong:
I did run through some smoke test. I noticed two things in the case of offline;
1. count not rendering (makes sense since its offline, but in this case should we show the button at all?)
image
2. there are some weird React flush errors when it's offline. I suppose it may be coming from the fetch logic. I am not sure if this is a big deal since it does not occur in the online state.
image

Otherwise, this PR looks good to me and the description of the PR is detained and amazing!

packages/insomnia/src/ui/components/github-stars-button.tsx
dimitropoulos: yep! exactly like you guessed: there's no loading state (just a default value).  This is really important because a loading state would cause UI jank when the numbers are popping in and out of view, which would cause the whole top bar to expand and contract rapidly.  This is the same reason I used `fontVariantNumeric: 'tabular-nums'`.

regarding idle time: nothing.  It's not an important value to keep up to date right to "the ones digit", after all, may implementations of this button (e.g. GitHub itself) use, like `21.7k` instead of `21,721` like we have in our design.  The other rationale there was that we know quite well how much insomnia users dislike us putting things in the app that have network activity (hence the creation of incognito mode, itself), so best to not have it on an interval or anything.  As they use the app, it'll periodically update just because of mounting, and that's plenty.

packages/insomnia/src/ui/components/github-stars-button.tsx
dimitropoulos: thanks! I'm always glad we're using this modern pattern in the app!

packages/insomnia/src/ui/components/github-stars-button.tsx
dimitropoulos: it's declared as a variable because they're used in 3 places.  As for why it's a _local_ variable: pulling them out to the module scope means that they'll be more exposed to other components in the app, which... I was thinking is probably best to avoid just to keep clean encapsulation (i.e. not leaking this implementation detail out to elsewhere).

I'm not super opinionated on this: so if you change your mind after hearing my explanation and think it should be blocking, please just let me know and I'd be happy to change.

marckong (approved):

marckong:
I left minor comments on something trivial and this PR looks good to me. I ran some smoke tests as well and it looks working as expected. One edge case may be handling the offline status, but I haven't been able to reproduce it after first few times of reload.

wdawson (approved):
LGTM

jackkav (approved):
I would prefer to keep the github url intact so make it easier to search for should we ever need to update it, but non-blocking and kinda opinionated.

filfreire (approved):
LGTM!

⚠️ I found 2 weird behaviors, but I leave it up to you to see if they are worth investigating or fixing in this PR:

- There's a slight discrepancy in stargazers numbers when you disable Incognito. It's only noticeable up to the point you switch to another tab. For example, first we show `21700` stargazers, then upon switching to another Design Document tab, we re-render with a new number of `21749`
Screen Shot 2022-07-27 at 10 41 14

Screen Shot 2022-07-27 at 10 41 20


- When we work offline / network disabled, the element is a bit jittery. It tries to render the star count, only to be back to just the star button in a split second:

https://user-images.githubusercontent.com/11976836/181218553-983fa919-6c08-4bdc-b1d7-b88130d992f2.mp4

dimitropoulos:
thanks @filfreire for finding that.  https://github.com/Kong/insomnia/pull/5009/commits/d07c58f8e4512d023b1f5fec776a5a06b7781716 fixes it (and as a note for the future, @gatzjames / @jackkav / @filfreire all saw and approved of the approach in that commit, with the suggestion from @gatzjames that in a future PR we can improve this further by hardcoding the fallback (today, set to 21700) with an environment variable that's fed into the production build by the CI).

packages/insomnia/src/ui/components/github-stars-button.tsx
dimitropoulos: update: ended up being inlined in https://github.com/Kong/insomnia/pull/5009/commits/3074a0a39d00baecd72338ceafdce0b4d1483d82

dimitropoulos:
@jackkav I did https://github.com/Kong/insomnia/pull/5009/commits/3074a0a39d00baecd72338ceafdce0b4d1483d82 in response to your feedback -> thanks!

DanielRouxSA:
Ahh I hate to be that person, but is there perhaps a way to hide this control? I find it doesn't meaningfully add to my experience with the app, and I find it visually cluttering.

PR #5020, opened 2022-07-28
INS-1692 removes document preview toggle

dimitropoulos:
This PR removes the preview button for the swagger-ui viewer:

| BEFORE | AFTER |
| - | - |
| Screenshot_20220728_153010 | Screenshot_20220728_152710 |

Note, that the prior behavior when clicking this button was to just have an empty/blank screen in place of the swagger-ui:

Screenshot_20220728_153524

Back in May, we added analytics to this in https://github.com/Kong/insomnia/pull/4761 to make sure that it is not a widely used feature of the app, and since then we have determined that it is not.

From the product discovery we've done on this, we've concluded that it is not altogether very useful to users to have this button, and it clutters up the app bar at the top of the application.

____

As always with Insomnia, we're aiming for a sleek user experience.  If you are reading this some time in the future and you have a critical use-case that involved hiding the spec generation, please let us know.

### DIFF
--- a/packages/insomnia/src/models/workspace-meta.ts
+++ b/packages/insomnia/src/models/workspace-meta.ts
@@ -25,7 +25,6 @@ export interface BaseWorkspaceMeta {
   paneHeight: number;
   paneWidth: number;
   parentId: string | null;
-  previewHidden: boolean;
   sidebarFilter: string;
   sidebarHidden: boolean;
   sidebarWidth: number;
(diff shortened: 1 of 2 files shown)


marckong:
Another great example of awesome description!

marckong (approved):
LGTM! I ran a smoke test, and I haven't run into any issues.

wdawson (approved):
sounds good to me

PR #5028, opened 2022-08-01
Fix signing out when pressing enter on the git form

kreosus:
Closes INS-1607

changelog(Fixes): Fixed an edge-case issue where the user would get a signing out suggestion instead of proceeding with GitHub/GitLab Sync


filfreire (approved):

gatzjames (approved):
🚀

marckong (approved):

PR #5040, opened 2022-08-03
update lerna/markdown-lint

gatzjames:
Highlights:
- Lerna 4 -> [5 breaking changes](https://github.com/lerna/lerna/releases/tag/v5.0.0) shouldn't affect our pipeline
- Ran markdown-lint and fixed a warning

TODO:
- [ ] Figure out why running `npm run bootstrap` updates all package-lock.json files with tabs
- [ ] Update the [release start workflow](https://github.com/Kong/insomnia/blob/develop/.github/workflows/release-start.yml#L74-L75) and validate everything works as expected


filfreire (approved):

marckong (approved):
Just ran
```shell
npm run hard-reset
nom run app-package
```

and there wasn't any issue. However, I didn't check other packages but I don't foresee an issue with this yet. LGTM!

DMarby:
The new version of lerna should be supporting v2 package-lock.json files correctly, so with this upgrade we should also remove the following step in release start workflow: https://github.com/Kong/insomnia/blob/develop/.github/workflows/release-start.yml#L74-L75

Would be good to test it before merging however.

filfreire:
Heads up @gatzjames @DMarby - I was having a quick look into this one locally and there are a few issues that require attention before we merge. I suspect same applies to #5081.

- ⚠️ Running `bootstrap` indents the lock files from spaces to tabs, but in some cases also from tabs to spaces 😕 🤷 

- ✅ We seem to no longer indeed need the line that @DMarby mentioned on `release-start`, that bit seems to now be fixed, examples:
image
image

jackkav:
cleaning up stale PRs

PR #5044, opened 2022-08-04
Websockets IPC API

gatzjames:
Adds an API for creating connections on WebSocket Requests.

Highlights:
- Uses [ws](https://www.npmjs.com/package/ws) for networking
- Provides an api over IPC through preload

API functionality:
- Create/Close connection
- Get/Send/Subscribe to events
- Get/Subscribe to connection readyState

Closes INS-1696

### DIFF
--- a/packages/insomnia/src/preload.ts
+++ b/packages/insomnia/src/preload.ts
@@ -68,7 +68,8 @@ const webSocketConnection = {
   },
 };
 
-const main: Window['main'] & { webSocketConnection: typeof webSocketConnection } = {
+export type WSConnection = typeof webSocketConnection; // using 'WS' because main/network/websocket.ts already has WebSocketConnection reserved.
+const main: Window['main'] & { webSocketConnection: WSConnection } = {
   restart: () => ipcRenderer.send('restart'),
   authorizeUserInWindow: options => ipcRenderer.invoke('authorizeUserInWindow', options),
   setMenuBarVisibility: options => ipcRenderer.send('setMenuBarVisibility', options),
--- a/packages/insomnia/src/preload.ts
+++ b/packages/insomnia/src/preload.ts
@@ -1,6 +1,74 @@
-import { contextBridge, ipcRenderer } from 'electron';
+import { contextBridge, ipcRenderer, IpcRendererEvent } from 'electron';
 
-const main: Window['main'] = {
+import type { WebsocketEvent } from './main/network/websocket';
+
+const webSocketConnection = {
+  create: (options: { requestId: string }) => {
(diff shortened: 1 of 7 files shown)


packages/insomnia/src/preload.ts
```diff
@@ -1,6 +1,74 @@
...
+
+      ipcRenderer.on(channel, onNewEvent);
+
+      const unsubscribe = () => {
+        ipcRenderer.off(channel, onNewEvent);
+      };
+
+      return unsubscribe;
+    },
+    send(options: { requestId: string; message: string }) {
```
marckong, line 62: nit pick - I wonder if this returns Promise<void>

packages/insomnia/src/preload.ts
```diff
@@ -12,6 +80,7 @@ const main: Window['main'] = {
     ipcRenderer.on(channel, listener);
     return () => ipcRenderer.removeListener(channel, listener);
   },
+  webSocketConnection,
```
marckong, line 84: non-blocker - 
MainIpcBridge interface does not seem to have this interface defined yet.

I am also wondering if it would make sense to name it as webSocketAPI or just give an entirely new name space for this.

```typescript
const WebSocketAPI: Window['WebSocketAPI'] = { ...props };
if (process.contextIsolated) {
  contextBridge.exposeInMainWorld('main', main);
  contextBridge.exposeInMainWorld('dialog', dialog);
  contextBridge.exposeInMainWorld('app', app);
  contextBridge.exposeInMainWorld('shell', shell);
  contextBridge.exposeInMainWorld('WebSocketAPI', WebSocketAPI);
} else {
  window.main = main;
  window.dialog = dialog;
  window.app = app;
  window.shell = shell;
  window.WebSocketAPI = WebSocketAPI;
}
```
However I suppose we may have wanted to explicitly indicate it is from main. In such case, window.main.propertyName also makes sense.

packages/insomnia/src/ui/redux/modules/entities.ts
```diff
@@ -100,7 +100,7 @@ export const initialEntitiesState: EntitiesState = {
   protoDirectories: {},
   grpcRequests: {},
   grpcRequestMetas: {},
-  websocketRequests: {},
+  webSocketRequests: {},
```
marckong, line 103: Awesome this was what the model missed.

marckong (approved):
Left minor comments but this looks great to me! I tested with a high velocity ws traffic as well. I am wondering if we should merge this now and add unit tests in a follow up PR, so I can continue on the INS-1701 path.

PR #5046, opened 2022-08-04
[INS-1701] Create/Close Websocket Connection

marckong:
- Adding functionalities to connect and disconnect
- Integrating the IPC bridge for WebSocket with React hooks; useWSControl, useWSReadyState

| Before Connection | Description |
| --- | ----------- |
| image | Connect button is enabled by default |
| image | Connect button does not establish connection unless the url field is valid |
| image | when the field is filled, pressing 'Enter' key or clicking  'Connect' button can establish the connection |

The button is not disabled when the url field is empty in order to be consistent with the existing pattern.

| After Connection | Description |
| --- | ----------- |
| image | the url field is read-only |


gatzjames (approved):
🚀

marckong:
Keeping ReadyState enum for now, and WebSocket doesn't explicitly expose the enum for ready state.

PR #5048, opened 2022-08-05
Display WebSocket requests in the sidebar

gatzjames:
Highlights:
- Display WebSocket Requests in the sidebar.
- WebSocket Requests can be renamed.
- WebSocket Requests can be re-ordered or put inside folders (RequestGroups).
- Updates types and adds guards for places where we use a generic request.

ws-sidebar

Closes INS-1698


marckong (approved):
I ran a couple smoke test cases. LGTM!

PR #5050, opened 2022-08-05
[INS-1693] Add WS echo server for smoke tests

filfreire:
Closes INS-1693

This PR adds a simple WS echo server for smoke tests into our existing express server. 
It's available behind `ws://localhost:4010` and any other dummy path, e.g. you can start a connection to `ws://localhost:4010/echo` and it will still work.

### DIFF
--- a/packages/insomnia-smoke-test/package-lock.json
+++ b/packages/insomnia-smoke-test/package-lock.json
@@ -22,6 +22,7 @@
 				"@types/oidc-provider": "^7.8.1",
 				"@types/ramda": "^0.27.45",
 				"@types/uuid": "^8.3.4",
+				"@types/ws": "^8.5.3",
 				"concurrently": "^7.0.0",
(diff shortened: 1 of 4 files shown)


marckong (approved):

dimitropoulos (commented):
At the time of writing, this does not function as an echo sever.  If you send it text, it responds with binary data (i.e. it does not echo).





(interacting with this server also crashes websocket king):

marckong:
Confirming the echo with text works as expected with the most recent commit. image

dimitropoulos:
I have an approval and I can see that we have a preexisitng break on the websockets feature branch (e.g. https://github.com/Kong/insomnia/runs/7734126701) that this is breaking on, so I'm going to merge since this has been reviewed and functionally tested and approved.

marckong:
@dimitropoulos this PR will fix it. https://github.com/Kong/insomnia/pull/5064

PR #5052, opened 2022-08-06
[INS-1702] WebSocket Send Message

marckong:
#### WebSocketClientContext
This context is created to separate the actual electron ipc bridge for the websocket interface from the UI components. 

In this change,
- added hooks to operate websocket controls

#### UI
@gatzjames, @erickrico , and myself discussed that we would put the most rudimentary UI for the first iteration until we have a concluded design for the message preset.
image
- added send message panel


packages/insomnia/src/ui/context/nedb-client/use-get-qhere-query.ts
```diff
@@ -0,0 +1,62 @@
...
+    const changeHandler = (_: IpcRendererEvent, [operation, entity]: ChangeBufferEvent<T>) => {
+      // TODO: add more operation case handling
+      if (operation === 'update') {
+        setData(entity);
+        setLoading(false);
+      }
+    };
+
+    const unsubscribe = dbClient.onChange(channel, changeHandler);
+    return unsubscribe();
```
gatzjames: Not sure if you should call this or just return it?

gatzjames (approved):
Code looks good! I would double check the unsubscribe fn in useEffect. Loved the nedb query/mutation implementation especially how straightforward testing becomes because of it!

packages/insomnia/src/ui/context/nedb-client/use-get-qhere-query.ts
marckong: you were right! It just needed to be called only!

marckong:
@dimitropoulos and I had a chat and we may have to revisit the nedb context pattern another time. For now I removed it to keep delivering the websocket path.

PR #5054, opened 2022-08-08
[INS-1703] Display WebSocket messages - first pass

gatzjames:
Adds a view for WebSocket messages to the response pane.

Highlights:
- Different message types have different presentation
- Selecting a message displays it in a detail view

Next Steps (probably another PR):
- The query/subscribe api for messages needs pagination to handle race conditions
- The list view needs virtualisation for handling many messages
- Iconography and design can be improved

| When a WebSocket Event is received | Description |
| --- | ----------- |
| image | Shows messages |


marckong (approved):
Locally ran it and tested. It looks good as the first pass and I will continue on this. Merging it for integrating with send websocket purpose.

PR #5063, opened 2022-08-09
adds graphql behind oidc endpoint

dimitropoulos:
Many issues have arisen (e.g. https://github.com/Kong/insomnia/issues/3013) that requires us to test a GraphQL endpoint protected by OIDC.  Today on the stream we did exactly that, and since it was so small to implement, I thought it'd be best to just make a PR now.


dimitropoulos:
Turns out this didn't work like I thought it did.  See the stream https://www.youtube.com/watch?v=q9fuLqihzxk near the end to see why.

PR #5074, opened 2022-08-12
[INS-1703] View WebSocket Messages

marckong:
image

This PR is about rudimentary functionality with presentable styling. Since the design is evolving and the table view is most likely going to be refactored with virtualized list, I will not put so much effort into the design perfect state.

NOTE:
There was some performance hit with low latency traffic like Coinbase and I tried on the following options as quick spikes for the next pass;
- react-window virtualized list
- cursor based pagination (in-memory)
- canvas list upon subscription
- no renderer interactions at all with main

None of them seemed to really have addressed the issue (I could have done spikes incorrectly). Therefore my suspect was that the low latency operation really impacted the Main process performance, so I took another spike on forking a child process to handle websocket related operation. That seemed most performant, but not including the spike on this ticket.


packages/insomnia/src/ui/components/websockets/event-log-table.tsx
```diff
@@ -0,0 +1,233 @@
+import { SvgIcon } from 'insomnia-components';
```
marckong, line 2: Moved the component into its own file and added icons only

packages/insomnia/src/ui/components/websockets/event-log-view.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FC } from 'react';
+
+import { WebsocketEvent } from '../../../main/network/websocket';
+import { CodeEditor } from '../codemirror/code-editor';
+interface Props {
+  event: WebsocketEvent;
+}
+export const EventLogView: FC<Props> = ({ event }) => {
```
marckong, line 8: Moved the component into its own file from WebSocketResponsePane

packages/insomnia/src/ui/context/websocket-client/use-ws-connection-events.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from 'react';
+
+import { WebsocketEvent } from '../../../main/network/websocket';
+
+export function useWebSocketConnectionEvents({ requestId }: { requestId: string }) {
```
marckong, line 6: Moved into its own file

packages/insomnia/src/ui/components/websockets/websocket-response-pane.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+
+  useEffect(() => {
+    setSelectedEvent(null);
+  }, []);
+
+  return (
+    <Pane type="response">
+      <PaneHeader>
+        Response Pane Header
```
filfreire: Don't forget to remove this

packages/insomnia/src/ui/components/websockets/event-log-view.tsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <CodeEditor
+      hideLineNumbers
+      mode={'text/plain'}
+      defaultValue={JSON.stringify(event)}
```
filfreire, line 13: I wonder, do we want the whole `event` object or just the `event.data`?

image

packages/insomnia/src/ui/components/websockets/event-log-table.tsx
```diff
@@ -0,0 +1,247 @@
...
+  textOverflow: 'ellipsis',
+});
+const TableCellIconWrapper = styled.div({
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center',
+});
+
+const Timestamp: FC<{ time: Date | number }> = ({ time }) => {
+  const date = format(time, 'hh:mm:ss');
```
filfreire: I wonder a couple of things, but I guess I would classify as nitpicks:
- should we, or NOT, format this according to user's local timezone, or default to UTC
- or somewhere else mention that this will be already in their timezone

And if it's the latter, then I think we have a small bug, for example in my case, a timestamp is getting printed like:
`05:03:31` but I would expect it to be printed like `17:03:31`

filfreire (approved):
_It was revealed to me in a dream that I should approve this PR_

Just kidding, some comments bellow, nothing super blocking. Marking as approved so it's not blocking any merge into the `websockets` branch later today or on Monday.

packages/insomnia/src/ui/components/websockets/event-log-table.tsx
filfreire: so probably instead of `hh:mm:ss` it should maybe be `HH:mm:ss` ([more about this here](https://stackoverflow.com/questions/62546442/what-is-the-difference-in-hmmss-and-hhmmss-in-time-formats))

packages/insomnia/src/ui/components/websockets/event-log-view.tsx
marckong: That's a good point. I think in the future we may need to differentiate what we display;
- message event
- connect event
- close event
- ping event
- pong event
and so on.

PR #5094, opened 2022-08-18
INS-1788: Add control flow to improve responsiveness.

johnwchadwick:
This prevents events from flooding the UI thread; currently, when a lot of events are coming in, they come in one-by-one but completely block the UI. This causes events to come in at a much faster rate, although for now the UI gets very sluggish quickly. (When the viewport is virtualized, it should allow for events to come in at a very fast rate; right now, if you limit the rendering to a fixed number of events, there is no noticeable slowdown. So we're bottlenecked on rendering, now.)

### DIFF
--- a/packages/insomnia/src/preload.ts
+++ b/packages/insomnia/src/preload.ts
@@ -43,12 +43,12 @@ const webSocketConnection = {
     },
     subscribe: (
       options: { responseId: string },
-      listener: (webSocketEvent: WebsocketEvent[]) => any
+      listener: (webSocketEvents: WebsocketEvent[]) => any
     ) => {
       const channel = `webSocketRequest.connection.${options.responseId}.event`;
 
-      function onNewEvent(_event: IpcRendererEvent, webSocketEvent: WebsocketEvent[]) {
-        listener(webSocketEvent);
+      function onNewEvent(_event: IpcRendererEvent, webSocketEvents: WebsocketEvent[]) {
+        listener(webSocketEvents);
       }
 
       ipcRenderer.on(channel, onNewEvent);
--- a/packages/insomnia/src/preload.ts
+++ b/packages/insomnia/src/preload.ts
@@ -43,11 +43,11 @@ const webSocketConnection = {
     },
     subscribe: (
       options: { responseId: string },
-      listener: (webSocketEvent: WebsocketEvent) => any
+      listener: (webSocketEvent: WebsocketEvent[]) => any
     ) => {
       const channel = `webSocketRequest.connection.${options.responseId}.event`;
 
-      function onNewEvent(_event: IpcRendererEvent, webSocketEvent: WebsocketEvent) {
+      function onNewEvent(_event: IpcRendererEvent, webSocketEvent: WebsocketEvent[]) {
         listener(webSocketEvent);
       }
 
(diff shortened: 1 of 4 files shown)


johnwchadwick:
Rebased.

johnwchadwick:
If you need a test case, the following works fairly well.

Connect to Coinbase: `wss://ws-feed.exchange.coinbase.com/`

Subscribe to a couple of instruments:

```json
{
    "type": "subscribe",
    "product_ids": [
        "ETH-USD",
        "ETH-EUR"
    ],
    "channels": [
        "level2",
        "heartbeat",
        {
            "name": "ticker",
            "product_ids": [
                "ETH-BTC",
                "ETH-USD"
            ]
        }
    ]
}
```

Before this PR: The events roll in one-by-one, but not fast enough to keep up with the stream. The UI still renders and you can scroll, but the JS thread never gets a chance to do anything, so it's impossible to close the connection.

After this PR: The events roll in at a variable rate. The UI updates get more sluggish due to increasing render times, but the UI continues to respond even with long frame times, making it possible to close the connection. (If you limit the viewport to a few hundred events, the UI remains consistently responsive, suggesting that virtualizing the viewport would make this much faster. I also suspect not using table layout would as well, since table layout tends to be fairly slow.)

DMarby (approved):
When tested locally I can confirm that the UI no longer completely freezes up as it did previously, I am still able to disconnect/interact with the UI although slower then when under no load.

packages/insomnia/src/preload.ts
```diff
@@ -43,11 +43,11 @@ const webSocketConnection = {
     },
     subscribe: (
       options: { responseId: string },
-      listener: (webSocketEvent: WebsocketEvent) => any
+      listener: (webSocketEvent: WebsocketEvent[]) => any
     ) => {
       const channel = `webSocketRequest.connection.${options.responseId}.event`;
 
-      function onNewEvent(_event: IpcRendererEvent, webSocketEvent: WebsocketEvent) {
+      function onNewEvent(_event: IpcRendererEvent, webSocketEvent: WebsocketEvent[]) {
```
marckong: nit pick here - I suppose now it may need to be webSocketEvents?

packages/insomnia/src/ui/context/websocket-client/use-ws-connection-events.ts
```diff
@@ -30,12 +30,19 @@ export function useWebSocketConnectionEvents({ responseId }: { responseId: strin
...
+        events => {
           if (isMounted) {
-            setEvents(events => events.concat([event]));
+            setEvents(allEvents => allEvents.concat(events));
           }
+
+          // Wait to give the CTS signal until we've rendered a frame.
+          // This gives the UI a chance to render and respond to user interactions between receiving events.
+          // Note that we do this even if the component isn't mounted, to ensure that CTS gets set even if a race occurs.
+          window.requestAnimationFrame(clearToSend);
```
marckong, line 41: :+1:

marckong (approved):
Brilliant! I locally tested, and it works as expected. I noticed that there is still a little bit of slow interaction that youve mentioned during our call, but this is a great improvement in the performance! Amazing!

packages/insomnia/src/preload.ts
johnwchadwick: Fixed.

PR #5102, opened 2022-08-23
Display response headers

gatzjames:
Adds a tab that displays the response headers

Screen Recording 2022-08-23 at 10 32 54

Closes INS-1806


jackkav (approved):
slick 🏎️

PR #5107, opened 2022-08-23
[INS-1656] fixes bug closes

kreosus:
Closes  [INS-1656]

### DIFF
--- a/packages/insomnia/src/ui/css/components/forms.less
+++ b/packages/insomnia/src/ui/css/components/forms.less
@@ -382,3 +382,7 @@ label {
     }
   }
 }
+
+button {
+  cursor: pointer;
+}


kreosus:
All arrows are now represented by the "Pointing Icon"

packages/insomnia/src/ui/css/components/forms.less
```diff
@@ -382,3 +382,7 @@ label {
     }
   }
 }
+
+button {
```
marckong, line 386: We may want to wrap this class style under a component class name, and make sure this style does not leak.

wongstein:
Can confirm the PR does what it purports
pr_working

This is what latest GA does:
current_GA

wongstein (approved):
QA approval

packages/insomnia/src/ui/css/components/forms.less
gatzjames: This is rare but I think this is a truly global style since we want to default all buttons to use this cursor.
We might want to extend the selector to apply to aria role="button" but couldn't find any such elements for now

gatzjames (approved):
Looks good!

jackkav (approved):
🚀

marckong (approved):

filfreire (approved):

filfreire:
Merged without requirements, GitHub Actions was screwed

PR #5112, opened 2022-08-25
Add client certificate support to websocket connections

DMarby:
To facilitate testing this, I've also added a `https`/`wss` server to the smoke-test server, which will refuse any connections that does not present a valid client certificate signed by the local CA, as well as certs to accompany it. These certs were generated using [mkcert](https://github.com/FiloSottile/mkcert). 

In order to test this, disable `Validate Certificates` under `Request/Response` in the app preferences, and then set up a client certificate in the collection settings using the `client.localhost-client.pem` and `client.localhost-client-key.pem` files: 
image

After doing so, create a request pointing towards `wss://localhost:4011` and attempt to connect/send messages/etc:
image

Fixes INS-1809


jackkav (approved):
Wow nice job, code looks great! 🚀

PR #5116, opened 2022-08-29
INS-1785: Add Inso NPM package test

davidfluck-kong:
This changeset adds a test to catch a previous regression involving the built version of the Inso package. I've confirmed that it fails for the previous behavior, and succeeds for the current fixed behavior.

Closes INS-1785

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "inso-start": "npm start --prefix packages/insomnia-inso",
     "inso-package": "npm run build:sr --prefix packages/insomnia && npm run package --prefix packages/insomnia-inso",
     "inso-package:artifacts": "npm run artifacts --prefix packages/insomnia-inso",
+    "inso-test": "npm run bootstrap && npm run test:npm --prefix packages/insomnia-inso",
     "watch:app": "npm run build:main.min.js --prefix packages/insomnia && npm run start:dev-server --prefix packages/insomnia",
     "app-start": "npm start --prefix packages/insomnia",
     "app-build": "npm run build --prefix packages/insomnia",
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "inso-start": "npm start --prefix packages/insomnia-inso",
     "inso-package": "npm run build:sr --prefix packages/insomnia && npm run package --prefix packages/insomnia-inso",
     "inso-package:artifacts": "npm run artifacts --prefix packages/insomnia-inso",
-    "inso-test": "npm run bootstrap && npm run test:npm --prefix packages/insomnia-inso",
+    "test:bundled-inso": "npm run bootstrap && npm run test:npm --prefix packages/insomnia-inso",
     "watch:app": "npm run build:main.min.js --prefix packages/insomnia && npm run start:dev-server --prefix packages/insomnia",
     "app-start": "npm start --prefix packages/insomnia",
     "app-build": "npm run build --prefix packages/insomnia",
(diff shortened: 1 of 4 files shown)


packages/insomnia-inso/package.json
```diff
@@ -27,6 +27,7 @@
     "test": "jest --runInBand",
     "test:watch": "npm run test -- --watch",
     "test:snapshots": "npm run build && npm run test -- -u",
+    "test:npm": "./src/scripts/test-inso.sh",
```
jackkav:
```suggestion
    "test:npm": "./bin/inso run test "Sample Specification" -e "OpenAPI env" --src src/db/fixtures/git-repo",
```
would also work.

package.json
```diff
@@ -23,6 +23,7 @@
     "inso-start": "npm start --prefix packages/insomnia-inso",
     "inso-package": "npm run build:sr --prefix packages/insomnia && npm run package --prefix packages/insomnia-inso",
     "inso-package:artifacts": "npm run artifacts --prefix packages/insomnia-inso",
+    "inso-test": "npm run bootstrap && npm run test:npm --prefix packages/insomnia-inso",
```
jackkav: this test doesn't actually use npm for anything perhaps test:bundled-inso or something more specific would be clearer.

packages/insomnia-inso/src/scripts/test-inso.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+./bin/inso run test "Sample Specification" -e "OpenAPI env" --src src/db/fixtures/git-repo
```
DMarby: I think we might need to use another fixture or unit-test suite within the fixture since this one has a failing test in it, causing the ci to fail?

filfreire (commented):
@davidfluck-kong @DMarby general question - should we also add "tests" for the `generate config` command in this PR?

I reckon we could do something like copying some of the fixtures on [kubernetes](https://github.com/Kong/insomnia/tree/develop/packages/openapi-2-kong/src/kubernetes/fixtures) and [declarative-config](https://github.com/Kong/insomnia/tree/develop/packages/openapi-2-kong/src/declarative-config/fixtures) folders, and we call the `inso` binary and check the output has no diffs from the expected output.

jackkav:
@filfreire I think it may be best to restrict the scope here to test just the final build output before it goes to npm in order to keep the complexity down. 
@davidfluck-kong I am curious to see some description of the exact difference is between this test and the smoke test we have in app.test.ts. If we have this do we still need that test?
reference:
```js
  describe('run test', () => {
    it('should not fail running tests', () => {
      const { failed } = inso(
        'run',
        'test',
        srcInsoNedb,
        ['--env', 'Dev'],
        'Echo Test Suite',
        '--verbose',
      );

      expect(failed).toBe(false);
    });
  });
```

davidfluck-kong:
@jackkav My understanding is that whatever is in `app.test.ts` runs too late in the process and tests the packaged binary and not the NPM package itself. My goal was to do the simplest possible test that positively identified the bug. It could certainly run from elsewhere, and I'm happy if folks want to push changes to the branch to try to accomplish that.

packages/insomnia-inso/package.json
davidfluck-kong: Yeah, this was hanging around from when I wasn't sure if I'd need to run multiple statements, but I suppose now I don't.

DMarby (approved):

jackkav (commented):
Is there a way to do this without running bootstrap again? Does it cache? What's the delta in CI run time?

davidfluck-kong:
I don't know. I wanted to do the simplest thing I could to exercise the behavior.

jackkav (approved):

filfreire (approved):

PR #5126, opened 2022-09-01
add react-virtual to virtualize the event log view

gatzjames:
Adds virtualization to the event log to handle connections with many events.

Highlights:
- The view can now handle thousands (tested with ~.5 million events / 350mb+ event log)
- Updated the table to be a list with divs to avoid styling issues.
- react-virtual is on a v3 beta so we should keep an eye to keep it up to date

Screen Recording 2022-09-01 at 13 46 00

Closes INS-1834


marckong:
I ran it locally to test. Looking really good! I am wondering if this was expected not to add the messages received:
image

packages/insomnia/src/ui/components/websockets/event-log-view.tsx
```diff
@@ -1,18 +1,200 @@
...
-      hideLineNumbers
-      mode={'text/plain'}
-      defaultValue={JSON.stringify(event)}
-      uniquenessKey={event._id}
-      readOnly
-    />
+    <EventLog>
+      <HeadingRow>
+        <EventIconCell>
+          <div style={{ width: '13px' }} />
```
marckong, line 155: nit pick, I wonder if this inline style is required?

marckong (approved):
LGTM! Great work! I left one minor nit pick and discovered something about the timeline view, but they shouldn't be a blocker for merging this! Got virtualization done so simple!

jackkav (approved):

packages/insomnia/src/ui/components/websockets/event-log-view.tsx
gatzjames: Good one. I'll look for a better way to style this in a follow up

PR #5129, opened 2022-09-01
Remove digest auth temporarily

marckong:
As we discussed, we will add digest auth support later and remove the getAuthHeader function.


gatzjames (approved):
👍

jackkav (approved):

PR #5132, opened 2022-09-02
Set the environment for websocket responses

DMarby:
This makes the app option `Filter responses by environment` work correctly.

A side-effect of this change is that if you have an active websocket connection and switch environments (and have response filtering by environment enabled), you no longer see the active request history.
I believe we have upcoming tasks to potentially prompt users to close the connection when switching between requests which would resolve this, but if anyone have any thoughts on alternative approaches, eg closing connections on environment change, I'm happy to implement it.


johnwchadwick (approved):
Does what it says on the tin. LGTM.

filfreire (approved):

PR #5134, opened 2022-09-02
Implement nunjucks rendering for websocket urls, authentication, headers and urls

DMarby:
To avoid issues with the history/response pane et-al when switching environments, we now close all connections on environment changes.

Closes INS-1823
Closes INS-1849

### DIFF
--- a/packages/insomnia/src/main/network/websocket.ts
+++ b/packages/insomnia/src/main/network/websocket.ts
@@ -19,6 +19,7 @@ import { generateId } from '../../common/misc';
 import { websocketRequest } from '../../models';
 import * as models from '../../models';
 import { Environment } from '../../models/environment';
+import { RequestAuthentication, RequestHeader } from '../../models/request';
 import type { Response } from '../../models/response';
 import { BaseWebSocketRequest } from '../../models/websocket-request';
 import { getBasicAuthHeader } from '../../network/basic-auth/get-header';
@@ -120,7 +121,13 @@ const parseResponseAndBuildTimeline = (url: string, incomingMessage: IncomingMes
 
 const createWebSocketConnection = async (
   event: Electron.IpcMainInvokeEvent,
-  options: { requestId: string; workspaceId: string }
+  options: {
+    requestId: string;
+    workspaceId: string;
+    url: string;
+    headers: RequestHeader[];
+    authentication: RequestAuthentication;
+  }
 ): Promise<void> => {
   const existingConnection = WebSocketConnections.get(options.requestId);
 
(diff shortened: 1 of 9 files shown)


packages/insomnia/src/main/network/websocket.ts
```diff
@@ -150,18 +157,17 @@ const createWebSocketConnection = async (
...
 
-    // @TODO: Render nunjucks tags in these headers
     const reduceArrayToLowerCaseKeyedDictionary = (acc: { [key: string]: string }, { name, value }: BaseWebSocketRequest['headers'][0]) =>
       ({ ...acc, [name.toLowerCase() || '']: value || '' });
-    const headers = request.headers;
-    if (request.authentication.disabled === false) {
-      if (request.authentication.type === AUTH_BASIC) {
-        const { username, password, useISO88591 } = request.authentication;
+    const headers = options.headers;
+    if (options.authentication.disabled === false) {
```
marckong, line 163: I find this approach meet the separation of concerns more than currently querying the request model in the websocket context (in the main process currently). However, it would be helpful if we can get some explanation of reasoning on passing request related objects as options here.

packages/insomnia/src/ui/components/websockets/action-bar.tsx
```diff
@@ -90,13 +94,42 @@ const ConnectionCircle = styled.span({
...
+      headers: rendered.headers,
+      authentication: rendered.authentication,
+    });
   };
 
+  const uniquenessKey = `${environmentId}::${request._id}`;
+
+  // TODO: Use getAutocompleteConstants to get existing websocket URLs?
+
+  // TODO: How do we handle if you switch to an environment where the variable no longer resolves, and the connection is active?
```
marckong: I suppose when an environment changes, it may need to disconnect the socket connection?

packages/insomnia/src/ui/components/websockets/action-bar.tsx
```diff
@@ -90,13 +94,42 @@ const ConnectionCircle = styled.span({
...
+  // Close the websockets on environment change always, since it also messes with history?
```
marckong: makes sense!

packages/insomnia/src/ui/components/websockets/action-bar.tsx
```diff
@@ -107,15 +140,19 @@ export const WebSocketActionBar: FC<ActionBarProps> = ({ requestId, workspaceId,
...
       <Form aria-disabled={isOpen} id="websocketUrlForm" onSubmit={handleSubmit}>
-        <Input
-          name="websocketUrlInput"
-          disabled={readyState === ReadyState.OPEN}
-          placeholder="wss://example.com/chat"
-          defaultValue={defaultValue}
-          onChange={onChange}
-        />
+        <StyledUrlBar>
+          <OneLineEditor
```
marckong, line 156: awesome!

packages/insomnia/src/ui/components/websockets/action-bar.tsx
```diff
@@ -90,13 +94,42 @@ const ConnectionCircle = styled.span({
...
+      authentication,
+    }, renderContext);
+
+    // TODO: Handle error in rendering
+
+    window.main.webSocket.create({
+      requestId: request._id,
+      workspaceId,
+      url: rendered.url,
+      headers: rendered.headers,
```
marckong: Now this explains why we are passing them in "options" in the method. I think this also removes the necessity of querying websocket model inside the main process. Great thought!

marckong:
Tested locally on the input url bar with the environment variables.
image

As you noted in the code comments, switching environments still keeps the connection alive despite the invalidity of the environment variables. I suppose we will address this in a subsequent PR?
image
image

packages/insomnia/src/main/network/websocket.ts
DMarby: The reason for these values being passed in as options is that nunjucks rendering can only be done in the renderer process currently, as our tag/nunjucks/plugin systems depend on being in the renderer, which means we can't do it in `websocket.ts` as that runs in the main process.

packages/insomnia/src/ui/components/wrapper-debug.tsx
```diff
@@ -77,6 +77,13 @@ export const WrapperDebug: FC<Props> = ({
 
   const isTeamSync = isLoggedIn && activeWorkspace && isCollection(activeWorkspace) && isRemoteProject(activeProject) && vcs;
 
+  // Close all websocket connections when the active environment changes
+  useEffect(() => {
+    return () => {
+      window.main.webSocket.closeAll();
+    };
+  }, [activeEnvironment?._id]);
```
DMarby, line 85: Maybe worth noting that this will also trigger on switch to another view (eg design/test) and back to the dashboard.

filfreire (approved):

marckong (approved):
LGTM! Thank you for explaining!

jackkav (approved):
Nice.

PR #5135, opened 2022-09-03
Add close logic when switching environments

marckong:
- adding a useEffect hook to check if it is different environment id, and ready state is open, close the connection.

There could be a better way to achieve it, but this does the job. However, it still does not refresh the timeline. For that, it may need the parent to flush out changes above the request pane and response pane when environment id changes.

image
image


filfreire (commented):
@marckong heads up, I tested this locally yesterday and it seemed like it was working fine 👍 

Not marking it as approved for now because I think before merging we should double-check with the rest of the team and @subnetmarco what should the UX be for closing connections indirectly.

It's not clear, at least for me right now, if open connections should indeed close when:

- Switching environments
- Switching between requests
- Switching back to Dashboard or to another workspace (edge-case we need to account for)

PR #5136, opened 2022-09-04
Fixes #2241 by adding a JSON Linting CodeMirror helper that renders the contents before linting

### LINKED ISSUE #2241: Template tags result in JSON Parse error when used as number
**Describe the bug**
Sorry if this is a duplicate of an existing issue. This issue is related to https://github.com/Kong/insomnia/issues/973 but with a JSON body instead of a Query.

In my case I'm attempting to send a POST request to API that requires time to be provided as a number and not a string. If I use CTRL+Space and select the Unix Timestamp tag, the query correctly sends but the JSON parser will have an error due to the template syntax not being valid json.

For example:

```
[
	{
		"name": "blah",
		"interval": 10,
		"value": 100,
		"time": {% now 'unix', '' %}
	}
]
```

Results in this parse error:
image

Is pre-evaluating Template tags before the string is passed to the linter at all an option, or would that be dangerous/slow? I can see how for very large bodies with lots of tags that could be slow since the linter might run every time you make a change.

 - OS: Windows 10
 - Installation Method: Download from Website
 - App Version:
image

**Additional context**

bbbco:
changelog(Fixes): Fixed an issue where certain template tags would result in JSON Parse error when used as number

Closes #2241 by adding a JSON linting helper to CodeMirror that renders the contents before linting. 

Previously, we could not use the Unix timestamp template as an unquoted value, or Faker Date or Faker JSON template due to the JSON linter expecting the values to be quoted.

image

With this PR, we are attempting to render the nunjuck template values before linting JSON objects. This enables us to validly use the template objects ...

image

... but still errors appropriately:

image

---

~Note, this still broken in the Environment editor, primarily due to the handling of ordering the JSON keys. Will have to address in a future PR.~
Should not be an issue in the Environment editor when used in conjunction with the `JSONPath` filter.

### DIFF
--- a/packages/insomnia/src/global.d.ts
+++ b/packages/insomnia/src/global.d.ts
@@ -15,8 +15,6 @@ declare const __DEV__: boolean;
 declare namespace NodeJS {
   interface Global {
     __DEV__: boolean;
-    /** this is required by codemirror/addon/lint/json-lint */
-    jsonlint: any;
     /** this is required by codemirror/addon/lint/yaml-lint */
     jsyaml: any;
   }
--- a/packages/insomnia/src/jsonlint.d.ts
+++ b/packages/insomnia/src/jsonlint.d.ts
@@ -1,2 +1,22 @@
 declare module 'jsonlint-mod-fixed' {
+
+    interface ParseErrorHash {
+        expected?: string[];
+        line?: number;
+        loc?: {
+            first_column: number;
+            first_line: number;
+            last_column: number;
+            last_line: number;
+        };
+        message?: string;
+        text?: string;
+        token?: string | null;
+    }
+
+    export function parse(input: string): string;
+    export namespace parser {
+        export function parseError(str: string, hash: ParseErrorHash): void;
+    }
+
 }
--- a/packages/insomnia/src/templating/utils.ts
+++ b/packages/insomnia/src/templating/utils.ts
@@ -87,7 +87,7 @@ export function normalizeToDotAndBracketNotation(prefix: string) {
 }
 
 /**
- * Parse a Nunjucks tag string into a usable abject
+ * Parse a Nunjucks tag string into a usable object
  * @param {string} tagStr - the template string for the tag
  */
 export function tokenizeTag(tagStr: string) {
--- a/packages/insomnia/src/ui/components/codemirror/base-imports.ts
+++ b/packages/insomnia/src/ui/components/codemirror/base-imports.ts
@@ -38,22 +38,6 @@ import 'codemirror/addon/selection/selection-pointer';
 import 'codemirror/addon/display/placeholder';
 import 'codemirror/addon/lint/lint';
 
-declare global {
-  // eslint-disable-next-line no-var -- necessary, let will not work here
-  var jsonlint: unknown;
-}
-
-/**/
-/**
- * Unfortunately, the CodeMirror addon for linting makes use of a pattern whereby linting dependencies are required to be attached to `window` (i.e. `global`) at runtime.
- * For that reason, if you search our codebase, you will not find anywhere where these imports are used.
- */
-/**/
-// for the code that uses this json parser, see https://github.com/codemirror/CodeMirror/blob/master/addon/lint/json-lint.js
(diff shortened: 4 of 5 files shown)


filfreire (approved):
Hi @bbbco thanks for contributing this!

I managed to test this out locally and seems to be working ok now:

- Before:
before

- After:
after

Tagging someone else from @Kong/team-insomnia to also double-check this PR.

bbbco:
Thanks so much @filfreire !

I did have a question though, I was hoping you could help address. 

As I mentioned in my OP, I noticed that this functionality is still broken in the Environments configuration due to the way we parse the environment data as JSON to turn it into an ordered object list. Since my current use case is to support this functionality (i.e., use an unquoted value of a template tag as the value for an environment key), such as:
```
{ 
  "foo": {% response 'body', 'req_2de2c09314934ffbb491cdbb91c3becb', 'b64::JA==::46b', 'never', 60 %}
}
```
where the response body returns something like:
```
{
  "hello": [
    "world",
    "kitty",
    "barbie",
    "my name is inigo montoya"
  ],
  "goodbye": [
    "world",
    "everyone",
    "forever"
  ]
}
```
so I can reference in my requests `_.foo.hello[0]` to return the value of `world`.

---

I was thinking about pre-parsing the data inside the `getValue()` and serializing `"foo": {% response 'body', 'req_2de2c09314934ffbb491cdbb91c3becb', 'b64::JA==::46b', 'never', 60 %}` into something more parsable by `JSON.parse()` like:
```
"foo": {
  "%": "response 'body', 'req_2de2c09314934ffbb491cdbb91c3becb', 'b64::JA==::46b', 'never', 60"
}
```
but then we have to add extra logic to deserialize from that output. Seems to get messy.
Thoughts?

bbbco:
> Thanks so much @filfreire !
> 
> I did have a question though, I was hoping you could help address.
> 
> As I mentioned in my OP, I noticed that this functionality is still broken in the Environments configuration due to the way we parse the environment data as JSON to turn it into an ordered object list. Since my current use case is to support this functionality (i.e., use an unquoted value of a template tag as the value for an environment key), such as:
> 
> ```
> { 
>   "foo": {% response 'body', 'req_2de2c09314934ffbb491cdbb91c3becb', 'b64::JA==::46b', 'never', 60 %}
> }
> ```
> 
> where the response body returns something like:
> 
> ```
> {
>   "hello": [
>     "world",
>     "kitty",
>     "barbie",
>     "my name is inigo montoya"
>   ],
>   "goodbye": [
>     "world",
>     "everyone",
>     "forever"
>   ]
> }
> ```
> 
> so I can reference in my requests `_.foo.hello[0]` to return the value of `world`.
> 
> I was thinking about pre-parsing the data inside the `getValue()` and serializing `"foo": {% response 'body', 'req_2de2c09314934ffbb491cdbb91c3becb', 'b64::JA==::46b', 'never', 60 %}` into something more parsable by `JSON.parse()` like:
> 
> ```
> "foo": {
>   "%": "response 'body', 'req_2de2c09314934ffbb491cdbb91c3becb', 'b64::JA==::46b', 'never', 60"
> }
> ```
> 
> but then we have to add extra logic to deserialize from that output. Seems to get messy. Thoughts?

Looks like this is not a concern after further investigations. I can just set the json response to a quoted variable, and then use the `JSONPath` Template filter to pull out the data I need in my requests.

---

The changes in this PR are still very necessary! Please let me know what else I can do to get this merged in.

bbbco:
@marckong any chance we can get a review from the team on this? Would love to get this in the upcoming release. :)

bbbco:
Fyi, still waiting for this to be pulled in to address an issue I've got opened in my Insomnia Plugin for Faker: https://github.com/bbbco/insomnia-plugin-faker/issues/12
Someone else just commented today on the issue agreeing that this is blocking to them. Would be great to see this fix in the next version of Insomnia!

JwanKhalaf:
Can we get this merged in please? This will fix a few issues we're having.

jackkav (approved):
Thanks @bbbco 🚀 🎸

filfreire:
@bbbco this will be merged and hopefull will go out in the next beta release iteration

PR #5138, opened 2022-09-06
[INS-1843] Add more checks to WS smoke test

filfreire:
Closes INS-1843

In this PR we add simple coverage for supported auth methods of WS connections as well as for handling redirects. 

Not done in this PR: handling "negative"/401 http status scenarios or different scenarios, e.g. Trying to close WS connection in different ways. That will need to be added once we finalize UX for how connections should be closed.


jackkav (approved):
noice

PR #5146, opened 2022-09-07
INS-1844: Implement basic event view functionality.

johnwchadwick:
First pass at a useful event view. This handles only message events, does not treat binary data specially, and only handles pretty printing/highlighting for JSON-formatted messages. Still, it's a start.

There's some things I had difficulty in untangling, so I cut it short and did things a bit simpler. For example, I would've liked to reuse the existing PreviewMode dropdown, but it proved too difficult, so this one contains a Websocket-specific dropdown instead.

I left TODOs for the issues that should be addressed.


filfreire (approved):
Tested it locally and LGTM!

jackkav (approved):
minor css fix though on the left side of the Preview button to line it up with the Events one above.

image

PR #5148, opened 2022-09-07
Reverse the order of WebSocket events

gatzjames:
Reverse the order of WebSocket events so that the latest is on top.
Also filters by the last message timestamp so there's no duplicate events in the view.

Screen Recording 2022-09-07 at 17 32 34

Closes INS-1949


jackkav (approved):
LGTM

marckong (approved):

filfreire (approved):

PR #5154, opened 2022-09-08
[INS-1800] Remove value validity check in the WebSocket headers

marckong:
- Removed the validity check on the value when parsing headers for WebSocket

Closes [INS-1800](https://linear.app/insomnia/issue/INS-1800/support-headers-with-empty-value-for-websockets)

### DIFF
--- a/packages/insomnia/src/main/network/websocket.ts
+++ b/packages/insomnia/src/main/network/websocket.ts
@@ -142,7 +142,7 @@ const createWebSocketConnection = async (
     }
 
     const lowerCasedEnabledHeaders = headers
-      .filter(({ value, disabled }) => !!value && !disabled)
+      .filter(({ disabled }) => !disabled)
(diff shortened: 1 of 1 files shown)


filfreire (commented):
@marckong heads up there's a small bug that the PR introduces.

The functionality of allowing you to send a header with empty value is working ok ✅ 

BUT, now we are not ignoring header entries that also have no header key, we throw an error instead of filtering those out silently like we do for HTTP requests:

Screenshot 2022-09-09 at 10 33 15


Screenshot 2022-09-09 at 10 28 01
Screenshot 2022-09-09 at 10 30 39

marckong:
@filfreire amazing you catch that! I just pushed the fix for it!

image

filfreire (approved):

jackkav (approved):

PR #5162, opened 2022-09-09
[NO-TICKET] Add style for enabled status

marckong:
Before connection is established
image

After connection is established
image

### DIFF
--- a/packages/insomnia/src/ui/components/websockets/websocket-request-pane.tsx
+++ b/packages/insomnia/src/ui/components/websockets/websocket-request-pane.tsx
@@ -34,6 +34,9 @@ const SendButton = styled.button({
   ':hover': {
     filter: 'brightness(0.8)',
   },
+  ':enabled': {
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/components/websockets/websocket-request-pane.tsx
```diff
@@ -37,6 +37,9 @@ const SendButton = styled.button({
   ':hover': {
     filter: 'brightness(0.8)',
   },
+  ':enabled': {
+    background: 'var(--color-surprise)',
```
filfreire, line 38: @marckong I think the text when the button is enabled is not super clear, perhaps we should also change it's color. Here's a suggestion:
```suggestion
    background: 'var(--color-surprise)',
    color: 'var(--color-font-surprise)',
```

Before:
image

After:
image

gatzjames (approved):
🥳

PR #5192, opened 2022-09-16
Fixes 944 | Remove the check for returning more than one JSON result in Response Plugin

### LINKED ISSUE #944: [Feature Request] Allow multiple results for JSONPath in JSON bodies
- Insomnia Version: 5.16.2
- Operating System: macOS

## Details

I make heavy use of chaining requests with template tags which is very powerful. However, I find it rather restrictive when working with JSON bodies in a POST request.

### Use case

I query a REST endpoint which returns an array of elements. In the response preview, I can use a JSONPath such as `$.data[*].id` to drill down to an array of id strings.

I would like to chain this to another POST request and provide this array as the JSON body. However, when I try to do so, the request fails to render with the following error: `Returned more than one result: $.data[*].id`

It would be great if JSON arrays could be rendered within JSON bodies.

bbbco:
changelog(Fixes): Fixed an issue where the Response template wouldn't return more than one result when parsing a JSON response by path

Fixes https://github.com/Kong/insomnia/issues/944

Removes the check for returning more than one JSON result in the Response Plugin. If the result is an array of objects, the code following the removed check will just stringify the result. This allows the response to be more easily chained and parsed by the JSONPath plugin.

### DIFF
--- a/plugins/insomnia-plugin-response/__tests__/index.test.js
+++ b/plugins/insomnia-plugin-response/__tests__/index.test.js
@@ -150,7 +150,7 @@ describe('Response tag', () => {
       }
     });
 
-    it('fails on more than 1 result', async () => {
+    it('allows more than 1 result', async () => {
       const requests = [{ _id: 'req_1', parentId: 'wrk_1' }];
 
       const responses = [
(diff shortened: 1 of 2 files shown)


filfreire:
Tested locally, LGTM

image

filfreire (approved):

bbbco:
Thanks for pulling this in!

PR #5193, opened 2022-09-17
FEAT | Global workspace authentication settings

carlosmfreitas2409:
### Current Problematic Behavior
Since each request has its own authentication, the desire to change the authentication of all requests is currently not possible today. This can end up causing a headache, no matter if you have enviroments, because you may want to change some fields without having to change the enviroment.

### Expected Behavior
This Pull Request introduces a feature to configure the authentication of ALL requests with just a few taps.
Now, under the "Management" dropdown, you will have the possibility to configure the "workspace authentication".
When setting up authentication, you can go to the request and under the "Auth" dropdown, you will have the "Inherit" option, which automatically adds the same workspace authentication settings.
So when you want to change some field of authentication, all requests that have this option will change.

### Reproduction Steps
1. Open an workspace.
2. Go to debug tab.
3. Open "Management" dropdown, next to "Enviroment".
4. Click on "Authentication".

### Preview and video
You can see a video preview of this feature here:

[](https://user-images.githubusercontent.com/32655274/190877293-c2302021-605c-46dd-b34c-ab4317254b07.mp4)

image


filfreire:
Hi @carlosmfreitas2409, we're going to be revamping the way workspaces work completely, and there will be a lot of breaking changes in the future. We'll also be including global auth into new workspace model, but for now we're not going to include it.

Feel free to open other issues. Apologies in advance for not reacting sooner to the PR.

PR #5195, opened 2022-09-19
Fix pasting curl to url bar not updating the request view

gatzjames:
Use fromSync=true when storing the curl data to the database to indicate that the ui should be refreshed

Closes INS-1990


filfreire (approved):
Tested locally, LGTM 👍

PR #5209, opened 2022-09-22
Websocket request settings

gatzjames:
Closes INS-1820

Adds the settings modal for the WebSocket request:
- Rename
- Description
- Send/Store cookies
- Follow Redirects
- Move/Copy to Workspace


filfreire (commented):
Heads up @gatzjames the redirect config is not working after this change.

Seems we always redirect, even if we disable that.

You can check with something like:
```shell
curl --request GET \
  --url https://httpbin.org/absolute-redirect/5 \
  --header 'accept: text/html'
```

and see if you see on the `Timeline` the 5 redirects, or it stops at the first 302 request.

Same seems to apply for websockets for the `ws://localhost:4010/redirect` endpoint, we seem to always get the upgraded 101 status response back.

filfreire (approved):

marckong (approved):
LGTM!

jackkav (approved):

PR #5221, opened 2022-09-23
Fix/ins 1656 send button doesnt show hand icon

kreosus:


kreosus:
no action needed, clicked on wrong button

PR #5224, opened 2022-09-26
[INS-2010] update the query tab panel when we import params from url

gatzjames:
Closes INS-2010


filfreire (approved):
Tested locally, LGTM

PR #5226, opened 2022-09-27
[INS-2014] Add Move/Copy to option in gRPC request settings

filfreire:
Closes INS-2014
Closes INS-267

image

This PR adds move/copy to option to gRPC request settings.

It also handles duplicating the related proto files of the gRPC requests being moved/copied and assigning those duplicates as the related protofiles in the new workspace. 

## Tested
- Users can copy and move a single gRPC request to another collection/doc
- A user can update the new proto file associated with the new moved/copied request independently of the original gRPC request

## Side-effects
- Copying and moving a requestGroup doesn't work properly yet, the duplicate gRPC requests still point to the original protoFileId ⚠️ 
- This adds a new side-effect - if we duplicate a grpc request within the same collection, this will be duplicating proto files instead of referencing the existing one in the collection ⚠️

### DIFF
--- a/packages/insomnia/src/models/grpc-request.ts
+++ b/packages/insomnia/src/models/grpc-request.ts
@@ -1,5 +1,5 @@
 import { database as db } from '../common/database';
-import type { BaseModel } from './index';
+import { BaseModel, protoDirectory, protoFile } from './index';
 
 export const name = 'gRPC Request';
 export const type = 'GrpcRequest';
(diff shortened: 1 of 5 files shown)


filfreire:
Closing this one for now due to the side-effects:
- Copying and moving a requestGroup doesn't work properly yet, the duplicate gRPC requests still point to the original protoFileId ⚠️

We can revisit this at some point when we look at gRPC again

PR #5236, opened 2022-09-30
[INS-1979] Add disconnect dropdown

marckong:
Reflecting on the most recent discussion; we want to have it like the split button we have for the Send button
- added split border between the button and dropdown
- modified the base dropdown styles to remove some extra styles
- added e2e tests

Working state demo:
https://www.loom.com/share/7c02e4325df24533ac1830bfc36b47e8

Hover state 1
image

Hover state 2
image

Dropdown Open state
image

changelog(Improvements): Adding WebSocket Disconnect dropdown to support Disconnect All

### DIFF
--- a/packages/insomnia-smoke-test/tests/websocket.test.ts
+++ b/packages/insomnia-smoke-test/tests/websocket.test.ts
@@ -48,86 +48,14 @@ test('can make websocket connection', async ({ app, page }) => {
   await expect(statusTag).toContainText('101 Switching Protocols');
   await page.click('[data-testid="response-pane"] >> [role="tab"]:has-text("Timeline")');
   await expect(responseBody).toContainText('WebSocket connection established');
-});
 
-test('can close a websocket connection', async ({ app, page }) => {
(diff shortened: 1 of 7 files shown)


filfreire (commented):
Heads up @marckong 

- ✅ The basic parts seem to work (disconnecting one or all requests) 
- ❔ We should probably double-check UX wise with @subnetmarco if we want the Disconnect button itself to also disconnect current request, and the button would behave as a split button. I was expecting that just clicking on the arrow 🔽  part would open dropdown and `Disconnect` part would still work.

Depending on UX we would need also to fix the smoke tests - right now they consider the Disconnect button to actually disconnect instead of opening a dropdown.

filfreire (approved):
LGTM!

@marckong heads up, I simplified your previous edit of the tests. We could do the disconnect all connections in the same main websocket scenario and that way keep the test duration to a minimum, e.g. , roughly 15 secs instead of 3 times 15 secs.

gatzjames (approved):
Works great! 🚀
Tried using styled for the dropdown item but the way the dropdown is designed makes it really hard. Something we should fix on the reiteration of dropdowns in the app.

PR #5239, opened 2022-10-03
Fix/ins 1660 tooltip issues

gatzjames:
Fixes tooltip positioning and open/close state issues.

Highlights:
- Adds react-aria to make the tooltip accessible and handle positioning
- Removes the usage of insomnia-components tooltip. Kept out of scope completely removing it since it's used by other insomnia-components internals.

[React-Aria](https://react-spectrum.adobe.com/react-aria/why.html) pros/cons

changelog(Fixes): Fixed an issue with tooltip

### DIFF
--- a/packages/insomnia/package-lock.json
+++ b/packages/insomnia/package-lock.json
@@ -132,6 +132,7 @@
 				"ncp": "^2.0.0",
 				"objectpath": "^2.0.0",
 				"react": "^17.0.2",
+				"react-aria": "^3.19.0",
 				"react-dnd": "^7.4.5",
 				"react-dnd-html5-backend": "^7.4.4",
 				"react-dom": "^17.0.2",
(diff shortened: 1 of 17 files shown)


marckong:
It took some time to read upon code of react-stately as it didn't seem to have good documentation. It's a cool headless UI, I have to admit.

One interesting thing I discovered though is captured here:
https://www.loom.com/share/7595338814c7478099b2c94bbcd8153b. The video demonstrates some delay to actually pop a tooltip popover. However once it is open, you can switch between different tooltip blocks and show tooltip popovers immediately.

I wonder if we should discuss what benefits this library would entail. I am looking at it positively, but how will we incorporate theme context with this for instance? Will it ever have a conflict if we put all the design tokens into the [theme context/provider](https://styled-components.com/docs/api#themeprovider) and read from it?

marckong (approved):
Ran smoke tests and it works as expected. I wonder if we can shorten the delay time to display but that is just my personal opinion. Thanks for explanation on the call.

gatzjames:
> I wonder if we can shorten the delay time to display but that is just my personal opinion.

Good one! Moving to a function component I didn't keep the default value of 400ms.
Did some quick search in the codebase and we seem to have different delay values for our tooltips (400/500/800/1000 ms). I don't see any reason for having different delays but will keep it out of this PR's scope to fix them and do some research before updating this.

PR #5240, opened 2022-10-03
Chore: bump playwright to 1.26.1

filfreire:


marckong (approved):
It looks like from 1.23 to 1.26 that there were many big addons to the releases, compared to 1.23. - https://github.com/microsoft/playwright/releases. However, none of the new additions seem to break our tests. Therefore, it looks good to me for the merge!

PR #5244, opened 2022-10-04
[INS-2022] Initial exploration multiple test projects

filfreire:
Closes INS-2022

## About

This is a spike to allow us having different kinds of e2e checks in our project, by making better use of Playwright's [Configuration Objects and creating distinction of Projects](https://playwright.dev/docs/test-advanced#configuration-object).

For now we keep everything similar, we just make sure that our existing smoke tests are caught under the `--project=Smoke` argument when running Playwright.

All other tests we might add would be run under the default project, which will be triggered only in Release flows. That will be explored in another PR.

## Previous Context

The gist of it all is, regardless of how we name them, is to create 3 distinctions:
- checks that should run in all CI (high-confidence) - **sanity**?
- checks that cover a larger base, and port over the hands-on work of going through release checklist, and help us catch bugs in basic features - **smoke? regression?**
- checks that repro major bugs, that we want to write once and forget about it, but we dont want the same bug to reappear - **regression?**

All of then running at an E2E level - meaning - they mimic a user touching a live running instance of the app

The naming of the projects is still up to be decided. One idea would be:
```
npx playwright test --project=PR
npx playwright test --project=Release
```


package.json
```diff
@@ -29,10 +29,11 @@
     "app-package": "npm run build --prefix packages/insomnia && npm run package --prefix packages/insomnia",
     "app-bump-version": "npm run bump-version --prefix packages/insomnia",
     "preinstall": "node ./scripts/check-version.js",
-    "test:smoke:dev": "npm run test:dev --prefix packages/insomnia-smoke-test",
-    "test:smoke:build": "npm run test:build --prefix packages/insomnia-smoke-test",
-    "test:smoke:package": "npm run test:package --prefix packages/insomnia-smoke-test",
-    "test:smoke:cli": "npm run test:cli --prefix packages/insomnia-smoke-test",
+    "test:prerelease:dev": "npm run test:dev --prefix packages/insomnia-smoke-test -- --project=PreRelease",
```
filfreire: This is probably unnecessary. I added the project to the existing `smoke` aliases, but this could be our oportunity to define proper new easy to use aliases.

packages/insomnia-smoke-test/playwright.config.ts
```diff
@@ -1,6 +1,26 @@
...
+    },
+    {
+      // Broad coverage PreRelease only checks (PreRelease Only)
+      name: 'PreRelease',
+      testMatch: /.*prerelease\/.*.test.ts/,
+      retries: 0,
+    },
+    {
+      // Run all tests
+      name: 'Default',
```
filfreire: Did a quick experiment and `Default` indeed picks up all of them if we don't define the `--project` argument.

image

jackkav (approved):

PR #5245, opened 2022-10-04
[INS-1575] Remove mousemove handleActivity on Team Sync

filfreire:
Paired with @kreosus on this one. We are not sure the `this._handleUserActivity` call when folks move their mouse is no longer needed.

We did some testing, and Team Sync seemed to work fine without it.

### DIFF
--- a/packages/insomnia/src/ui/components/dropdowns/sync-dropdown.tsx
+++ b/packages/insomnia/src/ui/components/dropdowns/sync-dropdown.tsx
@@ -167,14 +167,12 @@ class UnconnectedSyncDropdown extends PureComponent<Props, State> {
         await this.refreshMainAttributes();
       }
     }, REFRESH_PERIOD);
-    document.addEventListener('mousemove', this._handleUserActivity);
   }
 
   componentWillUnmount() {
     if (this.checkInterval !== null) {
       clearInterval(this.checkInterval);
     }
-    document.removeEventListener('mousemove', this._handleUserActivity);
   }
 
   componentDidUpdate(prevProps: Props) {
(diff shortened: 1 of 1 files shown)


filfreire:
Heads up converting back to draft, we need to double check 2 things before considering merging this:
- This stuff might be still used to check if there are any remote changes to pull/push updates periodically as the user does stuff. So if someone else has pushed to a remote branch, the sync menu actually updates to show you that you have stuff to pull, without you needing to interact with the sync menu. We need to confirm this still works.
- If we do remove it, we should probably remove the entire `checkInterval` setInterval function too

packages/insomnia/src/ui/components/dropdowns/sync-dropdown.tsx
```diff
@@ -197,10 +195,6 @@ class UnconnectedSyncDropdown extends PureComponent<Props, State> {
     }
   }
 
-  _handleUserActivity() {
```
marckong, line 200: I could be wrong, but it seems that we are manually tracking some metrics based on this value here:

https://github.com/Kong/insomnia/blob/6399a971b37658079c977d6a6f40cff62af30fcf/packages/insomnia/src/ui/components/dropdowns/sync-dropdown.tsx#L110

https://github.com/Kong/insomnia/blob/6399a971b37658079c977d6a6f40cff62af30fcf/packages/insomnia/src/ui/components/dropdowns/sync-dropdown.tsx#L165

Is this something we do not want to continue doing at all?

I ran some smoke test on this and discovered this:
https://www.loom.com/share/1fa194834efc4afa8ca474ec888150ac

The video may demonstrate that it is now hiding the branch name when you go back to dashboard and come back to the collection.

filfreire: @marckong seems the hiding of the branch name is also a current bug in our latest beta, unrelated to this change.

PR #5257, opened 2022-10-10
update use-vcs hook to use a single instance and await changes

gatzjames:
Updates the vcs reference to be a singleton since we use the use-vcs hook in many places now.
Also awaits the vcs operations so that it is reflected properly in state.

Closes INS-2047


jackkav (approved):

filfreire (commented):
heads up @gatzjames, commenting here so it doesn't get lost

The fix seems to do the trick for the branch name part. There still seems to be a problem though with the icons, they don't seem to get updated on the background. So for example if there are new changes to Pull, the Pull icon still remains greyed out until we click on the dropdown.

filfreire (approved):
Nevermind the above comment, LGTM

Issue #5088, opened 2022-08-17
Crash on Fedora 36 after got "Error: SSL connect error" when working with https websites


trunglebka:
>  ℹ️ **Note from Insomnia team**: We don't have a fix for this issue yet, but please refer to the workaround fix mentioned in ~~[this comment](https://github.com/Kong/insomnia/issues/5088#issuecomment-1228651988)~~ ~~[(UPDATE) here](https://github.com/Kong/insomnia/issues/5088#issuecomment-1398675567)~~. 

## Workaround
```
# remove existing insomnia installations
sudo dnf remove insomnia

# install latest insomnia
sudo dnf install https://github.com/Kong/insomnia/releases/download/core%402022.7.5/Insomnia.Core-2022.7.5.rpm

# download cups 2.4.1 libs and copy them to insomnia
curl -O https://kojipkgs.fedoraproject.org//packages/cups/2.4.1/7.fc36/x86_64/cups-libs-2.4.1-7.fc36.x86_64.rpm
rpm2cpio cups-libs-2.4.1-7.fc36.x86_64.rpm | cpio -idmv
sudo cp ./usr/lib64/libcups* /opt/Insomnia/.

# run insomnia
insomnia
```

### Expected Behavior

Working normally 

### Actual Behavior

After sending a request to a https website like https://google.com, got an error: `Error: SSL connect error`

### Reproduction Steps

1. Send request to any https website. E.g. https://google.com
2. Send request again (any https)
3. App crash

### Is there an existing issue for this?

- [ ] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Additional Information

Timeline:
```
* Preparing request to https://google.com/
* Current time is 2022-08-17T02:41:53.423Z
* Enable automatic URL encoding
* Using default HTTP version
* Enable SSL validation
*   Trying 172.217.24.110:443...
* Connected to google.com (172.217.24.110) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* TLSv1.3 (OUT), TLS alert, internal error (592):
* OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to google.com:443 
* Closing connection 0
```

Log from console:
```
(base) ➜  ~ /opt/Insomnia/insomnia
09:41:43.262 › Running version 2022.5.0
09:41:43.271 › [electron client protocol] FAILED to set default protocol 'insomnia://'
[...]

PR #5281, opened 2022-10-14
Add release-fedora to build fedora compatible RPM

filfreire:
## [SEE UPDATE 23 January 2023](https://github.com/Kong/insomnia/issues/5088#issuecomment-1400859407)
_____ 
In this PR we add another job ~~to the release-publish~~, `release-fedora` where we build Insomnia `.rpm` with `buildDependenciesFromSource` setting from within a fedora 36 container. 

This helps provide a working workaround build that solves the issue folks are running into for #5088.

When we do builds with `buildDependenciesFromSource` set to `true` from within a Fedora 36 environment, we no longer get the SSL crash:

Screenshot 2022-10-14 at 11 46 03

TL;DR: Exploration to try and provide some sort of workaround for #5088 

 > ⚠️ This should probably not be solved in the insomnia build, what we’re doing is making it build node-libcurl from source which means it will be different from other platforms and not statically compiled. So it’ll have other potential issues depending on the environment it’s ran on, and not necessarily behave the same as other platforms. The proper solution is probably to change the build of the prebuilt node-libcurl binaries so they’re compatible instead.


.github/workflows/release-publish.yml
```diff
@@ -197,3 +197,32 @@ jobs:
...
+          docker create --name builder fedora-insomnia-builder:latest
+          docker cp builder:/insomnia/packages/insomnia/dist ./artifacts/
+          docker rm -f builder
+
+      - name: Upload artifacts to release
+        uses: xresloader/upload-to-github-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          release_id: ${{ steps.core_tag_and_release.outputs.id }}
```
filfreire: not sure if `steps.core_tag_and_release.outputs.id` picks up the proper step from the other job

.github/workflows/release-publish.yml
```diff
@@ -197,3 +197,32 @@ jobs:
...
+        with:
+          ref: ${{ env.RELEASE_BRANCH }}
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Build Insomnia from sources for Fedora
+        run: |
+          docker build --tag fedora-insomnia-builder:latest -f fedora.Dockerfile .
+          docker create --name builder fedora-insomnia-builder:latest
+          docker cp builder:/insomnia/packages/insomnia/dist ./artifacts/
```
filfreire: full path is copied into artifacts folder, needs to be adapted bellow

.github/workflows/release-publish.yml
filfreire: fixed

azzamsa:
Is this fixes will land on master? 
I can't upgrade other packages because of pinned version of cups.

Thanks for insomnia! :heart:

leonardehrenfried:
By the way, I tried the artifact from here: https://github.com/Kong/insomnia/actions/runs/3481857586

I still see the SSL connect error followed by a crash.

filfreire:
Hi @azzamsa @leonardehrenfried , while we don't have a more definitive solution, in order to unblock you in the meantime, I've compiled an experimental build on my fork which you can download and use:

https://github.com/filfreire/insomnia/releases/tag/fedora-experiment%408ae4fe71

Contains changes up to [8ae4fe71](https://github.com/Kong/insomnia/commit/8ae4fe710dc7d80fcd14265e315f5c2f179821ab) and was built using: https://github.com/Kong/insomnia/pull/5281

Please let me know if you have any issues with it.

leonardehrenfried:
@filfreire I just upgraded all of my dependencies and re-installed the build you provided. I can now indeed use insomnia again! Thank you for this.

This will be good enough for me for a while.

filfreire:
## [SEE UPDATE 23 January 2023](https://github.com/Kong/insomnia/issues/5088#issuecomment-1400859407)

PR #5327, opened 2022-10-25
Feat/project route data loading

gatzjames:
Highlights:
- Moves the Project route to use react-router's action/loader for data fetching.
- Expands the routes to contain the project and workspace
- Introduces a sync function to keep navigation state in sync between redux/react-router as a migration path.


jackkav (commented):
:rocket:

jackkav (approved):

PR #5348, opened 2022-10-28
fix/bundle plugins

filfreire:


jackkav (approved):

filfreire:
Disabled auto-merge, we might need to change package.json depending on if #5347 gets merged beforehand or not

PR #5355, opened 2022-10-31
Run full smoke test suite nightly Mon-Fri

filfreire:
This will run all smoke tests, Monday to Friday, at 1:30 UTC.

Adding this so we have recurring nightly feedback against latest develop for the entirety of the smoke tests.


jackkav (commented):
is there a way to arrange this to avoid the not logic?

filfreire:
Closing for now

PR #5365, opened 2022-11-01
fix: prevent git request size limit errors

MichaelDarr:
When using git sync to push changes, I've occasionally encountered the following error:

```
Request body larger than maxBodyLength limit
```

These changes add options to the git http client to prevent this issue.


MichaelDarr:
@filfreire or @jackkav - Could you approve workflows for this PR? Thanks!

filfreire:
Hi @MichaelDarr, just as follow-up, we've hit this sort of issue while working in https://github.com/Kong/insomnia/pull/5320 and we found out that setting `Infinity` doesn't seem to work completely as we thought it would. We'll get back to this PR once we manage to finish work on upgrading axios dependency.

MichaelDarr:
Hi @filfreire, I appreciate the update! If there's anything I can do to help out with the axios upgrade, I'd be more than happy to assist.

filfreire:
@MichaelDarr we're closing this one for now, but we'll return to this once we focus back on axios and git sync work.

PR #5395, opened 2022-11-08
Add organization routes

gatzjames:
Updates the app to include the organizations view.

Highlights:
- New dashboard UI that separates previously shown projects under organizations.
- Added new organization routes
- Updated UI page layout
- Closes INS-2127 bug, where users could not re-pull a Team Sync Collection deleted locally previously

changelog(Improvements): Added new UI improvements to the Dashboard

Closes INS-2114, INS-2115


filfreire (approved):
LGTM. 10/10 experience

image

packages/insomnia/src/ui/components/app-header.tsx
```diff
@@ -28,9 +26,10 @@ export interface HeaderProps {
 }
 
 const StyledHeader = styled.div({
+  gridArea: 'Header',
```
pavkout, line 29: @gatzjames Do we use grid templates?

pavkout (approved):

packages/insomnia/src/ui/components/app-header.tsx
gatzjames: yep it's inside the layout in root.tsx

packages/insomnia/src/ui/components/organizations-navbar.tsx
```diff
@@ -0,0 +1,178 @@
...
+          </ExternalLink>
+        </div>
+      </ModalFooter>
+    </Modal>
+  );
+};
+
+export const OrganizationsNav: FC = () => {
+  const { organizations } = useLoaderData() as RootLoaderData;
+  const { organizationId } = useParams() as { organizationId:string };
```
marckong, line 123: nit pick;
I would prefer something like this below instead of casting, but we need to disable the lint rule.
```ts
// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
type UseOrganizationParams = {
  organizationId: string;
};

const { organizationId } = useParams<UseOrganizationParams>();
```

marckong (approved):
LGTM! I ran smoke tests as well.

packages/insomnia/src/ui/components/organizations-navbar.tsx
gatzjames: Agree I'll try to tackle typesafety in routing in a follow up PR. Afaik react-router might offer some utils in the next release

PR #5398, opened 2022-11-08
Additional option "skipAnnotations" for the export of spec

mod42:
changelog(Inso CLI): Added a new `--skipAnnotations` option to `inso export spec` command. If the flag is set all `x-kong-*` annotations are removed.


mod42:
I added an additional option "skipAnnotations" for the export of spec. When we export the spec and the user wants to share the spec with others, outside the organization, the annotations reveal sensitive information about the configuration of the kong gateway, like client_ids and clients_secrets. If the flag -s is set, all x-kong-* annotations are removed. I also adjusted the tests.

filfreire:
@mod42 did you push all your changes? Only see a new test added on files here.

filfreire (approved):

PR #5399, opened 2022-11-08
Add API Key auth

Asamsig:
Supports header / query parameter in both webrequest and websocket

changelog(Improvements): Added API Key Authentication method for HTTP and WebSocket requests

### DIFF
--- a/packages/insomnia-importers/src/importers/fixtures/postman/api-key-default-v2_1-input.json
+++ b/packages/insomnia-importers/src/importers/fixtures/postman/api-key-default-v2_1-input.json
@@ -1,0 +1,44 @@
+{
+	"info": {
+		"_postman_id": "test",
+		"name": "New Collection",
+		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json",
+		"_exporter_id": "test"
+	},
+	"item": [
+		{
+			"name": "New Request",
+			"request": {
+				"auth": {
+					"type": "apikey",
+					"apikey": [
+						{
+							"key": "value",
+							"value": "test",
+							"type": "string"
+						},
+						{
+							"key": "key",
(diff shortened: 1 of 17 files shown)


Asamsig:
Amended the commit, had an unused import.

Asamsig:
Added support for `cookie` as well.

I started to look into updating the importers (specifically the OpenApi 3 importer), and it seems pretty straight forward, except it hits a limitation in the sense that we can only support ONE `securityScheme` at a time, but this is common for all authentication types in Insomnia.
I.e. if an endpoint support several security schemes, the importer will have to just pick one, it already does this for other auth types, but currently for API Key auth it actually supports several since just places them in the `header`, `query params`, and `cookie` tabs that supports lists, which the auth tab does not.

I guess it might be seen as a downgrade for some people, if we change the importer in that way, any thoughts/input?

filfreire (approved):
@Asamsig I tested this locally, LGTM, thank you for your contribution!

Regarding your [comment above](https://github.com/Kong/insomnia/pull/5399#issuecomment-1307939772):
- I think the default behavior could be kept the OpenAPI importer, meaning there's 1 or more api keys on the security schema, when requests are generated, the api keys go into headers like before 
- For Postman importer I've taken the liberty of pushing and adding it to this PR, hope you don't mind.

From my side I think this is in mergeable state, waiting for someone else from @Kong/team-insomnia to also give the 👍

Asamsig:
@filfreire Yeah, so if there is only one security scheme, we could use the new auth, but otherwise preserve the current behaviour, that seems reasonable.
I'll take a look over the next few days, and create a separate PR for that.

Awesome that you already updated the Postman importer!

gatzjames (approved):
👍 Good one! Thanks for the PR!

PR #5417, opened 2022-11-14
upgrade vite to v3

gatzjames:
Motivation: we would like v3 in order to use folder routing

Upgrades Vite to v3.

Highlights:
- The [CommonJS Externals plugin](https://github.com/xiaoxiangmoe/vite-plugin-commonjs-externals) stops working with the new api. It is now replaced with our own inspired by [Electron Renderer plugin](https://github.com/electron-vite/vite-plugin-electron-renderer).
- Swagger UI React is now moved into an internal component that uses the dist version of swagger-ui. Solves issues where peer deps get in the way of upgrading and of how the swagger-ui-react library was bundled.


packages/insomnia/src/main/window-utils.ts
```diff
@@ -63,6 +63,7 @@ export function createWindow() {
     }
   }
 
+  console.log('😭😭😭😭😭😭😭😭 [main] Creating window');
```
filfreire:
```suggestion

```

gatzjames: 😂

filfreire (approved):

jackkav (approved):
🎸

PR #5425, opened 2022-11-15
OpenAPI 3 Import: Add support for importing API Key Authentication as authentication

Asamsig:
Now imports API Key as the authentication scheme, if there is only one API Key security schema.
Otherwise it will preserve the existing behavior and use query parameters, or headers.


filfreire (approved):

jackkav (approved):
🎸

PR #5431, opened 2022-11-16
remove swagger-ui-react and use the dist version

gatzjames:
Motivation:
The way this package is compiled has proven problematic for our bundling alongside with peer deps limiting us from upgrading.
By including the component in our codebase we only rely on swagger-ui for rendering specs.


gatzjames:
Closed in favor of #5417

PR #5435, opened 2022-11-17
Layout UI updates

gatzjames:
Introduces some UI updates:
- Increases the status bar size since it appears quite small.
- The header content is now horizontally centered.

image

- Updates the copy for the new org modal for logged out users

image


jackkav (approved):

filfreire (approved):
🚀

PR #5455, opened 2022-11-23
Remove insomnia-importers, move it to insomnia

filfreire:
Removes insomnia-importers, moving it into insomnia package

Pros: speeds up bootstrap and one less package to maintain

Cons: we are deleting an old "insomnia-importers" CLI that folks could use 
before when we published to npmjs as a package.


gatzjames (approved):
🥇

jackkav (approved):

marckong (approved):

PR #5473, opened 2022-11-30
[INS-2174] fixes create new plugin issue

filfreire:
changelog(Fixes): Fixed an error that showed up when attempting to create a new plugin from the plugin preferences menu
Closes INS-2174

Solves a case where folks will get an error when trying to create a new plugin:
Screenshot 2022-11-30 at 17 23 57


gatzjames (approved):
🎉

PR #5480, opened 2022-12-02
Add prerelease grpc smoke tests

filfreire:
Closes INS-2175

Adapts [route guide example](https://github.com/grpc/grpc/blob/master/examples/node/dynamic_codegen/route_guide/route_guide_server.js) and we get a separate grpc server for prerelease tests, so we can check all connection types.


jackkav (approved):
I'm going to miss that old grpc server, thanks for this @filfreire !

PR #5541, opened 2022-12-16
Fix inso cli lint fail on relative path INS-2219

filfreire:
Closes INS-2219

changelog(Inso CLI): Fixed an issue where `lint spec` didn't work for relative file paths


gatzjames (approved):
🚀

PR #5603, opened 2023-01-03
Preferences prelease test

kreosus:
Closes INS-2085


packages/insomnia-smoke-test/tests/prerelease/preferences-interactions.test.ts
```diff
@@ -0,0 +1,20 @@
+import { expect } from '@playwright/test';
+
+import { test } from '../../playwright/test';
+
+test('Preferences through click', async ({ page }) => {
+
+  await page.locator('[data-testid="settings-button"]').click();
+  await page.locator('text=Insomnia Preferences – v2022.7.0-alpha.0GeneralDataThemesKeyboardAccountPluginsU >> button').first().click();
```
gatzjames: This selector will break whenever we change versions (v2022.7.0-alpha.0) so we should change it to be more generic.

packages/insomnia-smoke-test/tests/prerelease/preferences-interactions.test.ts
```diff
@@ -5,16 +5,15 @@ import { test } from '../../playwright/test';
 test('Preferences through click', async ({ page }) => {
 
   await page.locator('[data-testid="settings-button"]').click();
-  await page.locator('text=Insomnia Preferences – v2022.7.0-alpha.0GeneralDataThemesKeyboardAccountPluginsU >> button').first().click();
+  await page.locator('text=Insomnia Preferences ').first().click();
```
wongstein: nit pick: the text has a trailing space "Insomnia Preferences " but in line 17 it doesn't have a trailing space.  It looks to me like they are the same element, and while I don't think a trailing space matters a lot in this instance, I could see edge cases where the difference between line 8 and line 17 causes unnecessary test failures.

My suggestion is to pick 1 (with a trailing space, without a trialing space) and make it the same between lines 8 and 17.

filfreire (approved):

PR #5666, opened 2023-01-13
Fix INS-2232 default theme darker

filfreire:
changelog(Improvements): Edited the default theme so previously bright modals are now darker. The previous default theme is still available as the Legacy theme 

Closes INS-2232

- Makes default theme darker, while preserving `default` to not break theming after people upgrade
- Created `legacy` theme in case folks want to go back to previous one

Before:
Screenshot 2023-01-13 at 12 47 09

After:

Screenshot 2023-01-13 at 12 51 05

Pros: 
- removes CS 1.6 flashbang effect when opening a modal like cookie editor or preferences. Also sparks joy

### DIFF
--- a/plugins/insomnia-plugin-core-themes/index.js
+++ b/plugins/insomnia-plugin-core-themes/index.js
@@ -1,5 +1,6 @@
 module.exports.themes = [
   require('./themes/default'),
+  require('./themes/legacy'),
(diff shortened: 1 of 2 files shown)


gatzjames (approved):
🚀

jackkav (approved):
💄 📸

PR #5681, opened 2023-01-16
Reproduce GraphQL unknown operation named issue

### LINKED ISSUE #5665: GraphQL Operations Menu is not disabled and causes "operationName" to be empty
### Expected Behavior

If there's no options, the Operations button should be disabled.

### Actual Behavior

There´s an empty option in the menu, that if selected, it adds `"operationName":""` to the query, which leads to the error `"Unknown operation named \"\".",`, and once selected, there's no way to remove the selection.
image
image

### Reproduction Steps

- Update to version 2022.7.2
- Create a GraphQL request
- Select the empty option in the Operations Menu
(note: some querys may have the option pre-selected, i haven't figured why)

### Additional Information

i have updated to the `2022.7.2` version and noticed that some mutations where returning an error about an `Unknown operation`, so i noticed that the Operations menu had an empty option that i can not remove the selection. i asked some members of my team if they could send an screenshot of their version and the Operations button disabled. the latest version where this issue doesn't happen seems to be version `2022.7.0`, here are the screenshots:
image
image

### Insomnia Version

2022.7.2

Windows

### Operating System Version

Windows 11 Pro 22H2

### Installation method

from insomnia.rest

2022.7.0

filfreire:
changelog(Fixes): Fixed an issue that caused GraphQL requests to fail in some cases with an unknown operation-like error

Basic reproduction of #5665 
Closes #5665

Reproduction for INS-2268

image

### DIFF
--- a/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
+++ b/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
@@ -36,6 +36,34 @@ import { Toolbar } from '../../key-value-editor/key-value-editor';
 import { useDocBodyKeyboardShortcuts } from '../../keydown-binder';
 import { TimeFromNow } from '../../time-from-now';
 
+function getGraphQLContent(body: GraphQLBody, query?: string, operationName?: string, variables?: string): string {
+  // the body object is one dimensional, so we don't need to worry about shallow copying.
+  const { query: originalQuery, ...optionalProps } = body;
+  const content: GraphQLBody = { query: originalQuery };
+
+  if (optionalProps.operationName) {
+    content.operationName = optionalProps.operationName;
+  }
+
+  if (optionalProps.variables) {
+    content.variables = optionalProps.variables;
+  }
+
+  if (query) {
+    content.query = query;
+  }
+
+  if (operationName) {
+    content.operationName = operationName;
+  }
+
+  if (variables) {
+    content.variables = variables;
+  }
+
+  return JSON.stringify(content);
+}
+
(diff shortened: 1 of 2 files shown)


marckong:
I pushed a quick hack to see if my assumption for the root cause is valid. It seems that the root cause is passing an empty string to the operationName, which is expected to throw an error according to the graphql-js library - https://github.com/graphql/graphql-js/blob/522f4950cea3bff53c919e0b3bca295c5696a618/src/execution/__tests__/executor-test.ts#L814

I admit that the committed code is just a hack to fix this, and we should probably make an actual fix for this. Will shortly make a better fix for this.

packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -36,6 +36,34 @@ import { Toolbar } from '../../key-value-editor/key-value-editor';
 import { useDocBodyKeyboardShortcuts } from '../../keydown-binder';
 import { TimeFromNow } from '../../time-from-now';
 
+function getGraphQLContent(body: GraphQLBody, query?: string, operationName?: string, variables?: string): string {
```
marckong, line 39: Please feel free to improve this.

filfreire (commented):
I think @marckong changes do the trick, LGTM 👍

marckong (approved):

gatzjames (approved):
getContent looks good 🎉

bosko:
I've just checked and problem still exists in 2022.7.4

Version: Insomnia 2022.7.4
Build date: 1/17/2023
OS: Darwin arm64 22.2.0
Electron: 22.0.0
Node: 16.17.1
V8: 10.8.168.20-electron.0
Architecture: arm64

PR #5723, opened 2023-01-25
Copy as curl

kreosus:
Closes INS-1757

changelog(Improvements): Added a small prompt when using `Copy as Curl` feature on a request


filfreire (approved):

Shawyeok:
Now `Copy as cURL` is way more complex before. I'm curiosity why should we add a confirm prompt here, is there any risk of accidental operation?

1. left click to select request
2. move cursor to right then left click to open dropdown (not responding right click anymore)
3. click `Copy as cURL`
4. click `Click to confirm`

PR #5729, opened 2023-01-26
fix windows pdf request smoke test flake

filfreire:
The pdf preview load causes the click on the dropdown `Raw Data` option to be missed due to focus loss.

Dropdown opening
Screenshot 2023-01-26 at 17 31 01

Dropdown closes when preview opens
Screenshot 2023-01-26 at 17 31 09


pavkout (approved):
LGTM 👍

PR #5778, opened 2023-02-14
Fix dropdown closing early issue

filfreire:
changelog(Fixes): Fixed an issue where some dropdown options no longer worked (e.g. deleting a folder, deleting request history, ...).

When we have an `ItemContent`/`DropdownItem` inside a `DropdownSection` and the item has both an `onClick` call and also a `withPrompt` property, we need to set the `closeOnSelect={false}` value on the parent Dropdown so it doesn't exit early when a prompt is opened.

Closes INS-2359


gatzjames (approved):
🚀

PR #5806, opened 2023-02-27
Add ability to delete default project

gatzjames:
Highlights:
- [x] Allows users to delete default project
- [x] A default project is created only the first time the user launches the app.

Future work:
- [ ] Empty View for organization/team without projects
- [ ] Add an Organization route above the project once we update the SidebarLayout component


gatzjames:
Closing since we don't have a default project anymore

PR #5852, opened 2023-03-17
Added the ability to escape percent character

### LINKED ISSUE #4818: Query can't handle number after %
### Expected Behavior

The URI encoding of % isn't handle correct when a number is entered.

```
http://localhost?test=%30%
```

should pass the query `test` with value `%30%`

or

```
http://localhost?test=%2530%25
```

should pass the query `test` with value `%2530%25`

### Actual Behavior

Screen Shot 2022-05-26 at 6 14 12 PM

Screen Shot 2022-05-26 at 6 28 05 PM

Screen Shot 2022-05-26 at 6 28 33 PM

### Reproduction Steps

1. Create a new GET request
2. Go to Query Tab
3. Enter a `test` in New name field
4. Enter `%30%` in New value field
5. Url in preview turns into `0%25`

### Additional Information

`curl http://localhost?test=%30%` and `curl http://localhost?test=%2530%25` both send the correct 

I found this post stating the same issue I'm experiencing:

https://github.com/Kong/insomnia/issues/605#issuecomment-825546206

### Insomnia Version

v2022.3.0

macOS

### Operating System Version

macOS Monterey Version 12.3.1

### Installation method

insomina.rest

_No response_

Bert210:
changelog(Fixes): Fixed an issue where url encoding couldn't handle % with two digits behind it without perceiving it as an escape code.

This PR adds the ability to escape percent character (%) with the backslash (\\).

e.g. \\% is converted to %25

I'm also open to any other ideas on how to handle this issue.

This closes #4818 
Screen Shot 2023-03-17 at 4 55 02 PM

### DIFF
--- a/packages/insomnia/src/utils/url/querystring.test.ts
+++ b/packages/insomnia/src/utils/url/querystring.test.ts
@@ -120,6 +120,15 @@ describe('querystring', () => {
 
       expect(str).toBe('foo=bar%3F%3F&hello=&hi%20there=bar%3F%3F&=bar%3F%3F&=');
     });
+
+    it('builds from params handle escaped %', () => {
+      const str = buildQueryStringFromParams([
+        { name: 'foo', value: '\\%30\\%' },
+        { name: 'bar', value: '%30\\%' },
+      ]);
+
+      expect(str).toBe('foo=%2530%25&bar=%30%25');
+    });
   });
 
   describe('deconstructToParams()', () => {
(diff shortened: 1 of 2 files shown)


myplacedk:
It seems to me that it solves the wrong problem.

We should need to encode anything. %30 is not an escape code. I would expect Insomnia to encode the % into %25.

If I want the query parameter in the url to say %2530%25, I would type in %30%.

If I want the query to actually say test=%30%, I wouldn't expect to be able to do that in that interface. I would expect to have to type in the URL manually as test=%30%, and even then I would expect that to turn into test=%30%25, which is actually just 0%.

gatzjames:
Closing as stale

PR #5958, opened 2023-05-12
INS-2627: sanitize path-variable name regex capture for kong 3.x

filfreire:
Closes INS-2627
Closes FTI-5060

changelog(Fixes): Added a fix to handle an edge of how path variable names are captured from regex when parsing OAS specs for Kong 3.x.

Minor change, to mimic what was done in `kced` - see https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L50

### DIFF
--- a/packages/openapi-2-kong/src/common.ts
+++ b/packages/openapi-2-kong/src/common.ts
@@ -77,6 +77,15 @@ export function pathVariablesToRegex(p: string, legacy: Boolean = true) {
     result = '~' + result;
   }
 
+  // Remove illegal chars from path-variable name.
+  // see https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L50
+  if (!legacy) {
+    result = result.replace(/-/g, '_');
+    if (result.startsWith('_')) {
+      result = 'a' + result;
+    }
+  }
+
   return result + '$';
 
 }
--- a/packages/openapi-2-kong/src/common.ts
+++ b/packages/openapi-2-kong/src/common.ts
@@ -77,17 +77,22 @@ export function pathVariablesToRegex(p: string, legacy: Boolean = true) {
     result = '~' + result;
   }
 
-  // Remove illegal chars from path-variable name.
-  // see https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L50
   if (!legacy) {
-    result = result.replace(/-/g, '_');
-    if (result.startsWith('_')) {
-      result = 'a' + result;
-    }
+    result = sanitizeRegexCapture(result);
   }
-
   return result + '$';
 
+}
+
+// Remove illegal chars from path-variable name.
+// see https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L50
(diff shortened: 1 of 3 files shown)


packages/openapi-2-kong/src/common.ts
```diff
@@ -77,6 +77,15 @@ export function pathVariablesToRegex(p: string, legacy: Boolean = true) {
...
   }
 
+  // Remove illegal chars from path-variable name.
+  // see https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L50
+  if (!legacy) {
+    result = result.replace(/-/g, '_');
+    if (result.startsWith('_')) {
+      result = 'a' + result;
+    }
+  }
```
Tieske, line 82: Can we externalize this in to its own function? that function can then be called from here, as well as from the validation-generator, to ensure both will have the same name, and the user doesn't have to manually update his/her OAS spec

here's the Go-function; https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L47-L57

called when generating the regex: https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L819

called when generating the validation: https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/validator.go#L60-L64

packages/openapi-2-kong/src/common.ts
```diff
@@ -77,6 +77,15 @@ export function pathVariablesToRegex(p: string, legacy: Boolean = true) {
     result = '~' + result;
   }
 
+  // Remove illegal chars from path-variable name.
+  // see https://github.com/Kong/go-apiops/blob/8842f7b0ecf3654f62f8e0bd867b5b67766794c9/convertoas3/oas3.go#L50
+  if (!legacy) {
+    result = result.replace(/-/g, '_');
```
Tieske: To do this properly, use some slugification (to fix non-ascii chars), and then reduce to [A-Za-z0-9_] whilst starting with [A-Za-z].

Tieske (changes requested):

packages/openapi-2-kong/src/common.ts
Tieske: from: https://www.pcre.org/current/doc/html/pcre2pattern.html#SEC16

> When PCRE2_UTF is not set, they may contain only ASCII alphanumeric characters and underscores, but must start with a non-digit.

I don't know if PCRE2_UTF is set in our implementation, but better safe than sorry I guess.

packages/openapi-2-kong/src/common.ts
filfreire: Done, moved it into it's own function, and has the same name 👍

packages/openapi-2-kong/src/common.ts
filfreire: Added slugification is done ✅ 

The PCRE2_UTF part not sure how to do it though, but should be fine as is.

jackkav (approved):

Tieske (approved):

packages/openapi-2-kong/src/common.ts
Tieske: what I meant mostly was removing accents etc from characters; eg. reduce `èëé` into `eee`

PR #5974, opened 2023-05-17
update test to expose bug in regex generation

Tieske:
sanitizing the path-variable name is not done correctly, as it sanitizes the entire path. So it also makes unintended changes.

changelog(Fixes): Fixed an issue when sanitizing the path-variable names in Inso CLI which was not done correctly, as it sanitized the entire path. So it also makes unintended changes. Now only matched path-variables are sanitized.


filfreire (approved):

Tieske (commented):
lgtm, just can't approve, since I'm the author of the original PR 🤷‍♂️

PR #5977, opened 2023-05-18
INS-2639: add kong 3.x support for inso generate k8s config

filfreire:
changelog(Improvements): Added minor Kong 3.x support for generating k8s configs using Inso CLI

While ApiOps v2 is not in full GA/production yet, adding this small QoL improvement to config generation to unblock folks still using Inso CLI for kong on k8s config generation cc @mheap @rspurgeon

### DIFF
--- a/packages/openapi-2-kong/src/kubernetes/generate.test.ts
+++ b/packages/openapi-2-kong/src/kubernetes/generate.test.ts
@@ -304,6 +304,11 @@ describe('index', () => {
       expect(generateServicePath(serverBasePath)).toBe('/api/.*/.*');
     });
 
+    it('adds /~ for kong 3.+ non legacy configs', () => {
+      const serverBasePath = '/api/.*';
+      expect(generateServicePath(serverBasePath, '', false)).toBe('/~/api/.*/.*');
+    });
+
     it.each(['/', '/specificPath'])(
       'does not add closing wildcard if using specific path: [%o]',
       specificPath => {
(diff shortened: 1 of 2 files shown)


packages/openapi-2-kong/src/kubernetes/generate.test.ts
```diff
@@ -304,6 +304,11 @@ describe('index', () => {
       expect(generateServicePath(serverBasePath)).toBe('/api/.*/.*');
     });
 
+    it('adds /~ for kong 3.+ non legacy configs', () => {
+      const serverBasePath = '/api/.*';
+      expect(generateServicePath(serverBasePath, '', false)).toBe('/~/api/.*/.*');
```
mheap:
```suggestion
      expect(generateServicePath(serverBasePath, '', false)).toBe('/~/api/.*');
```

filfreire: @mheap should we also edit then the `const serverBasePath = '/api/.*';` above to be just `/api` or the wildcard should just not be added if it's already on the path?

mheap: Ah, `.*` isn't supported in the base path

mheap (approved):

jackkav (approved):
Still don't agree with legacy defaulting to true, but since this feature is moving to a Go implementation, I'm happy to approve one liners.

PR #6078, opened 2023-07-02
Add feature: synchronization to workspace

NickRimmer:
These changes will allow plugins to synchronize a workspace as before - without generating new IDs for everything, and include environments, cookies, API spec, etc.

Nothing existing was touched, so it should be safe and not break anything. I was trying to be careful and save consistency in the code, but if you see anything to refactor - I'm open to updating it. I added tests with a few negative as well, I hope covered most of the cases.

Idea discussion can be found here #6077


NickRimmer:
As no one from Insomnia team reacted so far, I'm closing my PR

Issue #6035, opened 2023-06-15
Unable to duplicate request leaving name empty


rstain:
### Expected Behavior

When duplicating request with empty name it should allow to left the name empty.

### Actual Behavior

When duplicating request with empty name it is required to provide a name. Otherwise a new request is not created.

### Reproduction Steps

1. Navigate to a request that have no name.
2. Press Ctrl+D.
3. In the Duplicate Request dialog left the New Name field empty.
4. Click Create.
Result: no new request is created.

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Insomnia Version

2022.7.3

### What operating system are you using?

Windows

### Operating System Version

Windows 10 Pro, 22H2

### Installation method

download from insomnia.rest

PR #6138, opened 2023-07-11
Fix: Allow empty name when duplicate request

pilotpirxie:
changelog(Improvements): Empty strings are now allowed when duplicating a request.

When duplicating request, an empty prompt with a name can be accepted but it's not processed. Insomnia handles empty names and displays the entire URL. Not handling empty names looks there like a bad user experience.

This PR allows proceeding with an empty string when duplicating a request.
 
Screenshot 2023-07-11 at 23 08 19

### DIFF
--- a/packages/insomnia/src/ui/components/modals/prompt-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/prompt-modal.tsx
@@ -69,7 +69,7 @@ export const PromptModal = forwardRef<PromptModalHandle, ModalProps>((_, ref) =>
     event.preventDefault();
     if (inputRef.current) {
       const result = inputRef.current.type === 'checkbox' ? inputRef.current.checked.toString() : inputRef.current.value;
-      if (result) {
+      if (result || inputRef.current?.type === 'text') {
         state.onComplete?.(state.upperCase ? result?.toUpperCase() : result);
       }
       modalRef.current?.hide();


jackkav (approved):
I agree, however its debatable that we should allow empty request names as it could cause some confusion, but then we would need validation. Thanks.

filfreire:
@pilotpirxie @jackkav this will cause a few unhandled cases in terms of user experience, for example Rename which also uses prompt modal:

image

Right now we won't display anything:
image

jackkav:
closes #6035

jackkav:
We could add `required` to the prompt modal input as an alternative to allowinf empty, but I don't think its so bad to be empty naming things is sometimes too much trouble, perhaps an autofilled name and required field would be better than allowing empty. This PR is still okay though since empty names shouldn't cause any damage elsewhere, just confusion for our future selves.

pilotpirxie:
@filfreire 
When renaming seems also fine:

https://github.com/Kong/insomnia/assets/10637666/b55e032e-5324-47ea-a338-e286357e7922

pilotpirxie:
@jackkav so, can it be merged? Checks have passed but I don't have permission to merge

filfreire:
@pilotpirxie you can claim a free tshirt for this PR contribution - find more at https://konghq.com/community/open-source-contribution

Issue #5793, opened 2023-02-22
REST HTTP Header "Content-Disposition" used entirely as filename


veso0308:
### Expected Behavior

When downloading a file from a source which has following header:
`Content-Disposition: attachment; filename=bambi.csv`
The filename should be `bambi.csv`

### Actual Behavior

Actually when downloading a file, the name is not `bambi.csv`, but `attachment; filename=bambi.csv`.

### Reproduction Steps

Do to the desired resource which provides a resource with an header like:
`Content-Disposition: attachment; filename=bambi.csv`
Click on the arrow next to the "Send" button and click on "Download After Send", then choose a destination folder.
After this click on the now changed button "Download".
You will see, the output on the right: `Saved to _<DESTINATIONFOLDER>_\attachment; filename=bambi.csv`

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Insomnia Version

2022.7.5

### What operating system are you using?

Windows

### Operating System Version

Windows 10 Enterprise 21H2

### Installation method

download from insomnia.rest

v4rn:
Can I pick this issue?

veso0308:
> Can I pick this issue?

Sure you can!

schopr9:
looks like this is [fixed](https://github.com/Kong/insomnia/pull/6159) and can be closed. @veso0308

danrodagu:
It doesn't seem working on the last version of the Windows client at least. The filename remains empty to be manually set, or uses the last filename used

PR #6159, opened 2023-07-16
fix(download): properly parse content-disposition filename

### LINKED ISSUE #4252: Filename is incorrectly determined from Content-Disposition header
### Expected Behavior

When saving a file, I expect the filename to be named according to the Content-Disposition header.

GIven a Content-Disposition header of 'attachment; filename="foo.json"', I expect the saved file to be named "foo.json"

### Actual Behavior

The file is named according to the entire contents of "Content-Disposition" and not just the filename.

In the above example, the file is named 'attachment; filename="foo.json"'

### Reproduction Steps

1. Create a get request for an endpoint that will return a Content-Disposition header
2. Right click on "Request" and select "Start auto-download"
3. Choose a directory location
4. Click download

### Additional Information

The bug occurs on [this line](https://github.com/Kong/insomnia/blob/177d6adf38852aae2aef72717fa28dc4d9fef35e/packages/insomnia-app/app/ui/containers/app.tsx#L770) where the entire contents of the header is used, rather than being parsed to see if it contains a filename.

Please note the following recommendations [from MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) on handling of the filename directive of this header:

> Is followed by a string containing the original name of the file transmitted. The filename is always optional and must not be used blindly by the application: path information should be stripped, and conversion to the server file system rules should be done. This parameter provides mostly indicative information. When used in combination with Content-Disposition: attachment, it is used as the default filename for an eventual "Save As" dialog presented to the user.

### Insomnia Version

2021.6.0

Ubuntu

### Operating System Version

Ubuntu 20.04.2

### Installation method

download from insomnia.rest

_No response_

micc83:
Hi, first of all thank you for Insomnia, I love it's simplicity of use.

This PR should fix the issue with content-disposition not being parsed on file download so that the entire content-disposition was used to name the downloaded file.
### Steps to reproduce:
* GET `https://images.pexels.com/photos/17490386/pexels-photo-17490386.jpeg`
* Select "Download after send" or "Send and download" from the "Send" dropdown

Before this PR the file would have been downloaded with the following name:
```
path/attachment;filename="pexels-anastasiia-chaikovska-17490386.jpg"
```
After:
```
path/pexels-anastasiia-chaikovska-17490386.jpg
```
This is my first PR to the repo (actually haven't been doing much open source lately, but feels good to be back) so I'm not sure I did everything right, also not sure where tests lies, if you point me out I'd be glad to add some.

One last thing. I tried to keep my changes as small as possible but as I see that `getContentDispositionHeader` is used only here we might rename it to `getFilenameFromContentDisposition` and move all the logic there (we could also wrap all the implementation in a try catch returning null on error so to be sure that the impact of malformed content-disposition or any issue with the external library would be minimal).

Closes #4252 #5793


alan-m-thomas:
haustorium Keibler lekshmipura

jackkav (approved):

filfreire:
@micc83 you can claim a free tshirt for this PR contribution - find more at https://konghq.com/community/open-source-contribution

micc83:
Lovely, thank you @filfreire!

danrodagu:
It doesn't seem working on the last version of the Windows client at least. Just keep the filename as empty to be manually set, or uses the last filename used

PR #6160, opened 2023-07-16
Add feature: Plugins update

NickRimmer:
Hi guys,

I'm thrilled to suggest a new feature for Insomnia! This feature will make updating installed plugins easier and more intuitive for all users. Previously, the update process required a bit of maneuvering, as users had to reinstall their plugins. I've essentially added a touch of UI enhancement to streamline this process. I'm eager to hear everyone's feedback on this!

### DIFF
--- a/packages/insomnia/src/ui/components/settings/plugins.tsx
+++ b/packages/insomnia/src/ui/components/settings/plugins.tsx
@@ -106,6 +106,9 @@ export const Plugins: FC = () => {
     }
 
     // update plugins
+    updates.message = 'Downloading updates...';
+    setState(state => ({ ...state, updates }));
+
     for (const update of updates.plugins.filter(p => p.status === 'pending')) {
       update.status = 'in-progress';
       setState(state => ({ ...state, updates }));
@@ -123,6 +126,7 @@ export const Plugins: FC = () => {
     }
 
     await refreshPlugins();
+    updates.message = 'Updates complete';
     updates.completed = true;
     setState(state => ({ ...state, updates }));
   }
--- a/packages/insomnia/src/ui/components/settings/plugins.tsx
+++ b/packages/insomnia/src/ui/components/settings/plugins.tsx
@@ -84,7 +84,7 @@ export const Plugins: FC = () => {
 
     // build the list of plugins to udpate
     plugins = plugins.filter(plugin => plugin.directory);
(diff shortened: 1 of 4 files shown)


packages/insomnia/src/ui/components/settings/plugins.tsx
```diff
@@ -160,6 +232,33 @@ export const Plugins: FC = () => {
...
+            </button>
+          )}
+
+          <div>
+            <p style={{ marginBottom: 5 }}>{updates.message}</p>
+            <ul>
+              {updates.plugins.map(update => (
+                <li key={update.name}>
+                  <span style={{ marginRight: 5 }}>
+                    {update.status === 'in-progress' && (<i className="fa fa-refresh fa-spin" title="Updating in progres" />)}
```
bbbco: `progres` => `progress`

packages/insomnia/src/ui/components/settings/plugins.tsx
```diff
@@ -160,6 +232,33 @@ export const Plugins: FC = () => {
...
+                    {update.status === 'error' && (<i className="fa fa-exclamation-triangle" style={{ color: 'var(--color-warning)' }} title={update.statusMessage ?? 'Plugin cannot be udpated'} />)}
```
bbbco: `Plugin cannot be udpated` => `Plugin cannot be updated`

packages/insomnia/src/ui/components/settings/plugins.tsx
NickRimmer: Oh, thank you, updated (;

packages/insomnia/src/ui/components/settings/plugins.tsx
NickRimmer: Thank you \~_\~ updated

HeatherLemieux:
I think this is a great feature, thank you so much for your work on this! However, `Check Updates` implies that it will only `Check` whether there are updates or not and present that information to me. I would actually prefer that approach as well. I don't want all my plugins to be updated without me being able to first review. Ideally, I would be able to click `Check Updates`, see what plugins have updates available, and then be able to individually pick and choose which plugins will get updated.

NickRimmer:
As no one from Insomnia team reacted so far, I'm closing my PR

PR #6193, opened 2023-07-25
chore: use npm workspaces (herecles)

filfreire:
## TODO

- [x] replace version script to get version (equivalent to old `lerna version`)
- [x] inso package uses the wrong node version, either use electron or rebuild with node18(v108) from remote
- [x] fix import paths of workspaces?
- [x] lint
- [x] build
- [x] test (fails)
- [x] type-check (currently fails)
- [x] package (packaged app fails to fully boot)
- [x] inso tests
- [x] inso package
- [x] smoke tests (app fails to fully boot)
- [x] app-start

future work
- [x] remove iteration and alpha options in github actions
- [x] document new inso dx `node_modules/.bin/node-pre-gyp install --update-binary --directory node_modules/@getinsomnia/node-libcurl`
- [ ] remove duplicate deps in workspaces
- [ ] less tsconfigs please
- [ ] prettier > eslint-typescript
- [ ] vitest > jest

Closes INS-2867

## issues

- release process with npm version rather than leran version encourages a more conventional approach where its not required to set the iteration version, rather it is calculated
- on CI we get `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` and have to force the `--max_old_space_size` flag
- `build` doesn't seem to run as fast locally - also in subsequent runs it rebuilds everything from scratch
- running dev without a build step makes bundling inso with a build step more painful, perhaps its better to remove the build step from inso or simplify it, since it currently forces us to have tsconfigs lying around everywhere
- would be nice if we could replace ts-jest with vitest in order to get support for ts 5, more speed and better upstream support


jackkav (approved):
🎸

PR #6223, opened 2023-08-04
chore: force window size on smoke tests

filfreire:


filfreire:
closing for now, we'll need to recheck this later

PR #6448, opened 2023-09-05
chore: increase electron-builder compression to maximum

filfreire:


filfreire:
@Kong/team-insomnia closing this one, not worth taking into consideration for the time being.

 There's negligible save in size when using the maximum compression option and the packaging for example for release-recurring increases from 20 mins to 30 mins.

PR #6471, opened 2023-09-08
Update settings.ts to default not send user data

andreseloysv:
Let the users choose if they want to help improve the software or not


filfreire:
@andreseloysv thank you for suggesting this, we won't be taking this change at the moment

PR #6539, opened 2023-09-20
feat: move to jq

### LINKED ISSUE #777: [Feature Request] Use jq for JSON filters
JSON filters on Insomnia are pretty useful, but there's some cases that aren't (nicely) covered by it. I have to regularly interact with APIs that return arrays of big JSON objects (40+ properties) of which I usually need a small subset of fields. JSON path allows you to do `$.results[*]["asd","id"]` but that prints an array with no keys which is confusing when the types of the fields are similar. 

[jq](https://stedolan.github.io/jq/) is really powerfull JSON processor which can transform just about any JSON input to any output you can imagine (while being really really fast). The query `.results[] | {asd, id}` would produce an array of objects with the proper keys, instead of simply the values.

### LINKED ISSUE #2342: JSONpath evaluator not working as expected
**Issue**
The JSON path evaluator (https://github.com/dchester/jsonpath) which is being used is not working in the app as expected when multiple keys are given in an expression.

**To Reproduce**
Steps to reproduce the behavior as per the JSON object example given at https://github.com/dchester/jsonpath
1. Try the expression: $.store.book[*].[category,price]
2. No results in Insomnia app

**Expected behaviour**
With the same expression when tried at  https://jsonpath.com/ gave the below result:
[
  "reference",
  8.95,
  "fiction",
  12.99,
  "fiction",
  8.99,
  "fiction",
  22.99
]

**Using OSX Desktop app:**
 - macOS 10.14.6 
 - Installation Method: Directly from the web
 - App Version: 2020.2.2

### LINKED ISSUE #3155: JSONPath array filter can't refer to root object ($)
The following JSONPath expression:
```$.included[?(@.id == $.data.relationships.target.id)]```
would be expected work with the following JSON, but does not: http://api.mocki.io/v1/9c54d246

Individually, 
```$.included[?(@.id == "foo")]```
and
```$.data.relationships.target.id```
work, but not when combined.

 - OS: Big Sur
 - App Version: v2021.1.0

### LINKED ISSUE #3258: Filtering all the keys not working
**Describe the bug**
I'm trying to filter out all the keys from the JSON response of WordPress API. When going to https://jsonpath.com/ and trying 

```bash
$.routes.*~
```

I get the keys of all routes. When I type this in the Insomnia I get an empty array.

**To Reproduce**
JSON example

```json
{
  "name": "WordPress",
  "description": "Just another WordPress site",
  "url": "https:\/\/mysite.test",
  "home": "https:\/\/mysite.test",
  "gmt_offset": "0",
  "timezone_string": "",
  "namespaces": [
    "oembed\/1.0",
    "menus\/v1",
    "acf\/v3",
    "wc-bookings\/v1",
    "wc\/store",
    "wc-analytics",
    "wc-admin",
    "wc\/v1",
    "wc\/v2",
    "wc\/v3",
    "wccom-site\/v1",
    "wp\/v2",
    "wp-site-health\/v1"
  ],
  "authentication": {
    "application-passwords": {
      "endpoints": {
        "authorization": "https:\/\/mysite.test\/wp-admin\/authorize-application.php"
      }
    }
  },
  "routes": {
    "\/": {...},
    "/batch/v1": {...},
    "/oembed/1.0": {...},
    "/oembed/1.0/embed": {...},
    "/oembed/1.0/proxy": {...}
  }
}
```

**Expected behavior**
I expect to see something like

```json
[
  "/",
  "/batch/v1",
  "/oembed/1.0",
  "/oembed/1.0/embed",
  "/oembed/1.0/proxy"
]
```

**Screenshots**

 - OS: macOS 11.2.3 Big Sur
 - Installation Method: https://insomnia.rest/download
- Version: Insomnia 2021.1.1
- Release date: 10/03/2021
- OS: Darwin x64 20.3.0
- Electron: 9.1.1
- Node: 12.14.1
- V8: 8.3.110.13-electron.0
- Architecture: x64
- node-libcurl: libcurl/7.69.1 OpenSSL/1.1.1g zlib/1.2.11 brotli/1.0.7 libidn2/2.1.1 libssh2/1.9.0 nghttp2/1.41.0

kobenguyent:
Motivation: jq, while being more powerful than JSONPath, nevertheless enables one to write queries that are usually as simple as, and sometimes even simpler than, the corresponding JSONPath queries.

resolves #777 
resolves #3258
resolves #2342
resolves #3155

https://github.com/Kong/insomnia/assets/7845001/5e97f72b-a2ec-4d30-9497-6394d691c8c4


bbbco:
Considering `jq` uses a downloaded executable, you might want to consider something more integrated to Javascript like [JSONata](https://jsonata.org/)

kobenguyent:
Thanks. I'll check it out. 😄

kobenguyent:
I think we would still use node-jq instead of using another as jq is quite popular and its simplicity as well as performance. 

For this PR, I guess we should have backwards compatibility with users are using JSONPath in their collections. As we drop the JSONPath, it would ruin the chain requests where the JSONPath is used to query.

kobenguyent:
Well, I think I'll try with this approach then. Copying the node-jq lib along with the binaries and minimizing the lib source code inside insomnia. That would somehow be a better way. Love to hear the feedbacks.

gatzjames:
@kobenguyent thanks for the pr 👍. Adding extra binaries makes the build process highly dependent on them (currently we use node-libcurl) and makes testing and deploying to different platforms more difficult. So we avoid adding them unless the benefits are worth it. It will be helpful if you explain the pros/cons of jq and what are your motivations/features this will support

kobenguyent:
thanks @gatzjames for looking into this. So yeah, maybe this explains better than myself 
https://github.com/stedolan/jq/wiki/For-JSONPath-users

IMHO, jq resolves many issues as we are currently having with JSONPath also its simplicity and performance are two pros worth mentioning 😄

smartrejames:
Ping, I'm really excited about this feature and can’t wait to see it in action!

gatzjames:
Closing as including jq in creates many issues with bundling

PR #6551, opened 2023-09-22
Sync Organizations

gatzjames:
Currently we refetch organizations on navigation.

This caches the results in memory and only updates the organizations when:
- we navigate to `/organization` (currently redirected there after logging in) **OR**
- when there is an `OrganizationChanged` event in the presence API

Closes INS-3024, INS-3094, INS-3023

### DIFF
--- a/packages/insomnia/src/ui/routes/organization.tsx
+++ b/packages/insomnia/src/ui/routes/organization.tsx
@@ -34,6 +34,7 @@ import FileSystemDriver from '../../sync/store/drivers/file-system-driver';
 import { MergeConflict } from '../../sync/types';
 import { shouldRunMigration } from '../../sync/vcs/migrate-to-cloud-projects';
 import { getVCS, initVCS } from '../../sync/vcs/vcs';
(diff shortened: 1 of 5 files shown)


filfreire (approved):

jackkav (commented):
This is a good chance to clean up the sort code

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -94,12 +144,51 @@ export const indexLoader: LoaderFunction = async () => {
...
+            showModal(SyncMergeModal, {
+              conflicts,
+              handleDone: (conflicts?: MergeConflict[]) =>
+                resolve(conflicts || []),
+            });
+          });
         });
+      }
+
+      const organizationsResult = await window.main.insomniaFetch<OrganizationsResponse | void>({
```
marckong, line 142: Shouldn't this be Promise.all() or .allSettled()? This is forming waterfall request and the page won't be returned until they are resolved in waterfall manner.

jackkav (approved):

PR #6518, opened 2023-09-17
feat(shortcut): shortcut close env form

kobenguyent:
Motivation: have a shortcut to close the environment modal as I don't need to move my mouse and click the button.

I updated the UX a bit, when on Edit mode, 
- press the Esc to exit the edit mode then press the shortcut to save/close the modal, this behavior will be the same like user clicks on the Close button.
- press the Esc to exit the edit mode then press Enter to save/close the modal, this behavior will be the same like user clicks on the Close button.

Resolves #6511
Resolves #6370 

https://github.com/Kong/insomnia/assets/7845001/388c2336-c4f3-4011-b89d-d3d31539884f

PR #6573, opened 2023-09-27
fix #4877: add customizable keybind for closing modals

### LINKED ISSUE #4877: Escape key closes environment editor modal in vim mode
### Expected Behavior

When editing an environment, I'd like to use `Esc` to exit insert mode, without closing the modal. Ideally, when pressing the escape key:

* In insert mode: return to normal mode
* If anything is selected: deselect what is selected
* In normal mode: close the environment editor

In practice that might be tricky, so maybe an alternative would be to simply pass the escape keypress to the vim mode handler and prevent closing of the modal with it entirely (in which case, hopefully another keyboard shortcut could be used to close it? `:q` or `:wq` would be nice).

### Actual Behavior

Pressing escape closes the modal, even when in insert mode.

### Reproduction Steps

1. Open `Preferences > General`
2. Set `Text Editor Key Map` to `Vim`
3. Open `Manage Environments` (cmd+E)
4. press `i` and type some text OR select some text
6. press Escape

### Additional Information

Previously closed issue (although seemingly with slightly different behavior): #208

Perhaps this is a regression from the PR delivered against that issue (#1429)?

### Insomnia Version

2022.4.1

macOS

### Operating System Version

macOS Monterey 12.4 (21F79)

### Installation method

Homebrew 

Unsure, but I do think it previously worked as expected.

ian-h-chamberlain:
Fixes #4877

I got frustrated enough by this issue to try tackling it myself! This creates a new keyboard shortcut for closing Modals, which defaults to `Escape` (the previously hardcoded behavior). I'll probably set it to Ctrl-C for myself to restore the old-old behavior before #4877 was reopened.

There are a couple other places that had 'Escape' hardcoded for some kind of keydown event, but it seems mostly harmless - AFAIK this was the biggest painful hardcoded shortcut.

A couple notes on the implementation — I'm happy to add a test somewhere if that's wanted, but I wasn't sure exactly where to look. I did manual testing and it seems to work but I could use a pointer if there's a UI test that could correspond to this change.

Also, happy to take feedback on the description of the keyboard shortcut as well as the help text, I just plumbed in something quick to verify functionality.

### DIFF
--- a/packages/insomnia/src/common/hotkeys.ts
+++ b/packages/insomnia/src/common/hotkeys.ts
@@ -10,6 +10,7 @@ import { strings } from './strings';
 export const keyboardShortcutDescriptions: Record<KeyboardShortcut, string> = {
   'workspace_showSettings': `Show ${strings.document.singular} / ${strings.collection.singular} Settings`,
   'request_showSettings': 'Show Request Settings',
+  'workspace_closeModal': 'Close Modal Dialog',
   'preferences_showKeyboardShortcuts': 'Show Keyboard Shortcuts',
   'preferences_showGeneral': 'Show App Preferences',
   'request_quickSwitch': 'Switch Requests',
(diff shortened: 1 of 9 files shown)


kobenguyent:
FYI @ian-h-chamberlain https://github.com/Kong/insomnia/pull/6518#issuecomment-1735165221

ian-h-chamberlain:
Dang, thanks for the pointer, I guess this change might not be accepted either. I'd ask the maintainers to consider it as a stopgap until the refactor they mentioned is ready, but if that's happening soon then I suppose it's fine, as long as the use case in #4877 is still supported! 

Guess I should have tried to fix this much longer ago and maybe it could have worked in the meantime 😅

PR #6615, opened 2023-10-01
Remove message : "Welcome to the local Scratch Pad..." 

Jxtopher:
Hello team,

I suggest this modification because the message in top disrupts concentration. 
Is not ergonomic to work efficiently when user is not login.


filfreire:
@Jxtopher thanks for contributing this. We'll add a way to dismiss the message in another PR soon.

PR #6661, opened 2023-10-06
fix: only prettify if is a valid JSON #2506

### LINKED ISSUE #2506: Exception does not contain spaces
**Describe the bug**
I'm using java to develop a webservice and when I throw an exception it returns the correct response code, however the message does not contain spaces.

**To Reproduce**
1. throw exception that contains spaces in its messages in the app

**Expected behavior**
The returned message should contain spaces.

**Screenshots**
Insomnia:
image

Postman:
image

 - OS: Windows 10
 - Installation Method: Chocolatey
 - App Version: 2020.3.3

jacksjm:
changelog(Fixes): Fixed an issue that affected prettifying JSON request bodies

It was checked whether the return is a valid JSON, if so, it maintains the embellishment, if not, it directly returns the value obtained.

Following @johanhammar suggestion

Closes #2506

### DIFF
--- a/packages/insomnia/src/ui/components/codemirror/code-editor.tsx
+++ b/packages/insomnia/src/ui/components/codemirror/code-editor.tsx
@@ -221,7 +221,13 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(({
           jsonString = '[]';
         }
       }
-      return jsonPrettify(jsonString, indentChars, autoPrettify);
+      try {
+        // Try to parse the JSON to make sure it's valid
+        JSON.parse(jsonString);
+        return jsonPrettify(jsonString, indentChars, autoPrettify);
+      } catch (_) {
+        return jsonString;
+      }
     } catch (error) {
       // That's Ok, just leave it
       return code;
--- a/packages/insomnia/src/ui/components/codemirror/code-editor.tsx
+++ b/packages/insomnia/src/ui/components/codemirror/code-editor.tsx
@@ -221,16 +221,7 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(({
           jsonString = '[]';
         }
       }
-
-      // Check if it's stringified JSON or just a string
-      jsonString = jsonString.trim();
-      const isBaseString = jsonString.startsWith('"') && jsonString.endsWith('"');
-      const isBaseJson = jsonString.startsWith('{') && jsonString.endsWith('}');
-      if (isBaseString || isBaseJson) {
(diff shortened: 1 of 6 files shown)


packages/insomnia/src/ui/components/codemirror/code-editor.tsx
```diff
@@ -221,7 +221,13 @@ export const CodeEditor = forwardRef<CodeEditorHandle, CodeEditorProps>(({
           jsonString = '[]';
         }
       }
-      return jsonPrettify(jsonString, indentChars, autoPrettify);
+      try {
+        // Try to parse the JSON to make sure it's valid
+        JSON.parse(jsonString);
```
jackkav: The idea behind [jsonPrettify](https://github.com/jackkav/insomnia/blob/05d98968f972cee0968b72f0a143f3f95ab477c7/packages/insomnia/src/utils/prettify/json.ts#L30) is for it to work without parsing it into javascript, if we JSON parse it here, there would be no reason to use this jsonPrettify function. and this would serve us better. `JSON.stringify(jsonString,null,2)`

Another approach would be to add a case to the jsonPrettify function rather than throwing an error.

Also please provide a test case description.

jackkav (changes requested):

packages/insomnia/src/ui/components/codemirror/code-editor.tsx
jacksjm: @jackkav I really appreciate the feedback!
Do you believe the test case in jsonPrettify could use parsing in js?
I'm thinking about inserting the same validation into jsonPreffity after checking that the json is not empty, but I'm not sure if this is the best way. If you agree, I'll do it like this :smile: 
In order, to test the feature, I used a mocking REST (unfortunately via beeceptor: "https://test-mock-text.free.beeceptor.com", as mockbin does not allow this action) that returns a string without quotes. This way, when performing a GET on this service, the response in the Preview and SourceCode views are presented without space:
image
After the fixes:
image

jacksjm: @jackkav I had a different idea, I checked if the structure starts or ends on a String or JSON base.
If not, use JSON.stringify to convert to an understanding codification.
Please, review again if is a satisfactory solution.

Tests with GET endpoints:
image
image
image

jackkav:
Thanks @jacksjm I've pushed a commit that changes the underlying function and added a test. What do you think? @gatzjames

jackkav (approved):

filfreire:
@jacksjm you can claim a free tshirt for this PR contribution - find more at https://konghq.com/community/open-source-contribution

PR #6689, opened 2023-10-13
Local vault or remote project option for migration

gatzjames:
changelog(Improvements): Show an option for how the migration should happen.
- [x] Local vault (default)
- [x] Cloud projects

Highlights:
- [x] We keep using the existing migration logic to make sure all projects are moved under the organization locally.
- [x] After that we check the user preference (local/remote) and if remote try to move the local projects to remote. If anything fails (Api error,offline etc) projects are still under the org as local and users can turn them into remote from the project settings.


filfreire (approved):
Tested locally, LGTM

PR #6701, opened 2023-10-18
fix: presence is not refreshed in switching projects [INS-3221]

ihexxa:
changelog(Fixes): Fixed an issue where user presence feature was not refreshed when users switch between projects

Changes:
- For fetching collaborator effects, use `projectLoader` data as dependency instead of `projectsLoader`, because `projectLoader` is not refreshed in switching team project, it is only refreshed in entering one organization.
- Removed projectsLoader as it is duplicated


gatzjames (dismissed):

gatzjames:
Great catch! 
We might also need to update `present-users.tsx` with the same logic and remove the `projectLoader` in project.tsx and the loader from index.tsx

gatzjames (approved):

PR #6752, opened 2023-11-03
fix: interpolate graphQL vars before parsing them [INS-3256]

ihexxa:
WIP


ihexxa:
will push another one

PR #6753, opened 2023-11-03
Add learning feature and update the scope filter UI

gatzjames:
- [x] Adds a new dismissible section to the project sidebar about 'Insomnia Learning'. The contents of the section are fetched dynamically from the API.
- [x] Updates the size of the scope filters and removes the indentation
- [x] Fixes surprise button colour contrast

### DIFF
--- a/packages/insomnia/src/ui/routes/project.tsx
+++ b/packages/insomnia/src/ui/routes/project.tsx
@@ -406,18 +406,22 @@ export const loader: LoaderFunction = async ({
   };
 
   if (!window.localStorage.getItem('learning-feature-dismissed')) {
-    learningFeature = await window.main.insomniaFetch<{
-      active: boolean;
-      title: string;
-      message: string;
-      cta: string;
-      url: string;
-    }>({
-      method: 'GET',
-      path: '/insomnia-production-public-assets/inapp-learning.json',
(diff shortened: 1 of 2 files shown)


jackkav (commented):
Why localStorage over other persistence mechanisms?

filfreire (dismissed):
Checked locally, LGTM!

Screenshot 2023-11-08 at 10 18 52

gatzjames:
> Why localStorage over other persistence mechanisms?

Consistency since we use the same mechanism for dismissing the ScratchPad banner info

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -389,8 +397,32 @@ export const loader: LoaderFunction = async ({
...
+  let learningFeature = {
+    active: false,
+    title: '',
+    message: '',
+    cta: '',
+    url: '',
+  };
+
+  if (!window.localStorage.getItem('learning-feature-dismissed')) {
+    learningFeature = await window.main.insomniaFetch<{
```
gatzjames: Should have a try/catch around this since we don't want to fail the whole route in case this fetch fails

jackkav (approved):

PR #6757, opened 2023-11-05
feat: add reset passphrase [INS-3206]

marckong:
### Description
Adding the support for reset passphrase. When resetting the encryption passphrase, we are nullifying the encrypted content records in the remote (as users are the only ones to decrypt and no longer can decrypt it without knowing the passphrase).

This means we need to reconcile any local data that may still exist.
1. relinking these local files to the remote project
2. re-versioning these files and syncing to the cloud (that the users have opted in previously)
3. clean up the remote book-keeping record

The backend book-keeps the deleted file ids when resetting the passphrase and the electron app can know the account has reset the passphrase. It will nullify the existing session as the session data has to do with the encryption key, and the user is asked to re-sign in with the new passphrase (this is the case when user has chosen to reset passphrase during the sending invite flow, but still signed into the electron app => invitation requires passphrase for project encryption. Therefore, the encryption keys must be reconciled with the new keys).

The user will be redirected to the data recovery page where database operation and versioning operation run. During this operation, the remote will also clean up these records, so the user won't be in the state of reset passphrase flag all the time. The electron app will clear the session storage, and move onto the organization page.

### Steps to test
1. sign into your account in staging
2. create 3 team projects with cloud option
3. create files in each of them
4. invite a member
5. go to the forgot passphrase page and reset your passphrase
6. re-sign into the app
7. confirm all the files were where they were

### DIFF
--- a/packages/insomnia/src/account/session.ts
+++ b/packages/insomnia/src/account/session.ts
@@ -25,6 +25,7 @@ export interface WhoamiResponse {
   planId: string;
   canManageTeams: boolean;
   maxTeamMembers: number;
+  hasResetPassphrase: boolean;
 }
 
 export interface SessionData {
@@ -56,6 +57,7 @@ export async function absorbKey(sessionId: string, key: string) {
     accountId,
     firstName,
     lastName,
+    hasResetPassphrase,
   } = await _whoami(sessionId);
   const symmetricKeyStr = crypt.decryptAES(key, JSON.parse(encSymmetricKey));
 
@@ -73,6 +75,15 @@ export async function absorbKey(sessionId: string, key: string) {
     JSON.parse(publicKey),
     JSON.parse(encPrivateKey),
   );
+
+  if (hasResetPassphrase) {
+    // doesn't really matter what's saved here (it's true or false)
+    // session storage has this means the user reset the passphrase
+    // when resetting the passphrase, the session gets expired and requires re-login (mandatory due to encryption key update)
+    // therefore, it is 100% that the re-signing after reset in will get this value. => should clear it up after the cleanup screen
+    // TODO(@marckong): the whole session storing into localStorage probably needs redesign to be more robust. For now, we need to do this to ship reset passphrase
+    window.sessionStorage.setItem('hasResetPassphrase', 'true');
+  }
 
   window.main.loginStateChange();
 }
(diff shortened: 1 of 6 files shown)


filfreire (commented):
Tested locally, some notes

- **(FIXED IN https://github.com/Kong/insomnia/pull/6757/commits/070a3f759dad228ea787f7f6eb2c140d8897653f )** On reset of owner's organization  - the commit history was not preserved (if there was commit history already locally)
- Other personal organization members syncing after the owner of personal org resets passphrase can cause orphaned workspaces on the cloud - on the backend we need to add the same restriction to reset passphrase that we have for Team Organizations - of first removing the invited members and cancelling invites - before allowing to reset personal passphrase - this is a backend issue though - created ticket here INS-3308

packages/insomnia/src/account/session.ts
```diff
@@ -25,6 +25,7 @@ export interface WhoamiResponse {
   planId: string;
   canManageTeams: boolean;
   maxTeamMembers: number;
+  hasResetPassphrase: boolean;
```
gatzjames, line 28: What happens if I reset the passphrase another time?

packages/insomnia/src/account/session.ts
```diff
@@ -74,6 +76,15 @@ export async function absorbKey(sessionId: string, key: string) {
     JSON.parse(encPrivateKey),
   );
 
+  if (hasResetPassphrase) {
+    // doesn't really matter what's saved here (it's true or false)
+    // session storage has this means the user reset the passphrase
+    // when resetting the passphrase, the session gets expired and requires re-login (mandatory due to encryption key update)
+    // therefore, it is 100% that the re-signing after reset in will get this value. => should clear it up after the cleanup screen
+    // TODO(@marckong): the whole session storing into localStorage probably needs redesign to be more robust. For now, we need to do this to ship reset passphrase
+    window.sessionStorage.setItem('hasResetPassphrase', 'true');
```
gatzjames, line 85: This could happen again for the same account so not sure how we will handle it then.
Also what will happen when someone logs in with another account and this flag is true?

marckong: That case should be fine. This recovery relies on the data from the backup records table. If you sign into another account, this person would not receive the backup records. It will bypass. Ideally though we should clear the session storage out when logging out as well. Let me push that

Honestly though, our app cannot handle multi-account with the current database architecture.

marckong:
@filfreire [INS-3308](https://linear.app/insomnia/issue/INS-3308) addressed in the backend. It's merged in.

PR #6787, opened 2023-11-10
Fix: [Performance] - Transform sync to action instead of loader

gatzjames:
Use an action for sync to avoid refetching vcs data.


gatzjames:
Closing as this creates other syncing issues. A better approach would be to make VCS re-fetch only when it needs to to make it more performant

PR #6814, opened 2023-11-14
Fix: Make pinned requests always visible

gatzjames:
changelog(Fixes): Pinned requests are now always visible


marckong (approved):

marckong:
Has the smoke test been failing? and also would you mind adding some description or ticket number in the description for internal record?

gatzjames:
Tests probably failed on a flake. They pass now. It's too small of a change for a ticket and description would be duplicating the title 😅

PR #6822, opened 2023-11-16
fix: graphql request variables must be interpolated before parsing [INS-3256]

ihexxa:
Changes:
- interpolating graphql vars before parsing it


ihexxa:
This approach doesn't work

PR #6834, opened 2023-11-21
Cache VCS related queries

gatzjames:
Uses a simple cache to store VCS query results:
- Loader tries to fetch data if the cache is empty
- If there are cached data the loader uses only those.
- The cache gets refreshed explicitly by the new action.

Note:
This is a temp solution to avoid refetching the VCS related queries on the loader path.
It will to be replaced by a proper global network cache in the app.


filfreire (approved):
tested locally, LGTM

jackkav (commented):
Tasty 😋

PR #6837, opened 2023-11-21
refactor: add error route with override message

marckong:
- Implementing the product requirement to add error handling with a message
- Following the existing pattern as is for now

Closes INS-3354

### DIFF
--- a/packages/insomnia/src/ui/index.tsx
+++ b/packages/insomnia/src/ui/index.tsx
@@ -160,7 +160,7 @@ const router = createMemoryRouter(
           id: '/organization',
           loader: async (...args) => (await import('./routes/organization')).loader(...args),
           element: <Suspense fallback={<AppLoadingIndicator />}><Organization /></Suspense>,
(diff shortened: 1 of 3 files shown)


jackkav (commented):
It would be ideal if this over rode only the failed fetches rather than all errors. Since other errors may not be temporary and are important to uncover and handle in their own respective way.

marckong:
@jackkav Wouldn't the error stack and Sentry give us more details on the error? We could also log the error in the error.log specifically as well, but this was the requirement:
image

The error that we want to handle with this requirement happened during the rendering, not from the fetch failure directly. insomniaFetch does not throw the error, and it silently fails, so it continues rendering.

Another way would be hijack the error and set the message there, but I am not sure if that is really meaningful either. What would you suggest?

jackkav:
Lets not override the erorr messages, as we will need them for reproductions. 
Just add this message above the existing stuff.

marckong:
@jackkav would it be acceptable we display the default message (instead of override) when the error thrown without message? please LMK

filfreire (dismissed):

gatzjames (approved):

filfreire (approved):

PR #6843, opened 2023-11-22
Fix: deleting a key combination

gatzjames:
Deleting a key combination doesn't work.

This PR mutates the appropriate list of combinations that we pass to update the settings.

Next steps:
- Mutating a list like we do is error prone and can lead to other issues like this one.
- We should improve the api for adding/removing/updating a key combination in the hotkey registry 

changelog(Fixes): Fixed an issue where removing an existing keyboard shortcut would not work as expected

### DIFF
--- a/packages/insomnia/src/ui/components/settings/shortcuts.tsx
+++ b/packages/insomnia/src/ui/components/settings/shortcuts.tsx
@@ -112,13 +112,15 @@ export const Shortcuts: FC = () => {
                                 withPrompt
                                 onClick={() => {
                                   let toBeRemovedIndex = -1;
-                                  keyCombosForThisPlatform.forEach((existingKeyComb, index) => {
+                                  const keyCombs = getPlatformKeyCombinations(hotKeyRegistry[keyboardShortcut]);
+                                  keyCombs.forEach((existingKeyComb, index) => {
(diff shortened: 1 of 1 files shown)


filfreire (approved):

packages/insomnia/src/ui/components/settings/shortcuts.tsx
```diff
@@ -112,13 +112,15 @@ export const Shortcuts: FC = () => {
...
-                                  keyCombosForThisPlatform.forEach((existingKeyComb, index) => {
+                                  const keyCombs = getPlatformKeyCombinations(hotKeyRegistry[keyboardShortcut]);
+                                  keyCombs.forEach((existingKeyComb, index) => {
                                     if (areSameKeyCombinations(existingKeyComb, keyComb)) {
                                       toBeRemovedIndex = index;
                                     }
                                   });
                                   if (toBeRemovedIndex >= 0) {
-                                    keyCombosForThisPlatform.splice(toBeRemovedIndex, 1);
+                                    keyCombs.splice(toBeRemovedIndex, 1);
```
ihexxa, line 122: nit: just out of curious, where is the `keyCombs` consumed?

ihexxa (approved):

PR #6861, opened 2023-11-23
fix: graphql render issue caused by illegal Object.includes()

### LINKED ISSUE #6857: Render failure for Graphql on 8.4.3 and 8.4.4
### Expected Behavior

To Render the Query/Request window to work as is without any rendering issue.

### Actual Behavior

Render failure: e.includes is not a function

Screenshot 2023-11-22 at 10 59 47 AM

### Reproduction Steps

1. Upgrade Insomnia to 8.4.3 or 8.4.4
2. Open an Insomnia Project after upgrade (existing one not new)
3. and boom Request/Graphql Query Window throws the below error

Screenshot 2023-11-22 at 10 57 03 AM

### Additional Information

```
Version: Insomnia 8.4.4
Build date: 11/22/2023
OS: Darwin arm64 22.6.0
Electron: 27.0.3
Node: 18.17.1
Node ABI: 118
V8: 11.8.172.16-electron.0
Architecture: arm64

```

### Insomnia Version

8.4.4

macOS

### Operating System Version

Ventura 13.6

### Installation method

upgrade to existing one

8.4.2

marckong:
changelog(Fixes): Fixed issue #6857 related to GraphQL render issue caused by illegal Object.includes() method call. Also fixed an issue where GraphQL variables were not properly sent as an object when sending requests (both from Debug and Test view).

It seems we've tried to fix one issue and caused another issue. I took a very conservative approach here to fix the issue on top of the previous fix.

## Root Cause
Basically variables are passed as Object. Object does not contain 'includes()' method, and, therefore,  it throws an error.

## Solution
The fix here is in jsonPrettify method, we should give a default value as an empty string and stringify the object if passed instead of string in trycatch block. If it fails, we should return the default value, so it does not break the app. Ideally, I would prefer to show a toast when it happens, so the user can be aware. Currently it silently fails but the app still runs.

## TODO
- [x] solved issue with graphql variables field being sent as a string when it should be sent as an objectg
- [x] added reproduction smoke tests to check cases where we send Integer variables, Object variables, ... in Graphql requests

The codebase has very low test coverage, and in general, we should try increasing the coverage (despite little time we have)

Huge thanks to [camiloclc](https://github.com/camiloclc) who didn't mind jumping on the call and helped us reproduce it. 

Closes #6857 and all the other duplicate issues

### DIFF
--- a/packages/insomnia-smoke-test/fixtures/graphql.yaml
+++ b/packages/insomnia-smoke-test/fixtures/graphql.yaml
@@ -29,6 +29,58 @@ resources:
     settingRebuildPath: true
     settingFollowRedirects: global
     _type: request
+  - _id: req_63e5c8a8d1674bb39ba6df1d0eb452a1
+    parentId: wrk_d8bfe72fd18b42daa55d2ccc08c9eecb
+    modified: 1655289849596
+    created: 1655289826760
+    url: localhost:4010/graphql
+    name: GraphQL request with number
+    description: ""
+    method: POST
+    body:
+      mimeType: application/graphql
+      text: '{"query": "query($inputVar: Int) { echoNum(intVar: $inputVar)}", "variables": "{ \"inputVar\": {{ _.intVar }} }"}'
+    parameters: []
+    headers:
+      - name: Content-Type
+        value: application/json
+        id: pair_91da32f74847489ab06da669db0b425f
+    authentication: {}
+    metaSortKey: -1655289826761
+    isPrivate: false
+    settingStoreCookies: true
+    settingSendCookies: true
+    settingDisableRenderRequestBody: false
+    settingEncodeUrl: true
+    settingRebuildPath: true
+    settingFollowRedirects: global
+    _type: request
+  - _id: req_63e5c8a8d1674bb39ba6df1d0eb452a2
+    parentId: wrk_d8bfe72fd18b42daa55d2ccc08c9eecb
+    modified: 1655289849596
+    created: 1655289826760
+    url: localhost:4010/graphql
+    name: GraphQL request with variables
+    description: ""
+    method: POST
+    body:
+      mimeType: application/graphql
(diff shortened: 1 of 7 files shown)


suuunly:
When can we expect this to have been reviewed? :)
Currently, we are unable to work, as we rely heavily on GraphQL :)

lukepearce:
Your app is broken because it was able to release with basic JS errors?

mealbarracin10:
Hi all, When can we expect this to have been merged? I’m block 😕

CosminPerRam:
> Your app is broken because it was able to release with basic JS errors?

Irrelevant comment, this could and does happen from time to time in any project, what matters is fixing the problem as fast as possible and analyze how this got through and plan steps/procedures that avoids this from happening again.
Instead of asking aggressive questions we could ask how come this got through and if any plans to avoid this are being planned.

filfreire (dismissed):

marckong:
Thank you for your patience and support. We are actively working on testing the side effect of this. Going forward, we will add more Playwright smoke tests as much as we can. Especially, to prevent regression bugs.

@CosminPerRam well said. Thank you for your kindness.

To answer some of points you mentioned, we are working in codebase that lacks unit tests or smoke tests. We are also running thin. Many of us, including myself, were deployed to tackle engineering problems in the backend API side (the API that lives in the cloud, not in the electron codebase). Now some of us are brought back to the desktop app side, so we will do our best to improve as a team. When I hopped on a call with @camiloclc yesterday, I promised that we would put a PR for a fix in few hours and we did. We solved this problem as quickly as we could, and we are doing our best to unblock our users today.

Things will improve. Thanks for the patience.

Also this is not just "Basic JS errors" as it looks. It has multi-layers of things that contributors need to be aware of. This kind of comment is not very productive.

Thanks for collaboration on this - @camiloclc, @ihexxa, @gatzjames, @filfreire

suuunly:
@marckong thank you for the swift update :)
Most of us know the struggle of surprise bugs. 😆

PR #6870, opened 2023-11-24
 test: add tests for grpc operations

ihexxa:
### Background
Originally it aimed to fix a grpc issue but ends up with adding tests. : \

### Changes
- [x] test: add 2 cases for bidirectional and client side streaming


marckong:
Is this PR making style changes or actually adding or refactoring the changes?

nit:

Table driven tests are great when there are persistent and SIMPLE patterns for tests, but they are no longer great when such patterns break or done with complex patterns. I actually don't prefer table driven test approach at all for readability and difficulty to extend the test when it is needed.

In my opinion, tests should be so dumb that anyone reading a particular failed case can understand it easily. Table driven approach fails at this for me.

ihexxa:
@marckong yes, I'm checking if I'm able to quickly fix one gRPC requirement for specifying headers.

Good point of table driven testing 👍 , ideally tests should provide great readability and cover as more as possible of code, while practically tests will become bloated quickly if we keep adding test cases. Then it would introduce another kind of readability problem in maintaining, for example, we may need to read and refactor each test case if one new step is added to the workflow, as it is not DRY.

But currently it is not a problem for grpc, let me rollback to the traditional style.

ihexxa:
It might useful at sometime, closing this at the moment.

PR #6883, opened 2023-11-28
fix(untracked-workspaces): Migrate untracked workspaces and display them in the UI

gatzjames:
Handle untracked workspaces in the app.

- [x] Show the list of untracked workspaces that we can't automatically migrate in the `Settings > Data` tab like the untracked projects. We need to duplicate the workspace to avoid any document Id and backendProject conflicts

Future exploration:
- [ ] Show a warning to the user (toast notification) if there are any untracked workspaces/projects in the app. The warning has a link to open the `Settings > Data` panel so that users can view their data.
- [ ] Automatically migrate workspaces that belong to a project that the user has access to.

Closes INS-3373


marckong:
The API should be good to go in the feature branch. Currently in review. @gatzjames 

image

marckong:
@gatzjames also please keep in mind that insomniaFetch does not throw error when the request fails, including 400s, so we need to do something like

```ts
if ('error' in response) {
  // handle the error
}
```

jackkav (approved):

PR #6924, opened 2023-12-14
fix(tls): Use the system CA certificates to verify https servers by default

notjaywu:
depends on https://github.com/Kong/node-libcurl/pull/23 to be merged

by setting `tls.rootCertificate` on CAINFO_BLOB we were blocking macos and window from accessing the system CA store(keychain)

this PR will remove the setting of CAINFO_BLOB in the default case, unless a user adds a CA certificate manually in which case the system store will be ignored.

we will no longer use the nodejs provided `tls.rootCertificate`.

changelog(Improvements): Add support for system CA certificates

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2027,9 +2027,9 @@
       }
     },
     "node_modules/@getinsomnia/node-libcurl": {
-      "version": "2.3.6-21",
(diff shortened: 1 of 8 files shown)


jackkav (approved):
Tested on windows with Trusted Root Certification Authorities.

gatzjames (approved):

PR #6927, opened 2023-12-18
feat(untracked-data): Display a warning when there are untracked data in the app

gatzjames:
- Shows a warning in the status bar when there are untracked data in the app.
- Clicking the warning opens `Preferences > Data` so that users can view them

image

changelog(Improvements): Show a warning in the status bar when there are untracked data in the app


pavkout (approved):
LGTM

PR #6980, opened 2024-01-11
Fix parsing urlencoded pairs with empty values.

adamroyle:
There is a bug with parsing curl statements with empty values. This fixes the bug and adjusts the tests to expected behavior.

### DIFF
--- a/packages/insomnia/src/utils/importers/importers/curl.test.ts
+++ b/packages/insomnia/src/utils/importers/importers/curl.test.ts
@@ -123,7 +123,7 @@ describe('curl', () => {
       { flag: '--data-urlencode', inputs: ['=value'], expected: [{ name: '', value: 'value' }] },
 
       // --data-urlencode URI encoding
-      { flag: '--data-urlencode', inputs: ['a='], expected: [{ name: '', value: 'a=' }] },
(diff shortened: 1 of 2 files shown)


pexa-joeyzhou (approved):

filfreire:
Hi @adamroyle thanks for submitting this! We're eager to help you get this into a merged state.

While I was testing this PR out locally I've found some cases where we will still not properly render urlencoded pairs. Can you provide an example you used for testing this PR?

adamroyle:
I'm just using a simple example:

```
curl 'http://localhost/' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  --data-raw 'param1=&param2=&param3=value'
```

As an aside, however, I found another issue where Bash-style ANSI-C quoted strings (eg. --data-raw $'....') were not decoded properly. Google Chrome's "Copy as cURL" feature uses them for JSON strings that contain special chars.

eg.
```
curl 'http://localhost/' \
  -H 'Content-Type: application/json;charset=UTF-8' \
  --data-raw $'{"name":"O\'Reilly"}'
```

I spent a while trying to get it to work but ran into too many issues (mainly with general shell parsing libraries not implementing Bash-style quotes), and instead built a parser for "Copy as fetch (Node.js)" instead which has been much more reliable.

jackkav (approved):

PR #6994, opened 2024-01-18
Request path parameters

gatzjames:
Introduces path parameters to requests.
Path parameters are path segments that start with a colon `:` and can be used to describe variable segments of url paths.

This PR adds the ability to manage this parameters for requests in the request panel.

Highlights:
- [x] The Query tab on the response panel becomes the Parameters tab and has additional UI to manage and edit path parameters
- [x] Path parameter values are stored in the request/websocket-request models
- [x] The value of a path parameter can be a variable and is being rendered when the request is sent.

image

Closes INS-3442

changelog(Improvements): Introducing Path parameters. Path parameters are path segments that start with a colon `:` and can be used to describe variable segments of URLs.


marckong (approved):

ardaorkin:
That's great!

PR #7102, opened 2024-02-22
feat(hidden-window): enable baseEnvironment in the pre-request scripting [INS-3379]

ihexxa:
Changes:
- [x] : enable base environment in pre-request scripting
- [x] : make base environment modification work in rendering
- [x] : add tests for cover it

### DIFF
--- a/packages/insomnia-smoke-test/fixtures/smoke-test-collection.yaml
+++ b/packages/insomnia-smoke-test/fixtures/smoke-test-collection.yaml
@@ -177,7 +177,7 @@ resources:
     settingFollowRedirects: global
     _type: request
   - _id: req_89dade2ee9ee42fbb22d588783a9df3c
-    parentId: wrk_5b5ab67830944ffcbec47528366ef403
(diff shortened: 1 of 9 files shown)


filfreire (dismissed):
LGTM, only has one problem where maybe base environment should not override what we send on the pre-request


```
insomnia.baseEnvironment.set("testfilipe", "filipe")
pm.collectionVariables.set("testfilipe", "george")
```

gatzjames (commented):
We should also test environment overrides on the folder level.
If you put a request inside a folder, the folder has an option to add environment overrides that override environment variables

ihexxa:
Thank you for comments! The overriding issue is fixed.
Regarding folder environments, this is a really good point. The workflow might be a bit not intuitive: currently scripts mainly manipulates environment from "collection" level, although script is at request level. Then script's environment modifications will be overridden by folder environments. When the per-request variables are introduced, it can also be manipulated by script, which will override folder environments.

filfreire (approved):

PR #7187, opened 2024-03-18
fix: Export popup has small z-index [INS-3640]

### LINKED ISSUE #7176: Export popup has small z-index
### Expected Behavior

1) Go to «Preferences → Data»
2) Click to «Export the collection»

### Actual Behavior

Export dialog is not showing. If enable Electron's devtools, find there a dialog and add to it z-index couple of zeros it appeared

### Reproduction Steps

_No response_

### Additional Information

Video with steps to reproduce (in case of problems with playing [direct link to this video](https://inoy.dev/files/Insomnia%202024%20export.mp4)):

https://github.com/Kong/insomnia/assets/1235286/37e44f8e-902c-4d9f-8d89-4ed758c87380

### Insomnia Version

8.6.1

Windows

### Operating System Version

Windows 11

### Installation method

Downloaded from Github releases

Latest 2022 version (don't remember wich exactly)
```[tasklist]
### Tasks
```

CurryYangxx:
Close #7176 

Changes

- remove the global variable(globalZIndex) in base modal component
- add a new hook to admin every modal z-index(Even without using the base modal component)
- change the export request modal z-index, use globalModalZIndex instead

How to test

- Click Preferences → Data -> Export the collection, and you can see the export modal

### DIFF
--- a/packages/insomnia/src/ui/components/modals/settings-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/settings-modal.tsx
@@ -41,7 +41,7 @@ export const SettingsModal = forwardRef<SettingsModalHandle, ModalProps>((props,
   }), []);
 
   return (
-    <Modal ref={modalRef} tall {...props}>
+    <Modal className='!z-10' ref={modalRef} tall {...props}>
       <ModalHeader>
         {getProductName()} Preferences
         <span className="faint txt-sm">


ihexxa (dismissed):
Thanks for the quick fix, good job👍.

jackkav (commented):
~~Rather than syncing the zlndex's with a hook we could try to remove the zindex settings?~~ @gatzjames ?

gatzjames (commented):
We are moving away from the base/modal and using react-aria-components modals instead.
I'd suggest changing the settings-modal.tsx to have the same z-index as other modals instead.

In [settings-modal.tsx](https://github.com/gatzjames/insomnia/blob/cf9908a6a358a93131339b74ae985c05f150829c/packages/insomnia/src/ui/components/modals/settings-modal.tsx#L44) line 44
```js
<Modal className='!z-10' ref={modalRef} tall {...props}>
   ...
</Modal>
```
That way the last modal to open will always be on top and there's no need to keep calculations in state.

CurryYangxx:
> We are moving away from the base/modal and using react-aria-components modals instead. I'd suggest changing the settings-modal.tsx to have the same z-index as other modals instead.
> 
> In [settings-modal.tsx](https://github.com/gatzjames/insomnia/blob/cf9908a6a358a93131339b74ae985c05f150829c/packages/insomnia/src/ui/components/modals/settings-modal.tsx#L44) line 44
> 
> ```js
> <Modal className='!z-10' ref={modalRef} tall {...props}>
>    ...
> </Modal>
> ```
> 
> That way the last modal to open will always be on top and there's no need to keep calculations in state.

This way works. And do we have other similar scenes that import the base/modal? I'm worried that if there are similar scenes that are not covered by the same z-index, there may be the same issue. @gatzjames

gatzjames (approved):

gatzjames:
> This way works. And do we have other similar scenes that import the base/modal? I'm worried that if there are similar scenes that are not covered by the same z-index, there may be the same issue. @gatzjames

There are definitely more modals that use the previous approach.
I don't think there are other cases where we have this issue especially since opening a modal on top of another modal is offering a bad UX. So our goal is to eliminate these issues by migrating these modals instead of investing more on the previous approach

PR #7192, opened 2024-03-20
feat(session): Store session data in the db instead of LocalStorage

gatzjames:
Local Storage is not reliable to persist data across app restarts.

Highlights:
- [x] Creates a new model in the database to persist the user session
- [x] Uses the new database model in the app instead of Local Storage to read/write the session
- [x] Migrate existing sessions stored in local storage to the db?

Closes INS-3633

Next steps:
- [ ] Cleanup the session module and simplify the logic
- [ ] On later versions remove the migration from localStorage
- [ ] On unauthorised response from our API we should log out the user
- [ ] Explore multi-database per account or similar approach

### DIFF
--- a/packages/insomnia/src/account/session.ts
+++ b/packages/insomnia/src/account/session.ts
@@ -40,8 +40,8 @@ export interface SessionData {
   encPrivateKey: crypt.AESMessage;
 }
 export function onLoginLogout(loginCallback: LoginCallback) {
-  window.main.on('loggedIn', () => {
-    loginCallback(isLoggedIn());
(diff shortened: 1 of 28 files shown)


marckong:
@gatzjames will the user model be excluded from the git and cloud syncs?

gatzjames:
> @gatzjames will the user model be excluded from the git and cloud syncs?

Correct it's not shared like the settings model

marckong (dismissed):

marckong (approved):

Issue #7154, opened 2024-03-07
Path parameter evaluation fails if endpoint includes a colon


pgoldtho:
### Expected Behavior

![Screenshot 2024-03-07 at 6 46 09 PM](https://github.com/Kong/insomnia/assets/13651001/c7b6a3f5-b21e-4308-a701-d8023403bce9)

### Actual Behavior

![Screenshot 2024-03-07 at 6 45 56 PM](https://github.com/Kong/insomnia/assets/13651001/8de92dc4-851d-406a-aa3c-d2ca40407e8f)

### Reproduction Steps

Enter a URL with a path parameter as the final argument immediately followed by `:someString`

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Additional Information

Google Chronicle APIs include some endpoints that use a colon to indicate a function call. For example the [forwarders.generateForwarderFiles](https://cloud.google.com/chronicle/docs/reference/rest/v1alpha/projects.locations.instances.forwarders/generateForwarderFiles) method follows the form:

`GET https://chronicle.googleapis.com/v1alpha/{name}:generateForwarderFiles`

### Insomnia Version

Version: Insomnia 8.6.1 Build date: 2/5/2024 OS: Darwin x64 23.2.0 Electron: 28.0.0 Node: 18.18.2 Node ABI: 119 V8: 12.0.267.8-electron.0 Architecture: x64

### What operating system are you using?

macOS

### Operating System Version

14.2.1 (23C71)

### Installation method

download from insomnia website

### Last Known Working Insomnia version

n/a

pgoldtho:
Was able to fix the issue in a dev environment by adding a colon to the PATH_PARAMETER_REGEX at line 90 of 

packages/insomnia/src/models/request.ts

Before:
```
export const PATH_PARAMETER_REGEX = /\/:[^/?#]+/g;
```

After:
```
export const PATH_PARAMETER_REGEX = /\/:[^/?#:]+/g;
```

I don't have permissions to push branches to the repo so can't submit a pull request

andrea-mauro:
@pgoldtho have you tried to fork the repository and push the change to the forked version? Once the forked repo is ahead of the original you should be able to create a Pull Request.

<img width="929" alt="image" src="https://github.com/Kong/insomnia/assets/54074100/c118726a-4163-4508-95bc-44a45ac01ef2">

pgoldtho:
[...]

PR #7200, opened 2024-03-23
Support for colon after a path parameter

pgoldtho:
fix for #7154


filfreire (approved):
LGTM, checked locally, handles the case mentioned by the user in #7154

PR #7205, opened 2024-03-25
Fix local state updates for GraphQL Query Variables

### LINKED ISSUE #7204: Unable to remove/clear `Query Variables` from an API test request
### Expected Behavior

Deleting the JSON data in the `Query Variables` section should save when sending a request/returning to the request document page. This also stops anyone from switching through multiple requests in the same test document if any of the requests require different query variables. 

### Actual Behavior

Reloading the page in any way (ex. clicking off+on the page or sending a request) will return the JSON body to the input box. 

### Reproduction Steps

1. Open a new test document
2. Begin editing a GraphQL request to an endpoint 
3. Add `Query Variables` to the request 
4. Make a request with those variables (successfully)
5. Try to remove those variables and send a new request which doesn't require the variables

### Additional Information

Full insomnia version info: 

```
Version: Insomnia 8.6.1
Build date: 2/5/2024
OS: Darwin arm64 23.4.0
Electron: 28.0.0
Node: 18.18.2
Node ABI: 119
V8: 12.0.267.8-electron.0
Architecture: arm64
```

### Insomnia Version

v8.6.1

macOS

### Operating System Version

MacOS 14.4 (23E214)

### Installation method

Web download

_No response_

MKHokinson:
Closes #7204

### DIFF
--- a/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
+++ b/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
@@ -47,7 +47,7 @@ function getGraphQLContent(body: GraphQLBody, query?: string, operationName?: st
     content.query = query;
   }
 
-  // The below items are optional; should be set to empty if undefined
+  // The below items are optional; should be set to undefined if present and empty
   if (operationName !== undefined) {
     content.operationName = operationName.length ? operationName : undefined;
   }
--- a/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
+++ b/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
@@ -48,11 +48,12 @@ function getGraphQLContent(body: GraphQLBody, query?: string, operationName?: st
   }
 
   // The below items are optional; should be set to undefined if present and empty
-  if (operationName !== undefined) {
+  const isString = (value?: string): value is string => typeof value === 'string' || (value as unknown) instanceof String;
+  if (isString(operationName)) {
     content.operationName = operationName.length ? operationName : undefined;
   }
 
-  if (variables !== undefined) {
+  if (isString(variables)) {
     content.variables = variables.length ? variables : undefined;
   }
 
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -47,12 +47,13 @@ function getGraphQLContent(body: GraphQLBody, query?: string, operationName?: st
     content.query = query;
   }
 
-  if (operationName) {
-    content.operationName = operationName;
+  // The below items are optional; should be set to undefined if present and empty
+  if (operationName !== undefined) {
```
ihexxa: We may also check if it is null or `operationName.length` would fail below.

packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
```diff
@@ -47,12 +47,13 @@ function getGraphQLContent(body: GraphQLBody, query?: string, operationName?: st
...
+    content.operationName = operationName.length ? operationName : undefined;
```
ihexxa, line 53: We might check it with `Array.isArray`

ihexxa (commented):
Thanks for the fix 👍  , added some minor comments.

packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx
MKHokinson: I don't think `Array.isArray` will work for validating a `string` which would be expected for `operationName`, but we could cover both checks by asserting that the `operationName` and `variables` params are strings. I'd argue that `null` should never be given as an input anyway and it would be caught by the type checker, but I don't think there's harm in making the check anyway. 

Pushed an updated approach

MKHokinson: Not sure what your thoughts are on including the `(value as unknown) instanceof String` check is. I included it for completeness but I'm guessing there's no usage of the `String` wrapper object anywhere in the codebase.

ihexxa: My bad, it's fine to include `(value as unknown) instanceof String` although `typeof` should be good enough for most of cases.

ihexxa (approved):

PR #7229, opened 2024-04-02
fix: url is encoded during execute pre-request script [INS-3681]

CurryYangxx:
ipt

close #7219 

When user use pre-request script, the url in request data will be encoded during execute the script.
In order to recognize variable, we need to decode url

PR #7249, opened 2024-04-07
fix(pre-request script): avoid encoding tags in parsing request urls - INS-3379

ihexxa:
Background:
In pre-request script execution, originally active request urls will be parsed with URL web API. But in some cases, tags like nunjucks tags would be encoded which would introduce problems in manipulation and sending request.
This change parses urls manually for avoiding this problem.

Changes:
- [x] avoid encoding tags in parsing request urls
- [x] clean up request body transforming
- [x] add tests


CurryYangxx:
Does this change  #7229  need to be reverted?

ihexxa:
@CurryYangxx Good catch 👍, updated.

jackkav (dismissed):

jackkav (approved):

PR #7259, opened 2024-04-11
feat: add fallback for searching a proper client certificate for a host [INS-3680]

CurryYangxx:
changes:

- [x] add a util function filterClientCertificates and add fallback mechanism (don't check port matching) for finding client certificate
- [x] add a new param `needCheckPort` in urlMatchesCertHost
- [x] use filterClientCertificates in both https and wss logic
- [x] add test case


packages/insomnia/src/network/__tests__/url-matches-cert-host.test.ts
```diff
@@ -38,6 +38,12 @@ describe('urlMatchesCertHost', () => {
       expect(urlMatchesCertHost(certificateHost, requestUrl)).toBe(false);
     });
 
+    it('should return true if the request URL and certificate host have different ports and the needCheckPort parameter is false', () => {
```
notjaywu, line 41: Does the case below return true?
```
      const requestUrl = 'https://www.example.org:1234/some/resources?query=1';
      const certificateHost = 'https://www.example.org';
```

ihexxa (dismissed):
Probably we should consider to add a test case as @notjaywu proposed, otherwise lgtm.

packages/insomnia/src/network/__tests__/url-matches-cert-host.test.ts
CurryYangxx: Yes, I added more test cases and the results are meet the requirements

ihexxa (approved):

PR #7265, opened 2024-04-12
fix: wildcard pattern matching

### LINKED ISSUE #7181: Regression: wildcard values don't work for the host of certificates
### Expected Behavior

Wildcard values such as `*.mydomain.com` should work for host matching certificates for mTLS

### Actual Behavior

Wildcard values don't match properly and each subdomain must be added individually. 

Seems to be a regression of https://github.com/Kong/insomnia/issues/243

### Reproduction Steps

_No response_

### Additional Information

_No response_

### Insomnia Version

8.6.1

macOS

### Operating System Version

Sonoma 14.4

### Installation method

insomnia.rest

_No response_

hvitoi:
Closes #7181

## Rationale

The `new URL` keyword was previously instantiating an unknown class which does encode the hostname characters so that `https://*.myhost.com/` was turned into `https://%2A.myhost.com` and the latter would not match with the actual url being requested and thus the certificate would not be attached to the request itself.

This fix forces the use of the [node:url module](https://github.com/nodejs/node/blob/v20.11.0/lib/url.js) which does return a more predictable output (and not encoding the star symbol).

It's not clear why this bug is not caught by the test suite since this exact test case [is already validated](https://github.com/Kong/insomnia/blob/2d9e2da52ee6c4e403cb378e36ef4568caf113b6/packages/insomnia/src/network/__tests__/url-matches-cert-host.test.ts#L117). If you have any suggestion we could explore other test scenarios.

### DIFF
--- a/packages/insomnia/src/network/url-matches-cert-host.ts
+++ b/packages/insomnia/src/network/url-matches-cert-host.ts
@@ -1,4 +1,4 @@
-import { parse as urlParse } from 'url';
+import { parse as urlParse, URL } from 'url';
 
 import { escapeRegex } from '../common/misc';
 import { setDefaultProtocol } from '../utils/url/protocol';


subnetmarco:
Thanks for the contribution. @filfreire can you review it?

hvitoi:
What are the next steps on this? @filfreire Is the any action left?

hvitoi:
What are the next steps on this? @filfreire Is there any action left?

filfreire (approved):

filfreire:
@jackkav @gatzjames @CurryYangxx if possible please give a second pass to this

jackkav:
Nice catch, can confirm. I suspect the tests pass because they run in a node context so web apis are polyfilled with node ones.

image
image

jackkav (approved):

PR #7300, opened 2024-04-23
feat(event-log): Improve UX of the event log for WS and SSE responses

gatzjames:
Highlights:

- [x] Adds navigation with keyboard
- [x] The event view is now resizable
- [x] Navigating through the events displays the event in the event view


jackkav (approved):
🎸 🪨 🧻

PR #7330, opened 2024-04-28
Fix/environment edit modal save

### LINKED ISSUE #5612: Environment variables at folder level not saving state
Description:
When updating an environment variable at the folder level, exiting and then going back into environment variables doesn't reflect and reverts to previous assigned value.

Step 1: Create new variable "Test"
image
Step 2: Edit Variable and set header name to 'A'
image
Step 3: Exit to folder view
Step 4: Open env variables for the same folder.
Step 5: Open test assignment.
Step 6: Variables have not been updated.
image

Sometimes we found waiting ~1 minute allows config/env variable to save.

oahmed-OS:
### Background
For environment override modifications, the current behavior did not persist these changes consistently. This is not consistent with the expected behavior of persisting changes as the user closes and re-opens the modal after submitting some change.

This was caused by two issues:
- onBlur was not firing consistently on modal close
- Updated values was not loaded in debug route collection and was not re-rendered on change

### Changes

- [x] Capture save on hide rather then on blur
- [x] Update collection in debug route when change detected using new collection state
- [x] add smoke test

Closes #5612


oahmed-OS:
failing smoke tests, found more acceptable solution - will open a separate PR.

Issue #4241, opened 2021-11-23
Insomnia very slow and hangs on a >6000 line file


chaosKP:
### Expected Behavior

If the DESIGN is used the toggle of a menu entry inside the preview is nearly instantious.
If a change is made inside the editor the lag is neglectable (1-2 seconds).

### Actual Behavior

If a path entry inside the preview is toggled it takes up to 5 seconds to respond.
If a change is made inside the editor it takes more than 25 seconds until the tool is updated and responsive again.

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Additional Information

There is no difference if Preview is On or Off.

We also tried a local version of swagger editor and the expected behavior is given

### Insomnia Version

2021.6.0

### What operating system are you using?

Windows

### Operating System Version

Windows 10 Build 19043

### Installation method

download from insomnia.rest

tuscen:
Can confirm, the same behaviour on macOS 12.1 on Macbook with Intel Core i7. I have an openapi schema with 17k lines and Insomnia handles it very badly. Changing from design to debug tab can take up to several seconds (4-6 seconds).

```
Version: Insomnia 2021.7.2
Release date: 12/10/2021
OS: Darwin x64 21.2.0
Electron: 11.2.3
Node: 12.18.3
V8: 8.7.220.31-electron.0
Architecture: x64
node-libcurl: libcurl/7.73.0 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.9 zstd/1.4.9 libidn2/2.1.1 libssh2/1.9.0 nghttp2/1.42.0
```

poiuytrez:
I struggle to edit a 1.2k lines file in the design tab. Every time I type a few characters, the editor hangs for about 5-10 seconds which makes the edition very painful.

This is what I often get:
![Screenshot 2022-04-08 at 10 20 44](https://user-images.githubusercontent.com/1167433/162395676-7ac75511-c1e6-44eb-9960-b4b7b0443af2.png)

Any tip to disable real-time syntax checking is welcome.

```
Version: Insomnia 2022.2.1
Release date: 3/21/2022
OS: Darwin x64 20.6.0
Electron: 17.1.0
Node: 16.13.0
V8: 9.8.177.11-electron.0
Architecture: x64
```
I am on macOS 11.6.5 on Macbook Air 2018.

ArnaudPel:
[...]

PR #7374, opened 2024-05-06
perf: improve performance when spec lint [INS-3724]

### LINKED ISSUE #4653: Insomnia has become unresponsive open api specs 3
### Expected Behavior

Load Specs design document properly.

### Actual Behavior

Showing an error of Insomnia has become unresponsive, and there is no way i can remove this import from my workspace. 

### Reproduction Steps

1. Goto Settings > Data > Import data > From File
2. Open API spec with some error in specification and it has to be large. 

### Additional Information

Version: Insomnia 2022.2.1
Release date: 22/03/2022
OS: Windows_NT x64 10.0.19043
Electron: 17.1.0
Node: 16.13.0
V8: 9.8.177.11-electron.0
Architecture: x64

### Insomnia Version

2022.2.1

Windows

### Operating System Version

Windows_NT x64 10.0.19043

### Installation method

download from insomnia.rest

_No response_

CurryYangxx:
Close https://github.com/Kong/insomnia/issues/4653 https://github.com/Kong/insomnia/issues/4241

**Move spec lint logic to web worker, can improve both opening file and editing file**

Changes:

- [x] move spectral lint login to web worker both in the Design component and router action
    - [x] add dedicate worker file
    - [x] delete all the spectral in main process
- [x] open electron nodeIntegrationInWorker option
- [x] add vite worker plugin config so we can use electron require magic in worker
- [x] wrap code-editor with memo to reduce rerender

How to test:
- [x] trigger lint multiple times and only the latest task result will be updated
- [x] use custom spectral ruleset and it it works


jackkav:
@CurryYangxx very elegant approach. With node access in webworkers we have opened up some interesting performance options.

tsconfig.base.json
```diff
@@ -10,7 +10,7 @@
     "esModuleInterop": true,
     "lib": ["DOM", "ES2019"],
     "module": "CommonJS",
-    "moduleResolution": "node",
```
jackkav: image
curious

packages/insomnia/src/ui/routes/design.tsx
```diff
@@ -251,24 +229,71 @@ const Design: FC = () => {
...
   useEffect(() => {
     CodeMirror.registerHelper('lint', 'openapi', async (contents: string) => {
-      const diagnostics = await window.main.spectralRun({
-        contents,
-        rulesetPath,
+      // @TODO: Conisderations and things we need to verify/test:
+      // - Should we spawn a new worker on every lint command or use a shared worker?
+      // - Does this function handle race conditions? e.g. if a lint finishes late does it replace another lint that ran with latest contents?
+      // - Proper error handling from the worker and here. Perhaps the worker returns a result type like Result = Diagnostics | Error and we handle it here
+      // - Don't reload the spectral ruleset on every lint command. Instead the ruleset should be re-calculated when the underlying file changes due to e.g. git pull or the filename changes
```
jackkav: > Should we spawn a new worker on every lint command or use a shared worker?

it depends on the error handling story

> Does this function handle race conditions?

no if you always spawn a new one, although since it can happen on keystroke we might need to throttle the creation of workers, or resetting a shared one.

> Proper error handling from the worker and here.

the simplest way I can imagine is wrapping the worker in a promise and using abort and reject when listening on the worker error event.

> Don't reload the spectral ruleset on every lint command
agreed, but perhaps we can consider this is an optimisation for later.

CurryYangxx:
> Should we spawn a new worker on every lint command or use a shared worker?

Now it will create just one worker when component first time render.And use a unique id to distinguish worker message.

> Does this function handle race conditions?

Use the unique id for every message, only the result which id match the latested id will be updated.

> Don't reload the spectral ruleset on every lint command

use a cache object in worker, only reload when ruleset path change

CurryYangxx:
We still need to save spectralRun IPC because it is used in this action:https://github.com/Kong/insomnia/blob/cc421d1a3c041cebc059f5d421c4406c8fe75baf/packages/insomnia/src/ui/routes/actions.tsx#L736

packages/insomnia/src/ui/routes/actions.tsx
```diff
@@ -772,7 +776,9 @@ export const generateCollectionAndTestsAction: ActionFunction = async ({ params
     `version-control/git/${workspaceMeta?.gitRepositoryId}/other/.spectral.yaml`,
   );
 
-  const results = (await window.main.spectralRun({ contents: apiSpec.contents, rulesetPath })).filter(isLintError);
+  const spectralRunner = new SpectralRunner();
+
+  const results = (await spectralRunner.runDiagnostics({ contents: apiSpec.contents, rulesetPath })).filter(isLintError);
```
ihexxa, line 781: do we also need termination here?

packages/insomnia/src/ui/worker/spectral.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const cachedRuleset: {
+  path: string;
+  ruleset: RulesetDefinition;
+} = {
+  path: '',
+  ruleset: oas as RulesetDefinition,
+};
+
+const loadRuleset = async (rulesetPath: string,) => {
```
ihexxa: nit: `(rulesetPath: string,)` the comma might be unnecessary.

packages/insomnia/src/ui/worker/spetral-run.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  terminate() {
+    this.worker.terminate();
+  }
+
+  public async runDiagnostics({ contents, rulesetPath }: { contents: string; rulesetPath: string }) {
+    this.taskId = ++this.taskId;
+    return new Promise<ISpectralDiagnostic[]>(resolve => {
+      this.worker.onmessage = e => {
```
ihexxa: nit: Probably we might also handle `postMessage(err);` and add some logs for debugging.

ihexxa (dismissed):
Discussed offline and resolved several concerns, just added some minor comments. Overall it looks good to me but let's wait a bit as there might be some comments from other reviewers.

CurryYangxx:
> Discussed offline and resolved several concerns, just added some minor comments. Overall it looks good to me but let's wait a bit as there might be some comments from other reviewers.

@ihexxa Good catch, update more logs and necessary termination

gatzjames (dismissed):

ihexxa (approved):

PR #7396, opened 2024-05-13
feat: enable extended assertion chains on `insomnia.response`

ihexxa:
Closes: INS-3824

### Changes:
- [x] feat: enable extended assertion chains on `insomnia.response`
- [x] add unit tests

### DIFF
--- a/packages/insomnia/src/ui/components/editors/request-script-editor.tsx
+++ b/packages/insomnia/src/ui/components/editors/request-script-editor.tsx
@@ -58,6 +58,24 @@ const updateRequestAuth =
 );`;
 const requireAModule = "const atob = require('atob');";
 
+const getStatusCode = 'const statusCode = insomnia.response.code;';
+const getStatusMsg = 'const status = insomnia.response.status;';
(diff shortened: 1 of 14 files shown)


packages/insomnia/src/ui/components/editors/request-script-editor.tsx
```diff
@@ -380,6 +398,70 @@ export const RequestScriptEditor: FC<Props> = ({
...
+            />
+          </DropdownItem>
+          <DropdownItem textValue='Get cookies' arial-label={'Get cookies'}>
+            <ItemContent
+              icon="circle-info"
+              label='Get cookies'
+              onClick={() => addSnippet(getCookies)}
+            />
+          </DropdownItem>
+        </Dropdown>
```
CurryYangxx: Can we use an data array to generate these DropdownItems?

ihexxa: Good point, which will reduce some duplications. They are kept because sometimes we add components like `DropdownSection`s which might need additional efforts to map them.

CurryYangxx (approved):

PR #7426, opened 2024-05-17
fix: migrate loader redirect

CurryYangxx:
The `shouldMigrateProjectUnderOrganization` is an async function, so we need to add `await`


jackkav (commented):
Has this been tested? Seems like a sensitive code path. @gatzjames seems like the existing e2e test isn't covering this case correctly

gatzjames (approved):

PR #7433, opened 2024-05-20
fix(Sync): Display errors on sync branches modal

gatzjames:
Currently errors are not displayed properly inside the modal for syncing branches.

Highlights:
- [x] Adds proper error handling

Next steps:
- [ ] Move the branches modal functionality in the sidebar to improve UX as alerts on top of a modal are creating a lot of friction

Closes INS-3850


jackkav (approved):
this use effect idle checker pattern is smelly and error prone, is there any hope of something better?

gatzjames:
> this use effect idle checker pattern is smelly and error prone, is there any hope of something better?

React 19/router v7 will have the new actions api which will remove all this dependency management. Currently not sure if there's a better approach

PR #7435, opened 2024-05-20
Added Organization to OAuth2.

CallumJohnson:
Closes: https://github.com/Kong/insomnia/discussions/7209

This is an optional parameter on OAuth2 endpoints for 'Business Users' in OAuth2 applications.
image
This required affiliation with an organization, therefore requires this option to be specified on the request.

### DIFF
--- a/packages/insomnia-sdk/src/objects/__tests__/auth.test.ts
+++ b/packages/insomnia-sdk/src/objects/__tests__/auth.test.ts
@@ -169,6 +169,7 @@ describe('test auth transforming', () => {
             pkceMethod: 'pkceMethod',
             responseType: 'id_token',
             origin: 'origin',
+            organization: 'org',
         };
 
         [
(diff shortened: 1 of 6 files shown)


jackkav (commented):
I wonder if it might be better to add a custom values table rather than a new hard coded key, I'll discuss with the team, thanks for your contribution!

CallumJohnson:
Thanks for your quick response; a custom values table would be great for the OAuth2 flow.

jackkav:
I saw in postman they have three phases with custom query params, would some like that satisfy your use case? Or am I misunderstanding something?
image

CallumJohnson:
Absolutely, we're currently using the Auth Requests section to provide this context to our OAuth2 flow. Adding a table like this would be very powerful!

Thanks again for your response.

subnetmarco:
@jackkav What would be the next step here?

jackkav:
@subnetmarco we're tracking it in INS-3876 but it needs to be assigned.

gatzjames:
Thanks for the PR.
To fully support this along with other parameters for OAuth2 we need to add a generic parameter editor.
Closing this for now as we will address it in an upcoming PR and is already planned internally

PR #7465, opened 2024-05-29
feat(Key-Value Editor): Improve accessibility and make the items re-orderable

### LINKED ISSUE #5686: Can no longer reorder headers in the request panel
### Expected Behavior

In previous versions of Insomnia as can be seen by this screenshot, I was able to reorder my headers from the request panel with a little drag panel that was seen on the left of each item. I was also able to add new headers by just typing into the "new header" box that appeared at the bottom of the list.

image

### Actual Behavior

Now when opening Insomnia 2022.7.1, I don't see the reorder panel or the "New Header" box and can only add headers via the onscreen button:

image

I'm not sure if this is some sort of bug I'm facing or intentional removal, but it seems like a regression from earlier versions as I can no longer find a way to order my headers except for deleting them and adding them again. This removes a ton of convenience and organization from my requests and makes it harder to parse through my headers at a glance. The new header box was also convenient as it saved a click when I wanted to add a new header, but for me it's not as drastic an adjustment as the lack of reordering. 

If this is an intentional change, I would hope that there could at least be an option to restore the old behavior as the data organization aspects are why I use Insomnia in the first place. But I'm not seeing anything like that in the preferences right now. 

### Reproduction Steps

1. Create a new request
2. Add 2 headers in the Header panel
3. Observe that there is no way to change the header order

### Additional Information

Here is the "About Info" provided by Insomnia: 

Version: Insomnia 2022.7.1
Build date: 1/9/2023
OS: Windows_NT x64 10.0.19044
Electron: 22.0.0
Node: 16.17.1
V8: 10.8.168.20-electron.0
Architecture: x64

### Insomnia Version

2022.7.1

Windows

### Operating System Version

Windows 10

### Installation method

Downloaded from internet

2022.5(?)

gatzjames:
Highlights:

- [x] The key-value editor is now inside a gridlist and has more accessible keyboard navigation for actions and re-ordering
- [x] Key-value pairs can now be re-ordered with keyboard/drag-and-drop interactions

Extras:
- [x] Cleans up unused components
- [x] Fix height of graphql editor

TODO:
- [x] Updating the list of key-value pairs should happen on an event instead of a useEffect during render

Closes #5686


packages/insomnia/src/ui/components/editors/request-headers-editor.tsx
```diff
@@ -78,6 +78,8 @@ export const RequestHeadersEditor: FC<Props> = ({
     headersString += `${header.name}: ${header.value}\n`;
   }
 
+  const handleChange = useCallback((headers: RequestHeader[]) => patcher(id, { headers }), [id, patcher]);
```
CurryYangxx: the `patcher` in the dependency param seems will change in every render, because `useRequestPatcher` always return a new function

gatzjames: Good one 👍 
I'm not a fun of useCallback especially with the react-compiler being OSS now
Added this when I was firing onChange in a useEffect but as you say it's not working as expected and also not needed anymore since onChange fires on events

CurryYangxx: By the way,if we wan't a memorized function, we could return a memorized function in `useRequestPatcher `, and use `fetcher.submit` as dependency. Because fetch.submit is stable(https://github.com/remix-run/react-router/pull/10336).
But I am not sure if we need to change it in this situation

CurryYangxx (approved):

PR #7473, opened 2024-06-03
fix: minor typo in interpolator

pimothyxd:
While skimming through the codebase I noticed a small typo.


ihexxa (approved):
Thanks for fixing it 👍.

PR #7474, opened 2024-06-03
fix: add a missing assertion chain for the response entity - INS-3917

ihexxa:
### Changes
- Supported negative assertion chain such as `insomnia.response.to.not.have.status(200);`
- Added utests
- Only keep one instance for `collectionVariables` and `baseEnvironment` so that operations triggered from different entities will take effect.

Ref: - INS-3917

### DIFF
--- a/packages/insomnia-sdk/src/objects/insomnia.ts
+++ b/packages/insomnia-sdk/src/objects/insomnia.ts
@@ -126,7 +126,6 @@ export async function initInsomniaObject(
     const baseEnvironment = new Environment(rawObj.baseEnvironmentName || '', rawObj.baseEnvironment);
     // TODO: update "iterationData" name when it is supported
     const iterationData = new Environment('iterationData', rawObj.iterationData);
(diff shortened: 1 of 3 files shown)


CurryYangxx (approved):

jackkav (approved):

packages/insomnia-sdk/src/objects/insomnia.ts
```diff
@@ -140,7 +139,7 @@ export async function initInsomniaObject(
     const variables = new Variables({
         globals,
         environment,
-        collection: collectionVariables,
+        collection: baseEnvironment,
```
gatzjames, line 142: Is collection accepting environments? If so maybe we should change the name to make more clear

ihexxa: Good catch, it tries to keep one one base environment instance so that operations triggered from `variables` would also be applied to the base environment.
Let me start another PR to rename them.

PR #7477, opened 2024-06-03
fix(Response Tabs): Tabs with a menu inside are not accessible - Response Panes

gatzjames:
Highlights:

- [x] Updates the request pane's tabs to be keyboard accessible by moving the menu inside the tab panel
- [x] Fixes an issue where choosing `Cancel` when switching body type won't work
- [x] Improves GraphQL editor overflow by adding resizable panes and making navigation more accessible
- [x] Fixes overflow issue with tab panel content not scrolling properly
- [x] Fixes overflow issue with menus not scrolling when exceeding the screen limits

TODO:
- [x] Update e2e tests

Next steps:
- [ ] Fix tab issues in Response Panes
- [ ] Look into other tabs that might have the same issue

### DIFF
--- a/packages/insomnia/src/ui/components/dropdowns/content-type-dropdown.tsx
+++ b/packages/insomnia/src/ui/components/dropdowns/content-type-dropdown.tsx
@@ -151,8 +151,6 @@ export const ContentTypeDropdown: FC = () => {
   const hasParams = body && 'params' in body && body.params;
   const numBodyParams = hasParams ? body.params?.filter(({ disabled }) => !disabled).length : 0;
 
-  console.log('bodytype', body.mimeType);
-
   return (
     <Select
       aria-label="Change Body Type"
--- a/packages/insomnia/src/ui/components/dropdowns/content-type-dropdown.tsx
+++ b/packages/insomnia/src/ui/components/dropdowns/content-type-dropdown.tsx
@@ -1,4 +1,6 @@
+import { IconName } from '@fortawesome/fontawesome-svg-core';
 import React, { FC } from 'react';
+import { Button, Collection, Header, ListBox, ListBoxItem, Popover, Section, Select, SelectValue } from 'react-aria-components';
 import { useParams, useRouteLoaderData } from 'react-router-dom';
 
 import {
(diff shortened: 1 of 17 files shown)


packages/insomnia/src/ui/components/dropdowns/content-type-dropdown.tsx
```diff
@@ -54,152 +131,93 @@ export const ContentTypeDropdown: FC = () => {
...
   const hasParams = body && 'params' in body && body.params;
   const numBodyParams = hasParams ? body.params?.filter(({ disabled }) => !disabled).length : 0;
 
-  const getIcon = (mimeType: string | null) => {
-    const contentType = activeRequest?.body && 'mimeType' in activeRequest.body ? activeRequest.body.mimeType : null;
-    const contentTypeFallback = typeof contentType === 'string' ? contentType : EMPTY_MIME_TYPE;
-
-    return mimeType === contentTypeFallback ? 'check' : 'empty';
-  };
+  console.log('bodytype', body.mimeType);
```
CurryYangxx: Does this need to be deleted？

packages/insomnia/src/ui/components/editors/request-script-editor.tsx
```diff
@@ -139,6 +140,207 @@ function getRequestScriptSnippets(insomniaObject: InsomniaObject, path: string):
...
+    },
+    {
+      'id': 'get-cookies',
+      'name': 'Get cookies',
+      'snippet': getCookies,
+    },
+  ],
+};
+
+const miscMenu: SnippetMenuItem = {
```
CurryYangxx, line 320: Could we move the config or bojects outside of the components to a external file, may be a util dir for UI? So that the tsx file will smaller and easy to read.

packages/insomnia/src/ui/components/dropdowns/content-type-dropdown.tsx
gatzjames: Yep no need to keep this as it is handled by the check icon in the ListBoxItem

packages/insomnia/src/ui/components/editors/request-script-editor.tsx
gatzjames: For me it's less readable tbh.
Moving something to another file:
- Extra hop to figure out what the data/component/styles look like and how it works
- You can delete this file and forget the others since there's no lint/ts errors
- If it becomes an export it can be used by other modules and now you accidentally created an abstraction that has to evolve to cater to other module's use cases
I'd suggest to avoid moving things to other files unless we know it's an abstraction we want to share between modules and has clear intent

packages/insomnia/src/ui/components/dropdowns/content-type-dropdown.tsx
CurryYangxx: I mean do we need to remove this `console.log`? Is it just use for debug in dev mode or we need it in prod?

gatzjames: 😄 sorry thought it was about the deleted code above

CurryYangxx (approved):

PR #7492, opened 2024-06-05
perf: App start improvement [INS-3957]

CurryYangxx:
close INS-4026 INS-3957
**objective: Reduce the App start duration**
#### Loading strategy
- [x] use the `findBestRoute` function to try to return the path with orgId and projectId in:
    - [x] initialEntry
    - [x] organizationId indexLoader
- [x] sync task: submit action in useEffect instead of syncing in index route loader (organization index & organizationId index)
#### return deferred value in project loader

> This solution is beneficial for both `app start` and `projects change`.

- [x] return deferred data in project loader and wait promise in component

### Test

- [x] user has no cache data
- [x] user has cache data

Other approaches we tried:
- [x] Use the router navigation events:
   - Added more complexity on the routing side and relies on private router APIs
   - Simplifies the `useEffect` logic
   - Adds another way to fetch and store data and caching still needs another solution
   - `router.subscribe` executes many times in one navigation, we need use a flag to control execution times
   - There's uncertainty about the execution timing (e.g. has the component rendered before we fetch the data or after?)
   - Having a status of which data has been fetched is still hard to achieve
- [x] Use a cache and move the fetching to loaders using defer as a mechanism to update
  - Better UX and DX
  - When using defer we have a mechanism to show stale data on the UI and update when fresh data come in.
  - The problem that remains is cache invalidation and fetching too many times since fetching now happens on the hot path of the app


CurryYangxx:
have some `console.log` for testing now, will be deleted before merge

packages/insomnia/src/ui/hooks/use-organization-async-task.ts
```diff
@@ -0,0 +1,58 @@
+import { useEffect, useMemo } from 'react';
+import { useFetcher, useLocation } from 'react-router-dom';
+
+import { findPersonalOrganization, Organization } from '../../models/organization';
+import { UserSession } from '../../models/user-session';
+import { AsyncTask } from '../routes/organization';
+
```
jackkav: you can use a keyed fetcher instead of this hook 
https://reactrouter.com/en/main/hooks/use-fetcher#key

in the docs it says if you need fetcher state from outside a component just use a matching key

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -164,52 +172,98 @@ async function syncOrganization(sessionId: string, accountId: string) {
...
+    return {};
+  } catch (error) {
+    console.log('Failed to run async task', error);
+    return {
+      error: error.message,
+    };
+  }
+};
+
+async function migrateProjectsUnderOrganization(personalOrganizationId: string, sessionId: string) {
```
jackkav, line 205: its really dangerous to mess with migrations, I'd scope this out if you can without introducing more risk

jackkav:
can you provide more evidence that this approach is addressing the lionshare of performance impact? from what i can see this PR is doing at least two things. in a perf PR please only focus on a single measurable improvement and give a rationale about why this perf change was chosen in context of other perf impacts which were also investigated.

eg. 
```
thing a 2%
thing b 10%
thing c 88%
i chose c because its was worth optimizing.
```

packages/insomnia/src/ui/routes/organization.tsx
CurryYangxx: The origin migration logic is in `/organizationId` index route loader, after the initial entry change, the user will not match this index route, so we need to put the migration in action.If we could add some test to reduce the risk?

CurryYangxx:
> can you provide more evidence that this approach is addressing the lionshare of performance impact? from what i can see this PR is doing at least two things. in a perf PR please only focus on a single measurable improvement and give a rationale about why this perf change was chosen in context of other perf impacts which were also investigated.
> 
> eg.
> 
> ```
> thing a 2%
> thing b 10%
> thing c 88%
> i chose c because its was worth optimizing.
> ```

Do you mean the two things is modity initial entry and add async task hooks? This is because if we only modify the initial entry, the logic in the `organization` and `organizationId` index route loader will not run, so I put the two changes together to make logic of the app be complete.

packages/insomnia/src/utils/router.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (!hasSeenOnboardingV9) {
+      return '/onboarding';
+    }
+
+    const hasUserLoggedInBefore = window.localStorage.getItem('hasUserLoggedInBefore');
+
+    const user = await models.userSession.getOrCreate();
+    if (user.id) {
+      // return '/organization';
+      const path = await getWholePath(user.accountId);
```
jackkav: interesting idea in the PR #1 nest the route history logic inside the initial entry in order for app start to go to the best possible route

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -164,52 +172,98 @@ async function syncOrganization(sessionId: string, accountId: string) {
...
+        taskPromiseList.push(migrateProjectsUnderOrganization(personalOrganizationId, sessionId));
+      }
+
+      if (task === AsyncTask.SyncProjects) {
+        invariant(organizationId, 'organizationId is required');
+        taskPromiseList.push(syncProjects(organizationId));
+      }
+    }
+
+    await Promise.all(taskPromiseList);
```
jackkav: Using signals with actions is a new concept. I still don't understand the benefit of this over creating 3 separate actions. This kind of out of the box thinking is cool but really hard to reason about. I don't want it in our happy path if I can help it.

CurryYangxx: One reason is 3 seperate action will case loader rerun 3 times.So I merge into only action

packages/insomnia/src/models/organization.ts
```diff
@@ -28,3 +28,12 @@ export const isOwnerOfOrganization = ({
   accountId: string;
 }) =>
   organization.metadata.ownerAccountId === accountId;
+
+export const findPersonalOrganization = (organizations: Organization[], accountId: string) => {
+  return organizations.filter(isPersonalOrganization)
+    .find(organization =>
+      isOwnerOfOrganization({
```
ihexxa, line 35: nit: Probably need double check if It is good enough to say that it is a personal org.

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -402,6 +425,55 @@ const OrganizationRoute = () => {
...
+
+  const location = useLocation();
+  const asyncTaskList = useMemo(() => {
+    return location?.state?.asyncTaskList || [] as AsyncTask[];
+  }, [location?.state?.asyncTaskList]);
+  const syncOrgsFetcher = useFetcher();
+  const syncProjectsFetcher = useFetcher();
+  const migrateProjectsFetcher = useFetcher();
+
+  useEffect(() => {
```
CurryYangxx, line 441: if we fetch the existing action instead of combine to a new action, there are 3 action submit, and the loader will re-run 3 times.This will result in repeated requests in the project loader several times.I'd suggest merging into a new action to reduce the number of requests.
image

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -402,6 +425,55 @@ const OrganizationRoute = () => {
...
+    if (asyncTaskList.includes(AsyncTask.MigrateProjects)) {
+      console.log('running migrate projects task');
+      const submit = migrateProjectsFetcher.submit;
+      submit({}, {
+        action: '/organization/migrate-projects',
+        method: 'POST',
+      });
+    }
+  }, [asyncTaskList, migrateProjectsFetcher.submit]);
+
```
CurryYangxx: @gatzjames @jackkav if we fetch the existing action instead of combine to a new action, there are 3 action submit, and the loader will re-run 3 times.This will result in repeated requests in the project loader several times.I'd suggest merging into a new action to reduce the number of requests.
image

packages/insomnia/src/utils/router.ts
```diff
@@ -0,0 +1,86 @@
...
+      const personalOrganization = findPersonalOrganization(organizations, user.accountId);
+      // If the personal org is not found in local storage go fetch from org index loader
+      if (!personalOrganization) {
+        return '/organization';
+      }
+      const personalOrganizationId = personalOrganization.id;
+      return {
+        pathname: await findBestRoute(personalOrganizationId),
+        state: {
+          // async task need to excute when fisrt entry
```
ihexxa: minor: typo `first`

packages/insomnia/src/ui/routes/organization.tsx
CurryYangxx: using defer also can not cancel the fetch request in loader

filfreire (commented):
Some tests on my machine:

Latest beta packaged version:
- 956 ms loading screen until we see dashboard
- 1633 ms activity until fully loaded (affected by network loading of collections and status)

This PR (using packaged app, from `npm run app-package`):
- 428 ms loading screen until we see dashboard
- 1500 ms activity until fully loaded (affected by network loading of collections and status)

Loading time difference is noticeable.

Switching between Organizations also is fast, but there's a split second when it still is not showing all the unsynced collections right after going to another project/organization.

CurryYangxx:
> Switching between Organizations also is fast, but there's a split second when it still is not showing all the unsynced collections right after going to another project/organization.

@filfreire Thank you very much for your test.Not showing all the unsynced files right away is as expected, because we return a deferred data in react-router loader and wait the promise in component render phase.

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -402,6 +425,55 @@ const OrganizationRoute = () => {
   };
   const [status, setStatus] = useState<'online' | 'offline'>('online');
 
+  // ideas:
+  // 1. useEffect with controlled execution
+  // 2. use fetcher history state to avoid running the same task multiple times
+  // 3. create an event after logged in user is detected in initial entry and listen.once in this component
```
jackkav:
```suggestion
  // ideas:
  // 1. useEffect with controlled execution
  // 2. use fetcher history state to avoid running the same task multiple times
  // 3. create an event after logged in user is detected in initial entry and listen.once in this component
  // 4. rather than use an action hack to avoid fetching in the loader, we can now use a deferred promise in the loader since they are now non-blocking.
```

jackkav: Try this:

// 4. rather than use an action hack to avoid fetching in the loader, we can now use a deferred promise in the loader since they are now non-blocking.

CurryYangxx: I think this idea is not feasible for two reasons
1. Because in current solution, we find best route in InitialEntry, and will skip the index route loader, so we are not able to keep fetching in the index route loader.
2. If we don't find best route and keep using '/organization' as the initial entry, the logic is executed to index route loader. But the index route loader just use for refresh cache data and return a redirect, we can't return a deferred promise because no component connected to the loader .

ihexxa: The async loading strategy of workspaces seems straightforward with low maintaining effort and which can work with organizations, projects loading in parallel. So I propose to move forward with this part at first.

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -400,6 +438,21 @@ const OrganizationRoute = () => {
     projectId?: string;
     workspaceId?: string;
   };
+  const syncOrgsAndProjects = useFetcher();
+
+  useEffect(() => {
+    console.log('run async task in useEffect');
```
gatzjames:
```suggestion
```

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -624,13 +581,21 @@ const ProjectRoute: FC = () => {
...
   const [isLearningFeatureDismissed, setIsLearningFeatureDismissed] = useLocalStorage('learning-feature-dismissed', '');
   const { organizationId, projectId } = useParams() as {
     organizationId: string;
     projectId: string;
   };
+  const [learningFeature] = useLoaderDeferData<LearningFeature>(learningFeaturePromise);
+  const [remoteFiles] = useLoaderDeferData<InsomniaFile[]>(remoteFilesPromise);
+
+  const finalFiles = useMemo(() => {
+    console.log(remoteFiles, 'remoteFiles');
```
gatzjames:
```suggestion
```

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -624,13 +581,21 @@ const ProjectRoute: FC = () => {
...
   } = useLoaderData() as ProjectLoaderData;
   const [isLearningFeatureDismissed, setIsLearningFeatureDismissed] = useLocalStorage('learning-feature-dismissed', '');
   const { organizationId, projectId } = useParams() as {
     organizationId: string;
     projectId: string;
   };
+  const [learningFeature] = useLoaderDeferData<LearningFeature>(learningFeaturePromise);
+  const [remoteFiles] = useLoaderDeferData<InsomniaFile[]>(remoteFilesPromise);
+
+  const finalFiles = useMemo(() => {
```
gatzjames:
```suggestion
  const allFiles = useMemo(() => {
```

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -668,7 +633,7 @@ const ProjectRoute: FC = () => {
   const isUserOwner = organization && userSession.accountId && isOwnerOfOrganization({ organization, accountId: userSession.accountId });
   const isPersonalOrg = organization && isPersonalOrganization(organization);
 
-  const filteredFiles = files
+  const filteredFiles = finalFiles
```
gatzjames:
```suggestion
  const filteredFiles = allFiles
```

packages/insomnia/src/utils/router.ts
```diff
@@ -0,0 +1,86 @@
...
+import * as models from '../models';
+import { Organization } from '../models/organization';
+import { findPersonalOrganization } from '../models/organization';
+import { Project } from '../models/project';
+export const enum AsyncTask {
+  SyncOrganization,
+  MigrateProjects,
+  SyncProjects,
+}
+export const findBestRoute = async (orgId: string) => {
```
gatzjames:
```suggestion
export const getInitialRouteForOrganization = async (orgId: string) => {
```

packages/insomnia/src/utils/router.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (user.id) {
+      const organizations = JSON.parse(localStorage.getItem(`${user.accountId}:organizations`) || '[]') as Organization[];
+      const personalOrganization = findPersonalOrganization(organizations, user.accountId);
+      // If the personal org is not found in local storage go fetch from org index loader
+      if (!personalOrganization) {
+        return '/organization';
+      }
+      const personalOrganizationId = personalOrganization.id;
+      return {
+        pathname: await findBestRoute(personalOrganizationId),
```
gatzjames:
```suggestion
        pathname: await getInitialRouteForOrganization(personalOrganizationId),
```

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -72,6 +72,7 @@ import { showModal } from '../../ui/components/modals';
 import { AskModal } from '../../ui/components/modals/ask-modal';
 import { insomniaFetch } from '../../ui/insomniaFetch';
 import { invariant } from '../../utils/invariant';
+import { findBestRoute } from '../../utils/router';
```
gatzjames:
```suggestion
import { getInitialRouteForOrganization } from '../../utils/router';
```

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -227,60 +237,13 @@ export const indexLoader: LoaderFunction = async ({ params }) => {
...
-  // Check if the org has any projects and redirect to the first one
-  const projectId = allOrganizationProjects[0]?._id;
-
-  if (!projectId) {
-    return redirect(`/organization/${organizationId}/project`);
-  }
-  invariant(projectId, 'No projects found for this organization.');
-
-  return redirect(`/organization/${organizationId}/project/${projectId}`);
+  const bestRoute = await findBestRoute(organizationId);
```
gatzjames:
```suggestion
  const initialOrganizationRoute = await getInitialRouteForOrganization(organizationId);
```

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -227,60 +237,13 @@ export const indexLoader: LoaderFunction = async ({ params }) => {
...
+  return redirect(bestRoute);
```
gatzjames:
```suggestion
  return redirect(initialOrganizationRoute);
```

gatzjames (commented):
Great investigation on the pros/cons of different approaches to improve data fetching.
Let's clean up the code and get this merged!

packages/insomnia/src/utils/router.ts
```diff
@@ -0,0 +1,86 @@
...
+import { database } from '../common/database';
+import * as models from '../models';
+import { Organization } from '../models/organization';
+import { findPersonalOrganization } from '../models/organization';
+import { Project } from '../models/project';
+export const enum AsyncTask {
+  SyncOrganization,
+  MigrateProjects,
+  SyncProjects,
+}
```
jackkav, line 12: consider writing a unit test together for this? to help search for simplifications

jackkav (dismissed):

gatzjames (dismissed):

packages/insomnia/src/utils/router.ts
CurryYangxx: Good advice. Added to my todo list

ihexxa (dismissed):

ihexxa (approved):

PR #7525, opened 2024-06-11
fix: add certificateSubjectName [INS-3983]

filfreire:
Closes INS-3983

Based on https://stackoverflow.com/a/72267165 and https://www.electron.build/configuration/win#WindowsConfiguration-certificateSubjectName this change could do the trick to make it so it's not just the installer that is code signed?

Update: converting back to draft, this doesn't work the way we think it does, and release-recurring fails trying to find a certificate locally.


subnetmarco:
@filfreire What's the status here?

PR #7531, opened 2024-06-12
fix: copy password

### LINKED ISSUE #5811: Can't copy value of the password field in the basic auth section
### Expected Behavior

Select the text, press ctrl+c or right click and copy and it should be added to my clipboard.

### Actual Behavior

I can right click and copy, and also ctrl+c, but the value is not added to my clipboard.

### Reproduction Steps

1. Create a request with basic auth
2. Add any text to the password field
3. Try to copy the value

### Additional Information

I understand this might be a "Security" feature, if we were talking about a non dev tool. Insomnia is used to test API's, i should be able to copy values freely.

I'm testing an API, the password has 128 chars, the only way i can retrieve the password now is typing char by char, since i cannot copy it from the input.

### Insomnia Version

2022.7.5

Windows

### Operating System Version

Windows 11

### Installation method

From insomnia.rest

_No response_

arbezerra:
Let the user copy the password if the input is not masked.
Closes #5811


gatzjames (changes requested):
Thanks for the PR!
I think we should remove the `password` type from the input instead to avoid all these issues. It's already complicated as is and will make things even more complex.

Let me know if you'd like to work on that otherwise we'll update it

arbezerra:
Do you mean remove just the copying block? Or remove the password type, showPasswords config and the props canBeMasked/isMasked?

gatzjames:
> Do you mean remove just the copying block? Or remove the password type, showPasswords config and the props canBeMasked/isMasked?

Yes remove the password type and the masked props. They create far more problems than what they solve atm

PR #7540, opened 2024-06-13
fix: Maximum update depth exceeded error in key-value editor

CurryYangxx:
The generateId cause the component Infinite loop render


gatzjames:
Closing in favor of https://github.com/Kong/insomnia/pull/7541

PR #7546, opened 2024-06-13
spike: proxy error on login [INS-4009]

filfreire:
Closes, barely, INS-4009. _hic sunt dracones_

Some users will see the "Failed to Fetch" due to having a Proxy configured, either on their machines or on Insomnia Settings, which doesn't seem to work well with InsomniaFetch during login or the whoami call. One extra hypothesis, untested, is that some users might be facing problems due to CORS, not sure yet

I've tried capturing exception similar to `ERR_PROXY_CONNECTION_FAILED` but it gets swallowed by sentry instrumentation on `fetch()` function, regardless of what failure happens of fetch, it's swallowed with a `TypeError: Failed to fetch`.

By the time we try to capture a http status code, we always get the above exception.

image

Tried a handful of things to circunvent this but nothing seemed to work, e.g.

- using [beforeSend](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using--1) on sentry on sentry config
- using [ignoreErrors](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-) on sentry config
- filtering out Http from integrations on sentry config

Other related findings:
- https://github.com/getsentry/sentry-javascript/issues/6376
- https://forum.sentry.io/t/disabling-automatic-error-capturing-in-javascript-sdk/5039/2
- https://forum.sentry.io/t/typeerror-failed-to-fetch-reported-over-and-overe/8447

## Useful steps to repro/test

- Set Proxy for http and https on insomnia settings with a value like `localhost:8005` that will force the exception
- If you want to avoid having to re-login on every code-edit, you can use the electron console instead, manually try to send `fetch()` by hand to whoami endpoint or a url that doesn't exist.
```
await fetch('https://api.insomnia.rest/whoami').catch(error => {console.log(error)})
```

image

### DIFF
--- a/packages/insomnia/src/ui/routes/auth.authorize.tsx
+++ b/packages/insomnia/src/ui/routes/auth.authorize.tsx
@@ -6,6 +6,7 @@ import { invariant } from '../../utils/invariant';
 import { SegmentEvent } from '../analytics';
 import { getLoginUrl, submitAuthCode } from '../auth-session-provider';
 import { Icon } from '../components/icon';
+import { showSettingsModal } from '../components/modals/settings-modal';
 import { Button } from '../components/themed-button';
 
 export const action: ActionFunction = async ({
@@ -16,7 +17,7 @@ export const action: ActionFunction = async ({
   invariant(typeof data?.code === 'string', 'Expected code to be a string');
   const error = await submitAuthCode(data.code);
   if (error) {
-    const humanReadableError = error?.message === 'Failed to fetch' ? 'Network failed, please try again. Note: if this happens many times, check if you have any Proxy configuration enabled on Insomnia settings and disable it.' : error?.message;
+    const humanReadableError = error?.message === 'Failed to fetch' ? 'Network failed, please try again. If this happens repeatedly, check if you have any Proxy enabled on Insomnia Preferences or on your device and try to disable it.' : error?.message;
     return {
       errors: {
         message: humanReadableError,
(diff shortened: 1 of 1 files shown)


jackkav:
Is the preferences panel visible during login in order for a user to respond to your message? And is insomnia proxy config the cause of this error? Could it also be system proxy?

filfreire:
> Is the preferences panel visible during login in order for a user to respond to your message? And is insomnia proxy config the cause of this error? Could it also be system proxy?

Preferences is visible by doing the keyboard shortcut - not good UX, perhaps a "link" that would open preferences would also help.

It can be either Insomnia proxy config and system proxy.

jackkav:
Perhaps indicate the hot key and generalise the wording to compensate

filfreire:
@jackkav added preferences button and changed the error text a bit:

image

packages/insomnia/src/ui/routes/auth.authorize.tsx
```diff
@@ -16,7 +17,7 @@ export const action: ActionFunction = async ({
   invariant(typeof data?.code === 'string', 'Expected code to be a string');
   const error = await submitAuthCode(data.code);
   if (error) {
-    const humanReadableError = error?.message === 'TypeError: Failed to fetch' ? 'Network failed, try again.' : error?.message;
+    const humanReadableError = error?.message === 'Failed to fetch' ? 'Network failed, please try again. If this happens repeatedly, check if you have any Proxy enabled on Insomnia Preferences or on your device and try to disable it.' : error?.message;
```
jackkav: We might not need both a button and a generalised explanantion. How about this?
```suggestion
    const humanReadableError = error?.message === 'Failed to fetch' ? 'Network failed, please try again. If the problem persists, check you rnetwork and proxy settings.' : error?.message;
```

jackkav (approved):

PR #7574, opened 2024-06-21
feat(Environment Picker): Simplify the UI

gatzjames:
Highlights:

- [x] Collection and Global environments are on the same column
- [x] There is a new column that displays the active environments in the hierarchy they are applied

image


jackkav (dismissed):

jackkav (approved):

AlbinBackstom:
Is there any way to display the current set environment as it was before? Now it only displays the </> Manage Environments button and its impossible to know what environment you are using without opening the new menu. 
image

jonataswalker:
Please revert this not helpful change.

fadelchaidarf:
Now I have to be extra careful just in case I hit the production environment. Before, I know what environment I'm in just by a glance, the environment color was super nice too

JoakoV3:
Can we add this fix? It was super nice to see the color of the environment you are working on #7607

AlbinBackstom:
Thanks for fixing it!

PR #7580, opened 2024-06-24
fix(KeyValue Editor): Pressing enter on an input should update the value

gatzjames:
Fixes an issue where pressing Enter would use the previous value instead of the current


pavkout (approved):
LGTM

PR #7582, opened 2024-06-25
perf: change org performance improvement [INS-3968]

CurryYangxx:
**Changes**

- [x] generate whole route(with orgId and projectId) when click organization icon
- [x] pass the sync project task into location state, and execute sync project through action
- [x] when switch to a new organization first time, we can't get projectId from db, so we need redirect route with projectId in `syncOrgsAndProjectsAction` after sync project.

**Spike**
May be we can try to sync project in advance when app is idle

### DIFF
--- a/packages/insomnia/src/ui/routes/organization.tsx
+++ b/packages/insomnia/src/ui/routes/organization.tsx
@@ -692,6 +692,7 @@ const OrganizationRoute = () => {
                           navigate(routeForOrganization, {
                             state: {
                               asyncTaskList: [
+                                // we only need sync projects when user switch to another organization
                                 AsyncTask.SyncProjects,
                               ],
                             },
--- a/packages/insomnia/src/ui/routes/organization.tsx
+++ b/packages/insomnia/src/ui/routes/organization.tsx
@@ -32,7 +32,7 @@ import { database } from '../../common/database';
 import { userSession } from '../../models';
 import { updateLocalProjectToRemote } from '../../models/helpers/project';
 import { findPersonalOrganization, isOwnerOfOrganization, isPersonalOrganization, isScratchpadOrganizationId, type Organization } from '../../models/organization';
-import type { Project } from '../../models/project';
+import { type Project, type as ProjectType } from '../../models/project';
 import { isDesign, isScratchpad } from '../../models/workspace';
 import { VCSInstance } from '../../sync/vcs/insomnia-sync';
 import { migrateProjectsIntoOrganization, shouldMigrateProjectUnderOrganization } from '../../sync/vcs/migrate-projects-into-organization';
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -659,40 +677,51 @@ const OrganizationRoute = () => {
...
+                    <Link className="outline-none">
+                      <div
+                        className={`select-none text-[--color-font-surprise] hover:no-underline transition-all duration-150 bg-gradient-to-br box-border from-[#4000BF] to-[#154B62] font-bold outline-[3px] rounded-md w-[28px] h-[28px] flex items-center justify-center active:outline overflow-hidden outline-offset-[3px] outline ${isActive
                           ? 'outline-[--color-font]'
                           : 'outline-transparent focus:outline-[--hl-md] hover:outline-[--hl-md]'
-                        } ${isPending ? 'animate-pulse' : ''}`
-                      }
-                      to={`/organization/${organization.id}`}
+                          }`}
+                        onClick={async () => {
```
ihexxa, line 699: As we should calculate route and async tasks in switching org, do we need to support `CTRL+P`?

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -166,30 +168,37 @@ async function syncOrganizations(sessionId: string, accountId: string) {
...
-    taskPromiseList.push(migrateProjectsUnderOrganization(personalOrganization.id, sessionId));
-    invariant(organizationId, 'organizationId is required');
-    taskPromiseList.push(syncProjects(organizationId));
+    if (asyncTaskList.includes(AsyncTask.SyncOrganization)) {
+      invariant(sessionId, 'sessionId is required');
+      invariant(accountId, 'accountId is required');
+      taskPromiseList.push(syncOrganizations(sessionId, accountId));
+    }
+
+    if (asyncTaskList.includes(AsyncTask.MigrateProjects)) {
```
ihexxa, line 189: Just would like to double check, should we always check and migrate projects before `SyncProjects`?

ihexxa (commented):
Added several minor comments.

packages/insomnia/src/ui/routes/organization.tsx
CurryYangxx: migrate projects only need to be execute in initial entry

packages/insomnia/src/ui/routes/organization.tsx
CurryYangxx: It seems `CTRL+P` already get the whole route. So we dont't need do it here.

filfreire (commented):
@CurryYangxx initial tests I did show a 100 to 150 ms improvement from the click event until we draw another organization.

There's a couple of points I think could help the performance improvement:

- Whenever we switch between Orgs, we have some duplicate API requests being done:
image

- Reclicking on the exact same organization sends some requests which might not be necessary:
image

CurryYangxx:
> @CurryYangxx initial tests I did show a 100 to 150 ms improvement from the click event until we draw another organization.
> 
> There's a couple of points I think could help the performance improvement:
> 
> * Whenever we switch between Orgs, we have some duplicate API requests being done:
> 

@filfreire Thanks for your test. Duplicate API requests are due to the remix's revalidate mechanism. I will start working on reducing the repeat requests next step.

ihexxa (approved):

PR #7635, opened 2024-07-02
perf: return deferred data in permission loader

CurryYangxx:
> Any [fetcher.load](https://reactrouter.com/en/main/hooks/use-fetcher#fetcherloadhref-options) calls that are active on the page will be re-executed as part of revalidation
We use `fetcher.load` to load `organizationPermissionsLoader`, and it will block route navigation (such as switch between organization).

So any fetcher.load will block route navigation such as switch between orgs or projects

**Changes**

- [x] return deferred data in `organizationPermissionsLoader`
- [x] use `useLoaderDeferData` hook to deal with the deferred data in every place we fetch this loader

### DIFF
--- a/packages/insomnia/src/ui/routes/organization.tsx
+++ b/packages/insomnia/src/ui/routes/organization.tsx
@@ -312,9 +312,9 @@ export interface StorageRule {
 }
 
 export interface OrganizationFeatureLoaderData {
-  features: FeatureList;
-  billing: Billing;
-  storage: 'cloud_plus_local' | 'cloud_only' | 'local_only';
+  featuresPromise: Promise<FeatureList>;
+  billingPromise: Promise<Billing>;
+  storagePromise: Promise<'cloud_plus_local' | 'cloud_only' | 'local_only'>;
 }
 
 export const organizationPermissionsLoader: LoaderFunction = async ({ params }): Promise<OrganizationFeatureLoaderData> => {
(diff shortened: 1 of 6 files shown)


packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -348,28 +348,27 @@ export const organizationPermissionsLoader: LoaderFunction = async ({ params }):
   }
 
   try {
-    const response = await insomniaFetch<{ features: FeatureList; billing: Billing } | undefined>({
+    const featuresResponse = insomniaFetch<{ features: FeatureList; billing: Billing } | undefined>({
```
ihexxa, line 351: nit: Just minor comment, personally I prefer keeping `await` and also use `Promise.resolve` for 363-365.

ihexxa (dismissed):

gatzjames (commented):
Should we wait for the new API that combines these things instead of merging this?

CurryYangxx:
> Should we wait for the new API that combines these things instead of merging this?

Thanks for reminding. Let me first see if the new API combines is a trivial thing first.

notjaywu:
> Should we wait for the new API that combines these things instead of merging this?

@CurryYangxx @gatzjames I'd suggest we merge this PR, release a beta version, and create another PR to integrate the merged endpoint. What do you think?

CurryYangxx:
> > Should we wait for the new API that combines these things instead of merging this?
> 
> @CurryYangxx @gatzjames I'd suggest we merge this PR, release a beta version, and create another PR to integrate the merged endpoint. What do you think?

@notjaywu I think that is ok. This pr is already tested and it works for performance.

ihexxa (approved):

PR #7638, opened 2024-07-02
test: clean up sdk cookie and add transforming tests

ihexxa:
Changes:
- add a test for transforming cookies between renderer and sdk
- keep specific cookie properties from renderer
- change `maxAge` into number


ihexxa:
This part has been merged.

PR #7661, opened 2024-07-04
fix: Use SSE for storage control updates

pavkout:
## Fixes Implemented:
- [x] Implemented immediate display of error messages when admins change storage rules for the cloud application.
- [x] Added listener for the new StorageRuleChanged message.
- [x] Display the server error messages for PROJECT_STORAGE_RESTRICTION errors.
- [x] Addressed the issue where attempting to create a new cloud project as a member displayed an error modal, which reappeared when trying to create a local vault project subsequently.
- [x] Always show a warning message when creating or updating a project if the storage rule is not cloud_plus_local.

## Notes for Team Discussion:
1. The fourth fix on the list is a quick solution. If preferred, I can revert this change and address it more efficiently in a separate PR.
2. For handling storage rules, I utilized network cache instead of localStorage. If there are concerns about this approach, let's use localStorage or discuss about alternative solutions.

Closes INS-4113
Related BE PR INS-4108

### DIFF
--- a/packages/insomnia/src/ui/routes/organization.tsx
+++ b/packages/insomnia/src/ui/routes/organization.tsx
@@ -256,10 +256,18 @@ async function syncStorageRule(sessionId: string, organizationId: string) {
       }),
     ]);
 
-    console.log('[storageRule] Storage rule', { storageRule });
     invariant(storageRule, 'Failed to load storageRule');
 
-    localStorage.setItem(`${organizationId}:storage-rule`, JSON.stringify(storageRule));
+    const cacheName = 'organization-storage-rules';
+    const cacheKey = `${organizationId}:storage-rule`;
+
+    // Open cache
+    const cache = await caches.open(cacheName);
+
+    // Cache the fetched response
+    const response = new Response(JSON.stringify(storageRule));
+    await cache.put(cacheKey, response.clone());
+
   } catch (error) {
     console.log('[storageRule] Failed to load storage rules', error);
   }
@@ -303,7 +311,7 @@ export const syncOrganizationsAction: ActionFunction = async () => {
 
 export const syncOrganizationStorageRuleAction: ActionFunction = async ({ params }) => {
   const { organizationId } = params;
-  console.log('[storageRule] Syncing storage rule for organization', { organizationId });
+
   invariant(organizationId, 'Organization ID is required');
 
   const { id: sessionId } = await userSession.getOrCreate();
@@ -375,13 +383,21 @@ export const organizationStorageLoader: LoaderFunction = async ({ params }): Pro
   const { organizationId } = params as { organizationId: string };
   const { id: sessionId } = await userSession.getOrCreate();
 
-  // Load from cache
-  const { storage } = JSON.parse(localStorage.getItem(`${organizationId}:storage-rule`) || '[]') as StorageRule || {};
-  console.log('from cache storage', { storage, organizationId });
-  // If we have a cached value return it
-  if (storage) {
+  // Create a cache key and name
+  const cacheName = 'organization-storage-rules';
+  const cacheKey = `${organizationId}:storage-rule`;
+
+  // Open cache
+  const cache = await caches.open(cacheName);
+
+  // Try to load from cache
+  const cachedResponse = await cache.match(cacheKey);
+
+  if (cachedResponse) {
+    const storageRule = await cachedResponse.json();
+
     return {
-      storage,
+      storage: storageRule.storage,
     };
   }
 
@@ -395,8 +411,9 @@ export const organizationStorageLoader: LoaderFunction = async ({ params }): Pro
 
     invariant(storageRule, 'Failed to load storageRule');
 
-    // Cache the value
-    localStorage.setItem(`${organizationId}:storage-rule`, JSON.stringify(storageRule));
+    // Cache the fetched response
+    const response = new Response(JSON.stringify(storageRule));
+    await cache.put(cacheKey, response.clone());
 
     // Return the value
     return {
--- a/packages/insomnia/src/ui/routes/organization.tsx
+++ b/packages/insomnia/src/ui/routes/organization.tsx
@@ -391,15 +391,14 @@ export const organizationStorageLoader: LoaderFunction = async ({ params }): Pro
       sessionId,
     });
 
-    const storageRule = await storageRuleResponse.then(res => res);
-
-    invariant(storageRule, 'Failed to load storageRule');
-
-    inMemoryStorageRuleCache.set(organizationId, storageRule);
-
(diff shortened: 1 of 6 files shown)


packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -268,17 +309,68 @@ export interface Billing {
...
+    return {
+      storage: storageRule.storage,
+    };
+  }
+
+  // Otherwise fetch from the API
+  try {
+    const storageRule = await insomniaFetch<StorageRule | undefined>({
+      method: 'GET',
+      path: `/v1/organizations/${organizationId}/storage-rule`,
```
CurryYangxx, line 390: we already have a [loader](https://github.com/Kong/insomnia/blob/2ccc2326ab34a6e7eef20e0e6d58039426f35005/packages/insomnia/src/ui/routes/organization.tsx#L282) to fetch /storage-rule. And will run fetch.load('/permission') in Project route. So is this `organizationStorageLoader` unnecessary

pavkout: @CurryYangxx That's why I have removed the `/storage-rule` from this loader and keep it only in one place.

filfreire:
@pavkout we rebased your branch with latest develop to help test for one of @marckong 's PRs in the API

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -164,6 +164,33 @@ async function syncOrganization(sessionId: string, accountId: string) {
   }
 }
 
+async function syncStorageRule(sessionId: string, organizationId: string) {
+  try {
+    const [storageRule] = await Promise.all([
```
CurryYangxx: Just only one request, seems like the Promise all is unnecessary.

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -164,6 +164,33 @@ async function syncOrganization(sessionId: string, accountId: string) {
...
+      }),
+    ]);
+
+    invariant(storageRule, 'Failed to load storageRule');
+
+    const cacheName = 'organization-storage-rules';
+    const cacheKey = `${organizationId}:storage-rule`;
+
+    // Open cache
+    const cache = await caches.open(cacheName);
```
CurryYangxx: Could you explain the reason for using `cache` API? I see it is mainly used in service workers.

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -839,9 +852,15 @@ const ProjectRoute: FC = () => {
     key: `${organizationId}-create-new-project`,
   });
 
+  const [createNewProjectFetcherData, setCreateNewProjectFetcherData] = useState(createNewProjectFetcher.data);
+
+  useEffect(() => {
+    setCreateNewProjectFetcherData(createNewProjectFetcher.data);
+  }, [createNewProjectFetcher.data]);
```
CurryYangxx: Why we can't use `createNewProjectFetcher.data` directly

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -268,17 +309,68 @@ export interface Billing {
...
+  const cacheName = 'organization-storage-rules';
+  const cacheKey = `${organizationId}:storage-rule`;
+
+  // Open cache
+  const cache = await caches.open(cacheName);
+
+  // Try to load from cache
+  const cachedResponse = await cache.match(cacheKey);
+
+  if (cachedResponse && Object.keys(cachedResponse).length > 0) {
```
CurryYangxx: there is a case: If the storage data change when the App is close.Next time when user open App, the cache will never update.

packages/insomnia/src/ui/routes/organization.tsx
pavkout: @CurryYangxx You are right, I forgot to remove it.

packages/insomnia/src/ui/routes/organization.tsx
CurryYangxx: Can we keep fetching /storage-rule in the `organizationPermissionsLoader`? Seems it is not necessary to create a new seperate loader.

pavkout: @CurryYangxx Is the storage rule is actually permissions?

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -611,23 +614,39 @@ const ProjectRoute: FC = () => {
...
 
+  useEffect(() => {
+    const isIdleAndUninitialized = storageRuleFetcher.state === 'idle' && !storageRuleFetcher.data && !isScratchpadOrganizationId(organizationId);
+
+    if (isIdleAndUninitialized || lastLoadedOrganizationId !== organizationId) {
+      storageRuleFetcher.load(`/organization/${organizationId}/storage-rule`);
+      // Update the state tracking the last loaded organization ID
+      setLastLoadedOrganizationId(organizationId);
+    }
+  }, [lastLoadedOrganizationId, organizationId, storageRuleFetcher]);
```
CurryYangxx:
```suggestion
  useEffect(() => {
    if (!isScratchpadOrganizationId(organizationId)) {
      const load = storageRuleFetcher.load;
      load(`/organization/${organizationId}/storage-rule`);
    }
  }, [organizationId, storageRuleFetcher.load]);
```
In this way, we don't need lastLoadedOrganizationId anymore

packages/insomnia/src/ui/routes/project.tsx
```diff
@@ -1541,6 +1563,188 @@ const ProjectRoute: FC = () => {
             </Dialog>
           </Modal>
         </ModalOverlay>
+        <ModalOverlay
```
CurryYangxx, line 1574: It seems we have another similiar modal, if we colud use it? packages/insomnia/src/ui/components/dropdowns/project-dropdown.tsx

packages/insomnia/src/ui/routes/project.tsx
pavkout: @CurryYangxx I test it locally and confirm that it works

packages/insomnia/src/ui/routes/project.tsx
pavkout: @CurryYangxx After a discussion with @gatzjames we said let's keep both modals

pavkout:
@CurryYangxx Just finished the changes we discussed. Feel free to test it and review them one more time, and let's have a try it session one more time tomorrow morning.

CurryYangxx (dismissed):

CurryYangxx (approved):

PR #7669, opened 2024-07-05
feat(sdk): log request testing results and return them to the main renderer

ihexxa:
### Changes
- [x] in after-resp script execution, collect test results to the `requestTestResults` and return them to the main renderer
- [x] merge it to `feat/test-result-pane` temporarily
- [x] disabled `test` related tests
- [x] enabled `insomnia.test.skip`

Ref: INS-4162

### DIFF
--- a/packages/insomnia-sdk/src/objects/index.ts
+++ b/packages/insomnia-sdk/src/objects/index.ts
@@ -8,3 +8,4 @@ export * from './cookies';
 export * from './console';
 export * from './request-info';
 export * from './async_objects';
+export * from './test';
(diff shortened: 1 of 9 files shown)


filfreire (approved):
@ihexxa can you leave some notes but approving since this is for a feature branch;

- I don't see the logs  related to the test or the check/assertion anymore, not sure if this intended, unless i add a console.log
- the skip feature seems to work ok.

image

ihexxa:
@filfreire yes, it is expected.

Originally test results are logged and persisted to disk through `log(xxx)`, while currently they are collected by `pushRequestTestResult` and returned as a part of context. But it seems a good idea to keep log in timeline at the same time.

PR #7678, opened 2024-07-08
chore: update-changelog job not working [no-ticket]

filfreire:
Related to https://github.com/Kong/insomnia/pull/7677

Not sure yet why the job wouldn't trigger. Could be a permissions issue.

In this one we're changing the event type to publish just to see if anything at all shows up.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release


filfreire:
closing for now, we can reopen later if needed

PR #7681, opened 2024-07-09
fix: ui improvement when return deferred data in loader

CurryYangxx:
After return deferred data in loader, there is an ui issue:
When we switch project from `A` to `B`, the stale remote project in A will stay for a short time before the remote data fetching finish in `B project`. So we need a key to help us to dertimine if we need to keep stale data.

There are 2 cases:
1. switch to another project: we shouldn't keep stale data
2. other action such as rename a project: the projectId not change, so we should keep the stale data.
 
**In project A**
image
**Click to switch another project B**
The stale data stay for a short time
image
**Project B fetching finish**
image

### DIFF
--- a/packages/insomnia/src/ui/hooks/use-loader-defer-data.ts
+++ b/packages/insomnia/src/ui/hooks/use-loader-defer-data.ts
@@ -14,8 +14,8 @@ export const useLoaderDeferData = <T>(deferedDataPromise?: Promise<T>, keepStale
     (async () => {
       try {
         // use keepStaleDataKey to let us know if we should keep the stale data or not
-        // sometimes we want to keep the stale data if we are fetching the same data again
-        // sometimes we want to clear the stale data if we are fetching different data
+        // sometimes we want to keep the stale data (rename a collection -> action submit -> run loader)
+        // sometimes we want to clear the stale data if we are fetching different data (switch to another project -> navigate -> run loader)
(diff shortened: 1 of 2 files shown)


ihexxa (dismissed):

filfreire (commented):
After some initial testing it seems it takes more or less the same time to switch to the view, so perhaps @CurryYangxx something else might be slowing down the switch in such a way that the "keepStaleData" trick does not seem to help a lot.

Both latest `develop` and this branch both seem to take 400 to 450 ms from the event click until we first draw the new organization - when we click on a different one.


image

filfreire (approved):
Approving in the meantime as I reckon the focus of this PR is not for speeding up, but for UI

CurryYangxx:
> Approving in the meantime as I reckon the focus of this PR is not for speeding up, but for UI

Yeah, this pr is for UI improvement

CurryYangxx:
> After some initial testing it seems it takes more or less the same time to switch to the view, so perhaps @CurryYangxx something else might be slowing down the switch in such a way that the "keepStaleData" trick does not seem to help a lot.
> 
> Both latest `develop` and this branch both seem to take 400 to 450 ms from the event click until we first draw the new organization - when we click on a different one.
> 

@filfreire  Yes, the `organizationPermissionsLoader` loader will slow down switching organization. There is another pr I am working on to fix it. https://github.com/Kong/insomnia/pull/7635.

PR #7738, opened 2024-07-18
fix: lost <disabled> in header transforming and blank req body

ihexxa:
### Background
When scripts is enabled, it failed to keep `disabled` property for request headers.
so add `disabled` in both initing insomnia object and generating the updated request.

### Changes
- [x] fix: lost <disabled> in header transforming
- [x] fix: lost blank req body

Ref: INS-4196


subnetmarco:
@ihexxa is this still a draft? The review process cannot start until this PR is a non-draft.

CurryYangxx (approved):

PR #7763, opened 2024-07-26
Fix backslash in environment key freeze app [INS-4157]

### LINKED ISSUE #7286: Double backslash at end of Environment JSON key freezes entire app, corrupts collection
### Expected Behavior

Double-backslash at end of environment JSON key should be valid

### Actual Behavior

Double-backslash at end of environment JSON key freezes entire app and corrupts collection

### Reproduction Steps

1. Create a new "Request collection". Name it anything
2. Click the cog icon near "Base environment"
3. Enter the following JSON:

```
{
  "What": "dafuq"
}
```

4. Click after the letter t in What to move the cursor there. Type in `-\\`, resulting in the following JSON:
```
{
  "What-\\": "dafuq"
}
```

Note that shortly after typing the two backslashes, the app will freeze

### Additional Information

Screencap: https://imgur.com/jD6c4S4

This corrupts the collection. Whenever the collection is loaded again, the app freezes. On macOS, the "unresponsive app" warning dialog shows.

Sometimes "\\" alone will not freeze the app. Sometimes there must be a \ somewhere else in the string, or a - immediately before it. I'm not sure what other characters or patterns cause this.

One way to fix it is to manually edit `insomnia.Environment.db` to remove the backslashes. Then the collection will load just fine

### Insomnia Version

8.6.1

macOS

### Operating System Version

macOS Sonoma 14.2.1 (23C71)

### Installation method

homebrew

_No response_

yaoweiprc:
Close #7286 
objectPath.normalize will enter endless loop if last char in key name is backslash, just remove backslashes at the end of key name.
It's just a workaround. It will introduce problems when reference an environment variable which's name ends with backslash.
image
image


packages/insomnia/src/templating/utils.ts
```diff
@@ -83,7 +83,9 @@ export function forceBracketNotation(prefix: string, key: string | number) {
 }
 
 export function normalizeToDotAndBracketNotation(prefix: string) {
-  return objectPath.normalize(prefix);
+  // objectPath.normalize will enter endless loop if last char in key name is backslash, just remove backslashes at the end of key name
+  // workaround for https://github.com/Kong/insomnia/issues/7286
+  return objectPath.normalize(prefix.replace(/\\+']/g, '\']'));
```
filfreire: (EDIT)

Maybe we prevent users from entering the backslash altogether

filfreire (dismissed):
I think this is OK as is.

There's more options:
-  we could prevent users inputting backslash
- or delete it from the keys when saving environment
-  or we display an error and don't save environment, like when we forget a , or a " on json, not allowing users to save environment until they delete \ present in keys
- or we keep your approach as is, and accept this as lesser evil - between crashing the app and leaving environment key dangling, this is good enough for now
- something else?

jackkav:
Could this be done without using a regex, would it be more readable?

yaoweiprc:
This problem is caused by objectpath library, in its parse function, it could enter endless loop when the last char in key is backslash. It seems to be no longer maintained.
The previous commit was just a workaround to delete trailing backslashes from key. 
In this commit I copied its source code to packages\insomnia\src\templating\third_party\objectPath.ts and fixed the bug within line 58 to 80.

jackkav (approved):
this looks like it would work, but a step deeper would be attempting to eliminate the library completely and just reimplement the two functions use cases with less code, and eliminate the while loop that causes the problem. Not required for me to approve this PR but worth considering if you like a challenge. 🎸 

EDIT: Actually there are two things that could push you a little to improve this PR
- write a couple unit tests for forceBracketNotation and normalizeToDotAndBracketNotation
- attempt to simplify the object path implementation in the pr

filfreire (approved):

yaoweiprc:
I need some time to familiarize myself with Vitest. I just merged this PR, I'll improve in other threads.

PR #7772, opened 2024-07-29
Measure back to dashboard duration from collection and report to sentry

CurryYangxx:
**Changes:**

- [x] Use User Timing API to measure the time it takes to return to the dashboard from collection
- [x] create a new sentry metric: back_to_dashboard


jackkav (commented):
How long are you planning to keep this in the code? 
What do you plan to do with the data?
Are any of the useEffects likely to over execute? 
At what interval are they intended to execute?

CurryYangxx:
> How long are you planning to keep this in the code? What do you plan to do with the data? 

These codes are used to measure operation time. The metric data will be reported to the sentry platform for performance analysis(Let us know if the performance gets better or worse).
If we don't find a better approach to do measurement, the code will always be kept.


> Are any of the useEffects likely to over execute? At what interval are they intended to execute?

The newly added useEffect will only be executed once when the component is mounted because the second dependencies params in useEffect is an empty array.

filfreire (approved):

PR #7794, opened 2024-08-01
fix(Sync Staging Modal): use action to update selected items

gatzjames:
Highlights:

- [x] Updates the lists on the sync modal to not use selection and use actions instead. This allows esc to propagate and doesn't require us to manage selection state


CurryYangxx (approved):

CurryYangxx:
When is it appropriate to use onSelectionChange and when to use onAction?

gatzjames:
> When is it appropriate to use onSelectionChange and when to use onAction?

Selection comes with many features like multiple selection, select-deselect (allowing empty selection or not), checkbox support etc. while onAction is a more simple api that only fires when an item is pressed/clicked.

Since we don't use any of the advanced features of selection it's simpler to use action here.
If we'd want to allow multiple selection to e.g. select many items and drag-n-drop them to another list we would need to use the selection api.

Also note that they can be used in combination where you can have an action on press and a selection using a checkbox (with different keyboard shortcuts too)

Issue #2111, opened 2020-05-04
Keep equal sign for empty query parameters


gbogard:
Hi :) 
I've been using Insomnia for over 2 years now, and it's such a great piece of software. However, I am currently facing an issue that stops me from using on a particular project.

I need to call a legacy web server that has a very *not standard* way of dealing with query parameters. There is only one endpoint, and all parameters of a request, including the nature of the request itself is passed via query parameters. 
Some endpoints require query parameters with no value. However the server doesn't recognize the parameters when they are sent without an equal sign.

**Example**

This works
```
/terminologies?query=getconcept&terminologyid=indexation&conceptId=
``` 

But this yields a 400 Bad Request

```
/terminologies?query=getconcept&terminologyid=indexation&conceptId
```

The issue is Insomnia strips the equal sign for empty query parameters.

It has been mentioned [here](https://github.com/Kong/insomnia/issues/264) by another user, who has since then been able to solve the problem by updating the server. But I have no control whatsoever on the server in that case.

I think the best solution would be to have a checkbox somewhere to control this behavior.
Alternatively, is there any workaround to prevent Insomnia from doing that? I've tried passing `=""` and `=''` but the server is very picky and won't accept that either.

gbogard:
Update: I have found a workaround for my particular use case. When putting the equal sign in the parameter name itself in the `Query` tab, the server picks it up as if it wasn't part of the parameter name, even though it is url-encoded.

However this only works because of the way this particular application handles query parameters. I believe the feature request is still relevant, although much less urgent :)

kreosus:
Hi @gbogard 
Do you see the same behavior with our latest release located here: https://github.com/Kong/insomnia/releases/tag/core%402022.6.0-beta.4

nbenninghoff:
> 

I just had this problem and can verify it is present in 2022.6.0-beta.4
Equals sign is getting stripped from empty parameters.  This causes a problem in ASP .NET sites as each parameter gets treated as a "keyless" parameter and each one gets concatenated to a query param with a null key.
https://stackoverflow.com/questions/3130541/access-query-string-parameters-with-no-values-in-asp-net

Naist4869:
ASP .NET +1

> 你好@gbogard 您是否在我们位于此处的最新版本中看到相同的行为：[https ://github.com/Kong/insomnia/releases/tag/core%402022.6.0-beta.4](https://github.com/Kong/insomnia/releases/tag/core%402022.6.0-beta.4)

[...]

PR #7802, opened 2024-08-05
Fix: Keep equal sign for empty query parameters[INS-4228]

cwangsmv:
Relates to https://github.com/Kong/insomnia/issues/2111

Add a **strictNullHandling** option for smartEncodeUrl. 
By default, query string with equal sign(&foo=) or without equal sign(&foo) will converted to empty string as search param value. 
When this option is enabled, with equal sign(&foo=) will parsed as empty string in search param value while without equal sign(&foo) will parsed as **null** in search param value. After all, we can distinguish the different between two situations.

Also update the querystring unit test.


jackkav (approved):
Very thorough work, great job @cwangsmv !

filfreire (approved):

PR #7807, opened 2024-08-05
Moved the search bar to the right of the body so that it doesn't hide the string in initial row

### LINKED ISSUE #7752: The search bar hides the searched string
### Expected Behavior

The search result is always visible

### Actual Behavior

The search bar hides the searched string

### Reproduction Steps

type somthing on the search bar 
launch search
tap enter to go to next occurrence

### Additional Information

Capture d’écran 2024-07-24 094918
Capture d’écran 2024-07-24 094918

### Insomnia Version

9.3.2

Windows

### Operating System Version

Windows 11 23h2

### Installation method

exe

_No response_

Karthik7406:
Moved the search bar to the right of the request/response body by 50% so that the search bar doesn't hide with the text when the search bar is opened.

Closes #7752 

Screenshot
Request body
image

Response body
image


subnetmarco:
I seem to understand that this is the search bar for the JSON request body. It seems counterintuitive to show it all the way to the right in the response section, as the user may mistake it for a response filter.

Ideally, when we load the search bar, the request body should go down instead of overlying the search bar on top of the body. This would allow us to fix the issue, while still retaining the search bar in the right place.

gatzjames:
Thanks for the PR!
Closing as we are looking into alternatives for our code-editor and this solution has some issues especially when the editor view is small.
We will keep this UX improvement in mind on our iteration for the updated code editor.

PR #7810, opened 2024-08-06
fix: README.md typo to clarify who owns the projects

timodwhit:


subnetmarco:
Thank you @timodwhit - this was coincidentally already fixed in a latest update in the `README.md` which also reworded the paragraph in question.

PR #7811, opened 2024-08-06
feat(sync): support offline commit- [INS-4226]

CurryYangxx:
Support offline commit for Insomnia cloud sync.

- [x] Add fallback data when getting remote sync data

### DIFF
--- a/packages/insomnia/src/ui/routes/remote-collections.tsx
+++ b/packages/insomnia/src/ui/routes/remote-collections.tsx
@@ -254,7 +254,6 @@ interface SyncData {
     ahead: number;
     behind: number;
   };
-  remoteBackendProjects: BackendProject[];
 }
 
 const remoteBranchesCache: Record<string, string[]> = {};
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/ui/routes/remote-collections.tsx
```diff
@@ -320,27 +319,31 @@ export const syncDataLoader: LoaderFunction = async ({
...
     const status = await vcs.status(syncItems);
-    const compare =
+
+    let remoteBranches: string[] = [];
+    let compare = { ahead: 0, behind: 0 };
+    try {
+      remoteBranches = (
+        remoteBranchesCache[workspaceId] || (await vcs.getRemoteBranchNames())
+      ).sort();
+      compare =
```
gatzjames, line 335: Would it make sense to split these in two try/catch?
My thought is: In case one of them fails and the other works we would at least have the correct value for one of them

ihexxa: Also looking at this part and i have another question, will splitting introduce inconsistency? this seems for avoiding noise when offline, it might be great if we could differentiate different errors.

ihexxa (approved):
Overall lgtm and error handling might worth a bit discussion.

packages/insomnia/src/ui/routes/remote-collections.tsx
CurryYangxx: Previously, we returned an error if any of them failed, and did not show the entrance for sync. Now, I have wrapped the logic that relies on the network in a new try/catch block so that we can get some of the sync data when the request fails or the user is offline. This allows users to use the sync features that do not depend on the network.

CurryYangxx: Do you have some other suggestions for error handling?

gatzjames: Not sure I have other thoughts on this. I feel much more confident baed on your comment so LGTM

gatzjames (approved):

PR #7821, opened 2024-08-08
Import postman env in Insomnia project level [INS-4253]

yaoweiprc:
Before:
When importing a Postman environment in the Insomnia project level we create a new empty workspace and the environment doesn't get imported.
When importing a Postman global environment it's not supported.

Now:
When importing a Postman environment in the Insomnia project level we create a new global environment in current project and copy the variables.
When importing a Postman global environment in the Insomnia project level we create a new global environment in current project and copy the variables.

### DIFF
--- a/packages/insomnia/src/utils/importers/importers/postman-env.ts
+++ b/packages/insomnia/src/utils/importers/importers/postman-env.ts
@@ -20,13 +20,20 @@ type Data = {
   [key in EnvVar['key']]: EnvVar['value'];
 };
 
+const POSTMAN_ENV_TYPE: Record<string, string> = {
+  'GLOBAL': 'globals',
+  'ENVIRONMENT': 'environment',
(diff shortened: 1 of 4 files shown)


packages/insomnia/src/utils/importers/importers/postman-env.ts
```diff
@@ -20,13 +20,20 @@ type Data = {
   [key in EnvVar['key']]: EnvVar['value'];
 };
 
+const POSTMAN_ENV_TYPE: Record<string, string> = {
```
CurryYangxx: may be we can try to use TS enum to define it

CurryYangxx (dismissed):

filfreire (dismissed):
tested locally, LGTM

CurryYangxx (approved):

PR #7831, opened 2024-08-13
Fix: GraphQL request export curl body issue and GraphQL payload delete issue[INS-4281]

cwangsmv:
Background
1. When user exports GraphQL request as curl, the Query Variables value is exported as entire string rather than string from JSON.stringify. So user could not use the exported curl script.
2. When user delete both Query & Query Variables in GraphQL editor and switch to JSON body, it will be confusing that the content is not cleared.

- [x] Add the GraphQL request body parse logic when export curl 
- [x] Clear GraphQL request body when user delete both Query & Query Variables 
- [x] Update smoke test for GraphQL request export


ihexxa (dismissed):
LGTM, probably we could disable the failed ut temporarily in rebasing.

ihexxa (approved):

PR #7834, opened 2024-08-14
import postman data dump [INS-3810]

yaoweiprc:
Now we can import postman exported data dump. We can import all collections and environments from Postman data dump archive file. Make sure each request is in correct collection.

* Collections from Postman data dump will be created as new collections in current project with a top level folder. Just like what we do when import an individual collection json file from Postman.
* Environments from Postman data dump will be created as new global environments in current project.
* Not include global environments from postman because it's not included in data dump.
* The data dump does not preserve which workspaces in Postman exist, so we just import all Postman collections and envs to current project in Insomnia.
* Import data dump in the Insomnia Collection level is not allowed, you can just import in project level.

Introduced a new npm package `decompress` for this feature. I found it rather complicated to unzip archive files by node.js built-in zlib module.


notjaywu (approved):
lgtm

PR #7844, opened 2024-08-16
fix: test-util snippet to proper status code check [no-ticket]

filfreire:
`response.code` returns status code number. Before the snippet we were suggesting was wrong, checking the status code "string" instead.


gatzjames (approved):

ihexxa (approved):
LGTM, I also fixed this on the runner side.

PR #7856, opened 2024-08-21
feat: Add logic to redirect users based on their plan when creating a new organization

pavkout:
as per title - Add logic to redirect users based on their plan when creating a new organization via Electron App UI

Closes INS-4314

### DIFF
--- a/packages/insomnia/src/ui/routes/organization.tsx
+++ b/packages/insomnia/src/ui/routes/organization.tsx
@@ -849,25 +849,23 @@ const OrganizationRoute = () => {
                           return;
                         }
 
-                        // If user has a team or enterprise member plan show them an alert
                         if (currentPlan.type === 'enterprise-member') {
+                          // If user has a team or enterprise member plan show them an alert
                           showAlert({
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -832,16 +832,36 @@ const OrganizationRoute = () => {
...
                         }
+
+                        // If user has a free or individual plan redirect them to the landing page
+                        if (['free', 'individual'].includes(currentPlan.type)) {
+                          window.main.openInBrowser(
+                            `${getAppWebsiteBaseURL()}/app/landing-page`,
+                          );
+                        }
+
+                        // If user has a team or enterprise plan redirect them to the create organization page
```
filfreire: We are still running this bit of code if the user is of type `enterprise-member`, while we also show the alert.

we should probably not open any browser window in this case?

filfreire (dismissed):
Tested locally, aside from comment above, LGTM

filfreire (approved):

PR #7859, opened 2024-08-22
Add info and execution to Insomnia SDK with collection runner[INS-4320]

cwangsmv:
Expose more features in SDK with collection runner: 

Task List:
- [x] Add info and execution to SDK property
- [x] Integrate with request and collection runner for skipRequest and setNextRequest method
- [x] Add unit test

### DIFF
--- a/packages/insomnia/src/network/network.ts
+++ b/packages/insomnia/src/network/network.ts
@@ -1,6 +1,6 @@
 import clone from 'clone';
 import fs from 'fs';
-import type { ExecutionOption, RequestTestResult } from 'insomnia-sdk';
+import type { ExecutionOption, RequestContext, RequestTestResult } from 'insomnia-sdk';
 import orderedJSON from 'json-order';
 import { join as pathJoin } from 'path';
 
@@ -154,8 +154,8 @@ export const tryToExecutePreRequestScript = async (
   }: Awaited<ReturnType<typeof fetchRequestData>>,
   workspaceId: string,
   userUploadEnv?: UserUploadEnvironment,
-  iterationCount?: number,
   iteration?: number,
+  iterationCount?: number,
 ) => {
   const baseEnvironment = await models.environment.getOrCreateForParentId(workspaceId);
   const cookieJar = await models.cookieJar.getOrCreateForParentId(workspaceId);
@@ -203,6 +203,7 @@ export const tryToExecutePreRequestScript = async (
     iteration,
     iterationCount,
     ancestors,
+    eventName: 'prerequest',
   });
   if (!mutatedContext?.request) {
     // exiy early if there was a problem with the pre-request script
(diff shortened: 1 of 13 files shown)


packages/insomnia/src/ui/routes/runner.tsx
```diff
@@ -707,8 +707,18 @@ export const runCollectionAction: ActionFunction = async ({ request, params }) =
   });
 
   for (let i = 0; i < iterations; i++) {
+    // nextRequestIdOrName is used to manual set next request in iteration from pre-request script
+    let nextRequestIdOrName = '';
     for (let j = 0; j < requests.length; j++) {
-      const targetRequest = requests[j];
+      const targetRequest = requests[j] as { name: string; id: string; url: string };
+      if (typeof nextRequestIdOrName === 'string' && nextRequestIdOrName !== '') {
+        if (targetRequest.id === nextRequestIdOrName || targetRequest.name.trim() === nextRequestIdOrName.trim()) {
```
ihexxa, line 714: This condition seems satisfied in the first match while we noticed that it tries to find the last match.

packages/insomnia/src/network/network.ts
```diff
@@ -157,10 +166,7 @@ export const tryToExecutePreRequestScript = async ({
     activeGlobalEnvironment = await models.environment.getById(workspaceMeta.activeGlobalEnvironmentId) || undefined;
   }
 
-  const requestGroups = await db.withAncestors<Request | RequestGroup>(request, [
-    models.requestGroup.type,
-  ]) as (Request | RequestGroup)[];
-
+  const requestGroups = ancestors.filter(doc => isRequest(doc) || isRequestGroup(doc)) as RequestGroup[];
```
ihexxa, line 170: If this implementation is incorrect, would you mind also modify the same part in the after-response function.

packages/insomnia/src/network/network.ts
```diff
@@ -307,12 +321,20 @@ export const tryToExecuteScript = async (context: RequestAndContextAndOptionalRe
         clientCertificates,
         settings,
         cookieJar,
+        requestInfo: {
+          eventName: 'prerequest',
```
ihexxa: Besides `prerequest`, it could also be `test` when it is called by `trytoExecuteAfterResponseScript`

ihexxa (commented):
Had a quick review and will dive deeper later.

packages/insomnia/src/network/network.ts
cwangsmv: Good catch, I will add the missing test eventName

packages/insomnia/src/network/network.ts
cwangsmv: Yes, I have also change the same logic in https://github.com/Kong/insomnia/pull/7859/files#diff-cc2479e11623891c8c7a329087ad50c1f140db4b1e753ea94af01e59af95484dR451.

This change is an enhancement to remove duplicated db call. For both tryToExecuteAfterResponseScript and tryToExecutePreRequestScript function, they will accept requestData variable from fetchRequestData. And I have changed fetchRequestData to fetch all ancestors for the request, so in tryToExecuteAfterResponseScript and  tryToExecutePreRequestScript, they can just use the ancestors property of requestData` variable and reduce db call.

packages/insomnia/src/ui/routes/runner.tsx
cwangsmv: Fixed

ihexxa (approved):

PR #7860, opened 2024-08-22
fix(Git Sync): fix issue when switching to Insomnia Sync

gatzjames:
Highlights:

- [x] Update the git sync dropdown to match insomnia sync
- [x] Fixes an issue where there was an error thrown when switching to insomnia sync


filfreire (approved):

filfreire (commented):
tested locally, LGTM

PR #7870, opened 2024-08-26
Custom Authorization header name option

tandraehub:
Currently when using the OAuth2 in insomnia it always puts the access token in a header named Authorization. Even though the header almost always is called Authorization it causes inconvenience when using Insomnia with an API that has a different name for the auth header. 

This PR adds a new configuration in the Advanced Options section of the Auth tab where it is possible to specify a custom name of the authorization header. As default the option is not set and the name Authorization is then used.

Image below show the UI and a green circle is added in the image to highlight the new option:

custom_auth


gatzjames:
Closing for now since we need to expose the authorization sessions in the future and will address this in the near future.
For now a workaround is to use request chaining to use any kind of custom header or property for the authorization:
You could use the following template tag:
```
{% request 'oauth2', '', '' %}
```

PR #7897, opened 2024-09-02
fix: file not synced after switch sync method -[INS-4347]

CurryYangxx:
**Issue detail**
Sometimes after switch sync method. The file has become not synced.

**Root cause**
When using `models.xx.update` function, sometimes will update model by stale data. After switch to Insomnia Sync, we will delete `gitRepositoryId` in workspaceMeta and delete the corresponding gitrepository. But the workspaceMeta will be overwritten by the stale data, so there is `gitRepositoryId` in meta but no 
gitrepository data.
image


CurryYangxx:
I think we need try to use `updateByParentId` rather than `update` in most case.

jackkav (approved):

PR #7916, opened 2024-09-05
fix: minor fixes in styles, linting and UT

ihexxa:
### Changes

- [x] Added max width and min width for the runner pane one
- [x] Fixed some lint errors in the develop branch
- [x] Fixed an UT failure in the develop branch


jackkav (approved):
Thanks for fixing these, was a mess from an old PR that wouldn't merge properly

PR #7928, opened 2024-09-09
Fix duplicate file cause application error in collection view UI

cwangsmv:
**Issue:**

https://github.com/user-attachments/assets/d085b731-59ab-4d85-b535-bcf6736c8ccb

**Cause**
Race condition. 
Current page loader function in router will trigger due to create action in database. 
Meanwhile, new page loader function in router will also be triggered. So when new open is rendering, the loaderData will be overridden by old page loader function which cause this issue.

**Change**
Use old way of redirect function in duplicateWorkspaceAction


CurryYangxx (dismissed):

CurryYangxx:
May be we need a smoke test for this duplicate entrance. @yaoweiprc

jackkav (dismissed):

filfreire (dismissed):

CurryYangxx (approved):

PR #7936, opened 2024-09-10
perf: test-suit loader improvement - [INS-4400]

CurryYangxx:
**Issue**
When the nesting levels are deeper or there are too much children under workspace. The `withDescendants` will take a long time to find all requests.

**Changes**

- [x] Add a `queryType` param for `withDescendants` function to limit the types of database queries
- [x] pass `stopType` and `queryType` when call the `withDescendants` function in test-suit loader

### DIFF
--- a/packages/insomnia/src/common/database.ts
+++ b/packages/insomnia/src/common/database.ts
@@ -609,8 +609,20 @@ export const database = {
     return next([doc]);
   },
 
-  /** get all descendants of a document */
-  withDescendants: async function<T extends BaseModel>(doc: T | null, stopType: string | null = null): Promise<BaseModel[]> {
+  /**
+   * Get all descendants of a document.
+   *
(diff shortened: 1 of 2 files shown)


ihexxa (approved):

jackkav:
An alternative implementation would look like replacing stopType with queryType as they have overlapping behaviour.

Note: there are only 3 usages of stopType in the code and they all stop on request type, you could just set request as the query type and remove that argument completely.

image

CurryYangxx:
> An alternative implementation would look like replacing stopType with queryType as they have overlapping behaviour.
> 
> Note: there are only 3 usages of stopType in the code and they all stop on request type, you could just set request as the query type and remove that argument completely.
> 
@jackkav The request may not be a direct child of workspace, but a child of requestGroup. So we need to include both `request` and `requestGroup` in queryType, and let the query stop at reques type.

PR #7949, opened 2024-09-12
Manage members in organization [INS-2705]

yaoweiprc:
Now users can 
- invite new members
- change member's role
- remove existing member
- revoke invitation
- search members and invites
in app. It's basically migrated from insomnia-website-app. The logic for permission validation and error messages is essentially the same as in the web version. Except that:
- existing members and pending invites are combined into one list
- users can choose role when inviting
image

Other changes:
- Add an react hook `useStateRef` to get latest state immediately after setState, rather than in next round of render.
- Add an extra param `onlyResolveOnSuccess` to insomniaFetch so that it can reject when the response status code is not 2xx.


.vscode/settings.json
```diff
@@ -6,7 +6,7 @@
     "*.njson": "ndjson"
   },
   "files.insertFinalNewline": true,
-  "editor.formatOnSave": true,
+  "editor.formatOnSave": false,
```
jackkav: please keep dx changes in a separate PR, add a PR description, and mention anything you could have done but didn't want to include in the scope. For more clarity do the some PRs reviews over zoom with the team.

yaoweiprc: Current formatter use 4 spaces as tab on my computer. I did not find where to set it. So I just disabled formatOnSave temporarily. 
I have changed it back. I'll try to use some other formatter like Prettier in a separate PR when I'm free.

cwangsmv:
It is better to show a skeleton like web when user opens the invite modal. Users will see "no member" first for now.

https://github.com/user-attachments/assets/128e3c67-6de0-4a21-bcc8-818d4c79e97f

cwangsmv:
Should we change the title and hide stack trace for RBAC error:
Screenshot 2024-09-20 at 14 33 42

Web behavior:

Screenshot 2024-09-20 at 14 35 09

.vscode/settings.json
```diff
@@ -15,4 +15,9 @@
   "[jsonc]": {
     "editor.defaultFormatter": "vscode.json-language-features"
   },
+  "cSpell.words": [
```
CurryYangxx, line 18: What are these configurations for?

packages/insomnia/src/ui/routes/organization.tsx
```diff
@@ -645,16 +647,19 @@ const OrganizationRoute = () => {
...
                     <span className="truncate">
                       Invite
                     </span>
                   </Button>
+                  <InviteModalContainer
+                    {...{
+                      isOpen: isInviteModalOpen,
+                      setIsOpen: setIsInviteModalOpen,
+                    }}
+                  />
```
CurryYangxx, line 662: Why we need destruction here?

CurryYangxx:
This pr include some changes that are not part of this feature:

- add a new react hook
- add a new param  in insomniaFetch

It is better to explain these changes in pr description to make review easier

packages/insomnia/src/ui/components/modals/invite-modal/organization-invite-form.tsx
```diff
@@ -0,0 +1,520 @@
...
+        let json;
+        try {
+          json = await error.response.json();
+        } catch (e) {
+          throw new Error(`Failed to invite ${inviteeEmail}`);
+        }
+        if (json?.message) {
+          throw new Error(json.message);
+        }
+      }
```
CurryYangxx: I think we can do these logic in insomniaFetch function.

packages/insomnia/src/ui/components/modals/invite-modal/organization-invite-form.tsx
```diff
@@ -0,0 +1,520 @@
...
+you don’t have more than one project. Since you have more than one project, you need to upgrade to
+Individual or above to continue.`
+        );
+      }
+      if (json?.error === NEEDS_TO_INCREASE_SEATS_ERROR) {
+        throw new Error(needToIncreaseSeatErrMsg);
+      }
+      if (json?.message) {
+        throw new Error(json.message);
+      }
```
CurryYangxx, line 260: same thing

packages/insomnia/src/ui/insomniaFetch.ts
```diff
@@ -40,6 +59,12 @@ export async function insomniaFetch<T = void>({ method, path, data, sessionId, o
     if (uri) {
       window.main.openDeepLink(uri);
     }
+    if (onlyResolveOnSuccess && !response.ok) {
+      throw new ResponseFailError(
+        `${response.status} ${response.statusText}`,
+        response,
+      );
+    }
```
CurryYangxx:
```suggestion
    if (onlyResolveOnSuccess && !response.ok) {
      const res = await response.json();
      throw new ResponseFailError(
        `res.message`,
      );
    }
```

packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
```diff
@@ -0,0 +1,671 @@
...
+    return 0;
+  }
+  const validMaxPage = Math.floor(total / PAGE_SIZE);
+  if (page > validMaxPage) {
+    return validMaxPage;
+  }
+  return page;
+}
+
+const usePagination = (permissionRef: MutableRefObject<Record<Permission, boolean>>) => {
```
CurryYangxx, line 459: This seems like a hook, should we need to move it to hooks dir?

CurryYangxx:
There are much typescript definition and util functions in component file, we can discuss if it is better to move them to some other dir together.

packages/insomnia/src/ui/components/modals/invite-modal/organization-invite-form.tsx
```diff
@@ -0,0 +1,520 @@
...
+export const InviteForm = ({ onInviteCompleted, allRoles }: {
+  onInviteCompleted?: () => void;
+  allRoles: Role[];
+}) => {
+  const selectedRoleRef = React.useRef<Role>(
+    allRoles.find(role => role.name === defaultRoleName) as Role,
+  );
+  const organizationId = useParams().organizationId as string;
+  const [inputStr, setInputStr] = useState('');
+  const [addedMembers, setAddedMembers, addedMembersRef] = useStateRef<string[]>([]);
```
CurryYangxx, line 24: why we use this `useStateRef` hook here?

.vscode/settings.json
yaoweiprc: For spell check.

packages/insomnia/src/ui/routes/organization.tsx
yaoweiprc: To pass props to components. I prefer this style.

packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
yaoweiprc: This is a hook specifically used here for pagination and is not reusable.

packages/insomnia/src/ui/components/modals/invite-modal/organization-invite-form.tsx
yaoweiprc: To get latest state immediately after setState, rather than in next round of render.

packages/insomnia/src/ui/components/modals/invite-modal/organization-invite-form.tsx
yaoweiprc: I've talked about it with Jack before. Previously, when the status code of a request response was not 2xx, we were unable to retrieve the status code information from the insomniaFetch return. When the status code is not 2xx, the response may contain different information. As a general request method, insomniaFetch should not handle specific failure cases but should return the failed response to the caller, allowing the caller to handle the failure according to the specific API. This modification enables minimal intrusion to insomniaFetch while maintaining compatibility.

packages/insomnia/src/ui/insomniaFetch.ts
yaoweiprc: When response.ok is false, the content-type of the response is not necessarily JSON, and the JSON structure may not always contain a message property. I believe we should return the response directly to the caller and let the caller decide how to handle the error.

yaoweiprc:
> This pr include some changes that are not part of this feature:
> 
> * add a new react hook
> * add a new param  in insomniaFetch
> 
> It is better to explain these changes in pr description to make review easier

Yeah, I have added it.

notjaywu:
> Should we change the title and hide stack trace for RBAC error: Screenshot 2024-09-20 at 14 33 42
> 
> Web behavior:
> 
> Screenshot 2024-09-20 at 14 35 09

image

yaoweiprc:
> It is better to show a skeleton like web when user opens the invite modal. Users will see "no member" first for now.
> 
>  Screen.Recording.2024-09-20.at.14.23.57.mp4

image
Add a spin icon when loading.

yaoweiprc:
> Should we change the title and hide stack trace for RBAC error: Screenshot 2024-09-20 at 14 33 42
> 
> Web behavior:
> 
> Screenshot 2024-09-20 at 14 35 09

image
Use another alert modal.

packages/insomnia/src/ui/insomniaFetch.ts
yaoweiprc: Solved in https://github.com/Kong/insomnia/pull/7949/commits/e98eb1df8bd0b7db48846ca4e2c63f5e8a766a4a

cwangsmv (dismissed):
lgtm

cwangsmv (approved):

PR #7981, opened 2024-09-20
feat: missing variable improvement - [INS-4396]

CurryYangxx:
**Changes**

show all the missing environment variables in request
image
image

### DIFF
--- a/packages/insomnia/src/templating/utils.ts
+++ b/packages/insomnia/src/templating/utils.ts
@@ -1,4 +1,5 @@
 import type { EditorFromTextArea, MarkerRange } from 'codemirror';
+import _ from 'lodash';
 
 import type { DisplayName, PluginArgumentEnumOption, PluginTemplateTagActionContext } from './extensions';
 import objectPath from './third_party/objectPath';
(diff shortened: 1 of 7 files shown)


jackkav:
On a related note, in order to feedback useful context to a CLI user, we may want to change how we handle these. Theres a ticket to address it here INS-4142, provided this PR doesn't make that any more difficult it should be fine.

CurryYangxx:
> On a related note, in order to feedback useful context to a CLI user, we may want to change how we handle these. Theres a ticket to address it here [INS-4142](https://linear.app/insomnia/issue/INS-4142), provided this PR doesn't make that any more difficult it should be fine.

@jackkav  This PR just get all missing variable at once and change some UI, will not affect the original error handling. So I think this will not let it more difficult.

packages/insomnia/src/templating/utils.ts
```diff
@@ -280,13 +281,24 @@ export function decodeEncoding<T>(value: T) {
...
+    let variable = match[1];
+    if (variable.includes('_.')) {
+      variable = variable.split('_.')[1];
+    }
+    // Check if the variable is not present in the context
+    if (_.get(templatingContext, variable) === undefined) {
+      missingVariables.push(variable);
+    }
+  }
+  return missingVariables;
```
CurryYangxx: compare with template context and get all empty variables in one editor.
image

ihexxa (approved):

ihexxa:
image
Checked and did some tests, looks good.

Issue #7946, opened 2024-09-11
Insomnia working very slow after upgrade to 10


shubhendu-enable:
### Expected Behavior

We are using git sync for insomnia , after upgrading to version 10 it should work seamless as before, 

### Actual Behavior

Getting following error when trying to click folders in document 

TypeError: Cannot destructure property 'activeRequest' of '$(...)' as it is undefined.
    at Wc (file:///Applications/Insomnia.app/Contents/Resources/app.asar/debug-BKonWNXc.js:45:11262)
    at Nh (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:43:18469)
    at cj (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:45:3395)
    at Vk (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:45:47812)
    at Uk (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:45:42469)
    at Tk (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:45:42399)
    at Ik (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:45:42252)
    at Nk (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:45:38395)
    at Gk (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:45:37273)
    at Tt (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:30:1726)
    at Immediate.zt [as _onImmediate] (file:///Applications/Insomnia.app/Contents/Resources/app.asar/mainWindow-DV0jvs_A.js:30:2135)
    at process.processImmediate (node:internal/timers:483:21)

### Reproduction Steps

Create Document with multiple nested folders 

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Insomnia Version

10.0.0

### What operating system are you using?

macOS

### Operating System Version

Sonoma 14.6.1

### Installation method

download from insomnia website

### Last Known Working Insomnia version

9.3.3

[...]

PR #7989, opened 2024-09-23
fix: improve git sync slowness

CurryYangxx:
Changes:

- [x] check git changes every 30s
- [x] use promise all to improve speed when get git changes
- [x] add changes&push check after some of git action

Next step:

- [ ] we need try to use [statusMatrix](https://isomorphic-git.org/docs/en/statusMatrix.html) instead of `status`


cwangsmv (approved):
lgtm

CurryYangxx:
related #7946

PR #7992, opened 2024-09-23
Fix bug: Free user see alert when they switch to default project in some case [INS-4464]

yaoweiprc:
Free user can either have one project with multiple members or have multiple project with one member. So if you are a free user and only have one project. You invite someone to your project. Then you create a new project. An Alert of 'Could not create project' shows. Then you remove the other member from your project. You can successfully create a new project. But when you switch to the previous project, the alert shows again.
It's introduced by react router's data router, which persists fetch.data.


gatzjames (approved):
This is much simpler than the useEffect approach but some things to keep in mind:
- It will create sync issues if we do async stuff that then need to update state
- We should upgrade this to an action with the [new react router v7 version](https://github.com/remix-run/react-router/pull/11521/files#diff-0788b922993eedbe2768f25e1f0cdf6ff0455e49ed69e084d8cfedf79efa8730R1077)
- We should hold back from doing the same thing on other actions since it will create a lot of tech debt

PR #8004, opened 2024-09-25
feat: enable run folder mode

ihexxa:
### Background
Make the runner support "Run folder" capability.

### Changes
- [x] Filtering requests according to the target (collection or folder)

image

Ref: INS-4473


filfreire (approved):
LGTM

Small improvements/suggestion:
- Don't hide the Collection Run button when in collection run view in case i want to go from collection run view of my folder into the one for the entire workspace
- If user opens collection run for a folder - automatically "Select All" requests, otherwise I have to 1. open collection run for folder, 2. select all requests  and 3. only then can I run


Alternate idea:
- Run folder would still open the same workspace collection run - and just pre-select all requests & folders under the folder selected

ihexxa:
@filfreire Good ideas.

PR #8007, opened 2024-09-26
fix(runner): display no test message and enable history item deletion

ihexxa:
### Changes
- [x] Display a message when there's no test for one request
- [x] Enable runner history item deletion

Screenshot 2024-09-26 at 10 33 37

Screenshot 2024-09-26 at 10 32 59

Ref: INS-4502


cwangsmv (approved):
lgtm

PR #8021, opened 2024-09-27
Fix new property in request causes existing request show as modified in git sync [INS-4511]

cwangsmv:
**Background**
Currently, When we add a new property to models like Request, we will add the default value of this property in model init method. Otherwise the new property can not be updated due to current design in database.update method which will check if updated keys is valid.

But this will cause all existing models show as modified in git sync, due to current implementation of using YAML stringify method to compare local db content with git origin content. The YAML stringify method will add a new property to the yaml file even the value is undefined.

**Solution**
Instead of adding a new property to request, I have moved the logic to detect if a GraphQL request operation is subscription or not into request model, so no more extra property is needed

### DIFF
--- a/packages/insomnia/src/utils/graph-ql.ts
+++ b/packages/insomnia/src/utils/graph-ql.ts
@@ -1,4 +1,3 @@
-
 import { getOperationAST, parse } from 'graphql';
 
 import { CONTENT_TYPE_GRAPHQL } from '../common/constants';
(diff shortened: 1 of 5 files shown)


packages/insomnia/src/utils/graph-ql.ts
```diff
@@ -1,5 +1,9 @@
+
```
ihexxa: nit: probably unnecessary empty line

ihexxa (dismissed):

packages/insomnia/src/utils/graph-ql.ts
cwangsmv: Fixed

ihexxa (approved):

PR #8082, opened 2024-10-15
fix: stop interval when window not focused - [INS-4571]

CurryYangxx:
Changes:

- add `isWinsowFocused` bridge
- listen main window `focus` and `blur` event
- get window focus states in the renderer, and when the window is blurred, we should stop getting git changes interval
- add a flag `isCheckingGitChanges ` to record the task status, if the last time task is not finished, skip this time.

Background:
We use a timer in the collection to periodically check the status of the git repository and update db, to display uncommitted changes in the dashboard page.
image

### DIFF
--- a/packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
+++ b/packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
@@ -108,7 +108,7 @@ export const GitSyncDropdown: FC<Props> = ({ gitRepository, isInsomniaSyncEnable
     isCheckingGitChanges.current = true;
     await checkGitChanges(workspaceId);
     isCheckingGitChanges.current = false;
-  }, isWindowFocused ? 1000 : null);
+  }, isWindowFocused ? 30 * 1000 : null);
 
   useEffect(() => {
     if (shouldFetchGitRepoStatus) {
--- a/packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
+++ b/packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
@@ -1,5 +1,5 @@
 import type { IconName, IconProp } from '@fortawesome/fontawesome-svg-core';
-import React, { type FC, useEffect, useState } from 'react';
+import React, { type FC, useEffect, useRef, useState } from 'react';
 import { Button, Collection, Menu, MenuItem, MenuTrigger, Popover, Section, Tooltip, TooltipTrigger } from 'react-aria-components';
 import { useFetcher, useParams, useRevalidator } from 'react-router-dom';
 import { useInterval } from 'react-use';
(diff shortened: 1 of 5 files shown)


packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
```diff
@@ -85,11 +103,14 @@ export const GitSyncDropdown: FC<Props> = ({ gitRepository, isInsomniaSyncEnable
...
-    });
-  }, 30 * 1000);
+  useInterval(async () => {
+    if (isCheckingGitChanges.current) {
+      return;
+    }
+    isCheckingGitChanges.current = true;
+    await checkGitChanges(workspaceId);
+    isCheckingGitChanges.current = false;
+  }, isWindowFocused ? 1000 : null);
```
filfreire: should the interval here still be `30 * 1000` instead of `1000`?

CurryYangxx: Good catch, I forgot to restore it after testing😅

filfreire (approved):

packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
```diff
@@ -85,11 +103,14 @@ export const GitSyncDropdown: FC<Props> = ({ gitRepository, isInsomniaSyncEnable
...
-    requestIdleCallback(() => {
-      checkGitChanges(workspaceId);
-    });
-  }, 30 * 1000);
+  useInterval(async () => {
+    if (isCheckingGitChanges.current) {
+      return;
+    }
+    isCheckingGitChanges.current = true;
+    await checkGitChanges(workspaceId);
```
gatzjames, line 109: Not sure this is updating the UI since it doesn't use the routes and the app doesn't re-render

packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
```diff
@@ -48,11 +49,28 @@ export const GitSyncDropdown: FC<Props> = ({ gitRepository, isInsomniaSyncEnable
...
   const gitStatusFetcher = useFetcher<GitStatusResult>();
 
+  const isCheckingGitChanges = useRef(false);
+
   const loadingPush = gitPushFetcher.state === 'loading';
   const loadingPull = gitPullFetcher.state === 'loading';
   const loadingFetch = gitFetchFetcher.state === 'loading';
   const loadingStatus = gitStatusFetcher.state === 'loading';
 
+  useEffect(() => {
```
gatzjames, line 59: Not sure we need the bridge here. Could you share some more info on your approach?
Tried with `document.hasFocus()` and it works as expected if the app is not focused since we only have one tab in one window renderer

gatzjames (commented):
I think we can use the `document.hasFocus()` method and avoid the bridge here.
Also the checkGitChanges doesn't seem to work as expected.

packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
CurryYangxx: @gatzjames do you mean the checkGitChanges function? We don't need to update the UI in this scenario， we just need to update the `hasUncommitChange` db field when the user is working on the collection page, and we use this field to show notifications on the other router page(dashboard page).

packages/insomnia/src/ui/components/dropdowns/git-sync-dropdown.tsx
CurryYangxx: @gatzjames I haven't used `document.hasFocus` before, I can try it and if it works I can use it instead. And I think listening to the BrowserWindow `focus` and 'blur'  event is necessary here.

CurryYangxx:
After discussing with the team, we have some concern about using focus to control when the interval starts may introduce some side effects. Also after James's pr #8027, the performance of git status has been greatly improved.
So we decide to open another pr: #8086 to just make sure the previous task is completed before executing the next one .

PR #8086, opened 2024-10-16
feat: skip changes checking if last task is not finish

CurryYangxx:
Changes

- Add a flag to record git repo changes checking task status, if the last task is running, skip this time.

This is a small optimization for #7989 to prevent task accumulation caused by slow task execution.

related to INS-4571


cwangsmv (approved):
lgtm

PR #8095, opened 2024-10-17
fix(Git): issue with including unit-tests and protofiles

gatzjames:
Highlihts:

- [x] Updates the filters we pass to withDescendants in order to include unit-tests and proto-files inside test-suites and proto-directories
- [x] Fixes the issue where adding new tests/proto-files the result is not being shown in git status

### DIFF
--- a/packages/insomnia/src/sync/git/ne-db-client.ts
+++ b/packages/insomnia/src/sync/git/ne-db-client.ts
@@ -171,6 +171,14 @@ export class NeDBClient {
         typeFilter = [models.protoDirectory.type, type];
       }
 
+      if (type === models.mockRoute.type) {
+        typeFilter = [models.mockServer.type, type];
+      }
+
+      if (type === models.webSocketPayload.type) {
+        typeFilter = [models.webSocketRequest.type, type];
+      }
+
       const children = await db.withDescendants(workspace, null, typeFilter);
       docs = children.filter(d => d.type === type && !d.isPrivate);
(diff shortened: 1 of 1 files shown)


CurryYangxx:
I think we also need to deal with `models.mockRoute.type`

gatzjames:
> I think we also need to deal with `models.mockRoute.type`

Good one also added websocket payload.

Looking at the models one by one atm but let me know if you find any other model I haven't caught

filfreire (approved):

packages/insomnia/src/sync/git/ne-db-client.ts
```diff
@@ -156,8 +156,29 @@ export class NeDBClient {
       ];
     } else if (type !== null && id === null) {
       const workspace = await db.get(models.workspace.type, this._workspaceId);
+      let typeFilter = [type];
+
       const modelTypesWithinFolders = [models.request.type, models.grpcRequest.type, models.webSocketRequest.type];
-      const typeFilter = modelTypesWithinFolders.includes(type) ? [models.requestGroup.type, type] : [type];
+      if (modelTypesWithinFolders.includes(type)) {
```
yaoweiprc, line 162: If we want to get all requests in a workspace, It's unintuitive that we need to pass [requestGroup, request] as the queryTypes argument to withDescendants and then filter out the requestGroups from result. Maybe we can do it within withDescendants method.

PR #8101, opened 2024-10-18
fix: should not print error message with exit code 1 when status != 200

ihexxa:
### Changes
- [x] Disable response code check and error messaging in runner execution

More details in INS-4586


filfreire (approved):

cwangsmv (approved):
lgtm

PR #8109, opened 2024-10-21
fix(Git Sync): Fix issue with conflicts when cloning a repo

gatzjames:
Fixes an issue where cloning a repo would throw a conflict error

Cause of the issue:
- We shallow clone a repo, check if there are many workspaces or an existing one and if not we then put things in the database
- Putting things in the database goes through the db migration logic adding new properties or updating existing ones
- When we actually clone later git figures out we already have some files (objects in db) that are changed compared to the repo
- Throws conflict errors

The minimal change I could think of (what I did in this PR):
- Keep shallow cloning and if the checks pass actually clone instead of manually updating the db
- Keep the needsFullClone functionality in case there are network errors while actually cloning e.g. network errors etc. since we already create some resources in db

Other ideas that require more fundamental changes:
- Find a way to remove the double clone requirement (e.g. no shallowClone moving forward)
- If clone fails rollback any db changes or anything that shouldn't be in the app if it didn't work as expected

### DIFF
--- a/packages/insomnia/src/ui/routes/git-actions.tsx
+++ b/packages/insomnia/src/ui/routes/git-actions.tsx
@@ -10,7 +10,6 @@ import * as models from '../../models';
 import type { GitRepository } from '../../models/git-repository';
 import { createGitRepository } from '../../models/helpers/git-repository-operations';
 import {
-  isWorkspace,
   WorkspaceScopeKeys,
 } from '../../models/workspace';
 import { fsClient } from '../../sync/git/fs-client';
@@ -467,12 +466,12 @@ export const cloneGitRepoAction: ActionFunction = async ({
   });
   repoSettingsPatch.needsFullClone = true;
 
-  const fsClient = MemClient.createClient();
+  const inMemoryFsClient = MemClient.createClient();
 
   const providerName = getOauth2FormatName(repoSettingsPatch.credentials);
   try {
     await shallowClone({
-      fsClient,
+      fsClient: inMemoryFsClient,
       gitRepository: repoSettingsPatch as GitRepository,
     });
   } catch (e) {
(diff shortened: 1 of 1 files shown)


filfreire (approved):

ryan-willis (approved):

packages/insomnia/src/ui/routes/git-actions.tsx
```diff
@@ -467,12 +466,12 @@ export const cloneGitRepoAction: ActionFunction = async ({
...
+  const inMemoryFsClient = MemClient.createClient();
 
   const providerName = getOauth2FormatName(repoSettingsPatch.credentials);
   try {
     await shallowClone({
-      fsClient,
+      fsClient: inMemoryFsClient,
       gitRepository: repoSettingsPatch as GitRepository,
     });
   } catch (e) {
```
CurryYangxx, line 477: If we still need `shallowClone` here?

yaoweiprc:
Now you call GitVCS.initFromClone in cloneGitRepoAction and set gitRepository.needsFullClone to false. So that we do not clone the repo again when user clone a repository in project dashboard and then enter the cloned workspace.
So are there any situations that needsFullClone is true when executing gitRepoLoader now?

gatzjames:
The only reason I could think of is that one of the network related operations fails during cloning and the user is left with an empty workspace. In that case needsFullClone would be false and pick up from where it left off

packages/insomnia/src/ui/routes/git-actions.tsx
gatzjames: I think we do in case there are many workspaces or if the workspace we try to clone already exists and we want to throw an error.

CurryYangxx (approved):

PR #8134, opened 2024-10-28
fix: timers is not reset if no timing entry has been found

ihexxa:
### Changes
- [x] timers should be reset when no timing information has been found.

More details: INS-4616


packages/insomnia/src/ui/hooks/use-execution-state.ts
```diff
@@ -14,6 +14,8 @@ export function useExecutionState({ requestId }: { requestId?: string }) {
       const targetSteps = await window.main.getExecution({ requestId });
       if (targetSteps) {
         isMounted && setSteps(targetSteps);
+      } else {
+        setSteps([]);
```
gatzjames: Should we check for isMounted here too?

ihexxa: Good catch 👍

cwangsmv (approved):

PR #8143, opened 2024-10-31
fix(cli): enable setNextRequest for CLI

ihexxa:
### Background
This change basically tries to keep most of the changes from the proposed solution although some work might still be required to make the solution more comprehensive. Therefore this PR tries to make minimal changes to enable `setNextRequest` in CLI and also keep tests and enhancements from the proposed solution.

### Changes
- [x] Enabled `setNextRequest` for CLI
- [x] Also included test results from pre-request tests 
- [x] Added tests

Ref: INS-4604


cwangsmv (approved):
Test different setNextRequest situation and LGTM

PR #8151, opened 2024-11-06
feat: multiple tabs - [INS-4225]

CurryYangxx:
**Changes**

- [x] create a `TabList` and `Tab` Component, and import it in every type of file page.
- [x] store tab data in localstorage, use `InsomniaTabContext` to manage state of the tabs
- [x] create a `useInsomniaTab` hook
  - [x] get tab type according to the location pathname
  - [x] check if tab exist
  - [x] pack tab data
  - [x] add new tab
- [x] sync tab data with db
- [x] click "+" to add tab
- [x] right click context menu: Close all & Close others
  - [x] rename the origin IPC `show-context-menu` to `show-nunjucks-context-menu`
  - [x] add a new IPC `show-context-menu` as the common context menu method
- [x] support drag and drop to re-arrange tabs
- [x] collection runner improvements: #8244 
- [x] keep long connection for openning tabs: #8266 

image

related to #6108


filfreire (commented):
LGTM

Found issue where tabs disappear when pressing OPTION + TAB in the UI after a few times

some suggestions for this PR or other follow-up PRs:
- allow for pressing CTRL + TAB / CTRL + SHIFT + TAB to alternate between tabs
- perhaps also allow the same for pressing CTRL + 1  (1 to 9) to go to 1st -> 9th tab, 0 being last tab? 
- allow middle-mouse click to close a tab
- capture COMMAND+W (and windows/linux equivalent) to close tabs

gatzjames (dismissed):

filfreire (dismissed):

kaynansc:
it would be nice capture COMMAND+W (and windows/linux equivalent) to close tabs

packages/insomnia/src/ui/context/app/insomnia-tab-context.tsx
```diff
@@ -0,0 +1,272 @@
...
+      return;
+    }
+
+    const index = currentTabs.tabList.findIndex(tab => tab.id === id);
+    if (index === -1) {
+      return;
+    }
+    const newTabList = currentTabs.tabList.filter(tab => tab.id !== id);
+    if (currentTabs.activeTabId === id) {
+      navigate(newTabList[index - 1 < 0 ? 0 : index - 1]?.url || '');
```
yaoweiprc: Math.max(index - 1, 0)

packages/insomnia/src/ui/components/document-tab.tsx
```diff
@@ -0,0 +1,35 @@
...
+      ].map(item => (
+        <NavLink
+          key={item.id}
+          to={`/organization/${organizationId}/project/${projectId}/workspace/${workspaceId}/${item.id}`}
+          className={({ isActive, isPending }) =>
+            `${isActive
+              ? 'text-[--color-font] bg-[--color-surprise]'
+              : ''
+            } ${isPending ? 'animate-pulse' : ''} text-center rounded-full px-2`
+          }
```
cwangsmv: Use classnames lib can make this more clear like:
```
classnames('text-center rounded-full px-2', {
  'text-[--color-font] bg-[--color-surprise]': isActive,
  'animate-pulse': isPending,
})
```

CurryYangxx:
Update: (https://github.com/Kong/insomnia/pull/8151/commits/5e9072681d43d26ba1930696e6ae9bf5540e8351)
Because we can move a request or a folder to another collection. So, we need to update tab data after the move action.

- add patches field to database changeBuffer, so that we can know if user change the parentId
- request move
  - update tab by id (workspaceId, workspaceName, url)
  - revalidate workspace activeRequest
- folder move
  - batch update (workspaceId, workspaceName, url)
  - revalidate workspace activeRequest (if the activeRequest is in the folder, we need to clear it)
image

filfreire:
@CurryYangxx, I took a pass at the branch again and tried after your commit and seems copying and moving seems to be working.

I noticed one other thing - for example if you have 2 tabs open for 2 requests, and maybe you are working on pre-request/after-response scripts, or docs, or headers.... etc. in both, the moment you switch tabs, you are not redirected to the "sub-tab" you were at when browsing the request.

So each time we switch between tabs we don't remember where inside the tab specifically the user was.

CurryYangxx:
support drag and drop to re-arrange: https://github.com/Kong/insomnia/pull/8151/commits/bb5d15863db68314e3c62285ca5f523dfb511568

CurryYangxx:
> I noticed one other thing - for example if you have 2 tabs open for 2 requests, and maybe you are working on pre-request/after-response scripts, or docs, or headers.... etc. in both, the moment you switch tabs, you are not redirected to the "sub-tab" you were at when browsing the request.
> 
> So each time we switch between tabs we don't remember where inside the tab specifically the user was.

@filfreire Currently, we only have one component as the tab content for multiple tabs, also we only have one front-end state. Every time we switch to another tab, we do router navigation, and the component re-renders, but the state will not update except that is stored in the database.
If we want to remember the user's operation status in different tabs, we need to maintain the corresponding react state for different tabs in context or somewhere else, this will be more complex. I think it is ok now, we can ship multi-tabs first and determine if we need to do that according to the user feedback.

smiller0924:
Where can I find the instructions to add the Tabs functions to Insomnia?

CurryYangxx:
@smiller0924 This feature is still in progress, and have no instructions yet. When we release this feature, we will add some instruction to our docs.

packages/insomnia/src/common/database.ts
```diff
@@ -709,10 +710,11 @@ let changeListeners: ChangeListener[] = [];
 
 /** push changes into the buffer, so that changeListeners can get change contents when database.flushChanges is called,
  * this method should be called whenever a document change happens */
-async function notifyOfChange<T extends BaseModel>(event: ChangeType, doc: T, fromSync: boolean) {
+async function notifyOfChange<T extends BaseModel>(event: ChangeType, doc: T, fromSync: boolean, patches: Patch<T>[] = []) {
   const updatedDoc = doc;
 
-  changeBuffer.push([event, updatedDoc, fromSync]);
+  // TODO: Use object is better than array
+  changeBuffer.push([event, updatedDoc, fromSync, patches]);
```
jackkav, line 717: can you explain what behavior a ui tabs feature has which needs a new persistence layer feature. Is there an alternative way to achieve this behavior?
The reason I ask is this code is already hacked to achieve a sync side-effect so this smell like a hack too.

packages/insomnia/src/ui/components/tabs/tab.tsx
```diff
@@ -0,0 +1,159 @@
...
+  Folder = 'folder',
+  Env = 'environment',
+  Mock = 'mock-server',
+  MockRoute = 'mock-route',
+  Document = 'document',
+  Collection = 'collection',
+  Runner = 'runner',
+  TEST = 'test',
+  TESTSUITE = 'test-suite',
+};
```
jackkav: string union types are superior to enums eg 'request' | 'folder'
https://www.typescriptcourse.com/string-literal-unions-over-enums

packages/insomnia/src/ui/components/tabs/tabList.tsx
```diff
@@ -0,0 +1,380 @@
...
+    const innerWidth = tabListInnerRef.current?.clientWidth;
+    const wrapperWidth = tabListWrapperRef.current?.clientWidth;
+    if (innerWidth && wrapperWidth && innerWidth > wrapperWidth) {
+      setIsOverFlow(true);
+    } else {
+      setIsOverFlow(false);
+    }
+  };
+
+  const debouncedOnResize = _.debounce<(size: Size) => void>(onResize, 500);
```
jackkav: we don't use underscorejs or lodash in this project except to support legacy postman scripts. https://github.com/kong/insomnia/blob/eb826ff48e35e15a6d0b85bbd9ccbe51aeda4c98/packages/insomnia/src/common/misc.ts#L119

packages/insomnia/src/common/database.ts
CurryYangxx: We sync the tab states with the database by listening to the database's `onChange` event, such as changing a request's name or request method. The `onchange` event is triggered frequently, we only want to perform the corresponding update operation when certain fields are changed, because some of the update have heavy logic, such as move request or folder to another workspace. So, I add `patches` to the buffer to help us avoid some unnecessary operations.
https://github.com/Kong/insomnia/blob/d0e9218af0dfa6bfc5625270330465d79f603474/packages/insomnia/src/ui/components/tabs/tabList.tsx#L153

packages/insomnia/src/ui/components/tabs/tabList.tsx
CurryYangxx: Updated. By the way, why we don't want to use lodash? Some of the functions are useful and convenient.

packages/insomnia/src/ui/components/tabs/tab.tsx
CurryYangxx: I think whether to use enums or union depends on the specific scenario. I use enums here for easier maintenance because we have many places using this type and when we want to change one of them, we only need to modify one place, and it will not cause lots of changes. If using union, we may forget to modify some places and cause some bugs when the type definition is incomplete. 
https://medium.com/@gubrani.sanya2/typescript-enums-or-union-types-which-one-should-you-use-377ecae947af

jackkav: the type-check command will fail if the string union is not me. The article you link to suggests picking one and sticking to it, it this project we have chosen string union please respect consistency.

packages/insomnia/src/ui/components/tabs/tabList.tsx
jackkav: lodash was made obsolete in 2021, when most of the meaningful utility functions were added to javascript and the remainder are one liners. I don't expect to debate this here. If you read about it and still have questions reach out on slack or zoom.

jackkav:
improvement ideas for now or later:

analyse state choices against the choices in the rest app
localstorage, refs.
main map+useeffect?

db.onchange to trigger rerender
consider using the router loader instead of a surgical listener to get rerenders after state changes?

look for a way to remove the ref

look for a way to remove the event bus, figure out how to persist runner info on disk or local storage?

look for a way to remove the context provider,  flatten useInsomniaTabContext into tablist and pass down function needed for tabs in react props

duplicate useInsomniaTabContext

enum to union

solve sync problems in runner upfront before this pr so this pr doesn't have to treat it as a special thing that needs lots of custom code

packages/insomnia/src/ui/components/tabs/tab.tsx
yaoweiprc: I think enum is a better choice here.

jackkav (commented):
In order to help move this forward my suggestion would be to flatten the context providers and hooks back into the components they are used in and we can then review if any abstractions are required with the team.

jackkav:
1. Theres an inconsistency with the tab headers, the icons for mocks and requests are different
image

2. this dropdown is inconsistent with others too

image


3. the onresize observer code could be replaced with simple css

4. there are too many refs introduced in this pr there ideally should be none
 https://blog.logrocket.com/why-you-should-use-refs-sparingly-in-production/

CurryYangxx:
> 1. Theres an inconsistency with the tab headers, the icons for mocks and requests are different
> 
> image
This a mistake when I converting enum to union, already fixed it. Thanks for the catch.

packages/insomnia/src/ui/routes/workspace.tsx
```diff
@@ -315,6 +315,24 @@ export const workspaceLoader: LoaderFunction = async ({
   };
 };
 
+export const revalidateWorkspaceActiveRequest = async (requestId: string, workspaceId: string) => {
```
ihexxa, line 318: nit: probably name it as `deactiveWorkspaceActiveRequest` as it sets active to null?

packages/insomnia/src/ui/components/tabs/tabList.tsx
```diff
@@ -0,0 +1,382 @@
...
+      <div className='flex items-center justify-start flex-shrink-0 flex-grow border-b border-solid border-[--hl-sm]'>
+        <MenuTrigger>
+          <Button aria-label="Menu" className="w-[40px] text-center">
+            <Icon icon="plus" className='cursor-pointer' />
+          </Button>
+          <Popover>
+            <Menu className='border max-w-lg select-none text-sm border-solid border-[--hl-sm] shadow-lg bg-[--color-bg] py-2 rounded-md overflow-y-auto max-h-[85vh] focus:outline-none'>
+              {currentPage === 'debug' && (
+                <MenuItem className="aria-disabled:opacity-30 aria-disabled:cursor-not-allowed flex gap-2 px-[--padding-md] aria-selected:font-bold items-center text-[--color-font] h-[--line-height-xs] w-full text-md whitespace-nowrap bg-transparent hover:bg-[--hl-sm] disabled:cursor-not-allowed focus:bg-[--hl-xs] focus:outline-none transition-colors" onAction={addRequest}>
+                  add request to current collection
```
ihexxa: We should make `add` capitalized.

packages/insomnia/src/ui/components/tabs/tabList.tsx
```diff
@@ -0,0 +1,382 @@
...
+                </MenuItem>
+              )}
+              <MenuItem className="aria-disabled:opacity-30 aria-disabled:cursor-not-allowed flex gap-2 px-[--padding-md] aria-selected:font-bold items-center text-[--color-font] h-[--line-height-xs] w-full text-md whitespace-nowrap bg-transparent hover:bg-[--hl-sm] disabled:cursor-not-allowed focus:bg-[--hl-xs] focus:outline-none transition-colors" onAction={addRequestToCollection}>
+                add request to other collection
```
ihexxa: Ditto

packages/insomnia/src/ui/routes/workspace.tsx
CurryYangxx: It set active to null only when the active request is removed, so may be revalidate is better

CurryYangxx:
> 4. there are too many refs introduced in this pr there ideally should be none
>    https://blog.logrocket.com/why-you-should-use-refs-sparingly-in-production/

The use scenario of ref mentioned in this article is different from that in this pr. First, the article points out that ref is used to directly operate DOM, and uses class components as an example. However, in react function components, useRef can not only obtain the reference of DOM, but also be used to reference a variable to avoid re-rendering or side effects. The scenario in pr is the latter.

jackkav:
https://refine.dev/blog/react-useref-hook-and-ref/#best-practices-when-working-with-refs

> Avoid over-reliance on refs. You should use refs as an escape hatch to access DOM elements, browser APIs, and work with systems external to your React application. If you find yourself over-relying on refs, probably there is something you're doing wrong.



I have made a PR to show how you could go about a first pass of simplification https://github.com/Kong/insomnia/pull/8322/files

I would encourage you to keep taking passes aiming to remove one hook, context or ref with each pass.

yaoweiprc (approved):

Issue #8148, opened 2024-11-04
Bug when running a folder of request


akramsarkis:
### Expected Behavior

I have a folder of requests, the aim is to run the folder to test scenarios.

<details>
<summary> Variables set are not used in the next query of the folder </summary>

- I have a login request from which I am setting variables to use them for my next request

````
insomnia.test("Verify login status code", () => {
    // Extract values from the response and set them as environment variables
	  console.log(insomnia.response.json().token)
    
    // Set variables
    insomnia.variables.set('token_admin', insomnia.response.json().token.token);
    insomnia.variables.set('project_id', insomnia.response.json().projects[0].saas_project_id);

    // Log the values to confirm
    console.log('token_admin:', insomnia.variables.get('token_admin'));
    console.log('project_id:', insomnia.variables.get('project_id'));

    // Verify the response status code
    insomnia.expect(insomnia.response.code).to.eql(200);
});

````
- When adding a pre request script to my next query to use them, I am expecting to have the values set used

````
// Log the existing variables to ensure they are available
console.log(insomnia.variables.get('token_admin'));
console.log(insomnia.variables.get('project_id'));

````
</details>

<details>
<summary> In the pre request of a query I am generating I value I want to use as path parameter and it is not </summary>

In the pre request of a query I am generating I value I want to use as path parameter and I am expecting it to work

````
// Log the existing variables to ensure they are available
console.log(insomnia.variables.get('token_admin'));
console.log(insomnia.variables.get('project_id'));

// Function to generate a random alphanumeric string of specified length
function generateRandomString(length) {
    let result = '';
    const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
    const charactersLength = characters.length;
    for (let i = 0; i < length; i++) {
        result += characters.charAt(Math.floor(Math.random() * charactersLength));
    }
    return result;
}

// Generate a random string of length 8
const randomString = generateRandomString(8);
const sectionname = `section+${randomString}`;

// Set the generated section name to a variable
insomnia.variables.set("generated_sectionname", sectionname);

// Log the generated section name to confirm
console.log('Generated Section Name:', sectionname);

````

</details>

### Actual Behavior

1) For the first point the actual behaviour is to not use set variables in the next query, it is using the values I have set in my environment variable

Here the variable set by the first login query

````

< HTTP/2 200 
< date: Mon, 04 Nov 2024 15:00:57 GMT
[...]

PR #8157, opened 2024-11-08
fix(runner): extend insomnia.variables.set lifetime to whole execution period

ihexxa:
### Background
When `insomnia.variables.set` sets a variable, the variable should exist during whole runner execution period. Currently it is dropped after requesting.

### Changes
- [x] Wired up predefined local variables from runner to script runtime
- [x] Input local variables to the rendering step 
- [x] Added and updated tests

Ref: INS-4665, #8148


yaoweiprc (approved):
lgtm

PR #8172, opened 2024-11-14
fix(cli): insomnia.variables.set should persist variables during runner execution period

ihexxa:
This change focuses on CLI.

### Changes

- [x] Init transient variables in CLI runner
- [x] Wire transient variables up to the sendRequest and scriptExecutor

Ref: #8148, INS-4647


cwangsmv (commented):
Test local variables across different requests in the same runner and it works as expected

cwangsmv (approved):

Issue #5059, opened 2022-08-08
could not parse PKCS12 file, check password, OpenSSL error error:0308010C:digital envelope routines::unsupported


haijunsu:
# See workaround solution: https://github.com/Kong/insomnia/issues/5059#issuecomment-1674101495

### Expected Behavior
Return response body without errors

### Actual Behavior
Cannot connect to the server because there is an error while parses the PKCS12 file.

### Reproduction Steps

1. Document settings
2. Client Certificates tab
3. New Certificate
4. Add a p12 certificate for a domain
5. Create request and send to the domain which is created in step 4.

### Is there an existing issue for this?

- [x] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Additional Information

Timeline logs:
```
* Preparing request to https://mytest_domainname
* Current time is 2022-08-08T16:53:18.142Z
* Using libcurl/7.79.1-DEV OpenSSL/3.0.0 zlib/1.2.11 brotli/1.0.9 WinIDN libssh2/1.9.0_DEV nghttp2/1.45.1
* Using default HTTP version
* Disable timeout
* Enable automatic URL encoding
* Disable SSL validation
* Enable cookie sending with jar of 4 cookies
* Adding SSL P12 certificate
*   Trying xxx.xxx.xxx.xxx:443...
* Connected to apigatewayb2b.test.bnymellon.com (10.8.77.4) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* could not parse PKCS12 file, check password, OpenSSL error error:0308010C:digital envelope routines::unsupported
* Closing connection 0
```

### Insomnia Version
2022.1.1 and newer versions

### What operating system are you using?
Environment: Windows 10 21H2

### Installation method
Download from website and click the executable file.

### Last known Working Insomnia version
2021.7.2

filfreire:
Hi @haijunsu, could you double check if you can also reproduce this issue in our [latest stable release (2022.5.0)](https://github.com/Kong/insomnia/releases/tag/core%402022.5.0)?

We had a similar issue (#4543) a while back related to one of the dependencies we use ([node-libcurl](https://github.com/Kong/node-libcurl)) and it might have something to do with the PKCS12 file being in a format not supported by recent curl/node-libcurl versions ([other different example](https://stackoverflow.com/a/72600724)).

haijunsu:
[...]

PR #8192, opened 2024-11-21
spike: try openssl legacy providers and weak ciphers

filfreire:
Possibly helps with #5059

related to INS-4508

##  Note for posterity 

_rabbit-hole. dragons. it's the slot machine you already spent too much money on and don't want to leave your chair so someone else will take the big win._

Don't waste too much time retrying to enable legacy and weak ciphers on node-libcurl side. We should burn it all with fire.

One alternative path can be to try to solve the problem on insomnia side with UX:

- Users import the p12 / pfx file - we check the OS - if it's windows or linux we warn them we will also run an extra step, where we use `pkcs12` from `node-forge` (see [this example on SO](https://stackoverflow.com/a/55110537)) to run the equivalent to these two commands bellow to get the key and .crt file (which will work in Windows), and then use those instead right before saving the client certificate:
```bash
openssl pkcs12 -in cert.pfx -clcerts -nokeys -out the-crt-file.crt -legacy
openssl pkcs12 -in cert.pfx -nocerts -out the-key-file.key -legacy
```
- We should also probably have a warning tooltip or similar or hide the pfx/p12 option from windows users we if can guarantee it never works there
- We can also probably print out the above `openssl` commands in a warning modal or something to let users know of the potential solution.
- In the cases where this problem is further compounded due to use of weak ciphers when the users first got the .pfx/.p12 files, which we preemptively don't know - the trick is almost always better for them to use `legacy` magic on openssl on their side, than us trying to add support for those deprecated ciphers on our side.

Life finds a way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1688,9 +1688,9 @@
       }
     },
     "node_modules/@getinsomnia/node-libcurl": {
-      "version": "2.31.3",
-      "resolved": "https://registry.npmjs.org/@getinsomnia/node-libcurl/-/node-libcurl-2.31.3.tgz",
-      "integrity": "sha512-pek6a7gVthdanjjpf585Pmd2uHrxa+g/3qc0b9aMhBIC7MYPUv3fkkkHyVBppIg/T1FCPRGpkOT7Cmfn3QrCNg==",
+      "version": "2.31.4-alpha.0",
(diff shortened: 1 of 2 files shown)


filfreire:
doesn't do the trick, alternative is https://github.com/Kong/insomnia/issues/5059#issuecomment-1674101495

PR #8194, opened 2024-11-22
feat(runner/settings): enable logging switch configuration

ihexxa:
### Changes
- [x] Supported turning on/off logging by introducing runtime
- [x] Added smoke test

Ref: INS-4421

### DIFF
--- a/packages/insomnia/src/network/network.ts
+++ b/packages/insomnia/src/network/network.ts
@@ -33,7 +33,7 @@ import type { WebSocketRequest } from '../models/websocket-request';
 import { isWorkspace, type Workspace } from '../models/workspace';
 import * as pluginContexts from '../plugins/context/index';
 import * as plugins from '../plugins/index';
-import type { SendActionRuntime } from '../ui/routes/request';
+import { defaultSendActionRuntime, type SendActionRuntime } from '../ui/routes/request';
 import { invariant } from '../utils/invariant';
 import { serializeNDJSON } from '../utils/ndjson';
 import {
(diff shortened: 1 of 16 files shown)


packages/insomnia/src/network/network.ts
```diff
@@ -610,6 +623,7 @@ export async function sendCurlAndWriteTimeline(
   settings: Settings,
   timelinePath: string,
   responseId: string,
+  runtime: SendActionRuntime,
```
cwangsmv: Is it better to make the **runtime** parameter optional and give it a default value like:
```
const runtime = runtime || defaultSendActionRuntime
```
This can avoid change **sendCurlAndWriteTimeline** functional calling many places with defaultSendActionRuntime as runtime parameter value.

Since this function is used widely, maybe we can change the function parameter to object so that adding new parameter is much easier to implement
```
interface SendCurlAndWriteTimelineOptions = {
  renderedRequest: RenderedRequest;
  clientCertificates: ClientCertificate[];
  caCert: CaCertificate | null;
  settings: Settings;
  timelinePath: string;
  responseId: string;
  runtime?: SendActionRuntime;
}
sendCurlAndWriteTimeline(options: SendCurlAndWriteTimelineOptions)
```

packages/insomnia/src/ui/routes/request.tsx
```diff
@@ -482,15 +495,17 @@ export const sendActionImplementation = async ({
   iteration,
   iterationCount,
   transientVariables,
+  runtime,
```
cwangsmv: Same issue here, we can make runtime optional and set defaultSendActionRuntime as default value

cwangsmv (dismissed):
Turn on/off log feature in collection runner is working as expected with the change.

packages/insomnia/src/network/network.ts
ihexxa: Good point 👍 , the argument has been updated with default values.

packages/insomnia/src/ui/routes/request.tsx
ihexxa: Applied similar changes.

cwangsmv (approved):
lgtm

PR #8221, opened 2024-12-02
Fix: File path fails when used in env vars on Windows [INS-4742]

### LINKED ISSUE #5754: Reading file in an environment variable does not translate into the query
### Expected Behavior

Under windows:
I define an environment variable that points to a file:
```
{
   "file": "File"
}
```
I should be able to use `.file` in the query.

### Actual Behavior

The path fails. It's because the `\\` in the path, get removed when being sent as a string. So the file cannot be read and the `.file` is marked in red.

If I use a custom command in the environment variable definition and switch `\\` to `/` then `.file` works in the query

### Reproduction Steps

1.  define an environment variable that points to a file:
```
{
   "file": "File"
}
```
2. Try to use `.file` in a query
3. Go back to the environment variable, file on `File`, switch to **custom**, and change `\\` to `/`
4. The `.file` env variable can now be used

### Additional Information

Version: Insomnia 2022.7.5
Build date: 1/19/2023
OS: Windows_NT x64 10.0.19045
Electron: 22.0.0
Node: 16.17.1
V8: 10.8.168.20-electron.0
Architecture: x64

### Insomnia Version

2022.7.5

Windows

### Operating System Version

Windows 10 22H2 19045.2486

### Installation method

download from insomnia.rest

_No response_

filfreire:
## Fix

When we do a double-escape of the `\` slash on `tag-editor.tsx` it solves the issue.

It seems before we were sanitizing the tag in different ways, and that didn't work, but it's unclear how it may have worked years ago. For now this change seems to do the trick

Explored some alternatives, but going down the path of JSON.stringify'ing the argument elsewhere, or trying to sanitize it elsewhere, becomes a trap quickly and doesn't fully work.

<details>

<summary>Investigation details here</summary>

> Note: dragons 🐉 and spaghetti 🍝 , 
>
> this is failing as far back as 2023. Tried with `10.1.1`, `8.5.1` and even `2023.5.8`. Found a bug report about it here: https://github.com/Kong/insomnia/issues/5754 
>
> this is a sandworm/Shai-hulud type of issue for spending time

When using File template tags as value for environment variables in the JSON / non-table view of the Environment Editor, the file path gets stored with a bad path on the environment. 

- Does not seem to happen when setting the file template tag in the new environment table view editor, it's specific to the JSON view
- Throughout the codebase there's a handful of places where we replace `\\\\` to `\\` 
- There's also places where we do the oposite
- By the time we get the values from Env Editor, it's showing up as `\\` - which then magically dissappears when attempting to use the Env variable in a request, so a string like `C:\\Users\\filipe\\Desktop\\test.txt` ends up as `C:\UsersfilipeDesktoptest.txt`
- In other cases it ends up weirder, particularly when using the Packaged app instead of the local development, where it ends up adding the `AppData\Local\insomnia\app-10.2.0` path to the full bad path.
image

image

image
</details>

Closes INS-4742

Closes #5754

### DIFF
--- a/packages/insomnia/src/templating/utils.ts
+++ b/packages/insomnia/src/templating/utils.ts
@@ -328,6 +328,6 @@ export interface nunjucksTagContextMenuOptions extends Exclude<ReturnType<typeof
 
 export const responseTagRegex = new RegExp('{% *response *.* %}');
 
-export function sanitizeStrForWin32(path: string) {
-  return path.replace(/\\/g, '\\\\\\\\');
+export function sanitizeStrForWin32(str: string) {
+  return str.replace(/\\/g, '\\\\\\\\');
 }
--- a/packages/insomnia/src/templating/utils.ts
+++ b/packages/insomnia/src/templating/utils.ts
@@ -327,3 +327,7 @@ export interface nunjucksTagContextMenuOptions extends Exclude<ReturnType<typeof
 }
 
 export const responseTagRegex = new RegExp('{% *response *.* %}');
+
+export function sanitizeStrForWin32(path: string) {
+  return path.replace(/\\/g, '\\\\\\\\');
+}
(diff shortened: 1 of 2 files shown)


gatzjames (approved):
Tested on Windows and macOS LGTM

PR #8223, opened 2024-12-03
fix(cli): misc issues to keep order in running cli runner

ihexxa:
### Changes
- [x] UI should always generate runner command with workspace id
- [x] CLI should not reject runner command with workspace and `-i` args
- [x] Sort requests by path when sorting with manual order
- [x] Support running folder for cli
- [x] `getRequestsToRunFromListOrWorkspace` doesn't include the root folder itself. 

Ref: INS-4764, INS-4763


notjaywu (approved):
lgtm

PR #8232, opened 2024-12-05
fix(cli): failed to run script in cli if timeline dir is not inited

ihexxa:
### Changes:
- [x] Init the path for persisting timeline file before executing script
- [x] Updated smoke tests

Ref: INS-4774


notjaywu (approved):
lgtm

PR #8243, opened 2024-12-11
fix(sdk): support async function in insomnia.test

ihexxa:
### Changes
- [x] Monitor all tests in the script and suspending the script execution until all sync tests or async tests are finished or rejected
- [x] Test it 

image

Ref: INS-4765

### DIFF
--- a/packages/insomnia/src/hidden-window.ts
+++ b/packages/insomnia/src/hidden-window.ts
@@ -39,8 +39,8 @@ window.bridge.onmessage(async (data, callback) => {
 
 // as insomnia.test accepts an async function, prepend await to it as user don't write it
 function translateTestHandlers(script: string): string {
-  const replacedTests = script.replace('insomnia.test(', 'await insomnia.test(');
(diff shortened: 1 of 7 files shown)


packages/insomnia/src/hidden-window.ts
```diff
@@ -37,6 +37,12 @@ window.bridge.onmessage(async (data, callback) => {
   }
 });
 
+// as insomnia.test accepts an async function, prepend await to it as user don't write it
+function translateTestHandlers(script: string): string {
+  const replacedTests = script.replace(/insomnia\.test\(/g, 'await insomnia.test(');
+  return replacedTests.replace(/insomnia\.test\.skip\(/g, 'await insomnia.test.skip(');
+}
```
cwangsmv, line 44: A minor issue here:
If user has already input await, it will translate the script to 
```
await await insomnia.test('xxx', () => {})
``` 
It's fine for me since double await could execute as expected, maybe we can enhance in the future

cwangsmv (approved):
LGTM, only a minor issue which could fix in later releases

PR #8249, opened 2024-12-12
fix: pre-request script variable hostname certificate resolution [INS-4733]

ryan-willis:
This fixes two distinct, but closely related issues:

When using a variable reference in the request URL for the hostname and a pre-request script with non-empty contents, client certs would be filtered using the un-rendered template text before sending the request and yield an error (part 1 of the new smoke test).

When gathering context to send a request that has a pre-request script with non-empty contents, the first cert in the list was always selected, disregarding the disabled flag and the hostname on the cert (part 2 of the new smoke test).

### DIFF
--- a/packages/insomnia-sdk/src/objects/insomnia.ts
+++ b/packages/insomnia-sdk/src/objects/insomnia.ts
@@ -168,20 +168,15 @@ export async function initInsomniaObject(
         localVars: localVariables,
     });
 
-    const reqUrl = toUrlObject(rawObj.request.url);
-    reqUrl.addQueryParams(
-        rawObj.request.parameters
-            .filter(param => !param.disabled)
-            .map(param => ({ key: param.name, value: param.value }))
-    );
+    // replace all variables in the raw URL before parsing it
+    const sanitizedRawUrl = `${rawObj.request.url}`.replace(/{{\s*\_\./g, '{{');
+    const renderedRawUrl = variables.replaceIn(sanitizedRawUrl);
 
-    // hack: sanitize a template reference if it exists in the hostname
-    // so we can filter certificates without a full render on the request
-    const host = reqUrl.getHost().replace(/{{\s*\_\./g, '{{');
-    const renderedHost = variables.replaceIn(host);
+    // parse the URL to get the host + protocol
+    const renderedUrl = toUrlObject(renderedRawUrl);
 
-    const renderedBaseUrl = toUrlObject(`${reqUrl.protocol}//${renderedHost}`);
-
+    // filter client certificates by the rendered base URL
+    const renderedBaseUrl = toUrlObject(`${renderedUrl.protocol || 'http:'}//${renderedUrl.getHost()}`);
     const filteredCerts = filterClientCertificates(rawObj.clientCertificates || [], renderedBaseUrl.toString());
     const existingClientCert = filteredCerts != null && filteredCerts.length > 0 && filteredCerts[0];
     const certificate = existingClientCert ?
@@ -201,6 +196,13 @@ export async function initInsomniaObject(
         rawObj.settings.httpsProxy,
         rawObj.settings.proxyEnabled,
         rawObj.settings.noProxy,
+    );
+
+    const reqUrl = toUrlObject(rawObj.request.url);
+    reqUrl.addQueryParams(
+        rawObj.request.parameters
+            .filter(param => !param.disabled)
+            .map(param => ({ key: param.name, value: param.value }))
     );
 
     const reqOpt: RequestOptions = {
--- a/packages/insomnia-sdk/src/objects/insomnia.ts
+++ b/packages/insomnia-sdk/src/objects/insomnia.ts
@@ -177,7 +177,7 @@ export async function initInsomniaObject(
 
     // hack: sanitize a template reference if it exists in the hostname
     // so we can filter certificates without a full render on the request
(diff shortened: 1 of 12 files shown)


marckong (dismissed):
lgtm!

marckong (dismissed):

packages/insomnia-sdk/src/objects/insomnia.ts
```diff
@@ -167,16 +168,31 @@ export async function initInsomniaObject(
         localVars: localVariables,
     });
 
-    const existClientCert = rawObj.clientCertificates != null && rawObj.clientCertificates.length > 0;
-    const certificate = existClientCert && rawObj.clientCertificates[0] ?
+    const reqUrl = toUrlObject(rawObj.request.url);
```
ihexxa: This makes sense to me to move this part from below, pls notice that the url parsing is not reliable for all cases as there could be templates in it (there's one card for it). such as representing both host and path with tag `{{_.srvr}}{{ _.path }}`, then this url will not be rendered as expected.

packages/insomnia-sdk/src/objects/insomnia.ts
```diff
@@ -167,16 +168,31 @@ export async function initInsomniaObject(
...
+    const renderedHost = variables.replaceIn(host);
+
+    const renderedBaseUrl = toUrlObject(`${reqUrl.protocol}//${renderedHost}`);
+
+    const filteredCerts = filterClientCertificates(rawObj.clientCertificates || [], renderedBaseUrl.toString());
+    const existingClientCert = filteredCerts != null && filteredCerts.length > 0 && filteredCerts[0];
+    const certificate = existingClientCert ?
         {
-            disabled: rawObj.clientCertificates[0].disabled,
+            disabled: existingClientCert.disabled,
```
ihexxa, line 184: You might notice that there seems no perfect way to transform it between 2 sides, I'm thinking if we could just leave the certificate undefined at the beginning, if user specified the cert in the script, we prepend to the cert list, or return the original certs, pls let me know what you think.

packages/insomnia-sdk/src/objects/insomnia.ts
```diff
@@ -167,16 +168,31 @@ export async function initInsomniaObject(
...
+            .filter(param => !param.disabled)
+            .map(param => ({ key: param.name, value: param.value }))
+    );
+
+    // hack: sanitize a template reference if it exists in the hostname
+    // so we can filter certificates without a full render on the request
+    const host = reqUrl.getHost().replace(/{{\s*\_\./g, '{{');
+    const renderedHost = variables.replaceIn(host);
+
+    const renderedBaseUrl = toUrlObject(`${reqUrl.protocol}//${renderedHost}`);
```
ihexxa: There could be one edge case here when user doesn't specify the protocol then `reqUrl.protocol` will be empty.

packages/insomnia-smoke-test/tests/smoke/mtls.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  // ensure disabling the cert actually disables it
+  await page.getByRole('button', { name: 'Add Certificates' }).click();
+  await page.locator('[data-test-id="client-certificate-toggle"]').click();
+  await page.getByRole('button', { name: 'Done' }).click();
+  await page.getByLabel('Request Collection').getByTestId('pet 2').press('Enter');
+
+  await page.getByRole('button', { name: 'Send', exact: true }).click();
+  await expect(statusTag).toContainText('401 Unauthorized');
+  await expect(responseBody).toContainText('Client certificate required');
+
```
ihexxa, line 72: 👍

ihexxa (commented):
Added some comments and feel free to let me know your thoughts, also thanks for looking into this complicated issue.

packages/insomnia-sdk/src/objects/insomnia.ts
ryan-willis: Good point. I tried to avoid running the url through a full render but it looks like we might need to do that to cover this case

packages/insomnia-sdk/src/objects/insomnia.ts
ryan-willis: I see. I'll default to `http`

packages/insomnia-sdk/src/objects/insomnia.ts
ryan-willis: Now that I'm looking at this again, I _could_ perform the template render hack prior to calling `getHost`, what do you think?

packages/insomnia-sdk/src/objects/insomnia.ts
ryan-willis: I'd much rather return original certs, but I'm thinking we revamp this later

ihexxa (approved):

packages/insomnia-sdk/src/objects/insomnia.ts
ihexxa: We also have a plan to improve the `replaceIn` method, let's optimize it later.

packages/insomnia-sdk/src/objects/insomnia.ts
ihexxa: Yes we can revamp it later and the prepend operation is for those cases who would like to update the cert through script.

PR #8256, opened 2024-12-18
fix: runner select all

CurryYangxx:
a keyboard shortcut `⌘ Cmd + A` can select all grid list items and let `selectedKeys` to be `all`, we need to handle this case


packages/insomnia/src/ui/routes/runner.tsx
```diff
@@ -476,10 +476,10 @@ export const Runner: FC<{}> = () => {
       ? Array.from(reqList.items)
         .filter(item => item.ancestorIds.includes(targetFolderId))
```
CurryYangxx: is this filter necessary?

ihexxa: It looks like we can drop it.

ihexxa (dismissed):
Thanks for the fix 👍 .

CurryYangxx:
will merge after 10.3

ihexxa (approved):

PR #8269, opened 2024-12-25
fix: editor did not unlock after websocket disconnect

### LINKED ISSUE #8265: URL lock occurs after WebSocket connection and cannot be unlocked until switching requests
### Expected Behavior

    Given I have a WebSocket request in Insomnia
    When I connect to the WebSocket server
    Then the URL input field should be locked and not editable
    And the URL input field should remain locked
    But the URL input field should unlock when the WebSocket connection is closed without switching to another request
    And I should be able to change the parameters in the URL input field after unlocking

### Actual Behavior

the URL input field remains locked until I switch to a different request

- Local only

### Insomnia Version

Version: Insomnia 10.3.0

macOS

### Operating System Version

15.1.1 (24B91)

### Installation method

download

idk (made a request in discussions for help)

CurryYangxx:
close #8265


cwangsmv (approved):
lgtm, update the editor readonly settings after readOnly property is changed

PR #8274, opened 2025-01-06
fix: smoke failed in github action

### LINKED ISSUE #42510: [Bug]: The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now.
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/electron/electron/blob/main/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/electron/electron/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://www.github.com/electron/electron/issues) for a bug report that matches the one I want to file, without success.

### Electron Version

30.1.0

Ubuntu

### Operating System Version

Kubuntu 24.04

### What arch are you using?

x64

### Last Known Working Electron version

Unknown

### Expected Behavior

Upgrading Kubuntu from `23.10` to `24.04` should not cause this error even after clearing `node_modules` and running `yarn cache clean`.

I also manually ensured there were no cached versions of Electron.

### Actual Behavior

This error persists and I cannot run our project.

### Testcase Gist URL

_No response_

### Additional Information

Only began after upgrading Kubuntu to 24.04.

CurryYangxx:
The smoke test action failed in the `ubuntu-latest` image: https://github.com/Kong/insomnia/actions/runs/12629124914/job/35186437717

So we fixed the ubuntu image version to `ubuntu-22.04` temporarily.
Consider reset after the issue is fixed: https://github.com/electron/electron/issues/42510

### DIFF
--- a/.github/workflows/release-recurring.yml
+++ b/.github/workflows/release-recurring.yml
@@ -30,7 +30,7 @@ jobs:
             build-targets: "zip"
           - os: "windows-latest"
             build-targets: "portable"
-          - os: "ubuntu-latest"
+          - os: "ubuntu-22.04"
             build-targets: "tar.gz"
     steps:
       - name: Checkout branch
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ concurrency:
 jobs:
   Test:
     timeout-minutes: 20
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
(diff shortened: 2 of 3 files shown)


ihexxa (approved):
We can update it back when our tests work for ubuntu 24.

PR #8068, opened 2024-10-11
fix: disable request error modal - [INS-4559]

CurryYangxx:
close [INS-4559](https://linear.app/insomnia/issue/INS-4559/disable-the-request-failure-alert-modal#comment-42c35ae3)

Changes:

- [x] disable request error alert modal (keep modal for collection runner), always show error in response preview panel
- [x] write script execution error into db and show error in response preview panel
- [x] fix smoke test

PR #8277, opened 2025-01-07
fix: request render error not display - [INS-4833]

CurryYangxx:
In https://github.com/Kong/insomnia/pull/8068, we removed the request error modal because some of the errors are also displayed in the preview panel.
But we ignored the request render error, so we restored the modal box and used it to display rendering errors.

### DIFF
--- a/packages/insomnia/src/ui/routes/request.tsx
+++ b/packages/insomnia/src/ui/routes/request.tsx
@@ -433,6 +433,7 @@ export const sendAction: ActionFunction = async ({ request, params }) => {
   } catch (err) {
     console.log('[request] Failed to send request', err);
     const e = err.error || err;
+    const url = new URL(request.url);
 
     // when after-script error, there is no error in response, we need to set error info into response, so that we can show it in response viewer
     if (err.response && err.requestMeta && err.response._id) {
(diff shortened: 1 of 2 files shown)


cwangsmv (approved):
Tried the changes and LGTM.

filfreire:
Tested locally, LGTM

image

filfreire (approved):

packages/insomnia/src/ui/routes/request.tsx
```diff
@@ -445,15 +446,16 @@ export const sendAction: ActionFunction = async ({ request, params }) => {
...
       const response = existingResponse || await models.response.create(err.response, err.maxHistoryResponses);
       await models.requestMeta.update(err.requestMeta, { activeResponseId: response._id });
+    } else {
+      // if the error is not from response, we need to set it to url param and show it in modal
+      url.searchParams.set('error', e);
+      if (e?.extraInfo && e?.extraInfo?.subType === RenderErrorSubType.EnvironmentVariable) {
+        url.searchParams.set('envVariableMissing', '1');
+        url.searchParams.set('undefinedEnvironmentVariables', e?.extraInfo?.undefinedEnvironmentVariables);
+      }
     }
```
jackkav, line 456: 🇮🇹 🍝

jackkav: Could this logic be cleaned up? Or expressed more simply.

CurryYangxx: Currently passing the error message through the URL parameter is not a good way. We could keep the status quo first. I plan to find a new way to replace this

PR #8289, opened 2025-01-10
fix: prevent unparseable response cookie from causing request to fail [INS-4851]

ryan-willis:
In order to prevent malformed cookies from throwing application errors for otherwise successful requests, this PR changes to loose cookie parsing so that a set-cookie header sent as `"=value"` will be transformed into `key: "", value: "value"`. We are choosing to retain RFC-compliant types so a missing key will be an empty string as opposed to `null` or `undefined`.

One caveat is: since the cookie is malformed, it will not be added to the cookie jar.


jackkav (commented):
insomnia cookies have parsing logic pre-dating this feature which behave differently if i recall correctly, its worth checking for deviation this code may have introduced.

jackkav (approved):

Issue #8240, opened 2024-12-09
[Bug] Unable to send File in multipart form request that includes pre-request script


eboozer:
### Expected Behavior

Multipart form fields should not be affected when adding pre-request script that does not touch request body.

### Actual Behavior

Multipart from-data does not include `filename` for a form field that includes file data.
* If the form-data field does NOT include a name, then the field is excluded in the final request.
* If the form-data field DOES include a name, then the field is included in the final request but it doesn't include the `filename`, nor any file data.
  * Note, the `filename` aspect is required for some APIs (like Ktor) to understand that the FormData field is actually a file, and not a normal form field.

### Reproduction Steps

1. Create new POST request
2. Set Body to use `multipart`
3. Add new form-data field for the body that includes a file
4. Set new pre-request script
5. Update the request's auth according to standard documentation:
```js
insomnia.request.auth.update(
  {
    type: 'bearer',
    bearer: [
      { key: 'token', value: 'XXX' },
      { key: 'prefix', value: 'Bearer' },
    ],
  },
  'bearer'
);
```
6. Send request
7. Note that the console outputs for the request structure

### Is there an existing issue for this?

- [X] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Which sync method do you use?

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [X] Local only

### Additional Information

The expected behavior should continue to work as it does as if there is no pre-request script on the request. The behavior does work fine without any script, and adding the auth update should not have any affect on the request body formdata inputs.

### Insomnia Version

v10.2.0

### What operating system are you using?

macOS

### Operating System Version

Darwin arm64 23.6.0

### Installation method

Homebrew

notjaywu:
@ihexxa to take a look

usmajeed:
also experiencing

eboozer:
Just following up on this post. I tried to be as detailed as I could, but please let me know if you have any issues recreating this issue. This seems like, at the very least, unexpected behavior, but I'm happy to test any alternative workflows if there is another way of accomplishing this. My end goal is simply to have the pre-request script set the authentication for a multi-part form request, so if there are additional steps to accomplish that, I'm happy to test them out to see if they work for my purposes.

ihexxa:
Thanks @eboozer the steps are detailed, will look into it.

ellyssonmike:
Same problem here.

After remove the request script it works.

[...]

PR #8297, opened 2025-01-15
fix(sdk): misc fixes and improvements

ihexxa:
Also with some minor fixes.

### Changes
- [x] Request Body: in transforming, included some request body properties, such as multiline, fileName and disabled
- [x] Console: support escape sequences such as `\n` and print params in one line (instead of multiple lines)
- [x] Reject setting `null` to environment
- [x] In test output, also show completed version of Actual value and Expected value, originally it is truncated.

### Multipart body
Screenshot 2025-01-15 at 15 36 01

### log output
Screenshot 2025-01-17 at 14 11 22

### In test, actual got value is 200, and expected value is a long string.
Screenshot 2025-01-16 at 17 10 35

Ref: INS-4795, #8240

### DIFF
--- a/packages/insomnia-sdk/src/objects/console.ts
+++ b/packages/insomnia-sdk/src/objects/console.ts
@@ -14,8 +14,11 @@ export class Console {
     // TODO: support replacing substitution
     printLog = (rows: Row[], level: LogLevel, ...values: any) => {
         try {
+            const content = values.map((a: any) => JSON.stringify(a, null, 2).replace('\\n', '\n'))
+                .join(' ');
             const row = {
-                value: `${level}: ` + values.map((a: any) => JSON.stringify(a, null, 2)).join('\n'),
+                // TODO: also support formating e.g. \t
+                value: `${level}: ${content}`,
                 name: 'Text',
                 timestamp: Date.now(),
             };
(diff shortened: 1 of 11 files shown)


packages/insomnia-sdk/src/objects/environments.ts
```diff
@@ -22,6 +23,10 @@ export class Environment {
     };
 
     set = (variableName: string, variableValue: boolean | number | string) => {
+        if (variableValue === null) {
```
CurryYangxx, line 26: The `varaibleValue` seems inconsistent with the type definition.

packages/insomnia-sdk/src/objects/console.ts
```diff
@@ -14,8 +14,11 @@ export class Console {
     // TODO: support replacing substitution
     printLog = (rows: Row[], level: LogLevel, ...values: any) => {
         try {
+            const content = values.map((a: any) => JSON.stringify(a, null, 2).replace('\\n', '\n'))
```
CurryYangxx: When will it include `\\n`?

packages/insomnia-sdk/src/objects/environments.ts
ihexxa: Good catch, updated.

packages/insomnia-sdk/src/objects/console.ts
ihexxa: `JSON.stringify` may transform `\n` into `\\n` in `a` (if it is a string).

CurryYangxx (approved):

PR #8310, opened 2025-01-22
Rename local-storage to electron-storage

cwangsmv:
**Changes:**
Currently, we have a `local-storage.ts` file used to persistent data in local file.
Since the name is confused with browser native `localStorage` in render process, rename it to electron-storage to make it more clear.

After this change, `localStorage` points to the browser native localStorage in render process.
`electron-storage` is the persistent storage to save data in local files in main process


CurryYangxx (approved):

jackkav (approved):
❤️

ryan-willis (approved):

PR #8314, opened 2025-01-23
Improve WSDL import [INS-4291]

yaoweiprc:
When importing WSDL files, the file may depend on other XSD files. We haven't parsed referenced XSD files before.
Now we pass the file path of the WSDL to apiconnect-wsdl to let it parse all referenced xsd files.

Also, there are restrictions that prevent it from generating a full sample request. It's hard-coded in apiconnect-wsdl, so we use patch-package to patch the files in apiconnect-wsdl to remove the restrictions
https://www.npmjs.com/package/patch-package

Finally, I added the logic to show the detailed error message within the importer as our customer requested.

### DIFF
--- a/packages/insomnia/src/utils/importers/entities.ts
+++ b/packages/insomnia/src/utils/importers/entities.ts
@@ -100,12 +100,12 @@ interface ContentStrImporter extends BaseImporter {
   convert: Converter;
 }
 
-interface FileImporter extends BaseImporter {
+interface FilePathImporter extends BaseImporter {
   acceptFilePath: true;
   convert: FilePathConverter;
 }
 
-export type Importer = ContentStrImporter | FileImporter;
+export type Importer = ContentStrImporter | FilePathImporter;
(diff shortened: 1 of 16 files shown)


packages/insomnia/src/utils/importers/entities.ts
```diff
@@ -79,13 +79,36 @@ export interface ImportRequest<T extends {} = {}> extends Comment {
...
   name: string;
   description: string;
+}
+
+interface ContentStrImporter extends BaseImporter {
+  acceptFilePath?: false;
   convert: Converter;
 }
+
+interface FileImporter extends BaseImporter {
```
CurryYangxx: FilePathImporter maybe a better name instead of FileImporter

CurryYangxx:
I'd suggest adding a readme file in `patches` dir to explain the background and how to use it.

CurryYangxx (approved):

PR #8335, opened 2025-02-05
feat: preference for key map with one-line text fields

ryan-willis:
(due to my own personal reeling from a one-line vim cursor)

key map will be disabled when editing a `literal` (one-line: key/value, url), and only apply to `code` (multi-line: body, script) by default

users can turn it back on in preferences


jackkav (dismissed):
Looks good, but since I think it pairs with another setting it's a good time to confirm that the default is reasonable. I don't use a custom key map but if I did I guess I would only expect it to work in multi line inputs unless I toggle something on. I might be mistaken on this but worth considering before naming the something hard to change.

ryan-willis:
> I would only expect it to work in multi line inputs unless I toggle something on - @jackkav 

I agree, and I'm leaning toward that as well. I feel that enrolling all users into a default preference that changes the UX would better suit the next major version, so perhaps we should wait on this one until then?

CurryYangxx (dismissed):

jackkav (approved):

PR #8338, opened 2025-02-06
fix: template render error [INS-4962]

filfreire:
This PR closes INS-4962. Changes:

- A "dirty" fix to the issue - follows same approach as https://github.com/Kong/insomnia/blob/b7e959758bc012e102730266931fa2ceeb031970/packages/insomnia/src/ui/components/codemirror/extensions/nunjucks-tags.ts#L283-L288
- Reproduction of the bug added to `pre-request-script-window` smoke test to help test a better a fix in the future.

## context

The bug was introduced after `10.3.0` due to the sanitization code we added in order to solve another bug (see https://github.com/Kong/insomnia/commit/90bd25c55f1b8fa9e9505cd4f0fd44e8e685da4b#diff-28ee43a50d28cabb98317967663a3abd59d3e11c9fafa39bca88eed9d290c0b6R173). 

Something about the `variables.replaceIn(sanitizedRawUrl)` call that happens on the insomnia sdk does not work well with this particular format of template tags `{{ _['host_environment'] }}`.

## notes

One important thing to note, is that the code-path to render template tags in `insomnia-sdk` is a different one than the one we use on the app (for example for rendering the templates in the preview of the URL )

- The app we use this render function: https://github.com/Kong/insomnia/blob/b7e959758bc012e102730266931fa2ceeb031970/packages/insomnia/src/common/render.ts#L262
- But in the SDK we use this one: https://github.com/Kong/insomnia/blob/b7e959758bc012e102730266931fa2ceeb031970/packages/insomnia-sdk/src/objects/interpolator.ts#L11

We might want to consider merging both to make use of old edge case fixes we already have done before the sdk.

Another side-note is that to debug this issue is a bit tricky as it requires:
1. start hidden browser window
2. send a request
3. show hidden browser window
4. going to dev tools of hidden browser window

image

### DIFF
--- a/packages/insomnia-sdk/src/objects/insomnia.ts
+++ b/packages/insomnia-sdk/src/objects/insomnia.ts
@@ -170,7 +170,11 @@ export async function initInsomniaObject(
 
     // replace all variables in the raw URL before parsing it
     const sanitizedRawUrl = `${rawObj.request.url}`.replace(/{{\s*\_\./g, '{{');
-    const renderedRawUrl = variables.replaceIn(sanitizedRawUrl);
+    let renderedRawUrl = sanitizedRawUrl;
+
+    // FIXME: hack to solve INS-4962, remove underscores, single quotes & square brackets
+    const tmpSanitizedRawUrl = sanitizedRawUrl.replace(/[_'[\]]/g, '');
+    renderedRawUrl = variables.replaceIn(tmpSanitizedRawUrl);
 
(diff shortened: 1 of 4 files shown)


ryan-willis (dismissed):

gatzjames (dismissed):
✅

gatzjames (dismissed):

gatzjames (approved):

PR #8341, opened 2025-02-07
Render no-standard content-encoding response [INS-4910]

yaoweiprc:
[https://linear.app/insomnia/issue/INS-4910/support-custom-content-encoding](url)
One of our customers reported that we couldn't render the response, the issue stems from the http response using the non-standard Content-Encoding header 'base64'. Currently, the libcurl library we use supports decompressing the response body with deflate, gzip, and br algorithms.

When the Content-Encoding response header uses a value other than deflate, gzip, or br, libcurl throws an error: "Unrecognized or bad HTTP Content or Transfer-Encoding", with error code 61.

At present, when libcurl throws the "Unrecognized or bad HTTP Content or Transfer-Encoding" error, we cannot retrieve the response body content.

To achieve the same behavior as Postman, we would first send a request. Upon encountering the "Unrecognized or bad HTTP Content or Transfer-Encoding" error, we would set ACCEPT_ENCODING in libcurl to null and resend the request. This way, libcurl will no longer throw the "Unrecognized or bad HTTP Content or Transfer-Encoding" error, and we can retrieve the compressed response body and Try to render it.

### DIFF
--- a/packages/insomnia/src/main/network/libcurl-promise.ts
+++ b/packages/insomnia/src/main/network/libcurl-promise.ts
@@ -29,6 +29,8 @@ export interface CurlRequestOptions {
   caCertficatePath: string | null;
   socketPath?: string;
   authHeader?: { name: string; value: string };
+  // make libcurl not decompress the response content
+  noDecompress?: boolean;
 }
 interface RequestUsedHere {
   headers: any;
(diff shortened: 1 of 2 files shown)


packages/insomnia/src/main/network/libcurl-promise.ts
```diff
@@ -260,22 +269,29 @@ export const createConfiguredCurlInstance = ({
...
   socketPath,
+  noDecompress = false,
 }: {
   req: RequestUsedHere;
   finalUrl: string;
   settings: SettingsUsedHere;
   certificates: ClientCertificate[];
   caCert: string | null;
   socketPath?: string;
+    noDecompress?: boolean;
```
notjaywu: ident

packages/insomnia/src/main/network/libcurl-promise.ts
```diff
@@ -225,6 +228,12 @@ export const curlRequest = (options: CurlRequestOptions) => new Promise<CurlRequ
       curl.isOpen && curl.close();
       await waitForStreamToFinish(responseBodyWriteStream);
 
+      // If libcurl can't decompress the response, retry without decompression
+      if (code === CurlCode.CURLE_BAD_CONTENT_ENCODING) {
```
CurryYangxx: avoid Infinite retry
if (code === CurlCode.CURLE_BAD_CONTENT_ENCODING && noDecompress === false) {}

CurryYangxx (approved):

PR #8354, opened 2025-02-11
Prettify json in response viewer [INS-4966]

### LINKED ISSUE #1556: [Enhancement] Friendlier JSON in Preview mode
It looks like Insomnia always escapes the forward slash:
`\/`

Can this be an option in the response window, or the IDE as global option?

This bit us, as a very long encoded value was behaving one way in Insomnia (escaping the forward slash), but was different in another REST client (Postman: it doesn't escape forward slashes).

Thanks!

yaoweiprc:
Close #1556 

1. Use native method JSON.stringify to prettify json before it's passed to CodeEditor, which can solve the problem of forward slash escape mentioned in the issue.
2. Use useCallback hook to wrap functions in ResponseViewer component, to avoid creating new functions every time the component renders.

### DIFF
--- a/packages/insomnia/src/ui/components/viewers/response-viewer.tsx
+++ b/packages/insomnia/src/ui/components/viewers/response-viewer.tsx
@@ -1,6 +1,7 @@
 import iconv from 'iconv-lite';
 import React, {
   Fragment,
+  useCallback,
   useRef,
   useState,
 } from 'react';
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/components/viewers/response-viewer.tsx
```diff
@@ -235,6 +239,40 @@ export const ResponseViewer = ({
...
+    contentType === 'application/json'
+  ) {
+    let bodyStr = getBodyAsString();
+    // Although there is a prettifier for json inside the CodeEditor, but it is to prettify json strings that is being edited which may have syntax errors.
+    // There are some cases that the prettifier inside the CodeEditor can not handle.
+    // See https://github.com/Kong/insomnia/issues/1556
+    // Here the CodeEditor is readonly and the bodyStr is supposed to be a valid json string.
+    // So we try to use the native JSON.stringify to prettify the json string better. The native way can handle the issue.
+    try {
+      bodyStr = JSON.stringify(JSON.parse(bodyStr));
```
CurryYangxx, line 254: Why do we need to parse first and then stringify?

yaoweiprc: To remove the extra forward slash from the response.
See https://github.com/Kong/insomnia/issues/1556

CurryYangxx (approved):

PR #8358, opened 2025-02-11
Short-ciruit string rendering if require invocation is detected

godfrzero:
This PR adds a temporary means of preventing string rendering if a require invocation is detected. This has been tested locally and verified to be functional (enough) against strings in the URL, body, and headers but should cover all flows which render strings with the exception of ones introduced via plugins.

Closes SEC-1323


jackkav (approved):

filfreire (approved):
Tested locally, LGTM

PR #8360, opened 2025-02-12
Fix table view environment not updated issue[INS-4875]

cwangsmv:
**Changes:**
Fix Environment key-value editor content is not updated when user uses script to update environment variable including:
- Collection environment
- Global environment
- Folder environment


ihexxa (approved):
🚀

Issue #8357, opened 2025-02-11
Fail on Script processing


cburgosro9303:
### Expected Behavior

when I have a request with an After-response script it should execute correctly.

### Actual Behavior

when I have a request with an After-response script the execution of the script fails

<img width="1328" alt="Image" src="https://github.com/user-attachments/assets/8cb460f5-c9f2-4013-96dc-b247d0491af2" />

### Reproduction Steps

1. create a request
2. add After-response script 
3. execute de request

<img width="1328" alt="Image" src="https://github.com/user-attachments/assets/14c9fb8b-c3b6-4ce8-b82d-8e8150160bd9" />

### Is there an existing issue for this?

- [x] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Which sync method do you use?

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [ ] Local only

### Insomnia Version

10.3.1

### What operating system are you using?

Windows

### Operating System Version

Windows 11 26100.3037

### Installation method

update

### Last Known Working Insomnia version

10.3.0

ebektasiadis:
Same here, seems that `.response` is missing from `insomnia`.

masonwolff:
Same here- started happening when I upgraded to latest version earlier today

yshavit:
Happens for pre-request scripts, too.

lebster:
I'm experiencing this as well. It seems like the issue might be related to env vars in the URL. Removing them fixed it for me.

yshavit:
> I'm experiencing this as well. It seems like the issue might be related to env vars in the URL. Removing them fixed it for me.

Confirmed that it fixed it for me, too.

jnazaren:
I'm experiencing this as well since the update to 10.3.1.

vamaral-kno2:
I'm experiencing the same behavior using 10.3.1

yochail:
removing after-response scripts solve this for me as a temp workaround

ryan-willis:
a fix for this is now available in the beta release channel: https://github.com/Kong/insomnia/releases/tag/core%4010.3.2-beta.0

PR #8361, opened 2025-02-12
Fix/sdk/tagged url

### LINKED ISSUE #8363: Response -> Body Attribute feature does not work
### Expected Behavior

Ability to use `Response -> Body Attribute` feature:

Image

### Actual Behavior

No way to use `Response -> Body Attribute` feature, it breaks collection with error:

Image

---
Error is:
message: undefined; stack: Error: message: (unknown path) [Line 1, Column 31]
  unknown block tag: response; stack: Template render error: (unknown path) [Line 1, Column 31]
  unknown block tag: response
    at Object.P [as _prettifyError] (file:///snap/insomnia/305/resources/app.asar/sentry-RNTkbMCO.js:5:39127)
    at ie.render (file:///snap/insomnia/305/resources/app.asar/sentry-RNTkbMCO.js:19:6753)
    at ie.renderString (file:///snap/insomnia/305/resources/app.asar/sentry-RNTkbMCO.js:19:4938)
    at M5.render (file:///snap/insomnia/305/resources/app.asar/sentry-RNTkbMCO.js:76:16068)
    at $V.replaceIn (file:///snap/insomnia/305/resources/app.asar/sentry-RNTkbMCO.js:89:1270)
    at vre (file:///snap/insomnia/305/resources/app.asar/sentry-RNTkbMCO.js:89:10187)
    at h (file:///snap/insomnia/305/resources/app.asar/hiddenBrowserWindow-Cz0sW5QM.js:1:747)
    at file:///snap/insomnia/305/resources/app.asar/hiddenBrowserWindow-Cz0sW5QM.js:1:374
    at port.onmessage (/snap/insomnia/305/resources/app.asar/hidden-window-preload.js:128695:36)
    at port.onmessage (/snap/insomnia/305/resources/app.asar/preload.js:92:18)

---
It's defined correctly: 

Image

### Reproduction Steps

1. Have a collection with `Response -> Body attribute` feature  (so at least 2 requests with passing data between them using it)
2. Execute it

- [ ] Git sync.
- [x] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

- How do I know that this is the problem:

Hardcoded uuid, instead of using `Response -> Body attribute` feature works fine:

Image

- It's blocker for me, cannot use Insomnia at all in this moment

### Insomnia Version

10.3.1

Ubuntu

### Operating System Version

Ubuntu 22.04.5 LTS

### Installation method

insomnia.rest

10.3.0

ihexxa:
fixes #8363, #8357

### DIFF
--- a/packages/insomnia-smoke-test/fixtures/pre-request-collection.yaml
+++ b/packages/insomnia-smoke-test/fixtures/pre-request-collection.yaml
@@ -326,7 +326,7 @@ resources:
     parentId: fld_01de564274824ecaad272330339ea6b2
     modified: 1636707449231
     created: 1636141014552
-    url: http://127.0.0.1:4010/echo
+    url: "http://127.0.0.1:4010/echo"
     name: test certificate manipulation
     description: ""
     method: POST
@@ -344,10 +344,10 @@ resources:
     settingFollowRedirects: global
     preRequestScript: |-
       // print the original one
-      console.log('key:', insomnia.request.certificate.key.src);
-      console.log('cert:', insomnia.request.certificate.cert.src);
+      console.log('key:', insomnia.request.certificate.key);
+      console.log('cert:', insomnia.request.certificate.cert);
       console.log('passphrass:', insomnia.request.certificate.passphrass);
-      console.log('pfx:', insomnia.request.certificate.pfx.src);
+      console.log('pfx:', insomnia.request.certificate.pfx);
       // update
       insomnia.request.certificate.update({
           disabled: true,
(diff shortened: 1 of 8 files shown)


packages/insomnia-smoke-test/fixtures/pre-request-collection.yaml
```diff
@@ -357,6 +357,40 @@ resources:
...
+    isPrivate: false
+    settingStoreCookies: true
+    settingSendCookies: true
+    settingDisableRenderRequestBody: false
+    settingEncodeUrl: true
+    settingRebuildPath: true
+    settingFollowRedirects: global
+    preRequestScript: |-
+      insomnia.environment.set('serverURL', 'http://127.0.0.1:4010');
+      insomnia.clientCertificates = []; // remove all existing ones or it will still be selected
```
ryan-willis, line 411: I don't think any users that are calling `insomnia.request.certificate.update` in a pre-request script are going to know that this line is required in order for that call to work as it once did

ihexxa: Really good catch 👍 , this part contains a lot of details and there's a a reason behind it for testing, it seems that if the test just updates it with fake key and cert, so finally it will be ignored and fallback to the original one. If the user specified a valid one, this line should not be required. Ideally we should call `.update` with a valid cert, it requires a valid cert path while it requires some additional effort to get it for different envs. Pls correct me if you detected something wrong.

ryan-willis (approved):

ihexxa:
Let me merge firstly this to unblock users, and improve readability afterwards.

vamaral-kno2:
When will the version with this fix be created?

ryan-willis:
@vamaral-kno2 this fix is now available in the beta release channel

PR #8362, opened 2025-02-12
Spike: remove part of multiple abstraction-[INS-4911]

CurryYangxx:
Remove useRequestRunnerList and useInsomniaTab abstraction


CurryYangxx:
@jackkav I removed part of the abstraction and put the logic back to the same file. Hope this can help you understand the logic better.

PR #8372, opened 2025-02-17
fix: do not append headers with empty names

ryan-willis:


marckong (dismissed):

ihexxa (dismissed):
Probably also trim names before checking.

ihexxa (approved):

PR #8375, opened 2025-02-18
fix: reduce sse

CurryYangxx:
The `organizations` change in every render because of the react-router loader revalidation, causing repeated sse requests.
So we get it from localstorage directly.


cwangsmv (approved):
lgtm

PR #8377, opened 2025-02-18
feat(Git Sync): move git logic in git-service

gatzjames:
Highlights:

- [x] Git logic is now running in main

Next steps:

- [ ] There is some functionallity like continueMerge that needs to be moved in actions/loaders for consistency and state synchronization


CurryYangxx:
After this change, is it still necessary to run the git related logic in the loader or action? For example, we only call the ipc in this `gitRepoLoader`. 
```
export const gitRepoLoader: ActionFunction = async ({
  params,
}): Promise<GitRepoLoaderData> => {
  return window.main.git.loadGitRepository({ projectId, workspaceId });
};
```

Currently we have this process: 
```
create a fetcher -> run fetcher.load -> call IPC function in the loader -> get new data from fetcher.data
```
Could we call the IPC function directly in UI like:
```
const res = await window.main.git.loadGitRepository({ projectId, workspaceId });
```
This looks more intuitive, less code(we can get the result through 1 line), and we put the logic in just one place.

gatzjames:
> After this change, is it still necessary to run the git related logic in the loader or action? For example, we only call the ipc in this `gitRepoLoader`.
> 
> ```
> export const gitRepoLoader: ActionFunction = async ({
>   params,
> }): Promise<GitRepoLoaderData> => {
>   return window.main.git.loadGitRepository({ projectId, workspaceId });
> };
> ```
> 
> Currently we have this process:
> 
> ```
> create a fetcher -> run fetcher.load -> call IPC function in the loader -> get new data from fetcher.data
> ```
> 
> Could we call the IPC function directly in UI like:
> 
> ```
> const res = await window.main.git.loadGitRepository({ projectId, workspaceId });
> ```
> 
> This looks more intuitive, less code(we can get the result through 1 line), and we put the logic in just one place.

The loader and action helps us manage fetching -> loading/error/result states in a consistent way.
Replacing it with our own abstraction is possible but then we would need:
- States for error, loading and data
- Manual revalidation logic (we would need to know exactly what changed after each mutation and reload the portion of the data that got affected
- Manual synchronization logic, e.g. running the same mutation many times can cause ordering issues if the second mutation result arrives first
- Another approach to loading data in the future for React 19 (client/server components, not recommended to fetch data on useEffects etc.)

While it looks simpler to do `await doAsyncStuff() -> store result in state` the moment we want to handle all these extra cases I think the code is much simpler using routing and eliminates categories of bugs that we would have to manually handle in the other case.

filfreire (approved):

PR #8378, opened 2025-02-19
Chore: Fix vault case failure and add one more test for vault key

cwangsmv:
**Changes:** 
- Fix the flaky case of the newly added vault related smoke add
- Add one more test case for vault key check
- Move check enterprise logic for one-line-editor only when subscription key matches the editor id

### DIFF
--- a/packages/insomnia-smoke-test/tests/smoke/insomnia-vault.test.ts
+++ b/packages/insomnia-smoke-test/tests/smoke/insomnia-vault.test.ts
@@ -5,6 +5,20 @@ import { test } from '../../playwright/test';
 
 test.describe('Test Insomnia Vault', async () => {
 
+  test('check vault key display and can copy', async ({ page, app }) => {
+    await page.locator('[data-testid="settings-button"]').click();
+    await page.locator('text=Insomnia Preferences').first().click();
+    // get text under div with data-testid="vault-key"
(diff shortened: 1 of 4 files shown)


gatzjames (approved):

packages/insomnia-smoke-test/tests/smoke/insomnia-vault.test.ts
```diff
@@ -5,6 +5,20 @@ import { test } from '../../playwright/test';
...
 test.describe('Test Insomnia Vault', async () => {
 
+  test('check vault key display and can copy', async ({ page }) => {
+    await page.locator('[data-testid="settings-button"]').click();
+    await page.locator('text=Insomnia Preferences').first().click();
+    // get text under div with data-testid="vault-key"
+    const expectedVaultKeyValue = 'eyJhbGciOiJBMjU2R0NNIiwiZXh0Ijp0cnVlLCJrIjoia';
+    const vaultKeyValue = await page.getByTestId('VaultKeyDisplayPanel').innerText();
+    await expect(vaultKeyValue).toContain(expectedVaultKeyValue);
+    await page.getByTitle('Copy Vault Key').click();
```
gatzjames, line 15: Are there other ways for this check without using the clipboard api?

filfreire (approved):

PR #8388, opened 2025-02-20
fix(sdk): change insomnia.test mechanism to improve the nested test cases

ihexxa:
### Changes
- [x] Change the mechanism of `insomnia.test` to improve the nested test cases
- [x] Support `await insomnia.test`
- [x] Enable autocomplete for `insomnia.test` and `insomnia.expect`
- [x] Added smoke tests

Add sync or async cases in script to test it:

```javascript
          insomnia.test('asyncHappyTestInFunc', () => {
            insomnia.expect(200).to.eql(200);
          });

          insomnia.test('asyncHappyTestInFunc', async () => {
            insomnia.expect(200).to.eql(200);
          });
```
INS-5005

### DIFF
--- a/packages/insomnia-sdk/src/objects/insomnia.ts
+++ b/packages/insomnia-sdk/src/objects/insomnia.ts
@@ -82,6 +82,22 @@ export class InsomniaObject {
 
         this.requestTestResults = new Array<RequestTestResult>();
         this.parentFolders = rawObj.parentFolders;
+
+        return new Proxy(this, {
+            get: (target, prop, receiver) => {
+                if (prop === 'test') {
+                    const testHandler: TestHandler = async (msg: string, fn: () => Promise<void>) => {
+                        await this._test(msg, fn, this.pushRequestTestResult);
+                    };
+                    testHandler.skip = async (msg: string, fn: () => Promise<void>) => {
+                        await this._skip(msg, fn, this.pushRequestTestResult);
+                    };
+
+                    return testHandler;
+                }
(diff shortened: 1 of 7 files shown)


yaoweiprc (commented):
Could there be any compatibility issues since the function passed to test() is executed asynchronously now?

ihexxa:
> Could there be any compatibility issues since the function passed to test() is executed asynchronously now?

@yaoweiprc This is a good question, theoretically, for those who pass function without `async`, as those tests are still executed synchronously, so the behavior should not change. for those who pass function with `async`, originally their code returned an error and didn't work, then the currently patch should work.

cwangsmv:
Nit: Why don't we expose `insomnia.test` `insomnia.expect` method in snippets?
I see there's logic making test/expect as private property and do not expose them in `getRequestScriptSnippets` method while postman has hints for `pm.test` `pm.exepct`.

ihexxa:
> Nit: Why don't we expose `insomnia.test` `insomnia.expect` method in snippets? I see there's logic making test/expect as private property and do not expose them in `getRequestScriptSnippets` method while postman has hints for `pm.test` `pm.exepct`.

@cwangsmv Good catch, it just tries to wrap `_expect` and `_test` so that we can add some manual validation, let me tinker it a bit to see if I can expose them.

ihexxa:
Updated PR with following changes:
- Support `await insomnia.test`
- Enable autocomplete for `insomnia.expect` and `insomnia.test`
- Added a test

cwangsmv (approved):
lgtm

PR #8284, opened 2025-01-08
fix: paths in gRPC requests and proto state

ryan-willis:
To test the majority of this, follow the steps in my comment [here](https://github.com/Kong/insomnia/issues/7744#issuecomment-2578764796)

Changes:

- Makes proto files de-selectable in the dialog
- Fixes request switching losing method list (closes #6963)
- Adds path prefix support (closes #5203, closes #7744)
- Adds user-friendly error messages for common issues (closes #6380)
- Switches to using [Kong/grpc-reflection-js](https://github.com/Kong/grpc-reflection-js)

PR #8410, opened 2025-02-25
fix: unable to use unix socket with grpc - [INS-5036]

### LINKED ISSUE #8387: Unable to use unix sockets with gRPC
### Expected Behavior

I can use a path like `unix:///path/to/my/socket.sock` and it successfully connects to the socket

### Actual Behavior

A "URL not specified" error is presented.

### Reproduction Steps

1. Start a gRPC server on a unix socket
2. Create a gRPC request in Insomnia with the url pointing to the socket file
3. Try to make the request

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [x] Local only

### Additional Information

Unix sockets seem to work fine for HTTP requests, so I presume this is a bug but if not please help me redirect 🙏 

### Insomnia Version

10.3.1

macOS

### Operating System Version

macOS Sequoia 15.3

### Installation method

Downloaded from the website

_No response_

CurryYangxx:
Changes:
- close #8387 

Reason:
The unix URL is different from the normal grpc URL. For example: `unix:///path/to/my/socket.sock`,  the protocol is `unix`, and the host should be the whole URL. There is no pathname in this URL.
So the `new URL()` method is not suitable for handling unix grpc request.

### DIFF
--- a/packages/insomnia/src/network/grpc/parse-grpc-url.ts
+++ b/packages/insomnia/src/network/grpc/parse-grpc-url.ts
@@ -2,6 +2,14 @@ export const parseGrpcUrl = (grpcUrl: string): { url: string; enableTls: boolean
   if (!grpcUrl) {
     return { url: '', enableTls: false, path: '' };
   }
+
+  if (grpcUrl.includes('unix:')) {
+    return {
+      url: grpcUrl,
+      enableTls: false,
+      path: '',
+    };
+  }
   const url = new URL((grpcUrl.includes('://') ? '' : 'grpc://') + grpcUrl.toLowerCase());
   return {
     url: url.host,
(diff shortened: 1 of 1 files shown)


CurryYangxx:
@ryan-willis I noticed that this seems related to this pr(https://github.com/Kong/insomnia/pull/8284). Please help to take a look if this is an appropriate approach to handling this unix situation.

CurryYangxx:
[unix-grpc.zip](https://github.com/user-attachments/files/18961505/unix-grpc.zip)
This a js script for starting a test unix socket grpc server locally.

ryan-willis (dismissed):

packages/insomnia/src/network/grpc/parse-grpc-url.ts
```diff
@@ -2,6 +2,14 @@ export const parseGrpcUrl = (grpcUrl: string): { url: string; enableTls: boolean
   if (!grpcUrl) {
     return { url: '', enableTls: false, path: '' };
   }
+
+  if (grpcUrl.includes('unix:')) {
```
notjaywu: Should it use `startsWith('unix:')` instead?

yaoweiprc (approved):

PR #8411, opened 2025-02-25
feat(Git Projects): Update storage type

gatzjames:
Highlights:
- [x] UX for creating/updating a project with all possible storage options (git/cloud/local)
- [x] Display warnings when switching storage type
- [x] Update logic in update action to handle switching to/from git projects


pavkout (approved):
Awesome work @gatzjames 🎸 LGTM

PR #8419, opened 2025-02-26
feat: Create checkbox UI Component

pavkout:
image

Closes INS-5035


godfrzero:
Let's wait until after v11 is shipped to merge additional changes, by the way.

packages/insomnia/src/ui/components/base/indeterminate-checkbox.tsx
```diff
@@ -1,25 +0,0 @@
...
-  checked: boolean;
-}
-
-export const IndeterminateCheckbox: FC<Props> = ({ checked, indeterminate, ...otherProps }) => {
-  const checkRef = useRef<HTMLInputElement>(null);
-
-  useEffect(() => {
-    if (checkRef.current) {
-      checkRef.current.checked = checked;
-      checkRef.current.indeterminate = indeterminate;
```
marckong, line 14: how do we handle indeterminate status with the new checkbox?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

packages/insomnia/src/ui/components/checkbox.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { Checkbox as RACheckbox, type CheckboxProps } from 'react-aria-components';
+
+type Props = {
+  children: React.ReactNode;
+} & CheckboxProps;
+
+export const Checkbox = ({ children, ...props }: Props) => {
```
marckong, line 8: We might want to pass a ref object here in case if it needs a handle to be controlled by its parent?

yaoweiprc (approved):

packages/insomnia/src/ui/components/base/indeterminate-checkbox.tsx
pavkout: @marckong We never used this component.

PR #8425, opened 2025-02-27
Fix issue of the typo file

cwangsmv:
**Changes**
- Fix duplicate file with typo file naming caused by rebase
- Remove the duplicate file and fix the wrong file importing


ihexxa:
It seems that multi-tab is not enabled when running the smoke test, while it seems unrelated to this change.

cwangsmv:
> It seems that multi-tab is not enabled when running the smoke test, while it seems unrelated to this change.

Rebased and all smoke test are passed

ihexxa (approved):

PR #8427, opened 2025-02-27
fix: disable arch for linux on electron-builder [INS-5053]

filfreire:
Closes INS-5053


notjaywu (approved):
lgtm, to test on EE as well

hexchain:
May I know the reason for this change, and when the arm64 artifact will probably be added back?

PR #8442, opened 2025-03-07
Solve the bug related to importing multiple workspaces in one file [INS-5077]

### LINKED ISSUE #8424: Import of Collections results in  Error: New Requests missing `parentId`
### Expected Behavior

The insomnia import tool ought to have imported the entire collection set without throwing any errors.

### Actual Behavior

The import fails with the message. `Error: New Requests missing 'parentId'`. Noticed that in the JSON file, the parentId for "workspace" resources types was null. Tried to replace the `"parentId": null` with `"parentId": "proj_scratchpad"` but that didn't help.

### Reproduction Steps

1. Launched Insomnia 10.3.1 on my corporate laptop having Windows 11
2. Logged into Insomnia account
3. Exported the entire collections under the sole insomnia project. There are ~20 workspaces. A ~76 MB JSON file was created
4. The laptop was reset/reimaged
5. Reinstalled Insomnia after laptop was re-imaged.
6. Launched Insomnia 10.3.1 and logged in
7. Proceeded to import the Collection

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [x] Local only

### Additional Information

Image
Image
Image
Image
Image
Image

### Insomnia Version

10.3.1

Windows

### Operating System Version

Windows 11 Enterprise 24H2, OS Build 26100.3194, Windows Feature Experience Pack 1000.26100.48.0

### Installation method

download from insomnia site

10.3.1

yaoweiprc:
Close #8424 
If the user imports a file containing more than one workspace, we create a new workspace for each workspace in the file. Then, all resources in the file are added to all newly created workspaces, regardless of whether they belong to the workspace.
That is why the 'Error: New Requests missing parentId' showed up.

Now we check whether a resource belongs to a workspace before it's added to the workspace.

### DIFF
--- a/packages/insomnia-smoke-test/tests/smoke/import.test.ts
+++ b/packages/insomnia-smoke-test/tests/smoke/import.test.ts
@@ -1,14 +1,16 @@
-import path from 'node:path';
-
 import { expect } from '@playwright/test';
 
+import { loadFixture } from '../../playwright/paths';
 import { test } from '../../playwright/test';
 
-test('can import multiple workspaces from single file', async ({ page }) => {
+test('can import multiple workspaces from single file', async ({ app, page }) => {
+  const text = await loadFixture('multiple-workspaces.yaml');
+  await app.evaluate(async ({ clipboard }, text) => clipboard.writeText(text), text);
   await page.getByLabel('Import').click();
-  await page.locator('[data-test-id="import-file-input"]').setInputFiles(path.join(__dirname, '../..', 'fixtures', 'multiple-workspaces.yaml'));
+  await page.locator('[data-test-id="import-from-clipboard"]').click();
   await page.getByRole('button', { name: 'Scan' }).click();
   await page.getByRole('dialog').getByRole('button', { name: 'Import' }).click();
+  // Have two collections in current project
   await expect(page.getByLabel('Collection 1')).toBeAttached();
   await expect(page.getByLabel('Collection 2')).toBeAttached();
 });
--- a/packages/insomnia-smoke-test/tests/smoke/import.test.ts
+++ b/packages/insomnia-smoke-test/tests/smoke/import.test.ts
@@ -1,0 +1,14 @@
+import path from 'node:path';
+
+import { expect } from '@playwright/test';
+
+import { test } from '../../playwright/test';
+
+test('can import multiple workspaces from single file', async ({ page }) => {
+  await page.getByLabel('Import').click();
+  await page.locator('[data-test-id="import-file-input"]').setInputFiles(path.join(__dirname, '../..', 'fixtures', 'multiple-workspaces.yaml'));
+  await page.getByRole('button', { name: 'Scan' }).click();
+  await page.getByRole('dialog').getByRole('button', { name: 'Import' }).click();
+  await expect(page.getByLabel('Collection 1')).toBeAttached();
+  await expect(page.getByLabel('Collection 2')).toBeAttached();
+});
--- a/packages/insomnia/src/common/import.ts
+++ b/packages/insomnia/src/common/import.ts
@@ -239,7 +239,7 @@ export async function importResourcesToProject({ projectId }: { projectId: strin
       continue;
     }
 
-    const workspaceResources = resources.filter(isWorkspace);;
+    const workspaceResources = resources.filter(isWorkspace);
 
     // No workspace, so create one
     if (workspaceResources.length === 0) {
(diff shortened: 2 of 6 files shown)


ryan-willis (dismissed):
the filtering/traversal code is a bit if a brain teaser, but I trust you

packages/insomnia/src/common/import.ts
```diff
@@ -239,19 +239,69 @@ export async function importResourcesToProject({ projectId }: { projectId: strin
...
+    await Promise.all(workspaceResources
+      .map(workspace => {
+        if (workspaceResources.filter(({ _id }) => _id === '__WORKSPACE_ID__').length > 1) {
+          console.warn(`There are more than one workspace with id __WORKSPACE_ID__ in the resources, the importer is ${resourceCacheItem.importer.name}`);
+        }
+        // Here if there is only one workspace in the resources, we import all resources to it
+        let resourcesInCurrentWorkspace = resources;
+        // If there are more than one workspace in the resources, we filter in the resources that belong to the current workspace
+        if (workspaceResources.length > 1) {
+          resourcesInCurrentWorkspace = filterResourcesInWorkspace(resources, workspace);
```
CurryYangxx, line 260: We can try to use a one-time traverse to find the resources belong to each workspace

filfreire (dismissed):

packages/insomnia-smoke-test/tests/smoke/import.test.ts
```diff
@@ -0,0 +1,14 @@
+import path from 'node:path';
+
+import { expect } from '@playwright/test';
+
+import { test } from '../../playwright/test';
+
+test('can import multiple workspaces from single file', async ({ page }) => {
+  await page.getByLabel('Import').click();
+  await page.locator('[data-test-id="import-file-input"]').setInputFiles(path.join(__dirname, '../..', 'fixtures', 'multiple-workspaces.yaml'));
```
gatzjames: Using the file api tends to create some flakiness in tests.
I'd suggest to use the loadFixture api instead and use the clipboard if possible

packages/insomnia/src/common/import.ts
```diff
@@ -239,19 +239,69 @@ export async function importResourcesToProject({ projectId }: { projectId: strin
       continue;
     }
 
+    const workspaceResources = resources.filter(isWorkspace);;
```
gatzjames:
```suggestion
    const workspaceResources = resources.filter(isWorkspace);
```

packages/insomnia/src/common/import.ts
gatzjames: As a next step I think we should handle all these edge cases using schemas for all the importers.
We can embed the validation logic into the schema and apply transformations as required.
The final result would be a workspace or series of workspaces with resources or plain resources that we need to import into a workspace.
This will drastically simplify all this logic and create a convention on importing data to the app

packages/insomnia/src/common/import.ts
```diff
@@ -239,19 +239,69 @@ export async function importResourcesToProject({ projectId }: { projectId: strin
...
+    if (track.has(id)) {
+      return id;
+    }
+    track.add(id);
+    const parentId = idToParentIdMap.get(id);
+    if (!parentId) {
+      return id;
+    }
+    return findRootId(parentId, track);
+  }
```
gatzjames:
```suggestion
  function findRootId(id: string, existingResourceIds: Set<string>) {
    // avoid infinite loop
    if (existingResourceIds.has(id)) {
      return id;
    }
    existingResourceIds.add(id);
    const parentId = idToParentIdMap.get(id);
    if (!parentId) {
      return id;
    }
    return findRootId(parentId, existingResourceIds);
  }
```

packages/insomnia/src/common/import.ts
```diff
@@ -450,19 +499,24 @@ const importResourcesToNewWorkspace = async (
     const model = getModel(resource.type);
 
     if (model) {
+      const newParentId = ResourceIdMap.get(resource.parentId);
+      if (!newParentId) {
+        console.warn(`Could not find new parent id for ${resource.name} ${resource._id}`);
```
gatzjames, line 504: Does this mean that this resource will not be imported?
If so what will happen if it is a resource that contains other resources e.g. requestGroup?

packages/insomnia-smoke-test/tests/smoke/import.test.ts
yaoweiprc: @gatzjames Yeah I found that we use clipboard in other test cases.

packages/insomnia/src/common/import.ts
yaoweiprc: Since we have filtered the resources before this function 'importResourcesToNewWorkspace' is called. Only resources that belong to the new workspace are in resourceCacheItem. So each resource here is supposed to have a valid parentId. I do a double-check here because the methods we use in the following codes to insert a record to the database like 'models.grpcRequest.create', 'models.request.create' will throw errors if the parentId is undefined.

packages/insomnia/src/common/import.ts
yaoweiprc: @CurryYangxx Yes, but let's leave it as is for now, since performance is not an urgent concern here.

yaoweiprc: @gatzjames I agree.

gatzjames (approved):

PR #8451, opened 2025-03-13
securely wrap execution

ryan-willis:
resolves INS-5048

what this does:

- builds the normal Electron executable into `insomnia.dll` (so that it can be code-signed)
- builds a wrapper into `Insomnia.exe` (also code-signed)
- when the wrapper runs, makes a copy of `insomnia.dll` called `insomnia-$VERSION.exe` and executes it with custom Windows process mitigation policies
- when Insomnia exits, the wrapper deletes the `insomnia-$VERSION.exe` file
- standard output is piped directly

this is meant to be a stop-gap for CVE-2025-1353 and will very likely be superseded by other measures (NSIS/MSI installing into protected directories, etc)

to test locally, run `./build-secure-wrapper.sh` from a bash-like shell on Windows with MSYS2/MinGW installed, then run `packages/insomnia/dist/squirrel-windows/Insomnia.Core-X.X.X.exe` to install Insomnia.
everything should run normally when launching/installing/uninstalling

this installer launches cleanly on a fresh Windows 11 as well

references:
https://chromium.googlesource.com/chromium/src.git/+/master/docs/security/faq.md#why-arent-physically_local-attacks-in-chromes-threat-model

https://learn.microsoft.com/en-us/windows/security/threat-protection/overview-of-threat-mitigations-in-windows-10#table-4functions-available-to-developers-for-building-mitigations-into-apps


marckong (approved):
LGTM. We tested it together in a call and the fix seems to behave as expected.

PR #8456, opened 2025-03-13
feat: allow choosing install dir with NSIS - [INS-5109]

ryan-willis:
## Support new NSIS Windows installer

Changes
- package NSIS installer
  - `electron builder config`: Add new Windows NSIS target and NSIS-related config
  - NSIS install script (`nsisInstall.nsh`)
    - Uninstall the existing Squirrel app
    - Inject a file to mark nsis installer so that we can distinguish the current installer and use a different auto-update strategy
- auto-update for NSIS (`nsisUpdate.ts`)
  - introduce `electron-updater` package
  - Add `publish` config in `electron-builder-config.js`
- secure wrapper
  - Use different wrapper strategy for the NSIS installer, because we cannot do a file copy operation when the user installs to `/Program files` dir.
- 

**Auto Update strategy**
(the `electron.autoUpdater` we use for squirrel dose not support nsis, so we need to introduce electron-updater pkg)
image

**Secure Wrapper**
image


CurryYangxx:
@ryan-willis seems it is already able to detect existing insomnia.

packages/insomnia/src/cpp/insomnia.cpp
```diff
@@ -196,32 +196,81 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
...
+
+  HANDLE hFile = ::CreateFileW(installJsonPath.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+  if (hFile == INVALID_HANDLE_VALUE) {
+    ::DebugLog(L"installer-info.json not found or cannot be opened.");
   } else {
-    // if it's a directory, then exit and prompt the user to uninstall
-    if (attrs != INVALID_FILE_ATTRIBUTES) {
-      ::DebugLog(L"File is a directory, exiting.");
-      return ::ExitWithWarning(nCmdShow, L"Insomnia installation is corrupted. Please reinstall.");
+    DWORD fileSize = ::GetFileSize(hFile, NULL);
```
notjaywu: Is it NSIS installer when this file exists?

notjaywu (commented):
left 1 comment

build-secure-wrapper.sh
```diff
@@ -58,7 +59,7 @@ if [ ! $BUILD_CONTEXT ]; then
   npm run package:windows:dist -w insomnia
 
   echo "Resetting to prevent accidental loops..."
-  mv $DEST_DIR/insomnia.dll $DEST_DIR/Insomnia.exe
+  mv $DEST_DIR/Insomnia.dll $DEST_DIR/Insomnia.exe
```
notjaywu: Why rename `insomnia.dll` to `Insomnia.dll`?

build-secure-wrapper.sh
```diff
@@ -32,6 +32,7 @@ if [ ! $BUILD_CONTEXT ]; then
 fi
 
 cp $DEST_DIR/Insomnia.exe $DEST_DIR/insomnia.dll
+cp $DEST_DIR/Insomnia.exe $DEST_DIR/Insomnia-origin.exe
```
notjaywu: I'd suggest naming it as `Insomnia-{version}.exe`, to keep it the same as what we do for Squirrel.

packages/insomnia/src/cpp/insomnia.cpp
CurryYangxx: Yes. I think we don't need to read the file content here.

build-secure-wrapper.sh
CurryYangxx: This is a typo...

notjaywu (approved):

PR #8486, opened 2025-03-19
feat: temporary tabs - [INS-5088]

CurryYangxx:
**Background**

> Add an option not to open a new tab when clicking on a folders (i have to many folders and need to close tabs... boring)
(https://github.com/Kong/insomnia/discussions/6108#discussioncomment-12263700)

support temporary tabs to avoid creating too many useless tabs.

**Scope**
only for Request Collection

**Highlights**

- Click on request/folder in the sidebar, create a temporary tab with a different style(italic font style).
- only one temporary tab exists, click another request will replace the existing temporary tab.
- double click & edit request & send request will convert the temporary tab to a permanent one.
- create new request will create a permanent tab

https://github.com/user-attachments/assets/743900af-6444-4a4b-a1c8-a19c89a29f84


LunaticoCR:
@CurryYangxx  If I may suggest, this is an improvement but still not a solution for many of us that don't want to use tabs and think one of Insomnia's strengths is that it has no tabs, which removes a lot of clutter.
I think opening tabs on double click is perfectly fine, but opening them on edit or send request is not, in my view. Or ano option to allow users to switch them off completely.

yaoweiprc (approved):

Issue #8507, opened 2025-03-21
Unable to scroll on cookies page with Version: 11.0.0


rajiv2814:
### Expected Behavior

Recently there was upgrade for us to this version 11.0.0. We cannot edit or scroll on cookies window. This is problem for us as we have setup multiple cookies for our different environments. Happening to other folks with this version as well. We cannot even search our cookies.

### Actual Behavior

Version upgraded to 11.0.0

### Reproduction Steps

1. Go to Manage cookies
2. Add new cookie and enter details
3. Click on Add new cookie and scroll on the cookies modal.  scroll does not work.

### Is there an existing issue for this?

- [x] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Which sync method do you use?

- [ ] Git sync.
- [x] Insomnia Cloud sync.
- [ ] Local only

### Insomnia Version

11.0.0

### Operating System Version

macOS Sequoia Version 15.3.2 (24D81)

### Installation method

brew

pavkout:
@rajiv2814 Thank you for reporting this issue. We will fix it ASAP

pavkout:
@rajiv2814 We have fixed this issue in 11.0.1-beta.0

rajiv2814:
I don't think I will be able to use the beta version in our org.. Is there any new version which is published? I can ask our devops team to upgrade this if it's ready..

PR #8508, opened 2025-03-22
fix: Fixed cookie modal issue [INS-5210] 

pavkout:
image

Closes INS-5210, [8507](https://github.com/Kong/insomnia/issues/8507)

### DIFF
--- a/packages/insomnia/src/ui/components/modals/cookies-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/cookies-modal.tsx
@@ -135,7 +135,6 @@ export const CookiesModal = ({ setIsOpen }: Props) => {
     updateCookieJar(cookieJar._id, cookieJar);
   };
 
-  console.log(filteredCookies);
   return (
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/components/modals/cookies-modal.tsx
```diff
@@ -107,7 +97,16 @@ export const CookiesModal = ({ setIsOpen }: Props) => {
 
   const handleAddCookie = () => {
     const updatedActiveCookieJar = activeCookieJar;
```
CurryYangxx, line 99: `const updatedActiveCookieJar = {...activeCookieJar};`
It would be better to copy the origin object data first rather than modify props directly.

pavkout: @CurryYangxx Yes, we can do it. Let's do it in a separate PR and check all the places we need to change.

gatzjames (approved):

PR #8535, opened 2025-03-27
chore(Formatting): Use prettier

gatzjames:
Highlights:

- [x] Based on top [of the oxlint PR](https://github.com/Kong/insomnia/pull/8534) to verify there are no formatting conflicts
- [x] Had to also update prettier on packages/insomnia to avoid version conflicts.
  - [ ] We use it for formatting things inside the app so this needs some testing
- [x] Added the tailwind plugin to make sure classNames are sorted correctly
  - [ ] This could lead to style changes if we had className collisions so we need to verify this
- [x] Added prettier as the default formatter in vscode settings and recommended extension
- [x] Applied formatting to the whole codebase 🫡


gatzjames:
Done in another pass

PR #8551, opened 2025-03-31
fix(Git Sync): discard changes should delete requests and other entities if they don't exist in the previous version

gatzjames:
### The issue:

Inside a Git Project, when a user creates a request in a file and then discards the changes that have been made in that file, any entities e.g. new request that were created are not properly removed from the db.
This means that a user will chose to discard changes but those changes will persist.

Highlights:

- [x] When git updates a file we clean the previous workspace and it's descendants and use the resources provided by that version of the file.

Closes INS-5245


pavkout (approved):
LGTM

PR #8606, opened 2025-04-14
refactor(sdk): only display error message without stacktrace for minor errors

ihexxa:
### Change
- [x] Throw a string message instead an error -- Originally I introduced a custom error to do this, but after discussed within team, we plan to avoid introducing error abstraction too early.
- [x] Reformat outputs into message and stack.

Without stack:
image

With stack:
image

### DIFF
--- a/packages/insomnia-scripting-environment/src/objects/environments.ts
+++ b/packages/insomnia-scripting-environment/src/objects/environments.ts
@@ -1,5 +1,4 @@
 import { getExistingConsole } from './console';
-import { toMinorScriptError } from './interfaces';
 import { getInterpolator } from './interpolator';
 
 export class Environment {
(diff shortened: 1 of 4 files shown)


packages/insomnia-scripting-environment/src/objects/environments.ts
```diff
@@ -155,28 +156,28 @@ export class Vault extends Environment {
...
   };
 
   clear = () => {
-    throw new Error('Vault can not be cleared in script');
+    throw toMinorScriptError(new Error('Vault can not be cleared in script'));
   };
 
   set = () => {
-    throw new Error('Vault can not be set in script');
+    throw toMinorScriptError(new Error('Vault can not be set in script'));
```
jackkav:
```suggestion
    throw 'Vault can not be set in script';
```

If your goal is to avoid a stack trace from being included in the error.message. This approach should work without adding any new abstractions.

image

ihexxa: Good point, let's discuss this later.

ihexxa: Updated code and description as discussed.

jackkav (approved):

PR #8634, opened 2025-04-22
chore: optimize some interfaces & fix typo

ZxBing0066:
### Changes
- [x] Fix some typos
- [x] Code clean
- [x] Optimize some interface definition

### DIFF
--- a/packages/insomnia-scripting-environment/vitest.config.ts
+++ b/packages/insomnia-scripting-environment/vitest.config.ts
@@ -1,0 +1,7 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    hideSkippedTests: true,
+  },
+});
--- a/packages/insomnia/src/models/index.ts
+++ b/packages/insomnia/src/models/index.ts
@@ -17,6 +17,7 @@ import {
   EXPORT_TYPE_WORKSPACE,
 } from '../common/constants';
 import { generateId } from '../common/misc';
+import { typedKeys } from '../utils';
 import * as _apiSpec from './api-spec';
 import * as _caCertificate from './ca-certificate';
 import * as _clientCertificate from './client-certificate';
@@ -217,15 +218,13 @@ export async function initModel<T extends BaseModel>(type: string, ...sources: R
   // If we put those keys in init method, all related models will show as modified in git sync.
   const modelOptionalKeys: string[] = 'optionalKeys' in model ? model.optionalKeys || [] : [];
   // Prune extra keys from doc
-  for (const key of Object.keys(migratedDoc)) {
+  for (const key of typedKeys(migratedDoc)) {
     if (!(key in objectDefaults) && !modelOptionalKeys.includes(key)) {
-      // @ts-expect-error -- mapping unsoundness
       delete migratedDoc[key];
     }
   }
 
-  // @ts-expect-error -- TSCONVERSION not sure why this error is occurring
-  return migratedDoc;
+  return migratedDoc as T;
 }
 
 export const MODELS_BY_EXPORT_TYPE: Record<string, any> = {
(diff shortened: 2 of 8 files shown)


packages/insomnia-scripting-environment/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vitest/config';
```
ZxBing0066, line 1: When I use the official Vitest VSCode extension, it doesn't list the tests in `insomnia-scripting-environment`, because it cannot find the config file. Add a simple one to solve this issue.

packages/insomnia/src/models/index.ts
```diff
@@ -217,15 +218,13 @@ export async function initModel<T extends BaseModel>(type: string, ...sources: R
   // If we put those keys in init method, all related models will show as modified in git sync.
   const modelOptionalKeys: string[] = 'optionalKeys' in model ? model.optionalKeys || [] : [];
   // Prune extra keys from doc
-  for (const key of Object.keys(migratedDoc)) {
+  for (const key of typedKeys(migratedDoc)) {
```
ZxBing0066, line 221: Use a helper to solve the type inference issue.

packages/insomnia/src/models/index.ts
```diff
@@ -217,15 +218,13 @@ export async function initModel<T extends BaseModel>(type: string, ...sources: R
...
     if (!(key in objectDefaults) && !modelOptionalKeys.includes(key)) {
-      // @ts-expect-error -- mapping unsoundness
       delete migratedDoc[key];
     }
   }
 
-  // @ts-expect-error -- TSCONVERSION not sure why this error is occurring
-  return migratedDoc;
+  return migratedDoc as T;
```
ZxBing0066, line 227: `as` should be better than the magic comment.

packages/insomnia/src/network/cancellation.ts
```diff
@@ -92,11 +87,11 @@ export const cancellableCurlRequest = async (requestOptions: CurlRequestOptions)
   }
 };
 
-export const cancellablePromise = ({ signal, fn }: { signal: AbortSignal; fn: Promise<any> }) => {
+export const cancellablePromise = <T = any>({ signal, fn }: { signal: AbortSignal; fn: Promise<T> }) => {
```
ZxBing0066, line 88: Enhance the type inference of `cancellablePromise`, now it can automatically get the return type from `fn`.

packages/insomnia/src/network/cancellation.ts
```diff
@@ -54,12 +54,7 @@ export const cancellableRunScript = async (options: { script: string; context: R
       fn: process.type === 'renderer' ? window.main.hiddenBrowserWindow.runScript(options) : nodejsRunScript(options),
     });
 
-    return result as {
-      request: Request;
-      environment: object;
-      baseEnvironment: object;
-      cookieJar: CookieJar;
-    };
+    return result;
```
ZxBing0066, line 55: The result should be `RequestContext`, remove the unnecessary `as`.

packages/insomnia/src/network/network.ts
```diff
@@ -482,7 +482,7 @@ export const tryToExecuteScript = async (context: RequestAndContextAndOptionalRe
     if ('error' in originalOutput) {
       return { error: `Script executor returns error: ${originalOutput.error}` };
     }
-    const output = originalOutput as RequestContext;
+    const output = originalOutput;
```
ihexxa: nit: probably we can just name originalOutput as output at line 438.

ihexxa (dismissed):
Good job and thanks for sorting out these 👍.

packages/insomnia/src/network/network.ts
ZxBing0066: @ihexxa Good catch! I've renamed it, please have another review. 🙏

ihexxa (approved):
🚀

PR #8639, opened 2025-04-23
fix: Enforce stricter validation rules for the plugin name. [INS-5310 & INS-5408 & INS-5333]

pavkout:
Enforce stricter validation rules for the plugin name.

### Empty state - without plugins
image

### With Installed plugin list
image

### Initial View
image

### Error exists
image

Closes: INS-5310, INS-5310


packages/insomnia-smoke-test/tests/smoke/plugins-interactions.test.ts
```diff
@@ -1,17 +1,74 @@
...
+import { expect } from '@playwright/test';
 
 test.describe('Plugins', () => {
-  test('Open plugins menu and generate plugin', async ({ page }) => {
+  test.beforeEach(async ({ page }) => {
     // Opening settings
-    await page.getByTestId("settings-button").click();
+    await page.getByTestId('settings-button').click();
     // Switching to Plugins tab
     await page.locator('div[role="tab"]:has-text("Plugins")').click();
```
jackkav: An improved test would combine these into a single test and avoid using a before each

packages/insomnia/src/plugins/create.ts
```diff
@@ -2,22 +2,61 @@ import electron from 'electron';
...
+  }
+
+  // 2. Sanitize moduleName to remove any unexpected characters or sequences
+  const sanitizedModuleName = moduleName.replace(/\.\.(\/|\\)/g, ''); // Remove '../' or path traversal attempts
+
+  // 3. Get base directory
+  const baseDir = path.resolve(
+    process.env['INSOMNIA_DATA_PATH'] ||
+    (process.type === 'renderer' ? window : electron).app.getPath('userData'),
+    'plugins'
```
jackkav: you can avoid a lot of this complexity by comparing the encodeURI output as they do in this package https://github.com/npm/validate-npm-package-name/blob/main/lib/index.js

packages/insomnia/src/ui/components/settings/plugins.tsx
```diff
@@ -282,6 +250,15 @@ export const Plugins: FC = () => {
...
         </Button>
+        {showCreatePluginModal &&
+          <CreatePluginModal
+            onClose={() => setShowCreatePluginModal(false)}
+            onComplete={() => {
+              setShowCreatePluginModal(false)
+              refreshPlugins()
+            }}
+          />
+        }
```
jackkav: nice job with the plugin modal

packages/insomnia/src/plugins/create.ts
pavkout: @jackkav I kept some of the checks because we need to cover cases like this
`insomnia-plugin--` 

And this is true
`encodeURIComponent('insomnia-plugin--') === 'insomnia-plugin--' `

godfrzero:
Let's maintain the warning-type styling for enabling elevated plugin access. This has a significant impact on plugin access and the visual styling should emphasize that it's not a checkbox to be ticked lightly.

image

packages/insomnia-smoke-test/tests/smoke/plugins-interactions.test.ts
```diff
@@ -1,17 +1,89 @@
...
+  // Reject plugin name with consecutive dashes
+  await page.locator('text=New Plugin').click();
+
+  for (const { name, expectedError } of invalidNames) {
+    await page.getByTestId('plugin-name-input').fill(name);
+    await page.getByTestId('generate-plugin-button').click();
+
+    await expect.soft(page.getByTestId('plugin-name-error')).toBeVisible();
+    await expect.soft(page.getByTestId('plugin-name-error')).toHaveText(expectedError);
+  }
```
jackkav, line 79: very cool

packages/insomnia/src/__tests__/install-plugin.test.ts
```diff
@@ -0,0 +1,378 @@
...
+    normalize: vi.fn(path => path),
+    sep: '/',
+  },
+  resolve: vi.fn((...args) => '/mock/app/path/' + args.join('/')),
+  dirname: vi.fn(() => '/mock/app/path'),
+  join: vi.fn((...args) => args.join('/')),
+  normalize: vi.fn(path => path),
+  sep: '/',
+}));
+
```
jackkav: this looks like hard work to write hard work to read, its fine but low ROI.
I'd encourage refactoring code to make tests easier to read and write, not a blocker.

jackkav (dismissed):

pavkout:
> Let's maintain the warning-type styling for enabling elevated plugin access. This has a significant impact on plugin access and the visual styling should emphasize that it's not a checkbox to be ticked lightly.
> 
> image

@godfrzero I have changed it like this 

image

packages/insomnia-smoke-test/tests/smoke/plugins-interactions.test.ts
pavkout: 🎸

packages/insomnia/src/__tests__/install-plugin.test.ts
pavkout: @jackkav Let me check if it's possible to simplify it a bit before we merge it.

packages/insomnia/src/utils/plugin.ts
```diff
@@ -0,0 +1,159 @@
...
+
+// Pattern for common shell metacharacters
+const unsafeShellPattern = /[|;&$`\\]/;
+
+export function validatePluginName(pluginName: string, allowScopedPackageNames = false): string | null {
+  const pluginNameWithoutPrefix = pluginName.replace(/^insomnia-plugin-/, '');
+
+  // Check the length of the plugin name
+  // Plugin name must be less than 250 characters
+  if (pluginNameWithoutPrefix.trim().length === 0 || pluginNameWithoutPrefix.length > 250) {
```
godfrzero: How did we arrive at the 250 character limit? It looks like npm doesn't allow names longer than 214 characters, inclusive of the scope. Ref: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#name

pavkout: @godfrzero I had come across 250 as a "safe" limit in some repo or forum and went with it just to be on the safe side, without cross-checking the official npm docs. Thanks for pointing it out, I've updated it to align with the 214 character limit.

gatzjames (approved):

PR #8654, opened 2025-04-28
fix: update cookieJar methods to keep the name and params consistent

ZxBing0066:
## Background

When using a template to get the cookie, it throws an error.
```js
{% request 'cookie', 'test', '' %}
```
iShot_2025-04-28_17 21 16

## Root cause

There are two issues:
- We mostly use `cookieJar.getOrCreateForParentId`, but use `getOrCreateForWorkspace` in `templating-worker-database.ts` to match the message.
- `templating-worker-database.ts` uses `body.id` as the parent id, but the real param name used in `base-extension-worker.ts` is `body.parentId`.

## Changes

- Unify all the `getOrCreateForWorkspace` to `getOrCreateForParentId` to avoid similar issues.
- Change the `getOrCreateForParentId` param from workspace to parentId to match the new name
- Fix the wrong used parameter


cwangsmv (approved):
lgtm

PR #8660, opened 2025-04-29
refactor: default hide the swagger preview

marckong:
- we want to hide the swagger panel by default until enabled

### DIFF
--- a/packages/insomnia/src/ui/routes/design.tsx
+++ b/packages/insomnia/src/ui/routes/design.tsx
@@ -197,7 +197,7 @@ const Design: FC = () => {
   const updateApiSpecFetcher = useFetcher();
   const generateRequestCollectionFetcher = useFetcher();
   const [isLintPaneOpen, setIsLintPaneOpen] = useState(false);
-  const [isSpecPaneOpen, setIsSpecPaneOpen] = useState(true);
+  const [isSpecPaneOpen, setIsSpecPaneOpen] = useState(false);
 
   const { components, info, servers, paths } = parsedSpec || {};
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/routes/design.tsx
```diff
@@ -197,7 +197,7 @@ const Design: FC = () => {
   const updateApiSpecFetcher = useFetcher();
   const generateRequestCollectionFetcher = useFetcher();
   const [isLintPaneOpen, setIsLintPaneOpen] = useState(false);
-  const [isSpecPaneOpen, setIsSpecPaneOpen] = useState(true);
+  const [isSpecPaneOpen, setIsSpecPaneOpen] = useState(false);
```
gatzjames: Based on the details of the ticket I think we should check if the spec has contents and decide on what the default state is.
```suggestion
  const [isSpecPaneOpen, setIsSpecPaneOpen] = useState(Boolean(parsedSpec));
```

gatzjames (commented):
Not sure if the Preview feature will be discoverable by new users if we default to disabling it in all cases.
Maybe we should display the Preview when there is a spec available

godfrzero:
> Not sure if the Preview feature will be discoverable by new users if we default to disabling it in all cases. Maybe we should display the Preview when there is a spec available

Yes, let's only hide the Preview by default if there's no specification to render. There is, of course, a discoverability problem we need to address but we can pick that up as a separate item.

ryan-willis (approved):

ryan-willis (approved):

PR #8665, opened 2025-04-30
Fix smoke test flaky cases in slower runners

cwangsmv:
**Changes**
Fix flaky smoke test cases, especially in slower runner.

### DIFF
--- a/packages/insomnia-smoke-test/tests/smoke/insomnia-tab.test.ts
+++ b/packages/insomnia-smoke-test/tests/smoke/insomnia-tab.test.ts
@@ -28,6 +28,8 @@ test.describe('multiple-tab feature test', () => {
     await page.getByRole('textbox', { name: 'GET New Request' }).fill('foo');
     await page.getByLabel('Create in collection').click();
     await page.getByLabel('HTTP Request').click();
+    // check new request created
+    await page.getByTestId('New Request').isVisible();
     await page.getByTestId('New Request').dblclick();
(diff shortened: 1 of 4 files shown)


packages/insomnia-smoke-test/tests/smoke/insomnia-tab.test.ts
```diff
@@ -28,6 +28,8 @@ test.describe('multiple-tab feature test', () => {
     await page.getByRole('textbox', { name: 'GET New Request' }).fill('foo');
     await page.getByLabel('Create in collection').click();
     await page.getByLabel('HTTP Request').click();
+    // check new request created
+    await page.getByTestId('New Request').isVisible();
```
ZxBing0066, line 32: Just curious about this line, as I know `page.getByTestId('New Request')` will wait for the element to exist, why do we need to use `isVisible()` here? Is there anything that causes the element to be created but cannot be clicked? 🤔

ZxBing0066 (dismissed):
👍

packages/insomnia-smoke-test/tests/smoke/insomnia-tab.test.ts
cwangsmv: From the trace file screenshot,```await page.getByTestId('New Request').dblclick();``` clicks on the wrong element. So add a ```isVisible ``` first to ensure that the newly created element appears on UI before double click.
Screenshot 2025-04-30 at 12 49 08

ZxBing0066 (approved):

PR #8677, opened 2025-05-06
feat: support to set cookie hostOnly property

### LINKED ISSUE #8653: "__Host-" prefixed cookies are not send
### Expected Behavior

The cookie is attached and sent successfully 

### Actual Behavior

When adding a cookie which has a prefix of `__Host-` it is not added to the request. I tried to remove the `__Host_xxxx` at the cookie name and it is attached normally

logs:

```
* Enable cookie sending with jar of 3 cookies
* Found bundle for host: 0x12406dc2440 [can multiplex]
* Re-using existing connection #20 with host xxxxxxx
* Connected to xxxx.com (x.x.x.x) port 443 (#20)
* h2h3 [:method: GET]
* h2h3 [:path: /api/xxx/v2/xxxx]
* h2h3 [:scheme: https]
* h2h3 [:authority: xxxxxxx]
* h2h3 [cookie: authToken=xxxxx; GCP_IAP_UID=xxxx]
```

### Reproduction Steps

1. Create a new cookie with the prefix of `__Host-`
2. Click send with cookies attached
3. The cookie is not send

- [ ] Git sync.
- [x] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

_No response_

### Insomnia Version

11.0.2

macOS

### Operating System Version

macOS Sequoia 15.3.1 (24D70)

### Installation method

download

_No response_

ZxBing0066:
## Background

When a cookie has a `__Host-` prefix, it cannot be sent with the request.

## Root Cause

- A cookie with the `__Host-` prefix is special; it can be sent when Secure is true, path is `/`, and when the cookie is hostOnly.
- When a cookie is set without a domain, its host-only-flag is true
- A cookie should be set to a specific URL, but in Insomnia, we don't have such a URL, so we reuse the domain attribute in the cookie as the identifier. It works for most cases, but with this design, we cannot create a hostOnly cookie since we cannot omit a cookie's domain, otherwise we can't know which request needs this cookie.

## References

- [Cookie Name Prefixes](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis/#name-cookie-name-prefixes)
- [host-only-flag](https://datatracker.ietf.org/doc/html/rfc6265#section-5.3)
- [MDN Cookies hostOnly](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/Cookie#hostonly)

## Changes

Add `hostOnly` control to the cookie panel

Closes #8653

### DIFF
--- a/packages/insomnia-smoke-test/tests/smoke/cookie-editor-interactions.test.ts
+++ b/packages/insomnia-smoke-test/tests/smoke/cookie-editor-interactions.test.ts
@@ -58,4 +58,31 @@ test.describe('Cookie editor', () => {
     await page.getByRole('tab', { name: 'Console' }).click();
     await expect.soft(page.getByText('foo2=bar2')).toBeVisible();
   });
+
+  test('support __Host- prefix', async ({ page }) => {
+    // Open cookie editor
+    await page.click('button:has-text("Cookies")');
+
+    // Create a new cookie
+    await page.getByRole('button', { name: 'Add Cookie' }).click();
+
+    // Edit the new cookie
+    await page.getByRole('button', { name: 'Edit' }).first().click();
+    await page.getByText('HostOnly').click();
+    await expect.soft(page.locator('input[name="hostOnly"]')).toBeChecked();
+    await page.getByRole('tab', { name: 'Raw' }).click();
+    await page
+      .locator('text=Raw Cookie String >> input[type="text"]')
+      .fill('__Host-foo=bar; Expires=Tue, 19 Jan 2038 03:14:07 GMT; Secure; Domain=localhost; Path=/');
+    await page.locator('text=Done').nth(1).click();
+    await page.click('text=Done');
+
+    // Send request
+    await page.getByLabel('Request Collection').getByTestId('example http').press('Enter');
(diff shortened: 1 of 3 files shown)


ZxBing0066:
To solve the hostOnly issue, we have two solutions:
1. Add a URL input to the cookie edit panel, so that we know the cookie will be set to which host, and the domain field in the cookie could be omitted.
2. Add a hostOnly control, so the users can manually flag whether a cookie is hostOnly.

For solution 1
- It added too much mental burden on the users, every cookie needs to set a URL manually, and needs to handle sync domain with the URL in most cases.
- Need a new property on the model, hard to maintain, and hard to keep compatible with the previous model.
- It's a bit weird to be a part of a cookie.

For solution 2
- The hostOnly should be considered as a hidden flag.
- And hostOnly already exists in a response cookie, even though we cannot see it from the UI.

On balance, I think solution 2 is much better.

yaoweiprc (dismissed):

pavkout (approved):
Good job @ZxBing0066 🎸 

LGHT

ZxBing0066:
Thanks, @pavkout

Issue #8540, opened 2025-03-28
Inso CLI 11.0.1 is throwing error on Jenkins


ywidiastuti:
### Expected Behavior

Inso CLI v11.0.1 should work on Jenkins

### Actual Behavior

Inso CLI v11.0.1 is throwing error on Jenkins while executing the tests. 

This is the error i got from the console:
```
[network] Response failed req=req_a4ff8ebe50054c07834c9fd952f25720 err=Error: Failure when receiving data from the peer
at Context.<anonymous> (/tmp/insomnia-testing/a24ce0e0-024e-456e-8712-8339e4fc1c35-test.ts:11:32)
at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

### Reproduction Steps

1. Install Inso CLI v11.0.1 on Jenkins, this one `inso-linux-x64-11.0.1.tar.xz`
2. Check `./inso --version `
3. Execute the test `./inso run test TestPlan --env Environment -w insomnia-directory`

### Is there an existing issue for this?

- [x] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Which sync method do you use?

- [x] Git sync.
- [ ] Insomnia Cloud sync.
- [ ] Local only

### Insomnia Version

11.0.1

### What operating system are you using?

Ubuntu

### Operating System Version

Mint 21.3

### Installation method

download from insomnia.rest

### Last Known Working Insomnia version

10.3.0

ywidiastuti:
The problem started since inso CLI version 10.3.1

ywidiastuti:
could it be that the problem begins since you provided the -x64- for the inso-cli?

ywidiastuti:
@gatzjames @subnetmarco 
any plan to have a look at this issue? thank you!

ihexxa:
Hi @ywidiastuti Did you enable any proxy on the UI (the desktop app preference).

ywidiastuti:
@ihexxa i didn't enable any proxy on the UI. The only thing i disable is the SSL certificates validation for Request/Response, could it be the problem?

ihexxa:
@ywidiastuti I don't think the validation setting could cause this.
If possible, could you join the slack channel: https://chat.insomnia.rest and I might reach out to you for more information if I still can't reproduce it.

ywidiastuti:
@ihexxa you can contact me anytime if you need more info, also in the slack channel :)

ihexxa:
[...]

PR #8700, opened 2025-05-09
fix(cli): proxy options do not work and stop loading settings and cookies from exports - INS-5429

ihexxa:
### Background
Potentially related to #8540
- CLI proxy options (#8257) are not working, as options are overridden in the deeper call stack.
- No longer load settings and cookies from exports(#8111), as they seems not necessary in CLI execution.

### Changes
- [x] Remove db loading to make CLI setting overriding work
- [x] Stop loading settings and cookies from exports
- [x] Disabled a test for loading settings
- [x] Added a test for proxying
- [x] Also accept proxy options in lowercase.

### DIFF
--- a/packages/insomnia-inso/src/cli.test.ts
+++ b/packages/insomnia-inso/src/cli.test.ts
@@ -152,12 +152,12 @@ describe('inso dev bundle', () => {
     // Disabled this test
     // currently settings loading is disabled, as it could include values like proxies from UI, which might not make sense for the cli.
     // it can be re-enabled if necessary
-    xit('send request with settings enabled (by testing followRedirects)', async () => {
-      const input =
-        '$PWD/packages/insomnia-inso/bin/inso run collection -w packages/insomnia-inso/src/db/fixtures/nedb --requestNamePattern "withSettings" --verbose "Insomnia Designer" wrk_0b96eff';
-      const result = await runCliFromRoot(input);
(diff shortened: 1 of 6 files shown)


gatzjames (commented):
Since this is removing reading the reading settings from the network.ts does it affect how proxy works in the app?

ihexxa:
@gatzjames Good point, theoretically settings will be loaded by fetchRequestData in UI, and passed in as an argument.

ZxBing0066 (dismissed):
LGTM 🚀

gatzjames:
> @gatzjames Good point, theoretically settings will be loaded by fetchRequestData in UI, and passed in as an argument.

Can we verify this behaviour? Ideally with a e2e test or at least manually.

My only other concern is if this breaks existing functionality.
E.g. someone is using inso to run a collection that they have in the app. Will this affect how the collection runs or will the proxy and other settings be loaded appropriately?

ihexxa:
@gatzjames I checked this, there's [one test which already covers it](https://github.com/Kong/insomnia/blob/b229956097233dceabfb951cd064558aa11fbc40/packages/insomnia-smoke-test/tests/smoke/pre-request-script-features.test.ts#L376), as the main modification is related to the `tryToExecuteScript`, and I've also added one without scripts.
Regarding the concern, the existing behavior seems only loading in the local nedb case, so it basically won't affect real CI use cases (e.g. #8540), wdyt.

ZxBing0066 (approved):

PR #8701, opened 2025-05-09
fix: encoded value be decoded issue

### LINKED ISSUE #975: [Bug] URL escaped UTF-8 code units get de-encoded
- Insomnia Version: 5.16.6

## Details

I'm using query params to send a request towards an API but the URL isn't rendered how I want it to.
This is what I'm trying to achive using the query tab in in the request:

```
 https://example.com?text=%4E%C3%A4%72%20%6D%61%6E%20%74%65%73%74%61%72%20%74%65%73%74%61%72%20%6D%61%6E
```
But gets transformed into:
```
https://example.com?text=N%C3%A4r%20man%20testar%20testar%20man
```

I'm using a service that requires me to make the request in the first format and so I can't test this service using Insomnia.

ZxBing0066:
## Background

When users pass percent-encoded strings in the query parameters, the encoded characters will be decoded. E.g. %4E.

## Root Cause

1. `smartEncodeUrl` calls `deconstructQueryStringToParams` before `buildQueryStringFromParams`, `deconstructQueryStringToParams` always decodes the params, the original proposal I guess is to keep all the strings to be only encoded once. Although `encodeURIComponent` and `decodeURIComponent` are often thought of as complementary, **they are not exact opposites**. This is because `decodeURIComponent` will decode all percent-encoded sequences, whereas [`encodeURIComponent` only encodes necessary characters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#description).
2. When there is a `,` in the string, it'll be decoded to `%2C`, which seems it cause issues before, so `%2C` will always be replaced with `,`. But if there is `%2C` in the original input, we should consider it as `%2C` instead of `,`.

## Changes
1. Add `skipDecode` to `deconstructQueryStringToParams`
2. Use `ignore` param instead of replace all `%2C` to `,`
3. Remove unused whitespace replacement
4. Fix some typos

Closes #975

### DIFF
--- a/packages/insomnia/src/utils/url/querystring.test.ts
+++ b/packages/insomnia/src/utils/url/querystring.test.ts
@@ -69,7 +69,7 @@ describe('querystring', () => {
     });
   });
 
-  describe('build()', () => {
+  describe('buildQueryParameter()', () => {
     it('builds simple param', () => {
       const str = buildQueryParameter({ name: 'foo', value: 'bar??' });
       expect(str).toBe('foo=bar%3F%3F');
@@ -85,6 +85,21 @@ describe('querystring', () => {
       expect(str).toBe('foo');
     });
 
+    it('builds param with encoded value and name', () => {
+      const str = buildQueryParameter({ name: 'foo%11%09%2A%9D%0F', value: 'bar%11%09%2A%9D%0F' });
+      expect(str).toBe('foo%11%09%2A%9D%0F=bar%11%09%2A%9D%0F');
+    });
+
+    it('builds param with null value and disable strict', () => {
+      const str = buildQueryParameter({ name: 'foo', value: null }, false);
+      expect(str).toBe('foo=');
+    });
+
+    it('builds param with spaces', () => {
+      const str = buildQueryParameter({ name: 'foo bar', value: 'baz qux' });
+      expect(str).toBe('foo%20bar=baz%20qux');
+    });
+
     it('builds param with null value and enable strictNullHandling', () => {
       const str = buildQueryParameter({ name: 'foo', value: null }, true, { strictNullHandling: true });
       expect(str).toBe('foo');
@@ -93,6 +108,11 @@ describe('querystring', () => {
     it('builds param with empty string value', () => {
       const str = buildQueryParameter({ name: 'foo', value: '' });
       expect(str).toBe('foo');
+    });
+
+    it('builds param with empty string value and disable strict', () => {
+      const str = buildQueryParameter({ name: 'foo', value: '' }, false);
+      expect(str).toBe('foo=');
     });
 
     it('builds param with empty string value and enable strictNullHandling', () => {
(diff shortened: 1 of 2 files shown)


packages/insomnia/src/utils/url/querystring.test.ts
```diff
@@ -112,9 +132,34 @@ describe('querystring', () => {
...
+      const str = buildQueryParameter({ name: 'foo,', value: '1,2,3' });
+      expect(str).toBe('foo%2C=1,2,3');
+    });
+
+    it('builds param without decode the commas back', () => {
+      const str = buildQueryParameter({ name: 'foo', value: '1%2C' });
+      expect(str).toBe('foo=1%2C');
+    });
+
+    it.fails('build query param with __ENC__ returns incorrect result', () => {
```
ZxBing0066, line 146: There are some risks associated with the current solution.

packages/insomnia/src/utils/url/querystring.ts
```diff
@@ -245,10 +247,7 @@ export const smartEncodeUrl = (url: string, encode?: boolean, options?: IQuerySt
  * @param str string to encode
  * @param ignore characters to ignore
  */
-export const flexibleEncodeComponent = (str = '', ignore = '') => {
-  // Sometimes spaces screw things up because of url.parse
-  str = str.replace(/%20/g, ' ');
```
ZxBing0066, line 250: iShot_2025-05-12_10 50 14

This line was from 9 years before, I guess the original proposal is to avoid `%20` to be double-encoded when `decodeURI` or `decodeURIComponent` throws errors. I think it's safe to remove from the current code since it's using `__ENC__` to avoid double-encoding now.

packages/insomnia/src/utils/url/querystring.ts
```diff
@@ -224,16 +231,11 @@ export const smartEncodeUrl = (url: string, encode?: boolean, options?: IQuerySt
...
-    const encodedQsParams = [];
-    for (const { name, value } of qsParams) {
-      encodedQsParams.push({
-        name: flexibleEncodeComponent(name),
-        value: strictNullHandling && value === null ? null : flexibleEncodeComponent(value as string),
-      });
-    }
+    // Skip decode if already encoded, to avoid some decoded characters cannot be encoded again.
+    // E.g. %4E -> N, encodeURIComponent(decodeURIComponent('%4E')) is N which is not expected.
+    const qsParams = deconstructQueryStringToParams(parsedUrl.query, true, { strictNullHandling, skipDecode: true });
```
ZxBing0066, line 236: `buildQueryStringFromParams` will encode the name and value, and already ensure the string is not double-encoded. It's unnecessary to call `flexibleEncodeComponent` here.
And decoding the string will cause some strings cannot be encoded back since decodeURI and encodeURI are not exact opposites

PR #8707, opened 2025-05-13
chore: Support auto restart electron when main processor related code changes

cwangsmv:
**Changes:**
- Modify esbuild script to restart electron automatically when preload/main/hiddenWindow scripted changed
- Add a new ```dev:autoRestart``` script to start app with both front-end live-reload and main processor auto-restart
- Close the inspector when app quits to unblock electron app exit in development mode

[INS-507](https://konghq.atlassian.net/browse/INS-507)

[INS-507]: https://konghq.atlassian.net/browse/INS-507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -23,6 +23,7 @@
   ],
   "scripts": {
     "dev": "npm start -w insomnia",
+    "dev:main-watch": "npm run start:main-watch -w insomnia",
```
ZxBing0066: 👍
Just curious, is there any concern about enabling the hot reload for the main process by default? Why add another command instead of updating the existing command? 🤔

packages/insomnia/esbuild.main.ts
```diff
@@ -68,15 +72,84 @@ export default async function build(options: Options) {
...
+          console.log('[Dev Build] Electron exited');
+          startElectron();
+        });
+
+        //Shutdown electron first. Existing debugger inspector will be closed in quit event of the app.
+        electronProcess.kill();
+      }
+    };
+
+    startElectron();
```
yaoweiprc: I think we need to put startElectron after Context.watch().

cwangsmv: Good point, fixed

package.json
cwangsmv: Just to keep backward compatibility in case people do not want auto restart electron.

yaoweiprc (dismissed):

yaoweiprc (approved):

PR #8719, opened 2025-05-14
feat: improve design doc samples [INS-5526]

ryan-willis:
https://github.com/user-attachments/assets/67f87bbe-0655-4c18-92ae-a202964afd84

- disables linting/formating checks if spec contents are empty
- adds two new samples
  - Todo List API spec
  - Blank API spec


chayglass84:
Love it. 

Can we fix the petstore spec not to be riddled with warnings while we're at it? Even just whatever ChatGPT does to it?

Other then that, LGTM

gatzjames (approved):
LGTM

PR #8722, opened 2025-05-15
Fix the issue that SSE connection can not be disconnected in some cases [INS-5464]

yaoweiprc:
When connecting the Insomnia client to a locally running backend service, switching between multiple organizations causes HTTP requests to remain in a pending state.
image

I captured and analyzed the client’s network traffic and found that when switching organizations, the old SSE connections are not being closed. Once the number of SSE connections exceeds 6, all new requests remain in a pending state.
image

In the renderer process, the EventSource.close method is explicitly called to close old connections when switching organizations, but it clearly doesn’t take effect.

In the Insomnia app, we are not sending requests directly from the renderer process — instead, they are forwarded from the main process using electron.net.fetch.

net.fetch is indeed built on top of Chromium's native networking library, so it makes sense that it's subject to Chromium's per-domain connection limits.

The root cause of the issue was that under certain network conditions, Electron's net.fetch method fails to properly close SSE (Server-Sent Events) connections.

To work around this problem, I explored alternative solutions. I eventually found that using net.request inside protocol.handle, and manually constructing a Response object, provided a viable workaround. By using a ReadableStream as the Response.body and listening for its cancel event, I was able to accurately detect when the SSE connection was closed from the renderer process.

However, this workaround introduced a new problem. Since we're also forwarding regular HTTP requests through protocol.handle, the SSE requests made with net.request are inadvertently intercepted again by the handler designed for HTTP forwarding. That handler uses net.fetch, which brings us back to the original issue—SSE connections still can't be properly closed.

Electron does not provide a way to make net.request bypass other protocol handlers.

As a result, the only viable solution is to avoid using Electron’s built-in networking APIs for SSE entirely. Instead, I switched to using node-libcurl, which successfully establishes and closes SSE connections as expected.

### DIFF
--- a/packages/insomnia/src/main/api.protocol.ts
+++ b/packages/insomnia/src/main/api.protocol.ts
@@ -56,8 +56,6 @@ export async function registerInsomniaProtocols() {
           const url = new URL(`${apiURL}/${originalRequest.url.replace(`${insomniaStreamScheme}://`, '')}`);
           const urlStr = url.toString();
 
-          console.log('sse connecting main', apiURL);
-
           const sessionId = new URLSearchParams(url.search).get('sessionId');
 
           const curl = new Curl();
@@ -99,7 +97,6 @@ export async function registerInsomniaProtocols() {
 
           curl.on('stream', async (stream: Readable, _code: number, [headersWithStatus]: HeaderInfo[]) => {
             const { result, ...headers } = headersWithStatus;
-            console.log('sse headers', headers);
             resolve(
               new Response(Readable.toWeb(stream) as ReadableStream, {
                 status: _code,
--- a/packages/insomnia/src/main/api.protocol.ts
+++ b/packages/insomnia/src/main/api.protocol.ts
@@ -2,7 +2,7 @@ import { Readable } from 'node:stream';
 
 import { Curl, CurlAuth, CurlFeature, CurlProxy, CurlSslOpt, type HeaderInfo } from '@getinsomnia/node-libcurl';
 import { app, net, protocol, session } from 'electron';
-import { parse as urlParse } from 'url';
+import { parse as urlParse } from 'node:url';
 
 import { getApiBaseURL } from '../common/constants';
 import { get as getSettings } from '../models/settings';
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/main/api.protocol.ts
```diff
@@ -35,14 +47,73 @@ export async function registerInsomniaProtocols() {
...
   if (!protocol.isProtocolHandled(insomniaStreamScheme)) {
-    protocol.handle(insomniaStreamScheme, async request => {
-      const apiURL = getApiBaseURL();
-      const url = new URL(`${apiURL}/${request.url.replace(`${insomniaStreamScheme}://`, '')}`);
-      const sessionId = new URLSearchParams(url.search).get('sessionId');
-      request.headers.append('X-Session-Id', sessionId || '');
-      return net.fetch(url.toString(), request);
+    protocol.handle(insomniaStreamScheme, async originalRequest => {
+      const settings = await getSettings();
+      // here we use libcurl to forward the SSE request because the SSE request sent by net.fetch can not be disconnected correctly in some cases
```
gatzjames, line 59: One suggestion to link the open issue on the Electron repo to keep track of it's progress and provide a link for future reference if we want to revisit this.

packages/insomnia/src/main/api.protocol.ts
```diff
@@ -35,14 +47,73 @@ export async function registerInsomniaProtocols() {
...
+          const apiURL = getApiBaseURL();
+          const url = new URL(`${apiURL}/${originalRequest.url.replace(`${insomniaStreamScheme}://`, '')}`);
+          const urlStr = url.toString();
+
+          const sessionId = new URLSearchParams(url.search).get('sessionId');
+
+          const curl = new Curl();
+          curl.setOpt(Curl.option.URL, urlStr);
+          curl.setOpt(Curl.option.ACCEPT_ENCODING, '');
+          curl.setOpt(Curl.option.SSL_OPTIONS, CurlSslOpt.NativeCa);
```
gatzjames, line 68: I think this will keep working as before but maybe worth checking if custom CAs and system proxy settings are respected in a similar way with net.fetch

packages/insomnia/src/main/api.protocol.ts
```diff
@@ -35,14 +47,73 @@ export async function registerInsomniaProtocols() {
...
+
+          if (!settings.proxyEnabled) {
+            curl.setOpt(Curl.option.PROXY, '');
```
gatzjames: In net.fetch we use the system proxy option if proxy is not enabled.
🤔 Not sure if this will have the same result in cURL

gatzjames (dismissed):
🙏👏 Awesome write-up and explanation!
 
> Electron does not provide a way to make net.request bypass other protocol handlers.

Curious if you tried the [bypassCustomProtocolHandlers](https://www.electronjs.org/docs/latest/api/net#:~:text=the%20non%2Dstandard-,bypassCustomProtocolHandlers,-option%20is%20set) option in net.fetch.

We should double check the system Certificates and Proxy options work as before 
but overall looks good to me! 👍

packages/insomnia/src/main/api.protocol.ts
yaoweiprc: You are right, I'll take a look into it.

packages/insomnia/src/main/api.protocol.ts
yaoweiprc: https://github.com/electron/electron/issues/47097
I'll add it.

packages/insomnia/src/main/api.protocol.ts
yaoweiprc: > Curious if you tried the [bypassCustomProtocolHandlers]

That was exactly what we used before, but net.fetch can not close the SSE connection properly.
https://github.com/electron/electron/issues/47097

packages/insomnia/src/main/api.protocol.ts
yaoweiprc: I've added the logic to follow the system proxy.
CurlSslOpt.NativeCa means using system's native CA and I think it's consistent with net.fetch.

gatzjames (dismissed):

gatzjames (approved):

PR #8727, opened 2025-05-19
fix: preserve empty strings during export

ZxBing0066:
## Background 

INS-5463

Empty string environments are filtered out during export, which breaks the template render

## Changes

- Preserve empty strings
- Add test case


packages/insomnia/src/common/insomnia-v5.ts
```diff
@@ -42,12 +42,7 @@ import {
 type WithExportType<T extends models.BaseModel> = T & { _type: string };
 
 function filterEmptyValue(value: string | number | boolean | null | undefined) {
-  return (
-    value !== null &&
-    value !== undefined &&
-    value !== '' &&
-    !(typeof value === 'object' && Object.keys(value).length === 0)
```
ZxBing0066, line 49: I guess the original proposal is to make the exported file clean. Not sure if filtering the empty objects will also cause issues.

ryan-willis (dismissed):

packages/insomnia/src/common/insomnia-v5.ts
notjaywu: Perhaps @gatzjames has further insights on this.

gatzjames: Correct this is here only to make the exported file less cluttered. Imports should not be affected if we remove this whole function tbh

packages/insomnia/src/common/__tests__/insomnia-v5.test.ts
```diff
@@ -0,0 +1,96 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import * as models from '../../models';
+import type { Workspace } from '../../models/workspace';
+import { database } from '../database';
+import { getInsomniaV5DataExport } from '../insomnia-v5';
+
+vi.mock('../database', async () => ({
```
gatzjames: We should try to avoid mocking modules as much as we can since they are hard to maintain.
Instead we could expand the fixtures on the e2e tests to add empty strings and make sure no previous exports would break imo

ZxBing0066: Thank you for the reminder. I've revised the test, please take another review. @gatzjames

ihexxa (approved):

gatzjames (approved):

PR #8746, opened 2025-05-22
Chore(E2E): Remove arbitrary waits in e2e test

cwangsmv:
**Changes**
- Add a delay for the click action to replace the arbitrary wait for editor debounce to take effect
- Add expecting a change for new request tab/modal to appear rather than arbitrary waits


jackkav (approved):
Cheeky delay, but better

PR #8761, opened 2025-05-28
fix(Git Project migration): fix issue with matching the filepath with the document id

gatzjames:
Highlights:

- [x] Fixes an issue where the filepath could missmatch the document id


pavkout (approved):
LGTM

PR #8765, opened 2025-05-29
chore(analytics): capture abnormal run test events - INS-5623

ihexxa:
### Changes
- [x] chore(analytics): capture abnormal run test events
- [x] Also include org id in segment tracking

### DIFF
--- a/packages/insomnia/src/main/analytics.ts
+++ b/packages/insomnia/src/main/analytics.ts
@@ -1,7 +1,6 @@
 import crypto from 'node:crypto';
 
 import { Analytics } from '@segment/analytics-node';
-import * as Sentry from '@sentry/electron/main';
 import { net } from 'electron';
 import { v4 as uuidv4 } from 'uuid';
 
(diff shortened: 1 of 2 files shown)


packages/insomnia/src/main/analytics.ts
```diff
@@ -95,6 +96,18 @@ export async function trackSegmentEvent(event: SegmentEvent, properties?: Record
       );
     } catch (error: unknown) {
       console.warn('[analytics] Unexpected error while sending segment event', error);
+    } finally {
+      if (!userSession?.hashedAccountId && [SegmentEvent.unitTestRun, SegmentEvent.unitTestRunAll].includes(event)) {
```
CurryYangxx, line 100: Seems you just want to send a specific Sentry event here, but trackSegmentEvent is a common function.
I think it is better to report Sentry event directly in the corresponding action.

ihexxa: Yes, this is added temporarily and finally we will remove this, or we may need to add several duplicated captures calls with some logic for getting session.

CurryYangxx (approved):

PR #8780, opened 2025-06-04
fix: apply specific environment when sending requests in the template

ZxBing0066:
## Background

INS-5180
When running tests via the Inso CLI, users need to specify the environment. The request sent from the test via `insomnia.send` respects the specific environment, but the request sent from the template doesn't. Which leads to the `environment render error`.

## Changes

- Unify `fetchInsoRequestData` and  `fetchRequestData` to avoid inconsistent issues
- Pass `environmentId` when sending requests in the template


jackkav (approved):
I worry a bit about creating a new way to set an environmentId, as I feel there might be another but I could be wrong. Nice job cutting down the fetchRequestData drift.

ZxBing0066:
> I worry a bit about creating a new way to set an environmentId, as I feel there might be another but I could be wrong.

Yeah, good point. I think maybe we can set the environmentId into the workspace and override the one in the workspace meta in the database, then we don't need to carry the environmentId across the entire process. Not sure about the effort and risk, maybe we can consider it when we meet another related issue/requirement.

ihexxa (approved):

PR #8794, opened 2025-06-11
optimize(Git): move long-running git operations to a separate process to prevent blocking main thread

ZxBing0066:
## Background

INS-5761

When there are more than 1000 requests in one collection, some git operations will significantly block the main process, and users can't do anything during that time. This has seriously affected the user experience.

## Changes

- Move `gitStatus`, `diffFileLoader`, and `gitChangesLoader` to a separate process to avoid them blocking the main process.

Still reuse the current git service and database, so currently only support read only operations in the git process.

### Current
https://github.com/user-attachments/assets/8306f9d9-56db-4360-a4e1-e1298589f464

### After
https://github.com/user-attachments/assets/7a3046b0-e1ce-4ecc-a77c-bae803d2a07b

### DIFF
--- a/packages/insomnia/esbuild.main.ts
+++ b/packages/insomnia/esbuild.main.ts
@@ -71,7 +71,24 @@ export default async function build(options: Options) {
     ],
   });
 
-  return Promise.all([main, preload, hiddenBrowserWindowPreload]);
+  const gitWorker = esbuild.build({
+    entryPoints: ['./src/git-worker.ts'],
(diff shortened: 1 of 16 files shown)


ZxBing0066:
There are write behaviors in `models.initModel` and it's used everywhere. So it's unsafe to directly call the database in multiple processes, even only the read functions.
Currently, close this PR as it introduces risks, and will reopen it if it's necessary.
Maybe we could solve the risks by moving all the database operations into another process. But it changes too much, too risky as a short-term solution.

PR #8804, opened 2025-06-17
fix: tab disappear when discard changes - INS-757

CurryYangxx:
close https://konghq.atlassian.net/browse/INS-757

**Changes**
Since we listen to the database change event to sync the tabList with the database. We will close the tab when the entity is deleted.
https://github.com/Kong/insomnia/blob/403218cda5b2be92fc112839ef574021798425b9/packages/insomnia/src/ui/components/tabs/tab-list.tsx#L226

This issue is related to #8551
The current pr only deletes the entities which are not exist in the new git file version.


gatzjames (approved):
LGTM

PR #8811, opened 2025-06-20
fix(Git Sync): Display up to date content in code editors when data change through git sync

gatzjames:
# Overview

Currently when content changes because of changes coming through git, the code editor doesn't reflect them immediately but requires another update to happen.

## Highlights:
- [x] Introduces a new property in the gitRepository model `statusIdentifier` that holds a hash of the latest commit alongside any changes that are not commited yet in git
- [x] The identifier gets updated when any operations related to git happen in git service.
- [x] We use this identifier in the code-editor key that updates it's state if it changes and therefore displays up to date content in these scenarios

Closes INS-5227, INS-5626

### DIFF
--- a/packages/insomnia/src/main/git-service.ts
+++ b/packages/insomnia/src/main/git-service.ts
@@ -177,6 +177,10 @@ export async function loadGitRepository({ projectId, workspaceId }: { projectId:
         legacyInsomniaWorkspace = await containsLegacyInsomniaDir({ fsClient });
       }
 
+      await models.gitRepository.update(gitRepository, {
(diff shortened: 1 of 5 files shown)


ZxBing0066:
With this change, `statusIdentifier` will be called frequently, and as you know, its performance is suboptimal and could cause the main process to become unresponsive.
Would it be possible to defer this merge until after your refactoring of the Git subsystem is complete?

gatzjames:
Closing in favor of https://github.com/Kong/insomnia/pull/8851

PR #8821, opened 2025-06-30
fix(Import): Folder headers are not imported correctly

### LINKED ISSUE #8714: Insomnia v5 Exported Data import failed
### Expected Behavior

I've export project in Insomnia v5 format and confirmed that .yaml file contains collection level's header setting as below:
Image

It supposed to correctly imported as well.

### Actual Behavior

Those headers disappeared when I switched to another device and import this .yaml file.

### Reproduction Steps

1. Install Insomnia 11.1.0
2. Create collection and add collection level's header setting.
3. Export collection in Insomnia v5 format (should create one yaml file)
4. Import this collection and check collection level's header setting.

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [x] Local only

### Additional Information

_No response_

### Insomnia Version

11.1.0

macOS

### Operating System Version

macOS Sequoia 15.1

### Installation method

Download from Website

Before Version 11

gatzjames:
# Overview

Fixes an issue where importing a a folder with headers would not import the headers for that folder

Closes #8714

### DIFF
--- a/packages/insomnia/src/common/insomnia-v5.ts
+++ b/packages/insomnia/src/common/insomnia-v5.ts
@@ -300,6 +300,7 @@ function getCollection(file: InsomniaFile): (Request | WebSocketRequest | GrpcRe
             _type: EXPORT_TYPE_REQUEST_GROUP,
             name: item.name || 'Imported Folder',
             parentId,
+            headers: item.headers?.map(({ name, value }) => ({ name: name || '', value: value || '' })) || [],
             preRequestScript: item.scripts?.preRequest || '',
             afterResponseScript: item.scripts?.afterResponse || '',
             authentication: item.authentication || {},


packages/insomnia/src/common/insomnia-v5.ts
```diff
@@ -300,6 +300,7 @@ function getCollection(file: InsomniaFile): (Request | WebSocketRequest | GrpcRe
             _type: EXPORT_TYPE_REQUEST_GROUP,
             name: item.name || 'Imported Folder',
             parentId,
+            headers: item.headers?.map(({ name, value }) => ({ name: name || '', value: value || '' })) || [],
```
ZxBing0066, line 303: https://github.com/Kong/insomnia/blob/bf1433d13615141b0bc099f27bca4662d36132b2/packages/insomnia/src/common/insomnia-v5.ts#L329

Should we keep them consistent?

ZxBing0066 (approved):
🚀

packages/insomnia/src/common/insomnia-v5.ts
gatzjames: This is because of a typing issue with the schema.
WebSocket Request is using the same schema and gets the correct type but nor RequestGroup.
Will make another PR to upgrade zod to the newer version and consolidate inconsistencies to address this

ZxBing0066: Got it.

PR #8839, opened 2025-07-07
feat(Inso): add import collections

ZxBing0066:
INS-5861

## Background

Support importing collections with Inso CLI.

## Changes

- Import collections into an existing Cloud-Sync project: `inso import collections --from <path> --toProject <project-id>`
- Import collections into a folder: `inso import collections --from <path> --toDir <folder-name>`
- Extract shared session code to `src/common/session`
- Make `crypt` node compatible


packages/insomnia/src/common/account/crypt.ts
```diff
@@ -11,18 +9,6 @@ export interface AESMessage {
   ad: string;
 }
 
-/**
```
ZxBing0066, line 14: deriveKey and generateKeyPairJWK are not used.

packages/insomnia/src/common/import.ts
```diff
@@ -73,22 +75,6 @@ export interface PostmanDataDumpRawData {
   envList: ImportEntry[];
 }
 
-export async function getFilesFromPostmanExportedDataDump(filePath: string): Promise<PostmanDataDumpRawData> {
```
ZxBing0066, line 76: It's renderer-only code, should move out from `src/common`.

PR #8846, opened 2025-07-08
fix: squirrel windows can not start the new version after update - [INS-856]

CurryYangxx:
This should be introduced after #8451 
Issue list:

- The shortcut target is wrong; the correct one should be `/AppData/Local/insomnia/Insomnia.exe`
- When the user clicks restart in the update modal, it can not launch the new version

This is because we copy the Insomnia.exe to Insomnia-$version.exe and then execute in the wrapper, when Squirrel try to launch the new version, it will try to open the same exe file(`path.basename(process.execPath)`) in the new version folder, but there is no that file in the new version folder.
```
[02/07/25 18:49:59] info: Program: Starting Squirrel Updater: --processStartAndWait insomnia-11.2.0.exe
[02/07/25 18:50:00] info: Program: Want to launch 'C:\Users\curry.yangkonghq.com\AppData\Local\insomnia\app-11.3.0-beta0\insomnia-11.2.0.exe'
[02/07/25 18:50:00] error: Program: File C:\Users\curry.yangkonghq.com\AppData\Local\insomnia\app-11.3.0-beta0\insomnia-11.2.0.exe doesn't exist in current release
[02/07/25 18:50:00] fatal: Finished with unhandled exception: System.ArgumentException: Value does not fall within the expected range.
   at Squirrel.Update.Program.ProcessStart(String exeName, String arguments, Boolean shouldWait)
   at Squirrel.Update.Program.executeCommandLine(String[] args)
   at Squirrel.Update.Program.main(String[] args)
```

Changes:

- Fix the short cut creation process in cpp wrapper, should same with [electron-squirrel-startup](https://github.com/mongodb-js/electron-squirrel-startup/blob/master/index.js)
- Run `update.exe  --processStartAndWait Insomnia.exe` instead of calling `quitAndInstall` in electron.autoUpdater

reference the source code in electron:
https://github.com/electron/electron/blob/5ef6897bc73fa0301df583717be8bc53485dc239/lib/browser/api/auto-updater/auto-updater-win.ts#L11

### DIFF
--- a/packages/insomnia/src/cpp/insomnia.cpp
+++ b/packages/insomnia/src/cpp/insomnia.cpp
@@ -5,6 +5,7 @@
 #include <iostream>
 #include <string>
 #include <windows.h>
+#include <filesystem>
 
 const wchar_t *INSOMNIA_VERSION = L"__VERSION__";
 
@@ -122,7 +123,9 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
 
     // Squirrel.Windows install
     std::wstring shortcut = QuotePathIfNeeded(insomniaExecutable);
-    std::wstring args = std::wstring(L"--createShortcut=") + shortcut;
+    std::wstring shortcutTarget = std::filesystem::path(shortcut).filename().wstring();
+    ::DebugLog((L"Shortcut target: " + shortcutTarget).c_str());
+    std::wstring args = std::wstring(L"--createShortcut=") + shortcutTarget;
     ::ShellExecuteW(0, L"open", updatePath.c_str(), args.c_str(), NULL, SW_HIDE);
 
     return 0;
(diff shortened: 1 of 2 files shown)


ryan-willis (approved):

packages/insomnia/src/cpp/insomnia.cpp
```diff
@@ -122,7 +123,9 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
 
     // Squirrel.Windows install
     std::wstring shortcut = QuotePathIfNeeded(insomniaExecutable);
-    std::wstring args = std::wstring(L"--createShortcut=") + shortcut;
+    std::wstring shortcutTarget = std::filesystem::path(shortcut).filename().wstring();
```
ryan-willis, line 126: Can you help me understand why we need to do this? If I understand correctly `shortcutTarget` will always be `Insomnia.exe`. What does removing the path do for us?

CurryYangxx:
```
|--Insomnia
    |-- app-10.0
        |-- Insomnia.exe
    |-- app-11.0
        |-- Insomnia.exe
    |-- update.exe
    |-- Insomnia.exe
```
This is the insomnia directory tree installed by Squirrel. The origin `shortcut` is the full path pointing to `Insomnia/app-10.0/Insomnia.exe`, and it will not change after auto-update. So the user can not launch the new version.
According to the [electron-squirrel-startup](https://github.com/mongodb-js/electron-squirrel-startup/blob/master/index.js), it uses the basename, so the shortcut should be the `Insomnia.exe` at the same level as `update.exe`, because this one could find the latest version and launch the exe file in the latest app folder.

PR #8866, opened 2025-07-15
fix: only reconnect SSE when switch organization-[INS-779]

CurryYangxx:
Reconnect SSE only when switching organizations.


ZxBing0066 (dismissed):
🚀

ZxBing0066 (approved):

PR #8876, opened 2025-07-17
fix: Incorrect handling of events-[INS-779]

CurryYangxx:
- FileDeleted
Delete a workspace file, we don't need to sync projects
- FileChange
Add a new file or modify the file, we only need to submit `insomnia-sync/sync-data` action when the current workspace is the same as the remote file
If on the project page, revalidate to get new workspace list.
- BranchDeleted
We only need to submit `insomnia-sync/sync-data` action when the current workspace is the same as the remote file

### DIFF
--- a/packages/insomnia/src/ui/context/app/insomnia-event-stream-context.tsx
+++ b/packages/insomnia/src/ui/context/app/insomnia-event-stream-context.tsx
@@ -1,9 +1,10 @@
 import React, { createContext, type FC, type PropsWithChildren, useContext, useEffect, useState } from 'react';
-import { useFetcher, useParams, useRouteLoaderData } from 'react-router';
+import { useFetcher, useParams, useRevalidator, useRouteLoaderData } from 'react-router';
 import { useLatest } from 'react-use';
 
 import { CDN_INVALIDATION_TTL } from '../../../common/constants';
 import type { Organization } from '../../../models/organization';
+import { VCSInstance } from '../../../sync/vcs/insomnia-sync';
 import { insomniaFetch } from '../../../ui/insomniaFetch';
 import { avatarImageCache } from '../../hooks/image-cache';
 import type { ProjectIdLoaderData } from '../../routes/project';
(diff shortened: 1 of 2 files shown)


packages/insomnia/src/ui/context/app/insomnia-event-stream-context.tsx
```diff
@@ -75,6 +76,22 @@ interface UserPresenceEvent extends UserPresence {
...
+  if (
+    currentBackendProject &&
+    currentBackendProject?.id === remoteWorkspaceId &&
+    currentBackendProject.rootDocumentId === workspaceId
+  ) {
+    return true;
+  }
+  return false;
+};
+
```
ZxBing0066, line 94: It could be simpler.

```suggestion
const isSameWorkspaceWithRemote = (workspaceId: string | undefined, remoteWorkspaceId: string | undefined) => {
  const vcs = VCSInstance();
  const currentBackendProject = vcs.getActiveBackendProject();
  if (
    workspaceId && remoteWorkspaceId &&
    currentBackendProject?.id === remoteWorkspaceId &&
    currentBackendProject?.rootDocumentId === workspaceId
  ) {
    return true;
  }
  return false;
};

```

packages/insomnia/src/ui/context/app/insomnia-event-stream-context.tsx
```diff
@@ -231,19 +245,25 @@ export const InsomniaEventStreamProvider: FC<PropsWithChildren> = ({ children })
                 },
               );
             } else if (
-              ['BranchDeleted', 'FileChanged'].includes(event.type) &&
+              (event.type === 'FileChanged' || event.type === 'BranchDeleted') &&
               event.team === organizationId &&
               latestRemoteId.current &&
               event.project === latestRemoteId.current
```
ZxBing0066, line 251: Could we reuse `isSameWorkspace` here?

ZxBing0066 (approved):

PR #8905, opened 2025-07-24
fix(GitRemoteBranchSelect): update fetcher key to use deferred URL value

gatzjames:
# Overview:

In the new remote-branch-select we use the deferred url to fetch the latest branches which can lead to a race condition and fetch the previous data.

We use the deferred url as a cache key for the fetcher instead to make sure these two are synchronized


pavkout (approved):
LGTM 🚀

PR #8927, opened 2025-07-31
Expose NodeJS crypto native module to scripts.

OneHoopyFrood:
RE: Discussion: #8276

This will also need an update to documentation.

I'm unclear if this will require a change to testing. I'd be happy to add such with some minimal guidance.

I have included a deprecation warning, but I don't love where I've put it. If there is a better way or you prefer not to put such a notice in code, do let me know and I will make the needed changes.


packages/insomnia-testing/src/run/run.ts
```diff
@@ -1,5 +1,4 @@
-import { unlink, writeFileSync } from 'node:fs';
-import fs from 'node:fs';
+import fs, { unlink, writeFileSync } from 'node:fs';
```
OneHoopyFrood, line 1: This was automatically done by VSCode. I see no reason to refuse the mild optimization.

packages/insomnia-testing/src/run/run.ts
```diff
@@ -20,17 +19,18 @@ function prependInterceptedRequireToSource(source: string): string {
     if (
       [
         // node.js modules
-        'path',
         'assert',
         'buffer',
-        'util',
-        'url',
+        'crypto',
```
OneHoopyFrood, line 24: I added the 'crypto' module but also sorted the lines since I wasn't sure where to add it in the list. Since it didn't appear to matter, I figured I'd just alphabetize the list.

packages/insomnia/src/requireInterceptor.ts
```diff
@@ -61,6 +62,9 @@ export const requireInterceptor = (moduleName: string): any => {
       'xml2js',
     ].includes(moduleName)
   ) {
+    if (moduleName == "crypto-js") {
+      console.warn("crypto-js has been decommissioned in favor of NodeJS native \"crypto\" which is available through require. Consider using it instead as support for crypto-js will soon be dropped.")
+    }
```
OneHoopyFrood, line 67: The warning I mentioned. Happy to remove if you prefer.

packages/insomnia/src/requireInterceptor.ts
```diff
@@ -28,17 +28,18 @@ export const requireInterceptor = (moduleName: string): any => {
...
+      'path',
       'punycode',
       'querystring',
-      'string_decoder',
       'stream',
+      'string_decoder',
       'timers',
-      'events',
+      'url',
+      'util',
```
OneHoopyFrood, line 42: I added the 'crypto' module but also sorted the lines since I wasn't sure where to add it in the list. Since it didn't appear to matter, I figured I'd just alphabetize the list.

packages/insomnia-testing/src/run/run.ts
OneHoopyFrood, line 1: The string-code in this file is a bit strange. I explored possibly having it make use of the code from the other package, but that was complex so I just decided to update it despite the duplication. I assume that's why it was done in the first place.

PR #8940, opened 2025-08-05
fix(Git Sync): Custom repository clone invalid URL

gatzjames:
# Overview:

When cloning a repository from the custom form the user sees an error of `invalid URL`

The reason this happens is because we disable the form fields if the `uri` value exists instead of the `defaultValues.uri`
When the form get's submitted after that the disabled fields don't make it to the form submission and are missing from the clone action.

This PR fixes the issue by using the default uri in order to disable the form or not.


pavkout (approved):
LGTM

PR #8947, opened 2025-08-06
Closes #8946. OAuth2 plugin refresh token issue using basic auth for client credentials.

### LINKED ISSUE #8946: Insomnia fails to refresh token as it doesn't provide client credentials to token endpoint.
This code even has a question "Why not send headers here?". Credentials are certainly required by the token endpoint, headers or otherwise.
https://github.com/Kong/insomnia/blob/2f746f3e600c4a2045fea62a386c5384fa7b2253/packages/insomnia/src/network/o-auth-2/get-token.ts#L296-L297

ulisesbocchio:
Closes #8946

Fixes issue when using Basic Auth for client credentials on OAuth2 authentication.
When using Basic Auth for client credentials, the refresh token functionality fails as it doesn't send the credentials in Authorization header.


yaoweiprc (approved):
Appreciate the fix—great catch!

PR #8948, opened 2025-08-07
Refactor/db get descandents - [INS-1075]

CurryYangxx:
- [x] create new function for get descandents
- [x] generate descandent map from query type
- [x] complete the descandents map for all data models
- [x] replace origin withDescandents one by one


Dustin4444 (approved):
Duplicate of #

packages/insomnia/src/models/index.ts
```diff
@@ -281,3 +285,47 @@ export const WORKSPACE_EXPORT_TYPES_DESCENDANT_MAP: Record<string, string[]> = {
...
+      return;
+    }
+    visited.add(child);
+    const parents = childToParents[child];
+    if (parents && parents.length) {
+      for (const p of parents) {
+        if (!map[p]) {
+          map[p] = [];
+        }
+        if (!map[p].includes(child)) {
```
CurryYangxx: check if necessary

packages/insomnia/src/common/database.ts
```diff
@@ -606,6 +606,67 @@ export const database = {
...
+  getWithDescendants: async function <T extends BaseModel>(doc: T, types: models.ModelTypes = []) {
+    if (db._empty) {
+      return _send<T[]>('getWithDescendants', ...arguments);
+    }
+
+    if (!doc) return [];
+
+    let docsToReturn: BaseModel[] = [doc];
+
+    const queryTypesDescendantMap = types?.length ? models.generateDescendantMap(types) : models.DESCENDANT_MAP;
```
ZxBing0066: Nit: Since `types` has a default value, `?` could be removed.

ZxBing0066: If `types` is provided, it seems it returns all the descendants and the doc itself. Should we only return the items with the corresponding types or add an option for it? I think most of the scenarios we only need these items.

packages/insomnia/src/models/index.ts
```diff
@@ -276,11 +277,64 @@ export const WORKSPACE_EXPORT_TYPES_DESCENDANT_MAP: Record<string, string[]> = {
     unitTestSuite.type,
     protoDirectory.type,
     protoFile.type,
+    workspaceMeta.type,
```
ZxBing0066: Is `socketIORequest` missing here? Maybe we can add a temp array of the requests to reduce repetitive statements. `const allRequests = [request.type, ....];`

packages/insomnia/src/models/index.ts
```diff
@@ -276,11 +277,64 @@ export const WORKSPACE_EXPORT_TYPES_DESCENDANT_MAP: Record<string, string[]> = {
...
-  [unitTestSuite.type]: [unitTest.type],
+  [unitTestSuite.type]: [unitTest.type, unitTestResult.type],
+  [unitTest.type]: [unitTestResult.type],
   [protoDirectory.type]: [protoFile.type],
 };
+
+export const generateDescendantMap = (queryTypes: ModelTypes): Record<string, string[]> => {
+  const result: Record<string, string[]> = {};
+
+  const childToParents: Record<string, string[]> = {};
```
ZxBing0066, line 316: Nit: `childToParents` never changes, may we cache it?

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.tsx
```diff
@@ -361,8 +361,11 @@ export const revalidateWorkspaceActiveRequest = async (requestId: string, worksp
 };
 
 export const revalidateWorkspaceActiveRequestByFolder = async (requestGroup: RequestGroup, workspaceId: string) => {
-  const docs = await database.withDescendants(requestGroup, models.request.type, [
+  const docs = await database.getWithDescendants(requestGroup, [
```
ZxBing0066, line 364: Seems to be an existing issue? 🤔

ZxBing0066: So many places need explicit declarations of the types; it's risky when we want to add a new request type.

packages/insomnia/src/ui/components/modals/proto-files-modal.tsx
```diff
@@ -165,7 +165,7 @@ export const ProtoFilesModal: FC<Props> = ({ defaultId, onHide, onSave }) => {
       }
 
       // Try parse all loaded proto files to make sure they are valid
-      const loadedEntities = await db.withDescendants(createdDir);
+      const loadedEntities = await db.getWithDescendants(createdDir);
```
ZxBing0066, line 168: Maybe could use `protoFile` as the types here.

packages/insomnia/src/ui/components/tabs/tab-list.tsx
```diff
@@ -184,11 +184,13 @@ export const OrganizationTabList = ({ showActiveStatus = true, currentPage = ''
               url: `/organization/${organizationId}/project/${projectId}/workspace/${workspace._id}/debug/request/${doc._id}`,
             });
           } else if (isRequestGroup(doc)) {
-            const folderEntities = await database.withDescendants(doc, models.request.type, [
+            const folderEntities = await database.getWithDescendants(doc, [
```
ZxBing0066, line 187: Another bug found? 🙈

packages/insomnia/src/ui/components/templating/tag-editor.tsx
```diff
@@ -84,7 +84,8 @@ export const TagEditor: FC<Props> = props => {
     for (const type of models.types()) {
       allDocs[type] = [];
     }
-    for (const doc of await db.withDescendants(props.workspace, models.request.type)) {
+    const descendants = await db.getWithDescendants(props.workspace, [models.request.type]);
```
ZxBing0066, line 87: I'm not sure if it's safe to use request as the types here. If it needs something like cookieJar, it will be broken.

ZxBing0066 (commented):
👍

packages/insomnia/src/ui/components/tabs/tab-list.tsx
CurryYangxx: move request also have the same issue

packages/insomnia/src/models/index.ts
CurryYangxx: move it outside the function

packages/insomnia/src/ui/components/modals/proto-files-modal.tsx
CurryYangxx: protoDir only has child protoDir and profoFile inside, so we don't need to pass the second param.

packages/insomnia/src/models/index.ts
CurryYangxx: Good catch!

packages/insomnia/src/ui/components/templating/tag-editor.tsx
CurryYangxx: It should be safe, we only use request and requestGroup here.

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.tsx
CurryYangxx: Yes, this is an existing issue, will fix it in a separate pr.

packages/insomnia/src/ui/components/tabs/tab-list.tsx
CurryYangxx: The same issue for tab,  will fix in a separate pr.

packages/insomnia/src/common/insomnia-v5.ts
```diff
@@ -489,17 +488,10 @@ export async function getInsomniaV5DataExport({
...
-      [],
-      WORKSPACE_EXPORT_TYPES_DESCENDANT_MAP,
-    );
-
-    const exportableTypes = Object.values(models.MODELS_BY_EXPORT_TYPE).map(model => model.type);
+    const workspaceDescendants = await database.getWithDescendants(workspace, models.EXPORTABLE_TYPES);
 
     const exportableResources = workspaceDescendants.filter(resource => {
-      if (exportableTypes.includes(resource.type)) {
+      if ((models.EXPORTABLE_TYPES as string[]).includes(resource.type)) {
```
ZxBing0066: Nit: IMO, this would be more accurate.
```suggestion
      if (models.EXPORTABLE_TYPES.includes(resource.type as models.ModelTypes[number])) {
```

packages/insomnia/src/common/__tests__/database.test.ts
```diff
@@ -740,42 +740,70 @@ describe('withDescendants()', () => {
...
 
     await expect(
-      db.withDescendants(workspace, null, [], {
-        Workspace: ['RequestGroup'],
-      }),
-      'Should return only workspace and top request group',
-    ).resolves.toEqual([workspace, folder1]);
+      db.getWithDescendants(workspace, [models.requestGroup.type]),
+      'Should return workspace with all request groups',
+    ).resolves.toEqual(expect.arrayContaining([workspace, folder1, folder2]));
```
ZxBing0066: Nit: `arrayContaining` cannot cover the case that the expected value contains more items. [sameMembers](https://vitest.dev/api/assert.html#samemembers) may be better.

ZxBing0066 (dismissed):
👍 Looks great. But I'm not sure if the descendant map is complete. May need other folks who have enough knowledge help to check.

packages/insomnia/src/common/database.ts
```diff
@@ -607,74 +607,46 @@ export const database = {
...
-          }
+        const queryTypes = Array.from(uniqueDescendantTypes);
+
+        for (const type of queryTypes) {
+          // If the doc is null, we want to search for parentId === null
+          const promise = database.find(type, { parentId: { $in: parentIdsMap.get(type) || [] } });
+          promises.push(promise);
         }
 
         for (const more of await Promise.all(promises)) {
```
cwangsmv: Nit: we can also enhance this. Instead of create a new array on each iteration, we can use flat to do it only once
```
const docBatches = await Promise.all(promises);
foundDocs.push(...docBatches.flat());
```

packages/insomnia/src/common/__tests__/database.test.ts
CurryYangxx: use [samedeepmembers](https://vitest.dev/api/assert.html#samedeepmembers) here

ZxBing0066 (dismissed):

ZxBing0066 (approved):

PR #8953, opened 2025-08-08
feat: add data folder constraint for read file tags - INS-1062

ihexxa:
### Changes
- [x] Add getSettings to the rendering context
- [x] In read file interface, get data folders from settings and validate template's file path
- [x] Return error message in tag modal and preview
- [x] Make it work in worker and renderer
- [x] Updated smoke test

image

image

### DIFF
--- a/packages/insomnia/src/templating/base-extension-worker.ts
+++ b/packages/insomnia/src/templating/base-extension-worker.ts
@@ -179,8 +179,13 @@ export default class BaseExtension {
       meta: renderMeta,
       renderPurpose,
       util: {
-        readFile: async (path: string, encoding?: string) =>
-          fetchFromTemplateWorkerDatabase('readFile', { path, encoding }),
+        readFile: async (path: string, encoding?: string) => {
+          const allowed = renderContext?.getSettings().dataFolders.some((folder: string) => folder !== '' && path.startsWith(folder));
+          if (!allowed) {
+            throw `Insomnia cannot access the file ‘${path}’. You can adjust this in Preferences → Security.`;
+          }
+          return fetchFromTemplateWorkerDatabase('readFile', { path, encoding });
+        },
         nodeOS: async () => fetchFromTemplateWorkerDatabase('nodeOS', {}),
         decode: async (buffer: Buffer, encoding?: string) =>
           fetchFromTemplateWorkerDatabase('decode', { buffer, encoding }),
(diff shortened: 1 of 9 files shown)


packages/insomnia/src/templating/base-extension.ts
```diff
@@ -118,6 +118,11 @@ export default class BaseExtension {
           };
         },
         readFile: async (path: string, encoding = 'utf8') => {
+          const allowed = renderContext?.getSettings().dataFolders.some((folder: string) => folder !== '' && path.startsWith(folder));
```
gatzjames, line 121: We should also check for paths that might escape this by using '../..' or similar.
An example of how to do this can be found in the electron docs: https://www.electronjs.org/docs/latest/api/protocol#:~:text=//%20NB%2C%20this%20checks,%7D)%0A%20%20%20%20%20%20%7D

ihexxa: I think this is a good point which adds more flexibility and probably we might regard it as an enhancement. As currently specifying absolute paths is straightforward then it can quickly be understood. Relative ones are more powerful and also introduce more possibilities.

packages/insomnia/src/templating/base-extension-worker.ts
```diff
@@ -179,8 +179,13 @@ export default class BaseExtension {
       meta: renderMeta,
       renderPurpose,
       util: {
-        readFile: async (path: string, encoding?: string) =>
-          fetchFromTemplateWorkerDatabase('readFile', { path, encoding }),
+        readFile: async (path: string, encoding?: string) => {
+          const allowed = renderContext?.getSettings().dataFolders.some((folder: string) => folder !== '' && path.startsWith(folder));
```
cwangsmv, line 183: Nit: Instead of adding a new render context to get settings, we can use the existing way of ``` await fetchFromTemplateWorkerDatabase('settings.get', {})```

packages/insomnia/src/templating/base-extension.ts
cwangsmv, line 121: Nit: Instead of adding a new render context to get settings, we can use ```models.settings.get``` directly to get settings

cwangsmv (approved):

sunderhus:
I think it would be great to have a quick-add button on the popup message for unauthorized paths.
This way, with a single click, you could add the current file you’re trying to use in a request, instead of asking the user to select the path from the alert message, copy it, and add it manually to the Settings.

PR #8955, opened 2025-08-08
fix: enable dataFolders constraint for CLI - INS-1064

ihexxa:
### Changes
- [x] Add dataFolders arg for explicitly specifying directories that can be accessed
- [x] Add path validators


packages/insomnia/src/common/send-request.ts
```diff
@@ -97,6 +98,8 @@ export async function getSendRequestCallbackMemDb(
     // skip plugins
     const renderedRequest = renderedResult.request;
 
+    isFsAccessingAllowed(renderedRequest, mutatedContext.settings);
```
cwangsmv: Should we also consider ```requestData.caCert```? It will include the path of the customized ca cert file

cwangsmv: ```mutatedContext.clientCertificates``` will also contain path on cert attribute

ihexxa: Good catch, will add them.

cwangsmv (dismissed):
lgtm

cwangsmv (approved):

sunderhus:
@ihexxa I think it would be great to have a quick-add button on the popup message for unauthorized paths.
This way, with a single click, you could add the current file you’re trying to use in a request, instead of asking the user to select the path from the alert message, copy it, and add it manually to the Settings.

ihexxa:
@sunderhus Thanks for the feedback and ill also forward this to the team, ux improvement is in progress and will be shipped in the next release.

PR #8960, opened 2025-08-11
fix: introduce data folder as one constraint of file accessing

ihexxa:
### Changes
- [x] Enable fs path checking in accessing files from scripts, body editor and form params.
- [x] Added one component for array settings 
- [x] Updated UTs

image

image


jackkav (approved):
Nice first pass

PR #8961, opened 2025-08-11
fix: Put the collection filter into localStorage

### LINKED ISSUE #8591: Collection filter is reset when switching recent requests (top menu)

xiaodemen:
Closes [INS-1056](https://konghq.atlassian.net/browse/INS-1056)

Fixes #8591 

[INS-1056]: https://konghq.atlassian.net/browse/INS-1056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.tsx
+++ b/packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.tsx
@@ -38,6 +38,7 @@ import {
   useParams,
   useSearchParams,
 } from 'react-router';
+import { useLocalStorage } from 'react-use';
 
 import { DEFAULT_SIDEBAR_SIZE, getProductName, SORT_ORDERS, type SortOrder, sortOrderName } from '~/common/constants';
 import { type ChangeBufferEvent, database as db } from '~/common/database';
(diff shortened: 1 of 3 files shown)


jackkav (approved):
Much better. Also be aware we are planning an upgrade soon which will mean removing all the memos. https://react.dev/blog/2024/04/25/react-19-upgrade-guide#strict-mode-improvements

CurryYangxx:
I don't think using localStorage to store filter state is a good solution. We need to use in-memory state to manage state like this.

gatzjames (approved):

PR #8973, opened 2025-08-12
fix: Upgrade node-libcurl version to 2.36.12 and electron to 36.4.0

balugeorge:
## What does this PR do?
* Upgrade node-libcurl to 2.36.12
* Upgrade electron to 36.4.0 for to support the node-libcurl upgrade


balugeorge:
Addressed under https://github.com/Kong/insomnia/pull/8963

PR #9021, opened 2025-08-20
fix: Bump the aria-components version to resolve the slow search issue

### LINKED ISSUE #8914: quick search loading too long (>5 seconds)
### Expected Behavior

The quick search should appear and be immediately ready for typing after pressing `Cmd+P`. Small delay is acceptable

### Actual Behavior

After pressing Cmd+P, it takes too long to load.

https://github.com/user-attachments/assets/8db56924-b9cd-4ba9-9fa5-d9a8d407e0f8

### Reproduction Steps

_No response_

- [ ] Git sync.
- [x] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

I have three projects, eight collections, and over 100 requests. I also installed the `insomnia-plugin-request-body-hmac` plugin, but even when it’s disabled, the issue persists.

### Insomnia Version

v11.3.0

macOS

### Operating System Version

MacOS Sequoia 15.5

### Installation method

download

_No response_

xiaodemen:
Closes [INS-1111](https://konghq.atlassian.net/browse/INS-1111)

TODO:
- [ ] Modals like AlertModal/SettingsModal etc. in modals.tsx should use RAC modals instead of the custom implementation to avoid issues with focus, order, pressing esc to close etc. We also need to maintain the global showModal() API that is used in the app and by plugins. A great example of this is the showToast implementation that achieves a similar outcome for global toast notifications.

Fixes #8914 

[INS-1111]: https://konghq.atlassian.net/browse/INS-1111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


xiaodemen:
Yes, there are some ordering issues when our modal is used together with the RAC modal.

PR #9025, opened 2025-08-20
fix: Fix crash when viewing repo settings without a connected repo [INS-1173]

pavkout:
Closes INS-1173

### DIFF
--- a/packages/insomnia/src/main/git-service.ts
+++ b/packages/insomnia/src/main/git-service.ts
@@ -140,6 +140,10 @@ async function getGitRepository({ projectId, workspaceId }: { projectId: string;
   const project = await models.project.getById(projectId);
   invariant(project, 'Project not found');
   invariant(project.gitRepositoryId, 'Project is not linked to a git repository');
+  invariant(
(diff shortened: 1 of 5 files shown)


packages/insomnia/src/main/git-service.ts
```diff
@@ -2040,6 +2044,20 @@ const getRepositoryDirectoryTree = async ({
   repositoryTree: FileTree;
   folderList: Record<string, string[]>;
 }> => {
+  const project = await models.project.getById(projectId);
+
+  if (project?.gitRepositoryId === 'empty') {
```
gatzjames: Since we are using this convention in many places I think we should move it to a utility function in the project.ts like this:
```ts
// This is used to identify Git Projects that are not connected to a remote yet
export const EMPTY_GIT_PROJECT_ID = 'empty';

export function isEmptyGitProject(project: Project) {
  return project.gitRepositoryId === EMPTY_GIT_PROJECT_ID
}
```
and use it in all places we use this convenience.

pavkout: @gatzjames Good catch. Let me really quick change it

gatzjames (approved):

PR #9030, opened 2025-08-22
[INS-1234] Finesse experience of loading remote branches when creatin…

godfrzero:
This PR addresses UX issues in the Create a new project modal, especially evident when configuring access to a repository using the "Git" tab rather than using the Github/Gitlab tabs.

| Behavior | Example |
|:--|---|
| It now loads branches associated with a repository regardless of the order in which the URL, username, and password fields are filled out. Previously it would only work as expected if the credentials were filled in before the URL. | Kapture 2025-08-21 at 19 42 41 |
| It keeps the "Refresh" button enabled in case of an error, so it can be used as an escape hatch to manually retry fetching branches. | image |

Other than UX, it also avoids sending API requests which are obviously malformed by waiting until the URL and credentials are both available.

Changes were tested by creating new projects by connecting to Github, Gitlab, using a PAT for Github, and a PAT for Gitlab.

### DIFF
--- a/packages/insomnia/src/ui/components/git-credentials/git-remote-branch-select.tsx
+++ b/packages/insomnia/src/ui/components/git-credentials/git-remote-branch-select.tsx
@@ -1,5 +1,6 @@
-import React, { useDeferredValue, useEffect } from 'react';
+import React, { useDeferredValue } from 'react';
 import { Button, ComboBox, Input, Label, ListBox, ListBoxItem, Popover } from 'react-aria-components';
+import * as reactUse from 'react-use';
 import { z } from 'zod/v4';
 
 import type { GitCredentials } from '~/models/git-repository';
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/ui/components/git-credentials/git-remote-branch-select.tsx
```diff
@@ -18,22 +17,42 @@ export const GitRemoteBranchSelect = ({
 }) => {
   const uri = useDeferredValue(url);
   const remoteBranchesFetcher = useGitRemoteBranchesActionFetcher({ key: `branch-select:${uri}` });
-  const { organizationId } = useParams() as { organizationId: string };
```
godfrzero, line 21: Removed because it wasn't being used anywhere anymore.

gatzjames (approved):

Issue #9031, opened 2025-08-22
Pre-request script copied from Postman causes request to time out


MrMingFang:
### Expected Behavior

A pre-request script copied from a working Postman request should execute successfully within Insomnia's time limit. After the script completes, the main HTTP request should be sent as configured.

<img width="1299" height="669" alt="Image" src="https://github.com/user-attachments/assets/5c23ba07-788b-4197-ba1f-d41d2660c0c2" />

### Actual Behavior

When I copy a pre-request script directly from Postman and paste it into Insomnia's "Pre-Request" script editor, the request execution hangs and eventually results in a timeout error. The main HTTP request is never sent. The exact same script works perfectly and executes quickly in Postman.

<img width="1631" height="865" alt="Image" src="https://github.com/user-attachments/assets/8c742d02-8792-47f3-8cdc-271bd31007c5" />

### Reproduction Steps

Create a request in Postman with a functional pre-request script blow.
Copy the entire content of the Postman pre-request script.
In Insomnia, create a new HTTP request.
Navigate to the Script tab and select the pre Request section.
Paste the script copied from Postman into the editor.
Click "Send" to execute the request.
Observe that the request loading indicator runs indefinitely until a timeout error is displayed.

postman pre-script：
```
console.log('Pre-request Script from Request start');

if (pm.request.method !== 'GET' && !(pm.request.headers.has('x-csrf-token'))) {

  var csrfRequest = pm.request.clone();
  csrfRequest.method = 'GET';
  if (pm.request.method === 'POST') {
    csrfRequest.url = pm.request.url + '?$top=1';
  }

  csrfRequest.upsertHeader({
    key: 'x-csrf-token',
    value: 'fetch'
  });

  pm.sendRequest(csrfRequest, function(err, res) {
    console.log('pm.sendRequest start');
    if (err) {
      console.log(err);
    } else {
      var csrfToken = res.headers.get('x-csrf-token');
      if (csrfToken) {
        console.log('csrfToken fetched:' + csrfToken);
        pm.request.headers.upsert({
          key: 'x-csrf-token',
          value: csrfToken
        });
      } else {
        console.log('No csrf token fetched');
      }
    }
    console.log('pm.sendRequest end');
  });
}

console.log('Pre-request Script from Request end');
```
COPY TO postman
```
console.log('Pre-request Script from Request start');

if (insomnia.request.method !== 'GET' && !(insomnia.request.headers.has('x-csrf-token'))) {

[...]

Issue #8897, opened 2025-07-23
insomnia.sendRequest is causing the after response script execution forever.


Sravanchitiprolu:
### Expected Behavior

while calling the insomnia.sendRequest, expect to call the API with the details passed and retrun response in the after response script.

### Actual Behavior

It is making the after script execute forever with out and end.

### Reproduction Steps

1. Open the API request
2. Go to script tab
3. Click on after response tab
4. Script below is failing and making the after response script to execute for ever.
Example 1:
		const resp = await 
			insomnia.sendRequest(
				{
					url: '<API URL>',
					method: 'POST',
					headers: [
						{ name: 'Content-Type', value: 'application/json' },
						{ name: 'Authorization', value: authToken }
					],
					body: {
						mimeType: 'application/json',
						text: JSON.stringify(postData)
					}
				}
		);
		console.log('Pricing API response:', resp.statusCode, resp.body);

Example 2: 
		const resp = await new Promise((resolve, reject) => {
			insomnia.sendRequest(
				{
					url: '<API URL>',
					method: 'POST',
					headers: [
						{ name: 'Content-Type', value: 'application/json' },
						{ name: 'Authorization', value: authToken }
					],
					body: {
						mimeType: 'application/json',
						text: JSON.stringify(postData)
					}
				}, (err, resp) => {
							if (err != null) {
									reject(err);
							} else {
									resolve(resp);
							}
					}
			)
		});
Both of this didn't work.

### Is there an existing issue for this?

- [x] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Which sync method do you use?

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [ ] Local only

### Insomnia Version

11.3.0

### What operating system are you using?

macOS

### Operating System Version

15.5

### Installation method

downloaded dmg from insomnia site

ihexxa:
This is because the approach that you are specifying body and headers are not supported, btw both examples didn't specify url, will take a look at the bubbling up error though:

```
const resp = await new Promise((resolve, reject) => {
    insomnia.sendRequest(
        {
            url: 'httpbin.org/anything',
            method: 'POST',
            header: {
                'Content-Type': 'text/plain',
		'Authorization': "authToken",
            },
            body: {
		mode: 'raw',
                raw: JSON.stringify({}),
            }
        },
        (err, resp) => {
            if (err != null) {
                reject(err);
            } else {
                resolve(resp);
            }
[...]

PR #9059, opened 2025-08-26
fix: sendRequest doesnt return error in some cases - INS-1262

ihexxa:
### Changes
- [x] Return error when there's error in sendRequest
- [x] `requestToCurlOptions` always returns request body mime type as the body could be empty

Tested cases from: #9031 #8897

### DIFF
--- a/packages/insomnia-scripting-environment/src/objects/send-request.ts
+++ b/packages/insomnia-scripting-environment/src/objects/send-request.ts
@@ -188,23 +188,23 @@ function requestToCurlOptions(req: string | Request | RequestOptions, settings: 
       settings,
       certificates: finalReq.certificate
         ? [
-          {
-            host: finalReq.certificate?.name || '',
-            passphrase: finalReq.certificate?.passphrase || '',
-            cert: finalReq.certificate?.cert?.src || '',
-            key: finalReq.certificate?.key?.src || '',
-            pfx: finalReq.certificate?.pfx?.src || '',
-            // unused fields because they are not persisted
-            disabled: false,
-            isPrivate: false,
-            _id: '',
-            type: '',
-            parentId: '',
-            modified: 0,
-            created: 0,
-            name: '',
-          },
-        ]
+            {
+              host: finalReq.certificate?.name || '',
+              passphrase: finalReq.certificate?.passphrase || '',
+              cert: finalReq.certificate?.cert?.src || '',
+              key: finalReq.certificate?.key?.src || '',
+              pfx: finalReq.certificate?.pfx?.src || '',
+              // unused fields because they are not persisted
+              disabled: false,
+              isPrivate: false,
+              _id: '',
+              type: '',
+              parentId: '',
+              modified: 0,
+              created: 0,
+              name: '',
+            },
+          ]
         : [],
       caCertficatePath: null, // the request in pre-request script doesn't support customizing ca yet
       socketPath: undefined,
--- a/packages/insomnia-scripting-environment/src/objects/send-request.ts
+++ b/packages/insomnia-scripting-environment/src/objects/send-request.ts
@@ -15,11 +15,11 @@ export async function sendRequest(
   cb: (error?: string, response?: Response) => void,
   settings: Settings,
 ): Promise<Response | undefined> {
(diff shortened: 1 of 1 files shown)


packages/insomnia-scripting-environment/src/objects/send-request.ts
```diff
@@ -30,20 +30,32 @@ export async function sendRequest(
...
+          if (cb) {
+            cb(e, undefined);
+          } else {
+            reject(e);
+          }
         });
     } catch (err: any) {
       if (err.name === 'AbortError') {
-        cb(`Request was cancelled: ${err.message}`, undefined);
+        if (cb) {
```
ZxBing0066: 🤔 From the interface `cb` always exists, is it necessary to check? Or is this only for the script, so users may not pass in cb when they write the script?

ZxBing0066:
Seems there's some boilerplate code. Could we make the code simpler? like:

```ts
const promise = new Promise<Response | undefined>(async (resolve, reject) => {
  try {
    // ...
    resolve(result);
  } catch (e) {
    // ...
    reject(e);
  }
})

if(cb) {
  try {
    const result = await promise;
    cb(undefined, result);
  } catch(e) {
    cb(e);
  }
}

return promise;

```

ZxBing0066 (dismissed):

packages/insomnia-scripting-environment/src/objects/send-request.ts
ihexxa: Good point, as our implementation is in ts while users are writing script in js, it just tries to avoid swallowing error if no function is there.

ihexxa:
> Seems there's some boilerplate code. Could we make the code simpler? like:
> 
> ```ts
> const promise = new Promise<Response | undefined>(async (resolve, reject) => {
>   try {
>     // ...
>     resolve(result);
>   } catch (e) {
>     // ...
>     reject(e);
>   }
> })
> 
> if(cb) {
>   try {
>     const result = await promise;
>     cb(undefined, result);
>   } catch(e) {
>     cb(e);
>   }
> }
> 
> return promise;
> ```

Cool, will refactor it a bit.

jackkav (commented):
This promise callback is an async function, why not just await the response? why use these promise then chains? Very difficult to understand error cases.

ihexxa:
@jackkav Firstly I agreed that await the response is more reasonable, there's background of it, we digged into this before, as the existing interface which only supports callback, which doesn't require `await` while it works in kindof synchronous way, the weird behavior makes the implementation tricky. I will take a look to see if i could make it more straightforward but honestly I'm not sure how much can be improved.

packages/insomnia-scripting-environment/src/objects/send-request.ts
```diff
@@ -15,35 +15,28 @@ export async function sendRequest(
...
+      if (cb) {
+        cb(e, undefined);
+        resolve(undefined);
       } else {
-        cb(`Something went wrong: ${err.message}`, undefined);
+        reject(e);
       }
-      resolve(undefined);
     }
   });
```
ZxBing0066, line 41: Nit: Just realized that, this is an async function, so we could directly remove this promise and use return/throw to replace the resolve/reject.

```suggestion
    try {
      const requestOptions = requestToCurlOptions(request, settings);
      const nodejsCurlRequest =
        process.type === 'renderer'
          ? window.bridge.curlRequest
          : (await import('insomnia/src/main/network/libcurl-promise')).curlRequest;

      const output = (await nodejsCurlRequest(requestOptions)) as CurlRequestOutput;
      const transformedOutput = await curlOutputToResponse(output, request);

      if (cb) {
        cb(undefined, transformedOutput);
      }
      return transformedOutput;
    } catch (e) {
      if (cb) {
        cb(e, undefined);
      } else {
        throw e;
      }
    }
```

ZxBing0066 (approved):

packages/insomnia-scripting-environment/src/objects/send-request.ts
ZxBing0066: Aligned offline with @ihexxa, thanks for the details explanation. 👍

PR #9068, opened 2025-08-28
[INS-1189] Redirect to organization page if project is not found.

gatzjames:
# Overview:

Currently when deleting a project there is a race condition where the project is deleted but some items inside the project e.g. (workspace,request) are not based on the way our database handles deleting related objects.
This leads to some items requesting the project loader data which ends up throwing an error.

This PR fixes the issue where the loader is called with a projectId for a deleted project by redirecting to the parent organization.

## Next steps:
- [ ] Investigate deleting everything under a project when the project is deleted.
- [ ] We are already working on removing using `database` logic from inside the UI which will eliminate this class of issues.


pavkout (approved):
LGTM

PR #9079, opened 2025-09-01
fix: clean unused notification call in the toast

ZxBing0066:
[INS-1296](https://konghq.atlassian.net/browse/INS-1296)

## Background
The notification api has been removed, and the code was never called at all because `checkForNotifications` is only called when there's a notification, then it returns.

## Changes
- Remove `checkForNotifications`

[INS-1296]: https://konghq.atlassian.net/browse/INS-1296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jackkav (approved):
Good initiative! @ZxBing0066

PR #9086, opened 2025-09-03
[INS-1054] Remove .git suffix requirement for custom git repository URL

godfrzero:
Not all providers will use a URL ending in .git. The requirement makes sense for Github/Gitlab since we know which conventions they follow, but not for a form which can connect to any git server. For example, Azure DevOps repo URLs don't end in `.git` and customers report that manually adding it results in a 404.

Note that I have not been able to test cloning an Azure DevOps repo myself, so this PR just makes two changes:
- Modifies validation to no longer require the URL to end in `.git` when adding a URL in the Git—a.k.a. custom repository settings—form.
- Requires at least one character to be present after the URL scheme for the validation to pass.

### DIFF
--- a/packages/insomnia/src/main/git-service.ts
+++ b/packages/insomnia/src/main/git-service.ts
@@ -100,22 +100,22 @@ export function vcsSegmentEventProperties(type: 'git', action: VCSAction, error?
   return { type, action, error };
 }
 
-function parseGitToHttpsURL(s: string) {
+function parseGitToHttpsURL(url: string) {
   // try to convert any git URL to https URL
-  let parsed = fromUrl(s)?.https({ noGitPlus: true }) || '';
+  let parsed = fromUrl(url)?.https({ noGitPlus: true }) || '';
 
   // fallback for self-hosted git servers, see https://github.com/Kong/insomnia/issues/5967
   // and https://github.com/npm/hosted-git-info/issues/11
   if (parsed === '') {
-    let temp = s;
+    let tempURL = url;
     // handle "shorter scp-like syntax"
-    temp = temp.replace(/^git@([^:]+):/, 'https://$1/');
+    tempURL = tempURL.replace(/^git@([^:]+):/, 'https://$1/');
     // handle proper SSH URLs
-    temp = temp.replace(/^ssh:\/\//, 'https://');
+    tempURL = tempURL.replace(/^ssh:\/\//, 'https://');
 
     // final URL fallback for any other git URL
-    temp = new URL(temp).href;
-    parsed = temp;
+    tempURL = (URL.canParse(tempURL) ? URL.parse(tempURL)?.href : url) || '';
+    parsed = tempURL;
   }
 
   return parsed;
--- a/packages/insomnia/src/ui/components/git-credentials/custom-repository-settings-form.tsx
+++ b/packages/insomnia/src/ui/components/git-credentials/custom-repository-settings-form.tsx
@@ -55,7 +55,7 @@ export const CustomRepositorySettingsFormGroup: FunctionComponent<Props> = ({ gi
       <TextField
         name="uri"
         type="url"
-        pattern="https?://.*\.git"
+        pattern="https?://.+"
         autoFocus
         defaultValue={uri}
         onChange={value => setUri(value)}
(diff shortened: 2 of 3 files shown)


packages/insomnia/src/ui/components/git-credentials/custom-repository-settings-form.tsx
```diff
@@ -52,10 +52,10 @@ export const CustomRepositorySettingsFormGroup: FunctionComponent<Props> = ({ gi
       }}
     >
       <TextField name="uri" className="col-span-2 flex w-full flex-col gap-1 px-0.5" isRequired>
-        <Label className="text-start text-sm font-semibold">Git URI (http/https, including .git suffix)</Label>
+        <Label className="text-start text-sm font-semibold">Git URI (http/https, including .git suffix if present)</Label>
         <Input
           type="url"
-          pattern="https?://.*\.git"
+          pattern="https?://.+"
```
pavkout: @godfrzero After testing this PR I noticed the following issue

The pattern we use now `https?://.+` it will match spaces, symbols, or query strings that might not be valid URLs.

- Examples 1

1. https://github.com/foo bar
2. https://github.com/@@@!!

image

image

- Example 2 

Allow multiple URLs
Screenshot 2025-09-04 at 13 50 10
 
Screenshot 2025-09-04 at 13 51 53

It's better to use this pattern `https?://[^\s]+`. This pattern is a safer choice for matching most common URLs with a path, while avoiding false matches or truncated links.

- Requires domain
- Stops at spaces
- Still allows .git or no .git

pavkout:
@godfrzero An alternative approach would be to rely on the browser’s built-in validation instead of enforcing a custom pattern, since the .git suffix is optional. I’ve also applied a small improvement to the parseGitToHttpsURL function to prevent potential crashes. This keeps the implementation simpler and more robust overall.

packages/insomnia/src/ui/components/git-credentials/custom-repository-settings-form.tsx
```diff
@@ -64,10 +65,8 @@ export const CustomRepositorySettingsFormGroup: FunctionComponent<Props> = ({ gi
           className="w-full rounded-sm border border-solid border-[--hl-sm] bg-[--color-bg] py-1 pl-2 pr-7 text-[--color-font] transition-colors placeholder:text-sm placeholder:italic focus:outline-none focus:ring-1 focus:ring-[--hl-md]"
         />
         <FieldError className="text-xs text-[--color-danger]">
-          {({ validationDetails, defaultChildren }) =>
-            validationDetails.patternMismatch
-              ? 'Please ensure the URL is valid and ends with a .git suffix.'
-              : defaultChildren
+          {({ isInvalid }) =>
+            isInvalid ? 'Please ensure the URL is valid. In most cases it should also end with a .git suffix.' : null
```
godfrzero, line 72: What was the purpose of `defaultChildren` here earlier?

pavkout: @godfrzero The defaultChildren is the default error message that would normally be displayed for that validation error.

godfrzero:
Refraining from merging this for now because I'm seeing some strange behavior in the local build. Will test more tomorrow and figure out what's going on.

godfrzero:
> Refraining from merging this for now because I'm seeing some strange behavior in the local build. Will test more tomorrow and figure out what's going on.

This was happening because the `.git` requirement was re-added when #9092 was merged. I've removed the requirement again and the GUI now accepts an Azure DevOps URL, e.g. https://dev.azure.com/fabrikam/Fabrikam-Fiber-Git/_git/Fabrikam

packages/insomnia/src/main/git-service.ts
```diff
@@ -100,22 +100,22 @@ export function vcsSegmentEventProperties(type: 'git', action: VCSAction, error?
...
-    temp = temp.replace(/^git@([^:]+):/, 'https://$1/');
+    tempURL = tempURL.replace(/^git@([^:]+):/, 'https://$1/');
     // handle proper SSH URLs
-    temp = temp.replace(/^ssh:\/\//, 'https://');
+    tempURL = tempURL.replace(/^ssh:\/\//, 'https://');
 
     // final URL fallback for any other git URL
-    temp = new URL(temp).href;
-    parsed = temp;
+    tempURL = (URL.canParse(tempURL) ? URL.parse(tempURL)?.href : url) || '';
```
godfrzero, line 117: @pavkout Should the code not throw an error instead of settling on `''` in case everything fails?

pavkout: @godfrzero This function’s purpose is to parse the given URL. In Node.js, calling `new URL(temp).href` behaves differently than in the browser, which can lead to issues. To handle that discrepancy, we introduced this logic. The actual validation, is performed on the client side.

pavkout (approved):
LGTM

godfrzero:
Verified with an Azure DevOps repo that cloning, pushing, and pulling changes all work as expected.

PR #9092, opened 2025-09-04
[INS-1234] Include token as backup credential when initially fetching repo contents

godfrzero:
# Overview

The `custom` Git tab was not using the correct credentials and was failing because it would send the credentials using the token field while the submit method expected it in the password field.

This PR addresses this issue by changing the input name from `token` to `password` and uses that as the proper credential for connecting to a remote.

Another issue was that the readOnly/disabled properties on the `Input` component do not show the `defaultValue` and therefore the form was showing empty fields on a configured git connection.
These properties are now properly moved to the `TextField` component and work as expected.

Changes were tested on different setups, on both existing and new projects for all possible Git connections (GitHub/GitLab/Custom with a PAT on GitLab)

### DIFF
--- a/packages/insomnia/src/ui/components/git-credentials/custom-repository-settings-form.tsx
+++ b/packages/insomnia/src/ui/components/git-credentials/custom-repository-settings-form.tsx
@@ -13,6 +13,7 @@ export interface Props {
 }
 
 export const CustomRepositorySettingsFormGroup: FunctionComponent<Props> = ({ gitRepository, onSubmit }) => {
+  const isReadOnly = Boolean(gitRepository?.uri);
   const linkIcon = <i className="fa fa-external-link-square" />;
   const defaultValues = gitRepository || {
     uri: '',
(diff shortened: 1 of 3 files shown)


gatzjames (dismissed):

gatzjames (approved):

packages/insomnia/src/ui/components/git-credentials/custom-repository-settings-form.tsx
```diff
@@ -51,15 +52,19 @@ export const CustomRepositorySettingsFormGroup: FunctionComponent<Props> = ({ gi
         });
       }}
     >
-      <TextField name="uri" className="col-span-2 flex w-full flex-col gap-1 px-0.5" isRequired>
+      <TextField
+        name="uri"
+        type="url"
+        pattern="https?://.*\.git"
```
godfrzero, line 58: @gatzjames This will conflict with changes in #9086 will it not?

gatzjames: Yep will make sure I resolve it in the rebase

PR #9098, opened 2025-09-04
[INS-1337] Fix commit-and-push not pushing the changes to the remote

gatzjames:
# Overview

In the Git staging modal, the commit and push functionality doesn't push the changes to the remote as expected.
This issue seems to have happened during a rebase that reverted the way commit-and-push worked to a previous version.

This PR brings back the missing working code that was using the submitter value to check if a push should also happen.


gatzjames:
Closing in favor of #9069

PR #9105, opened 2025-09-05
Chore: Avoid loading bundle plugin errors in development mode

cwangsmv:
**Changes:**

Do not show error stack traces when fails to load bundle plugin in development mode.


jackkav (approved):
:heart:

PR #9107, opened 2025-09-05
chore: audit deps for vulns, add dep overrides for unmaintained packages

ryan-willis:
Bumps:
- esbuild
- eslint-plugin-unicorn 

Overrides:
- tmp
- form-data

Also promotes the dependencies+source from `@getinsomnia/api-client` into the only place it was imported.


jackkav (commented):
I guess we can remove @getinsomnia/api-client in this diff too.

Nice change, lotsa scary CVEs recently

ryan-willis:
superseded by other PRs

PR #9108, opened 2025-09-06
fix: improve env selection/messaging in CI mode and silence warnings [INS-1345]

ryan-willis:
This PR changes how `inso` environments are selected when the `--ci` flag is used, and provides a more detailed message when an environment cannot be chosen.

If a single sub-environment exists within a workspace, it is selected by default when `--ci` is passed. Under the same conditions without the `--ci` argument, the user would be prompted to select it anyway, and is not given the option to proceed with the base environment.

It also silences the `punycode` deprecation warning (and all other Node warnings) to clean up the output a bit


jackkav (approved):
Looking good

PR #9141, opened 2025-09-15
fix: Discard actions now respect staged vs unstaged changes and prevent silent data loss [INS-1099]

pavkout:
This PR addresses an issue where using Discard or Discard All removed both staged and unstaged changes, regardless of the user’s intent. The previous behavior caused accidental loss of work, especially when staged and unstaged changes overlapped on the same line.

Closes INS-1099


gatzjames (approved):
Tested with different scenarios of creating/discarding changes!
LGTM!

PR #9149, opened 2025-09-17
feat: add access grant modal for stdio transport

ZxBing0066:
image


packages/insomnia/src/ui/components/mcp/mcp-url-bar.tsx
```diff
@@ -186,6 +210,155 @@ export const McpUrlActionBar = ({ request, environmentId, defaultValue, onChange
...
+        workspaceId={workspaceId}
+        projectId={projectId}
+        organizationId={organizationId}
+      />
     </>
   );
 };
+
+const isAllowedToRunSTDIO = async (requestId: string, modalRef: React.RefObject<MCPStdioAccessModalHandle>) => {
+  const request = await getMcpRequestById(requestId);
```
CurryYangxx: Can we get the request data from the loader data?

ZxBing0066: Yes, but if so, we need to add the request as a dependence of handleSubmit.

CurryYangxx (approved):

PR #9156, opened 2025-09-18
feat: create collection from mock server routes

shelby-moore:
Depends on https://github.com/Kong/insomnia-mockbin/pull/199

- Adds a feature to create a collection from an existing mock server. Each mock route is translated to a request in the new collection.
- Modifies create/update for mock server routes to allow unique combinations of path and method to be created as a mock route. Previously, path had to be unique for each route for a server. Both create and edit must be done via the same underlying modal. Full validation is performed on the backend. Path/method can no longer be modified in the editor, this is only possible from the modal.


ryan-willis (approved):
dynamite

PR #9157, opened 2025-09-19
feat: Enhance the real time event polling mechanism for websocket and socket-io 

cwangsmv:
# Background 
For real-time events coming from WebSocket and Socket.IO requests, we currently rely on a useInterval hook that polls the file content every 0.5 seconds.
Because each poll recreates the entire events array, the “Events” tab re-renders on every cycle and useMemo provides no benefit.

# Changes
This PR replaces the frequent polling with an observer pattern so that updates are pushed only when new events arrive.
**New workflow**

- Main process
   - When a new event is received from the server and written to the file, invoke a callback.
   - The callback function notifies a predefined channel that a new event is available.

- Render process
  - Add a listener on mount on pre-defined channel
  - When the listener is fired, get the event using existing method `window.main.<protocol>.event.findMany`
  - Update the events and trigger re-render

# Outcome

By moving from time-based polling to an event-driven observer pattern, the “Events” tab updates only when necessary, reducing unnecessary renders and improving overall responsiveness and performance.
This change would also bring huge benefit for the mcp integration since we need to active the latest event automatically but do not break user's selection if exists

Refer [INS-1423](https://konghq.atlassian.net/browse/INS-1423)

[INS-1423]: https://konghq.atlassian.net/browse/INS-1423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/main/network/websocket.ts
+++ b/packages/insomnia/src/main/network/websocket.ts
@@ -1,5 +1,5 @@
 import fs from 'node:fs';
-import { type IncomingMessage, request } from 'node:http';
+import { type IncomingMessage } from 'node:http';
 import path from 'node:path';
 import tls, { type KeyObject, type PxfObject } from 'node:tls';
 
--- a/packages/insomnia/src/main/network/websocket.ts
+++ b/packages/insomnia/src/main/network/websocket.ts
@@ -1,5 +1,5 @@
 import fs from 'node:fs';
-import { type IncomingMessage } from 'node:http';
+import { type IncomingMessage, request } from 'node:http';
 import path from 'node:path';
 import tls, { type KeyObject, type PxfObject } from 'node:tls';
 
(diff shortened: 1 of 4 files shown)


packages/insomnia/src/main/network/websocket.ts
```diff
@@ -354,8 +384,10 @@ const openWebSocketConnection = async (
         type: 'open',
         timestamp: Date.now(),
       };
-
-      eventLogFileStreams.get(options.requestId)?.write(JSON.stringify(openEvent) + '\n');
+      for (const window of BrowserWindow.getAllWindows()) {
+        window.webContents.send(readyStateChannel, ws.readyState === WebSocket.OPEN);
+      }
+      writeEventLogAndNotify({ requestId: options.requestId, data: JSON.stringify(openEvent) + '\n' });
```
CurryYangxx: Repeat with line 396

cwangsmv: fixed

CurryYangxx (approved):

PR #9164, opened 2025-09-22
feat(Git Commit): Automatically generate commits

gatzjames:
# Overview

Insomnia now uses an AI service to automatically generate multiple commits for git.
To achieve this it collects previous commit messages alongside a diff of the current git status and provides it as context to the AI service.

Highlights:
- [x] New bridge API to generate AI commits using `insomnia-ai`
- [x] New utility process to run the plugin for commit generation.
- [x] Expanded the git service to handle multiple commits
- [x] Drag-n-drop UX to move files between commits.
- [x] Allows editing commit messages for each generated commit
- [x] Do not commit section where the user can drag-n-drop files they don't want to include in the generated commits
- [x] Simplified the `git-project-staging-modal.tsx` code to handle the added complexity
- [x] Added a new form for the generated-commits that follows the same UX (handles pull/push if needed) with the manual commit flow.


packages/insomnia/src/routes/ai.generate-commit-messages.tsx
```diff
@@ -0,0 +1,72 @@
...
+  try {
+    const { changes } = await window.main.git.gitChangesLoader({ projectId });
+    if (changes.staged.length > 0) {
+      return {
+        error: 'You have staged changes. Please commit or unstage them and try again.',
+      };
+    }
+    const diff = await window.main.git.diff();
+
+    console.log('Diff for AI:\n', diff);
```
shelby-moore: Should we remove this?

packages/insomnia/src/main/ipc/main.ts
```diff
@@ -381,4 +387,21 @@ export function registerMainHandlers() {
       }
     },
   );
+
+  ipcMainHandle('generateCommitsFromDiff', async (_, input) => {
+    try {
+      const { generateCommitsFromDiff } = await import('@kong/insomnia-plugin-ai');
```
shelby-moore: Should we move the interaction with the plugin to a utility process? We've done that with the mock server gen in https://github.com/Kong/insomnia/pull/9210

packages/insomnia/src/routes/ai.generate-commit-messages.tsx
```diff
@@ -0,0 +1,72 @@
...
+        id: crypto.randomUUID(),
+        ...commit,
+      })),
+    };
+  } catch (err) {
+    showToast({
+      title: 'Failed to generate commit messages',
+      icon: 'star',
+      status: 'error',
+      description: 'There was an error communicating with the AI service. Please try again.',
```
shelby-moore: Should we surface the actual error? If it's an issue with something like the API key it would be preferable to surface that so it can be resolved before trying again

packages/insomnia/src/ui/components/modals/git-project-staging-modal.tsx
```diff
@@ -71,6 +76,729 @@ function getModificationClassName(type: string) {
...
+  id: string;
+  commitsSections: TreeData<CommitItem>;
+  files: TreeData<CommitItem>['items'];
+  emptyState?: ReactNode;
+  diffChanges: (params: { path: string; staged: boolean }) => void;
+}) => {
+  const { dragAndDropHooks } = useDragAndDrop({
+    // Provide drag data in a custom format as well as plain text.
+    getItems(keys) {
+      console.log('getItems', keys);
```
shelby-moore: Are the console.logs in here intentional?

gatzjames: Only for debugging 😅

packages/insomnia/src/main/ipc/main.ts
gatzjames: ✅ Done. Not sure if we want to spawn multiple utility processes for each ai feature or have a long running process instead. It's worth investigating pros/cons on the approach

packages/insomnia/src/routes/ai.generate-commit-messages.tsx
gatzjames: I appended the error to the message.
Since these errors can be very low level we might want to keep the user facing message simple and add the whole trace and complex error to the console instead.
Also we have been discussing on surfacing all app errors in a more user friendly error panel alongside some error codes/remediation strategies as future work

packages/insomnia/src/main/ipc/main.ts
shelby-moore: Agree, do you want to make sure we're tracking that somewhere?

packages/insomnia/src/routes/ai.generate-commit-messages.tsx
shelby-moore: >Since these errors can be very low level we might want to keep the user facing message simple and add the whole trace and complex error to the console instead.

Yeah, I'm good with that too. The main problem I want to solve is giving the user enough information to be transparent about what is failing, in case it's something on their side.

shelby-moore (approved):
PR looks good, I think we need to publish 1.0.3 of the plugin to resolve the failing type check CI.

PR #9177, opened 2025-09-26
customize more template

CurryYangxx:
Add more custom template, support for json schema `additionalProperties`, `oneof`, `allof`, `anyof`.

### DIFF
--- a/packages/insomnia/src/ui/components/base/select.tsx
+++ b/packages/insomnia/src/ui/components/base/select.tsx
@@ -17,9 +17,10 @@ interface SelectProps {
   onChange: (value: string | number) => void;
   className?: string;
   options: { label: string; value: string }[];
+  multiple?: boolean;
 }
 
-export const Select = ({ value, onChange, className, options, ...rest }: SelectProps & RaSelectProps) => {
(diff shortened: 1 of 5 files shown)


packages/insomnia/src/ui/components/base/select.tsx
```diff
@@ -17,9 +17,10 @@ interface SelectProps {
   onChange: (value: string | number) => void;
   className?: string;
   options: { label: string; value: string }[];
+  multiple?: boolean;
 }
 
-export const Select = ({ value, onChange, className, options, ...rest }: SelectProps & RaSelectProps) => {
+export const Select = ({ value, onChange, className, options, multiple, ...rest }: SelectProps & RaSelectProps) => {
```
ZxBing0066: It seems `multiple` is never used?

ZxBing0066 (approved):

packages/insomnia/src/ui/components/base/select.tsx
CurryYangxx: Good catch! Deleted it.

cwangsmv (approved):

PR #9183, opened 2025-09-28
perf: parallelize data fetching

mohd-akram:
This improves loader performance by 2x.


mohd-akram:
Thoughts on this PR? This loader is called a lot so any optimization helps, especially on less capable machines.

mohd-akram:
Any update on this?

PR #9190, opened 2025-09-29
feat: Improve global modals

gatzjames:
# Overview

Currently global modals in `modals.tsx` are using a unique mechanism to render modals in the app.
This PR adds the react-aria modal to those so that they behave similarly to all other modals in the app.

Next steps:
- [ ] There is a lot of complexity in sub-components like ModalHeader/Footer etc. This logic can be simplified or even removed for now since it introduces unnecessary complexity.
- [ ] Some of the modals in `modals.tsx` are only used by one component in the app and could be inlined instead
- [ ] Truly global modals like `AlertModal` could have a much simpler API (e.g. something similar to showToast(...)) instead of the current implementation.


jackkav (approved):
happy to throw this one in quick

PR #9196, opened 2025-09-30
fix: mcp issue

CurryYangxx:
**Fix:**
- Delete McpRequest and McpResponse when deleting a MCP workspace
- Error when getting readyState if MCP is not connected
- Repeat MCP payload creation


cwangsmv (approved):
lgtm

PR #9220, opened 2025-10-08
fix: ensure data dir exists prior to reading

ryan-willis:
Adds a new IPC handler to read a folder within the Insomnia data directory, and catches errors thrown to produce an empty list when enumerating `.gguf` files

### DIFF
--- a/packages/insomnia/src/main/ipc/main.ts
+++ b/packages/insomnia/src/main/ipc/main.ts
@@ -96,7 +96,9 @@ export interface RendererToMainBridgeAPI {
   writeFile: (options: { path: string; content: string }) => Promise<string>;
   readFile: (options: { path: string; encoding?: string }) => Promise<{ content: string; encoding: string }>;
   readDir: (options: { path: string }) => Promise<{ type: 'file' | 'directory'; name: string; path: string }[]>;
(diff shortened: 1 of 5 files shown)


packages/insomnia/src/main/ipc/main.ts
```diff
@@ -251,22 +270,13 @@ export function registerMainHandlers() {
...
-        .filter(file => file.type !== 'other');
-    } catch (err) {
-      throw new Error(`Failed to read directory: ${err}`);
-    }
+  ipcMainHandle('readDir', readDir);
+
+  ipcMainHandle('readDataDir', async (_, options: { folder: string }) => {
+    const dataPath = app.getPath('userData');
+    const folderPath = path.join(dataPath, options.folder);
+    mkdirSync(folderPath, { recursive: true });
```
shelby-moore, line 280: It's unexpected to see a function called readDataDir make the directory. Should we instead have a function to check directory existence and use that to gate reading the LLMs dir?

ryan-willis: lemme rename this to be a lil more accurate

shelby-moore (approved):

PR #9231, opened 2025-10-09
fix(deps): Move @react-router/dev to dev dependencies

gatzjames:
# Overview

- [x] Moves @react-router/dev to dev dependencies since it shouldn't be included in the final package
- [x] Removes the `x64ArchFiles` option from electron builder since it was added to support bundling this package


CurryYangxx (dismissed):

ryan-willis (approved):
We're going to be adding that `x64ArchFiles` option back in with the new plugin

PR #9232, opened 2025-10-10
feat: toggle ai features

shelby-moore:
Adds the ability for a user to toggle AI features on off. There are currently two AI features - one for mock server generation, and another for commit message generation. Both can be toggled on/off individually.

This PR also disables the functionality for using the AI features depending on whether or not they're toggled on/off. If the features have been toggled off at the organization level, the user cannot toggle them on/off, the toggles are simply disabled along with the features.


jackkav (commented):
I have some thoughts about terminology. 
Mock servers are currently an entity under workspace, which has within it a number of mock routes. This work appears to be introducing a new definition of mock servers and an idea that a mock server can be enabled and fully enabled. For the future, lets come up with some more specific names. Happy to join a call and spit ball some ideas. Might help me give a better review too.
I understand that this work has some time pressure an I don't mean to hold it up over a nuance, but this kind of nuance is less painful to add now than later.

shelby-moore:
>This work appears to be introducing a new definition of mock servers and an idea that a mock server can be enabled and fully enabled. For the future, lets come up with some more specific names. Happy to join a call and spit ball some ideas. Might help me give a better review too.

I don't think that's the case - these toggles are turning the AI generation for _creating_ a mock server on/off. It's not actually enabling a mock server itself. I can work on renaming to make this more clear

ryan-willis (approved):

PR #9249, opened 2025-10-11
fix: sentry sourcemap - [INS-1255]

CurryYangxx:
Upload sourcemaps to Sentry

- pack all JS resources into the sourcemap artifact
- correct the artifact name for Sentry sourcemaps upload

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -277,10 +277,8 @@ jobs:
         with:
           name: ${{ runner.os }}-${{ runner.arch }}-sentry
           path: |
-            packages/insomnia/build/*.js
-            packages/insomnia/build/*.map
-            packages/insomnia/build/client/assets/*.js
-            packages/insomnia/build/client/assets/*.map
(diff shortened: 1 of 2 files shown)


.github/workflows/release-build.yml
```diff
@@ -279,6 +279,8 @@ jobs:
           path: |
             packages/insomnia/build/*.js
             packages/insomnia/build/*.map
+            packages/insomnia/build/client/assets/*.js
+            packages/insomnia/build/client/assets/*.map
```
ZxBing0066: Could we use `packages/insomnia/build/**/*.js` to avoid adding new paths in the future?

ZxBing0066 (dismissed):

jackkav (dismissed):

.github/workflows/release-build.yml
CurryYangxx: Yes, we don't need to care about the subpath if we change to `packages/insomnia/build/**/*.js`

ZxBing0066 (approved):

PR #9279, opened 2025-10-20
fix: Enhance discardChanges to support unstaged and staged options [INS-1099]

pavkout:
Fixes an issue where discarding unstaged changes also removed staged changes, ensuring correct separation of Git state and preserving intended user actions.

### DIFF
--- a/packages/insomnia/src/sync/git/git-vcs.ts
+++ b/packages/insomnia/src/sync/git/git-vcs.ts
@@ -1893,7 +1893,7 @@ export class GitVCS {
     }
   }
 
-  async discardChanges(changes: { path: string; status: [git.HeadStatus, git.WorkdirStatus, git.StageStatus] }[]) {
+  async discardChanges(changes: { path: string; status: Status }[]) {
     for (const change of changes) {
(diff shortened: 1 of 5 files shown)


packages/insomnia/src/sync/git/git-vcs.ts
```diff
@@ -1747,20 +1747,88 @@ ${formatDiffChanges(status, 'Unstaged Changes')}`;
     }
   }
 
-  async discardChanges(changes: { path: string; status: [git.HeadStatus, git.WorkdirStatus, git.StageStatus] }[]) {
+  async discardChanges(
+    changes: { path: string; status: [git.HeadStatus, git.WorkdirStatus, git.StageStatus] }[],
+    options?: { discardStaged?: boolean; discardUnstaged?: boolean },
```
gatzjames: Since we can only discard unstaged changes should we remove the options.discardStaged option and always perform that action?
Imo if in the future we find that we need this functionality we can always add that extra option and code to handle this case.

gatzjames (approved):

PR #9289, opened 2025-10-22
fix: support the encrypt auth flow for mcp auth

ZxBing0066:
## Background

The mcp auth flow depends on the `authorizeUserInDefaultBrowser` but doesn't fit the new encrypt flow.

## Changes

Extract the encrypt flow and adopt into the mcp


yaoweiprc (approved):
LGTM

PR #9293, opened 2025-10-22
feat: track commit generation and saving events for analytics

gatzjames:
# Overview

Adds segment events for AI generated commits

- [x] recommend_commits_clicked - when the user initially clicked the generate button
- [x] recommend_commits_generated - when the results come back from the LLM. Includes **file_count** and **group_count** (number of groups it generated), **time_to_generate** in seconds and a **has_error** boolean indicator if there where any errors in the generation
- [x] recommend_commits_cancelled - when they user clicked the “Back to Manual” button
- [x] recommend_commits_saved - when the user actually submitted the generated commits. Includes the **group_count** (the number of groups they actually committed) and **file_excluded_count** (how many files they excluded)

### DIFF
--- a/packages/insomnia/src/routes/ai.generate-commit-messages.tsx
+++ b/packages/insomnia/src/routes/ai.generate-commit-messages.tsx
@@ -39,7 +39,7 @@ export async function clientAction(args: Route.ClientActionArgs) {
     });
 
     window.main.trackSegmentEvent({
-      event: SegmentEvent.recommend_commits_generated,
+      event: SegmentEvent.recommendCommitsGenerated,
       properties: {
         file_count: commits?.map(commit => commit.files?.length || 0)?.reduce((a, b) => a + b, 0),
         group_count: commits?.length || 0,
--- a/packages/insomnia/src/routes/ai.generate-commit-messages.tsx
+++ b/packages/insomnia/src/routes/ai.generate-commit-messages.tsx
@@ -1,5 +1,6 @@
 import { href } from 'react-router';
 
+import { SegmentEvent } from '~/ui/analytics';
 import { showToast } from '~/ui/components/toast-notification';
 import { createFetcherSubmitHook } from '~/utils/router';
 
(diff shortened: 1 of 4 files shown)


packages/insomnia/src/routes/ai.generate-commit-messages.tsx
```diff
@@ -36,6 +37,16 @@ export async function clientAction(args: Route.ClientActionArgs) {
         .join('\n'),
     });
 
+    window.main.trackSegmentEvent({
+      event: 'recommend_commits_generated',
+      properties: {
+        file_count: commits?.map(commit => commit.files?.length || 0)?.reduce((a, b) => a + b, 0),
+        group_count: commits?.length || 0,
+        time_to_generate: (performance.now() - startTime) / 1000,
```
shelby-moore: Should we include the unit to clarify it's in seconds?

packages/insomnia/src/routes/ai.generate-commit-messages.tsx
```diff
@@ -36,6 +37,16 @@ export async function clientAction(args: Route.ClientActionArgs) {
         .join('\n'),
     });
 
+    window.main.trackSegmentEvent({
+      event: 'recommend_commits_generated',
```
shelby-moore: We have an enum for `SegmentEvent` with each event type, should we follow that approach here?

packages/insomnia/src/ui/analytics.ts
```diff
@@ -50,6 +50,10 @@ export enum SegmentEvent {
   projectUpdated = 'Project Updated',
   exportStarted = 'Export Started',
   exportRequestsChosen = 'Export Requests Chosen',
+  recommend_commits_generated = 'Recommend Commits Generated',
```
shelby-moore: Should we use the same casing as the rest of the enum vals for consistency?

gatzjames: Good catch I forgot to check the other values 😅

shelby-moore (approved):

PR #9302, opened 2025-10-23
fix: dockerAsync shell command built from environment values 

odaysec:
https://github.com/Kong/insomnia/blob/bae945281b047d215e765cbf340a4212a756dc68/packages/insomnia/customSign.js#L1-L4
https://github.com/Kong/insomnia/blob/bae945281b047d215e765cbf340a4212a756dc68/packages/insomnia/customSign.js#L44-L52

fix this problem, remove string-based shell interpolation with `exec`, and switch to `child_process.spawn` (or `execFile`) with argument arrays. Specifically, the docker command line must be split into command and arguments and each dynamic value provided as a separate array element, avoiding unintended shell interpretation.

**Steps:**
- Replace `execAsync(dockerCommand)` with a call to `spawn` or `execFile` (preferably promisified).
- Split the constructed docker command into a command (`docker`) and an array of arguments (including `run`, `--rm`, `-v`, etc.).
- Pass dynamic values (`directoryPath`, environment vars, etc.) as array arguments, not within interpolated strings.
- Properly handle stdout/stderr as before.
- Add necessary import (`promisify`), and update function to await the spawned process.

### DIFF
--- a/packages/insomnia/customSign.js
+++ b/packages/insomnia/customSign.js
@@ -1,7 +1,7 @@
-const { exec } = require('child_process');
+const { spawn } = require('child_process');
 const util = require('util');
 const path = require('path');
-const execAsync = util.promisify(exec);
+// No need for execAsync now.
 
 // adapted from https://www.electron.build/win.html#how-do-delegate-code-signing
 // It was possible code-sign installer after packaging, but some files are only available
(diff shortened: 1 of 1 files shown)


poratoes (approved):

jackkav (approved):
Cool contribution thanks, this code is only run by electron-builder during the packaging process, I don't see an exploit vector, but it looks like it adheres to better secure coding practices.

PR #9306, opened 2025-10-23
feat: sparkles badge for AI features

ryan-willis:
In the preferences tab:
Screenshot 2025-10-24 at 08 22 55

In the mock server dialog:
Screenshot 2025-10-24 at 08 23 40

I didn't add it to the git commit feature, that one is a little more complicated, but this will at least make the component available


packages/insomnia/src/ui/components/base/badge.tsx
```diff
@@ -0,0 +1,19 @@
...
+  color: string;
+  icon?: React.ReactNode;
+  label: string;
+  textSize?: string;
+}
+
+export const Badge: FC<BadgeProps> = ({ color, icon, label, textSize = '0.8rem' }) => {
+  return (
+    <span
+      className={`mr-0.5 rounded-[2.5px] border-2 border-solid px-[0.25rem] py-[0.1rem] text-[${textSize}] text-[--color-${color}] border--[rgb(--color-${color}-rgb)]`}
```
gatzjames: The dynamic classes will only work if the combination of text + variable is already defined somewhere else in the code.
e.g. if text-[0.8rem] exists elsewhere tailwind will create a class and it will work here
If text-[5rem] doesn't exist in the code as a static string then tailwind will not create the class and it won't have any effect 😅

- For text-[size] could we use sm/base/lg etc? e.g. `{sm: 'text-sm', base: 'text-base'}[size]`
- For text and border color we could pass it as inline style since it's already a var
`style={{color: '--color-${color}'}}`

shelby-moore (dismissed):

packages/insomnia/src/ui/components/base/badge.tsx
gatzjames: > For text-[size] could we use sm/base/lg etc? e.g. {sm: 'text-sm', base: 'text-base'}[size]

Scratch that since we use rems we could inline-style all of the dynamic props imo 😅

ryan-willis: I've adjusted it to inherit the text size, and inlined all of the styles since they're oddly precise

godfrzero: Something to note about badges, icons, and similar UI elements: It's best to leave Tailwind behind and focus on precision. Visual aligment tends to be more important than numeric symmetry. I.e. we need a finely tuned "AI Badge" component rather than a "badge" component with an icon + text specified.

On a separate note, I believe we should be able to dynamically color the SVG (e.g. using currentColor?) so this doesn't happen:
image

gatzjames (approved):

PR #9313, opened 2025-10-27
[INS-1598] fix: remove unnecessary padding from feature list in onboarding view

gatzjames:
# Overview:

- [x] Removes extra padding from onboarding feature list
- [x] The height of the feature list is the same as the height of the feature preview

| Before | After |
|--------|-------|
| image | image |

### DIFF
--- a/packages/insomnia/src/routes/onboarding.$.tsx
+++ b/packages/insomnia/src/routes/onboarding.$.tsx
@@ -56,11 +56,11 @@ const FeatureWizardView = () => {
       <Route
         path="/"
         element={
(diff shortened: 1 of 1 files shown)


ryan-willis (dismissed):
I kinda think it looked better the way it was 🤷🏻

gatzjames:
> I kinda think it looked better the way it was 🤷🏻

I don't have a strong preference tbh. Also think the "before" is good enough until we come up with a new version of feature cards.

ryan-willis (approved):

gatzjames:
Closing as we will keep the current design as-is and update it in a future version

PR #9319, opened 2025-10-28
feat: update commit generation area

ryan-willis:
https://github.com/user-attachments/assets/dd617c32-a5bc-4b10-bf29-f687438aba4a


gatzjames (approved):
I'd fix the dep issue but other than that looks 🔥

PR #9348, opened 2025-11-06
Fix: Show oauth related error only when discover authorization server and dynamic register client success

cwangsmv:
Enter oauth workflow only when:
- By default no auth has been selected (no `type` property in authentication)
- OAuth type auth and enable it

[INS-1690](https://konghq.atlassian.net/browse/INS-1690)

[INS-1690]: https://konghq.atlassian.net/browse/INS-1690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/main/mcp/transport-streamable-http.ts
+++ b/packages/insomnia/src/main/mcp/transport-streamable-http.ts
@@ -9,12 +9,13 @@ import { isInitializeRequest } from '@modelcontextprotocol/sdk/types.js';
 import { BrowserWindow } from 'electron';
 
 import { timelineFileStreams, writeEventLogAndNotify } from '~/main/mcp/common';
-import type { McpOAuthClientProvider } from '~/main/mcp/oauth-client-provider';
+import { type McpOAuthClientProvider } from '~/main/mcp/oauth-client-provider';
 import type { McpAuthEventWithoutBase, OpenMcpHTTPClientConnectionOptions } from '~/main/mcp/types';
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/main/mcp/transport-streamable-http.ts
```diff
@@ -163,6 +164,20 @@ const wrappedFetch = async (
...
+    const mcpRequest = await models.mcpRequest.getById(requestId);
+    invariant(mcpRequest, 'MCP Request not found');
+    const { authentication } = mcpRequest;
+    // By default no authentication is set, authentication is an empty object
+    const isDefaultAuth = !('type' in authentication);
+    // Continue to oauth workflow only when the auth type is oauth and enable it
+    if (!isDefaultAuth) {
+      const isOauth2AuthType = authentication.type === 'oauth2';
+      const isAuthTypeEnabled = !authentication.disabled;
+      if (!(isOauth2AuthType && isAuthTypeEnabled)) {
```
ZxBing0066: Should we also check if auth.grantType === mcp_auth_flow here?

cwangsmv: good catch, only for mcp oauth to enter oauth workflow

ZxBing0066 (approved):

PR #9357, opened 2025-11-07
feat: add model config to segment even for mock creation

shelby-moore:
Adds the AI provider and model used to create a Mock Server to the segment event for Mock Server Creation.


ryan-willis (approved):

gatzjames (approved):
🚀

PR #9358, opened 2025-11-08
fix: prevent duplicate cookies due to lost id

### LINKED ISSUE #8791: Cookies not overwriting
### Expected Behavior

When a Set-Cookie header is sent by server, after a cookie with the same name exists in the client, Insomnia creates a second cookie rather than overwriting the existing cookie.

### Actual Behavior

When using insomnia to test api's that deal with login, i've seen a problem when the user sets the auth cookie more than once - Insomnia does not handle cookies properly. The expected behavior is for insomnia to overwrite the cookie, instead it adds it to the cookie store as an additional cookie.

### Reproduction Steps

1. Create a server which sets a different cookie on each request.
2. Send request 1: cookie A1 will be returned.
3. Send request 2: (cookie A1 will be sent) - cookie A2 will be returned.
4. Send request 3: cookie A1 will still be sent

- [ ] Git sync.
- [x] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

_No response_

### Insomnia Version

11.2.0

macOS

### Operating System Version

macos sequoia 15.3.1

### Installation method

insomnia website

_No response_

autoboxer:
### Context

When processing Set-Cookie headers, Insomnia was creating duplicate cookies instead of updating existing ones with the same name, domain, and path. This occurred because the custom `id` field was lost during the `tough-cookie` round-trip, causing each updated cookie to receive a new `id`.

The fix preserves original cookie `id`s by mapping them before `tough-cookie` processing and restoring them afterward, ensuring cookies are properly updated rather than duplicated.

Closes #8791

### Fix Details

Logic that caused duplicate cookies:

1. Insomnia stores cookies with a custom `id` field
2. when processing Set-Cookie headers, `jarFromCookies()` creates a `tough-cookie` jar from existing cookies
3. `tough-cookie` stores the `id` field as a custom property on the `Cookie` instance
4. when `setCookieSync()` is called, tough-cookie correctly overwrites the cookie
5. when `cookiesFromJar()` extracts cookies back out, tough-cookie's `toJSON()` excludes the custom `id` field
6. each cookie gets a new `id` assigned via `migrateCookieId()`
7. Insomnia sees these as different cookies because they have different `id`s

The fix:

1. before calling `jarFromCookies()`, create a map storing existing cookie `id`s keyed by `domain|path|key`
2. after calling `cookiesFromJar()`, restore the original `id`s from the map for cookies that match existing ones
3. when the cookie jar is saved, `migrateCookieId()` only assigns new `id`s to cookies that don't already have one
4. existing cookies retain their original `id`s, so they update instead of duplicate

NOTE: This change created the knock-on effect of rendering the cookie count in the UI incorrectly due to cookies not being deduplicated when merged.  This fixes that issue as well.


autoboxer:
@jackkav, I apologize for pushing on this, but if you would consider taking a look at this fix, it would be appreciated.  This PR is focused on a single issue that's been causing me and other users real headaches.

PR #9360, opened 2025-11-10
feat: support saving run collection result report

### LINKED ISSUE #273: Inso cli - collection run - global reporter and results save
### Code of Conduct

- [x] I have read and agree to the Insomnia Docs project's [Code of Conduct](https://github.com/Kong/insomnia-docs/blob/main/CODE_OF_CONDUCT.md).

### What article on docs.insomnia.rest is affected?

[Inso run collection options](https://docs.insomnia.rest/inso-cli/cli-command-reference/inso-run-collection)

### What part(s) of the article would you like to see updated?

Hi, 

I thought that it may be helpful to have the ability to see at the end of the inso cli run an overall summary regarding the tests such as the number of total checks and the number of failed and passed. This gives quicklier an overview of how much was hood and how bad so that we don't have to scroll on the entire run log in the terminal to see for each request/endpoint how many failed and passed.

Also it may be of benefit to us to have the abbility to save the results into a json file or xml so we also canhave the ability to store the results for further look and integrate better with Ci/CD pipelines. A big example here can be Azure which has the test plan and with a json file it's easy to make a script to update the tests, in the test plan.

What are your thoughts on this topic?

P.S.: In the docs/other issues I have not seen anything alike for collection run, not can use -o .
 

### Additional information

Quick current example: got a lot of iterations on my tests and I'd prefer to see at the end a summary of it...

Image

### Code of Conduct and Community Expectations

- [x] I agree to follow this project's Code of Conduct
- [x] I agree to abide by the Community Expectations

ZxBing0066:
## Background
[INS-554](https://konghq.atlassian.net/browse/INS-554)

Add a result report for the inso run collection

## Changes

- Add `--output` to save the report, `--includeFullData` to specify the data to be included in the report, `--acceptRisk` to use in the CI/CD
- Only include necessary data when default, and redact known sensitive data when `--includeFullData=redact`

closes Kong/insomnia-docs#273

[INS-554]: https://konghq.atlassian.net/browse/INS-554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia-inso/src/cli.ts
+++ b/packages/insomnia-inso/src/cli.ts
@@ -1,12 +1,15 @@
 import fs from 'node:fs';
 import { readFile } from 'node:fs/promises';
 import { homedir } from 'node:os';
-import path from 'node:path';
+import path, { dirname } from 'node:path';
 
 import * as commander from 'commander';
 import type { logType } from 'consola';
 import consola, { BasicReporter, FancyReporter, LogLevel } from 'consola';
 import { cosmiconfig } from 'cosmiconfig';
+// @ts-expect-error the enquirer types are incomplete https://github.com/enquirer/enquirer/pull/307
+import { Confirm } from 'enquirer';
+import { pick } from 'es-toolkit';
 import { isDevelopment, JSON_ORDER_PREFIX, JSON_ORDER_SEPARATOR } from 'insomnia/src/common/constants';
 import { getSendRequestCallbackMemDb } from 'insomnia/src/common/send-request';
 import type { Environment, UserUploadEnvironment } from 'insomnia/src/models/environment';
@@ -19,10 +22,13 @@ import orderedJSON from 'json-order';
 import { parseArgsStringToArgv } from 'string-argv';
 import { v4 as uuidv4 } from 'uuid';
 
+import type { Workspace } from '~/models/workspace';
+
 import { type RequestTestResult } from '../../insomnia-scripting-environment/src/objects';
 import packageJson from '../package.json';
 import { exportSpecification, writeFileWithCliOptions } from './commands/export-specification';
 import { getRuleSetFileFromFolderByFilename, lintSpecification } from './commands/lint-specification';
+import { RunCollectionResultReport } from './commands/run-collection/result-report';
 import type { Database } from './db';
 import { isFile, loadDb } from './db';
 import { insomniaExportAdapter } from './db/adapters/insomnia-adapter';
@@ -140,7 +146,7 @@ export const getDefaultProductName = (): string => {
 const localAppDir = getAppDataDir(getDefaultProductName());
 
 export const getAbsoluteFilePath = ({ workingDir, file }: { workingDir?: string; file: string }) => {
-  return file && path.resolve(workingDir || process.cwd(), file);
+  return file && path.resolve(workingDir ? dirname(workingDir) : process.cwd(), file);
 };
 export const logErrorAndExit = (err?: Error) => {
   if (err instanceof InsoError) {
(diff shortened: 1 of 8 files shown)


packages/insomnia/src/common/send-request.ts
```diff
@@ -130,9 +130,9 @@ export async function getSendRequestCallbackMemDb(
     }
     const { statusCode: status, statusMessage, headers: headerArray, elapsedTime: responseTime } = res;
 
-    const headers = headerArray?.reduce(
+    const headers = headerArray?.reduce<Record<string, string>>(
       (acc, { name, value }) => ({ ...acc, [name.toLowerCase() || '']: value || '' }),
-      [],
+      {},
```
ZxBing0066, line 135: This was a bug since headers should be an object, not an array.

packages/insomnia-inso/src/cli.ts
```diff
@@ -140,7 +146,18 @@ export const getDefaultProductName = (): string => {
 const localAppDir = getAppDataDir(getDefaultProductName());
 
 export const getAbsoluteFilePath = ({ workingDir, file }: { workingDir?: string; file: string }) => {
-  return file && path.resolve(workingDir || process.cwd(), file);
+  if (!file) {
```
ZxBing0066, line 149: There was a bug when the workingDir is a file, `getAbsoluteFilePath({ workingDir: '/workdir/insomnia.yml', file: 'filename.ext' })` returns `/workdir/insomnia.yml/filename.ext`.

packages/insomnia-inso/src/cli.ts
```diff
@@ -552,12 +581,57 @@ export const go = (args?: string[]) => {
...
+            try {
+              fs.accessSync(outputFilePath, fs.constants.W_OK);
+            } catch (err) {
+              logger.fatal(`Output file "${outputFilePath}" is not writable.`);
+              return process.exit(1);
+            }
+          }
+
+          // Show security disclaimer when outputting to a file with data
+          if (options.includeFullData && !options.acceptRisk) {
```
gatzjames, line 610: Is it necessary to add this many options for the cli?
Could we have a different option for output with full data?
e.g. `inso ... --output=somefile.txt` and `inso ... --unsafe-complete-output=somefile.txt`
I don't think `--acceptRisk` is easy to understand from a user's perspective.

ZxBing0066: Do you mean
- `--output=file.ext` to export the default report
- `--unsafe-redact-output=file.ext` to export the redacted report
- `--unsafe-complete-output=file.ext` to export the full report

Good point. Both are fine. I think the current one would be more flexible if we need to add more options for the report in the future.

> I don't think --acceptRisk is easy to understand from a user's perspective.
How about `--acceptUnsafeOutput`? Do you have any suggestions?

ZxBing0066: BTW, when there's no `--acceptRisk`, CI will show a risk disclaimer. And `--acceptRisk` is considered for the CI/CD run.
image

So even with the new `--unsafe` options, I think we'll also need another `--acceptRisk` option to explicitly hide the risk disclaimer.

packages/insomnia-inso/src/cli.ts
```diff
@@ -552,12 +581,57 @@ export const go = (args?: string[]) => {
...
+            const disclaimerMessage = [
+              'SECURITY WARNING',
+              'Outputting to a file could contain sensitive data like API tokens or secrets. Make sure you understand this, and the contents of your collection, before proceeding.',
+              'Are you sure you want to continue?',
```
ihexxa, line 614: nit: probably "Are you sure to continue?"

ihexxa (approved):
LGTM, just one minor comment.

packages/insomnia-inso/src/cli.ts
ZxBing0066: I've confirmed with GPT about the grammar, and it's OK. Since it has been reviewed, I'd like to keep the current one.

Issue #7713, opened 2024-07-12
pre request autocomplete with key `tab`


arthurvanl:
### Expected Behavior

To be able to use `TAB` key as vscode does 

### Actual Behavior

Adds spaces instead of the autocomplete I want to use.

### Reproduction Steps

1. Go to pre request script
2. Type `insomnia.`
3. Use key `TAB`

### Is there an existing issue for this?

- [x] I have searched the [issue tracker](https://www.github.com/Kong/insomnia/issues) for this problem.

### Additional Information

it does work with key `ENTER` but not with `TAB`. Would appreciate if it could be added

### Insomnia Version

v9.3.2

### What operating system are you using?

Windows

### Operating System Version

Windows 10 Pro 22H2 19045.4651

### Installation method

download from website

pavkout:
@arthurvanl Thanks for reporting this issue. Can you please clarify what is the behavior you expected? vscode TAB key behavior is also to add spaces.

arthurvanl:
> @arthurvanl Thanks for reporting this issue. Can you please clarify what is the behavior you expected? vscode TAB key behavior is also to add spaces. 

When typing in pre request script there is a intellisense screen popping up. In vscodd I am able to use tab to complete it

arthurvanl:
bump

gowrimanoharir:
@pavkout Looks like the default code mirror behavior was overridden in the original PR when autocomplete was implemented

https://github.com/Kong/insomnia/pull/112

<img width="398" height="36" alt="Image" src="https://github.com/user-attachments/assets/f738cb25-9f90-4b64-adc0-7bb9d8d5e8ab" />

Just need this override to be removed to allow tab to have default behavior accepting auto complete selection if one is available. Please let me know if I can create a PR?

https://github.com/Kong/insomnia/blob/3040436e8bc9443ee6a4dd56679cc839f6a88baa/packages/insomnia/src/ui/components/.client/codemirror/extensions/autocomplete.ts#L107

pavkout:
Hey @gowrimanoharir, you are right, I just checked the code and this could be the issue, please go ahead and create a PR for it, otherwise I can do it.

gowrimanoharir:
@pavkout have created the PR, since it was just removing override in codemirror extension config did not add tests, pls let me know if I should.

PR #9364, opened 2025-11-11
removed tab override in codemirror extension

gowrimanoharir:
Closes-#7713

**Background**

Looks like the default code mirror behavior was overridden in the original PR when autocomplete was implemented, to not allow `tab` key  to autocomplete

https://github.com/Kong/insomnia/pull/112

**Changes**

Removing the override so it reverts to default code mirror behavior allowing `tab` to autocomplete suggestions

Tested in local, can upload videos if needed.


pavkout (approved):
I tested it, and everything works as expected. Thanks for your help.

gowrimanoharir:
@pavkout thanks for the approval, the workflow still says Test Expected please let me know I any should be done for that.

PR #9376, opened 2025-11-13
feat: support ca cert config for mcp

ZxBing0066:
## Background
[INS-1649](https://konghq.atlassian.net/browse/INS-1649)

## Changes
Add an MCPCertificatesModal for users to configure the CA certificate for the MCP client and use it in the mcp fetch.

[INS-1649]: https://konghq.atlassian.net/browse/INS-1649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/main/mcp/common.ts
+++ b/packages/insomnia/src/main/mcp/common.ts
@@ -311,7 +311,9 @@ export const getFetchDispatcher = async (requestId: string) => {
 
   return new Agent({
     connect: {
-      ...(workspaceCaCert?.path && !workspaceCaCert?.disabled ? { ca: fs.readFileSync(workspaceCaCert.path) } : {}),
+      ...(workspaceCaCert?.path && !workspaceCaCert?.disabled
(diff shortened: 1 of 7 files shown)


ZxBing0066:
UI: 
image

packages/insomnia/src/main/mcp/common.ts
```diff
@@ -8,6 +8,7 @@ import {
   ListRootsResultSchema,
 } from '@modelcontextprotocol/sdk/types.js';
 import { BrowserWindow } from 'electron';
+import { Agent } from 'undici';
```
cwangsmv, line 11: Need to add `undici` in package.json. It exists now because it is a transitive dependency.

ZxBing0066: Good catch, the global fetch in node js depends on undici, which makes me consider it as a node-internal module. 😌

cwangsmv (approved):

PR #9377, opened 2025-11-13
feat: add mcp error details to the events and the UI

ZxBing0066:
## Background
When fetch throws an error, most of the details are in the cause, but currently, we only show the error.message. This makes it hard to find the root cause of an error.

## Changes

- Show cause in the transport events
- Show cause in the error message


ZxBing0066:
Before:
image

After:
image

cwangsmv (approved):

PR #9397, opened 2025-11-18
feat: refactor mcp code with context to support safe abort at any time

ZxBing0066:
## Background 
[INS-1638](https://konghq.atlassian.net/browse/INS-1638)
There are many steps when creating an mcp connection. When users close the tabs to force close the connections, they need a safe way to abort all connections, whatever which steps that are in.

## Changes
- Use context to store the execution environments instead of global variables
- Support aborting requests even when it's not ready

[INS-1638]: https://konghq.atlassian.net/browse/INS-1638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/insomnia/src/main/mcp/client-requests.ts
```diff
@@ -87,50 +87,54 @@ export const readResource = async (options: CommonMcpOptions & ReadResourceReque
...
-        ...result,
-        // @ts-expect-error - workaround to add the json rpc request id to the logged data
-        id: mcpClient._requestMessageId - 1,
-      },
-      direction: 'INCOMING',
-    };
-    writeEventLogAndNotify(requestId, messageEvent);
-    return result;
+  const context = getReadyActiveMcpConnectionContext(requestId);
+  const mcpClient = getActiveMcpClient(requestId);
```
cwangsmv: Nit: `getActiveMcpClient` has also used `getReadyActiveMcpConnectionContext`, we can use `context?.client` as `mcpClient`.

packages/insomnia/src/main/mcp/client-requests.ts
```diff
@@ -87,50 +87,54 @@ export const readResource = async (options: CommonMcpOptions & ReadResourceReque
...
```
cwangsmv: Same here

cwangsmv (dismissed):

packages/insomnia/src/main/mcp/client-requests.ts
ZxBing0066: Good catch, revised.

cwangsmv (approved):

PR #9398, opened 2025-11-18
Fix file delimiter escape issue in Windows [INS-1611]

### LINKED ISSUE #9292: Base64 URL encode produces wrong output if \ (backslash) at input
### Expected Behavior

LP204\VFRFIN
TFAyMDRcVkZSRklO

### Actual Behavior

LP204\VFRFIN
TFAyMDRcXFZGUkZJTg

Image

### Reproduction Steps

use base64 encoded and as a static value type LP204\VFRFIN

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

Revisiting the UI once again, no other action, shows that \ handling has more issues, \ clones itself very agressively:

LP204\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\VFRFIN

Image

### Insomnia Version

Version: Insomnia 11.6.1 Build date: 9/18/2025 OS: Windows_NT x64 10.0.22631 Electron: 37.2.6 Node: 22.17.1 Node ABI: 136 V8: 13.8.258.30-electron.0 Architecture: x64

Windows

### Operating System Version

Windows_NT x64 10.0.22631

### Installation method

download

_No response_

yaoweiprc:
Close https://github.com/Kong/insomnia/issues/9292

This issue is not related to our base64-URL implementation. It was introduced by a previous PR:
https://github.com/Kong/insomnia/pull/8221

The background of that PR was to address another issue: 
The user is using the JSON view to edit environment variables. When an environment variable’s value uses a File tag, and then that environment variable is referenced in a Request, the file content cannot be read correctly.
The details of the original issue:

https://konghq.atlassian.net/browse/INS-167

https://github.com/Kong/insomnia/issues/5754

In that earlier PR, to fix the issue, the logic for escaping and unescaping in the tag editor was modified. I believe the changes in that PR are not reasonable, as they affect the rendering of all Nunjucks tags that contain string-type parameters. I believe the original implementation was correct.

So in this PR, firstly, I reverted the changes from that PR, and it does fix the [issue](https://github.com/Kong/insomnia/issues/9292)

Secondly, for the issue that PR was trying to address, it should be fixed in the environment editor component. I’ve added the escape and unescape logic for the file tag’s parameters in this component, and now it works correctly.


jackkav (commented):
Since we known this issue is related to orderedJSON an unmaintained library, another possible apporach would be to write a wrapper function which performs the required transforms and replace all usages of orderedJSON with this wrapper function.

yaoweiprc:
> Since we known this issue is related to orderedJSON an unmaintained library, another possible apporach would be to write a wrapper function which performs the required transforms and replace all usages of orderedJSON with this wrapper function.

This issue has nothing to do with orderedJSON. When parsing JSON-formatted environment variables using the JSON.parse or orderedJSON.parse method, double backslashes are converted into a single backslash, which then causes problems during Nunjucks rendering.

For example, if the JSON-formatted environment variable contains two backslashes:

`{"foo": "{% file 'C:\\a.txt' %}"}`


After JSON.parse, it becomes a single backslash:

`{% file 'C:\a.txt' %}`


During rendering, Nunjucks interprets this single backslash as an escape character for 'a', causing an error.

My approach is simply to insert an extra backslash after JSON.parse to ensure that Nunjucks can correctly parse the file path later.

ZxBing0066 (approved):

PR #9408, opened 2025-11-20
fix: clone selected branch - [INS-1462]

CurryYangxx:
## Changes

- pass the `ref` option into `git.clone` function (for both github, gitlab, custom git)
- delete 2 unused git-related modal components
- delete some unused entry


CurryYangxx:
@gatzjames Could you help me take a look if it is safe to delete the 2 modals in this pr? There are some ts errors, but I found there is no way to trigger the modals to open.

yaoweiprc (dismissed):

gatzjames (dismissed):
🚀 LGTM

The modals were used on the previous flow and are no longer needed!

gatzjames (approved):

PR #9425, opened 2025-11-26
Fix: Using safeParser to parse Insomnia git yaml files

cwangsmv:
## Background
Currently we're using `InsomniaFileSchema.parse` to parse git yaml files directly. Whenever it meets malformed yaml files or yaml file that is not supported for current version, it will throw error and the application will crash. 

## Changes
- Using `InsomniaFileSchema.safeParser` to parse yaml files instead. 
- Throw error when `file.type` is not recognized rather than fall back to mock server workspace by default
- Add unit test to cover unsupported yaml file

[INS-1763](https://konghq.atlassian.net/browse/INS-1763)

[INS-1763]: https://konghq.atlassian.net/browse/INS-1763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/common/__tests__/import-v5-parser.test.ts
+++ b/packages/insomnia/src/common/__tests__/import-v5-parser.test.ts
@@ -501,6 +501,10 @@ describe('InsomniaFileSchema (discriminated union)', () => {
     expect(f.type).toBe('mock.insomnia.rest/5.0');
   });
 
+  it('rejects future or unsupported type', () => {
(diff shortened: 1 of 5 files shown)


yaoweiprc (approved):
LGTM

godfrzero:
This could potentially fix #9469 as well.

PR #9472, opened 2025-12-12
chore: Modify the windows app signing process to use the new certificate

balugeorge:
## What does this PR do?
* This will use the new certificate for windows app signing


godfrzero (approved):
Approving to verify signed artifacts during beta release.

PR #9485, opened 2025-12-15
Add test for main-workflow.test.ts

zhangxyweb:
# Transpile js bundle
npm run app-build

# Run tests
npm run test:smoke:build -- main-workflow.test.ts --reporter=html

# Test cases implemented:

- Normal request
- Request with cookies
- Request requiring environment variables
- Request requiring authentication
- Request with empty URL

Note: Authentication failure case and token-based authentication were not implemented due to time constraints.


jackkav (commented):
Welcome! In the spirit of keeping our builds fast we have two kinds of e2e test critical, every platform, smoke every PR run but only on linux platform.
This test overlaps existing coverage, so it might be a good idea to create a new kind of e2e test, perhaps exhaustive or workflows or something similar. That way you can contribute tests without having a full knowledge of existing test coverage and speed of execution.
Alternatively you can get familiar with existing tests and extend them to limit the time added to PR test runs in order to keep the team merging PRs quickly.
Thanks for the contribution.

PR #9495, opened 2025-12-18
fix: Insomnia V5.1 parser fails to parse legacy data with incomplete header data

cwangsmv:
Legacy Insomnia will export files with value property missing in header items.
Since the value property of the header item is missing, the latest parser will parse the request as `GRPCRequest` rather than normal `Request`.
Fix to add default value for name or value when it is missing in each header item.

[INS-1822](https://konghq.atlassian.net/browse/INS-1822)

[INS-1822]: https://konghq.atlassian.net/browse/INS-1822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yaoweiprc (approved):
LGTM

PR #9496, opened 2025-12-18
Remove unnecessary parameter in OAuth flow [INS-1859]

yaoweiprc:
We didn’t implement the OAuth2 flow strictly according to the spec.
According to the specification, the state parameter should only be sent to the Auth Provider during the authorization request. It does not need to be included when exchanging the authorization code for a token. However, we are currently also sending state in the code-to-token exchange. The user’s Auth server performs strict validation, detects the extra, unnecessary parameter, and therefore returns a 400 error.


cwangsmv (approved):
LGTM

PR #9507, opened 2025-12-22
fix: Add state and scope input UI for MCP OAuth parameter

### LINKED ISSUE #9505: MCP: allow extra properties for tool calls
Allow passing extra params to MCP tools besides those that the tool was initialized with.

**Why?**

Because VS Code (unlike other IDEs) caches the MCP tools and does not start MCP servers until they are required by an agent. Because of that, there are situations when MCP tools get updated, but VS Code still uses the old tool params for the first request. In case a param was removed, it still sends it. And I need to test this specific case, I need to handle that param for compatibility purposes.

**How does it currently work?**

I can add an additional param in the "Parameters Overview" pane, but I get an error `must NOT have additional properties` upon clicking the "Call Tool" button.

cwangsmv:
### Changes
- Adding `state` and `scope` parameter for MCP OAuth
- Enhance OAuth logic to support public OAuth client
- Do not block sending MCP requests when parameters are invalid

[INS-1861](https://konghq.atlassian.net/browse/INS-1861)
Closes https://github.com/Kong/insomnia/issues/9505

[INS-1861]: https://konghq.atlassian.net/browse/INS-1861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/main/mcp/oauth-client-provider.ts
+++ b/packages/insomnia/src/main/mcp/oauth-client-provider.ts
@@ -92,11 +92,15 @@ export class McpOAuthClientProvider implements OAuthClientProvider {
     }
 
     if ('clientId' in this.mcpRequest.authentication && this.mcpRequest.authentication.clientId) {
+      const { clientId, clientSecret, clientIdIssuedAt, clientSecretExpiresAt } = this.mcpRequest.authentication;
+      // https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/auth.ts#L223
+      // Set client_secret to undefined if it's not set or empty string
+      const parsedClientSecret = clientSecret && clientSecret.trim().length > 0 ? clientSecret : undefined;
       return {
-        client_id: this.mcpRequest.authentication.clientId,
-        client_secret: this.mcpRequest.authentication.clientSecret,
-        client_id_issued_at: this.mcpRequest.authentication.clientIdIssuedAt,
-        client_secret_expires_at: this.mcpRequest.authentication.clientSecretExpiresAt,
+        client_id: clientId,
+        client_secret: parsedClientSecret,
+        client_id_issued_at: clientIdIssuedAt,
+        client_secret_expires_at: clientSecretExpiresAt,
       };
     }
     return;
--- a/packages/insomnia/src/main/mcp/oauth-client-provider.ts
+++ b/packages/insomnia/src/main/mcp/oauth-client-provider.ts
@@ -93,7 +93,8 @@ export class McpOAuthClientProvider implements OAuthClientProvider {
 
     if ('clientId' in this.mcpRequest.authentication && this.mcpRequest.authentication.clientId) {
       const { clientId, clientSecret, clientIdIssuedAt, clientSecretExpiresAt } = this.mcpRequest.authentication;
-      // https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/auth.ts#L223
+
+      // https://github.com/modelcontextprotocol/typescript-sdk/blob/6b4d99f10b975d65392bb777cc8cb1151c20c972/packages/client/src/client/auth.ts#L223%20
       // Set client_secret to undefined if it's not set or empty string
       const parsedClientSecret = clientSecret && clientSecret.trim().length > 0 ? clientSecret : undefined;
       return {
--- a/packages/insomnia/src/main/mcp/oauth-client-provider.ts
+++ b/packages/insomnia/src/main/mcp/oauth-client-provider.ts
@@ -148,6 +148,13 @@ export class McpOAuthClientProvider implements OAuthClientProvider {
     });
     await this.refreshMcpRequest();
   }
+  // add state parameter to authorization url if present in authentication
+  async state() {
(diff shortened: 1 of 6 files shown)


packages/insomnia/src/ui/components/websockets/realtime-response-pane.tsx
```diff
@@ -314,7 +314,7 @@ const RealtimeActiveResponsePane: FC<RealtimeActiveResponsePaneProps & { readySt
         <TabPanel className="flex w-full flex-1 flex-col overflow-hidden" id="events">
           <PanelGroup direction="vertical" className="grid h-full w-full grid-rows-[repeat(auto-fit,minmax(0,1fr))]">
             {response.error && !isMCPAuthError ? (
-              <ResponseErrorViewer url={response.url} error={response.error} />
+              <ResponseErrorViewer url={response.url} error={response.error} isMcpResponse />
```
ZxBing0066: Why is it hard-coded to `isMcpResponse`? 🤔

packages/insomnia/src/ui/components/editors/auth/o-auth-2-auth.tsx
```diff
@@ -291,7 +291,7 @@ const getFieldsForGrantType = (authentication: Extract<RequestAuthentication, {
 
     advanced = [responseType, scope, state, tokenPrefix, audience];
   } else if (grantType === GRANT_TYPE_MCP_AUTH_FLOW) {
-    basic = [clientId, clientSecret, readonlyRedirectUri];
+    basic = [clientId, clientSecret, readonlyRedirectUri, state, scope];
```
ZxBing0066: Should we put state and scope into advanced for consistency?

packages/insomnia/src/main/mcp/transport-streamable-http.ts
```diff
@@ -160,7 +160,9 @@ const wrappedFetch = async (
       }
     }
 
-    const resourceMetadataUrl = extractResourceMetadataUrl(response);
+    const { resourceMetadataUrl, scope: scopeFromWWWAuthenticate } = extractWWWAuthenticateParams(response);
+    // Use scope from authenticate config if available, otherwise use scope from WWW-Authenticate header
+    const scope = 'scope' in authentication && authentication.scope ? authentication.scope : scopeFromWWWAuthenticate;
```
ZxBing0066: Nit: It may be better to trim(authentication.scope) before checking. Otherwise a space may blocks the scope from the header, which is hard to realize.

packages/insomnia/src/main/mcp/oauth-client-provider.ts
```diff
@@ -92,11 +92,15 @@ export class McpOAuthClientProvider implements OAuthClientProvider {
     }
 
     if ('clientId' in this.mcpRequest.authentication && this.mcpRequest.authentication.clientId) {
+      const { clientId, clientSecret, clientIdIssuedAt, clientSecretExpiresAt } = this.mcpRequest.authentication;
+      // https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/auth.ts#L223
+      // Set client_secret to undefined if it's not set or empty string
+      const parsedClientSecret = clientSecret && clientSecret.trim().length > 0 ? clientSecret : undefined;
```
ZxBing0066, line 99: So, this case is that the user cleans the input, and an empty string is sent to the backend, which causes issues?

packages/insomnia/src/main/mcp/oauth-client-provider.ts
```diff
@@ -92,11 +92,15 @@ export class McpOAuthClientProvider implements OAuthClientProvider {
     }
 
     if ('clientId' in this.mcpRequest.authentication && this.mcpRequest.authentication.clientId) {
+      const { clientId, clientSecret, clientIdIssuedAt, clientSecretExpiresAt } = this.mcpRequest.authentication;
+      // https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/auth.ts#L223
```
ZxBing0066: Let's use a permalink instead of this unstable one.

ZxBing0066 (dismissed):

packages/insomnia/src/ui/components/websockets/realtime-response-pane.tsx
cwangsmv: Refine the logic:
- For none MCP Auth related error, using `isMcpResponse` to decide whether to show settings or MCP certificate modal to `Disable SSL Validation`.
- For MCP Auth error, since users can view error details when clicking the event. Adding a new `showErrorDetails` to hide error details under this situation to avoid error details to cover events table by default

packages/insomnia/src/main/mcp/oauth-client-provider.ts
cwangsmv: Fixed

packages/insomnia/src/ui/components/editors/auth/o-auth-2-auth.tsx
cwangsmv: Good catch, moving `state` and `scope` to advanced pane

ZxBing0066 (approved):

PR #9511, opened 2025-12-23
refactor: IoC database implementation to get rid of runtime dependencies

ZxBing0066:
[INS-1819](https://konghq.atlassian.net/browse/INS-1819)
## Background
Current common/database depends on Node.js, Electron, and NeDB.

## Changes
IoC database implementation to make the common/database get rid of runtime dependencies.

- Move the main DB implementation to database-nedb
- Move the DB implementation to database.client and database.main

[INS-1819]: https://konghq.atlassian.net/browse/INS-1819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jackkav:
This PR looks like a good step forward.

packages/insomnia/src/common/database/database-nedb.ts
```diff
@@ -1,49 +1,39 @@
...
 
-/* eslint-disable prefer-rest-params -- don't want to change ...arguments usage for these sensitive functions without more testing */
+import os from 'node:os';
 import fsPath from 'node:path';
 
 import NeDB from '@seald-io/nedb';
-import electron from 'electron';
-import { v4 as uuidv4 } from 'uuid';
 
+import { database } from '~/common/database';
```
ZxBing0066, line 9: Hi @jackkav, I've removed the `database/index` reference from `database.main`, but for this file, since it's not the final implementation because for the main process, some of the methods need to be overridden in the `database.main`. So to ensure the function references between each other, like when calling `database.flushChanges` in `bufferChanges`, we need to ensure it's calling the function from the injected instance.
The current implementation seems inelegant. Do you think using `class MainDatabase extends NedbDatabase` would be better? Looking forward to your comments.

jackkav (dismissed):

packages/insomnia/src/common/database/database-nedb.ts
jackkav: I don't think classes offer us enough benefit for their cost in this case. Especially since this is not the final step we don't want to use abstractions that obscure the data flow.

packages/insomnia/src/main/ipc/database.ts
```diff
@@ -0,0 +1,3 @@
+export interface DatabaseBridgeAPI {
```
CurryYangxx, line 1: Why do we put this interface under `network` folder?

ZxBing0066: Good catch, I've moved it to the `main/ipc` folder

CurryYangxx (approved):

PR #9514, opened 2025-12-24
[INS-1725] Add Git Providers registry and unify Git Credentials

gatzjames:
# Overview

- [x] Add new Git Provider interface that allows us to group common logic for connecting and managing state and credentials for different git providers (e.g. github, gitlab)
- [x] Added a new Git Provider Registry to allow us to register providers during runtime
- [x] Migrated Git Credentials to a new unified format. All credentials are now referenced by `id` instead of having to handle custom logic for each individual case. Storing credentials in Git Repositories is now `@deprecated` and only kept to allow the migration to run due to the way our database works.
- [x] Since Git Providers have features that they support we no longer need specific logic for fetching repos, completing OAuth flows etc for each provider. These have now been standardized so instead of `completeGitHubSignIn` and `completeGitLabSignIn` we now have a single function to complete an OAuth flow for a provider that supports it.


packages/insomnia/src/sync/git/migrations.ts
```diff
@@ -0,0 +1,156 @@
...
+import { database } from '~/common/database';
+import ElectronStorage from '~/main/electron-storage';
+import { type GitRepository, isGitCredentialsOAuth } from '~/models/git-repository';
+
+import * as models from '../../models';
+import type { GitCredentials } from '../../models/git-credentials';
+
+const MIGRATION_KEY = 'GIT_CREDENTIALS_MIGRATION';
+
+const migrationStorage = new ElectronStorage('git-migrations');
```
yaoweiprc, line 36: Need to specify the absolute path here. Can reuse localStorage

yaoweiprc, line 36: Specify absolute path and reuse localStorage here.

packages/insomnia/src/sync/git/migrations.ts
```diff
@@ -0,0 +1,156 @@
...
+const hasRunMigration = () => migrationStorage.getItem(MIGRATION_KEY);
+const markMigrationComplete = () => migrationStorage.setItem(MIGRATION_KEY, 1);
+
+async function migrateGitHubConnectedRepositories(repositories: GitRepository[]) {
+  const githubCredentials = await models.gitCredentials.getByProvider('githubapp');
+
+  if (githubCredentials) {
+    await models.gitCredentials.update(githubCredentials, {
+      provider: 'github',
```
yaoweiprc, line 45: Add default name and renewalAttempts since they are required in the model.

packages/insomnia/src/sync/git/providers/index.ts
```diff
@@ -0,0 +1,115 @@
...
+ * Initialize built-in Git remote providers
+ *
+ * This function registers all built-in providers (GitHub, GitLab, Custom, Local)
+ * with the singleton registry. Call this once during app initialization.
+ */
+export function initializeGitRemoteProviders(): void {
+  console.log('[Git Providers] Initializing Git remote providers...');
+
+  // Register GitHub provider
+  gitRemoteProviderRegistry.register(
```
CurryYangxx, line 74: Can we register multiple provider at once?
```
const githubProvider = new GithubProvider();
const gitlabProvider = new GitlabProvider();

gitRemoteProviderRegistry.register([ githubProvider, gitlabProvider ])
```

packages/insomnia/src/sync/git/providers/types.ts
```diff
@@ -0,0 +1,185 @@
...
+export type GitRemoteProviderType = 'github' | 'gitlab' | 'custom';
+
+/**
+ * Base configuration for all providers
+ */
+export interface BaseProviderConfig {
+  type: GitRemoteProviderType;
+  displayName: string;
+  description?: string;
+  iconName?: IconProp; // For UI rendering
```
CurryYangxx, line 18: Do we need `supportsOAuth: boolean` here?

packages/insomnia/src/sync/git/utils.ts
```diff
@@ -1,58 +1,42 @@
...
   console.log(`[git-event] ${message}`);
 };
 
-const onAuthFailure = (credentials?: GitCredentials): AuthFailureCallback => {
-  let retryCount = 0;
-  const maxRetries = 2;
-
-  return async (message, auth) => {
+const onAuthFailure = (credentialsId?: string | null): AuthFailureCallback => {
+  return async message => {
```
CurryYangxx, line 17: this param should be `url` to be consistent with isomorphic-git type definition

packages/insomnia/src/main/git-service.ts
```diff
@@ -784,53 +753,51 @@ export const cloneGitRepoAction = async ({
...
 
       const inMemoryFsClient = MemClient.createClient();
 
-      const providerName = getOauth2FormatName(repoSettingsPatch.credentials);
+      let providerName = 'custom';
+      if (repoSettingsPatch.credentialsId) {
+        const credentials = await models.gitCredentials.getById(repoSettingsPatch.credentialsId);
+        invariant(credentials, 'Git Credentials not found');
+        providerName = credentials.provider;
+      }
```
CurryYangxx, line 800: Looks like duplicate logic.
We use this logic many places in this file, consider to extract to a separate function.

packages/insomnia/src/models/git-credentials.ts
```diff
@@ -48,11 +145,14 @@ export function create(patch: Partial<GitCredentials> = {}) {
...
 export async function getById(id: string) {
-  return db.findOne<GitCredentials>(type, { _id: id });
+  const doc = await db.findOne<GitCredentials>(type, { _id: id });
+  return doc ? migrate(doc) : null;
 }
 
-export async function getByProvider(provider: OauthProviderName) {
-  return db.findOne<GitCredentials>(type, provider === 'github' ? { provider: 'githubapp' } : { provider: 'gitlab' });
+export async function getByProvider(provider: GitRemoteProviderType | 'githubapp') {
+  const doc = await db.findOne<GitCredentials>(type, { provider });
```
CurryYangxx, line 153: Every provider can have multiple credentials, is it ok to use `findOne` here ?

packages/insomnia/src/sync/git/migrations.ts
CurryYangxx: Also for gitlab and custom provider below

packages/insomnia/src/sync/git/migrations.ts
```diff
@@ -0,0 +1,156 @@
...
+    });
+
+    for (const repo of repositories) {
+      await models.gitRepository.update(repo, {
+        credentialsId: githubCredentials._id,
+        credentials: undefined,
+        author: undefined,
```
CurryYangxx, line 52: We set the `credentials` and `authot` to undefined here. So do we need to change the ts definition in the gitRepository model?

CurryYangxx:
Close this one because this change is included in #9513

PR #9520, opened 2025-12-26
feat: credential setup when project creation

CurryYangxx:
Add git credential setup flow in the project creation Modal
image
image


yaoweiprc (approved):
👍

PR #9526, opened 2025-12-29
Support editing custom git credentials in the preference modal

yaoweiprc:
1. Add `optionalKeys` field to the git-credentials model so that extral keys can be preserved when initializing the model.
2. Support editing custom git credentials in the preference modal.

### DIFF
--- a/packages/insomnia/src/models/git-credentials.ts
+++ b/packages/insomnia/src/models/git-credentials.ts
@@ -8,8 +8,24 @@ export type GitRemoteProviderType = 'github' | 'gitlab' | 'custom';
 
 export type GitCredentials = BaseModel & BaseGitCredentials;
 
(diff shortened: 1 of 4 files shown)


packages/insomnia/src/models/git-credentials.ts
```diff
@@ -8,8 +8,24 @@ export type GitRemoteProviderType = 'github' | 'gitlab' | 'custom';
...
+  'token',
+  'refreshToken',
+  'expiresAt',
+  'scopes',
+  'emails',
+  'selectedEmail',
+  'username',
+  'password',
+];
+
```
CurryYangxx, line 21: this array should only include the field which is not exist in the init function

CurryYangxx (approved):

PR #9530, opened 2025-12-29
fix: Import collection variables as the base environment collection environment.

cwangsmv:
### Changes:
- Import collection variable as collection base environment in Insomnia
- When importing within workspace, add a new option to override or discard existing base environment with name conflict.

[INS-873](https://konghq.atlassian.net/browse/INS-873)

[INS-873]: https://konghq.atlassian.net/browse/INS-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/ui/components/modals/import-modal/import-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/import-modal/import-modal.tsx
@@ -1,7 +1,7 @@
 import classNames from 'classnames';
 import React, { type FC, Fragment, type ReactNode, useEffect, useId, useMemo, useRef, useState } from 'react';
 import { type DirectoryDropItem, type FileDropItem, OverlayContainer, useDrop } from 'react-aria';
-import { Heading, Switch } from 'react-aria-components';
+import { Heading } from 'react-aria-components';
 
 import { useImportResourcesFetcher } from '~/routes/import.resources';
 import { useScanResourcesFetcher } from '~/routes/import.scan';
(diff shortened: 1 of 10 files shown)


yaoweiprc:
Notes:
If we can delete this piece of code?
```
    // Each postman's collection has its variable, we map it to request group's environment in Insomnia
    // I think it's better to check if the resource's type is 'request_group' rather than to check it by index 0, but let's just leave it as it is
    if (resources.length > 0 && resources[0].variable) {
      resources[0].environment = resources[0].variable;
    }
```

packages/insomnia/src/ui/components/modals/import-modal/import-modal.tsx
```diff
@@ -376,19 +387,42 @@ const ImportResourcesForm = ({
...
+  const [overrideBaseEnvironmentData, setOverrideBaseEnvironmentData] = useState(true);
   return (
     <Fragment>
       <div className="flex max-h-[50vh] flex-col gap-(--padding-md) overflow-auto">
         <div className="overflow-y-auto">
           <ScanResultsTable scanResults={scanResults} />
+          {isImportingBaseEnvironmentToWorkspace && (
+            <div className="">
+              <label className="flex items-center gap-2">
+                <input
```
ZxBing0066: Maybe better to use the checkbox component from the ui components folder.

ZxBing0066 (dismissed):

yaoweiprc (dismissed):
👍

ZxBing0066 (approved):

PR #9532, opened 2025-12-30
fix: update project credential

CurryYangxx:
**Changes**

> if user select gitlab credential and input a custom git server repo url, will try renew token after authorization fails, this will lead to an infinite loop in isomorphic-git onAuthFailure callback.

- validate repo url when fetch remote branches
- support update credential and re-connect to remote if the project's credential is deleted


yaoweiprc (approved):

yaoweiprc (approved):
LGTM

PR #9549, opened 2026-01-08
fix: git credential migration state storage path

CurryYangxx:
Init electronStorage after we set the `userdata` path.

### DIFF
--- a/packages/insomnia/src/sync/git/migrations.ts
+++ b/packages/insomnia/src/sync/git/migrations.ts
@@ -25,6 +25,7 @@
  */
 
 import { database } from '~/common/database';
+import type ElectronStorage from '~/main/electron-storage';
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/sync/git/migrations.ts
```diff
@@ -33,9 +34,23 @@ import { type GitCredentials, isGitCredentialsV1 } from '../../models/git-creden
 
 const MIGRATION_KEY = 'GIT_CREDENTIALS_MIGRATION';
 
-const migrationStorage = initElectronStorage();
```
CurryYangxx, line 36: We initialize storage directly when this file is loaded. So the path is always point to `Insomnia` path

yaoweiprc (approved):

PR #9553, opened 2026-01-09
fix: new git credential flow

CurryYangxx:
Fix for the new git credentials flow #9513 
- Hide git repo&branch select UI when converting project type if the feature flag is closed
- Modify the clearing git repo logic when logging out to adapt to the new credential model


yaoweiprc (approved):
👍

PR #9567, opened 2026-01-14
test: verify git sync feature flag & storage rule - [INS-1908]

CurryYangxx:
**Add test for the cases:**

- Disable git sync when the feature flag is closed
- Disable git sync when git storage rule is closed


yaoweiprc (approved):
👍

PR #9568, opened 2026-01-14
Fix: Environment variable in MCP OAuth settings in not working

cwangsmv:
## Background
When using environment variables in MCP OAuth settings, the setting values are not rendered

## Changes
- Using the rendered authentication value in OAuth client provider rather than read the value from database
- Fix a minor issue that token scope will override user's scope setting

[INS-1942](https://konghq.atlassian.net/browse/INS-1942)

[INS-1942]: https://konghq.atlassian.net/browse/INS-1942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/main/mcp/oauth-client-provider.ts
+++ b/packages/insomnia/src/main/mcp/oauth-client-provider.ts
@@ -63,12 +63,15 @@ export class McpOAuthClientProvider implements OAuthClientProvider {
     const newAuthentication = {
       ...mcpRequest.authentication,
       ...auth,
-    } as RequestAuthentication;
+    };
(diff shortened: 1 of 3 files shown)


ZxBing0066 (dismissed):
👍
Nit: since we use the auth from the options, we may directly get them from the context instead of passing through the arguments.

cwangsmv:
> 👍 Nit: since we use the auth from the options, we may directly get them from the context instead of passing through the arguments.

Fixed, using the `context` instaed

ZxBing0066 (approved):

PR #9574, opened 2026-01-15
refactor: migrate projects request to insomnia-api package - [INS-1874]

CurryYangxx:
Migrate the following apis to the `insomnia-api` package:

- /v1/organizations/:organizationId/team-projects
- /v1/organizations/:organizationId/team-projects/:projectId
- /v1/organizations/:organizationId/git-projects
***
Since we set `onlyResolveOnSuccess` to always true when we register `fetch` api. So I changed all the error handling approaches for the places that send these requests.


packages/insomnia/src/ui/insomnia-fetch.ts
```diff
@@ -11,8 +11,28 @@ export class ResponseFailError extends Error {
     this.response = response;
   }
   response;
+  payload?: any;
 }
 
+export const parseInsomniaFetchError = (error: unknown) => {
```
ZxBing0066: I would suggest moving this check to the insomnia-api, as it is part of the api spec. Like the following:

```ts
import {createProject, isAPIError} from 'insomnia-api'

try {
  const newProject = await createProject({ ... })
} catch (e) {
  if(isAPIError(e)) {
    if (parsedError.name === 'FORBIDDEN') {
      errMessage = 'You do not have permission to create a cloud project in this organization.';
    } else if (parsedError.name === 'NEEDS_TO_UPGRADE') {
      errMessage = 'Upgrade your account in order to create new Cloud Projects.';
    } else if (parsedError.name === 'PROJECT_STORAGE_RESTRICTION') {
      errMessage = parsedError.message ?? 'The owner of the organization allows only Local Vault project creation.';
    }

    throw new Error(errMessage);
  }
  throw e
}
```

packages/insomnia/src/ui/insomnia-fetch.ts
```diff
@@ -11,8 +11,28 @@ export class ResponseFailError extends Error {
     this.response = response;
   }
   response;
+  payload?: any;
```
ZxBing0066: Is it necessary to add this payload? I didn't find anywhere use this.

packages/insomnia/src/models/helpers/project.ts
```diff
@@ -25,22 +26,10 @@ export async function updateLocalProjectToRemote({
...
-        error: string;
-        message?: string;
-      }
-  >({
-    path: `/v1/organizations/${organizationId}/team-projects`,
-    method: 'POST',
-    data: {
-      name: project.name,
-    },
+  const newCloudProject = await createTeamProject({
```
ZxBing0066: I think we should also update the error handling here.

packages/insomnia/src/ui/insomnia-fetch.ts
CurryYangxx: deleted it

packages/insomnia/src/models/helpers/project.ts
CurryYangxx: fixed

packages/insomnia/src/ui/insomnia-fetch.ts
CurryYangxx: I added an `isApiError` function into the `insomnia-api` package and replace the old error handling logic

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.update.tsx
```diff
@@ -43,42 +43,40 @@ export async function clientAction({ request, params }: Route.ClientActionArgs)
...
+          });
 
-        return {
-          error,
-        };
+          return {
+            error: errorMessage,
+          };
+        }
+        throw error;
```
ZxBing0066, line 79: To keep consistent, we should use `return` instead of throw here?

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.update.tsx
```diff
@@ -95,45 +93,43 @@ export async function clientAction({ request, params }: Route.ClientActionArgs)
...
```
ZxBing0066, line 132: Also here.

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.update.tsx
```diff
@@ -149,121 +145,106 @@ export async function clientAction({ request, params }: Route.ClientActionArgs)
...
-            title: 'Error updating project',
-            description: error,
-            icon: 'warning',
-            status: 'error',
-          });
-
-          return {
-            error,
-          };
+          throw error;
```
ZxBing0066, line 247: Same here.

packages/insomnia/src/routes/organization.$organizationId.project.new.tsx
```diff
@@ -142,8 +109,25 @@ export const createProject = async (organizationId: string, newProjectData: Crea
...
+        errMessage = 'You do not have permission to create a cloud project in this organization.';
+      } else if (error.name === 'NEEDS_TO_UPGRADE') {
+        errMessage = 'Upgrade your account in order to create new Cloud Projects.';
+      } else if (error.name === 'PROJECT_STORAGE_RESTRICTION') {
+        errMessage = error.message ?? 'The owner of the organization allows only Local Vault project creation.';
+      }
+      throw new Error(errMessage);
+    }
 
+    throw error;
```
ZxBing0066, line 126: And here

ZxBing0066 (approved):

PR #9575, opened 2026-01-15
Spike: Separate main and renderer global type definitions 

cwangsmv:
# Problem
The global type definitions (specifically Window interface) were being applied to all code including the main process. These types are only relevant for the renderer process in Electron, not the main process where window doesn't exist.

# Solution
- Adding `renderer-global.d.ts` file for window interface used in renderer process. 
- Adding `tsconfig.base.json` for common shared TS configs.
- Adding `tsconfig.main.json` for main code TS configs and `tsconfig.renderer.json` for renderer code TS configs.
- The `tsconfig.json` will take the  `tsconfig.main.json`  and `tsconfig.renderer.json` for reference.


jackkav:
This is a great idea, I'm curious if there are any other ways to achieve separation besides using 
```
  "exclude": ["../../types/renderer-global.d.ts"]
```

jackkav (commented):
❤️

cwangsmv:
Type-checking requires building the full dependency graph of the project, splitting into two TS configs will cause the issue that codes in the `main process` still imports code from the `modal`, so those files are type-checked under tsconfig.main and inherit its declare global environment. This leads to a large number of type errors in the modal code which are using `window.main` method. We’ll need to finish decoupling the modules before this separation can work properly.

PR #9590, opened 2026-01-21
refactor: migrate invite api - [INS-1875]

CurryYangxx:
Migrate apis to `insomnia-api` package:

- /v1/organizations/:organizationId/invites/:invitationId
- /v1/organizations/:organizationId/invites/:invitationId/reinvite

### DIFF
--- a/packages/insomnia-api/src/invite.ts
+++ b/packages/insomnia-api/src/invite.ts
@@ -1,0 +1,52 @@
+import { fetch } from './fetch';
+
+export const reinvite = ({
+  organizationId,
+  invitationId,
+  sessionId,
+}: {
+  organizationId: string;
+  invitationId: string;
+  sessionId: string;
+}) => {
+  return fetch<{ enabled: boolean }>({
+    method: 'POST',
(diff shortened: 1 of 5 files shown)


packages/insomnia-api/src/invite.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const updateInvitationRole = ({
+  organizationId,
+  invitationId,
+  roleId,
+  sessionId,
+}: {
+  organizationId: string;
+  invitationId: string;
+  roleId: string;
```
ZxBing0066, line 27: Nit: since the backend considers this as a string, do you think it's better to keep this for this function?

ZxBing0066 (dismissed):

packages/insomnia-api/src/invite.ts
CurryYangxx: The backend receives an array and uses its first item. So I think we could keep it for now, just pass a string into `updateInvitationRole` and convert to array when calling `fetch`.

ZxBing0066 (approved):

PR #9593, opened 2026-01-21
feat(git): select commit email from git provider per repo [INS-1474]

ryan-willis:
- Adds the ability to override the default author email address used in commits provided by GitHub/GitLab on the project setup/settings modal
  - It does not change the default email address on the _credential_ itself, as users may want/need to use different email addresses for commits on different projects/repos
  - Doesn't apply to custom credentials
- Updates styles on the credential list in preferences to clearly distinguish credential rows
- Only fetches from the provider when changing the author email

### DIFF
--- a/packages/insomnia/src/sync/git/providers/gitlab.ts
+++ b/packages/insomnia/src/sync/git/providers/gitlab.ts
@@ -134,8 +134,6 @@ export class GitLabProvider implements GitRemoteProvider<GitLabProviderConfig> {
   readonly supportsFetchEmails = true;
   readonly supportsAutoRenew = true; // GitLab supports refresh tokens
 
-  private repositoryCache = new Map<string, ProviderRepository[]>();
-
   constructor(config: GitLabProviderConfig) {
     this.config = config;
   }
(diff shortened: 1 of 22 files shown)


packages/insomnia/src/ui/components/project/git-repo-form.tsx
```diff
@@ -210,12 +274,12 @@ export const GitRepoForm: FC<Props> = ({
                       uri,
                     }))
                   }
-                  credentialsId={selectedCredentialsId}
+                  credentialsId={projectData.credentialsId || selectedCredentialsId}
```
yaoweiprc: `selectedCredentialsId` already contains `projectData.credentialsId`

packages/insomnia/src/sync/git/providers/gitlab.ts
```diff
@@ -222,44 +231,9 @@ export class GitLabProvider implements GitRemoteProvider<GitLabProviderConfig> {
...
-
-    const userData = await userResponse.json();
-
-    // Fetch all emails for the user
-    const emailsResponse = await net.fetch(`${this.config.apiUrl}/user/emails`, {
-      headers: {
-        Authorization: `Bearer ${credential.credentials?.token}`,
-      },
-    });
+    const userData = await this.fetchUserWithToken(credential.credentials?.token);
```
yaoweiprc, line 226: The token for Gitlab may have expired when this function is called.
Github does not have same issue because Github token does not expire.

yaoweiprc (changes requested):

packages/insomnia/src/ui/components/project/git-repo-form.tsx
ryan-willis: good call, fixed

yaoweiprc (approved):
👍

PR #9601, opened 2026-01-26
feat: navigate to imported workspace

jackkav:
motivation: when importing a single workspace, we don't currently navigate to it.
Goal: in order to save a click for most use cases when we receive an import with a single workspace we will redirect to it.

Tested with a single curl request import, and a large collection.

```
insomniadev://app/import?curl=curl 'http://en.wikipedia.org/' \
    -H 'Accept-Encoding: gzip, deflate, sdch' \
    -H 'Accept-Language: en-US,en;q=0.8' \
    -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36' \
    -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8' \
    -H 'Referer: http://www.wikipedia.org/' \
    -H 'Connection: keep-alive' --compressed
```

Notes
- Needed to update most of the tests to skip the navigate part after import.
- Renamed a test collection for clarity of debugging
- updated testing docs

### DIFF
--- a/packages/insomnia-smoke-test/README.md
+++ b/packages/insomnia-smoke-test/README.md
@@ -17,6 +17,7 @@ This project contains the smoke testing suite for Insomnia App.
     - [Getting traces from CI](#getting-traces-from-ci)
     - [Build and package methods](#build-and-package-methods)
     - [Non recurring tests](#non-recurring-tests)
+    - [Refresh certs](#refresh-certs)
 
 ## Quick-start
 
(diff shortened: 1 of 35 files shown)


ZxBing0066 (approved):
🚀

What will happen if we import by deep link when in a non-project organization?

PR #9608, opened 2026-01-27
fix: auto-close after commit+push with multiple commits [INS-1948]

ryan-willis:
- refactored the `useEffect` logic to embed success/error checks 
- displays toast post-close indicating a successful commit+push


yaoweiprc (approved):
👍

PR #9611, opened 2026-01-28
optimize: make debug view loader more robust

ZxBing0066:
## Background

[INS-525](https://konghq.atlassian.net/browse/INS-525)

Currently, it's very easy to enter the crash page for data check.

## Changes

- Redirect to the parent page when the resource is not found and show a toast instead of showing the crash page for known checks.
- Remove unnecessary checks.
- Add a max width to the toast to avoid it showing long messages with incorrect style.

[INS-525]: https://konghq.atlassian.net/browse/INS-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.request.$requestId.tsx
+++ b/packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.request.$requestId.tsx
@@ -94,7 +94,7 @@ export async function clientLoader({ params }: Route.ClientLoaderArgs) {
   if (!activeRequest) {
     showResourceNotFoundToast(`Request not found: ${requestId}`);
     if (activeWorkspace.scope === 'mcp') {
-      // Redirect to the project page if it is MCP workspace, as MCP workspace only and must have one request.
+      // Redirect to the project page if it is an MCP workspace, as an MCP workspace must have one request.
       throw redirect(
         href('/organization/:organizationId/project/:projectId', {
           organizationId,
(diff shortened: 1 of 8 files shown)


CurryYangxx:
Do we need to handle other route paths like:

- /environment
- /debug
- /debug/request-group/$id

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.request.$requestId.tsx
```diff
@@ -84,12 +84,33 @@ const getResponseModelName = (request: Request | WebSocketRequest | SocketIORequ
...
+  }
+
   const activeRequest = await requestOperations.getById(requestId);
   if (!activeRequest) {
-    throw redirect(`/organization/${organizationId}/project/${projectId}/workspace/${workspaceId}/debug`);
+    showResourceNotFoundToast(`Request not found: ${requestId}`);
+    if (activeWorkspace.scope === 'mcp') {
+      // Redirect to the project page if it is an MCP workspace, as an MCP workspace must have one request.
+      throw redirect(
+        href('/organization/:organizationId/project/:projectId', {
```
CurryYangxx, line 99: Do you think adding a common function `fallbackToDashboard` to replace this repeated redirect logic would be better?

ZxBing0066:
> Do we need to handle other route paths like:
> 
> * /environment
> * /debug
> * /debug/request-group/$id

Good catch, done.

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.request.$requestId.tsx
ZxBing0066: I don't want to add a function that only wraps one line of code and strongly depends on React Router. I have thought to extract something like 
```ts
const safeLoadResource = ({ organizationId, projectId, workspaceId }) => {
  const project = xxx;
  if(!project) {
    showToast();
    throw redirect('project')
  }
  const workspace = xxx;
  if(!workspace) {
    showToast();
    throw redirect('project/projectid')
  }
  // ...
}
```

But each handler has slight differences, so I haven't got a good idea yet. Maybe we can consider this after we replace all the data load logic in a service in the future.

CurryYangxx (approved):

PR #9626, opened 2026-02-02
fix: show static author email for all git auth types

ryan-willis:
Screenshot 2026-02-03 at 09 33 24


ZxBing0066:
The style is inconsistent with custom credential and OAuth credential with only one eamil, should we unify them?

image

image

ZxBing0066 (dismissed):

ryan-willis:
> @ZxBing0066 should we unify them?

That's how I had it at first, I'll change it back

ZxBing0066 (approved):

PR #9630, opened 2026-02-03
fix: filter legacy credential

CurryYangxx:
> In some edge cases (migration failed or user downgrade then upgrade again), there will be some legacy format git credentials.

Filter legacy Git credentials in credentials list


packages/insomnia/src/ui/components/project/git-repo-form.tsx
CurryYangxx, line 158: @ZxBing0066 Do you think we need to change the ts assertion here? 
Like `const provider = providers.find(p => p.type === selectedItem.provider)!`.

PR #9632, opened 2026-02-03
fix: oas should import requests where possible

jackkav:
This appears to have worked previously. Just had to fix the api spec detection

image

### DIFF
--- a/packages/insomnia/src/common/import.ts
+++ b/packages/insomnia/src/common/import.ts
@@ -173,8 +173,12 @@ export async function scanResources(importEntries: ImportEntry[]): Promise<ScanR
       let v5Error = null;
 
       try {
-        const { data: insomnia5Import, error } = tryImportV5Data(contentStr);
-        v5Error = error;
+        let insomnia5Import: ExportedModel[] = [];
+        if (contentStr.startsWith('type: ')) {
+          const { data, error } = tryImportV5Data(contentStr);
+          insomnia5Import = data as ExportedModel[];
+          v5Error = error;
+        }
         if (insomnia5Import.length > 0) {
           result = {
             type: {
@@ -183,7 +187,6 @@ export async function scanResources(importEntries: ImportEntry[]): Promise<ScanR
               description: 'Insomnia v5',
             },
             data: {
-              // @ts-expect-error -- TSCONVERSION
               resources: insomnia5Import,
             },
           };
@@ -234,6 +237,7 @@ export async function scanResources(importEntries: ImportEntry[]): Promise<ScanR
       });
 
       const requests = resources.filter(isRequest);
+      const requestGroups = resources.filter(isRequestGroup);
       const websocketRequests = resources.filter(isWebSocketRequest);
       const grpcRequests = resources.filter(isGrpcRequest);
       const socketIoRequests = resources.filter(isSocketIORequest);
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/common/import.ts
```diff
@@ -173,8 +173,12 @@ export async function scanResources(importEntries: ImportEntry[]): Promise<ScanR
...
 
       try {
-        const { data: insomnia5Import, error } = tryImportV5Data(contentStr);
-        v5Error = error;
+        let insomnia5Import: ExportedModel[] = [];
+        if (contentStr.startsWith('type: ')) {
+          const { data, error } = tryImportV5Data(contentStr);
+          insomnia5Import = data as ExportedModel[];
+          v5Error = error;
+        }
```
jackkav, line 181: I did this because the error spam was killing my debugging experience

packages/insomnia/src/common/import.ts
```diff
@@ -251,6 +255,7 @@ export async function scanResources(importEntries: ImportEntry[]): Promise<ScanR
         unitTests,
         unitTestSuites,
         requests: [...requests, ...websocketRequests, ...grpcRequests, ...socketIoRequests],
+        requestGroups,
```
jackkav, line 258: This was most of what was broken about imports before, if im not mistaken

packages/insomnia/src/common/import.ts
```diff
@@ -598,21 +603,14 @@ export const importResourcesToNewWorkspace = async ({
   const resources = resourceCacheItem.resources;
   const ResourceIdMap = new Map();
   let newWorkspace: Workspace;
-  // in order to support import from api spec yaml
-  if (resourceCacheItem?.importer?.id && isApiSpecImport(resourceCacheItem.importer)) {
+  // support import from both insomnia export and api spec yaml
+  if (resources.find(isApiSpec) || isApiSpecImport(resourceCacheItem.importer)) {
```
jackkav, line 607: this makes the next code support both v5 and oas

packages/insomnia/src/common/import.ts
```diff
@@ -624,95 +622,81 @@ export const importResourcesToNewWorkspace = async ({
...
   // Make sure the new workspace has required resources like base environment, cookie jar and workspaceMeta
   await models.environment.getOrCreateForParentId(newWorkspace._id);
-  await models.workspaceMeta.getOrCreateByParentId(newWorkspace._id);
+  const workspaceMeta = await models.workspaceMeta.getOrCreateByParentId(newWorkspace._id);
 
+  if (isGitProject(project)) {
+    await models.workspaceMeta.update(workspaceMeta, {
+      gitFilePath: `${newWorkspace.name}-${newWorkspace._id}.yaml`,
+    });
+  }
```
jackkav, line 699: moved this down since it was duplicated

gatzjames (approved):

packages/insomnia/src/common/import.ts
```diff
@@ -173,8 +173,12 @@ export async function scanResources(importEntries: ImportEntry[]): Promise<ScanR
       let v5Error = null;
 
       try {
-        const { data: insomnia5Import, error } = tryImportV5Data(contentStr);
-        v5Error = error;
+        let insomnia5Import: ExportedModel[] = [];
+        if (contentStr.startsWith('type: ')) {
```
gatzjames, line 177: Not sure if this will be true for all yaml imports.
Maybe we can improve the error logging instead?

jackkav: I'm confused why this is not a consistent convert function like the others, as there are already too many try catches here

PR #9633, opened 2026-02-03
remove duplicate route

jackkav:


CurryYangxx:
image
We can't delete this route; this route is used for this UI when the user has no project.
I am working on deleting unused code in this route and extract reusable component with the `project.$projectId._index` route (#9635 ).

PR #9640, opened 2026-02-05
feat: add URL backend for custom LLM endpoints

Ali-Sab:
### **Feature: Custom URL Backend**

Adds support for OpenAI-compatible LLM endpoints via custom URLs.

**Backend**

* Added `url` and `baseURL` to the `LLMConfig` interface.
* Updated config services to handle URL fields and filtering.
* Refactored code to use the `LLM_BACKENDS` constant.
* Added 15 new unit tests for backend operations.

**Frontend**

* **New URL Component:** Handles configuration, HTTP/HTTPS validation, and model fetching.
* **State Management:** Moved AI features to a unified `aiFeatures` object.
* **Navigation:** Added URL options to the AI settings menu.

**Consistency Updates**
Standardized the UI across Claude, Gemini, OpenAI, and GGUF:

* Added loading spinners and button disable logic.
* Standardized error messages and helper text.
* Linked documentation in the UI.

**Files Changed**

* `constants.ts` (Added URL backend type)
* `llm-config-service.ts` & `.test.ts` (Backend logic and tests)
* `ai-settings.tsx` & `url.tsx` (Settings UI and new component)
* `claude.tsx`, `gemini.tsx`, `openai.tsx`, `gguf.tsx` (UI polish)

### DIFF
--- a/packages/insomnia/src/common/constants.ts
+++ b/packages/insomnia/src/common/constants.ts
@@ -81,7 +81,7 @@ export const CHECK_FOR_UPDATES_INTERVAL = 1000 * 60 * 60 * 24;
 
 export const ACCEPTED_NODE_CA_FILE_EXTS = ['.pem', '.crt', '.cer', '.p12'];
 
-export const LLM_BACKENDS = ['gguf', 'claude', 'openai', 'gemini'] as const;
+export const LLM_BACKENDS = ['gguf', 'claude', 'openai', 'gemini', 'url'] as const;
 
 // Available editor key map
 export enum EditorKeyMap {
--- a/packages/insomnia/src/ui/components/settings/llms/url.tsx
+++ b/packages/insomnia/src/ui/components/settings/llms/url.tsx
@@ -4,13 +4,28 @@ import { Button, Input, Text } from 'react-aria-components';
 import type { LLMBackend, LLMConfig } from '~/main/llm-config-service';
 import { Icon } from '~/ui/components/icon';
 
-interface AnthropicModelData {
-  type: string;
+const URL_BACKEND: LLMBackend = 'url';
+
+interface LLMModelData {
   id: string;
-  display_name: string;
-  created_at: string;
+  object: string;
+  created?: number;
+  owned_by?: string;
 }
-export const Claude = ({
+
+const validateUrl = (urlString: string): boolean => {
+  if (!urlString) {
+    return false;
+  }
+  try {
+    const parsedUrl = new URL(urlString);
+    return parsedUrl.protocol === 'http:' || parsedUrl.protocol === 'https:';
+  } catch {
+    return false;
+  }
+};
+
+export const Url = ({
   saveLLMSettings,
   configuredLLMs,
   currentLLM,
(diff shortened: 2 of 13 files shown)


packages/insomnia/src/common/constants.ts
```diff
@@ -81,7 +81,7 @@ export const CHECK_FOR_UPDATES_INTERVAL = 1000 * 60 * 60 * 24;
 
 export const ACCEPTED_NODE_CA_FILE_EXTS = ['.pem', '.crt', '.cer', '.p12'];
 
-export const LLM_BACKENDS = ['gguf', 'claude', 'openai', 'gemini'] as const;
```
shelby-moore, line 84: https://github.com/Kong/insomnia/blob/develop/packages/insomnia/src/plugins/types.ts#L3-L25 is used in a few places and is missing the new backend type, along with the url field. Right now, the generate commit diffs and mcp sampling both use this interface and would likely fail if the url backend were configured

packages/insomnia/src/ui/components/settings/llms/url.tsx
```diff
@@ -0,0 +1,225 @@
...
+      const realUrl = urlOverride || url;
+      if (!validateUrl(realUrl)) {
+        setError('Please enter a valid HTTP or HTTPS URL.');
+        return;
+      }
+
+      try {
+        setIsLoadingModels(true);
+        setError(null);
+        const response = await fetch(`${realUrl}/models`);
```
shelby-moore: Should we be wrapping this in a timeout for this or some sort of option to prevent it from hanging if it's incorrect?

shelby-moore: Should we be using new URL to get better path safety?

packages/insomnia/src/common/constants.ts
Ali-Sab: Good catch. Let me update this, and some other fields in this file to show what the LLM URL option can support

packages/insomnia/src/ui/components/settings/llms/url.tsx
Ali-Sab: adding both of these in the next revision, let me know your thoughts on them 🙏

Ali-Sab:
Additional changes since initial commit:

- Added 'url' backend and url field to ModelConfig in types.ts so commit generation and MCP sampling work with the URL backend
- Improved URL fetch robustness: 10s timeout via AbortController, safe path construction with new URL(), JSON parse guard, and user feedback when no models are found
- Found issue where if MCP Server specifies a token count that is too low, than that parameter is passed on and can limit max tokens to a number that isn't enough for the LLM to reply, which results in failures that are difficult to debug. Fixed by enforcing min token count.

packages/insomnia/src/main/ipc/main.ts
```diff
@@ -544,13 +544,25 @@ export function registerMainHandlers() {
...
 
+      const finalModelConfig = {
+        ...modelConfig,
+        ...modelConfigFromSamplingRequest,
+      };
+
+      // Enforce minimum maxTokens to prevent unreasonably low limits from MCP servers
+      // that would result in truncated/incomplete responses and silent failures.
+      const MIN_MAX_TOKENS = 1000;
+      if (finalModelConfig.maxTokens !== undefined && finalModelConfig.maxTokens < MIN_MAX_TOKENS) {
```
shelby-moore: Should we instead be surfacing an error or enforcing this in the settings so this behaviour is more visible?

Ali-Sab: Ideally, yes. I'd rather not have some automatically enforced value and instead give the user the ability to specify a max token override. 

I can try adding something of that sort, like a Max Token Override in the settings. We should ideally have this for all types of AI models, but it felt like this PR might get too big. Thoughts going forward?

shelby-moore: Perhaps we should add this minimum max token enforcement as part of a follow-up in that case?

Ali-Sab: I'd be okay with that, and we can try a better quality solution with it. I'll make a rev to remove this out for now

shelby-moore (approved):

ryan-willis (approved):

PR #9649, opened 2026-02-09
fix: Improve preferences data access folders input UX with validation and guidance

### LINKED ISSUE #9064: Insomnia cannot access the file ... . You can adjust this in Preferences → Security.
### Expected Behavior

When I make a POST request query with a file as the request body it uploads the file

### Actual Behavior

Since the last Insomnia or Mac OS update it is no longer is able to do queries with files attached. 

Image

Giving Full Disk Access permissions to Insomnia in the Mac OS Privacy & Security settings does not resolve the issue. Unfortunately the Settings pane does not allow to give permission for the downloads folder, it needs to be requested by the app which it apparently does not do.

### Reproduction Steps

See screenshot above

- [x] Git sync.
- [ ] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

_No response_

### Insomnia Version

Version: Insomnia 11.5.0 Build date: 8/26/2025 OS: Darwin arm64 24.6.0 Electron: 37.2.6 Node: 22.17.1 Node ABI: 136 V8: 13.8.258.30-electron.0 Architecture: arm64

macOS

### Operating System Version

Mac OS Sequoia 15.6.1 (24G90)

### Installation method

Download pkg from Github

_No response_

miycctwn:
## Summary

Improves the UX of the "File Access" data folders setting (Preferences → General → Security -> "What folders can Insomnia access?") to address confusion reported in #9064 (specifically [this comment](https://github.com/Kong/insomnia/issues/9064#issuecomment-3323827519)), where users found the Add button misleading — it wasn't clear that a path must be entered first, and clicking Add with an empty input silently did nothing.

This PR does **not** close #9064 (which covers broader file access UX concerns) but addresses the specific input usability pain points discussed there.

 ## Changes

  - On-click validation for empty input: Clicking Add with an empty input now shows a red-bordered input with the message "Enter a folder path to add." instead of silently doing nothing.
  - Duplicate folder path validation: If the user enters a path that already exists in the list and clicks Add, the input is
  highlighted with a red border and a "Duplicate folders are not allowed." message is shown — preventing duplicate entries.
  - Trailing slash normalization: Paths like /Users/foo and /Users/foo/ are as duplicates by stripping trailing slashes before comparison.
  - Validation clears on edit: The error state is dismissed as soon as the user modifies the input.
  - Placeholder example: Added a placeholder (e.g., /Users/folder1) to the data folders input to guide users on the expected
  format.

https://github.com/user-attachments/assets/3d300146-f57d-4b18-9b1f-a26deb94e663

  Related to #9064

### DIFF
--- a/packages/insomnia/src/ui/components/settings/text-array-setting.tsx
+++ b/packages/insomnia/src/ui/components/settings/text-array-setting.tsx
@@ -8,7 +8,6 @@ import type { SettingsOfType } from '../../../common/settings';
 import { useSettingsPatcher } from '../../hooks/use-request';
 import { PromptButton } from '../base/prompt-button';
 import { HelpTooltip } from '../help-tooltip';
-import { Tooltip } from '../tooltip';
 
 export const TextArraySetting: FC<{
   disabled?: InputHTMLAttributes<HTMLInputElement>['disabled'];
@@ -22,6 +21,7 @@ export const TextArraySetting: FC<{
 
   const patchSettings = useSettingsPatcher();
   const [folderToAdd, setFolderToAdd] = useState('');
+  const [validationError, setValidationError] = useState('');
 
   let currentValue = settings[setting];
   if (!Array.isArray(currentValue)) {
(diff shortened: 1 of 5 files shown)


packages/insomnia/src/ui/components/settings/text-array-setting.tsx
```diff
@@ -29,12 +30,20 @@ export const TextArraySetting: FC<{
...
-    const exists = currentValue.includes(validValue);
-    if (folderToAdd !== '' && !exists) {
-      const updatedValue = [...currentValue, validValue];
-      patchSettings({ [setting]: updatedValue });
+    if (validValue === '') {
+      setValidationError('Enter a folder path to add.');
+      return;
+    }
+    const normalizedValue = validValue.replace(/\/+$/, '') || validValue;
+    const exists = currentValue.some(v => (v.replace(/\/+$/, '') || v) === normalizedValue);
```
shelby-moore: A few concerns with these regex replaces:

- if `validValue` is something like `///`, the regex replace will output an empty string, which will cause a fallback and `///` to be stored
- The regex makes sense for unix style paths, but it's not compatible with windows

An alternative here could perhaps be a combination of path.normalize (to collapse duplicate separators and resolve ./.. segments) and stripping trailing separators with `/[/\\]+$/` to handle both platforms.

miycctwn: Hello @shelby-moore , 
Great catch! I initially overlooked Windows path handling. Thanks for the suggestions and these give me ideas to add unit tests for the path handler function  `normalizeFolderPath` to ensure covering different OS systems and UX error messaging for invalid paths. 

The fix has been pushed! Please take another look whenever you have time. Thank you.

image

shelby-moore: Love the addition of tests :)

shelby-moore (approved):
Thank you for your contribution, @miycctwn! 🚀

PR #9660, opened 2026-02-19
feat: skip curl import process

jackkav:
motivation: lower friction from deeplink to request send time. Also support mcp.
If curl is valid, we redirect directly to the request.

TODO:
- [x] add inline import logic and redirect before starting ux
- [x] add scope param for mcp
- [x] add label for mcp workspace name, search in current project, or default new workspace name
- [x] add mcp and label drilling for regular ux

Cases supported

create new collection/add to existing with label name
`insomniadev://app/import?label=Testname&curl=curl 'http://en.wikipedia.org/'`
invalid input
`insomniadev://app/import?label=Testname&curl=curl broken input`
create new mcp client
`insomniadev://app/import?label=Testname&scope=mcp&curl=curl 'http://en.wikipedia.org/'`

### DIFF
--- a/packages/insomnia/src/root.tsx
+++ b/packages/insomnia/src/root.tsx
@@ -19,6 +19,7 @@ import {
 } from 'react-router';
 
 import { EXTERNAL_VAULT_PLUGIN_NAME, isDevelopment } from '~/common/constants';
+import { setDefaults } from '~/main/importers/utils';
 import * as models from '~/models';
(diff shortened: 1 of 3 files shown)


ryan-willis (approved):
lookin' good

ZxBing0066:
Cool. Is it better to use another key instead of curl when importing an MCP?

ryan-willis:
these changes have been rolled into https://github.com/Kong/insomnia/pull/9733

PR #9676, opened 2026-03-02
fix: Export raw response without any interpretation by default [INS-2147]

### LINKED ISSUE #9619: Response body "Export raw response" corrupts ZIP
### Expected Behavior

When an API returns a ZIP file (`Content-Type: application/zip`), using **"Export Raw Response"** should save the exact raw binary response without any modification, producing a valid ZIP file.

### Actual Behavior

After the latest Insomnia update:

- Insomnia shows the response size in the UI as ~271.7 KB
- When using **"Export Raw Response"**, the saved file becomes ~492 KB
- The resulting ZIP file is corrupted and cannot be opened or extracted
- Windows reports the archive as invalid/corrupted

The same POST request produces a valid ZIP file when tested with Postman.

Image

Image

Image

### Reproduction Steps

1. Create a POST request to an endpoint that returns a ZIP file.
2. Send the request in Insomnia.
3. Confirm the response size shown in the response panel (e.g. ~271.7 KB).
4. Click **"Export Raw Response"** and save the file as `.zip`.
5. Try to open or extract the ZIP file on Windows.
6. Observe that the ZIP file is corrupted.

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

- The issue is **reproducible on multiple machines**
  - Me
  - Two other teammates tested independently and encountered the same issue
- Backend response is confirmed to be valid
- Mobile app, extracting the data and working fine in production, there is no problem
- Postman exports the same response correctly without corruption
- Since **"Export Raw Response"** is used, this appears to be a binary handling issue (raw bytes not preserved)

### Insomnia Version

12.3.0

Windows

### Operating System Version

Windows 10 Pro 22H2

### Installation method

insomnia.rest

_No response_

godfrzero:
Insomnia's "Export raw response" feature has been interpreting the response buffer as a utf-8 string, since #9439, before writing to file. This leads to byte sequences which are invalid UTF-8, like PNG magic bytes or other sequences in Zip files, to be replaced by a Unicode replacement character �. Ref: https://nodejs.org/docs/latest-v22.x/api/buffer.html#buffers-and-character-encodings

```
# Other patterns: https://en.wikipedia.org/wiki/List_of_file_signatures

> sevenzip = Buffer.from([0x37, 0x7A, 0xBC, 0xAF, 0x27, 0x1C])
<Buffer 37 7a bc af 27 1c>
> sevenzip.toString('utf-8')
"7z��'\x1C"

> pngmagic = Buffer.from([0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a])
<Buffer 89 50 4e 47 0d 0a 1a 0a>
> pngmagic.toString('utf-8')
'�PNG\r\n\x1A\n'
```

This PR restores the previous behavior and writes an uninterpreted buffer to file by default, reserving utf-8 interpretation for prettified JSON. It also adds unit tests to cover various branches in the `downloadResponseBody` method, to prevent a similar bug in the future. Note that the tests don't _actually_ write to file, but just tests the data that's eventually passed to `fs.writeFile`.

Changes were verified locally by pulling raw bytes via httpbin, e.g. https://httpbin.org/bytes/32, and verifying that the exported file maintained the expected size:
image

Closes #9619 INS-2147


packages/insomnia/src/ui/components/panes/response-pane-utils.ts
godfrzero, line 1: Isolated to a separate file to avoid pulling react and other complex dependencies in for unit tests.

ryan-willis (approved):

PR #9689, opened 2026-03-04
refactor: migrate MCP-related models and services to use the new insomnia-data package

ZxBing0066:
## Background

[INS-2201](https://konghq.atlassian.net/browse/INS-2201)

- This PR shows how to migrate the existing models into insomnia-data to drop off the runtime dependencies and make it clearer for each module.

## Changes

- Split existing models to insomnia-data/models(only the static configurations and interfaces, and type guards), insomnia-data/services(the data operations), `services` depends on the runtime, so it uses IoC.
- Updated various components and routes to import MCP-related interfaces and services from '~/insomnia-data' instead of local models.

## Note

This PR focuses on reconstructing the code, not recoding the existing implementation.

The key is to decide the usage code of insomnia-data, other implementations could always be easily adjusted at any time.

```ts
import { type McpRequest, models, services } from '~/insomnia-data';

await models.mcpPayload.isMcpPayloadId('123');

await services.caCertificate.all();
```

This PR ensures all the data related interface, configurations, and operations export by the same way, and reduces mental load on devs.

[INS-2201]: https://konghq.atlassian.net/browse/INS-2201?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/insomnia/src/insomnia-data/node-src/services/mcp-payload.ts
```diff
@@ -0,0 +1,41 @@
...
+    const newPatch = Object.assign(
+      {
+        parentId,
+      },
+      patch,
+    );
+    return baseOperations.create(newPatch);
+  },
+
+  async getOrCreateByParentIdAndUrl(parentId: string, url: string) {
```
CurryYangxx, line 29: This function is unused, and the logic is incomplete. I think we can remove this function.

packages/insomnia/src/insomnia-data/node-src/services/base.ts
```diff
@@ -0,0 +1,30 @@
...
+  type T = ModelMap[U];
+
+  return {
+    async create(patch: Partial<T>): Promise<T> {
+      return await database.docCreate<T>(modelType, patch);
+    },
+    async getById(id: string): Promise<T | undefined> {
+      return await database.findOne<T>(modelType, { _id: id });
+    },
+    async getByParentId(parentId: string): Promise<T | undefined> {
```
CurryYangxx, line 14: Semantically, `getByParentId` can return a list, such as a project's collections, but we use `findOne` inside. We should let this function more general.

packages/insomnia/src/insomnia-data/node-src/services/mcp-response.ts
```diff
@@ -0,0 +1,76 @@
+import { database as db, type McpResponse, type Services } from '~/insomnia-data';
+import * as models from '~/models';
+
+import { createBaseOperations } from './base';
+
+const type = models.mcpResponse.type;
```
CurryYangxx, line 6: We still use the old model here.

ZxBing0066: Good catch, this is because we're using the legacy models in the implementation. This will be replaced at the final PR of model refactoring.

packages/insomnia/src/insomnia-data/node-src/services/mcp-payload.ts
ZxBing0066: Will do the clean work after the migration.

packages/insomnia/src/insomnia-data/node-src/services/base.ts
ZxBing0066: Will rename or clean after the migration.

packages/insomnia/src/entry.main.ts
```diff
@@ -51,6 +52,9 @@ initializeLogging();
 
 initializeSentry();
 
+// Initialize services for main process
+initServices(servicesNodeImpl);
```
cwangsmv: Since the services are database-dependent, can we move the `initServices(servicesNodeImpl)` right after `await initDatabase(mainDatabase);` which will not make the state fragile and follows the same pattern in `entry.client`

packages/insomnia/src/insomnia-data/src/services/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface Services {
+  caCertificate: CaCertificateService;
+  mcpRequest: McpRequestService;
+  mcpResponse: McpResponseService;
+  mcpPayload: McpPayloadService;
+}
+
+export function initServices(impl: Services) {
+  services = impl;
```
cwangsmv, line 19: Nit: For the service init and database init, should we add an init protection check to avoid duplicate init by mistake?

packages/insomnia/src/insomnia-data/src/services/base.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface ModelMap {
+  CaCertificate: CaCertificate;
+  McpRequest: McpRequest;
+  McpResponse: McpResponse;
+  McpPayload: McpPayload;
+}
+
+export type AllTypes = keyof ModelMap;
+
+export interface BaseServices<T extends AllTypes> {
```
cwangsmv, line 12: After introducing dependency injection, it is difficult to find real implementation, we can only find the interfaces in IDE. 
It's better to add navigation hints for real implementation in comment like:
```
/**
 * Implementations:
 *   - Node/main:  node-src/services/ca-certificate.ts → caCertificateService
 *                 node-src/services/mcp-request.ts    → mcpRequestService
 *                 node-src/services/mcp-response.ts   → mcpResponseService
 *                 node-src/services/mcp-payload.ts    → mcpPayloadService
 *
 *  - Renderer process:
 *   In the renderer, per-model services are exposed via the contextBridge as
 *   `window._dataServices` and consumed via `initServices()` in entry.client.tsx.
 *   Do NOT call node-src implementations directly from renderer code — always
 *   go through the `services` export from `~/insomnia-data/src/services/index.ts`.
 *
 */
```

ZxBing0066: Good point, but I have no idea about this, since we split the interface and the implementation. I don't want to add comments to all the models; it's hard to maintain. I added some flow chat in the insomnia-data readme. Hope that once all the team members know how it works, it will not be a problem.

packages/insomnia/src/entry.main.ts
ZxBing0066: 👍 Moved. But actually, the init order doesn't affect anything since init itself doesn't depend on others.

packages/insomnia/src/insomnia-data/src/services/index.ts
ZxBing0066: Good point, I added a variable to mark it.

CurryYangxx (approved):

PR #9697, opened 2026-03-05
refactor: migrate remaining apis - [INS-2162]

CurryYangxx:
INS-2162

Migrate remaining APIs into the `insomnia-api` package which includes:

- invite encrypt
- mockbin
- vcs graphql
- some fragmented apis

### DIFF
--- a/packages/insomnia-api/src/organizations.ts
+++ b/packages/insomnia-api/src/organizations.ts
@@ -4,9 +4,12 @@ interface Branding {
   logo_url: string;
 }
 
-export interface Metadata {
-  organizationType: string;
+export type OrganizationType = 'personal' | 'team' | 'enterprise';
+
+export interface OrganizationMetadata {
+  organizationType: OrganizationType;
   ownerAccountId: string;
+  description?: string;
 }
 
 export interface Organization {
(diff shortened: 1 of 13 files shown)


packages/insomnia-api/src/organizations.ts
```diff
@@ -4,17 +4,20 @@ interface Branding {
   logo_url: string;
 }
 
-export interface Metadata {
-  organizationType: string;
+export type OrganizationType = 'personal' | 'team' | 'enterprise';
+
+export interface OrganizationMetadata {
+  organizationType: OrganizationType;
   ownerAccountId: string;
```
ZxBing0066, line 11: In the previous invite-modal.tsx, this field is defined as optional. Is it safe to change as a required one?

packages/insomnia-api/src/organizations.ts
```diff
@@ -4,17 +4,20 @@ interface Branding {
...
+  description?: string;
 }
 
 export interface Organization {
   id: string;
   name: string;
   display_name: string;
   branding?: Branding;
```
ZxBing0066, line 19: This interface is also inconsistent. 
- Optional or required
- Missing field colors

ZxBing0066 (approved):

packages/insomnia-api/src/organizations.ts
CurryYangxx: https://github.com/Kong/insomnia-api/blob/main/internal/api/routes_organizations.go#L1408
According to the backend code, the `ownerAccountId` always exists.

PR #9700, opened 2026-03-06
feat: implement hotkey conflict detection and display warnings by toast notification on startup

ZxBing0066:
## Background

[INS-2184](https://konghq.atlassian.net/browse/INS-2184)

- When users customize the shortcut in an old version, and that shortcut is used as the default value of a new shortcut, it causes a conflict.
- The sidebar_focusFilter is not used anywhere

## Changes

- Detect hotkey conflict on app startup
- Remove sidebar_focusFilter
- Supports in the toast line breaks via `\n`

## UI

image

[INS-2184]: https://konghq.atlassian.net/browse/INS-2184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/insomnia/src/root.tsx
+++ b/packages/insomnia/src/root.tsx
@@ -1,5 +1,6 @@
 import { config } from '@fortawesome/fontawesome-svg-core';
 import type { IpcRendererEvent } from 'electron';
+import { once } from 'es-toolkit/function';
 import type { FC } from 'react';
 import { useEffect, useState } from 'react';
 import { Button } from 'react-aria-components';
@@ -19,6 +20,8 @@ import {
 } from 'react-router';
 
 import { EXTERNAL_VAULT_PLUGIN_NAME, isDevelopment } from '~/common/constants';
+import { detectHotKeyConflicts } from '~/common/hotkeys';
+import type { HotKeyRegistry } from '~/common/settings';
 import * as models from '~/models';
 import type { Settings } from '~/models/settings';
 import type { UserSession } from '~/models/user-session';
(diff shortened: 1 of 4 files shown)


packages/insomnia/src/root.tsx
```diff
@@ -1,5 +1,6 @@
 import { config } from '@fortawesome/fontawesome-svg-core';
 import type { IpcRendererEvent } from 'electron';
+import { once } from 'es-toolkit/function';
```
CurryYangxx, line 3: We don't have this package in the package.json

packages/insomnia/src/root.tsx
```diff
@@ -304,6 +307,25 @@ export const HydrateFallback = () => {
   );
 };
 
+// Only detect hotkey conflicts once per app launch, since the detection can be expensive and hotkey registry is unlikely to change frequently during a session. If we want to detect more frequently in the future, we can consider adding a manual "Detect Conflicts" button in the Keyboard Shortcuts settings page that calls the detectHotKeyConflicts function directly.
```
CurryYangxx, line 310: Do we need to check for conflicts when a user sets the hotkey?

packages/insomnia/src/root.tsx
ZxBing0066: Actually, we have it in the package of the scripting environment. It should be moved to the insomnia package, but not the scope of this PR.

packages/insomnia/src/root.tsx
ZxBing0066: Conflict check on set already exists.

CurryYangxx (approved):

PR #9715, opened 2026-03-17
Separate GUI & internal state of runner iteration count [INS-2130]

godfrzero:
This PR allows users to temporarily clear the Iterations input in the Runner without impacting the internal iterationCount that Insomnia tracks. In the current Insomnia version users can't change the iteration count in the input by deleting the existing value and typing a new one. For example, if the iteration count is set to 1 and the user would like to change it to 5. The following alternatives work:
- Highlighting 1 and typing 5.
- Using the Up arrow key to increment the number input to 5.

This PR separates the local state displayed in the GUI from the internal state that's actually used during execution. The two states only deviate when the local state is an empty string.

A couple smoke tests have been added. The changes were tested locally as well by:
- Clearing a single digit iteration count and then typing a different digit.
- Clearing a single digit iteration count and then typing the same digit.
- Clearing the iteration input and clicking Run without typing anything.
- Uploading a data file and verifying that the iteration count updates to match the number of rows.


packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.runner.tsx
ihexxa, line 521: nit: Hm probably we could also enable the mechanism for this input.

ihexxa (approved):
Thanks for fixing this 👍.

packages/insomnia/src/routes/organization.$organizationId.project.$projectId.workspace.$workspaceId.debug.runner.tsx
godfrzero: Good point! I'll send a separate PR.

godfrzero: @ihexxa I've created #9721 as a follow-up. I refrained from abstracting the logic into a separate component since it looked like it would only be used in two places at the moment.

PR #9717, opened 2026-03-17
feat: add sdk snippet generation integration

shelby-moore:
This PR introduces a secondary "mode" to the existing `Generate Client Code` feature for requests. When available, Stainless will be used to generate the client code snippet. When unavailable, the codepath falls back to the existing http snippet client code generation.


packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -38,32 +44,73 @@ export interface State {
...
+  } catch {
+    return undefined;
+  }
+}
+
+function capitalize(s: string) {
+  return s.charAt(0).toUpperCase() + s.slice(1);
+}
+
+type HarRequest = NonNullable<Awaited<ReturnType<typeof exportHarWithRequest>>>;
```
ryan-willis: Simpler: `import type { Request as HarRequest } from 'har-format';` (or re-export that from the common `har` module)

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -38,32 +44,73 @@ export interface State {
...
+
+async function fetchSdkSnippet(sdk: Sdk, language: string, har: HarRequest) {
+  const pathname = new URL(har.url).pathname;
+  const parameters = [
+    ...har.queryString.map(({ name, value }) => ({ in: 'query' as const, name, value })),
+    ...har.headers.map(({ name, value }) => ({ in: 'header' as const, name, value })),
+  ];
+  const body = har.postData?.text ? tryParseJson(har.postData.text) : undefined;
```
ryan-willis: Since this will be empty for other body formats we support, should we only show the Stainless options if there is no body or a set request body's `mimeType` can be parsed as JSON? (Maybe this is more a question for Stainless) can we translate from other formats to make the generate call?

Generating an SDK snippet without the body when the request itself has it doesn't feel right to me

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -38,32 +44,73 @@ export interface State {
...
+      snippet: '',
+      mode: 'httpsnippet',
+      sdkLoading: false,
+    };
   });
 
-  const [snippet, setSnippet] = useState<string>('');
+  // Refs to avoid stale closures in callbacks
+  const sdkRef = useRef(state.sdk);
+  sdkRef.current = state.sdk;
```
ryan-willis: This is a synchronization of reactive state to a ref that is run for every change (not just changes to `state.request` and `state.sdk`), it's better to refer to the state fields directly in the callback and add them to the  dependency arrays:

```ts
useCallback(() => {
  doSomething(state.sdk);
  doAnother(state.request);
}, [state.sdk, state.request]);
```

State in callbacks won't be stale if those dependencies are specified

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -38,32 +44,73 @@ export interface State {
   target?: HTTPSnippetTarget;
   client?: HTTPSnippetClient;
   targets: HTTPSnippetTarget[];
+  snippet: string;
+  mode: 'loading' | 'httpsnippet' | 'sdk';
```
ryan-willis: I don't think we should have `loading` as a mode, and I think this should use `stainless-sdk` instead of `sdk` since that mode is domain-specific to Stainless

`sdkLoading`->`loading` and can be used for loading indicators in any mode

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -15,14 +17,18 @@ import { ModalBody } from '../base/modal-body';
...
+// Maps language/target keys to CodeMirror modes where the name differs
+const LANGUAGE_MODE_MAP: Record<string, string> = {
   c: 'clike',
   java: 'clike',
   csharp: 'clike',
   node: 'javascript',
   objc: 'clike',
   ocaml: 'mllike',
+  typescript: 'javascript',
+  kotlin: 'clike',
```
ryan-willis, line 30: 🕺🏻

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -1,9 +1,11 @@
 import type { HTTPSnippetClient, HTTPSnippetTarget } from 'httpsnippet';
+import { generateSdkSnippet, getSdkByEndpoint, type Sdk } from 'insomnia-api';
```
ryan-willis: these import names are a bit too generic IMO, maybe we call the `insomnia-api` file `integrations.ts` and namespace these within to `generateStainlessSdkSnippet`, etc.

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -135,71 +265,113 @@ export const GenerateCodeModal = forwardRef<GenerateCodeModalHandle, Props>((pro
...
+        <div className="pad flex items-center justify-between">
+          <div className="flex items-center gap-2">
+            {mode === 'loading' ? null : mode === 'sdk' && sdk ? (
+              <Dropdown
+                aria-label="Select a language"
+                isDisabled={sdkLoading}
+                placement="bottom start"
+                triggerButton={
+                  <Button className="h-(--line-height-xs) rounded-md border border-solid border-(--hl-lg) px-(--padding-md) hover:bg-(--hl-xs)">
+                    {sdkLanguage ? capitalize(sdkLanguage) : 'n/a'}
```
ryan-willis: non-blocking: could use the `capitalize` Tailwind class instead for these

packages/insomnia/src/ui/components/modals/__tests__/generate-code-modal.test.ts
ryan-willis, line 1: There's a good chunk of mocking here that wouldn't be necessary if `fetchSdkSnippet` were pulled out of the component file into `insomnia-api` instead

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
shelby-moore: Yeah that's a good idea 👍

packages/insomnia/src/ui/components/modals/__tests__/generate-code-modal.test.ts
shelby-moore: I'll try pulling it in 👍

packages/insomnia/src/ui/components/modals/__tests__/generate-code-modal.test.ts
```diff
@@ -33,75 +15,86 @@ const baseHar = {
...
+    const result = harToSdkParams(mockSdk, 'typescript', {
+      ...baseHar,
+      method: 'POST',
+      postData: {
+        mimeType: 'application/x-www-form-urlencoded',
+        params: [{ name: 'username', value: 'alice' }, { name: 'age', value: '30' }],
+      },
+    });
+
+    expect(result.body).toEqual({ username: 'alice', age: '30' });
```
ryan-willis, line 92: yesss

packages/insomnia/src/ui/components/modals/generate-code-modal.tsx
```diff
@@ -46,28 +46,47 @@ export interface State {
...
-  ];
-  const body = har.postData?.text ? tryParseJson(har.postData.text) : undefined;
-  return generateSdkSnippet({ id: sdk.id, language, method: har.method, path: pathname, parameters, body });
+export function harToSdkParams(sdk: StainlessSdk, language: string, har: HarRequest) {
+  let body: Record<string, unknown> | undefined;
+  let bodyWarning: string | undefined;
+
+  if (har.postData) {
+    if (har.postData.params) {
+      body = Object.fromEntries(har.postData.params.map(({ name, value }) => [name, value ?? '']));
```
ryan-willis: TODO: how we'd support `this[that][more]` if the body structure in the spec is nested (and if this applies/works with the snippet api)

PR #9720, opened 2026-03-18
Update the logic for indentifing system changes [INS-2165]

yaoweiprc:
image

### Reworked system-change detection logic (significant-diff-detection.ts)

The previous implementation used the `diff` library to compare the original and modified text line-by-line, then searched for lines containing the string `"modified"` as a heuristic. This approach was fragile and only flagged the `modified` property specifically.

The new implementation:

- **Accepts a different input shape.** Instead of `(original, modified)` strings, `findSystemChangeLines` now takes `(modifiedYaml, lineChangeIntervals[])`, where `lineChangeIntervals` comes directly from Monaco's diff editor (`getLineChanges()`). This decouples system-change detection from having to re-diff the file independently.
- **Uses YAML AST parsing to locate `meta` blocks precisely.** The YAML document is parsed with `parseDocument` + `LineCounter`, and the AST is walked recursively to find every `meta` key and its value subtree. The exact 1-based line ranges are extracted from node offsets.
- **Applies a line-sweep algorithm to intersect intervals.** Change intervals (from Monaco) and system-change intervals (meta lines) are converted into start/end events, sorted, and swept in a single pass to find their overlapping regions. Adjacent or overlapping result intervals are then merged.
- **Returns `{ start, end }[]` instead of flat line number arrays.** This makes the result suitable for range-based decorations.

### Updated diff view decorations (`diff-view-editor.tsx`)

- `findSystemChangeLines` is now called inside a `setTimeout(..., 0)` to wait for Monaco's diff computation to complete before reading `getLineChanges()`.
- Pure-deletion blocks (`modifiedEndLineNumber < modifiedStartLineNumber`) are filtered out before processing, since there are no modified-side lines to decorate.
- `addLineDecoration` now takes `startLineNumber` and `endLineNumber` instead of a single line number. The first line of the range gets both the background highlight and the glyph margin icon; subsequent lines get only the background highlight, to avoid repeating the icon for multi-line meta blocks.


packages/insomnia/src/ui/components/diff-view-editor.tsx
```diff
@@ -101,18 +101,27 @@ export const DiffEditor = ({ original, modified, highlightSystemChange = false }
...
     // Add system change decorations if enabled
     if (highlightSystemChange) {
-      const systemChangeLines = findSystemChangeLines(original, modified);
-      systemChangeLines.originalLines.forEach(lineNumber => {
-        addLineDecoration(originalEditor, lineNumber);
-      });
-      systemChangeLines.modifiedLines.forEach(lineNumber => {
-        addLineDecoration(modifiedEditor, lineNumber);
-      });
+      setTimeout(() => {
```
CurryYangxx, line 106: https://microsoft.github.io/monaco-editor/typedoc/interfaces/editor_editor_api.editor.IDiffEditor.html#onDidUpdateDiff
We can try to use `onDidUpdateDiff ` api to replace settimeout, this seems to be more reliable

packages/insomnia/src/common/significant-diff-detection.ts
```diff
@@ -184,83 +183,159 @@ export function hasSignificantChanges(
...
-  const modifiedLines: number[] = [];
+  modifiedYaml: string,
+  lineChangeIntervals: { modifiedStartLineNumber: number; modifiedEndLineNumber: number }[],
+) {
+  const intersectIntervals: { start: number; end: number }[] = [];
 
   try {
-    const changes = diffLines(original, modified);
+    // Get all line numbers (1-based, inclusive) spanned by a YAML AST node
+    function getNodeLineInterval(node: any, lineCounter: LineCounter) {
```
CurryYangxx: Move this function out of `findSystemChangeLines`.

packages/insomnia/src/common/significant-diff-detection.ts
```diff
@@ -184,83 +183,159 @@ export function hasSignificantChanges(
...
+      if (!node?.range) return;
+      const [start, , end] = node.range as [number, number, number];
+      const startLine = lineCounter.linePos(start).line;
+      const endLine = end > start ? lineCounter.linePos(end - 1).line : startLine;
+      if (endLine < startLine) return;
+      return { start: startLine, end: endLine };
+    }
+
+    // Recursively find all line numbers belonging to 'meta' keys in a YAML string
+    function findMetaLineIntervals(yamlString: string) {
```
CurryYangxx: Move out of this scope.

packages/insomnia/src/common/significant-diff-detection.ts
```diff
@@ -184,83 +183,159 @@ export function hasSignificantChanges(
...
+      const lineCounter = new LineCounter();
+      const doc = parseDocument(yamlString, { lineCounter });
+      const retIntervals: { start: number; end: number }[] = [];
+
+      function walk(node: any) {
```
CurryYangxx: `node` should have a ts definition from the `yaml` pkg.

packages/insomnia/src/ui/components/diff-view-editor.tsx
```diff
@@ -141,4 +155,19 @@ function addLineDecoration(editor: monaco.editor.IStandaloneCodeEditor, lineNumb
...
+        options: {
+          className: 'system-change-line-bg',
+          hoverMessage: hoverMessage,
+          glyphMarginHoverMessage: hoverMessage,
+          isWholeLine: true,
+          showIfCollapsed: false,
+        },
+      },
+    ]);
+  }
```
CurryYangxx, line 172: We could calculate the linenumber first and call `createDecorationsCollection` only 1 time.

packages/insomnia/src/common/significant-diff-detection.ts
```diff
@@ -184,83 +183,159 @@ export function hasSignificantChanges(
...
-): {
-  originalLines: number[];
-  modifiedLines: number[];
-} {
-  const originalLines: number[] = [];
-  const modifiedLines: number[] = [];
+  modifiedYaml: string,
+  lineChangeIntervals: { modifiedStartLineNumber: number; modifiedEndLineNumber: number }[],
+) {
+  const intersectIntervals: { start: number; end: number }[] = [];
```
CurryYangxx: ```{ start: number; end: number }[]```
This type has appeared many times, we can define a common type for it.

packages/insomnia/src/ui/components/__tests__/diff-view-editor.test.ts
```diff
@@ -2,133 +2,107 @@ import { describe, expect, it } from 'vitest';
...
 
-describe('findSystemChangeLines()', () => {
-  it('should return empty arrays when original and modified are identical', () => {
-    const content = `name: test
-modified: 2024-01-01
-value: 123`;
-
-    const result = findSystemChangeLines(content, content);
+// Helper to build a lineChangeIntervals entry
+function interval(start: number, end: number) {
```
CurryYangxx: buildInterval could be a better name.

packages/insomnia/src/ui/components/diff-view-editor.tsx
yaoweiprc: I tried to use this api but sometimes I can not get the latest diff info inside onDidUpdateDiff

packages/insomnia/src/ui/components/diff-view-editor.tsx
yaoweiprc: The parameters passed to createDecorationsCollection between the two calls are different.
The first call has an additional glyphMarginClassName prop, which adds a glyph margin icon at the beginning of the line.

CurryYangxx (approved):

PR #9721, opened 2026-03-18
fix: Separate GUI & internal state of runner request delay [INS-2130]

godfrzero:
This PR implements a fix similar to the one in #9715, but for the request delay input in the runner. Details outlining the issue & fix are outlined in the previous. I've copied it over and adapted it for ease of access:

> This PR allows users to temporarily clear the Delay input in the Runner without impacting the internal delay that Insomnia tracks. In the current Insomnia version users can't change the delay in the input by deleting the existing value and typing a new one. For example, if the delay is set to 1 and the user would like to change it to 5. The following alternatives work:
> 
> * Highlighting 1 and typing 5.
> * Using the Up arrow key to increment the number input to 5.
> 
> This PR separates the local state displayed in the GUI from the internal state that's actually used during execution. The two states only deviate when the local state is an empty string.
> 
> A couple smoke tests have been added. The changes were tested locally as well by:
> 
> * Clearing a single digit delay and then typing a different digit.
> * Clearing a single digit delay and then typing the same digit.
> * Clearing the delay input and clicking Run without typing anything.

### DIFF
--- a/packages/insomnia-smoke-test/tests/smoke/runner.test.ts
+++ b/packages/insomnia-smoke-test/tests/smoke/runner.test.ts
@@ -239,6 +239,41 @@ test.describe('runner features tests', () => {
     await verifyResultRows(page, 0, 0, 4, expectedTestOrder, 1);
   });
 
+  test('delay input can be cleared to enter a new value', async ({ page }) => {
+    await page.getByTestId('run-collection-btn-quick').click();
+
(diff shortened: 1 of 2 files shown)


ihexxa (approved):
⚙️ 🔧

ryan-willis (approved):

PR #9724, opened 2026-03-19
feat: Extract vault and kv-pair utility methods out of environment model

cwangsmv:
## Background
For the environment model, certain functions do not depend on the DB and are not model operations. Move them to separate utils to simplify and clarify the model logic.

## Changes
- Move `encryptSecretValue` and `decryptSecretValue` to the vault utils.
- Move `getKVPairFromData`, `getDataFromKVPair` and `maskVaultEnvironmentData` to environment utils.
- Refine the `environment-utils` file to contain none-ui related logic. Extract a hook for common toggle environment type behavior.
- Unit test added for both utils.

[INS-2236](https://konghq.atlassian.net/browse/INS-2236)

[INS-2236]: https://konghq.atlassian.net/browse/INS-2236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/insomnia/src/ui/hooks/use-toggle-environment-type.tsx
```diff
@@ -0,0 +1,56 @@
+import { useCallback } from 'react';
+
+import { type Environment, type EnvironmentKvPairData, EnvironmentType } from '~/models/environment';
+import { showModal } from '~/ui/components/modals';
+import { AlertModal } from '~/ui/components/modals/alert-modal';
+import { AskModal } from '~/ui/components/modals/ask-modal';
+import { getKVPairFromData } from '~/utils/environment-utils';
+
+export function useToggleEnvironmentType() {
+  const toggleEnvironmentType = useCallback(
```
ZxBing0066, line 10: This function seems to be a pure function, and no need to be a hook. Is there any reason to change it from a function to be a hook?

ZxBing0066 (approved):
🚀

packages/insomnia/src/ui/hooks/use-toggle-environment-type.tsx
cwangsmv: My initial idea was to extract this as a pure function and place it in `environment-utils`.
However, since it contains UI-specific logic (modal interactions) and JSX, doing so would pollute `environment-utils` with UI logics and require it to become a `tsx` file.    

Since it is consumed across multiple files — route component, modal component and UI component. There's no better place except `ui/hooks/` to be the appropriate location.

ZxBing0066: Got it, thank you.

PR #9725, opened 2026-03-19
fix: Use flexbox layout in Invite modal to allow for scrolling on longer lists [INS-2101]

godfrzero:
Note that the diff is likely cleaner with whitespace-only changes ignored.

The "Invite collaborators" modal overflows beyond the fold when the list becomes long enough.

image

This PR centers the modal vertically to be consistent with other modals throughout the app and modifies the modal contents to use flexbox for layout throughout, allowing the list of collaborators to scroll vertically.

image

No automated tests have been added because testing for such scenarios is likely to just lead to unnecessary tests over time, but changes were manually tested in the following scenarios:
- Project with a single collaborator
- Project with multiple collaborators but not enough to cause overflow
- Project with enough collaborators to cause vertical overflow
- Project with enough collaborators to span multiple pages

I didn't manually test a case where there are _no_ collaborators on a project—I'm not sure if such a case is possible even though such a state is accounted for in the code.

### DIFF
--- a/packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
@@ -157,10 +157,10 @@ const InviteModal: FC<{
       onOpenChange={setIsOpen}
       className="theme--transparent-overlay fixed top-0 left-0 z-10 flex h-(--visual-viewport-height) w-full items-center justify-center bg-(--color-bg)"
     >
-      <Modal className="theme--dialog fixed top-[100px] h-fit w-full max-w-[900px] rounded-md border border-solid border-(--hl-sm) bg-(--color-bg) p-[32px] text-(--color-font)">
-        <Dialog className="relative outline-hidden">
+      <Modal className="theme--dialog flex max-h-[calc(var(--visual-viewport-height)-140px)] w-full max-w-[900px] flex-col overflow-hidden rounded-md border border-solid border-(--hl-sm) bg-(--color-bg) p-[32px] text-(--color-font)">
+        <Dialog className="relative flex h-full flex-1 flex-col overflow-hidden outline-hidden">
           {({ close }) => (
-            <>
+            <div className="flex h-full flex-col overflow-hidden">
               <Heading slot="title" className="mb-[24px] text-[22px] leading-[34px]">
                 Invite collaborators
               </Heading>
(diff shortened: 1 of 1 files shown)


godfrzero:
@pavkout I'd appreciate a pair of eyes on this whenever you get the time, and any thoughts in case you feel like the overall approach could be different.

packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
```diff
@@ -157,10 +157,10 @@ const InviteModal: FC<{
       onOpenChange={setIsOpen}
       className="theme--transparent-overlay fixed top-0 left-0 z-10 flex h-(--visual-viewport-height) w-full items-center justify-center bg-(--color-bg)"
     >
-      <Modal className="theme--dialog fixed top-[100px] h-fit w-full max-w-[900px] rounded-md border border-solid border-(--hl-sm) bg-(--color-bg) p-[32px] text-(--color-font)">
```
pavkout, line 160: @godfrzero Do we want to center the modal? Unless there is a specific reason for that, I’d suggest keeping it as `fixed top-[100px]`, since most of our modals use that approach.

packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
```diff
@@ -157,10 +157,10 @@ const InviteModal: FC<{
...
-        <Dialog className="relative outline-hidden">
+      <Modal className="theme--dialog flex max-h-[calc(var(--visual-viewport-height)-140px)] w-full max-w-[900px] flex-col overflow-hidden rounded-md border border-solid border-(--hl-sm) bg-(--color-bg) p-[32px] text-(--color-font)">
```
pavkout: @godfrzero We can avoid adding `overflow-hidden` here.

godfrzero: Ah, good catch. 🤦 
I've removed it.

packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
godfrzero: I've updated the PR to add a top padding on the ModalOverlay so we get the same visual outcome as keeping `fixed top-[100px]` on the Modal, mainly because having a flex parent with a single fixed child seems counterproductive.
I had a look at various modals throughout the app and there's inconsistencies in how the spacing is created—`py-[100px]`, `pt-[70px]` on the overlap, `top-[100px]` on the modal in other places—which is why I aimed for visual output.

pavkout:
> @pavkout I'd appreciate a pair of eyes on this whenever you get the time, and any thoughts in case you feel like the overall approach could be different.

@godfrzero  Tested this and everything looks good, just left a couple of small comments. 🚀

pavkout (approved):

packages/insomnia/src/ui/components/modals/invite-modal/invite-modal.tsx
pavkout: Yeah, I noticed the same inconsistency with the spacing across modals. Also, some modals are still centered on the screen. We can fix this in a separate PR.

ryan-willis (approved):

PR #9728, opened 2026-03-20
fix: Use json-bigint for parsing JSON in the response template tag to handle large numbers [#8988]

### LINKED ISSUE #8988: Unable to read 19/20 digit long value and then rounding it
### Expected Behavior

I'm using Postgres Enterprise DB version uses snowflake Id for primary key generation which gives me 19/20 digits long value. When REST API gives response for Request-X and then using that value to make a request to Y, then I see its rounding of that value.

I am using response body Tag and in the edit tag of JSONPath or XPath getting giving rounding value.

Image

Can you please look into this defect & fix it?

### Actual Behavior

I am using response body Tag and in the edit tag of JSONPath or XPath getting giving rounding value.

Image

Can you please look into this defect & fix it?

### Reproduction Steps

_No response_

- [ ] Git sync.
- [ ] Insomnia Cloud sync.
- [ ] Local only

### Additional Information

_No response_

### Insomnia Version

11.4.0

Windows

### Operating System Version

mac

### Installation method

insomnia.rest

_No response_

godfrzero:
#8988 is caused by JSON.parse incorrectly interpreting numbers more than 19-20 digits, resulting in "rounding" or scientific notation representation in unexpected places:
| Source Request | Referencing Request |
|----|----|
| image | image |

This PR uses `json-bigint` for parsing the JSON input instead, just for this particular tag. This allows large numbers in JSON to be parsed into JS strings, preserving the original "number" as it's being piped from response to request. The string representation shouldn't prove to be problematic since the resulting output is used in string form anyway.

Note that `json-bigint` is already an indirect dependency. The changes in this PR makes it a direct dependency. The package seems healthy enough to add as a dependency—15-20M weekly downloads, [no current vulnerabilities](https://security.snyk.io/package/npm/json-bigint). The last release was 5+ years ago, but that makes sense for a narrowly focused package.

Multiple unit tests have been added to cover various integer and floating point cases.

Closes #8988.


packages/insomnia/types/json-bigint.d.ts
ryan-willis: Add published types to `devDependencies` instead:
`npm -w insomnia i -D @types/json-bigint`

godfrzero:
@cwangsmv I added you as a reviewer on this one as well since it's a dependency of #9778, so I figured there's no point in reviewing that one if it'll end up just sitting around afterwards until this one is reviewed.

cwangsmv (approved):

PR #9738, opened 2026-03-23
Fix the bug that self-hosted mock server URL can not be edited when reating a new mock server [INS-2150]

yaoweiprc:
Fixed the bug that users can not edit the Self-hosted Mock Server URL when creating a self-hosted mock server within a local project.

For existing Mock Servers, changing the server type is no longer allowed. This is because we cannot automatically migrate existing Mock Routes to a new type. Permitting such changes would result in the mock routes becoming unavailable after the type modification. Only display the Mock Server type in the Mock Server Settings window.


packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
```diff
@@ -55,25 +55,19 @@ export const NewWorkspaceModal = ({
...
 }) => {
   const { organizationId } = useParams() as { organizationId: string; projectId: string };
 
-  const isLocalProject = !project.remoteId;
-  const isEnterprise = currentPlan?.type.includes('enterprise');
-  const isSelfHostedDisabled = !storageRules.enableLocalVault;
-  const isCloudProjectDisabled = isLocalProject || !storageRules.enableCloudSync;
-
-  const canOnlyCreateSelfHosted = isLocalProject && isEnterprise;
+  const isSelfHostedMockDisabled = !storageRules.enableLocalVault && !storageRules.enableGitSync;
```
ZxBing0066, line 69: It seems self-hosted mocks could also sync up using cloud sync. I think we should remove this check.

yaoweiprc: Our current app logic is:

Under a Cloud Project, users can create both cloud and self-hosted mock servers.
Under Local Projects and Git Projects, users can only create self-hosted mock servers.

This behavior is consistent with what we have in production.

ZxBing0066 (approved):

PR #9739, opened 2026-03-23
feat(Git Sync): Add support for canonical repository output

gatzjames:
# Overview

Prior to this work, Insomnia stored its git repositories with a layout that was internal to the app and invisible to standard Git tooling:

```
{userData}/version-control/git/{repoId}/
  git/          ← git internals (non-standard folder name)
  other/        ← all non-Insomnia files
  (Insomnia YAML files existed only in the NeDB database — never on disk)
```

This prevented users from running any native Git command (git log, git diff, git rebase, VSCode source-control, etc.) against a cloned repository, even though the data was conceptually a real Git repo. It also made it impossible to integrate Insomnia repositories into broader developer workflows.

## Goals

- Native Git compatibility — The repository on disk must be a first-class Git repository that any Git client can operate against without any knowledge of Insomnia internals.

- Bidirectional sync — Changes made inside Insomnia (new requests, environment edits, etc.) must appear as real file changes on disk. Changes made on disk (native git pull, merge, rebase, VSCode edits) must be reflected in the Insomnia UI and database.

- Zero data loss migration — Existing cloned repositories must be migrated silently and safely, without any action from the user.

- Correctness under concurrency — The sync loop must not produce duplicate writes, import loops, or data corruption when both Insomnia and native Git are operating near-simultaneously.

- Scope — The modern project-scoped code path (GitProjectNeDBClient) is the target. The legacy workspace-scoped path (NeDBClient / .insomnia/ directory) is left unchanged for backward compatibility.

### DIFF
--- a/packages/insomnia/src/main/git-service.ts
+++ b/packages/insomnia/src/main/git-service.ts
@@ -37,6 +37,7 @@ import { migrateToLatestYaml } from '../common/insomnia-schema-migrations';
 import { insomniaSchemaTypeToScope } from '../common/insomnia-v5';
 import * as models from '../models';
 import { fsClient } from '../sync/git/fs-client';
+import { migrateRepoStructureIfNeeded } from '../sync/git/git-repo-migration';
 import GitVCS, {
   fetchRemoteBranches,
   GIT_CLONE_DIR,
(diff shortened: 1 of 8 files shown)


packages/insomnia/src/main/git-service.ts
```diff
@@ -1255,6 +1277,9 @@ export const resetGitRepoAction = async ({ projectId, workspaceId }: { projectId
     });
   }
 
+  // Stop the file watcher for this repository (project-scoped flow only).
+  stopWatcher(repo._id);
+
```
CurryYangxx, line 1286: This `resetGitRepoAction` function will not be called in the project-scoped repos

PR #9765, opened 2026-03-31
Validate credentials

pavkout:
## Summary

- Adds a `validateCredentials` method to the GitHub and GitLab provider implementations that hits the provider's `GET /user` REST API endpoint — a lightweight, authoritative check that reliably returns HTTP 401 for revoked tokens or uninstalled GitHub Apps (unlike the git wire protocol, which can succeed anonymously on public repos).
- Adds an optional `validateCredentials` method to the `GitRemoteProvider` interface, with a fallback to `fetchRemoteBranches` for providers that don't implement it.
- Exposes a new `validateGitRepositoryCredentials` IPC handler in `git-service.ts` that looks up the stored credential for a project/workspace and delegates to the appropriate validation strategy.
- Adds a `git.validate-credentials` route and a `useGitValidateCredentialsFetcher` hook so the UI can trigger credential validation on demand.
- Credentials are also validated eagerly on `loadGitRepository` (after `GitVCS.init()`) to surface revoked tokens as HTTP 4xx errors, which triggers the existing re-auth banner logic.

## Test plan

- [ ] Connect a project to a GitHub/GitLab repo with a valid token — confirm no error is shown.
- [ ] Revoke the token (or uninstall the GitHub App) and reload the project — confirm the re-auth banner appears.
- [ ] Verify that a project linked to a public GitHub repo with a revoked token still shows the re-auth hint (regression case for the wire-protocol false-positive).
- [ ] Confirm the fallback path works for non-GitHub/GitLab providers (e.g. manual PAT credentials).

image

image

image

### DIFF
--- a/packages/insomnia/src/ui/components/project/project-settings-form.tsx
+++ b/packages/insomnia/src/ui/components/project/project-settings-form.tsx
@@ -205,7 +205,7 @@ export const ProjectSettingsForm: FC<Props> = ({
     }
   }, [showGitConnectionInfo, gitRepository?.uri, gitRepository?._id, project?._id, validateCredentialsFetcher]);
 
-  const gitRepoLoadErrors =
+  const credentialsValidationErrors =
     validateCredentialsFetcher.data && 'errors' in validateCredentialsFetcher.data
       ? validateCredentialsFetcher.data.errors
       : undefined;
@@ -316,10 +316,10 @@ export const ProjectSettingsForm: FC<Props> = ({
               <GitOauthAuthBanner
                 selectedCredential={selectedCredential}
                 gitRepository={gitRepository}
-                repoLoadErrors={gitRepoLoadErrors}
+                repoLoadErrors={credentialsValidationErrors}
                 provider={selectedProvider}
               />
-              {showEmailSelector && !gitRepoLoadErrors ? (
+              {showEmailSelector && !credentialsValidationErrors ? (
                 <div className="flex flex-col gap-2">
                   {isLoadingEmails ? (
                     <div className="flex items-center gap-2 text-sm">
@@ -397,7 +397,7 @@ export const ProjectSettingsForm: FC<Props> = ({
                     </div>
                   )}
                 </div>
-              ) : selectedCredential?.author.email && !gitRepoLoadErrors ? (
+              ) : selectedCredential?.author.email && !credentialsValidationErrors ? (
                 <div className="text-[12px]">
                   <div className="flex">
                     <div className="w-[110px] font-semibold">Author Email</div>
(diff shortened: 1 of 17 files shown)


packages/insomnia/src/ui/components/project/project-settings-form.tsx
```diff
@@ -203,15 +202,17 @@ export const ProjectSettingsForm: FC<Props> = ({
...
+      validateCredentialsFetcher.load({ projectId: project._id });
     }
-  }, [showGitConnectionInfo, gitRepository?.uri, gitRepository?._id, project?._id, gitRepoDataFetcher]);
+  }, [showGitConnectionInfo, gitRepository?.uri, gitRepository?._id, project?._id, validateCredentialsFetcher]);
 
   const gitRepoLoadErrors =
-    gitRepoDataFetcher.data && 'errors' in gitRepoDataFetcher.data ? gitRepoDataFetcher.data.errors : undefined;
+    validateCredentialsFetcher.data && 'errors' in validateCredentialsFetcher.data
+      ? validateCredentialsFetcher.data.errors
+      : undefined;
```
CurryYangxx, line 213: We could use a new name to replace `gitRepoLoadErrors` since we don't use gitRepoLoader here

packages/insomnia/src/ui/components/project/project-settings-form.tsx
```diff
@@ -203,15 +202,17 @@ export const ProjectSettingsForm: FC<Props> = ({
...
-      gitRepoDataFetcher.state === 'idle' &&
-      !gitRepoDataFetcher.data
+      validateCredentialsFetcher.state === 'idle' &&
+      !validateCredentialsFetcher.data
     ) {
-      gitRepoDataFetcher.load({ projectId: project._id });
+      validateCredentialsFetcher.load({ projectId: project._id });
     }
-  }, [showGitConnectionInfo, gitRepository?.uri, gitRepository?._id, project?._id, gitRepoDataFetcher]);
+  }, [showGitConnectionInfo, gitRepository?.uri, gitRepository?._id, project?._id, validateCredentialsFetcher]);
```
CurryYangxx: There will be a potential bug here, when the project id change,  `fetcher.load` will not run again because validateCredentialsFetcher.data is not empty, so that we always get the stale data from fetcher.
For the scenario in the pr, the form component will be destroyed every time the project id change, and it will create a new fetcher, so everythink is ok. But if the component is always active, the bug will appear.

I would suggest we use fetcher.loader like this:
```
const fetcher = useFetcher();
const load = fetcher.load;
useEffect(() => {
  if (otherDenpendency) {
    load({ id });
  }
}, [load, id, otherDependency])

```

This will not block this pr, I just wanted to raise this question for us to discuss

mcturco:
- Let's update the toast content to the following:
  - Push failed
  Connection has expired. Re-authenticate (underlined)
- Just noticed that I have misspelled "Re-authenticate" in my designs, sorry about that! Can we update that everywhere?
- For all links in the red banners, can we change the color from purple to white?
- In the red banners, can we reduce the space between the icon and the content by half?

pavkout:
> * Let's update the toast content to the following:
>   
>   * Push failed
>     Connection has expired. Re-authenticate (underlined)
> * Just noticed that I have misspelled "Re-authenticate" in my designs, sorry about that! Can we update that everywhere?
> * For all links in the red banners, can we change the color from purple to white?
> * In the red banners, can we reduce the space between the icon and the content by half?

image

image

@mcturco I have completed the suggested changes 🚀

gatzjames (approved):

PR #9785, opened 2026-04-06
feat: integrate v3 user endpoints

Ali-Sab:
Migrate user API calls from v1 to v3 SDK

- Introduces `@kong/insomnia-api-sdk` to insomnia-api and migrates user profile and encryption key fetching from v1 endpoints to `GET /v3/users/me` and `GET /v3/users/me/encryption-keys`. 
- `absorbKey` now makes parallel v3 calls instead of a single `whoami`. 
- Smoke test mocks updated accordingly.
- `GET /auth/whoami` is retained for the SRP login handshake.


packages/insomnia-api/src/user.ts
```diff
@@ -9,66 +10,48 @@ export const logout = ({ sessionId }: { sessionId: string }) => {
...
 }
 
-export const getUserProfile = async ({ sessionId }: { sessionId: string }) => {
-  return fetch<UserProfile>({
-    method: 'GET',
-    path: '/v1/user/profile',
-    sessionId,
-  });
+export const getUserProfile = async ({ sessionId }: { sessionId: string }): Promise<UserProfile> => {
+  const client = createV3Client(sessionId);
```
ZxBing0066: It seems to be risky to directly use the client from the SDK. Please use insomniaFetch as the fetch client, since:
- We need to customize some headers for tracking
- Some error catching related code there
- Deeplink support
- Native fetch may be necessary for proxy proposal

ZxBing0066 (dismissed):

packages/insomnia-api/src/user.ts
```diff
@@ -9,66 +10,48 @@ export const logout = ({ sessionId }: { sessionId: string }) => {
...
-  }
-  if (response && !response?.encSymmetricKey) {
-    throw new Error('Unexpected response: ' + JSON.stringify(response));
-  }
-  return response;
-};
-
-// GET /v1/user/profile
+// GET /v3/users/me
 export interface UserProfile {
```
ZxBing0066: Could we directly use the User type from the SDK?

Ali-Sab: Yes, but it would require some refactoring, across like 5 files? Let me give it a shot, but I think it's worthwhile since we want to migrate to v3 anyway.

packages/insomnia-api/src/user.ts
Ali-Sab: Hmm, that makes sense, those are important for sure. The only thing is, the generated fetch-based SDK uses fetchApi instead of fetch (which is a custom take that the SDK is built around using for all calls). 
I think we have two paths: 
1. I think I can make a version of insomnia-fetch heavily based on the existing but it uses fetchApi so I can plug that in. 
2. Or we can drop using SDK's builtin fetching and response parsing and just use the insomnia-fetch we have right now and keep the SDK types like User for example. 

With option 1, maybe we have some repeated code, with option 2, we miss the SDK's builtin handlers, but we can still keep types, and our custom fetch (insomnia-fetch)
Any preference here?

Ali-Sab: I'll go with option 2 because it feels simpler to just use the insomnia-fetch and skip the SDK's builtin handler, and I'll keep the generated `User` and `UserEncryptionKeys` types

Let me know if that's not preferred

packages/insomnia/types/json-bigint.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'json-bigint';
```
ZxBing0066: Is there any reason that we don't use @types/json-bigint?

packages/insomnia-smoke-test/server/insomnia-api.ts
```diff
@@ -78,47 +78,40 @@ let organizationFeatures = {
...
+  emails: [],
+  encryption_enabled: true,
+  is_externally_provisioned: false,
 };
 
-const whoami = {
-  sessionExpiry: 4_838_400,
-  publicKey: {
+const v3EncryptionKeys = {
+  public_key: JSON.stringify({
```
ZxBing0066, line 96: I'm curious why we use a JSON string instead of keeping the previous object. I think it's harder for parsing and checking.

packages/insomnia-api/src/user.ts
ZxBing0066: I'm fine with it. We can migrate to use the SDK once it supports customizing the fetch client.

ZxBing0066:
Considering the SDK module is currently private, it might be worth re-evaluating this approach. Keeping it private prevents other open-source contributors from installing and debugging effectively.

packages/insomnia-smoke-test/server/insomnia-api.ts
Ali-Sab: That is because the SDK's type also stores them as strings, so in the smoke test I just generate the JSON, and keep it as string to match the type correctly. Here is a snippet from the SDK (public)

```
export interface UserEncryptionKeys {
    /**
     * The user's encrypted driver key
     * @type {string}
     * @memberof UserEncryptionKeys
     */
    readonly enc_driver_key: string;
    /**
     * The user's encrypted private key
     * @type {string}
     * @memberof UserEncryptionKeys
     */
    readonly enc_private_key: string;
    /**
     * The user's encrypted symmetric key
     * @type {string}
     * @memberof UserEncryptionKeys
     */
    readonly enc_symmetric_key: string;
    /**
     * The user's public key
     * @type {string}
     * @memberof UserEncryptionKeys
     */
    readonly public_key: string;
    /**
     * Salt used for encryption key derivation
     * @type {string}
     * @memberof UserEncryptionKeys
     */
    readonly salt_enc: string;
}
```

packages/insomnia/types/json-bigint.d.ts
Ali-Sab: If you mean why does this file exist and is not used... actually, this file is a mistake, seems I did some compilation out of order and the file wasn't generated by @types for some reason 🤔 , so I will delete this file as it's not needed

Ali-Sab:
> Considering the SDK module is currently private, it might be worth re-evaluating this approach. Keeping it private prevents other open-source contributors from installing and debugging effectively.

So I think they should still be able to call the endpoints, since we are only actually using the generated types. And the dev can output a response to see the data inside (treated as `any`), but you are right in that an open-source contributor cannot actually see the types themselves within the code and modify it. However, since the variable is treated as `any`, a dev can technically work with the output of the `getUserProfile` and `getEncryptionKeys` functions if they want to. 

Given that, is it sufficient? Do we still think that making the SDK fully public would be better?

packages/insomnia-smoke-test/server/insomnia-api.ts
ZxBing0066: I mean, why the structure of the API has changed to this?

ZxBing0066:
> that an open-source contributor cannot actually see the types themselves within the code and modify it.

They will fail to install the packages and cannot dev and debug insomnia anymore.

Ali-Sab:
We will set the SDK to be a public npm package, to alleviate this issue. Will make a rev after I've got that setup

ZxBing0066:
> We will set the SDK to be a public npm package, to alleviate this issue. Will make a rev after I've got that setup

I'm fine with it. Please ensure it's safe to make the SDK public.

packages/insomnia-smoke-test/server/insomnia-api.ts
Ali-Sab: Oh, I thought it was like this before too. Maybe @shelby-moore would have a better idea actually

shelby-moore: They were also previously returned as strings, that hasn't changed.

packages/insomnia-api/src/user.ts
```diff
@@ -9,66 +13,17 @@ export const logout = ({ sessionId }: { sessionId: string }) => {
...
-    throw new TypeError('Unexpected plaintext response: ' + response);
-  }
-  if (response && !response?.encSymmetricKey) {
-    throw new Error('Unexpected response: ' + JSON.stringify(response));
-  }
-  return response;
+// GET /v3/users/me
+export const getUserProfile = async ({ sessionId }: { sessionId: string }): Promise<User> => {
+  const user = await fetch<User>({ method: 'GET', path: '/v3/users/me', sessionId });
+  console.debug('User profile', user);
```
ZxBing0066: Please clean the logs before merging the PR.

packages/insomnia-api/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    environment: 'node',
+  },
+});
```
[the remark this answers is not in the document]
ZxBing0066: @Ali-Sab please take a look at this.

packages/insomnia/src/account/__tests__/session.test.ts
```diff
@@ -0,0 +1,232 @@
...
+
+    const privateKey = await getPrivateKey();
+
+    expect(crypt.decryptAES).toHaveBeenCalledWith(MOCK_SYMMETRIC_KEY, MOCK_ENC_PRIVATE_KEY);
+    expect(privateKey).toEqual(mockPrivateKey);
+  });
+
+  it('throws when session is empty', async () => {
+    // Ensure no session exists by checking with blank id
+    const session = await getUserSession();
```
ZxBing0066: If this test is intend to test the `getUserSession` itself, we should reset the store data by `setSessionData`. If this test is for `getPrivateKey`. I would suggest to merge it into the above test.

packages/insomnia-smoke-test/server/insomnia-api.ts
ZxBing0066: Got it, thank you. It looks like the previous mock data is incorrect and is never used.

packages/insomnia-api/src/user.ts
Ali-Sab: Yes, I will do that for all changes, good call out.

packages/insomnia-api/vitest.config.ts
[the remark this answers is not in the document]
Ali-Sab: Adding in next revision

ZxBing0066 (dismissed):
🚀

Ali-Sab:
Had to rebase -> merging

PR #9790, opened 2026-04-08
Feat/git repo output sync queue

gatzjames:
# Overview

Improves on the existing design of the repo-watcher with guarantees for data consistency and no race conditions when reading/writing to/from the db/fs.

## Consistency

In order to achieve these guarantees we introduce a serial queue (not direction lock):

All sync work — both FS→DB and DB→FS — goes through one FIFO queue.
Tasks never run concurrently, eliminating all race conditions.
Throughput isn't a concern since sync events are infrequent.

## No infinite loops

To ensure there are no write -> read -> write loops we use Content-hash dedup (not timing-based suppression):

After writing a YAML file to disk, record its SHA-256 hash.
When the watcher sees the file change, compare hashes — if it matches, skip (it's our own write).
This is writer-agnostic and handles external tools, git ops, and timing edge cases.

mtime as fast-path: Before computing content hash, we check the mtime first (cheap stat call). Only read + hash the file if mtime changed.

## Other

- Removes the `project-nedb-client` and completely uses the new sync mechanism for all git operations which completely decouples the `git` functionality from the app data storage.
- The UI now refreshes when data on the repository (fs) change
- Added an open folder button and copy repo path in the Project Settings view


packages/insomnia/src/main/git-service.ts
```diff
@@ -223,18 +222,17 @@ async function getGitFSClient({
...
   // All git metadata in the GIT_INTERNAL_DIR directory is stored in a .git/ directory on the filesystem
   const gitDataClient = fsClient(baseDir);
 
-  // All non-YAML files are stored at the repository root (no separate 'other' subfolder)
-  // so that native Git tools can operate directly on the repository directory.
-  const otherDataClient = fsClient(baseDir);
+  // All files (YAML + non-YAML) are stored at the repository root so that
+  // native Git tools can operate directly on the repository directory.
+  // The RepoFileWatcher is solely responsible for syncing YAML ↔ NeDB.
+  const diskClient = fsClient(baseDir);
```
CurryYangxx, line 231: These 2 clients are the same. Could we merge them?

packages/insomnia/src/main/git-service.ts
```diff
@@ -264,7 +262,7 @@ export async function loadGitRepository({ projectId, workspaceId }: { projectId:
       if (!workspaceId) {
         legacyInsomniaWorkspace = await containsLegacyInsomniaDir({ fsClient });
         // Ensure watcher is running (idempotent)
-        startWatcher(gitRepository._id, baseDir, projectId);
+        await repoFileWatcherRegistry.startWatcher(gitRepository._id, baseDir, projectId);
```
CurryYangxx, line 265: We don't import all files when the repo exists. If the user made some changes in the external editor when the Insomnia app closed, then when the user opens the app, they will not see the latest content in the UI.
Also, if the user does some git action and triggers the flushNow function, the old DB data will overwrite the latest fs content.

packages/insomnia/src/sync/git/repo-file-watcher.ts
```diff
@@ -90,164 +155,150 @@ class RepoFileWatcher {
...
+      return;
+    }
+
+    // Cancel any pending debounced flush — we're doing it immediately
+    if (this.flushDebounce) {
+      clearTimeout(this.flushDebounce);
+      this.flushDebounce = null;
+    }
+
+    this.queue.enqueue(() => this.flushProjectWorkspacesToDisk());
```
CurryYangxx, line 204: I think there is still a narrow race here between inbound FS→DB imports and flushNow().

The design says all sync work goes through a single serial queue, but scheduleImport() debounces for 300ms before the import is actually enqueued. That means an external YAML edit can exist in the repo but still be "outside" the queue for a short window. If the user then triggers a git action that calls flushNow() during that window, the DB→FS flush can run first and overwrite the external edit before it is ever imported.

This is probably a low-frequency case if fs.watch fires promptly, but it still means the "no race conditions" / "all sync work is serialised" guarantee is not fully true today. It also becomes less narrow if fs.watch is delayed (in Linux we fallback to polling) or missed and we fall back to polling.

pavkout:
After running the project locally, I noticed that all these changes are appearing as part of my modifications.

image

image

Additionally, it is somewhat unclear why these changes are present.

image

gatzjames:
> After running the project locally, I noticed that all these changes are appearing as part of my modifications.

> Additionally, it is somewhat unclear why these changes are present.

@pavkout I think this is related to the migration not running before the diff gets called.

I can verify it happens for sure if you migrate the repo, switch back to a previous version and come back again. Since the repoId is marked as migrated in electron storage it doesn't re-run the migration so you see these files as changes.

We should follow up on the migration work:
- Ensure the migration has run before any git work happens
- We can store the migration result in another place (database?) so that if the user goes back to a previous version the migration needs to run again. This will depend on how we decide to run the migrations eventually.

packages/insomnia/src/sync/git/repo-file-watcher.ts
```diff
@@ -90,164 +155,150 @@ class RepoFileWatcher {
...
-
-          // Update the tracked mtime so the polling loop doesn't re-import the file
-          const stat = await fs.promises.stat(absPath);
-          this.lastSyncMtime.set(absPath, stat.mtimeMs);
-        } catch (err) {
-          console.warn('[repo-file-watcher] Could not flush workspace to disk:', workspace._id, err);
-        } finally {
-          // Unsuppress after a short delay to let the fs.watch event fire and be dropped
-          setTimeout(() => this.suppressedPaths.delete(absPath), DEBOUNCE_MS * 2);
+          await fs.promises.unlink(previousAbsPath);
```
CurryYangxx: I suggest moving this unlink action after export success.

packages/insomnia/src/main/git-service.ts
gatzjames: Good catch. I updated the initialization logic so that it doesn't require us to manually do the import before the watcher.

CurryYangxx (approved):

CurryYangxx (approved):

PR #9796, opened 2026-04-10
feat: move model migrate and initModel to database

ZxBing0066:
## Changes

- Move initModel to a new file since it should only be used in the database
- Unify the imports of service files
- Move some of the legacy model APIs to insomnia-data


packages/insomnia/src/sync/vcs/pull-backend-project.ts
```diff
@@ -28,15 +27,18 @@ export const pullBackendProject = async ({ vcs, backendProject, remoteProject }:
...
-    const workspace = await models.initModel<Workspace>(models.workspace.type, {
+    const workspace = await database.update<Workspace>({
+      ...models.workspace.init(),
       _id: backendProject.rootDocumentId,
       name: backendProject.name,
       parentId: remoteProject._id,
       scope: 'collection',
+      modified: Date.now(),
+      created: Date.now(),
+      isPrivate: false,
```
CurryYangxx, line 38: Is this field useful here?

packages/insomnia/src/sync/vcs/pull-backend-project.ts
```diff
@@ -28,15 +27,18 @@ export const pullBackendProject = async ({ vcs, backendProject, remoteProject }:
   // @TODO Revisit the UX for this. What should happen if there are other branches?
   // The default branch does not exist, so we create it and the workspace locally
   if (defaultBranchMissing) {
-    const workspace = await models.initModel<Workspace>(models.workspace.type, {
+    const workspace = await database.update<Workspace>({
+      ...models.workspace.init(),
```
CurryYangxx, line 31: I suggest we don't call `model.init()` directly outside the initModel function, because we need to add these model-unrelated fields like `modified/created` manually. It's easy to miss some of these fields.

packages/insomnia/src/network/o-auth-2/get-token.ts
```diff
@@ -436,27 +434,26 @@ const sendAccessTokenRequest = async (
...
+      mimeType: 'application/x-www-form-urlencoded',
+      params,
     },
-  );
+    _id: requestOrGroupId + '.other',
+    parentId: requestOrGroupId,
+    type: models.request.type,
+    modified: Date.now(),
+    created: Date.now(),
+  };
```
CurryYangxx, line 456: Same here

packages/insomnia/src/sync/vcs/pull-backend-project.ts
ZxBing0066: Yes, it's using the update API, and these fields are necessary. I think the API is not reasonable, but I don't want to make any changes to the business code.

CurryYangxx (approved):

PR #9802, opened 2026-04-13
feat: migrate model imports, base types, org model and helpers

ZxBing0066:
[INS-2204](https://konghq.atlassian.net/browse/INS-2204)

## Changes
- Move base types to insomnia-data
- Unify the imports of insomnia-data/models
- Migrate organization model
- Move model-schemas to sync since it's only used in the sync code
- Move replace-ids-in-fields to insomnia-data
- Move helpers to insomnia-data
- Change legacy model imports

[INS-2204]: https://konghq.atlassian.net/browse/INS-2204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jackkav (approved):

jackkav:
🚀

jackkav (dismissed):

packages/insomnia/src/plugins/context/response.ts
```diff
@@ -49,11 +50,31 @@ export function init(response?: MaybeResponse) {
       },
 
       getBody() {
-        return getBodyBuffer(response);
+        return services.helpers.getResponseBodyBuffer(response);
       },
 
       getBodyStream() {
-        return getBodyStream(response);
+        // To avoid break the plugin APIs, keep this API as synchronous and move the implementation here.
```
CurryYangxx, line 57: The getBody function above also returns a promise. Will there be the same problem?

packages/insomnia/src/main/ipc/main.ts
```diff
@@ -263,7 +267,13 @@ export function registerMainHandlers() {
...
       throw new TypeError(`Unknown service method: ${serviceName}.${methodName}`);
     }
-    return (fn as (...args: unknown[]) => unknown).call(service, ...args);
+    const result = await (fn as (...args: unknown[]) => unknown).call(service, ...args);
+    // Tag Buffer results before contextBridge serializes them as plain Uint8Array,
+    // so the preload can distinguish them from intentional Uint8Array returns.
+    if (Buffer.isBuffer(result)) {
+      return { __type: 'Buffer', data: Array.from(result as Buffer) };
+    }
+    return result;
```
CurryYangxx, line 276: Which case do we need to distinguish the data type?

ZxBing0066: It is only used for getResponseBodyBuffer as it returns a Buffer which will be serialized by IPC.

packages/insomnia/src/plugins/context/response.ts
ZxBing0066: No, since getBody keeps async, but getBodyStream is changed to async from sync

CurryYangxx (approved):

PR #9804, opened 2026-04-13
test: add core workflow smoke tests covering request CRUD lifecycle

starry0811:
Add Playwright E2E smoke tests that cover Insomnia primary end-to-end workflows: GET with env-var URL substitution, POST/201 creation, full CRUD chain using an after-response script to capture and reuse an ID, custom header propagation verified via echo endpoint, basic auth, manual request creation via the UI URL bar, and environment switching.


starry0811:
I have read the CLA Document and I hereby sign the CLA

PR #9807, opened 2026-04-13
fix: large responses now download correctly [INS-2365]

arora-r:
Users that are requesting endpoints with large responses (>= 5MB) are shown a "Save To File" button in the response pane. Clicking this saves an empty file. Similarly, in the Preview button dropdown the options to export as raw response, prettified response, and http debug were also saving blank files.

The "Export as HAR" option uses the stored response body instead of the current `activeResponse.bodyBuffer` making it work as expected regardless of the response size.

[Addresses INS-2365]


packages/insomnia/src/ui/components/panes/response-pane-utils.ts
```diff
@@ -1,10 +1,21 @@
...
 
 export async function downloadResponseBody(
   activeRequest: { name: string } | null | undefined,
-  activeResponse: { contentType: string; bodyBuffer?: Buffer | null } | null | undefined,
+  activeResponse:
+    | {
+        contentType: string;
+        bodyBuffer?: Buffer | null;
+        bodyPath?: string;
+        bodyCompression?: Compression;
```
arora-r: Note: Adding these to show where the file is stored locally, and if we need to decompress because we now can use `getBodyBuffer(activeResponse)` which is the file in disk instead of the data in memory

packages/insomnia-smoke-test/server/index.ts
```diff
@@ -39,6 +39,15 @@ app.get('/pets/:id', (req, res) => {
...
 });
 
+app.get('/large-json', (_req, res) => {
+  const items = Array.from({ length: 100_000 }, (_, i) => ({
+    id: i,
+    name: `item-${i}`,
+    value: 'x'.repeat(100),
+  }));
+  res.status(200).json({ items });
+});
```
arora-r, line 64: Generates ~13MB response size

packages/insomnia-smoke-test/fixtures/response-download-collection.yaml
```diff
@@ -0,0 +1,50 @@
+type: collection.insomnia.rest/5.0
```
arora-r, line 1: Created the collection in the insomnia app and then exported them for this file

packages/insomnia/src/ui/components/panes/response-pane-utils.ts
```diff
@@ -24,12 +33,41 @@ export async function downloadResponseBody(
   if (canceled) {
     return;
   }
+
+  // The loader only populates bodyBuffer for responses under LARGE_RESPONSE_MB, so reading it
+  // here would write an empty file for exactly the large responses this export needs to handle.
+  // Copy from disk in the main process instead, which also keeps the body out of renderer memory.
+  if (!prettify && activeResponse.bodyPath) {
+    await window.main.writeResponseBodyToFile({
```
fiosman, line 41: do we want to surface any error toast / modal to the user if this function call fails? Currently it will bubble up to the IPC layer where is a catch and a raw error is thrown but the user doesn't get any UI feedback

packages/insomnia/src/ui/components/dropdowns/preview-mode-dropdown.tsx
```diff
@@ -119,11 +125,25 @@ export const PreviewModeDropdown: FC<Props> = ({ download, copyToClipboard }) =>
...
         <DropdownItem aria-label="Export prettified response">
           {shouldPrettifyOption && (
-            <ItemContent icon="save" label="Export prettified response" onClick={handleDownloadPrettify} />
+            <ItemContent
+              icon="save"
+              label={
+                isLargeResponse ? `Export prettified response (must be <${LARGE_RESPONSE_MB}MB)` : 'Export prettified response'
+              }
+              isDisabled={isLargeResponse}
+              className={isLargeResponse ? 'opacity-50' : ''}
```
fiosman, line 147: why is this className necessary? If `isDisabled` prop is truthy shouldn't the component itself handle the styling?

packages/insomnia-smoke-test/tests/smoke/response-downloads.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    const buffer = Buffer.alloc(length);
+    const bytesRead = fs.readSync(fd, buffer, 0, length, 0);
+    return buffer.subarray(0, bytesRead).toString('utf8');
+  } finally {
+    fs.closeSync(fd);
+  }
+}
+
+test.describe('Response Downloads', () => {
+  test.slow(process.platform === 'darwin' || process.platform === 'win32', 'Slow app start on these platforms');
```
fiosman, line 28: i am curious why we need this `test.slow()`, is this based on CI runs to help with test flakiness on diff os?

packages/insomnia/src/ui/components/panes/response-pane-utils.ts
arora-r: Good call added this now

packages/insomnia/src/ui/components/dropdowns/preview-mode-dropdown.tsx
arora-r: the isDisabled doesnt add styling right now in the component. I didn't want to bake it into the `ItemContent` component incase other places where its used have other ideas

packages/insomnia-smoke-test/tests/smoke/response-downloads.test.ts
arora-r: Yea its just for consistency. Other tests already do this to ensure tests ran locally are less flaky

fiosman (approved):

PR #9809, opened 2026-04-14
refactor: import to main

jackkav:
Plan: https://github.com/Kong/insomnia/pull/9803
blocked by: #9827 
Removes 9 renderer node imports

## Purpose
- Stop renderer-reachable code from importing `src/main` implementations when only pure helpers, constants, or types are needed.
- Extract shared logic out of main-only modules while leaving privileged behavior in the main process.

## Files in scope
- `packages/insomnia/src/main/importers/importers/openapi-3.ts`
- `packages/insomnia/src/main/network/parse-header-strings.ts`
- `packages/insomnia/src/main/network/multipart.ts`
- `packages/insomnia/src/network/parse-header-strings.ts`
- `packages/insomnia/src/network/multipart-constants.ts`
- `packages/insomnia/src/common/import.ts`
- `packages/insomnia/src/common/path-with-params.ts`
- `packages/insomnia/src/common/misc.ts`
- `packages/insomnia/src/common/significant-diff-detection.ts`
- `packages/insomnia/src/utils/url/querystring.ts`
- `packages/insomnia/src/ui/components/modals/proto-files-modal.tsx`
- `packages/insomnia/config/renderer-node-import-baseline.json`

## Baseline entries expected to be removed
- `src/main/importers/importers/curl.ts -> url`
- `src/main/importers/importers/openapi-3.ts -> crypto`
- `src/main/importers/importers/openapi-3.ts -> url`
- `src/main/importers/importers/swagger-2.ts -> crypto`
- `src/main/network/parse-header-strings.ts -> url`
- Additional nearby baseline reductions landed where the extracted shared helpers removed renderer-reachable main imports.

## Any new preload or IPC surface added
- None. This PR is focused on module boundaries and shared helper extraction rather than new bridge surface.

## Any deliberate deferrals to later PRs
- The broader network and gRPC privileged boundary pass remains in PR 5.
- Sync and storage boundary work remains in PR 6.
- Plugin and templating boundary work remains in PR 7.

## Future Optimization Candidate:

- For large Postman workspace imports (zip files with hundreds of entries), consider profiling the `scanResources()` parsing performance in the main process. If bulk imports cause noticeable UI freeze, an Electron utility process dedicated to import parsing could isolate heavy workloads from the main thread. This would require IPC serialization cost analysis, but the overhead would likely be justified if parsing takes >1–2 seconds on large files. Defer investigation until user feedback indicates performance issues with large imports.

### DIFF
--- a/packages/insomnia/src/common/path-with-params.ts
+++ b/packages/insomnia/src/common/path-with-params.ts
@@ -1,0 +1,3 @@
+const VARIABLE_SEARCH_VALUE = /{([^}]+)}/g;
+
+export const pathWithParamsAsPathParameters = (path?: string) => path?.replace(VARIABLE_SEARCH_VALUE, ':$1') ?? '';
--- a/packages/insomnia/src/main/cloud-sync/vcs.ts
+++ b/packages/insomnia/src/main/cloud-sync/vcs.ts
@@ -37,15 +37,15 @@ const requestConflictResolution = (conflicts: MergeConflict[], labels: { ours: s
   const context = syncInvocationContext.getStore();
   invariant(context, 'Sync conflict resolution requires a renderer context');
 
-  const requestId = randomUUID();
+  const handlerId = randomUUID();
   context.sender.send('sync.merge-conflicts', {
-    requestId,
+    handlerId,
     conflicts,
     labels,
   });
 
   return new Promise<MergeConflict[]>((resolve, reject) => {
-    pendingConflictResolutions.set(requestId, {
+    pendingConflictResolutions.set(handlerId, {
       senderId: context.sender.id,
       resolve,
       reject,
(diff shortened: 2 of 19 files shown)


ZxBing0066:
IMO, the import should only run in the main process, not check the process type everywhere. `process.type === 'renderer'` is easily lead to undetectable runtime chaos

```
UI -> import -- IPC --> main.scanResource(files); main.importResourcesToXXX();
```

packages/insomnia/src/common/path-with-params.ts
```diff
@@ -0,0 +1,3 @@
+const VARIABLE_SEARCH_VALUE = /{([^}]+)}/g;
```
ZxBing0066, line 1: How about export `VARIABLE_SEARCH_VALUE` so openapi-3.ts can reused it to avoid has two duplicate variables.

packages/insomnia/src/main/ipc/main.ts
```diff
@@ -100,6 +100,17 @@ const writeResponseBodyToFile = async (
   _: unknown,
   options: { sourcePath: string; destinationPath: string; bodyCompression?: 'zip' | null },
 ) => {
+  // Validate sourcePath is within the expected responses directory to prevent a
```
ZxBing0066, line 103: Also have concerns about destinationPath, is it safe if an attacker attempts to override important system files? It should be rejected by the system, but maybe we should test.

ZxBing0066 (dismissed):

packages/insomnia/src/main/ipc/main.ts
jackkav: We have overlapping concepts, allowlisted directories, and blocklisted important system files. One allows the other blocks. Lets scope this in a security pass after the refactor. I'll add it to the ticket.

packages/insomnia/src/common/path-with-params.ts
jackkav: > Exporting a global RegExp with the 'g' flag (VARIABLE_SEARCH_VALUE) exposes a stateful object across modules; create a fresh RegExp per use or remove the 'g' flag.

oh well

packages/insomnia/src/main/cloud-sync/vcs.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+export interface PullRemoteBackendProjectOptions {
+  organizationId: string;
+  backendProjectId: string;
+  remoteId: string;
+}
+
+const syncInvocationContext = new AsyncLocalStorage<SyncInvocationContext>();
+const pendingConflictResolutions = new Map<string, PendingConflictResolution>();
```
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: uses handlerId to ensure uniqueness

ZxBing0066 (approved):

PR #9824, opened 2026-04-16
Repo Migration flow [INS-2256]

pavkout:
## Summary

- Introduce a new canonical on-disk git repository layout (`git/` → `.git/`, `other/` → repo root, YAML written to disk) so users can run native `git` CLI commands directly against Insomnia repos
- Add `git-repo-migration.ts` to migrate existing repos to the new layout at startup, idempotently and non-fatally
- Add `repo-file-watcher.ts` to detect changes made by external tools (VS Code, native git, etc.) and sync them back into the DB
- Rewire `project-routable-fs-client.ts` to dual-sync YAML: disk-first writes, disk-preferred reads, with DB as fallback

## What was fixed

- **Concurrent migration race** — in-memory guard prevents the same repo being migrated twice in parallel
- **Interrupted migration resume** — if a previous run crashed mid-copy (both `git/` and `.git/` exist), the migration now resumes rather than skipping
- **Symlinks broken during file move** — `copyFile` was silently dereferencing symlinks; now uses `readlink` + `symlink` to preserve them
- **`rmdir` failing on non-empty dirs** — replaced with `fs.promises.rm({ recursive: false })`
- **Stale `core.worktree` breaking native git** — strips the entry from `.git/config` after migration so native git resolves the worktree correctly
- **Non-atomic YAML writes** — replaced direct `writeFile` with a tmp-file + rename strategy to prevent truncated files on crash
- **Path traversal in stored `gitFilePath`** — validates resolved path stays within `baseDir` before writing
- **Empty YAML export written to disk** — skips write with a warning if export content is blank
- **Migration idempotency lost across sessions/rollbacks** — replaced `ElectronStorage` flag with a `repoMigrationVersion` field on `GitRepository`; older app versions naturally clear it via `docUpdate`, triggering a correct re-run after downgrade
- **N+1 DB queries on workspace flush** — batch-fetches all `WorkspaceMeta` records upfront instead of one query per workspace
- **DB `gitFilePath` not reconciled when file already existed** — DB update now always runs regardless of whether the file was just written

## Test plan

- [ ] Open a project with an existing repo — verify migration runs once and the old `git/` and `other/` directories are gone
- [ ] Verify `.git/` is present and `git log` works from the repo directory in a terminal
- [ ] Restart the app — verify migration does not re-run (idempotency)
- [ ] Edit a YAML file externally (e.g. via VS Code or `git checkout`) — verify changes appear in the Insomnia UI
- [ ] Rollback to an older build, open the same project, upgrade again — verify migration re-runs correctly
- [ ] Verify no data loss for workspaces after migration


packages/insomnia/src/sync/git/git-repo-migration.ts
```diff
@@ -15,34 +15,60 @@
...
+ * downgrade that recreates the old directories is always caught.
+ *
+ * Accepts a pre-fetched `gitRepo` so the caller avoids an extra DB round-trip.
+ */
+async function hasMigrated(baseDir: string, gitRepo: GitRepository | null | undefined): Promise<boolean> {
+  // Disk override: old layout directories mean migration is definitely needed.
+  // Both checks run in parallel — they're independent stat calls.
+  const [hasOldGit, hasOldOther] = await Promise.all([
+    dirExists(path.join(baseDir, 'git')),
+    dirExists(path.join(baseDir, 'other')),
```
gatzjames: We need to verify that this will not re-run the migration for a migrated repo that happens to have a folder named `other` in the root.

gatzjames (approved):

PR #9826, opened 2026-04-16
fix(UI update logic): revalidation should happen after actions are completed

gatzjames:
## Overview

Calling revalidate with active fetchers can lead to the following bug:
- The component that inits the fetcher hook is unmounted before the fetcher gets a result back
- We rely on redirect or the fetcher's state to update some UI

## Fix

Similar to how we solve this in the event stream context we can only revalidate after all actions have finished:
https://github.com/gatzjames/insomnia/blob/26e40175a5a554f3657d822efd4c40adee930c03/packages/insomnia/src/ui/context/app/insomnia-event-stream-context.tsx#L200-L202

### DIFF
--- a/packages/insomnia/src/root.tsx
+++ b/packages/insomnia/src/root.tsx
@@ -13,11 +13,13 @@ import {
   Outlet,
   Scripts,
   ScrollRestoration,
+  useFetchers,
   useNavigate,
   useParams,
   useRevalidator,
   useRouteLoaderData,
 } from 'react-router';
+import { useLatest } from 'react-use';
 
 import { EXTERNAL_VAULT_PLUGIN_NAME, isDevelopment } from '~/common/constants';
 import type { Settings, UserSession } from '~/insomnia-data';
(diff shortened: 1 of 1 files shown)


packages/insomnia/src/root.tsx
```diff
@@ -323,11 +325,17 @@ const Root = () => {
...
+
   useEffect(() => {
     return window.main.on('git.db-synced', () => {
-      revalidate();
+      if (!latestInSubmission.current) {
+        revalidate();
+      }
     });
-  }, [revalidate]);
+  }, [latestInSubmission, revalidate]);
```
CurryYangxx, line 338: In this case, the timeline is: `action start -> git operation -> trigger this event -> action  return result`.
So, do you want the revalidation to wait for all fetchers to finish, or not revalidate if there are any submissions in flight?
Seems we could only achieve the latter via this approach; the event will not trigger again after submission is finished.

gatzjames: Correct we don't want to run revalidate again!
When an action completes loaders are revalidated automatically so the changed data will be rendered correctly

CurryYangxx (approved):

PR #9827, opened 2026-04-16
refactor: move sync code to main

jackkav:
## Summary

This PR moves the cloud sync/VCS runtime behind a main-process bridge and updates the flows that need sync initialization to use that bridge instead of constructing VCS from renderer-owned code.

The goal is broader than just making import work: it is to make the non-UI sync surface main-owned while keeping renderer responsibilities limited to initiating actions and presenting UI such as merge-conflict resolution.

## What changed

- added a main-process sync bridge in `src/main/cloud-sync/*` and exposed it through preload as `window.main.sync`
- extracted `createVCS(...)` so a `VCS` instance can be constructed without renderer dependencies
- updated sync callers, workspace creation, and import flows to go through the main bridge for initialization and sync operations
- moved merge-conflict request/response plumbing to main while keeping the modal/UI in renderer
- extracted organization storage rule fetching into a main-safe shared module so main no longer depends on renderer utilities
- added tests around the main-side initialization path and injected-VCS behavior

## VCS instance lifetime / ownership

The new bridge currently uses a single lazily-created main-process `VCS` instance via `getMainVCS()`. In practice, that means the VCS instance lifespan is the Electron main-process lifespan.

That is an intentional scoping choice for this PR:

- keep ownership of sync state in main
- avoid mixing this refactor with a larger instance-registry / lifecycle-management redesign
- preserve the current assumption that these flows operate against one active sync context at a time

Related implementation details:

- pending merge-conflict resolutions are tracked in a module-scoped `Map`, keyed by request id and bound back to the originating renderer
- `AsyncLocalStorage` is used to preserve the invoking renderer context while sync work runs in main
- this PR does **not** introduce a module-scoped `Map` of `VCS` instances keyed by workspace/project/organization

## Why not a keyed VCS registry in this PR?

A per-workspace or per-project VCS registry is a reasonable follow-up if we need multiple concurrent sync contexts or stricter instance lifecycle control, but it would broaden the scope quite a bit. That design would need clear rules for instance creation, reuse, teardown, invalidation, and cross-renderer ownership.

For the flows being addressed here, a single main-owned VCS singleton is enough to complete the renderer-to-main move without also taking on the larger registry/lifespan problem.

## Out of scope

This PR does **not**:

- redesign sync around a keyed registry of `VCS` instances
- define eviction/disposal semantics for multiple long-lived VCS instances
- move all sync UI to main; renderer still owns merge-conflict presentation
- attempt a broader sync architecture rewrite beyond the main bridge and initialization paths


packages/insomnia-smoke-test/server/cloud-sync-api.ts
```diff
@@ -291,6 +291,17 @@ let deletedProjectIds: string[] = [];
 let cloudSyncApiEnabled = false;
 let remoteHasNewCommit = false;
 
+const resetCloudSyncTestState = () => {
```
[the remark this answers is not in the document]
jackkav: This is not production code, its a stub server used for testing. There is no need to consider security or best practices in this scope.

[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: files inside packages/insomnia-smoke-test/ are not production code. They are tests and stub servers.

packages/insomnia/src/sync/vcs/__tests__/insomnia-sync.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    global.window = {
+      main: {
+        sync: {
+          on,
+          resolveConflict: vi.fn(),
+          cancelConflict: vi.fn(),
+        },
+      },
+    } as Window & typeof globalThis;
```
ZxBing0066, line 29: Nit: this conversion doesn't work. It passes because we disable ts check for the test files.

packages/insomnia/src/entry.preload.ts
```diff
@@ -107,6 +113,73 @@ const secretStorage: secretStorageBridgeAPI = {
...
+  await refreshActiveBackendProject();
+  return result;
+};
+
+const sync: SyncBridgeAPI = {
+  archiveProject: () => invokeSyncMethodAndRefresh('archiveProject'),
+  checkout: (...args) => invokeSyncMethod('checkout', ...args),
+  compareRemoteBranch: () => invokeSyncMethod('compareRemoteBranch'),
+  fork: (...args) => invokeSyncMethod('fork', ...args),
+  getActiveBackendProject: () => activeBackendProject,
```
ZxBing0066: This uses the cache but may be missing on the first call or mismatch sometimes. Will it cause regressions?

packages/insomnia/src/main/cloud-sync/vcs.ts
```diff
@@ -0,0 +1,159 @@
...
+  invariant(
+    pendingConflictResolution.senderId === sender.id,
+    `Sync conflict request ${requestId} was cancelled by an unexpected renderer`,
+  );
+
+  pendingConflictResolutions.delete(requestId);
+  pendingConflictResolution.reject(new UserAbortResolveMergeConflictError());
+};
+
+export const pullRemoteBackendProjectWithSingleton = async (
```
ZxBing0066, line 116: This reintroduces the cross-workspace interference the previous code explicitly guarded against.
The new implementation skips that and mutates the shared main-process singleton directly. Between `removeBackendProjectsForRoot` and the final restore, every concurrent `sync.invoke` call (status, stage, push...) operates against the same singleton whose `_backendProject` has already been switched.

jackkav: You identify a possible race condition if an existing workspace sync's while another is attempting to be cloned.

jackkav: Where it actually breaks — the conflict resolution handler. If a merge/pull triggers a conflict dialog and the user has already switched workspaces, the dialog would appear in the wrong context (or not at all if the renderer has moved on). That's the one case "abandon"
   could be confusing rather than silently fine.

  Realistic frequency — sync operations are user-initiated and short. Two workspaces racing is an edge case. The current switchAndCreateBackendProjectIfNotExist already implicitly abandons by just overwriting _backendProject.

  So doing nothing is defensible. The status quo already silently abandons — you'd just be acknowledging that as the intended behavior rather than fixing it.

ZxBing0066: @jackkav I don't think it's an edge case; it's easy to reproduce when a user tries to load a new remote workspace and then switches to another workspace. It makes the data chaotic. Blobs will be pulled into the wrong workspaces.

Any concern about not keeping the previous logic here?

jackkav: You're right, and this is a real bug — not an edge case. The original code explicitly guarded against this with vcs.newInstance(), which shallow-cloned the VCS so the singleton's _backendProject was never mutated during a pull. That method was dropped in this refactor 
  without a replacement, and the try/finally save-and-restore doesn't hold up when async calls interleave.                                                                                                                                                                     
                                                                                                                                                                                                                                                                               
  The fix is to use createVCS() (already extracted in this PR) to spin up an isolated VCS instance for the pull, using the same data path. The singleton is only used for the remoteBackendProjects network call (read-only), and pullBackendProject gets the isolated         
  instance. This is functionally equivalent to the old newInstance() pattern — the same files on disk, independent _backendProject state.                                                                                                                                      
                                                                                                                                                                                                                                                                               
  I've pushed that change. Let me know if you'd prefer the remoteBackendProjects call also moves to the isolated instance.

ZxBing0066 (dismissed):

gatzjames (approved):

PR #9828, opened 2026-04-16
chore: isolate gRPC proto file preparation behind IPC boundary

jackkav:
## Purpose
Isolate gRPC proto temp-file creation behind the Electron preload boundary, removing `fs`, `os`, and `path` ownership from the renderer-reachable code path as part of the `nodeIntegration: false` migration.

## Single feature scope
Preparing proto directories and temp proto files for gRPC request execution.

## Implementation notes
- Moves proto temp-file creation behind a main-process IPC boundary.
- Keeps request assembly in the renderer.
- Avoids mixing in generic request execution, OAuth, or response archival work.

## Files in scope
- `packages/insomnia/src/network/grpc/write-proto-file.ts`
- `packages/insomnia/src/main/ipc/grpc.ts`
- `packages/insomnia/src/main/ipc/electron.ts`
- `packages/insomnia/src/entry.preload.ts`
- `packages/insomnia/src/ui/components/modals/proto-files-modal.tsx`
- `packages/insomnia/src/main/ipc/__tests__/grpc.test.ts`
- `packages/insomnia/config/renderer-node-import-baseline.json`

## Baseline entries removed
- `src/network/grpc/proto-directory-loader.tsx -> fs`
- `src/network/grpc/proto-directory-loader.tsx -> path`
- `src/network/grpc/write-proto-file.ts -> fs`
- `src/network/grpc/write-proto-file.ts -> os`
- `src/network/grpc/write-proto-file.ts -> path`

## Test automation plan
- Keeps existing focused coverage for proto file writing.
- Adds IPC contract coverage for the new `grpc.writeProtoFile` bridge.
- Leaves broader gRPC execution behavior unchanged.

## New preload or IPC surface added
- `grpc.writeProtoFile(protoFileId: string)` exposed on `window.main.grpc`

## Deliberate deferrals
- OAuth token helpers
- Generic network file IO
- Response archival helpers

### DIFF
--- a/packages/insomnia/src/main/ipc/grpc.ts
+++ b/packages/insomnia/src/main/ipc/grpc.ts
@@ -64,10 +64,23 @@ export interface gRPCBridgeAPI {
   writeProtoFile: (protoFileId: string) => Promise<{ filePath: string; dirs: string[] }>;
 }
 
+const grpcOptions = {
+  keepCase: true,
+  longs: String,
+  enums: String,
+  defaults: true,
+  oneofs: true,
+};
+
 export const writeProtoFileById = async (protoFileId: string): Promise<{ filePath: string; dirs: string[] }> => {
   const protoFile = await services.protoFile.getById(protoFileId);
   invariant(protoFile, `Proto file ${protoFileId} not found`);
-  return writeProtoFile(protoFile);
+  const result = await writeProtoFile(protoFile);
+  await protoLoader.load(result.filePath, {
+    ...grpcOptions,
+    includeDirs: result.dirs,
+  });
+  return result;
 };
 
 export function registergRPCHandlers() {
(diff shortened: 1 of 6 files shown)


packages/insomnia/src/main/ipc/grpc.ts
```diff
@@ -61,8 +61,28 @@ export interface gRPCBridgeAPI {
...
+  enums: String,
+  defaults: true,
+  oneofs: true,
+};
+
+export const writeProtoFileById = async (protoFileId: string): Promise<{ filePath: string; dirs: string[] }> => {
+  const protoFile = await services.protoFile.getById(protoFileId);
+  invariant(protoFile, `Proto file ${protoFileId} not found`);
+  const result = await writeProtoFile(protoFile);
+  await protoLoader.load(result.filePath, {
```
ZxBing0066, line 79: I'm not sure whether it same to call load from the main process and the renderer process. Is it only used for validation?

ZxBing0066 (approved):

packages/insomnia/src/main/ipc/grpc.ts
jackkav: I think so yes, should be the same.

PR #9829, opened 2026-04-16
chore: remove node:crypto from renderer-reachable OAuth code

jackkav:
## Purpose
Remove isolated OAuth crypto usage from renderer-reachable code without pulling in unrelated network or storage work.

## Single feature scope
OAuth token helper crypto operations.

## Implementation notes
- Moves OAuth 1 signing behind a narrow main-process IPC bridge.
- Replaces renderer `node:crypto` PKCE and relay-URL crypto usage with Web Crypto where appropriate.
- Keeps request sending, response handling, and broader network refactors out of scope.

## Files in scope
- `packages/insomnia/src/network/o-auth-1/get-token.ts`
- `packages/insomnia/src/network/o-auth-2/get-token.ts`
- `packages/insomnia/src/network/o-auth-2/utils.ts`
- `packages/insomnia/src/main/ipc/oauth.ts`
- `packages/insomnia/src/main/ipc/electron.ts`
- `packages/insomnia/src/main/ipc/main.ts`
- `packages/insomnia/src/entry.preload.ts`
- `packages/insomnia/src/main/ipc/__tests__/oauth.test.ts`
- `packages/insomnia/src/network/__tests__/authentication.test.ts`
- `packages/insomnia/src/main/mcp/oauth-client-provider.ts`

## Baseline entries removed
- `src/network/o-auth-1/get-token.ts -> crypto`
- `src/network/o-auth-2/get-token.ts -> crypto`
- `src/network/o-auth-2/utils.ts -> crypto`

## Test automation plan
- Extends OAuth unit coverage around signing, PKCE, and token-helper behavior.
- Adds contract coverage for the new OAuth crypto bridge.
- Avoids broad smoke-test expansion.

## New preload or IPC surface added
- `oauthCrypto.getOAuth1AuthHeader(...)` exposed on `window.main.oauthCrypto`

## Deliberate deferrals
- Generic request execution
- Response archival
- Import or sync persistence

### DIFF
--- a/packages/insomnia/src/main/ipc/oauth.ts
+++ b/packages/insomnia/src/main/ipc/oauth.ts
@@ -1,0 +1,136 @@
+import crypto from 'node:crypto';
+
+import OAuth1 from 'oauth-1.0a';
+
+import type { OAuth1SignatureMethod } from '~/network/o-auth-1/constants';
+import {
+  SIGNATURE_METHOD_HMAC_SHA1,
+  SIGNATURE_METHOD_HMAC_SHA256,
+  SIGNATURE_METHOD_PLAINTEXT,
+  SIGNATURE_METHOD_RSA_SHA1,
+} from '~/network/o-auth-1/constants';
+
+import { ipcMainHandle } from './electron';
+
+function hashFunction(signatureMethod: OAuth1SignatureMethod) {
+  if (signatureMethod === SIGNATURE_METHOD_HMAC_SHA1) {
+    return function (baseString: string, key: string) {
+      return crypto.createHmac('sha1', key).update(baseString).digest('base64');
+    };
+  }
+
+  if (signatureMethod === SIGNATURE_METHOD_HMAC_SHA256) {
+    return function (baseString: string, key: string) {
+      return crypto.createHmac('sha256', key).update(baseString).digest('base64');
+    };
+  }
+
+  if (signatureMethod === SIGNATURE_METHOD_RSA_SHA1) {
+    return function (baseString: string, privatekey: string) {
+      return crypto.createSign('RSA-SHA1').update(baseString).sign(privatekey, 'base64');
+    };
+  }
+
+  if (signatureMethod === SIGNATURE_METHOD_PLAINTEXT) {
+    return function (baseString: string) {
+      return baseString;
+    };
+  }
+
+  throw new Error(`Invalid signature method ${signatureMethod}`);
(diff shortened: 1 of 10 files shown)


packages/insomnia/src/main/ipc/oauth.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  if (signatureMethod === SIGNATURE_METHOD_HMAC_SHA256) {
+    return function (baseString: string, key: string) {
+      return crypto.createHmac('sha256', key).update(baseString).digest('base64');
+    };
+  }
+
+  if (signatureMethod === SIGNATURE_METHOD_RSA_SHA1) {
+    return function (baseString: string, privatekey: string) {
+      return crypto.createSign('RSA-SHA1').update(baseString).sign(privatekey, 'base64');
```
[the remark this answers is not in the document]
jackkav: Keeping `RSA-SHA1` here is intentional. This code only uses that algorithm when the request is explicitly configured for the OAuth 1.0a `RSA-SHA1` signature method, which is part of the interoperability surface we need to preserve. Switching this path to `RSA-SHA256` would change the signature and break servers that expect the spec-defined `RSA-SHA1` behavior.

packages/insomnia/src/network/o-auth-1/get-token.ts
```diff
@@ -2,124 +2,44 @@
...
-
-    // We override getSigningKey for RSA-SHA1 because we don't want ddo/oauth-1.0a to percentEncode the token
-    oauth.getSigningKey = function (tokenSecret) {
-      return tokenSecret || '';
-    };
-  }
-
-  const data = oauth.authorize(requestData, token);
-  return oauth.toHeader(data);
+  return window.main.oauthCrypto.getOAuth1AuthHeader({
```
ZxBing0066, line 26: `inso -> getSendRequestCallbackMemDb -> sendCurlAndWriteTimeline -> getAuthHeader -> getToken`

So this will be called in inso, won't it break the send request in inso?

jackkav: Oauth flows dont work in inso because they expect a callback url, which inso doesnt know how to handle.

jackkav: rather than only extract oauth1, ive move all of auth header over in #9834 and will close this

PR #9831, opened 2026-04-16
feat: konnect integration sanitizes nunjucks templates on sync

shelby-moore:
Insomnia uses Nunjucks templating to render request URLs, headers, and other fields at send time — evaluating expressions like {{ _.proxy_host }} against the active environment. This means any string stored on a request that contains {{ }} or {% %} syntax will be executed when the user sends the request.

During a Konnect sync, route data fetched from the Konnect API is written directly into request fields (URL path, headers, request name, etc.). If a route's fields were crafted to contain Nunjucks expressions, those expressions would be silently evaluated the next time the user sends a request.

This PR introduces sanitization to strip {{ }} and {% %} syntax from all route-sourced string fields before they are stored. This happens at a single point in the sync pipeline, after routes are fetched and before any transformation or persistence. {# #} comment blocks are intentionally left alone as they produce no output when rendered.

### DIFF
--- a/packages/insomnia/src/konnect/transform.ts
+++ b/packages/insomnia/src/konnect/transform.ts
@@ -1,4 +1,54 @@
-import type { KonnectProxyUrl } from './api';
+import type { KonnectProxyUrl, KonnectRoute } from './api';
+
+// ─── Template injection sanitisation ─────────────────────────────────────────
+
+/**
+ * Strips Nunjucks template syntax (`{{ }}`, `{% %}`) from a string
+ * sourced from external API data, preventing template injection when the value
+ * is later rendered by Insomnia's Nunjucks engine.
+ */
+function stripTemplateSyntax(value: string): string {
+  let prev = '';
+  let result = value;
+  while (result !== prev) {
+    prev = result;
+    result = result
+      .replace(/\{\{[\s\S]*?\}\}/g, '')
+      .replace(/\{%[\s\S]*?%\}/g, '');
+  }
+  return result;
+}
+
+/** Strips template syntax from each item, filters empties, and returns null if nothing remains. */
+function sanitizeStringArray(arr: string[] | null): string[] | null {
+  if (arr === null) { return null; }
+  const result = arr.map(stripTemplateSyntax).filter(s => s.trim() !== '');
+  return result.length > 0 ? result : null;
+}
+
+/**
+ * Returns a copy of the route with Nunjucks template syntax stripped from all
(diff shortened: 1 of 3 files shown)


packages/insomnia/src/konnect/transform.ts
```diff
@@ -1,4 +1,48 @@
...
+
+/**
+ * Strips Nunjucks template syntax (`{{ }}`, `{% %}`) from a string
+ * sourced from external API data, preventing template injection when the value
+ * is later rendered by Insomnia's Nunjucks engine.
+ */
+function stripTemplateSyntax(value: string): string {
+  return value
+    .replace(/\{\{[\s\S]*?\}\}/g, '')
+    .replace(/\{%[\s\S]*?%\}/g, '');
```
kwburns-kong: Tested the function locally, this circumvented with `{{% %}% TEST %}`.

shelby-moore: Made it more robust 👍

packages/insomnia/src/konnect/transform.ts
```diff
@@ -1,4 +1,52 @@
...
+import type { KonnectProxyUrl, KonnectRoute } from './api';
+
+// ─── Template injection sanitisation ─────────────────────────────────────────
+
+/**
+ * Strips Nunjucks template syntax (`{{ }}`, `{% %}`) from a string
+ * sourced from external API data, preventing template injection when the value
+ * is later rendered by Insomnia's Nunjucks engine.
+ */
+function stripTemplateSyntax(value: string): string {
```
kwburns-kong, line 10: It may be possible to perform the same smuggling twice, maybe something like this?  
```ts
const pattern = /\{\{[\s\S]*?\}\}|\{%[\s\S]*?%\}|\{\{|\}\}|\{%|%\}/g;
  let prev = '';
  let result = value;
  while (result !== prev) {
    prev = result;
    result = result.replace(pattern, '');
  }
  return result;
```

shelby-moore: Went with a slightly different implementation, this wasn't cleaning up the full template var in some cases:

When the regex has both {{ (fragment) and {{...}} (pair) as alternatives, the engine tries them left-to-right at each character. So given {{% %}:

Regex matches {{ (bare fragment) at position 0 -> removes it -> % %}}
Next iteration: %} matches -> removes it -> % }
Next iteration: no matches -> loop exits
Result: % }

kwburns-kong (approved):
lgtm!

PR #9834, opened 2026-04-17
refactor: auth header to main

jackkav:
replaces this PR: https://github.com/Kong/insomnia/pull/9829

## Summary

This PR moves request-auth header generation and the OAuth token acquisition paths out of renderer-owned network code and into the Electron main process.

The core scope is making auth header resolution main-owned so request execution no longer depends on renderer importing main-only auth/token helpers.

## What changed

- moved auth-header generation into `src/main/network/get-auth-header.ts`
- moved the OAuth 2 token flow into `src/main/network/o-auth-2/get-token.ts`
- moved the remaining OAuth 1 token-fetch path into main as well
- added/updated the preload + IPC bridge so renderer/network code calls into main for auth header and token work instead of importing that logic directly
- reduced `src/network/authentication.ts` down from being the implementation home for this logic to being a thinner renderer-side caller/orchestrator
- updated request execution call sites (`network.ts`, request connect route, HAR-related wiring) to use the bridge
- updated Electron/Vite renderer-node import enforcement so these main-only imports stay blocked from renderer code
- added/updated tests and smoke fixtures to cover auth types and the moved behavior

## Additional cleanup included in this scope

These small follow-on changes came along with the move because they were directly coupled to the auth-header/token path:

- removed deprecated `baseUrl` usage
- fixed a couple of bad-input/edge cases around cookie and API key auth handling
- cleaned up a few import/config issues needed to make the main-owned implementation build and test correctly

## Why this scope

The main motivation here is boundary correctness:

- request auth resolution depends on main-only capabilities
- renderer should call a bridge for that work rather than import it directly
- the renderer-node-import checks should fail if we regress back toward renderer importing main-only auth code

## Out of scope

This PR does **not**:

- move the entire request execution stack to main
- redesign the auth model or change the set of supported auth types
- refactor unrelated network code just because it sits near authentication
- do a broader OAuth architecture rewrite beyond relocating the active token/header flows needed for request execution
- change request/collection export formats or do broader Inso fixture restructuring beyond test coverage needed for this move


packages/insomnia/src/main/ipc/main.ts
```diff
@@ -287,7 +297,19 @@ export function registerMainHandlers() {
     }
   });
   ipcMainHandle('writeResponseBodyToFile', writeResponseBodyToFile);
-
+  ipcMainHandle('getAuthHeader', (_, renderedRequest: RenderedRequest, url: string) => {
```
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: we don't have authorization machinery

packages/insomnia/src/network/authentication.ts
```diff
@@ -1,146 +1,6 @@
...
-      name: 'Authorization',
-      value: generator(),
-    };
-  }
-
-  return;
-}
+import { QUERY_PARAMS } from './api-key/constants';
 
 export function getAuthQueryParams(authentication: RequestAuthentication) {
```
ZxBing0066: This doesn't need any more since you've move it to packages/insomnia/src/network/network.ts

packages/insomnia/src/main/network/o-auth-2/get-token.ts
```diff
@@ -0,0 +1,542 @@
+import crypto from 'node:crypto';
```
ZxBing0066, line 1: The previous get-token file still exists

packages/insomnia/src/main/network/get-auth-header.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  if (authentication.type === 'bearer' && authentication.token) {
+    const { token, prefix } = authentication;
+    return getBearerAuthHeader(token, prefix);
+  }
+
+  if (authentication.type === 'oauth2') {
+    try {
+      const tokenId = requestId.match(/\.graphql$/) ? requestId.replace(/\.graphql$/, '') : requestId;
```
ZxBing0066, line 74: Should we keep the previous comments?

packages/insomnia/src/common/constants.ts
```diff
@@ -258,6 +258,55 @@ export type AuthTypes =
 export const HAWK_ALGORITHM_SHA256 = 'sha256';
 export const HAWK_ALGORITHM_SHA1 = 'sha1';
 
+//oauth 1
+export type OAuth1SignatureMethod = 'HMAC-SHA1' | 'RSA-SHA1' | 'HMAC-SHA256' | 'PLAINTEXT';
```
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: we are an api client that offers flexibility, if a user wants to use an insecure hash its their choice.

ZxBing0066 (approved):

PR #9837, opened 2026-04-20
Feat: add custom npm registry mirror setting for plugin installation [INS-1269]

jeremyjpj0916:
### Summary
Adds a configurable npm registry URL setting in **Preferences > Plugins** to enable Insomnia plugin installation in corporate environments where direct access to \`registry.npmjs.org\` is blocked.

### Problem
Many enterprise environments block direct access to the public npm registry, requiring all npm traffic to route through internal mirrors (e.g., Artifactory, Nexus, Verdaccio). Previously, Insomnia hardcoded \`https://registry.npmjs.org/\` for plugin metadata fetching and installation, making it impossible to install plugins in these environments.

### Solution
- Added \`npmRegistryUrl\` setting to store a custom npm registry URL
- Modified \`install-plugin.ts\` to use the custom registry for both \`yarn info\` and \`yarn add\` commands
- Automatically adds the custom registry's hostname to the allowed tarball hosts list for security validation
- Falls back to the default npm registry when the setting is empty or invalid

### Files Changed
| File | Change |
|------|--------|
| \`packages/insomnia/src/common/settings.ts\` | Added \`npmRegistryUrl: string\` to Settings interface |
| \`packages/insomnia/src/insomnia-data/src/models/settings.ts\` | Added default empty value for \`npmRegistryUrl\` |
| \`packages/insomnia/src/main/install-plugin.ts\` | Core logic: dynamic registry URL, hostname allowlisting, helper functions |
| \`packages/insomnia/src/ui/components/settings/plugins.tsx\` | UI: Registry URL input field with tooltip and auto-save on blur |

### UI Changes
- New **"npm Registry"** section in **Preferences > Plugins**
- Text input with placeholder \`https://registry.npmjs.org/\`
- Help tooltip explaining the feature
- Auto-saves on blur (trimmed value)

### How It Works
1. User enters their corporate npm mirror URL (e.g., \`https://artifactory.example.com/api/npm/npm-remote/\`)
2. The registry is used for:
   - \`yarn info <plugin> --registry <url>\` — fetching plugin metadata
   - \`yarn add <plugin> --registry <url>\` — installing the plugin
3. The registry's hostname is automatically added to the allowed tarball hostnames for security validation
4. Works alongside existing **Proxy** and **Certificate** settings for complete corporate environment support

#### Setting the npm Mirror via the UI

1. Open **Preferences** (Cmd/Ctrl + ,)
2. Go to the **Plugins** tab
3. Under **npm Registry**, enter your corporate mirror URL (e.g., \`https://artifactory.example.com/api/npm/npm-remote/\`)
4. The setting is saved on blur. Leave empty to use the default (\`https://registry.npmjs.org/\`)

#### How It Works (user-facing)

- The custom registry URL is used for both fetching plugin metadata (\`yarn info\`) and installing plugins (\`yarn add\`).
- The hostname of the custom registry is automatically added to the allowed tarball hosts, so tarballs served by your mirror will pass validation.
- If the custom URL is invalid, Insomnia falls back to the default npm registry.
- This setting works alongside existing **Proxy** and **Certificate** settings (HTTP/HTTPS proxy, \`NODE_EXTRA_CA_CERTS\`) which may also be needed in corporate environments.

#### Typical Corporate Setup

For a corporate environment behind a firewall, you may need to configure all of the following in **Preferences > Plugins**:

| Setting | Example | Purpose |
|---------|---------|---------|
| npm Registry | \`https://artifactory.corp.com/api/npm/npm-remote/\` | Route npm traffic through your mirror |
| Certification | Upload your corporate CA bundle (\`.pem\`, \`.crt\`) | Trust your corporate TLS-intercepting proxy |

And in **Preferences > Proxy**:

| Setting | Example | Purpose |
|---------|---------|---------|
| HTTP Proxy | \`http://proxy.corp.com:8080\` | Route HTTP traffic through corporate proxy |
| HTTPS Proxy | \`http://proxy.corp.com:8080\` | Route HTTPS traffic through corporate proxy |

### Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Code comments added for complex logic
- [x] Changes generate no new warnings

### DIFF
--- a/packages/insomnia/src/ui/components/settings/plugins.tsx
+++ b/packages/insomnia/src/ui/components/settings/plugins.tsx
@@ -2,6 +2,7 @@ import React, { type FC, useEffect, useState } from 'react';
 import {
   Button,
   Checkbox,
+  FieldError,
   FileTrigger,
   GridList,
   GridListItem,
@@ -27,6 +28,24 @@ import { Icon } from '../icon';
 import { Tooltip } from '../tooltip';
 import { CreatePluginModal } from './create-plugin-modal';
 
+const getNpmRegistryUrlValidationError = (url: string): string | null => {
+  if (!url) {
+    return null;
+  }
+
+  try {
+    const parsedUrl = new URL(url);
+
+    if (parsedUrl.protocol !== 'http:' && parsedUrl.protocol !== 'https:') {
+      return 'Enter a valid HTTP or HTTPS URL.';
+    }
+
+    return null;
+  } catch {
+    return 'Enter a valid HTTP or HTTPS URL.';
+  }
+};
+
 interface State {
   plugins: Plugin[];
   npmPluginValue: string;
@@ -35,6 +54,8 @@ interface State {
   isInstallingFromNpm: boolean;
   isRefreshingPlugins: boolean;
   pluginNodeExtraCerts: string;
+  npmRegistryUrl: string;
+  npmRegistryUrlError: string | null;
 }
 
 export const Plugins: FC = () => {
@@ -50,6 +71,8 @@ export const Plugins: FC = () => {
       isRefreshingPlugins,
       npmPluginValue,
       pluginNodeExtraCerts,
+      npmRegistryUrl,
+      npmRegistryUrlError,
     },
     setState,
   ] = useState<State>({
@@ -60,6 +83,8 @@ export const Plugins: FC = () => {
     isInstallingFromNpm: false,
     isRefreshingPlugins: false,
     pluginNodeExtraCerts: settings.pluginNodeExtraCerts,
+    npmRegistryUrl: settings.npmRegistryUrl,
+    npmRegistryUrlError: null,
   });
 
   // If all plugins are enabled, we show the checked state
@@ -71,6 +96,10 @@ export const Plugins: FC = () => {
   useEffect(() => {
     setState(state => ({ ...state, pluginNodeExtraCerts: settings.pluginNodeExtraCerts }));
   }, [settings.pluginNodeExtraCerts]);
+
+  useEffect(() => {
+    setState(state => ({ ...state, npmRegistryUrl: settings.npmRegistryUrl, npmRegistryUrlError: null }));
+  }, [settings.npmRegistryUrl]);
 
   useEffect(() => {
     handleReloadPlugins();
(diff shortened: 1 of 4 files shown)


DEVELOPMENT.md
```diff
@@ -129,3 +129,37 @@ bump the following node and electron versions
 - `.nvmrc`
 - `packages/insomnia/package.json` electron and node-libcurl
 - `shell.nix`
+
+## npm Registry Mirror (Corporate Proxy/Firewall)
```
ryan-willis: This section reads as user documentation, which is best suited for https://github.com/Kong/developer.konghq.com instead of this development document.

Remove this markup from here and move it into the section in the PR description. Our docs team will make sure the setting is documented publicly and/or included in release notes when the next version is released

ryan-willis (changes requested):
Thanks for the contribution! One small change and it's good to go 🚀

jeremyjpj0916:
@ryan-willis hopefully tweaked it how you wanted, thx.

packages/insomnia/src/ui/components/settings/plugins.tsx
```diff
@@ -314,6 +321,59 @@ export const Plugins: FC = () => {
...
+                value={npmRegistryUrl}
+                onChange={value => {
+                  setState(state => ({ ...state, npmRegistryUrl: value }));
+                }}
+              >
+                <Input
+                  placeholder="https://registry.npmjs.org/"
+                  className="flex h-(--line-height-xs) w-full items-center rounded-md border border-solid border-(--hl-md) bg-(--hl-xxs) p-(--padding-sm) text-(--color-font) focus:border-(--hl-lg) focus:bg-transparent"
+                  onBlur={() => {
+                    patchSettings({ npmRegistryUrl: npmRegistryUrl.trim() });
```
ZxBing0066: It's better to check if it's valid before patching it.

ryan-willis (approved):

packages/insomnia/src/ui/components/settings/plugins.tsx
jeremyjpj0916: checking for http and https urls sufficient? I guess could be really strict and require typical trailing / but tbh most people will be copy pasting in their mirror to the box haha.

Added a check commit tho.

jeremyjpj0916:
Also is some internal Optum request ticket for this btw so may wanna tie this PR to that request just so yall know it will be done once this merges and don't later start dup work.

jeremyjpj0916:
@CurryYangxx  Did you resolve your findings? I Saw an earlier email but comment must have gotten deleted?

"
Hello, have you tested scenarios that require a certificate?
I use Verdaccio to build a custom registry server on my local laptop. When the server requires a cert, the installation failed, and I encountered this error:

Error: Error invoking remote method 'installPlugin': Error: Failed to fetch plugin tarball https://test-registry.local:4873/insomnia-plugin-github-theme/-/insomnia-plugin-github-theme-1.2.7.tgz: net::ERR_UNEXPECTED
    at IpcRenderer.invoke (node:electron/js2c/renderer_init:2:9840)
    at async invokeWithNormalizedError (/Users/[curry.yang@konghq.com/dev/Kong/insomnia-dev/packages/insomnia/src/entry.preload.min.js:23:12](http://curry.yang@konghq.com/dev/Kong/insomnia-dev/packages/insomnia/src/entry.preload.min.js:23:12))
    at async onPress (http://localhost:3334/src/ui/components/settings/plugins.tsx:309:21)
It seems electron's net.fetch can not recognize the NODE_EXTRA_CA_CERTS environment variable.
This step is just for checking in advance, not the installation process. I can install successfully after I comment out this step.
I am not sure if there is any difference between the local registry I built and the real corporate registry.
This is legacy logic not related to the current pr change, I just want to confirm with you whether my test approach is not correct.
"

jeremyjpj0916:
I think my PR exposed an existing mismatch between “Yarn install trust config” and “Electron tarball preflight trust config.”

> This is not caused by the npm registry mirror implementation. The registry URL is correctly passed to both yarn info and yarn add, and the custom registry host is allowed for tarball validation. The failing path is the existing net.fetch(info.dist.tarball) accessibility check, which uses Electron/Chromium networking instead of the same Node/Yarn environment that receives NODE_EXTRA_CA_CERTS. If the registry cert is trusted by the OS/Chromium trust store, it should pass; if it is only trusted via the plugin extra cert setting, the preflight can fail even though Yarn install would work. The appropriate fix would be to remove that preflight check or replace it with a check that uses the same cert/proxy configuration as Yarn, while keeping the allowed-host validation.

CurryYangxx:
@jeremyjpj0916, sorry for the late reply. I found the reason is that in the tarball preflight process, we did not pass the cert to the server. I think we don't need the preflight.
This issue is not related to you pr so I deleted the comment. We'll address this issue later. Thank you for your attention!

PR #9851, opened 2026-04-22
Refactor:use electron store for oauth session

jackkav:
prep work for oauth PR #9834

This code path is covered by oauth.test.ts

the biggest decision in this PR is the lifespan of the electronStore.

Option 1: Per-request / per-IPC-call (ephemeral)
A new instance is created for each request and discarded after the response. This is the most granular option — cheapest in terms of retained memory, but it means no state persists between calls. Only viable if every caller is fully self-contained and never needs to read back what it wrote.

Option 2: Per-window lifetime
The store instance is tied to a BrowserWindow — created when the window opens, destroyed when it closes. Since localStorage semantics in a renderer are naturally scoped to an origin per window, this maps intuitively to the mental model. Works well if different windows genuinely need isolated state.

Option 3: Application lifetime (singleton)
One instance is created at app startup (in the main process module or a dedicated service), lives for the duration of the process, and is shared across all IPC callers. Destroyed only when the app exits. Optionally flushed to disk for persistence across launches.

Option 4: Persistent / session-restored
Similar to application lifetime, but the store is serialised on quit and rehydrated on launch. Closer to what actual localStorage does — data outlives the process. Adds a write/read cycle around app lifecycle events.

Optimal choice: application lifetime singleton (Option 3)
For an Electron main-process context, the singleton is the right default, and the rationale is straightforward.
The main process is itself a singleton — there's exactly one of it, and it owns IPC. Any state stored there is already implicitly global to the app session. Creating a new store instance per IPC call throws away the entire point of having a store: accumulation and recall within a session. Per-window scoping is semantically correct for renderer-side localStorage, but in the main process you've already crossed the abstraction boundary — you're emulating the storage interface, not the storage scope, so per-window isolation is usually unnecessary overhead unless you have a concrete isolation requirement.
The application lifetime singleton also gives you a natural place to attach flush-on-quit behaviour later (upgrading to Option 4) without changing any caller code — you just add app.on('quit', () => store.flush()). That upgrade path is nearly free; the reverse (collapsing a per-request design into a singleton) requires more surgery.
The one condition that would push you toward Option 2 is if windows have meaningfully different users or security boundaries — for example, a settings window vs. a sandboxed preview window. In Insomnia's case, where windows share the same user session and data model, that boundary doesn't apply, and a singleton is both simpler and more consistent with how the rest of the main-process state is managed.

### DIFF
--- a/packages/insomnia/src/main/electron-storage.ts
+++ b/packages/insomnia/src/main/electron-storage.ts
@@ -3,11 +3,22 @@ import path from 'node:path';
 
 import { app } from 'electron';
 
-let electronStorage: ElectronStorage | null = null;
+import { isDevelopment } from '~/common/constants';
+
+import { userDataFolder } from '../../config/config.json';
+
+// resilience against userData folder being changed while the app is running
+// if it changes we just make a new ElectronStorage instance with the new path
+const electronStorageByPath = new Map<string, ElectronStorage>();
 export function getElectronStorage() {
-  const electronStoragePath = path.join(process.env['INSOMNIA_DATA_PATH'] || app.getPath('userData'), 'localStorage');
+  const dataPath =
+    process.env.INSOMNIA_DATA_PATH ||
+    path.join(app.getPath('userData'), '../', isDevelopment() ? 'insomnia-app' : userDataFolder);
+  const electronStoragePath = path.join(dataPath, 'localStorage');
+  let electronStorage = electronStorageByPath.get(electronStoragePath);
   if (!electronStorage) {
     electronStorage = new ElectronStorage(electronStoragePath);
+    electronStorageByPath.set(electronStoragePath, electronStorage);
   }
   return electronStorage;
 }
--- a/packages/insomnia/src/main/electron-storage.ts
+++ b/packages/insomnia/src/main/electron-storage.ts
@@ -7,14 +7,8 @@ import { invariant } from '~/utils/invariant';
 let electronStorage: ElectronStorage | null = null;
 export function initElectronStorage(dataPath: string) {
   const electronStoragePath = path.join(dataPath, 'localStorage');
-  if (electronStorage) {
-    // In dev, loud failure > silent no-op
-    invariant(
-      process.env.NODE_ENV !== 'development',
-      `ElectronStorage already initialized. Attempted re-init with: ${electronStoragePath}`,
-    );
-    return;
-  }
+  // Ensure that electronStorage is not yet initialized before creating a new instance. This prevents accidental re-initialization with a different path, which could lead to data loss.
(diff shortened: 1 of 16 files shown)


packages/insomnia/src/ui/spawn-oauth-window.ts
```diff
@@ -0,0 +1,15 @@
+import { v4 as uuidv4 } from 'uuid';
+
+import { OAUTH_WINDOW_SESSION_ID_KEY } from '~/common/constants';
+
+export async function initNewOAuthSession() {
```
ZxBing0066: This seems not used at all, should we remove it?

packages/insomnia/src/main/ipc/electron-storage.ts
```diff
@@ -0,0 +1,30 @@
+import { getElectronStorage } from '../electron-storage';
+import { ipcMainHandle } from './electron';
+
+export interface electronStorageBridgeAPI {
+  getItem: (key: string) => Promise<string | null>;
+  setItem: (key: string, value: string) => Promise<void>;
+}
+
+function validateElectronStorageKey(key: string): string {
```
ZxBing0066: 👍 Great guard to secure the IPC trust boundary.

How about move it inside the ElectronStorage since it also valid for the main process.

packages/insomnia/src/main/electron-storage.ts
```diff
@@ -1,6 +1,27 @@
...
 
+import { invariant } from '~/utils/invariant';
+
+// Intentional singleton: initialized once per process via initElectronStorage and shared across the app.
+let electronStorage: ElectronStorage | null = null;
+export function initElectronStorage(dataPath: string) {
+  const electronStoragePath = path.join(dataPath, 'localStorage');
+  if (electronStorage) {
+    // In dev, loud failure > silent no-op
+    invariant(
```
ZxBing0066: IMO, it's unnecessary to silent this when dev  since it should only be called once in the main process.

packages/insomnia-api/tsconfig.json
```diff
@@ -9,7 +9,6 @@
     "isolatedModules": true,
     "module": "ESNext",
     "sourceMap": true,
-    "baseUrl": ".",
```
jackkav, line 12: not strictly in scope, but minor.

packages/insomnia/src/main/ipc/electron.ts
```diff
@@ -120,6 +120,8 @@ export type HandleChannels =
   | 'readDir'
   | 'readOrCreateDataDir'
   | 'restoreBackup'
+  | 'electronStorage.getItem'
```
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: we have no way to track caller identity or resource ownership at this time.

ZxBing0066 (approved):
🚀

PR #9861, opened 2026-04-23
fix: socket.io server sent disconnect state

ryan-willis:
The fix was pretty simple; the socket.io client defaults to reconnect, which breaks the Insomnia-managed connectivity state


fiosman:
Despite my best effort I was not able to replicate this issue. The UI state in the app was immediately updated upon starting/stopping my local socket.io server.

PR #9889, opened 2026-05-04
feat: move plugin loading to dedicated hidden BrowserWindow (Phase 1)

jackkav:
## Summary

Moves all plugin loading and execution out of the UI renderer into a dedicated hidden \`BrowserWindow\` with \`nodeIntegration: true\`. The renderer communicates with the plugin window exclusively over IPC.

### What changed

- **New hidden plugin window** (\`src/plugin-window.html\` + \`src/entry.plugin-window.ts\`) — loads and executes all plugin code with Node.js access
- **Preload for plugin window** (\`src/entry.plugin-window-preload.ts\`) — provides \`window.app.getPath\` via \`ipcRenderer.sendSync\` so plugins can resolve \`userData\`
- **Main process manager** (\`src/main/plugin-window.ts\`) — creates the window; routes IPC via a pending-request map (UUID correlation, 10s ready timeout, sender validation, crash drain on close)
- **IPC database proxy** (\`src/main/database.plugin-window.ts\`) — plugin window reads from the main process NeDB connection via IPC instead of opening a second connection; services are initialized before the window signals readiness
- **Serializable type boundary** (\`src/plugins/bridge-types.ts\`) — all data crossing IPC is serializable; context modules run in the plugin window, not the renderer
- **Renderer bridge** (\`window.main.plugins.*\`) — all plugin execution replaced with bridge calls
- **Build** — two new esbuild entry points added; dev \`buildCount\` threshold updated from 3 → 6

### Architecture

```
Renderer
  window.main.plugins.executeAction({ type, pluginName, label, projectId, domainData })
    └─ ipcRenderer.invoke('plugins.executeAction', args)
         └─ ipcMain.handle → invokeInPluginWindow('executeAction', args)
              └─ pluginWindow.webContents.send('plugin-invoke', { id, method, args })
                   └─ entry.plugin-window.ts: finds action, builds context, calls action(ctx, domainData)
                        └─ ipcRenderer.send('plugin-invoke-result', { id, result })
                             └─ main pendingRequests.get(id).resolve(result)
                                  └─ renderer Promise resolves
```

### What's bridged (Phase 1 complete)

| Feature | Status |
|---|---|
| Theme listing | ✅ |
| Plugin listing / reload | ✅ |
| Bundle plugin listing (\`getBundlePlugins\`) | ✅ |
| Request / RequestGroup / Workspace / Document actions (list + **execute**) | ✅ |
| Template tag listing (\`getTemplateTags\`) | ✅ |
| Template tag action execution (\`runTemplateTagAction\`) | ✅ |
| Elevated plugin main actions (\`executePluginMainAction\`) | ✅ |
| Request hooks / Response hooks | ✅ |
| Sandbox hardening | ⏳ Phase 2 — \`contextIsolation: true\` once API surface is stable |

### Hook bridging design

Request and response hooks mutate their context objects by reference, which only works within a single process. To cross the IPC boundary the plugin window receives a serialized copy of the request/response, runs all registered hooks against it, and returns the fully mutated object — matching the same pattern used by pre-request scripts.

The built-in default-headers hook runs in the renderer without any IPC (it has no plugin dependency). A \`hasRequestHooks\`/\`hasResponseHooks\` result is cached in the main process after the first check and cleared on \`reloadPlugins\`, so requests incur zero plugin-window round-trips when no user plugins have hooks registered.

### Non-renderer process handling

\`_applyRequestPluginHooks\` and \`_applyResponsePluginHooks\` guard on \`process.type !== 'renderer'\` before using the IPC bridge. In the Electron main process (OAuth token exchange) and in the inso CLI (Node.js), hooks are applied directly via \`plugins.getRequestHooks()\` / \`plugins.getResponseHooks()\` without any IPC.

### Intentionally remaining in renderer

- \`applyColorScheme\` / \`getColorScheme\` — CSS/DOM manipulation, not plugin execution
- \`createPlugin\` — filesystem plugin scaffolding, not plugin execution

### Security fixes (from review)

- \`plugin-window-ready\` and \`plugin-invoke-result\` validate \`event.sender === pluginWindow.webContents\`
- Persistent \`ipcMain.on\` listener (registered once) replaces \`ipcMain.once\` — plugin window re-arms correctly after reload
- \`waitForReady()\` has a 10s timeout and listens to \`did-fail-load\`
- Request IDs use \`crypto.randomUUID()\` instead of a guessable counter
- Non-Error plugin hook throws are wrapped into \`Error\` instances before attaching \`.plugin\` metadata

### Playwright fixture

The plugin window is created after the main window's \`did-finish-load\` event, so Playwright's \`firstWindow()\` always returns the main app window. No fixture fallback or polling is needed.

## Test plan

- [x] All 1900 unit tests pass (\`npm test -w packages/insomnia\`)
- [x] TypeScript: no errors (\`tsc --noEmit\`)
- [x] E2E: \`plugin-bridge.test.ts\` — writes a test plugin, reloads via bridge, verifies the request action appears in the dropdown (16.7s, passes)
- [ ] Manual: themes settings panel — themes should populate
- [ ] Manual: reload plugins via keyboard shortcut — should complete without error
- [ ] Manual: plugin settings tab — installed plugins should list
- [ ] Manual: right-click a request — plugin actions should appear and execute
- [ ] Manual: cloud credentials settings — Azure auth flow should work

closes INS-2466


packages/insomnia/src/main/plugin-window.ts
```diff
@@ -0,0 +1,140 @@
...
+    show: false,
+    title: 'PluginWindow',
+    webPreferences: {
+      contextIsolation: false,
+      nodeIntegration: true,
+      preload: path.join(__dirname, 'entry.plugin-window-preload.min.js'),
+      backgroundThrottling: false,
+      devTools: process.env.NODE_ENV === 'development',
+    },
+  });
```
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: intended behaviour

packages/insomnia/src/main/plugin-window.ts
```diff
@@ -0,0 +1,171 @@
...
```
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: intended behaviour — this is a dedicated hidden BrowserWindow for plugin execution, isolated from user content. nodeIntegration is required for plugins to use Node APIs. Phase 2 will harden this with contextIsolation + sandbox.

packages/insomnia/src/main/plugin-window.ts
```diff
@@ -0,0 +1,140 @@
+import path from 'node:path';
+
+import { app, BrowserWindow, ipcMain } from 'electron';
+
+let pluginWindow: BrowserWindow | null = null;
+let windowReady = false;
+const pendingRequests = new Map<string, { resolve: (v: unknown) => void; reject: (e: Error) => void }>();
```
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: intentional design — the module-level Map is the pending-request registry, not a cache. Entries are always removed on resolve, reject, or window close (the closed handler drains all pending entries). No leak path exists.

packages/insomnia/src/main/plugin-window.ts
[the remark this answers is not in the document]
jackkav: @AikidoSec ignore: the plugin window needs node in order to preserve behaviour.

packages/insomnia/src/main/network/socket-io.ts
ryan-willis, line 1: clean 👍🏻

packages/insomnia/src/main/ipc/electron.ts
```diff
@@ -328,10 +342,10 @@ export function registerElectronHandlers() {
               },
               { type: 'separator' },
             ];
-        const localTemplate: MenuItemConstructorOptions[] = [...localTemplateTags, ...pluginTemplateTags]
+        const localTemplate: MenuItemConstructorOptions[] = ([...localTemplateTags, ...pluginTemplateTags] as any[])
```
ryan-willis: There's a lot of `any` casting, but aside from the actual plugin invocations, which previously were `Promise<any>`, is there a good way to make it less implicit?

packages/insomnia/PLUGIN_SYSTEM_POC.md
ryan-willis, line 1: I'd prefer to keep this in Confluence

jackkav: Its easier for the llm to reference here. I'll get rid of it once the work is done though and can add it to confluence for posterity.

ryan-willis (dismissed):

ZxBing0066:
Some plugin APIs that rely on the main window (e.g., showAlert, showWrapper) will trigger errors following this change. These APIs may require bridging to the main window to maintain functionality.

And if the previous plugins depend on some of the window APIs, like height and width, they also break. As well as if the plugins are trying to append DOMs. Is this acceptable?

jackkav:
feedback concern: 
chome web apis that show UI will need bridges, or be excluded from supported features. 
`await window.showOpenFilePicker();`

templating now runs in main, which means templating has access to nodejs again, temporarily.

injecting custom UI, in to the DOM won't work anymore, because it was using ambiant renderer libraries like React. We could add these to the plugin context and build bridges but this would be securable.

packages/insomnia/src/main/ipc/electron.ts
```diff
@@ -120,6 +119,20 @@ export type HandleChannels =
...
+  | 'plugins.runTemplateTagAction'
+  | 'plugins.getActivePlugins'
+  | 'plugins.getDocumentActions'
+  | 'plugins.getPlugins'
+  | 'plugins.getRequestActions'
+  | 'plugins.getRequestGroupActions'
+  | 'plugins.getThemes'
+  | 'plugins.getWorkspaceActions'
+  | 'plugins.reloadPlugins'
+  | 'plugin-ui-prompt'
```
ZxBing0066, line 135: Some of the plugin channels' naming style are inconsistent.

AGENTS.md
```diff
@@ -72,3 +81,47 @@ Organization
 ## Sensitive Data
 - **Vault system (AES-GCM):** For environment secrets (`EnvironmentKvPairDataType.SECRET`).
 - **Electron safeStorage:** Platform-native encryption (`window.main.secretStorage`).
+## cx — Semantic Code Navigation
+
+Prefer cx over reading files. Escalate: overview → symbols → definition/references → Read tool.
+
+### Quick reference
```
ZxBing0066, line 88: This seems more like a skill rather than an agent instruction. How do you think?

packages/insomnia/src/root.tsx
```diff
@@ -24,9 +24,10 @@ import { useLatest } from 'react-use';
 import { EXTERNAL_VAULT_PLUGIN_NAME, isDevelopment } from '~/common/constants';
 import type { Settings, UserSession } from '~/insomnia-data';
 import { models, services } from '~/insomnia-data';
-import { executePluginMainAction, reloadPlugins } from '~/plugins';
+import { executePluginMainAction } from '~/plugins';
```
ZxBing0066, line 27: Should we call executePluginMainAction from the bridge here?

packages/insomnia/src/entry.preload.ts
```diff
@@ -344,6 +361,28 @@ const main: Window['main'] = {
...
+    executeAction: (args: ExecutePluginActionArgs) => invokePluginBridgeMethod('executeAction', args),
+    getTemplateTags: () => invokePluginBridgeMethod('getTemplateTags'),
+    runTemplateTagAction: (args: RunTemplateTagActionArgs) => invokePluginBridgeMethod('runTemplateTagAction', args),
+    getBundlePlugins: () => invokePluginBridgeMethod('getBundlePlugins'),
+    executePluginMainAction: (args: ExecutePluginMainActionArgs) => invokePluginBridgeMethod('executePluginMainAction', args),
+    hasRequestHooks: () => invokePluginBridgeMethod('hasRequestHooks'),
+    hasResponseHooks: () => invokePluginBridgeMethod('hasResponseHooks'),
+    applyRequestHooks: (args: ApplyRequestHooksArgs) => invokePluginBridgeMethod('applyRequestHooks', args),
+    applyResponseHooks: (args: ApplyResponseHooksArgs) => invokePluginBridgeMethod('applyResponseHooks', args),
+    getBridgeMetrics: () => invokeWithNormalizedError('plugins.getBridgeMetrics'),
```
ZxBing0066, line 382: Should we use invokePluginBridgeMethod here instead of invokeWithNormalizedError here?

packages/insomnia/src/main/plugin-window.ts
```diff
@@ -0,0 +1,316 @@
...
+      promptPendingRequests.set(id, value => {
+        clearTimeout(timeout);
+        resolve(value);
+      });
+      mainWindow.webContents.send('plugin-ui-prompt', id, options);
+    });
+  });
+
+  ipcMain.on('plugin-ui-prompt-result', (_event, { id, value }: { id: string; value: string | null }) => {
+    const resolve = promptPendingRequests.get(id);
```
ZxBing0066, line 128: No sender check here as other handlers, is it missing?

packages/insomnia/src/entry.plugin-window.ts
```diff
@@ -0,0 +1,36 @@
...
+    ipcRenderer.send('plugin-invoke-result', { id, error: errMsg });
+  }
+});
+
+// Initialize database (via IPC proxy) and services before signalling readiness.
+// getPlugins() calls services.settings.get(), which requires this to be done first.
+(async () => {
+  try {
+    await initDatabase(pluginWindowDatabase);
+    initServices(servicesNodeImpl);
```
ZxBing0066, line 31: It's better to use servicesProxy instead of the servicesNodeImpl to avoid unnecessary imports to the hidden window.

ZxBing0066 (approved):

PR #9920, opened 2026-05-14
feat: Custom lint rules [INS-2338]

fiosman:
### Summary

1. Insomnia
   - Adds File Picker to the design document UI to enable a user to select a custom yaml/yml Spectral ruleset to lint their specs for local, cloud and git sync projects. Validator added to ensure the content is valid spectral syntax. Currently only supporting `extends` and `rules` as top level keys. The custom ruleset uploaded is project scoped (to ensure compatibility with the current linting implementation for git projects where a `.spectral.yaml` is placed at the root of the repo).
   - Added support for nesting linting rules (e.g. the spectral ruleset `extends` references a local file path which references another local path, etc.). Recursion is used to walk down the `extends` tree and flatten all local rulesets into a single valid spectral ruleset that can be synced to cloud/git projects and collaborators would be able to use. Note that we also validate remote urls in extends but we do not flatten them in the final merged ruleset, as that would mean users would need to re-upload their custom rulesets to pick up changes in remote rulesets. 
   - Patched vulnerabilities associated with Spectral rulesets (e.g. timeout limit for `lintSpec` to mitigate against ReDoS, not allowing custom functions, protecting against prototype pollution, mitigating SSRF via `extends` URLs). 

1. Inso 
   - Adds `--ruleset` flag to allow a user to input a file path to lint their target spec. 

### Test Plan for Insomnia/Inso
- Cloud Sync
  - [x] User A uploads a ruleset → User B on the same project pulls → B sees the same ruleset.
  - [x] A updates the ruleset → B pulls → update reflected.
  - [x] A clears the ruleset → B pulls → ruleset removed.
  - [x] fetch, branch checkout, branch create, branch merge, rollback, restore all work with a custom ruleset.
  - [x] Cloud<>Git and Cloud<>Local project conversions carries over the ruleset. 
  - [x] Upload invalid ruleset. Does the user get a descriptive error message?
  - [x] Test out nesting. Add a ruleset that `extends` from a remote URL. Add a ruleset that `extends` from a local file path. 
 
- Git Sync
  - [x] Upload a ruleset → .spectral.yaml appears in the commit modal UI; `git status` shows it.
  - [x] Edit `.spectral.yaml` externally → app picks up the change (record updates, lint reflects it).
  - [x] Delete .spectral.yaml externally → lint falls back to default.
  - [x] Commit, push, clone elsewhere → ruleset present on the clone.
  - [x] Pulling changes works as a collaborator
  - [x] Git<>Cloud and Git<Local> project conversions carries over the ruleset.
  - [x] Upload invalid ruleset. Does the user get a descriptive error message? 
  
- Local
  - [x] Upload a ruleset in-app → spec lints automatically 
  - [x] Remove a ruleset in app -> fall back to OAS 
  - [x] Local<>Cloud and Local<>Git project conversions carries over the ruleset. 
  - [x] Upload invalid ruleset. Does the user get a descriptive error message?

- Inso
  - [x] Pass a `--ruleset` flag pointing to a local file path, does the spec lint?
  - [x] Does passing no ruleset still use default OAS ruleset to lint spec?
  - [x] Does having a `.spectral.yaml` in the root dir still get automatically picked up if no --ruleset is provided?
  - [x] Upload invalid ruleset. Does the user get a descriptive error message?
 
- Backwards compatibility for Insomnia
  - [ ] User A, on this build, part of a cloud project uploads a custom ruleset and syncs it. User B, on a older build, attempts to pull the changes. Ensure no unintended action (i.e. data loss) happens.


packages/insomnia/src/common/spectral-ruleset-validator.ts
```diff
@@ -0,0 +1,231 @@
...
+export function isPrivateOrLoopbackHost(hostname: string): boolean {
+  if (hostname === 'localhost' || hostname.endsWith('.localhost')) {
+    return true;
+  }
+  const host = hostname.startsWith('[') && hostname.endsWith(']') ? hostname.slice(1, -1) : hostname;
+  if (!ipaddr.isValid(host)) {
+    return false;
+  }
+  return ipaddr.process(host).range() !== 'unicast';
+}
```
kwburns-kong: Confirmed this can be circumventable during domain resolution to hit `localhost`. If the domain resolves to `127.0.0.1`  it's possible to interact with internal services. You can URL `https://app.localtest.me/rules.yaml` to verify.

packages/insomnia/src/common/spectral-ruleset-validator.ts
```diff
@@ -0,0 +1,231 @@
...
+];
+
+// For security reasons we do not allow rulesets to contain certain tokens that could be used for JavaScript prototype pollution when used in certain Spectral properties (e.g. "field").
+const PROTOTYPE_POLLUTION_TOKENS = ['__proto__', 'prototype', 'constructor'];
+
+// For security reasons we only allow extends URLs with certain safe schemes and hosts.
+const SAFE_URL_SCHEMES = ['https:'];
+
+// Check if path is absolute file path (e.g. /foo/bar.yaml, C:\foo\bar.yaml, \\server\share\file.yaml)
+function isAbsoluteFilePath(value: string): boolean {
```
kwburns-kong: Any way to use `insecureReadFile()` to limit this based on what users grant access to within settings?

packages/insomnia/src/common/spectral-ruleset-validator.ts
```diff
@@ -0,0 +1,231 @@
...
+  return value.startsWith('/') || value.startsWith('\\\\') || /^[A-Za-z]:[\\/]/.test(value);
```
kwburns-kong: Haven't found a way to leverage this into anything, but came up anyway. Depending on OS the`value.startsWith('\\\\') || /^[A-Za-z]:[\\/]/.test(value);` will resolve as a relative path. Any way to do a conditional check of OS to prevent that?

packages/insomnia/src/main/bundle-spectral-ruleset.ts
```diff
@@ -0,0 +1,143 @@
...
+  // Remove duplicate entries while preserving order,
+  // then return the final flattened ruleset.
+  const uniqueExtends = [...new Set(remainingExtends)];
+  delete flattenedRuleset.extends;
+  return uniqueExtends.length > 0 ? { extends: uniqueExtends, ...flattenedRuleset } : flattenedRuleset;
+}
+
+export async function bundleSpectralRuleset(sourcePath: string): Promise<string> {
+  const rootDir = path.dirname(path.resolve(sourcePath));
+  const flattenedRuleset = await flattenRuleset(sourcePath, new Set(), 0, rootDir);
```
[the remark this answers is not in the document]
Ali-Sab: @AikidoSec ignore: We let user pick the file they want, and we validate that the `extends` files are within userData to be safe and avoid unintended `extends` references.

packages/insomnia/src/main/bundle-spectral-ruleset.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+  const rel = path.relative(rootDir, absolute);
+  if (rel.startsWith('..') || path.isAbsolute(rel)) {
+    throw new Error(`"extends" target must stay within the ruleset's root directory: ${absolute}`);
+  }
+}
+
+// Reads and parses a ruleset file
+async function readRuleset(absolute: string): Promise<Ruleset> {
+  const raw = await fs.promises.readFile(absolute, { encoding: 'utf8' });
```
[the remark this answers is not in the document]
Ali-Sab: @AikidoSec ignore: The path-traversal defense already exists in assertAllowed, which confines reads to rootDir and blocks bad extensions before any readFile happens.

packages/insomnia/src/common/spectral-ruleset-validator.ts
```diff
@@ -0,0 +1,231 @@
...
+      url = new URL(path);
+    } catch {
+      return `"extends" entry "${path}" is not a recognized Spectral identifier or a valid URL.`;
+    }
+
+    if (!SAFE_URL_SCHEMES.includes(url.protocol)) {
+      return `"extends" entry "${path}" must use https (got "${url.protocol}").`;
+    }
+
+    if (!url.hostname || isPrivateOrLoopbackHost(url.hostname.toLocaleLowerCase())) {
```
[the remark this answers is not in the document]
fiosman: @copilot this no longer exists in spectral-ruleset-validator.ts and elsewhere where it is being used I am now using .toLowerCase()

packages/insomnia-inso/src/commands/lint-specification.ts
```diff
@@ -31,11 +33,17 @@ export async function lintSpecification({
...
   try {
     if (rulesetFileName) {
+      const rulesetContent = await fs.promises.readFile(rulesetFileName, 'utf8');
+      const validation = validateSpectralRuleset(rulesetContent);
+      if (!validation.isValid) {
+        logger.fatal(`Invalid Spectral ruleset: ${validation.error}`);
+        return { isValid: false };
+      }
       ruleset = await bundleAndLoadRuleset(rulesetFileName, { fs });
     }
```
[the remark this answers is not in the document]
fiosman: @copilot fixed, now I am calling `bundleSpectralRuleset` which validates every remote extends entry.

packages/insomnia/src/common/bundle-spectral-ruleset.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+  const rel = path.relative(rootDir, absolute);
+  if (rel.startsWith('..') || path.isAbsolute(rel)) {
+    throw new Error(`"extends" target must stay within the ruleset's root directory: ${absolute}`);
+  }
+}
+
+// Reads a local ruleset file from disk and parses it.
+async function readRuleset(absolute: string): Promise<Ruleset> {
+  const raw = await fs.promises.readFile(absolute, { encoding: 'utf8' });
```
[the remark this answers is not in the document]
fiosman: @AikidoSec ignore: `assertAllowed` is already doing this and it gets called before `readRuleset` in `flattenRuleset`. It would be redundant to add this.

packages/insomnia/src/common/bundle-spectral-ruleset.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+    // Remote URL extends — validate upfront (SSRF + content checks), then preserve the URL
+    // in "extends" for Spectral to fetch fresh at lint time via spectralRuntime.fetch.
+    if (!isLocalFilePath(entry)) {
+      await validateRemoteExtends(parseRemoteExtendsUrl(entry), nextVisited, depth + 1);
+      remainingExtends.push(entry);
+      continue;
+    }
+    // Local file paths are recursively loaded and flattened.
+    const childRuleset = await flattenRuleset(path.resolve(baseDir, entry), nextVisited, depth + 1, rootDir);
```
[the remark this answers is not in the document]
fiosman: @AikidoSec ignore: This check is redundant. `assertAllowed` is always called before `readRuleset` in every code path, and it already performs a path traversal check against the original ruleset's root dir.

packages/insomnia/src/common/spectral-ruleset-validator.ts
fiosman: Added logic to check DNS resolution

packages/insomnia/src/common/spectral-ruleset-validator.ts
fiosman: @kwburns-kong I am now using `path.isAbsolute` (platform aware) in `bundle-spectral-ruleset.ts`. I have removed `isAbsoluteFilePath` and `isLocalFilePath` from `spectral-ruleset-validator.ts`.

packages/insomnia/src/main/ipc/main.ts
```diff
@@ -329,15 +332,64 @@ export function registerMainHandlers() {
...
+    }
+  });
   ipcMainHandle('writeResponseBodyToFile', writeResponseBodyToFile);
   ipcMainHandle('getAuthHeader', (_, renderedRequest: RenderedRequest, url: string) => {
     return getAuthHeaderInMain(renderedRequest, url);
   });
   ipcMainHandle('getOAuth2Token', (_, requestId: string, authentication: AuthTypeOAuth2, forceRefresh?: boolean) => {
     return getOAuth2TokenInMain(requestId, authentication, forceRefresh);
   });
+  ipcMainHandle('bundleSpectralRuleset', async (_, options: { sourcePath: string }) => {
```
jackkav, line 409: This handler accepts an arbitrary `sourcePath` from the renderer and passes it directly to `bundleSpectralRuleset`, which reads that file and all its local `extends` references from disk. A compromised renderer could call this with `/etc/passwd` or `~/.ssh/id_rsa` and get the contents back.

The `lintSpec` handler just below already has the right pattern — restricting `rulesetPath` to within `userData/` via `path.relative`. This handler should apply the same restriction (or use `isPathAllowed`/`secureReadFile`) before passing `sourcePath` to `bundleSpectralRuleset`.

packages/insomnia/src/main/ipc/main.ts
```diff
@@ -329,15 +332,64 @@ export function registerMainHandlers() {
       throw new Error(err);
     }
   });
+  ipcMainHandle('deleteFile', async (_, options: { path: string }) => {
```
jackkav: Same issue as `bundleSpectralRuleset` above: this handler accepts an arbitrary `path` from the renderer and deletes it with no restriction. Should be limited to files within `userData/` before calling `fs.promises.unlink`.

Also: `throw new Error(err)` when `err` is already an `Error` produces a double-wrapped `"Error: Error: ..."` message. Prefer `throw err instanceof Error ? err : new Error(String(err))`.

packages/insomnia/src/common/bundle-spectral-ruleset.ts
```diff
@@ -0,0 +1,232 @@
...
+    if (childRuleset.extends) {
+      remainingExtends.push(...childRuleset.extends);
+    }
+    mergeInto(flattenedRuleset, childRuleset); // child takes precedence over parent
```
jackkav: The comment `// child takes precedence over parent` is backwards. The parent is applied last (line 210) and wins over all children. What this line actually does is merge each child in order so that later `extends` entries override earlier ones. A clearer comment would be something like `// later extends entries override earlier ones`.

packages/insomnia/src/main/lint-process.mjs
```diff
@@ -1,15 +1,79 @@
...
+
+// Note: This is duplicated in inso's lint-specification.ts. Remember to mirror changes there as well.
+const safeHttpResolver = {
+  async resolve(ref) {
+    const href = ref.href();
+    if (!isSafeRefUrl(href)) {
+      throw new Error(`Failed to fetch "${href}". Only https URLs to public hosts are allowed.`);
+    }
+    await assertResolvesToPublicHost(new URL(href).hostname.toLowerCase());
+    const response = await fetch(href, { redirect: 'error' });
```
jackkav: No fetch timeout here. If a remote `$ref` endpoint hangs, this ties up the lint worker until the 30s process-kill fires. Consider adding `signal: AbortSignal.timeout(10_000)` to be consistent with `bundle-spectral-ruleset.ts` (`REMOTE_FETCH_TIMEOUT_MS`) and to fail faster per-request rather than waiting for the full worker timeout.

jackkav (commented):
Good overall — clean security model, thorough validation, and solid test coverage. A few issues worth addressing before merge, mainly around IPC path restrictions.

**Summary of findings:**

- `bundleSpectralRuleset` and `deleteFile` IPC handlers accept arbitrary renderer-supplied paths with no restriction — the same risk the `lintSpec` handler already guards against with `userData/` path restriction.
- The `safeHttpResolver` in `lint-process.mjs` has no per-fetch timeout (the 30s worker kill mitigates at process level, but `AbortSignal.timeout` would be cleaner and consistent with `bundle-spectral-ruleset.ts`).
- A comment in `flattenRuleset` contradicts the actual merge precedence.
- Three-way duplication of `isPrivateOrLoopbackHost` / HTTP resolver helpers is acknowledged in comments but worth tracking — divergence in security-critical code across copies is a maintenance hazard.
- No size limit on `rulesetContent` — worth adding a reasonable cap (e.g. 1 MB) during upload.
- The smoke test removes the `'Lint problems detected'` click and toggle panel lines without explanation — if those UI elements still exist, this is a coverage regression.

packages/insomnia/src/common/private-host.ts
```diff
@@ -0,0 +1,18 @@
+import ipaddr from 'ipaddr.js';
```
gatzjames: Could we use net.isIPv4/isIPv6 instead of adding the ipaddr package?

packages/insomnia/src/main/ipc/main.ts
fiosman: @jackkav Thanks! Correct me if I am wrong but by applying this restriction, the user would not be able to select a file outside of the `userData` directory (the renderer calls this handler via `handleSelectSpectralFile` so it is the source of truth; the user is selecting this path themselves). Users legitimately keep rulesets anywhere on their machine (e.g. ~/projects/...) so I feel that would impact the UX. 

`bundleSpectralRuleset` also calls `validateSpectralRuleset` which would reject anything that is not a valid spectral file format.

`lintSpec` is different because it reads the already _bundled_ ruleset which we wrote ourselves to `userData/`. It absolutely makes sense to have that restriction there. 

Those are my two cents, please let me know your thoughts :)

packages/insomnia/src/common/private-host.ts
fiosman: @gatzjames Done. I had unit tests in place for this function and they all pass per the new functionality which utilizes the net module from node.

gatzjames (approved):

jackkav (approved):
🚀

packages/insomnia/src/common/bundle-spectral-ruleset.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+  const rel = path.relative(rootDir, absolute);
+  if (rel.startsWith('..') || path.isAbsolute(rel)) {
+    throw new Error(`"extends" target must stay within the ruleset's root directory: ${absolute}`);
+  }
+}
+
+// Reads a local ruleset file from disk and parses it.
+async function readRuleset(absolute: string): Promise<Ruleset> {
+  const raw = await fs.promises.readFile(absolute, { encoding: 'utf8' });
```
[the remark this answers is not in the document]
fiosman: @AikidoSec ignore: assertAllowed is already doing this and it gets called before readRuleset in flattenRuleset. It would be redundant to add this.

PR #9924, opened 2026-05-14
fix(sandbox): INS-2464 additional checks  

kwburns-kong:
This PR adds a resolution for INS-2464, increases the existing unit test count and optimizes existing sandbox smoke test.


packages/insomnia/src/scripting/sandbox.ts
```diff
@@ -344,6 +388,25 @@ export async function prepareSandbox(
     ({ names: maskNames, values: maskValues } = alwaysOnPolicy.buildMaskScope(checkSandboxViolations));
   }
 
+  // Wrap eval so user-supplied source is checked and then evaluated in a scope that inherits the same masked bindings as the outer sandbox.
```
ihexxa, line 391: Will we also cover expressions like `const sum = new Function('a', 'b', 'return a + b');`, or composition of them?

packages/insomnia/src/scripting/sandbox.ts
```diff
@@ -106,36 +151,22 @@ export function checkSandboxViolations(
...
-      if (node.right.type === 'Identifier') {
-        const rightName = (node.right as acorn.Identifier).name;
-        const origin = blocked.get(rightName);
-        if (origin !== undefined) {
-          blocked.set(id.name, origin);
-        }
-      }
-      if (node.right.type === 'ThisExpression' && blockedRoots.has('this')) {
-        blocked.set(id.name, 'this');
+      const foundName = extractBlockedIdentifier(node.right, blocked, blockedRoots);
```
ihexxa, line 167: Hm will this also cover computed value?

ihexxa (commented):
I was occupied by some other stuff.. but will accelerate to review this.. 🏃

packages/insomnia/src/scripting/sandbox.ts
```diff
@@ -106,36 +151,22 @@ export function checkSandboxViolations(
   };
 
   walk.simple(tree, {
```
ihexxa, line 153: Will it also collect names that blocked ones which alias are assigned after? e.g:
```
const b = a;            // ← a not yet in blocked → b is missed
const a = globalThis;   // ← a added here
b.require('...');       // ← NOT blocked
```

packages/insomnia/src/scripting/sandbox.ts
ihexxa: Okay, it seems covered by `getMemberPropertyName`, resolving this.

packages/insomnia/src/scripting/sandbox.ts
ihexxa, line 129: Could the script be higher version (e,g., ecmaVersion 2025) which works but can't be parsed, then it is returned here.

ihexxa (commented):
Just several comments and questions.

PR #9935, opened 2026-05-19
chore: enable ts checks for test files

ZxBing0066:
## Changes

- Enable ts checks for the test files
- Fix some of the test files
- Off ban-ts-comment for the test files
- Fix a bug that `multiline` could be strings


packages/insomnia/src/insomnia-data/src/models/request.ts
```diff
@@ -205,7 +205,7 @@ export interface RequestBodyParameter {
   value?: string;
   description?: string;
   disabled?: boolean;
-  multiline?: boolean;
+  multiline?: boolean | string;
```
ZxBing0066, line 208: From the code, actually the multiline could be strings like `application/json`, set it as boolean only is a mistake.

Image

cwangsmv (approved):

PR #9936, opened 2026-05-19
feat: New navigation sidebar

cwangsmv:
# Background
 A common concern from the feedback came up around the lack of a sidebar. Because of that feedback, we think that a universal sidebar tree is the best pattern to use to help to communicate to users about how project data can be mapped to their existing workflows.

The goal here would be to make it easier for new and existing users to make sense of how data is organized in Insomnia and that gives them confidence to use the data model in however way they need to suit their needs.

# Changes
- New global navigation sidebar to replace the project list and request collection
   - Expand/Collapse, filterting support
   - New dropdown for project, workspace and request/request groups
   - Enhance drag & drop with cross project move support
- New breadcrumb and top menu to replace the old debug settings list
- Pin/Unpin request support
- Adapt the new sidebar with debug/Mcp/mock-server/design/environment page
- Integrate with Konnect support
- Fix all e2e tests due to navigation changes

[INS-2267](https://konghq.atlassian.net/browse/INS-2267)

[INS-2267]: https://konghq.atlassian.net/browse/INS-2267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CurryYangxx (approved):

packages/insomnia/src/common/project.ts
```diff
@@ -45,3 +86,224 @@ const lockGenerator = () => {
...
+    }
+
+    console.log(
+      '[getAllRemoteFiles] start fetching remote backend workspaces for project',
+      projectId,
+      `remoteId: ${remoteId}`,
+    );
+
+    const [allPulledBackendProjectsForRemoteId, allFetchedRemoteBackendProjectsForRemoteId] = await Promise.all([
+      window.main.sync.localBackendProjects().then(projects => projects.filter(p => p.id === remoteId)),
```
yaoweiprc, line 236: A backend project's ID can never be equal to a frontend project's remoteId.

PR #9940, opened 2026-05-21
Streamline workspace create & settings form [INS-2621]

godfrzero:
This PR addresses a few usability issues in the "Create a new *" and "* Settings" modals for Git Sync projects:
1. Editing just the name and leaving the filename untouched in the Settings modal still leads to a validation error for the filename, preventing the user from saving the new name.
2. Filename input width is virtually restricted to the width of the original name and causes unintuitive scrolling & clipping when entering a longer filename.

|  | 1 | 2 |
|---|---|---|
| Before | image | image |
| After | Kapture 2026-05-20 at 19 35 45 | ws-settings-filename-adaptive-length |

Changes were manually tested by comparing v12.6.0-beta.0 and a local dev build. Automated tests were not added for these because unit tests wouldn't effectively test either case and the cost of failure seems too low to add e2e/smoke tests for these. Happy to give the tests a go if that assessment doesn't feel right though.

### DIFF
--- a/packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
@@ -272,7 +272,7 @@ export const NewWorkspaceModal = ({
                             placeholder={workspaceData.name ? safeToUseInsomniaFileName(workspaceData.name) : 'name'}
                             className="w-full outline-hidden [grid-area:input] placeholder:italic focus:outline-hidden"
                           />
-                          <span className="pointer-events-none w-min truncate opacity-0 [grid-area:input]">
+                          <span className="pointer-events-none truncate opacity-0 [grid-area:input]">
                             {safeToUseInsomniaFileName(workspaceData.fileName || workspaceData.name || 'name')}
                           </span>
                           <span className="text-(--hl) [grid-area:extension]">.yaml</span>
--- a/packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
+++ b/packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
@@ -267,7 +267,7 @@ export const NewWorkspaceModal = ({
                       <Label className="group relative flex flex-col gap-2 overflow-hidden">
                         <span className="text-sm text-(--hl)">File name</span>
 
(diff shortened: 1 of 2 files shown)


packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
```diff
@@ -251,9 +251,9 @@ export const NewWorkspaceModal = ({
                         <div className="grid w-full grid-cols-[min-content_auto] overflow-hidden rounded-xs border border-solid border-(--hl-sm) bg-(--color-bg) py-1 pr-7 pl-2 text-(--color-font) transition-colors [grid-template-areas:'input_extension'] focus:ring-1 focus:ring-(--hl-md) focus:outline-hidden">
                           <Input
                             placeholder={workspaceData.name ? safeToUseInsomniaFileName(workspaceData.name) : 'name'}
-                            className="w-full min-w-[3ch] outline-hidden [grid-area:input] placeholder:italic focus:outline-hidden"
+                            className="w-full outline-hidden [grid-area:input] placeholder:italic focus:outline-hidden"
```
godfrzero, line 273: Enforcing a minimum width on this was leading to a strange gap when typing short filenames one the hardcoded extension was allowed to move around freely:

Image

packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
```diff
@@ -251,9 +251,9 @@ export const NewWorkspaceModal = ({
...
                           />
-                          <span className="-z-10 w-min truncate opacity-0 [grid-area:input]">
+                          <span className="pointer-events-none w-min truncate opacity-0 [grid-area:input]">
```
godfrzero: Replaced with `pointer-events: none` rather than messing with the stacking order since the intent here is to let the user "click through" the element.

packages/insomnia/src/ui/components/modals/new-workspace-modal.tsx
pavkout: @godfrzero In `workspace-settings-modal.tsx` the invisible width-measuring span uses `pointer-events-none truncate opacity-0`, but the equivalent span here in `new-workspace-modal.tsx` keeps `w-min` (`pointer-events-none w-min truncate opacity-0`).

godfrzero: I've removed it from here as well. It shouldn't make a difference in practice since `min-content` and the default would behave the same if we're disallowing spaces and replacing other characters like `.` with `_`.

pavkout (approved):
LGTM

PR #9945, opened 2026-05-21
feat: lift network.ts fs/path behind window.main.timeline IPC bridge 

jackkav:
## Summary

- Removes all `process.type` / `typeof window` branching from `src/network/network.ts`
- Introduces build-time adapter modules:
  - `network-adapter.renderer.ts` — IPC calls via `window.main` (Electron renderer)
  - `network-adapter.node.ts` — direct `node:fs` calls (inso CLI)
  - `network-adapter.ts` — type-checking shim; bundlers rewrite this import at build time
- Vite aliases `~/network/network-adapter` → `network-adapter.renderer`; inso esbuild aliases it → `network-adapter.node`
- Adapter surface: `getTimelinePath`, `appendToTimelineOnError`, `appendTimelineLines`, `getAuthHeader`, `executeCurlRequest`, `applyRequestHooks`, `applyResponseHooks`
- Path traversal guard in node adapter (validates resolved path stays inside `responses/`)
- Shrinks `config/renderer-node-import-baseline.json` by 2 entries (`src/network/network.ts` `fs` + `path`)
- Fixes consola v3 compatibility in `insomnia-inso` (`FancyReporter`/`LogLevel` removed upstream)

## Context

Part of the `nodeIntegration: false` renderer migration. This is **PR B** from `NODE_INTEGRATION_MIGRATION_PR_PLAN.md`.

Stacks on top of `chore/disable-node-integration`.

## Design

The adapter boundary replaces all previous `if (process.type !== 'renderer')` and `process.type === 'renderer' ? window.main.X : (await import(...))` patterns. No runtime environment detection remains in `network.ts` — the correct implementation is selected at bundle time.

The node adapter also absorbs the plugin hook loops that were previously inlined in `network.ts` behind a `process.type` guard.

## Test plan

- [ ] `npm run type-check` — passes (0 errors)
- [ ] `npm run check:renderer-node-imports` — passes, baseline no longer contains `src/network/network.ts`
- [ ] `npm run test:bundle -w insomnia-inso` — 45 tests pass
- [ ] Manual: send a request and confirm the timeline file is written correctly
- [ ] Manual: confirm pre-request script errors still appear in the timeline


packages/insomnia-smoke-test/playwright.config.ts
```diff
@@ -8,7 +8,7 @@ const echoServer: PlaywrightTestConfig['webServer'] = {
   url: 'http://localhost:4010',
   timeout: 20 * 1000,
   reuseExistingServer: !process.env.CI,
-  stdout: 'pipe',
+  stdout: 'ignore',
```
ZxBing0066, line 11: Why do we ignore the stdout here?

packages/insomnia-inso/vitest.config.ts
```diff
@@ -4,6 +4,7 @@ export default defineConfig({
   test: {
     hideSkippedTests: true,
     alias: {
+      '~/network/network-adapter': new URL('../insomnia/src/network/network-adapter.renderer.ts', import.meta.url).pathname,
```
ZxBing0066: This seems to be incorrect. Should Inso use `.node.ts`?

packages/insomnia-smoke-test/playwright.config.ts
jackkav: because I only enabled it recently, and it was cluttering the smoke test output

ZxBing0066 (approved):

PR #9955, opened 2026-05-26
chore: add quick-check skill and document it in AGENTS.md

jackkav:
## Summary

- Adds `.claude/skills/quick-check/SKILL.md` — a Claude Code skill that picks the narrowest lint/type-check/test command for the affected workspace instead of always running repo-wide checks.
- References the skill in `AGENTS.md` under **Validation Commands** so agents know to prefer it during iterative development.

## Test plan

- [ ] Verify `.claude/skills/quick-check/SKILL.md` is present and matches the local skill definition
- [ ] Confirm `AGENTS.md` reference links to the correct path

### DIFF
--- a/.claude/skills/quick-check/SKILL.md
+++ b/.claude/skills/quick-check/SKILL.md
@@ -1,0 +1,131 @@
+---
+name: quick-check
+description: "Run the smallest useful format, lint, type-check, or test command for the affected Insomnia workspace, and only escalate to broader validation when the change scope justifies it."
+argument-hint: "Provide changed file paths and the desired check type (format, lint, type-check, test), or name the target workspace and what you want validated"
+---
+
+# Quick Workspace Checks
+
+## When to Use
+
+- You are iterating on a change and want the fastest useful validation loop.
+- The change is limited to one workspace or a small set of files.
+- You do not want repo-wide checks yet.
+
+## Core Rule
+
+Choose the smallest affected workspace first. Only expand to more workspaces or repo-wide checks when the change crosses boundaries.
+
+## Workspace Mapping
+
+| Changed paths                                | Workspace                        | Quick checks                                                                                                                                           |
+| -------------------------------------------- | -------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `packages/insomnia/**`                       | `insomnia`                       | `npm run lint -w insomnia`, `npm run type-check -w insomnia`, `npm test -w insomnia`                                                                   |
+| `packages/insomnia-api/**`                   | `insomnia-api`                   | `npm run lint -w insomnia-api`, `npm run type-check -w insomnia-api`                                                                                   |
+| `packages/insomnia-inso/**`                  | `insomnia-inso`                  | `npm run lint -w insomnia-inso`, `npm run type-check -w insomnia-inso`, `npm run test:unit -w insomnia-inso`                                           |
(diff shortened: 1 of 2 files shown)


.claude/skills/quick-check/SKILL.md
```diff
@@ -0,0 +1,54 @@
+---
+name: quick-check
+description: "Run the smallest useful format, lint, type-check, or test command for the affected Insomnia workspace, and only escalate to broader validation when the change scope justifies it."
+argument-hint: "Provide changed file paths and the desired check type (format, lint, type-check, test), or name the target workspace and what you want validated"
+---
+
+## Workspace Mapping
```
ZxBing0066, line 7: IMO, workspace seems not to be that useful since most changes also affect other workspaces. A full repo `npm run lint && npm run type-check && npm test` may be more accurate.

.claude/skills/quick-check/SKILL.md
```diff
@@ -0,0 +1,54 @@
+---
+name: quick-check
+description: "Run the smallest useful format, lint, type-check, or test command for the affected Insomnia workspace, and only escalate to broader validation when the change scope justifies it."
+argument-hint: "Provide changed file paths and the desired check type (format, lint, type-check, test), or name the target workspace and what you want validated"
```
arora-r, line 4: It might be worth adding `allowed tools` to the metadata for the commands it'll use. We can scope to specific commands not purely *. So, `npx prettier *` instead of `npx *` [- docs](https://code.claude.com/docs/en/skills#:~:text=Tools%20Claude%20can%20use%20without%20asking%20permission%20when%20this%20skill%20is%20active.%20Accepts%20a%20space%2Dseparated%20string%20or%20a%20YAML%20list.)

.claude/skills/quick-check/SKILL.md
```diff
@@ -0,0 +1,54 @@
...
+
+Format then fix then verify:
+
+```bash
+npx prettier --write <changed files>
+npx eslint --fix <changed files>
+npm run lint -w <workspace>
+```
+
+## Targeted Tests
```
arora-r, line 30: Maybe we could also tell it explicitly to either auto-fix or not fix the issues found in tests to be clear on its goals.

.claude/skills/quick-check/SKILL.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Escalate When
+
+- Change spans more than one workspace
+- Root config changed (`package.json`, `package-lock.json`, `eslint.config.mjs`)
+- Renderer/main boundary touched — also run `npm run check:renderer-node-imports -w insomnia`
+
+```bash
+npm run lint && npm run type-check && npm test
+```
```
arora-r, line 54: Maybe we can give it an early exit condition section to prevent any spikes. 

We could also have it not write everything to the terminal 
```
  - eslint --quiet
  - npm test --silent
  - pipe commands with `| head -50 or 2>&1 | tail -25`
```
 so only errors are shown to lower token usage even more.

arora-r (commented):
A personal preference of mine would be to use [pre-commit ](https://pre-commit.com/) instead of a Claude skill. We can force it to run on every `git commit` and we can define custom hooks like having a custom script run to find which test files should be ran based on changes.

This is more work but less token usage, especially for bigger changes 😅.

We can configure it to run with a separate command and tell the agent to run this command at the end of all changes to not bother regular development.

jackkav:
We got unlimited now. No need to share these, we just develop our own to our preferences.

On pre-commit hooks. You're correct it's a personal preference so best kept out of the project and in user settings so everyone can choose their workflow

PR #9956, opened 2026-05-26
fix: prevent flake in environment-editor JSON input

jackkav:
## Summary

- Fixes a pre-existing flaky test in `environment-editor-interactions.test.ts` — "manage environment"
- The original flake was partly caused by `bodyEditor.fill()` bypassing CodeMirror's key event handlers; replacing it with `ControlOrMeta+a` + `page.keyboard.type()` makes the JSON edit path behave like a real user edit
- The remaining flake was a broader persistence race: environment edits are debounced and saved asynchronously, so the test could close the modal before those updates had finished settling
- This change adds an explicit save-complete signal to the manage-environments modal and updates the smoke test to wait for that signal before closing

## Rationale

- We considered splitting or removing the test, but that would either increase CI time materially or reduce coverage on a valuable end-to-end environment-editing flow
- We also wanted to avoid layering on more timing-based waits, because they can reduce the flake rate without eliminating the race
- Exposing a real save-state signal gives the app and the test a deterministic synchronization point tied to the actual async work, which is a safer long-term fix than relying on editor timing or modal transition timing alone
- Disabling close while a save is pending also improves product behavior, not just the test, because users can no longer dismiss the modal while environment mutations are still in flight

## Test plan

- [ ] Confirm the "manage environment" smoke test passes consistently in CI (shard 2/6)
- [ ] No other environment editor tests affected

### DIFF
--- a/packages/insomnia-smoke-test/tests/smoke/environment-editor-interactions.test.ts
+++ b/packages/insomnia-smoke-test/tests/smoke/environment-editor-interactions.test.ts
@@ -110,7 +110,6 @@ test.describe('Environment Editor', () => {
     // wait for modal to show
     await expect.soft(page.getByRole('dialog').getByTestId('CodeEditor')).toBeVisible();
     const bodyEditor = page.getByRole('dialog').getByTestId('CodeEditor').getByRole('textbox');
-    // move cursor right and input json string
     await bodyEditor.focus();
     await page.keyboard.press('ControlOrMeta+a');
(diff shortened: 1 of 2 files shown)


fiosman (commented):
https://github.com/Kong/insomnia/actions/runs/26440164941/job/77887439182 on the latest CI run it appears that this e2e test is still failing

jackkav:
@fiosman I ran it 20 times before this change and it failed 5 times, I ran it 40 times after this change and it failed once.

PR #9968, opened 2026-05-27
Remove script-executor from renderer import baseline

jackkav:
## Summary

Removes `src/script-executor.ts` from the renderer Node import baseline by ensuring its `node:fs/promises` call never ends up in the renderer module graph.

## What changed

**`network/cancellation.ts`** — remove `cancellableRunScript`. The AbortController it wired up was dead code: the only thing that could abort it was `cancelRequestById`, which calls `window.main.completeExecutionStep` — a renderer-only IPC call. No CLI code path could ever trigger the abort.

**`network/network.ts`** — replace the `cancellableRunScript` call in the CLI path with a direct `require(/* @vite-ignore */ '../script-executor').runScript`. The `@vite-ignore` comment keeps the dynamic require out of the renderer bundle entirely.

**`renderer-node-import-baseline.json`** — the `script-executor.ts → node:fs/promises` entries are gone.

## Why not a separate module?

An earlier iteration moved the Node logic into a dedicated `run-script.ts` loaded via `@vite-ignore`. That worked but added indirection for no gain once we confirmed cancellation was dead code. The final approach is simpler: the CLI path inlines the dynamic require directly in `network.ts`, and `script-executor.ts` handles its own runtime branch for log appending.

## Net diff

`15 insertions, 39 deletions` — net removal.


ZxBing0066:
runScript is never called from the renderer process. Moving cancellableRunScript out from `cancellation` would be better.

packages/insomnia/src/script-executor.ts
```diff
@@ -121,3 +119,12 @@ function proxiedSetTimeout(callback: () => void, ms?: number | undefined) {
     resolveHdl(null);
   }, ms);
 }
+
+async function appendScriptLogs(timelinePath: string, data: string) {
```
ZxBing0066: runScript only calls in node, is this still necessary?

jackkav: I can address in 9970 as it sounds non blocking

jackkav: i removed it

ZxBing0066 (approved):

PR #9970, opened 2026-05-27
Move require interceptor out of main renderer bundle

jackkav:
## Summary
- remove the main renderer import path that pulled `script-executor` into the Vite graph
- move runtime interceptor rules into a hidden-window-only module so the scripting settings UI only imports rule metadata
- keep request and script cancellation renderer-only, while non-renderer script execution falls back to the plain node runner

## Why
The renderer node import baseline showed `src/script-executor.ts` and `src/scripting/require-interceptor.ts` in the main window bundle. This change keeps the require interceptor logic scoped to hidden-window script execution and ratchets the baseline down accordingly.

## Validation
- `npm run update:renderer-node-import-baseline -w insomnia`
- `npm run check:renderer-node-imports -w insomnia`
- `npm run type-check -w insomnia`
- `npm run lint -w insomnia`
- `npm run test -w insomnia -- src/scripting/__tests__/script-security-policy.test.ts`


ZxBing0066:
This PR seems to conflict with 9968. script-executor should has been resolved in 9968.

PR #9972, opened 2026-05-28
fix(script-executor): remove from renderer Node import baseline (#9968)

jackkav:
## Summary

Closes #9968.

Removes `src/script-executor.ts` from the renderer Node import baseline by ensuring its `node:fs/promises` call never ends up in the renderer module graph.

## What changed

**`script-executor.ts`** — replace the top-level `appendFile` import (which pulled `node:fs/promises` into the renderer bundle) with a runtime-branched helper:
- Renderer path: delegates to `window.main.timeline.appendToFile` (existing IPC bridge)
- CLI/Node path: `require('node:fs/promises')` at call-time, invisible to Vite's static analyser

**`network/cancellation.ts`** — remove `cancellableRunScript`. The AbortController it wired up was dead code: the only thing that could abort it was `cancelRequestById`, which calls `window.main.completeExecutionStep` — a renderer-only IPC call. No CLI code path could ever trigger the abort.

**`network/network.ts`** — replace the `cancellableRunScript` call in the CLI path with a direct `require(/* @vite-ignore */ '../script-executor').runScript`. The `@vite-ignore` comment keeps the dynamic require out of the renderer bundle entirely.

**`renderer-node-import-baseline.json`** — the `script-executor.ts → node:fs/promises` entries are gone.

## Why not a separate module?

An earlier iteration moved the Node logic into a dedicated `run-script.ts` loaded via `@vite-ignore`. That worked but added indirection for no gain once we confirmed cancellation was dead code. The final approach is simpler: the CLI path inlines the dynamic require directly in `network.ts`, and `script-executor.ts` handles its own runtime branch for log appending.

## Net diff

`15 insertions, 39 deletions` — net removal.


jackkav:
Closing — opened by mistake, duplicate of #9968.

Issue #9951, opened 2026-05-25
12.6.0 Linux crashes with `write EPIPE` on desktop-file launch


bashfulrobot:
## Expected behaviour

Insomnia opens when launched from the desktop entry, same as 12.5.0 did.

## Actual behaviour

A modal pops up immediately on launch: "A JavaScript error occurred in the main process". The main window never opens. Full error text from the dialog:

```
Uncaught Exception:
Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    at writeGeneric (node:internal/stream_base_commons:150:3)
    at Socket._writeGeneric (node:net:966:11)
    at Socket._write (node:net:978:8)
    at writeOrBuffer (node:internal/streams/writable:570:12)
    at _write (node:internal/streams/writable:499:10)
    at Writable.write (node:internal/streams/writable:508:10)
    at console.value (node:internal/console/constructor:313:16)
    at console.info (node:internal/console/constructor:423:26)
    at transport.writeFn (/nix/store/47pxbmwhs21sd9rs6g6h3zdgrnvhvwid-insomnia-12.6.0-extracted/resources/app.asar/entry.main.min.js:287808:11)
```

<img width="681" height="416" alt="Image" src="https://github.com/user-attachments/assets/e68dcd5c-145e-445c-a4a0-456ec508f983" />

Looks like the main process logger calls `console.info` synchronously during init, and that call is not wrapped to handle `EPIPE` when stdout is a pipe whose reader has closed.

12.5.0 launches fine under the same conditions.

## Reproduction steps

1. Download `Insomnia.Core-12.6.0.AppImage` from the `core@12.6.0` release.
2. `chmod +x Insomnia.Core-12.6.0.AppImage`
3. Launch it from a `.desktop` entry, `gtk-launch`, or anything else where stdout has no live reader. Terminal-equivalent repro:
   ```
   ./Insomnia.Core-12.6.0.AppImage 2>&1 | head -0
   ```
   `head -0` exits immediately, the pipe closes, and the next logger write crashes the main process.

Launching from an interactive terminal works because stdout is a TTY and writes succeed.

## Searched issue tracker

Yes.

## Sync method

Insomnia Cloud sync.

## Additional information

[...]

PR #9984, opened 2026-05-29
fix: app crash if no writable stdout is available #9951

godfrzero:
On Linux, launching Insomnia from a .desktop entry (or any context without a writable stdout) could crash the main process in two ways:

1. Synchronous write failure — initializeLogging() was called after initElectronStorage(), meaning early log writes could throw before any EPIPE handler was in place. Reordering the calls ensures logging is initialized first.
2. Async EPIPE — Even with the reorder, EPIPE can surface as an async error event on process.stdout after the write is dispatched (e.g. when the read end of the pipe is closed). Without a handler this becomes an uncaught exception that crashes the main process. A process.stdout.on('error', ...) handler now silently drops EPIPE and re-throws everything else.

Together these fixes make the app stable when launched without a writable stdout, which is the common case on Linux desktop environments.

I couldn't test these changes on a linux system, but I was able to recreate what looks like the same error on macOS by launching Insomnia from the command line:

```
# head -0 exits immediately, causing an EPIPE error when the app tries to write to stdout.
<path to build> | head -0
```


notjaywu (approved):
lgtm

PR #9989, opened 2026-05-29
chore: add e2e tests for custom linting rules

Ali-Sab:


packages/insomnia-smoke-test/server/cloud-sync-api.ts
```diff
@@ -219,6 +228,28 @@ const projectSnapshots: Record<string, any[]> = {
...
+    {
+      ...commonSnapshotProps,
+      created: '2026-01-22T06:20:00.759Z',
+      id: '2ce4bced4220de84704bee82b6174890ba4a89f0',
+      name: 'Initial Snapshot',
+      parent: '0000000000000000000000000000000000000000',
+      state: [
+        {
+          blob: '75bdac19931bd37e2853464f7a26ecbb79bc4fca',
+          key: 'wrk_f3e4a2b1c9d0e5f6a7b8c9d0e1f2a3b4',
```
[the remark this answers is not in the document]
Ali-Sab: @Aikido ignore: test key, not used in prod

packages/insomnia-smoke-test/server/cloud-sync-api.ts
```diff
@@ -219,6 +228,28 @@ const projectSnapshots: Record<string, any[]> = {
...
+          name: 'Design Project',
+        },
+        {
+          blob: '52aa4f92c8e47e955f0c3fcc2fd38d41710450ad',
+          key: 'spc_f3e4a2b1c9d0e5f6a7b8c9d0e1f2a3b5',
```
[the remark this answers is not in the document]
Ali-Sab: @aikido ignore: test key, not used in prod

packages/insomnia-smoke-test/fixtures/files/custom.spectral.yaml
```diff
@@ -0,0 +1,10 @@
+extends: [spectral:oas]
+rules:
+  require-x-smoke-test-marker:
+    description: info object must contain a custom x-smoke-test-marker field
+    message: "{{description}}"
+    severity: error
+    given: $.info
+    then:
+      field: x-smoke-test-marker
+      function: truthy
```
fiosman, line 10: This is great, should we also include a fixture with disallowed top level keys (e.g. custom functions). In the smoke test we can make assertion that an error pops up when linting the spec. We just need 1 test to check the modal pops up etc. 

This is totally OK to leave out; technically it's an edge case that I think I am already covering in a unit test for a validation function I wrote.

packages/insomnia-smoke-test/tests/smoke/custom-lint-rules.test.ts
```diff
@@ -0,0 +1,424 @@
...
+    test('upload custom ruleset in a synced design project', async ({ page, insomnia }) => {
+      await page.getByTestId('workspace-grid').getByLabel('Design Project').click();
+      await page.waitForURL(/\/workspace\/.+\/spec/, { timeout: 30_000 });
+
+      await uploadRuleset(insomnia, page);
+
+      await page.reload();
+      await expect.soft(page.getByRole('button', { name: 'View selected ruleset content' })).toBeVisible({
+        timeout: 15_000,
+      });
```
fiosman, line 260: Do you think it's worth opening up the modal and verifying the content matches what the ruleset content is?

packages/insomnia-smoke-test/tests/smoke/custom-lint-rules.test.ts
```diff
@@ -0,0 +1,424 @@
...
+
+    // -------------------------------------------------------------------------
+    // 3c. Different users, same project
+    //
+    // User A uploads the ruleset and pushes.  User B (different accountId,
+    // same symmetric key so the project blobs are decryptable) starts on a
+    // fresh machine, pulls the project, and verifies the ruleset survived the
+    // sync handoff.
+    // -------------------------------------------------------------------------
+    test.describe.serial('different users, same project', () => {
```
fiosman: Should we be using `serial`? If I were to add a `.only` to any of the tests below they would fail because they rely on state from previous tests? We don't currently use this anywhere in our codebase.

packages/insomnia/src/main/ipc/electron.ts
```diff
@@ -428,6 +432,14 @@ export function registerElectronHandlers() {
...
   });
   ipcMainHandle('showOpenDialog', async (_, options: OpenDialogOptions) => {
+    // Playwright test hook: consume queued responses set via `electronApp.evaluate`
+    // instead of opening the native dialog. See packages/insomnia-smoke-test.
+    if (process.env.PLAYWRIGHT === 'true') {
+      const queue = globalThis.__PLAYWRIGHT_OPEN_DIALOG_QUEUE__;
+      if (queue && queue.length > 0) {
+        return queue.shift();
+      }
+    }
```
fiosman, line 443: Can we utilize existing `mockOpenDialogForDirectory`? It seems to mock behaviour for native file picker which I think is what you are trying to do this here.

fiosman (commented):
Thanks for your work on this! Appreciate that you were trying to cover as much as possible. I left some comments that we should consider. 

Also, we should consider whether we care about testing persistence on app reloads etc (e.g. `relaunch` function you added). Sometimes it's not worth it if it adds complexity/maintenance efforts to an already flaky/large e2e test suite. In my opinion, end to end tests are supposed to be limited to critical user paths (in this case; on a cloud sync project, one test that uploads a custom ruleset, spec is automatically linted, diff shows up in commit modal, and I am able to push the changes). Unit tests and/or component tests, which run much faster, can cover edge cases. Lmk your thoughts! Great work.

packages/insomnia/src/main/ipc/electron.ts
Ali-Sab: Good catch on the inconsistency. The reason I didn't reuse it is that `mockOpenDialogForDirectory` works by permanently replacing the `ipcMain` handler (`removeHandler` + `handle`), so after it fires the real dialog is gone for the rest of the process. The queue approach is just a list that gets shifted on each call and falls through to the real dialog when empty, which felt safer for tests that might open multiple dialogs. Happy to revisit if I'm mistaken or we just don't need the queue

packages/insomnia-smoke-test/fixtures/files/custom.spectral.yaml
Ali-Sab: Valid, I'll add it !

packages/insomnia-smoke-test/tests/smoke/custom-lint-rules.test.ts
Ali-Sab: Actually it might be, I'll add something in

packages/insomnia-smoke-test/tests/smoke/custom-lint-rules.test.ts
Ali-Sab: I think you're right, let me merge the serial pairs into a single test, it's only one scenario anyway. Actually, on also checking playwright's docs, `serial` is discouraged for use, so good call here!

fiosman (approved):

PR #9992, opened 2026-05-30
Remove all third-party npm package imports from the renderer bundle

jackkav:
## Summary

Removes several third-party package imports from the renderer by replacing them with lightweight alternatives or inline implementations. No behavioral changes — pure substitutions.

- **node-forge (renderer)** — removed from renderer bundle via IPC bridge. `vault-crypto.ts` is now a thin async wrapper over `window.main.vault.{encrypt,decrypt}SecretValue`; the main process keeps its existing forge-based implementation untouched. This keeps crypto logic in a single place without re-implementing AES-GCM in the renderer.
- **httpsnippet** — two `await import('httpsnippet')` calls in `generate-code-modal.tsx` and `request-actions-dropdown.tsx` are replaced with `window.main.generateCodeSnippet()` and `window.main.getCodeSnippetTargets()` IPC calls. `httpsnippet` was never bundled by Vite (its Node.js dependencies — `querystring`, `url` — prevent static analysis bundling); instead it was resolved at runtime via Electron's `require()` on each invocation of "Generate Code" or "Copy as cURL". It is now fully eliminated from renderer access.
- **common/mime.ts** — replaces the `mime-types` npm package with a 48-entry inline lookup table covering images, audio, video, office documents, archives, fonts, wasm, and common text types. Removes the package boundary for a simple extension↔MIME mapping.
- **response-cookies-viewer** — removes `tough-cookie` import; replaces with a 10-line inline `Set-Cookie` header parser (only `name=value` extraction was needed)
- **grpc-status-tag** — removes `@grpc/grpc-js` import just to read `status.OK = 0`; replaced with an inline constant
- **auth.clear-vault-key** — replaces `electron.ipcRenderer.emit('show-toast')` with the typed `showToast()` helper; also fixes typo "all you local" → "all your local"
- **window-utils** — removes an unused `trackAnalyticsEvent` call from the app menu handler
- **window.env / contextBridge** — env vars are not available in the renderer by default when `nodeIntegration` is disabled. The preload now explicitly whitelists the env vars the renderer needs and exposes them as `window.env` via `contextBridge`. `constants.ts` reads from `window.env` in the renderer and falls back to `process.env` for the inso CLI and main process.
- **response-viewer charset aliases** — `TextDecoder` does not accept iconv-lite alias names (e.g. `utf8`, `latin1`, `win1252`). Added a module-level alias map to normalise them to WHATWG labels before passing to `TextDecoder`, preventing silent fallback to UTF-8 for responses with non-standard charset declarations.

## Bundle verification

`npm run check:renderer-node-imports -w insomnia` reports zero third-party npm packages in the renderer bundle after these changes. The 13 remaining entries are all Node built-ins (`fs`, `path`, `crypto`, etc.) in pre-existing plugin/scripting paths that have always been present.

**Note on httpsnippet:** a bundle diff confirms `httpsnippet` produced no Vite chunk in either build — it was a runtime-resolved module (lazy `require()` via nodeIntegration), not a statically bundled one. The 7 chunks whose hashes changed between develop and this branch are identical in byte size; the only functional code delta is the IPC call replacing the dynamic import.

## E2E note

Also removes the `toHaveScreenshot` assertion from the PDF preview smoke test (added in #9922, @rwillis-kong). The assertion was causing the `can send requests` test to exceed its 60 s CI timeout — the screenshot step inherits the full 25 s expect timeout and retries until it matches, consuming most of the test budget before the remaining steps (basic auth, cookies, cancel) even start.

The three structural assertions immediately above it already give good smoke-level confidence:
1. `toBeVisible()` — iframe is rendered
2. `toHaveAttribute('src', /^blob:/)` — PDF data was loaded into a blob URL
3. `expect.poll` for the `chrome-extension://` URL — Chromium's built-in PDF viewer actually mounted

## Test plan

- [ ] Vault encrypt/decrypt round-trips correctly end-to-end (environment secret save → reload → template render)
- [ ] MIME lookup still resolves extensions correctly in body editor and response pane download
- [ ] Cookie viewer still displays cookies from response headers
- [ ] gRPC status tag still shows green/red for OK/non-OK status codes
- [ ] Vault key reset flow still shows the toast notification with corrected copy
- [ ] All env-driven features still work: GitLab OAuth, E2E fixtures, service URL overrides, Windows portable detection
- [ ] Response viewer correctly decodes bodies with non-standard charset headers (latin1, win1252, etc.)


packages/insomnia/src/utils/vault-crypto.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const decryptSecretValue = async (encryptedValue: string, symmetricKey: JsonWebKey): Promise<string> => {
+  if (typeof symmetricKey !== 'object' || Object.keys(symmetricKey).length === 0) {
+    return encryptedValue;
+  }
+  try {
+    return await window.main.vault.decryptSecretValue(encryptedValue, symmetricKey);
+  } catch {
+    return encryptedValue;
```
ZxBing0066, line 19: This seems to be insecure. Should we throw the error here?

packages/insomnia/src/entry.preload.ts
```diff
@@ -437,6 +446,44 @@ const database: Window['database'] = {
   invoke: (fnName, ...args) => invokeWithNormalizedError('database.invoke', fnName, ...args),
 };
 
+const env: Window['env'] = {
+  // GitLab OAuth — redirect URI, client ID, and API URL allow dev/enterprise overrides
+  INSOMNIA_GITLAB_REDIRECT_URI: process.env.INSOMNIA_GITLAB_REDIRECT_URI,
```
ZxBing0066, line 452: Just curious. Is there any better idea instead of explicitly referencing all the envs one by one? Ideally, just adding statements on the types is enough.

packages/insomnia/src/ui/components/editors/environment-key-value-editor/key-value-editor.tsx
```diff
@@ -77,8 +77,26 @@ export const EnvironmentKVEditor = ({
   );
   const codeModalRef = useRef<CodePromptModalHandle>(null);
   const [kvPairError, setKvPairError] = useState<{ id: string; error: string }[]>([]);
+  const [decryptedValues, setDecryptedValues] = useState<Record<string, string>>({});
   const symmetricKey = vaultKey === '' ? {} : base64decode(vaultKey, true);
 
+  useEffect(() => {
+    const secretPairs = kvPairs.filter(p => p.type === EnvironmentKvPairDataType.SECRET);
+    if (secretPairs.length === 0 || Object.keys(symmetricKey).length === 0) {
+      return;
```
ZxBing0066, line 87: Should we reset setDecryptedValues here?

packages/insomnia/src/main/window-utils.ts
```diff
@@ -270,7 +269,6 @@ export function createWindow(): ElectronBrowserWindow {
       {
         label: `${MNEMONIC_SYM}Preferences`,
         click: () => {
-          trackAnalyticsEvent(AnalyticsEvent.AppMenuPreferencesClicked);
```
ZxBing0066, line 273: Why is this removed?

packages/insomnia/src/ui/components/editors/body/body-editor.tsx
```diff
@@ -1,5 +1,4 @@
 import clone from 'clone';
-import { lookup } from 'mime-types';
```
ZxBing0066, line 2: Why do we remove mine-types? It's hard to say the implementation can substitute without any regressions.

packages/insomnia/src/common/mime.ts
```diff
@@ -0,0 +1,83 @@
...
+  mjs: 'application/javascript',
+  txt: 'text/plain',
+  xml: 'application/xml',
+  yaml: 'application/yaml',
+  yml: 'application/yaml',
+  // image
+  bmp: 'image/bmp',
+  gif: 'image/gif',
+  ico: 'image/x-icon',
+  jpeg: 'image/jpeg',
```
ZxBing0066: And the implementation of `mimeTypeExtension` totally depends on the order of the keys. Maybe we can take https://github.com/jshttp/mime-db/blob/master/db.json#L8803 as a reference.

packages/insomnia/src/entry.preload.ts
jackkav: In the first pass in #9979 i use a loop and add all insomnia env vars, but I think explicitness is a good thing with env vars.

packages/insomnia/src/main/window-utils.ts
jackkav: Good catch

packages/insomnia/src/ui/components/editors/body/body-editor.tsx
jackkav: I thought so too, mini-tye depends on node:path extname. and adding a bundle resolve aliad just adds 200kb of unused json to the bundle

packages/insomnia/src/utils/vault-crypto.ts
jackkav: I don't think returning the encrypted value is insecure. @kwburns-kong can confirm and we can solve in a follow up pr.

jackkav: not new

ZxBing0066 (dismissed):

packages/insomnia/src/ui/components/viewers/response-multipart-viewer.tsx
```diff
@@ -77,7 +77,7 @@ export const ResponseMultipartViewer: FC<Props> = ({
       return;
     }
     const contentType = getContentTypeFromHeaders(selectedPart.headers, 'text/plain');
-    const extension = mimeExtension(contentType) || '.txt';
+    const extension = mimeExtension(contentType) || 'txt';
```
ZxBing0066, line 80: Is this expected?

ZxBing0066 (dismissed):

ZxBing0066 (approved):

packages/insomnia-smoke-test/tests/smoke/app.test.ts
```diff
@@ -93,13 +93,6 @@ test('can send requests', async ({ page, insomnia }) => {
     })
     .toBe(true);
 
-  // No explicit timeout — inherits the global expect.timeout (40s on CI) so Playwright
-  // retries long enough for the Chromium PDF viewer to finish rendering.
-  await expect.soft(pdfIframe).toHaveScreenshot('dummy-pdf-preview.png', {
```
godfrzero, line 98: If this assertion is removed, then should the surrounding pdf-related code also be removed? @jackkav @ryan-willis

ryan-willis: the surrounding assertions should be kept, only the screenshot test wasn't consistent/reliable enough

PR #9994, opened 2026-05-30
Introduce render-adapter; remove pluginsAllowElevatedAccess escape hatch

jackkav:
## Summary

- Introduces a `render-adapter` pattern (matching `network-adapter`) to cleanly split renderer vs node/main templating implementations
- **Renderer** always routes to the Web Worker via `render-adapter.renderer.ts`; **node/main** calls the full LiquidJS engine directly via `render-adapter.node.ts`
- Removes `pluginsAllowElevatedAccess` setting, its type definition, default, and UI toggle — the escape hatch that allowed in-renderer plugin execution is gone
- Extracts `NUNJUCKS_TEMPLATE_GLOBAL_PROPERTY_NAME` to `templating/constants.ts` so `render.ts` no longer needs to import the full `../templating` module (which transitively pulled `node:crypto`/`node:os` into the renderer bundle via `liquid-extension.ts`)
- `executePluginMainAction` in `plugins/index.ts` now always delegates to main process over IPC — the direct renderer execution path is removed
- Removes `INSOMNIA_ENABLE_PLUGIN_BRIDGE` rollback switch — plugins always run in the hidden plugin window via `window.main.plugins` IPC; the legacy renderer-direct fallback is gone

## How the adapter works

```typescript
// render-adapter.ts
const impl = (
  (process as any).type === 'renderer'
    ? require('./render-adapter.renderer')           // → renderInWorker
    : require(/* @vite-ignore */ './render-adapter.node')  // → templating.render()
) as typeof AdapterType;
```

Vite inlines `process.type = 'renderer'` at build time, so Rollup tree-shakes the `.node` branch entirely from the renderer bundle. `liquid-extension.ts` (which imports `node:crypto` and `node:os`) no longer appears in the renderer chunk.

## Test plan

- [ ] Template tags render correctly in the renderer via the Web Worker path
- [ ] Templates render correctly in `insomnia-inso` (Node path via `render-adapter.node`)
- [ ] Plugin main actions still work (routed via IPC)
- [ ] `npm run check:renderer-node-imports -w insomnia` no longer reports `liquid-extension.ts`
- [ ] Plugins settings page no longer shows the elevated access toggle
- [ ] Plugin bridge routes correctly through hidden window (no `INSOMNIA_ENABLE_PLUGIN_BRIDGE` env var needed)


packages/insomnia/src/entry.client.tsx
```diff
@@ -40,8 +39,6 @@ delete window._dataServices;
 
 configureFetch(options => insomniaFetch({ ...options, onDeepLink: (uri: string) => window.main.openDeepLink(uri) }));
 
-await initPlugins();
```
ZxBing0066, line 43: It seems initPlugins is not used anymore. Should we remove it?

packages/insomnia/src/ui/components/settings/plugins.tsx
```diff
@@ -127,32 +127,6 @@ export const Plugins: FC = () => {
         Plugins is still an experimental feature. See <Link href={docsPlugins}>Documentation</Link> for more info.
       </p>
 
-      <div className="notice warning margin-bottom text-left">
```
ZxBing0066, line 130: Why do we remove the disclaimer?

